PR #23, opened 2024-03-19
change sample method to work for pandas

raisadz:


MarcoGorelli:
wonderful, thanks! fancy adding a little test which just makes an assertion on the length of the output if we use this?

raisadz:
I added `__invert__()` method in `narwhals/expression.py` and tests for `.is_null()` and `.sample()` functions. It looks like there are some issues with pandas indexing for `.sample()`  and `.drop_nulls()` methods. Also, `.sample()` returns an eager dataframe when the native dataframe is pandas and a lazy dataframe when the native is Polars and this results in the error `AttributeError: 'DataFrame' object has no attribute 'collect'` for `pandas` and `modin` inputs in test `test_expr_sample`.

MarcoGorelli (approved):

PR #34, opened 2024-03-30
More docstrings

frnmst:
Docstrings for:

- `narwhals.dataframe.DataFrame.unique`
- `narwhals.dataframe.DataFrame.sort`
- `narwhals.dataframe.DataFrame.join`

You can create a new base `docstrings` branch from your `main` before merging since I can't do it with my user

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -957,6 +957,143 @@ class DataFrame(BaseFrame):
         left_on: str | list[str],
         right_on: str | list[str],
     ) -> Self:
+        r"""
+        Join in SQL-like fashion.
+
+        Arguments:
+            other: DataFrame to join with.
+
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
nice one, thanks @frnmst !

I just kept join inner for now, as the others aren't (yet) supported - but I'll restore these example later when they are

frnmst:
> nice one, thanks @frnmst !
> 
> I just kept join inner for now, as the others aren't (yet) supported - but I'll restore these example later when they are

Ah, OK. I though the fact that was working as expected it meant that it was implemented.

But now I just noticed:

```python
        if how not in ["inner"]:                                                
            msg = "Only inner join supported for now, others coming soon"       
            raise ValueError(msg)
```

in `_pandas_like/dataframe.py`. But this exception wasn't triggered for those examples.

PR #37, opened 2024-04-01
Test: Trying to use hypothesis for Expr.is_null()

DeaMariaLeon:
There is an error I need to work on:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tests/test_expression.py:11: error: Untyped decorator makes function "test_isnull" untyped  [misc]
```
But feedback on the rest would be welcomed. There are 2 failures found by Hypothesis. I'm not sure if it's expected:

```
OverflowError: Python int too large to convert to C long
    | Falsifying example: test_isnull(
    |     integers=[0, 0, 9223372036854775808],
    |     datetimes=[datetime.datetime(2000, 1, 1, 0, 0),
    |      datetime.datetime(2000, 1, 1, 0, 0),
    |      datetime.datetime(2000, 1, 1, 0, 0)],
    |     floats=[0.0, 0.0, 0.0],
    |     text=['0', '0', '0'],
    | )
```
And:
```
assert  result_pd == result_pl
    | AssertionError: assert ┌────────────...─────────────┘ == ┌────────────...─────────────┘
    |   
    |   Full diff:
    |     ┌─────────────────────────────────────────────────┐
    |     | Narwhals DataFrame                              |
    |     | Use `narwhals.to_native()` to see native output |
    |     └─────────────────────────────────────────────────┘
    | Falsifying example: test_isnull(
    |     # The test always failed when commented parts were varied together.
    |     integers=[0, 0, 0],  # or any other generated value
    |     datetimes=[datetime.datetime(2000, 1, 1, 0, 0),
    |      datetime.datetime(2000, 1, 1, 0, 0),
    |      datetime.datetime(2000, 1, 1, 0, 0)],  # or any other generated value
    |     floats=[0.0, 0.0, 0.0],  # or any other generated value
    |     text=['0', '0', '0'],  # or any other generated value
    | )
```

It seems that the first error is related to: https://github.com/pola-rs/polars/issues/5031

Regarding the code, I was not able to have a separate function to generate the random data. I think I have to use the decorator. 
There is also a `pandas.DataFrame` strategy that could be used but then it would be missing for polars. 

Thanks for considering this PR. :-)


MarcoGorelli:
thanks! you can set min and max values for each strategy

i think is_null might not be the best place to start though because we don't have any nulls in your example - maybe some method which doesn't involve nulls?

DeaMariaLeon:
> you can set min and max values for each strategy

Sorry, is that to avoid the overflow error? 
I noticed that using UInt64 works. But shouldn't polars recognise that it needs that type (without us telling it)?
This works:
```
dfpl = pl.DataFrame({"integer": [9223372036854775808]}, schema= {"integer": pl.UInt64})
```
And if I set the maximum so it does not overflow, wouldn't Narwhals throw an error when someone tries to use a value this large? (but OK for polars?)..  

I'm just asking to understand, I'll do as you prefer of course.

MarcoGorelli:
It looks like the error happens in the Polars constructor, before even reaching Narwhals? In which case, I wouldn't worry about it tbh, it's not something we have control over here

DeaMariaLeon:
Added the mix and max values. 
The other previous error was due to zero values not being exactly the same (when using assert). 
I'm not sure I have to ping you... just in case: @MarcoGorelli.

MarcoGorelli (commented):
Nice, thanks @DeaMariaLeon ! Looks like you got the idea of hypothesis!

One comment would be `test_expression` - it might be good to make folder `tests/hypothesis` and to put such tests in there? This one could be `tests/hypothesis/test_basic_arithmetic` and we put some other functions in it?

I don't think we need to "go mad" with testing all arithmetic (by now, I hope both libraries have figured these out 😄 ) but I think it would be really good to test `join` and `concat`

Simple numeric columns are fine, something like:
- make two dataframes, both with 3 columns
- join on a random number of `left_on` columns
- same for `right_on`
- check that the output names and values match. Or, if one raises an error, check that the other one does too

It's tricky to think of all edge cases manually there so I think hypothesis might be really useful for this

MarcoGorelli (approved):
Nice one, thanks @DeaMariaLeon !

CI-nightly fails on PRs but passes when merged 🤔 I need to figure out how to resolve that 😄

PR #42, opened 2024-04-03
Docstrings DataFrame

frnmst:
This PR adds all the missing docstrings for the `narwhals.dataframe.DataFrame` class.

## Docstring problem

I had to use some tricks to let one of the `to_dict()` docstrings pass. I think there is a problem in the docstring "translation" process. Apparently tabs in the outputs are transformed to spaces before the output is checked by pytest. However, even setting the output to have spaces instead of tabs does not work. My workaround is to transform the tabs to spaces like this:

```python
p = pprint.pformat(df.to_dict(as_series=True)).replace('\t', '    ')
print(p)
```

instead of doing directly:

```python
df.to_dict(as_series=True)
```

The other solution would be just to skip the test with `# doctest: +SKIP`

## Error message

~~I changed the error message in the `__init__` because the class constructor should not accept Polars lazy frames.~~

Change reverted because of failing tests, but it seems sensible to me.

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -334,6 +334,58 @@ class DataFrame(BaseFrame):
             raise TypeError(msg)
 
     def to_pandas(self) -> Any:
+        r"""
+        Convert this DataFrame to a pandas DataFrame.
+
+        Returns:
+            A pandas DataFrame.
+
+        Notes:
+            This operation requires that `pandas` is installed.
+
+        Examples:
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> df_pl = pl.DataFrame(
+            ...     {
+            ...         "foo": [1, 2, 3],
+            ...         "bar": [6.0, 7.0, 8.0],
+            ...         "ham": ["a", "b", "c"],
+            ...     }
+            ... )
+            >>> df = nw.DataFrame(df_pl)
+            >>> df
+            ┌─────────────────────────────────────────────────┐
+            | Narwhals DataFrame                              |
+            | Use `narwhals.to_native()` to see native output |
+            └─────────────────────────────────────────────────┘
+            >>> df.to_pandas()
+               foo  bar ham
+            0    1  6.0   a
+            1    2  7.0   b
+            2    3  8.0   c
+
+            Null values in numeric columns are converted to `NaN`.
+
+            >>> df_pl = pl.DataFrame(
+            ...     {
+            ...         "foo": [1, 2, None],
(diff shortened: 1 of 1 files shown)


narwhals/dataframe.py
```diff
@@ -157,6 +157,150 @@ def join(
...
+        ╞═════╪═════╡
+        │ 1   ┆ 3   │
+        │ 2   ┆ 4   │
+        └─────┴─────┘
+
+        To specify a more detailed/specific frame schema you can supply the `schema`
+        parameter with a dictionary of (name,dtype) pairs...
+
+        >>> data = {"col1": [0, 2], "col2": [3, 7]}
+        >>> df_pl2 = pl.DataFrame(data, schema={"col1": pl.Float32, "col2": pl.Int64})
```
MarcoGorelli, line 197: this only works for Polars though, right? I don't think we should document it if it's Polars-specific

Same with the rest of the examples - they're showing how to create a Polars dataframe, which is then just passed to Narwhals. I think we can skip them, the initial simple example is enough

narwhals/dataframe.py
```diff
@@ -216,6 +443,99 @@ def __getitem__(self, col_name: str) -> Series:
...
+            └─────┴────────┴─────┴────────┴──────────┘
+            >>> import pprint
+            >>> pprint.pprint(df.to_dict(as_series=False))
+            {'A': [1, 2, 3, 4, 5],
+             'B': [5, 4, 3, 2, 1],
+             'cars': ['beetle', 'audi', 'beetle', 'beetle', 'beetle'],
+             'fruits': ['banana', 'banana', 'apple', 'apple', 'banana'],
+             'optional': [28, 300, None, 2, -30]}
+            >>> p = pprint.pformat(df.to_dict(as_series=True)).replace('\t', '    ')
+            >>> print(p)
```
MarcoGorelli, line 492: maybe we can just skip this example and only show the `as_series=False` one?

MarcoGorelli (commented):
Nice, thanks! Just got some requests to remove some Polars-specific things which aren't in Narwhals

narwhals/dataframe.py
frnmst: Yes, makes sense

narwhals/dataframe.py
frnmst: I change it to use `# doctest: +SKIP` so we can still have it in the docs.

MarcoGorelli (approved):
thanks @frnmst !

PR #43, opened 2024-04-04
Docstrings LazyFrame

frnmst:
I added the missing docstrings for the `narwhals.dataframe.LazyFrame` class.

The pre-commit error is for a single word in a docstring output. Codespell thinks it's a typo. I tried adding an inline ignore but it doesn't work.

About the other error in the CI, I think the docstring with the lambda operator is not supported in Python 3.8.

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -1316,9 +1316,160 @@ class LazyFrame(BaseFrame):
         return super().drop(*columns)
 
     def unique(self, subset: str | list[str]) -> Self:
+        """
+        Drop duplicate rows from this LazyFrame.
+
+        Arguments:
+            subset: Column name(s) to consider when identifying duplicate rows.
+                     If set to `None`, use all columns.
+
+        Returns:
+            LazyFrame: LazyFrame with unique rows.
+
+        Examples:
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> lf_pl = pl.LazyFrame(
+            ...     {
+            ...         "foo": [1, 2, 3, 1],
+            ...         "bar": ["a", "a", "a", "a"],
+            ...         "ham": ["b", "b", "b", "b"],
+            ...     }
+            ... )
+            >>> lf = nw.LazyFrame(lf_pl)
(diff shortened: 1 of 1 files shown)


narwhals/dataframe.py
```diff
@@ -1054,21 +1111,429 @@ def select(
...
+            shape: (3, 3)
+            ┌─────┬─────┬─────┐
+            │ coo ┆ car ┆ cam │
+            │ --- ┆ --- ┆ --- │
+            │ i64 ┆ i64 ┆ str │
+            ╞═════╪═════╪═════╡
+            │ 1   ┆ 6   ┆ a   │
+            │ 2   ┆ 7   ┆ b   │
+            │ 3   ┆ 8   ┆ c   │
+            └─────┴─────┴─────┘
```
MarcoGorelli, line 1171: let's skip this one if lambda isn't supported for all versions (it's not documented here anyway)

narwhals/dataframe.py
```diff
@@ -1054,21 +1111,429 @@ def select(
...
+
+        Arguments:
+            n: Number of rows to return.
+
+        Notes:
+            Consider using the `fetch` operation if you only want to test
+             your query. The `fetch` operation will load the first `n`
+             rows at the scan level, whereas the `head`/`limit` are
+             applied at the end.
+
```
MarcoGorelli, line 1187: we don't have this here (yet), let's remove this for now

narwhals/dataframe.py
```diff
@@ -1054,21 +1111,429 @@ def select(
...
+            shape: (3, 1)
+            ┌─────┐
+            │ ham │
+            │ --- │
+            │ str │
+            ╞═════╡
+            │ a   │
+            │ b   │
+            │ c   │
+            └─────┘
```
MarcoGorelli, line 1294: selectors aren't supported here yet, let's remove this one

MarcoGorelli (commented):
Nice one, well done @frnmst ! Just got a couple of minor comments

frnmst:
I removed the unnecessary stuff

MarcoGorelli (approved):
thanks @frnmst !

PR #49, opened 2024-04-12
Hypothesis testing DataFrame.join

DeaMariaLeon:
Repeating the data inside each frame, if not, `join` can't find data to link. Same for columns to join, they have to be the same left and right. Unless I'm missing something of course. 

Feedback very welcome.


MarcoGorelli (approved):
cool, thanks! nothing unexpected here then, that's nice

maybe as a follow-up it would be interesting to try injecting some null values?

PR #50, opened 2024-04-14
Add nulls & inf to hypothesis test for DataFrame.join

DeaMariaLeon:
Feedback welcome.

### DIFF
--- a/tests/hypothesis/test_join.py
+++ b/tests/hypothesis/test_join.py
@@ -63,3 +63,53 @@ def test_join(
     )
 
     assert_frame_equal(dframe_pd1, dframe_pd2)
+
+
+@given(
+    st.lists(
+        st.floats(allow_nan=True),
+        min_size=3,
+        max_size=3,
+    ),
+    st.lists(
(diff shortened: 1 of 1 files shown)


DeaMariaLeon:
Infinity is not included on what we previously had, because there are the min and max values. https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.floats

I'll get back to you for the rest. :)

MarcoGorelli:
ah thanks for explaining - maybe we could amend the previous example to have just
```
st.floats()
```
then? as far as I understood, the reason for the `min_value` and `max_value` where because of UInt64, but floats are fine?

DeaMariaLeon:
There had been an error also with the Floats64 too, so I added the max/min values for those as well.
But I checked and it seems to be working. 🤔

Ok, so I'll set them to`st.floats()`.

MarcoGorelli:
>  I checked and it seems to be working

love it when things "fix themselves" 😄

MarcoGorelli (approved):

PR #54, opened 2024-04-21
Bumped polars version on extremes.yml due to polars old bug

DeaMariaLeon:
Previous polars versions were failing when using `join` on a column with a `float=-0.0`.

Added specific example to make sure it is always tested.


MarcoGorelli (approved):
yay thanks @DeaMariaLeon

PR #55, opened 2024-04-22
Adding error message when different libraries are used

DeaMariaLeon:
Trying to close https://github.com/MarcoGorelli/narwhals/issues/46

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -134,7 +134,7 @@ class BaseFrame:
     ) -> Self:
         if how != "inner":
             raise NotImplementedError("Only inner joins are supported for now")
-        validate_same_library(self, other)
+        validate_same_library([self, other])
         return self._from_dataframe(
             self._dataframe.join(
                 self._extract_native(other),
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -134,8 +134,8 @@ class BaseFrame:
             raise NotImplementedError("Only inner joins are supported for now")
 
         if self._is_polars != other._is_polars or (
-         (hasattr(self, "implementation") and hasattr(other, "implementation")) and (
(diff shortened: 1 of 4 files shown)


MarcoGorelli (commented):
Ooh, nice, thanks!

I think the sample applies for `concat` though, right? Could we put this in `narwhals/utils.py` and then use it in both concat and join?

DeaMariaLeon:
I'm not sure if I should ping you @MarcoGorelli - no rush or anything.

narwhals/dataframe.py
```diff
@@ -132,6 +133,7 @@ def join(
     ) -> Self:
         if how != "inner":
             raise NotImplementedError("Only inner joins are supported for now")
+        validate_same_library(self, other)
```
MarcoGorelli, line 136: rather than making two paths inside `validate_same_library`, could we just call this one with
```python
validate_same_library([self, other])
```
?

DeaMariaLeon: Didn't think about it! 🤓

narwhals/utils.py
```diff
@@ -42,3 +42,13 @@ def parse_version(version: Sequence[str | int]) -> tuple[int, ...]:
 
 def isinstance_or_issubclass(obj: Any, cls: Any) -> bool:
     return isinstance(obj, cls) or issubclass(obj, cls)
+
+
+def validate_same_library(one_object: Any) -> None:
```
MarcoGorelli, line 47: maybe this could/should be a bit more specific? `items: Iterable[Any]`?

MarcoGorelli (approved):
nice! just got a minor comment then this can go 🚀

narwhals/utils.py
DeaMariaLeon: Done! :)

PR #72, opened 2024-05-02
Docstring for Cast

MatthewMec:
Created Docstring for cast.


MarcoGorelli:
that was fast, thanks!

I think you've staged too many files though, the diff shows 37 files changed. It should just show 1 file changed (i.e. `narwhals/expression.py`)

PR #75, opened 2024-05-02
Docstring cast

MatthewMec:
Added docstring for the cast function in narwhals.

### DIFF
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -454,9 +454,77 @@ class Expr:
         return self.__class__(lambda plx: self._call(plx).max())
 
     def n_unique(self) -> Expr:
+        """
+         Returns count of unique values
+
+        Examples:
+            >>> import polars as pl
+            >>> import pandas as pd
+            >>> import narwhals as nw
+            >>> df_pd = pd.DataFrame({'a': [1, 2, 3, 4, 5], 'b': [1, 1, 3, 3, 5]})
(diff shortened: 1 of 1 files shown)


narwhals/expression.py
```diff
@@ -74,6 +74,59 @@ def cast(
...
+                >>> "bar": [6.0, 7.0, 8.0],
+                >>> "ham": [date(2020, 1, 2), date(2021, 3, 4), date(2022, 5, 6)],})
+
+                Let's make it data frame agnostic:
+
+                >>> df = nw.from_native(df_pd)
+
+                Cast specific frame columns to the specified dtypes:
+
+                >>> df = df.cast({"foo": pl.Float32, "bar": pl.UInt8})
```
MarcoGorelli, line 102: close 😉  we need to use `Expr.cast` here, not `DataFrame.cast`

something like
```
df = df.select(nw.col('foo').cast(nw.Float32), nw.col('bar').cast(nw.UInt8))
```

MarcoGorelli (changes requested):
nice, this is close!

can you also run
```
pre-commit install
pre-commit run --all-files
```
?

MarcoGorelli (changes requested):
getting closer 💪 can you run `pytest narwhals --doctest-modules` and check it passes please? The code examples will need fixing up a bit

PR #78, opened 2024-05-03
Added Expr.max() docstring

brentomagic:


MarcoGorelli (changes requested):
this is close, thanks! can you run
```
pytest narwhals --doctest-modules
```
 and check it passes?

narwhals/expression.py
```diff
@@ -286,6 +286,38 @@ def min(self) -> Expr:
...
+        Returns the maximum value(s) from a column(s).
+
+        Examples:
+            >>> import polars as pl
+            >>> import pandas as pd
+            >>> import narwhals as nw
+            >>> df_pd = pd.DataFrame({'a': [10, 20], 'b': [50, 100]})
+            >>> df_pl = pl.DataFrame({'a': [10, 20], 'b': [50, 100]})
+
+            let's define a dataframe-agnostic function:
```
MarcoGorelli, line 299:
```suggestion
            Let's define a dataframe-agnostic function:
```

narwhals/expression.py
```diff
@@ -286,6 +286,38 @@ def min(self) -> Expr:
...
+
+            >>> def func(df_any):
+            ...    df = nw.from_native(df_any)
+            ...    df = df.select(nw.max('a', 'b'))
+            ...    return nw.to_native(df)
+
+            We can then pass either pandas or polars to `func`:
```
MarcoGorelli, line 306:
```suggestion
            We can then pass either pandas or Polars to `func`:
```

MarcoGorelli (approved):
thanks @brentomagic !

PR #80, opened 2024-05-04
std usage example added

anopsy:
I re-used the one sentence description from polars.Expr.std,
followed the style of examples that are already in the docs (max(), any())
I've chosen for df's/values that were somewhere halfway between the already present examples and the ones from polars examples, so I'm curious if they're okay


MarcoGorelli (approved):
legend, thanks @anopsy !

as a follow-up, if you're interested, we may want to add a `ddof` argument?

anopsy:
let's do that!

PR #82, opened 2024-05-05
mean usage example added

anopsy:
it worked - thank you!

### DIFF
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -274,6 +274,38 @@ class Expr:
         return self.__class__(lambda plx: self._call(plx).all())
 
     def mean(self) -> Expr:
+        """
+        Get mean value.
+
+        Examples:
+            >>> import polars as pl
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
nice, thanks @anopsy !

just fyi, in the future, it should work to push to the same branch (this will save you from having to open a new pull request)

anopsy:
you mean that I didn't have to create a new branch after rebase? or like I should in general push all the docstrings (the ones i've created and the ones I will create in the future) to one branch?

MarcoGorelli:
> you mean that I didn't have to create a new branch after rebase?

exactly, you can just push to the same one

> should in general push all the docstrings (the ones i've created and the ones I will create in the future) to one branch?

I'd suggest making a separate branch for each feature (this enables you to work on multiple things at the same time!), but if you've already created a pull request, then when you update it, you can just push to the same branch

PR #87, opened 2024-05-05
Added docstring for Expr.is_null

ugohuche:


narwhals/expression.py
```diff
@@ -410,6 +410,56 @@ def filter(self, other: Any) -> Expr:
...
+            ...             'a': [2, 4, None, 3, 5],
+            ...             'b': [2.0, 4.0, float("nan"), 3.0, 5.0]
+            ...         }
+            ... )
+
+            Let's define a dataframe-agnostic function:
+
+            >>> def func(df_any):
+            ...     df = nw.from_native(df_any)
+            ...     df = df.with_columns(nw.all().is_null()) # nan != null for polars
```
MarcoGorelli, line 437: how about showing:
```python
...     df = df.with_columns(
...         a_is_null = nw.col('a').is_null(),
...         b_is_null = nw.col('b').is_null(),
...     )
```
, so that the output is easier to reason about?

MarcoGorelli (commented):
Nice, thanks @ugohuche ! Just got a suggestion

narwhals/expression.py
ugohuche: Yeah, that makes more sense. I'll add it in the next pull request

MarcoGorelli: thanks! can you add it to this one please? if you make your change, add and commit them, and push them to this branch, then this pull request will be automatically updated

ugohuche: Okay, I've updated the branch now

MarcoGorelli (approved):
awesome work @ugohuche !

PR #88, opened 2024-05-05
Added ddof arg to std

anopsy:
This one is spicy, I'm very curious if I did it right

### DIFF
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -151,8 +151,8 @@ class PandasExpr:
     def mean(self) -> Self:
         return register_expression_call(self, "mean")
 
-    def std(self) -> Self:
-        return register_expression_call(self, "std")
+    def std(self, *, ddof: int = 1) -> Self:
+        return register_expression_call(self, "std", ddof=ddof)
 
     def any(self) -> Self:
         return register_expression_call(self, "any")
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -321,10 +321,10 @@ class PandasSeries:
     def std(
         self,
         *,
-        correction: float = 1.0,
+        ddof: int = 1,
     ) -> Any:
         ser = self._series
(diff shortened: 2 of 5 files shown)


anopsy:
Okay, I can see what's the problem, I need to check also why it passed the tests

MarcoGorelli:
Wow, nice! I'm impressed you figured it all out independently, well done! 

I think we need a unit test for this, in which we pass different values of ddof. Fancy trying to add one? Either in test_common.py, or in a new test file

anopsy:
geeeeez, that was tough, I'm still wondering what that "correction: float =1.0" parameter was. 
so happy it works

> I think we need a unit test for this, in which we pass different values of ddof. Fancy trying to add one? Either in test_common.py, or in a new test file

sure thing!

MarcoGorelli:
😆 I'd accidentally copied over `correction` from the array API, glad you figured out it was wrong

anopsy:
really curious why the pytest-coverage fails, I've run the pytest --cov==narwhals and it didn't complain:
image


Also: I kind of feel that writing the test with nw.all would be more elegant.

MarcoGorelli:
that was fast, cool - you definitely understated your skills 😄 🙌 

for the test, I think it's just that your branch is out-of-sync with main - can you run
```
git fetch upstream
git merge upstream/main
git push origin HEAD
```
?

You can also remove `if df_any is None: return`, it's no longer needed (it was just a "trick" I'd introduced to skip running the pyarrow-backed dataframe for CI runs when the pandas version was too old to support them)

anopsy:
image

This is what I get, should I then git pull? (as far as I remember it will ask me about rebasing and I have no idea how to proceed)

thank you, I could definitely improve my git game

MarcoGorelli:
> Also: I kind of feel that writing the test with nw.all would be more elegant, let me know what do you think

maybe, but what you've done is good, I think it's ok to merge as-is

MarcoGorelli (approved):
nice one @anopsy !

PR #93, opened 2024-05-06
Added sum() docstring

brentomagic:


narwhals/expression.py
```diff
@@ -349,6 +349,38 @@ def std(self, *, ddof: int = 1) -> Expr:
         return self.__class__(lambda plx: self._call(plx).std(ddof=ddof))
 
     def sum(self) -> Expr:
+        """
+        Return the sum value(s).
```
MarcoGorelli, line 353:
```suggestion
        Return the sum value.
```

MarcoGorelli: (it's only one value per Expr)

MarcoGorelli (approved):
thanks @brentomagic !

Issue #71, opened 2024-05-01
Fill in missing docstrings


MarcoGorelli:
# Add missing docstrings

Let's start with all the functions listed in https://narwhals-dev.github.io/narwhals/api-reference/expressions/. Feel free to divide the work up amongst yourselfs as you see fit. We need to gradually make sure they all have docstrings

## Steps

1. Pick a function you'd like to add a docstring for. For example, `Expr.alias`.
2. Make a new branch
3. Check the Polars docs to see what the function does: https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.alias.html
4. If applicable, re-use the same one-sentence summary and argument descriptions. Narwhals only implements a subset of Polars, so the Polars docs may have some extra bits which Narwhals doesn't support. If that's the case, leave the extra bits out
5. Add a one-sentence summary, and document the supported arguments
6. Make sure to include an example! More details below
7. run `mkdocs serve`, and open `http://127.0.0.1:8000/` (or whichever URL it gives you) in the browser. Find the example you added, and make sure it looks good
8. run `pytest narwhals --doctest-modules`, make sure it passes
9. `git add <files you've modified>`, `git commit -m '<you message goes here>'`, `git push origin HEAD`
10. open pull request

See https://github.com/narwhals-dev/narwhals/pull/74 for an example pull request

## Examples requirements

The examples should follow that pattern:
- import pandas, Polars, and Narwhals
- construct a `df_pd` pandas dataframe and a `df_pl` Polars dataframe
- write a dataframe-agnostic function `func` which uses the function you're writing a docstring for
- show the output when you pass either a pandas or a Polars dataframe

## Help

If you're stuck, feel free to ask here or on Discord. If you've got most of the way but are unsure about something, I'd suggest opening a draft pull request anyway - that's usually the fastest way to get help

## Contributing

See the [contributing guide](https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md) for how to get started

hathawayj:
I will tackle `drop_nulls()`

MatthewMec:
I will tackle csts(dftype).

ugohuche:
I and Matthew would work on the first three - ```alias()```,  ```all()```, and ```any()```

brentomagic:
i'll look into max().

ugohuche:
Next I'll work on ```drop_nulls()``` to ```is_null()```

anopsy:
[...]

PR #97, opened 2024-05-07
Add docstring for Expr.is_between()

shriyakalakata:
Towards #71

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -142,7 +142,17 @@ class PandasSeries:
         self, lower_bound: Any, upper_bound: Any, closed: str = "both"
     ) -> PandasSeries:
         ser = self._series
-        return self._from_series(ser.between(lower_bound, upper_bound, inclusive=closed))
+        if closed == "left":
+            res = ser.ge(lower_bound) & ser.lt(upper_bound)
+        elif closed == "right":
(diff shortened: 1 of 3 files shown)


MarcoGorelli (commented):
Thanks @shriyakalakata !

Looks like this one is going to be slightly harder, because the pandas API changed in version 1.3.0. But, it's a good learning opportunity :)

Instead of using [pandas.Series.between](https://pandas.pydata.org/docs/reference/api/pandas.Series.between.html), maybe we should use `pandas.Series.gt`, `pandas.Series.ge`, `pandas.Series.lt`, and `pandas.Series.le`?

So, concretely, in `narwhals._pandas_like.series.py`, we remove

https://github.com/narwhals-dev/narwhals/blob/a9995b5dc28764afd504b5f18525275749b7b5d5/narwhals/_pandas_like/series.py#L141-L145

and put something like

```
        ser = self._series
        # pandas API changed in 1.3.0, so we can't just use Series.between
        if closed == 'left':
            return self._from_series(ser.ge(lower_bound) & ser.lt(upper_bound))
        # fill in the cases for `closed=='right'`, `closed=='both'`, `closed=='none'`
        msg = f"Expected {'left', 'right', 'both', 'none'}, got {closed}"
        raise TypeError(msg)
```

Fancy trying this, and filling in the gaps? I think this should work across pandas versions

MarcoGorelli (approved):

PR #102, opened 2024-05-07
series drop nulls documentation

hathawayj:
I moved over to the series methods and tried the documentation for `drop_nulls()`. Followed directions in https://github.com/narwhals-dev/narwhals/issues/71.

### DIFF
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -138,32 +138,33 @@ class Series:
           >>> import polars as pl
           >>> import numpy as np
           >>> import narwhals as nw
-          >>> df_pd = pd.DataFrame({'a': [1, None, 2], 'b': [4, 5, np.nan]})
-          >>> df_pl = pl.DataFrame({'a': [1, None, 2, ], 'b': [4, 5, float("nan")]})
+          >>> s_pd = pd.Series([2, 4, None, 3, 5])
+          >>> s_pl = pl.Series('a', [2, 4, None, 3, 5])
 
           Now define a dataframe-agnostic function with a `column` argument for the column to evaluate :
 
-          >>> def func(df_any, column):
-          ...   df = nw.from_native(df_any)
-          ...   df = df.select(nw.col(column).drop_nulls())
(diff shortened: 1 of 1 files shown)


MarcoGorelli (commented):
That was fast - nice one!

This is close - however, `nw.col(column).drop_nulls()` is actually `Expr.drop_nulls`, not `Series.drop_nulls`. `nw.col(column)` is an `Expr`

It should be a simple change though:
- instead of making `df_pd` and `df_pl`, you can make `s_pd` and `s_pl`
- when you call `nw.from_native`, pass `series_only=True`

So, your function will look something like
```python
def func(s_any):
    s = nw.from_native(s_any, series_only=True)
    s = s.drop_nulls()
    return nw.to_native(s)
```

hathawayj:
Yeah. I was wondering about that.  I will update.

narwhals/series.py
```diff
@@ -123,6 +123,49 @@ def is_in(self, other: Any) -> Self:
...
+
+          Then we can pass either Series (polars or pandas) to `func`:
+
+          >>> func(s_pd)
+          0    2.0
+          1    4.0
+          3    3.0
+          4    5.0
+          dtype: float64
+          >>> func(s_pl)
```
MarcoGorelli, line 159:
```suggestion
          >>> func(s_pl)  # doctest:+SKIP
```

MarcoGorelli (approved):
thanks @hathawayj for your PR!

I came across this today too, and can't tell what the cause it. The docstring validation works fine for `dataframe.py` and `expression.py`, but not for `series.py`

For now, let's just add a `  # doctest: +SKIP` for the Polars one, but hopefully we can figure out why this is happening and remove it later

MarcoGorelli:
> For now, let's just add a   # doctest: +SKIP for the Polars one, but hopefully we can figure out why this is happening and remove it later

Found it - the Polars output uses tabs instead of spaces (!)

So I've set `# doctest: +NORMALIZE_WHITESPACE` for the Polars Series outputs

PR #114, opened 2024-05-08
Added docstring for Expr.drop_nulls

ugohuche:


narwhals/expression.py
```diff
@@ -721,6 +721,55 @@ def fill_null(self, value: Any) -> Expr:
...
+
+        Examples:
+            >>> import narwhals as nw
+            >>> import pandas as pd
+            >>> import polars as pl
+
+            >>> data = {
+            ...             "a": [2, 4, None, 3, 5],
+            ...             "b": [2.0, 4.0, float("nan"), 3.0, 5.0]
+            ...         }
```
MarcoGorelli, line 739: instead of making two columns and not using one of them, how about making column `'b'` have both` float('nan')` and `None`?

ugohuche: Alright

MarcoGorelli (approved):
look great, thanks @ugohuche !

PR #117, opened 2024-05-08
Added docstring for Series.max()

ugohuche:


narwhals/series.py
```diff
@@ -111,6 +111,32 @@ def min(self) -> Any:
...
+            We define a data-frame agnostic function:
+
+            >>> def func(s_any):
+            ...     s = nw.from_native(s_any, series_only=True)
+            ...     s = s.max()
+            ...     return nw.to_native(s)
+
+            We can then pass either pandas or Polars to `func`:
+
+            >>> func(s_pd) # doctest:+SKIP
```
MarcoGorelli, line 134: why do we need to skip this?

MarcoGorelli: actually, `s.max()` should just return a scalar

So for this one, you can skip the `nw.to_native(s)` part and just do
```
return s.max()
```

MarcoGorelli (approved):
that's the one, thanks @ugohuche !

PR #129, opened 2024-05-10
Update README.md

koaning:
This was probably an old url before the project moved to its own Github org.


MarcoGorelli (approved):
thanks @koaning !

PR #134, opened 2024-05-10
feat: narwhalify decorator

### LINKED ISSUE #133: Add decorator for end to end steps
# Description

Inspired by Vincent stream and being a fan of decorators, it could be possible to wrap functions that input and output a dataframe.

(Somewhat) reasonable implementation could be:

**decorator.py**
```py 

from functools import partial
from functools import wraps
from typing import Any
from typing import Callable
from typing import Concatenate
from typing import ParamSpec, TypeVar

from narwhals.dataframe import DataFrame
from narwhals.translate import from_native
from narwhals.translate import to_native

NativeDataFrame = TypeVar("NativeDataFrame")
NativeSeries = TypeVar("NativeSeries")

T = DataFrame | LazyFrame | Series
N = NativeDataFrame | NativeSeries
PS = ParamSpec("PS")


def find_me_a_good_name(
    func: Callable[Concatenate[T, PS], T] | None,
    from_kwargs: dict[str, Any] = {
        "strict": True,
        "eager_only": None,
        "series_only": None,
        "allow_series": None,
    },  # Ruff complains about `mutable data structures for argument defaults`
    to_kwargs: dict[str, Any] = {
        "strict": True,
    }, # Ruff complains about `mutable data structures for argument defaults`
) -> Callable[Concatenate[N, PS], N]:
    if func is None:
        return partial(find_me_a_good_name, from_kwargs=from_kwargs, to_kwargs=to_kwargs)

    @wraps(func)
    def wrapper(
        frame: NativeDataFrame, *args: PS.args, **kwargs: PS.kwargs
    ) -> NativeDataFrame:
        nw_frame = from_native(frame, **from_kwargs).pipe(func, *args, **kwargs)
        return to_native(nw_frame, **to_kwargs)

    return wrapper
```

Basic usage example:

```py
import narwhals as nw
from narwhals.decorator import find_me_a_good_name

@find_me_a_good_name
def func(df_any):
    return df_any.select(
        a_sum=nw.col('a').sum(),
        a_mean=nw.col('a').mean(),
        a_std=nw.col('a').std(),
    )

import pandas as pd
import polars as pl

df = pd.DataFrame({'a': [1, 1, 2]})
print(func(df))

df = pl.DataFrame({'a': [1, 1, 2]})
print(func(df))
```

What do you think?

(Edit: Type hinting would require some more love)

FBruzzesi:
# Description

Fix #133 

It fails mypy, but I spent more time trying to annotate the module and the tests than anything else, so I will take any help at this point 😂

### DIFF
--- a/docs/api-reference/narwhalify.md
+++ b/docs/api-reference/narwhalify.md
@@ -1,0 +1,7 @@
+# `narwhals.narwhalify`
+
+::: narwhals.decorator.narwhalify
+    handler: python
+    options:
+      show_source: false
+      show_bases: false
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -25,6 +25,7 @@ nav:
     - api-reference/expressions.md
     - api-reference/expressions_dt.md
     - api-reference/expressions_str.md
+    - api-reference/narwhalify.md
     - api-reference/dtypes.md
     - api-reference/dependencies.md
 theme:
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -1,5 +1,6 @@
 from narwhals.dataframe import DataFrame
 from narwhals.dataframe import LazyFrame
+from narwhals.decorator import narwhalify
 from narwhals.dtypes import Boolean
 from narwhals.dtypes import Datetime
 from narwhals.dtypes import Float32
@@ -58,4 +59,5 @@ __all__ = [
     "Boolean",
     "String",
     "Datetime",
+    "narwhalify",
 ]
--- a/narwhals/decorator.py
+++ b/narwhals/decorator.py
@@ -1,0 +1,123 @@
+from __future__ import annotations
+
+from functools import partial
+from functools import wraps
+from typing import Any
+from typing import Callable
+from typing import Concatenate
+from typing import ParamSpec
+from typing import TypeVar
+
+from narwhals.dataframe import DataFrame
+from narwhals.dataframe import LazyFrame
+from narwhals.series import Series
+from narwhals.translate import from_native
+from narwhals.translate import to_native
+
+T = DataFrame | LazyFrame | Series
+PS = ParamSpec("PS")
+AnyFrame = TypeVar("AnyFrame")
+
+
+def narwhalify(
+    func: Callable[Concatenate[tuple[T], PS], T] | None = None,
+    from_kwargs: dict[str, Any] | None = None,
+    to_kwargs: dict[str, Any] | None = None,
+) -> (
+    Callable[Concatenate[tuple[AnyFrame], PS], AnyFrame]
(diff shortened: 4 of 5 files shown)


FBruzzesi:
@MarcoGorelli this is the closest I got to be mypy compliant on my local machine😕

narwhals/narwhalify.py
```diff
@@ -0,0 +1,123 @@
...
+    dataframe/series-like is returned from it.
+
+    Warning:
+        `func` can take an arbitrary number of dataframe/series objects which will be converted using `from_native`.
+        However, as these arguments are parsed as positional, make sure to:
+
+        - Pass them as the first positional arguments.
+        - Explicitly pass any other argument as keyword.
+
+        Check examples section below to see an instance of that.
```
MarcoGorelli, line 38: this seems a little dangerous, are sure people will use it correctly?

maybe the decorator could just deal with the simplest case of:
- single frame input
- *args
- **kwargs

?

FBruzzesi: I don't know why but I didn't get notified on this comment.
Yes that was the original _fairly_ simple use case mentioned in the [issue itself](#133).

I am ok with having the simple case rollout first

MarcoGorelli: yup, sorry for having initially suggested to make it more complicated 🙈

FBruzzesi:
Older python versions seem to not update wrapper name even when using [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps)? That's new to me 😕

> Without the use of this decorator factory, the name of the example function would have been 'wrapper'

MarcoGorelli:
I'm not super-familiar with decorators, but if it's too complex to do like this, maybe we can just make
- `@narwhalify_frame` (default kwargs)
- `@narwhalify_eager_frame` (from kwargs have `eager_only=True`)
- `@narwhalify_series` (from kwargs have `series_only=True`)
?

Then, consumers would just decorate with `@narwhalify_frame`, without passing in the `from_kwargs` / `to_kwargs`?

FBruzzesi:
The way to implement case specific decorators would be to have one and then use `partial(narwhalify, ...)` on it.
Otherwise the codebase would easily get duplicated and unmaintainable for any change.

I will take a look on why for py3.{8,9} `functools.wraps` seems to not update the wrapper name

FBruzzesi:
@MarcoGorelli an alternative is to unpack args as follows:

```py
...
    @wraps(func)
    def wrapper(*args: PS.args, **kwargs: PS.kwargs) -> Any:
        
        frame, *args = args
        nw_frame = from_native(frame, **from_kwargs)
        result = func(nw_frame, *args, **kwargs)
        return to_native(result, **to_kwargs)
return wrapper
```
Then keyword only arguments would lead to 

> ValueError: not enough values to unpack (expected at least 1, got 0)

but we can customize it.

WDYT?

PR #135, opened 2024-05-10
Added millisecond and microsecond temporal methods

ugohuche:


MarcoGorelli (commented):
Nice effort here, thanks!

A few comments:
- if we're adding a new method, let's include a docstring in the PR straight away (rather than doing it separately). I'm aware that I didn't do this at the beginning, but going forwards, let's make sure that every new development has a docstring
- you'll also need to add a test for this new functionality. Please check `tests/test_dt.py`

ugohuche:
I tried adding docstrings but passing a pandas dataframe to 'func' in the example was giving issues that I couldn't resolve. I wanted to get this right first then I'll try resolving that, but definitely I'll add docstrings to subsequent ones.

MarcoGorelli:
> was giving issues that I couldn't resolve

could you show the issues please?

ugohuche:
It kept throwing this AttributeError: 'DatetimeProperties' object has no attribute 'millisecond'

MarcoGorelli:
right - you'll have to find another way to implement
```python
    def millisecond(self) -> PandasSeries:
        return self._series._from_series(
            self._series._series.dt.millisecond,
        )
```
then

MarcoGorelli (commented):
getting there 💪 

could you add docstrings + tests please?

MatthewMec (approved):

MarcoGorelli (approved):
wow, nice! well done working through all this

PR #136, opened 2024-05-10
Added docstring for Expr.dt.second and Expr.dt.minute

ugohuche:


MarcoGorelli (approved):
nice examples, thanks @ugohuche !

PR #166, opened 2024-05-13
DOC: fill in missing docstrings for `Expr.sample`

natmokval:
xref #71


MarcoGorelli (approved):
looks good, thanks - and good to skip such a non-deterministic doctest :)

PR #172, opened 2024-05-14
Series.minute

brentomagic:


DeaMariaLeon:
Thanks @brentomagic ! I noticed that the docstrings of your previous PR `series.hour` is included here as well.
Maybe the earlier one should be closed to avoid the overlap? 
Also, the one line description should match the one in Polars?

MarcoGorelli:
Yup - I think you do
```
git fetch upstream
git rebase -i upstream/main
```
and then drop the commit you don't want (the one from the previous PR), and then
```
git push origin HEAD -f
```
then it should work

PR #174, opened 2024-05-15
DOC: fill in missing docstrings for `Expr.sort`

natmokval:
xref #71

I am unsure if we can call 'sort' on groups to sort them.
For example:
```
df = pl.DataFrame(
    {
        "group": ["one", "one", "one", "two", "two", "two"],
        "value": [1, 98, 2, 3, 99, 4],
    }
)

df.group_by("group").agg(pl.col("value").sort()) 
```


narwhals/expression.py
```diff
@@ -679,6 +679,72 @@ def shift(self, n: int) -> Expr:
...
+            ┌──────┐
+            │ a    │
+            │ ---  │
+            │ i64  │
+            ╞══════╡
+            │ null │
+            │ 1    │
+            │ 2    │
+            │ 5    │
+            └──────┘
```
MarcoGorelli, line 727: thanks for working on this

this doesn't look quite right, looks like the results differ - maybe `na_position='first'` needs passing to pandas? fancy trying that?

narwhals/expression.py
```diff
@@ -679,6 +679,78 @@ def shift(self, n: int) -> Expr:
...
+            ...         return nw.to_native(df).sort_values(by='a', na_position='first')
+            ...     return nw.to_native(df)
+
+            >>> def func_descend(df_any):
+            ...     df = nw.from_native(df_any)
+            ...     df = df.select(nw.col('a').sort(descending=True))
+            ...     if isinstance(nw.to_native(df), pd.DataFrame):
+            ...         return nw.to_native(df).sort_values(
+            ...             by='a', ascending=False, na_position='first'
+            ...         )
```
MarcoGorelli, line 711: wait we can't do this 😄 we need to correct the definition itself, here:

https://github.com/narwhals-dev/narwhals/blob/1d2e387b2359359725d256c84ae19394d02c5184/narwhals/_pandas_like/series.py#L400-L408

MarcoGorelli (approved):
great, thanks @natmokval !

narwhals/expression.py
natmokval: Oops, now I got it. I corrected my mistake and updated the PR.

PR #177, opened 2024-05-16
added to_pandas docstring

brentomagic:


DeaMariaLeon:
Thank you @brentomagic for this. 
Just a question, do we need the part to convert from pandas to pandas? DataFrame.to_pandas() example doesn't have it. I would think that we don't need it here.. ?

MarcoGorelli (approved):
thanks @brentomagic  for your PR, and @DeaMariaLeon for your review

I think this is OK as-is, like this we show that `to_pandas`, when starting from pandas, is a no-op

Some of the DataFrame docstrings are a bit old and don't follow the "create pandas and Polars, write dataframe-agnostic function, show that you can pass both" template - we should probably update them

PR #185, opened 2024-05-17
DOC: fill in missing docstrings for Series.all and Series.sort

natmokval:
xref #71

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -317,6 +317,9 @@ class PandasSeries:
         return ser.any()
 
     def all(self) -> Any:
+        """
+        Return whether all values in the column are True.
+        """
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
Thanks! Though `_pandas_like` is private and doesn't show up in the docs, I'd suggest we instead focus on making sure that everything public has a docstring (so, the files directly under the narwhals folder)

natmokval:
> Thanks! Though _pandas_like is private and doesn't show up in the docs, I'd suggest we instead focus on making sure that everything public has a docstring (so, the files directly under the narwhals folder)
>

thank you, it's my mistake. I moved docstrings from private `_pandas_like` to public `series` and added examples.

MarcoGorelli (approved):
thanks @natmokval !

PR #186, opened 2024-05-17
feat: `Expr.name` namespace

### LINKED ISSUE #130: Add `.name.suffix`
From a request  [here](https://github.com/koaning/scikit-lego/pull/661#discussion_r1596527264)

FBruzzesi:
# Description

Introduces `Expr.name` namespace and 5 methods within it (`.map`, `.suffix`, `.prefix`, `.to_lowercase`, and `to_uppercase` (Fix #130).
Adds `starts_with` withing the `Expr.str` namespace.

## Current issues

If I am understanding correctly from polars, name namespace should address the _root_ name. 
As I was to implement these from scratch as an Expr, I am unsure if there are uncovered edge cases with the current implementation. ~~Once this is clarified I will add checks for the `Anonymous expressions` case.~~

### DIFF
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -261,6 +261,10 @@ class PandasExpr:
     def dt(self) -> PandasExprDateTimeNamespace:
         return PandasExprDateTimeNamespace(self)
 
+    @property
+    def name(self) -> PandasExprNameNamespace:
+        return PandasExprNameNamespace(self)
+
 
 class PandasExprStringNamespace:
     def __init__(self, expr: PandasExpr) -> None:
@@ -272,6 +276,14 @@ class PandasExprStringNamespace:
             "str",
             "ends_with",
             suffix,
+        )
+
+    def starts_with(self, prefix: str) -> PandasExpr:
+        return register_namespace_expression_call(
+            self._expr,
+            "str",
+            "starts_with",
+            prefix,
         )
 
     def head(self, n: int = 5) -> PandasExpr:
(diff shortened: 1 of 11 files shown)


narwhals/_pandas_like/expr.py
```diff
@@ -330,3 +342,166 @@ def total_microseconds(self) -> PandasExpr:
...
+            raise ValueError(msg)
+
+        def mapper(df: PandasDataFrame) -> list[PandasSeries]:
+            tmp = df.select(root_names)
+            return [
+                PandasSeries(
+                    tmp[n]._series.rename(function(n)), implementation=implementation
+                )
+                for n in root_names
+            ]
```
FBruzzesi, line 370: Is there a better way to iterate over the series-es that compose the dataframe?

narwhals/_pandas_like/expr.py
```diff
@@ -274,6 +278,14 @@ def ends_with(self, suffix: str) -> PandasExpr:
             suffix,
         )
 
+    def starts_with(self, prefix: str) -> PandasExpr:
+        return register_namespace_expression_call(
+            self._expr,
+            "str",
+            "starts_with",
+            prefix,
+        )
```
MarcoGorelli, line 287: this looks self-contained, and already correct - fancy splitting it out into a separate PR so we can merge it quickly?

narwhals/_pandas_like/expr.py
```diff
@@ -330,3 +342,166 @@ def total_microseconds(self) -> PandasExpr:
...
+        if root_names is None:
+            msg = (
+                "Anonymous expressions are not supported in `.name.map`.\n"
+                "Instead of `nw.all()`, try using a named expression, such as "
+                "`nw.col('a', 'b')`\n"
+            )
+            raise ValueError(msg)
+
+        def mapper(df: PandasDataFrame) -> list[PandasSeries]:
+            tmp = df.select(root_names)
```
MarcoGorelli, line 364: I don't think you want to do the select on the original dataframe, otherwise you lose all the other transformations that took place

e.g.:
```python
In [9]: df = pd.DataFrame({'a': [2, 1, 3], 'b': [4, 5, 6]})

In [10]: nw.to_native(nw.from_native(df).with_columns((nw.col('a')*2).name.suffix('_doubled')))
Out[10]:
   a  b  a_doubled
0  2  4          2
1  1  5          1
2  3  6          3
```

MarcoGorelli (commented):
Thanks for doing this! I'm impressed with how far you took it 💪 

Sorry for the short review, I'm heading out soon-ish - but tomorrow I've got a loooong train ride so I'll give some better feedback

narwhals/_pandas_like/expr.py
FBruzzesi: I see, let me take a better look at this and test it properly as well

narwhals/_pandas_like/expr.py
FBruzzesi: Very self contained :) Would you mind waiting for the rest? (Lazy mode activated from my side 😂)

narwhals/_pandas_like/expr.py
```diff
@@ -330,3 +342,117 @@ def total_microseconds(self) -> PandasExpr:
...
+        return self._expr.__class__(
+            lambda df: [
+                series.alias(function(series.name)) for series in self._expr._call(df)
+            ],
+            depth=self._expr._depth,
+            function_name=self._expr._function_name,
+            root_names=root_names,
+            output_names=[function(name) for name in root_names],
+            implementation=self._expr._implementation,
+        )
```
FBruzzesi, line 371: I will comment just one but logic applies for all the methods.

- This implementation works fine with chaining operations (cfr updated test suite)
- I am now calling `self._expr._call` and renaming the series using alias method, and name attribute
- The renaming is done on the `root_names`
- Unsure if the depth should be increased or not

MarcoGorelli: nice!

depth shouldn't be increased, as this is still just a simple aggregation, so we can use the fastpath in group_by

I'm not really proud of using `depth` to determine whether an aggregation is simple in group_by...maybe we can make a `SimplePandasExpr` subclass of `PandasExpr`, and then do an `isinstance` check...that should be clearer to read and to maintain

FBruzzesi: Not sure I completely follow along. I will need to come back at this 😂

MarcoGorelli: I'll  explain it in the 'how it works' PR - anyway, I think you're doing things correctly here 👍 

one way to check would be to do something like
```python
import pandas as pd
import narwhals as nw
dfpd = pd.DataFrame({'a': [1,1,2], 'b': [5,6,7]})
df = nw.from_native(df)
df = df.group_by('a').agg(pl.col('b').name.suffix('_baz').mean())
print(nw.to_native(df))
```
and check that:
1. you get the correct result
2. no perf warning is raised

FBruzzesi: If I add an _aggregation_ I get a warning which seems unrelated.

```py
import pandas as pd
import narwhals as nw
df_pd = pd.DataFrame({'a': [1,1,2], 'b': [5,6,7]})
df = nw.from_native(df_pd)
df = df.group_by('a').agg(nw.col('b').mean().name.suffix('_baz').mean())
print(nw.to_native(df))
```
results in:
```terminal
 .../narwhals/narwhals/_pandas_like/group_by.py:56: UserWarning: Found complex group-by expression, which can't be expressed efficiently with the pandas API. If you can, please rewrite your query such that group-by aggregations are simple (e.g. mean, std, min, max, ...).
  return agg_pandas(
   a  b_baz
0  1    5.5
1  2    7.0
```

MarcoGorelli: thanks for checking - looks fine then, the warning looks correct if there's a double-aggregation. will expand on it in the other PR 😇

MarcoGorelli:
I just tried this out , and I think we have a behaviour difference:
```python
In [3]: import pandas as pd

In [4]: import polars as pl

In [5]: import narwhals as nw

In [6]: data = {'a': [1,2,3], 'b': [4,5,6]}

In [7]: dfpd = pd.DataFrame(data)

In [8]: dfpl = pl.DataFrame(data)

In [9]: def func(df_any):
   ...:     df = nw.from_native(df_any)
   ...:     df = df.select(nw.col('a').alias('c').name.prefix('jedi_of_the_'))
   ...:     return nw.to_native(df)
   ...:

In [10]: func(dfpd)
Out[10]:
   jedi_of_the_c
0              1
1              2
2              3

In [11]: func(dfpl)
Out[11]:
shape: (3, 1)
┌───────────────┐
│ jedi_of_the_a │
│ ---           │
│ i64           │
╞═══════════════╡
│ 1             │
│ 2             │
│ 3             │
└───────────────┘
```

I think this one isn't going to be very easy to do without quite a bit refactor unfortunately

narwhals/_pandas_like/expr.py
```diff
@@ -345,3 +357,117 @@ def total_microseconds(self) -> PandasExpr:
...
+            raise ValueError(msg)
+
+        return self._expr.__class__(
+            lambda df: [
+                series.alias(series.name + suffix) for series in self._expr._call(df)
+            ],
+            depth=self._expr._depth,
+            function_name=self._expr._function_name,
+            root_names=root_names,
+            output_names=[name + suffix for name in root_names],
```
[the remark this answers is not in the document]
FBruzzesi: Sometimes github loses notification I guess! Anyway, due to 

> This will undo any previous renaming operations on the expression.

I thought applying to root_names was the correct way of doing it. But from the comment below I understand it is not enough :(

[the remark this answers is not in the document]
MarcoGorelli: you might be able to do it on the root names, and then apply the transformations to
```
            lambda df: [
                series.alias(function(series.name)) for series in self._expr._call(df)
            ],
```

but the number of output names might not match the number of root names. I'm a little worried about introducing bugs here, we may need to study the Polars code for this operation first (I've not looked at that part of the codebase yet)

FBruzzesi:
Closing in favor of #552

PR #159, opened 2024-05-13
feat: descriptive

FBruzzesi:
# Description

Adds dataframe methods:

- is_duplicated
- is_empty
- is_unique
- null_count

Fix #158

PR #187, opened 2024-05-18
feat: Series descriptive methods

FBruzzesi:
# Description

Similar to #159 but for series and expr methods ([complete list from polars](https://docs.pola.rs/py-polars/html/reference/series/descriptive.html)); introduces the following Series methods:

- is_duplicated
- is_empty (not in expr)
- is_unique
- null_count
- is_first_distinct
- is_last_distinct
- is_sorted (not in expr)
- value_counts (not in expr)

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -458,6 +458,25 @@ class PandasSeries:
     def is_last_distinct(self: Self) -> Self:
         return self._from_series(~self._series.duplicated(keep="last"))
 
+    def is_sorted(self: Self, *, descending: bool = False) -> bool:
+        if not isinstance(descending, bool):
+            msg = f"argument 'descending' should be boolean, found {type(descending)}"
+            raise TypeError(msg)
+
+        if descending:
+            return self._series.is_monotonic_decreasing  # type: ignore[no-any-return]
+        else:
+            return self._series.is_monotonic_increasing  # type: ignore[no-any-return]
+
+    def value_counts(self: Self, *, sort: bool = False, parallel: bool = False) -> Any:
+        """Parallel is unused, exists for compatibility"""
+        from narwhals._pandas_like.dataframe import PandasDataFrame
+
+        return PandasDataFrame(
+            self._series.value_counts(dropna=False, sort=sort).reset_index(),
+            implementation=self._implementation,
+        )
+
     @property
     def str(self) -> PandasSeriesStringNamespace:
         return PandasSeriesStringNamespace(self)
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -472,8 +472,14 @@ class PandasSeries:
         """Parallel is unused, exists for compatibility"""
         from narwhals._pandas_like.dataframe import PandasDataFrame
 
+        name_ = self._series.name or "index"
+        val_count = self._series.value_counts(dropna=False, sort=False).reset_index()
+        val_count.columns = [name_, "count"]
+        if sort:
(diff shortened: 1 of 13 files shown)


utils/check_api_reference.py
```diff
@@ -133,7 +133,9 @@
 if (
     extra := set(series)
     .difference(expr)
-    .difference({"to_pandas", "to_numpy", "dtype", "name", "shape", "to_frame"})
+    .difference(
+        {"to_pandas", "to_numpy", "dtype", "name", "shape", "to_frame", "is_empty"}
+    )
```
FBruzzesi, line 138: Would it make sense to get directly the set of methods that a polars Series has but are not an Expr? It would avoid having to manually add method names.

Something like:

```py
series_only = [i for i in dir(pl.Series) if not i[0].isupper() and i[0] != "_" and i not in dir(pl.Expr)]
```

Let me know what you think

MarcoGorelli: maybe, but there's also value in this being manual, so we can inspect whether it makes sense that a given method is missing

MarcoGorelli:
thanks @FBruzzesi ! Looks like `value_counts` is failing on the minimum pandas version, possibly related to https://github.com/pandas-dev/pandas/pull/49912

FBruzzesi:
> thanks @FBruzzesi ! Looks like `value_counts` is failing on the minimum pandas version, possibly related to [pandas-dev/pandas#49912](https://github.com/pandas-dev/pandas/pull/49912)

Thanks Marco, it seems that both naming **and** order is different (as if `sort=True` has no effect?). Should we manually address that? E.g. renaming and sorting afterwards?

MarcoGorelli:
I think order is fine, there are other places where the order isn't preserved (e.g. group_by) - shall we just document the lack of ordering guarantee, and fixup the names?

FBruzzesi:
Pandas `value_counts` has a `sort` keyword since forever I think ([v1.0](https://pandas.pydata.org/pandas-docs/version/1.0/reference/api/pandas.Series.value_counts.html) is the furthest I could find on the doc page). Setting `sort=True` and getting unsorted results is not ideal to me

MarcoGorelli:
yeah, sorry, hadn't looked carefully enough - sure, this one should be easy enough to work around and manually fixup

narwhals/_pandas_like/series.py
```diff
@@ -439,6 +439,50 @@ def to_pandas(self) -> Any:
...
+        if descending:
+            return self._series.is_monotonic_decreasing  # type: ignore[no-any-return]
+        else:
+            return self._series.is_monotonic_increasing  # type: ignore[no-any-return]
+
+    def value_counts(self: Self, *, sort: bool = False, parallel: bool = False) -> Any:
+        """Parallel is unused, exists for compatibility"""
+        from narwhals._pandas_like.dataframe import PandasDataFrame
+
+        name_ = self._series.name or "index"
```
MarcoGorelli, line 475: if `self._series.name` is something like `0` or `''`, will this overwrite it with `'index'`? should be specifically check for if `self._series.name is None`?

MarcoGorelli (approved):
Wow, absolute legend mate!

I can never remember the `is_first_distinct` in pandas, so I may need to refer to this if I ever need to write pandas again (and I probably will at some point 😄 )

Just made a minor comment

Approving, though I didn't sleep much tonight (got up early to catch a train) so I'm not super-"with it", feel free to merge if you're confident

narwhals/_pandas_like/series.py
FBruzzesi: Thanks! That's a great catch 🙈

PR #188, opened 2024-05-19
add how it works section

### LINKED ISSUE #182: Add "how it works" page
There should be a page explaining the basic idea, i.e. that an expression is a function from a dataframe to a Series

For example, if we start with `pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})`, then `nw.col('a')` takes `df` as input and returns `df['a']`

We can access that function with `._col`

e.g.:
```python
In [1]: import pandas as pd

In [2]: from narwhals._pandas_like.dataframe import PandasDataFrame

In [3]: from narwhals._pandas_like.namespace import PandasNamespace

In [4]: df = PandasDataFrame(pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]}), implementation='pandas')

In [5]: pn = PandasNamespace(implementation='pandas')

In [6]: pn.col('a')._call(df)[0]._series
Out[6]:
0    1
1    2
2    3
Name: a, dtype: int64
```

Or:
```python
In [7]: (pn.col('a') + 1)._call(df)[0]._series
Out[7]:
0    2
1    3
2    4
Name: a, dtype: int64
```

MarcoGorelli:
closes #182

### DIFF
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -1,0 +1,142 @@
+# How it works
+
+## Theory
+
+You might think that Narwhals runs on underwater unicorn magic. However, this section exists
+to reassure you that there's no such thing. There's only one rule you need to understand in
+order to make sense of Narwhals:
+
+> **An expression is a function from a DataFrame to a sequence of Series.**
+
+For example, `nw.col('a')` means "given a dataframe `df`, give me the Series `'a'` from `df`".
+Translating this to pandas syntax, we get:
+
+```python
+def col_a(df):
+    return [df.loc[:, 'a']]
+```
+
+Let's step up the complexity. How about `nw.col('a')+1`? We already know what the
+`nw.col('a')` part looks like, so we just need to add `1` to each of its outputs:
+
+```python
+def col_a(df):
+    return [df.loc[:, 'a']]
+
+def col_a_plus_1(df):
+    return [x+1 for x in col_a(df)]
+```
+
+Expressions can return multiple Series - for example, `nw.col('a', 'b')` translates to:
+
+```python
+def col_a_b(df):
+    return [df.loc[:, 'a'], df.loc[:, 'b']]
+```
+
+Expressions can also take multiple columns as input - for example, `nw.sum_horizontal('a', 'b')`
+translates to:
+
+```python
+def sum_horizontal_a_b(df):
+    return [df.loc[:, 'a'] + df.loc[:, 'b']]
+```
+
+Note that although an expression may have multiple columns as input,
+those columns must all have been derived from the same dataframe.
+
+By itself, an expression doesn't produce a value. It only produces a value once you give it to a
+DataFrame context. What happens to the value(s) it produces depends on which context you hand
+it to:
+
+- `DataFrame.select`: produce a DataFrame with only the result of the given expression
+- `DataFrame.with_columns`: produce a DataFrame like the current one, but also with the result of
+  the given expression
+- `DataFrame.filter`: evaluate the given expression, and if it only returns a single Series, then
+  only keep rows where the result is `True`.
+
+Now let's turn our attention to the implementation.
+
+## Polars implementation
+
+For Polars, Narwhals just "passes everything through". For example consider the following:
+```python
+import polars as pl
+import narwhals as nw
+
+df_pl = pl.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
+df = nw.from_native(df_pl)
+df.select(nw.col('a')+1)
+```
+
+`nw.col('a')` produces a `narwhals.expression.Expr` object, which has a private `_call` method.
+We can call `nw.col('a')._call(pl)`, then the result is actually `pl.col('a')`.
+
+We then let Polars do its thing. Which is nice, but also not particularly interesting.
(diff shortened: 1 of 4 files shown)


narwhals/dependencies.py
```diff
@@ -18,8 +18,7 @@ def get_pandas() -> Any:
 
 def get_modin() -> Any:  # pragma: no cover
     """Import modin (if available - else return None)."""
-    modin = sys.modules.get("modin", None)
-    if modin is not None:
+    if (modin := sys.modules.get("modin", None)) is not None:
```
MarcoGorelli, line 21: drive-by walrus

docs/how_it_works.md
```diff
@@ -0,0 +1,142 @@
...
+from narwhals._pandas_like.namespace import PandasNamespace
+
+pn = PandasNamespace(implementation='pandas')
+nw.col('a')._call(pn)
+```
+The result from the last line above is the same as we'd get from `pn.col('a')`, and it's
+a `narwhals._pandas_like.expression.PandasExpr` object, which we'll call `PandasExpr` for
+short.
+
+`PandasExpr` also have a `_call` method - but this one expects a `PandasDataFrame` as input.
```
FBruzzesi, line 97: Nit:

```suggestion
`PandasExpr` also has a `_call` method - but this one expects a `PandasDataFrame` as input.
```

docs/how_it_works.md
```diff
@@ -0,0 +1,142 @@
...
+Recall from above that an expression is a function from a dataframe to a sequence of series.
+The `_call` method gives us that function! Let's see it in action.
+
+Note: the following examples uses `PandasDataFrame` and `PandasSeries`. These are wrappers
+around pandas DataFrame and pandas Series, which are Narwhals-compliant. To get the native
+pandas objects out from inside them, we access `PandasDataFrame._dataframe` and `PandasSeries._series`.
```
FBruzzesi, line 103: Not the biggest fan of this phrasing. Maybe something like:

```suggestion
around pandas DataFrame and pandas Series, which are Narwhals-compliant. We can access the 
pandas objects via `PandasDataFrame._dataframe` and `PandasSeries._series`.
```

docs/how_it_works.md
```diff
@@ -0,0 +1,142 @@
...
+import narwhals as nw
+from narwhals._pandas_like.namespace import PandasNamespace
+from narwhals._pandas_like.dataframe import PandasDataFrame
+import pandas as pd
+
+pn = PandasNamespace(implementation='pandas')
+
+df_pd = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
+df = PandasDataFrame(df_pd, implementation='pandas')
+expression = pn.col('a') + 1
```
FBruzzesi, line 115: Should we use some fancy mkdocs-material to highlight the fact that this such line does not compute anything as is? Maybe [code annotation](https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#code-annotations)

docs/how_it_works.md
```diff
@@ -0,0 +1,142 @@
...
+0    2
+1    3
+2    4
+Name: a, dtype: int64
+```
+
+So indeed, our expression did what it said on the tin - it took some dataframe, took
+column 'a', and added 1 to it.
+
+If you search for `def register_expression_call`, you'll see that that's all
```
FBruzzesi, line 130: Let's handle the link ourselves:

```suggestion
If you search for [`register_expression_call`](https://github.com/narwhals-dev/narwhals/blob/f18c533a0c2ff870ad489a6f93bca8bef2a98eda/narwhals/_pandas_like/utils.py#L171){target="_blank"}, you'll see that that's all
```

FBruzzesi (commented):
That's a great write up! I got a much better understanding of the situation and internals 

The only addition I would do it to mention somewhere that pandas implementation is used also for the pandas-like APIs.

Other comments are just food for thought :)


ThanksThankYouGIF

docs/how_it_works.md
MarcoGorelli: the `{target="_black"}` didn't work for me when I rendered this. in any case, are sure? I think the link would quickly go out-of-date

MarcoGorelli:
thanks Francesco for taking a look! I've addressed most comments, those I wasn't able to do shall we leave them for a follow-up?

docs/how_it_works.md
FBruzzesi: The one I added is the permalink, not just the link to main and line, which in theory should be more stable to changes in the file. `{target="_blank"}` should open on a new tab/page, which is something I prefer but can live without 😂

MarcoGorelli: do i need to enable some extra feature for it to work? if i copy and paste your snippet, i get:

image

MarcoGorelli:
going to merge this as I wanted to link this to somebody, we can then address the permalink (https://github.com/narwhals-dev/narwhals/pull/188#discussion_r1615621451) in a separate PR

thanks for your review!

docs/how_it_works.md
FBruzzesi: I will take a look, but if you want to merge it, let's remove it for now 😇

PR #189, opened 2024-05-19
DOC: fill in missing docstrings for Series.is_in

natmokval:
xref #71

### DIFF
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -463,8 +463,8 @@ class Series:
             We can then pass either pandas or Polars to `func`:
 
             >>> func(s_pd)
-            1    False
-            2     True
+            0    False
+            1     True
             2     True
             dtype: boolean
             >>> func(s_pl)  # doctest: +NORMALIZE_WHITESPACE
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -441,7 +441,7 @@ class Series:
 
     def is_in(self, other: Any) -> Self:
         """
-        Check if elements of this sequence are present in the other Series.
+        Check if the elements of this Series are in the other sequence.
 
         Arguments:
             other: Sequence of primitive type.
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -441,10 +441,10 @@ class Series:
 
     def is_in(self, other: Any) -> Self:
         """
-        Check if elements of this Series are in the other Series.
+        Check if elements of this sequence are present in the other Series.
 
         Arguments:
-            other: Series or sequence of primitive type.
+            other: Sequence of primitive type.
 
         Examples:
             >>> import pandas as pd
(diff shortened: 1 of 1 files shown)


narwhals/series.py
```diff
@@ -440,6 +440,42 @@ def std(self, *, ddof: int = 1) -> Any:
         return self._series.std(ddof=ddof)
 
     def is_in(self, other: Any) -> Self:
+        """
+        Check if elements of this Series are in the other Series.
+
+        Arguments:
+            other: Series or sequence of primitive type.
```
MarcoGorelli, line 447: this should just be "sequence of primitive type"

narwhals/series.py
```diff
@@ -440,6 +440,42 @@ def std(self, *, ddof: int = 1) -> Any:
         return self._series.std(ddof=ddof)
 
     def is_in(self, other: Any) -> Self:
+        """
+        Check if elements of this Series are in the other Series.
```
MarcoGorelli, line 444:
```
Check if elements of this Series are in a given sequence.
```

we don't support `other` being a Series of lists here as Polars does, we can only accept a single sequence

narwhals/series.py
```diff
@@ -440,6 +440,42 @@ def std(self, *, ddof: int = 1) -> Any:
...
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> s_pd = pd.Series([1, 2, 3])
+            >>> s_pl = pl.Series([1, 2, 3])
+
+            We define a library agnostic function:
+
+            >>> def func(s_any):
+            ...     s = nw.from_native(s_any, series_only=True)
+            ...     s = s.is_in(pl.Series([3, 2, 8]))
```
MarcoGorelli, line 460: we shouldn't use anything polars-specific in these examples, can you just use `is_in([3, 2, 8])`?

narwhals/series.py
natmokval: thanks, I corrected the description of `other`

narwhals/series.py
natmokval: thanks, I updated summary.

narwhals/series.py
natmokval: done

narwhals/series.py
```diff
@@ -440,6 +440,42 @@ def std(self, *, ddof: int = 1) -> Any:
         return self._series.std(ddof=ddof)
 
     def is_in(self, other: Any) -> Self:
+        """
+        Check if elements of this sequence are present in the other Series.
+
```
DeaMariaLeon, line 445: We are checking if the Series is in the `other` but `other` is not a Series. 

So I think it should be:
Check if the elements of this Series are in the `other` sequence.

DeaMariaLeon:
Thanks @natmokval, I commented..

narwhals/series.py
natmokval: Thanks @DeaMariaLeon for the comment, I corrected the summary

DeaMariaLeon:
Thank you @natmokval!

Issue #101, opened 2024-05-07
improve error message for `is_in`


MarcoGorelli:
```python
import narwhals as nw
import pandas as pd
import polars as pl

data = {"sets": [[1, 2, 3], [1, 2], [9, 10]], "optional_members": [1, 2, 3]}

df_pd = pd.DataFrame(data)
df_pl = pl.DataFrame(data)

def func(df_any):
    df = nw.from_native(df_any)
    df = df.with_columns(contains=nw.col("optional_members").is_in("sets"))
    return nw.to_native(df)

func(df_pd)
```

this gives
```
TypeError: only list-like objects are allowed to be passed to isin(), you passed a `str`
```

We should improve the error message so it's clear that Narwhals less flexible than Polars here (where you can just pass expressions to `is_in`)

PR #195, opened 2024-05-21
DOC: usage example for is_in added in Expr

anopsy:
I noticed that the narwhals Expr.is_in() works slightly different than polars Expr.is_in() - 
it doesn't accept Series and Expressions as "other" argument.  Therefore I changed the description to "list"and "List-like objects"
because it was what I've found in the error description.

Edit: I see now, that has been addressed in #101 , should I then change "list", "list-like objects"to Iterable in the docstrings?

### DIFF
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -636,6 +636,48 @@ class Expr:
         )
 
     def is_in(self, other: Any) -> Expr:
+        """
+        Check if elements of this expression are present in the other list.
+
+        Arguments:
+            other: List-like objects.
+
+        Examples:
+            >>> import pandas as pd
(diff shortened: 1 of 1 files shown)


MarcoGorelli (commented):
Thanks for writing this! Yup, would be great to update based on the other PR (and also there is a merge conflict, feel free to write in #help if resolving it trips you up)

anopsy:
exciting, my first merge conflict. I went for fixing in on github, therefore couldn't test it.

MarcoGorelli (approved):
wonderful! well done with the merge conflict

(the first time I got one I just deleted everything and started over again, 2/10 would not recommend)

PR #196, opened 2024-05-21
added expr.filter docstring

brentomagic:


narwhals/expression.py
```diff
@@ -847,6 +847,42 @@ def is_in(self, other: Any) -> Expr:
...
+            >>> import pandas as pd
+            >>> import narwhals as nw
+            >>> df_pd = pd.DataFrame({'a': [2, 3, 4, 5, 6, 7], 'b': [10, 11, 12, 13, 14, 15]})
+            >>> df_pl = pl.DataFrame({'a': [2, 3, 4, 5, 6, 7], 'b': [10, 11, 12, 13, 14, 15]})
+
+            Let's define a dataframe-agnostic function:
+
+            >>> def func(df_any):
+            ...    df = nw.from_native(df_any)
+            ...    df = df.filter((nw.col('a') > 4) & (nw.col('b') > 12))
```
FBruzzesi, line 864: I think that `expr.filter` should exemplify what happens when filtering within a context.

The example I wrote in #197:

```py
import polars as pl

frame = pl.DataFrame({
    "a": [1, 2, 3, 1],
    "b": ["a", "a", "b", "c"]
})

frame.select(pl.col("a").filter(pl.col("b")=="a").mean().alias("a_mean_filtered"), pl.col("a").mean()
```

Here you are showing how [`df.filter`](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.filter) works

FBruzzesi (commented):
Hey there! Thanks for the PR! I left a feedback coming from #197 😇

PR #198, opened 2024-05-21
nanoseconds with docstring

MatthewMec:
nanoseconds, no test will work on that tonight.

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -546,6 +546,14 @@ class PandasSeriesDateTimeNamespace:
             self._series._series.dt.second,
         )
 
+    def microsecond(self) -> PandasSeries:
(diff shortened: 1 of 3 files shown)


MatthewMec:
Added test for nanosecond as well as docstrings.

narwhals/_pandas_like/series.py
```diff
@@ -546,6 +546,11 @@ def second(self) -> PandasSeries:
             self._series._series.dt.second,
         )
 
+    def nanosecond(self) -> PandasSeries:
+        return self._series._from_series(
+            ((self._series._series.dt.microsecond * 1e3).astype(int)),
```
MarcoGorelli, line 551: we don't currently have a good way of constructing datetimes with nanoseconds, but here you'll also need the nanosecond component:
```
self._series._series.dt.microsecond * 1_000 + self._series._series.dt.nanosecond
```

PR #207, opened 2024-05-24
feat: series and expr quantile

### LINKED ISSUE #203: Add Expr.quantile and Series.quantile
https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.quantile.html#polars.Expr.quantile

I'm _slightly_ worried that the default interpolation method might change in Polars at some point, so we should probably not put any default for it in Narwhals and force users to specify it explicitly

FBruzzesi:
# Description

Fix #203 by introducing `.quantile(...)` for Expr and Series.

I am opening this as draft because of two main reasons:

- [ ] I may need some support with type hinting
- [ ] Need decision for how to deal with non numeric formats

### DIFF
--- a/docs/api-reference/expressions.md
+++ b/docs/api-reference/expressions.md
@@ -26,11 +26,12 @@
         - null_count
         - n_unique
         - over
-        - unique
+        - quantile
         - sample
         - shift
         - sort
         - std
         - sum
+        - unique
       show_source: false
       show_bases: false
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 from typing import Iterable
+from typing import Literal
 
 from narwhals.dependencies import get_polars
 from narwhals.dtypes import translate_dtype
@@ -1417,7 +1418,11 @@ class Expr:
         """
         return self.__class__(lambda plx: self._call(plx).is_last_distinct())
 
-    def quantile(self, quantile: float, interpolation: str = "nearest") -> Expr:
+    def quantile(
+        self,
+        quantile: float,
+        interpolation: Literal["nearest", "higher", "lower", "midpoint", "linear"],
+    ) -> Expr:
         r"""Get quantile value.
 
         Arguments:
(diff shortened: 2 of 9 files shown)


docs/api-reference/expressions.md
```diff
@@ -26,11 +26,12 @@
...
         - n_unique
         - over
-        - unique
+        - quantile
         - sample
         - shift
         - sort
         - std
         - sum
+        - unique
```
FBruzzesi, line 35: I assumed it was supposed to be alphabetically sorted

FBruzzesi:
To check the pandas behavior in the [failing GHA](https://github.com/narwhals-dev/narwhals/actions/runs/9224134064/job/25378803025?pr=207), I found that the min version in which it doesn't break is `pandas==1.3.5`. Pandas versions previous to that one require `numpy<=1.21`.

MarcoGorelli:
thanks - maybe we can just pin `numpy` in the ci-minimum-versions CI job too then?

narwhals/expression.py
```diff
@@ -1417,6 +1418,54 @@ def is_last_distinct(self) -> Expr:
...
+
+            >>> func(df_pl)  # doctest: +NORMALIZE_WHITESPACE
+            shape: (1, 2)
+            ┌──────┬──────┐
+            │ a    ┆ b    │
+            │ ---  ┆ ---  │
+            │ f64  ┆ f64  │
+            ╞══════╪══════╡
+            │ 25.0 ┆ 75.0 │
+            └──────┴──────┘
```
MarcoGorelli, line 1463: i'm a little surprised that they output different results, have you checked why?

MarcoGorelli (commented):
thanks @FBruzzesi ! just got a question

narwhals/expression.py
FBruzzesi: True! The quantile falls in the middle of two numbers. numpy and pandas implementation of `method="nearest"` seem to round down while rust/polars to round up? 
Should we use a more stable example?

MarcoGorelli: yeah maybe 'linear' in the examples, and then add a note explaining that there may be some slight implementation differences when rounding?

PR #210, opened 2024-05-24
Update df docstrings (shape, to_dict, to_pandas, to_numpy)

georgescutelnicu:
I'm not sure if the approach i took to update ```to_dict```'s docstirng is the most effective and appealing one, perhaps i shouldn't have made it output both versions with ```as_series``` ```True``` and ```False``` because it feels like is too much. Anyways i'll leave it here for some feedback until tomorrow.

EDIT: i see the tests are failing for to_dict docstring (maybe i should just add # doctest: +SKIP like it was before) , but im gonna take a look into it tomorrow, thanks.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,4 +28,8 @@ repos:
       entry: python -m utils.check_api_reference
       language: python
       additional_dependencies: [polars]
-
+    - id: imports-are-banned
+      name: import are banned (use `get_pandas` instead of `import pandas`)
+      entry: (?<!>>> )import (pandas|polars|modin|cudf)
+      language: pygrep
+      files: ^narwhals/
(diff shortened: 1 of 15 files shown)


MarcoGorelli (commented):
Thanks for doing this! maybe for `as_dict` just `as_series=False` suffices

also, i made a little update in https://github.com/narwhals-dev/narwhals/pull/211

georgescutelnicu:
The tests seem to fail due to a discrepancy between the expected and actual output of ```to_dict``` (although the output matches the expected one). Am I missing something, and should the output actually be displayed in a single line?

MarcoGorelli (approved):
thanks @georgescutelnicu ! for the doctest failure, maybe you can just make the example much smaller so that it fits on a single line?

MarcoGorelli:
thanks @georgescutelnicu !

PR #215, opened 2024-05-26
DOC: fill in missing docstrings for GroupBy.agg

natmokval:
xref #71

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -28,6 +28,7 @@ nav:
     - api-reference/dtypes.md
     - api-reference/dependencies.md
     - api-reference/selectors.md
+    - api-reference/group_by.md
 theme:
   name: material
   font: false
--- a/narwhals/group_by.py
+++ b/narwhals/group_by.py
@@ -23,6 +23,87 @@ class GroupBy:
     def agg(
         self, *aggs: IntoExpr | Iterable[IntoExpr], **named_aggs: IntoExpr
     ) -> DataFrame:
+        """
+        Compute aggregations for each group of a group by operation.
+
+        Arguments:
+            *aggs: Aggregations to compute for each group of the group by operation,
+            specified as positional arguments.
+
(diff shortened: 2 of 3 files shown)


mkdocs.yml
```diff
@@ -28,6 +28,7 @@ nav:
     - api-reference/dtypes.md
     - api-reference/dependencies.md
     - api-reference/selectors.md
+    - api-reference/group_by.md
```
DeaMariaLeon, line 31: I wonder if we should order these alphabetically.. ?

DeaMariaLeon:
Thank you @natmokval, it's nice! .. but I did have a question so I leave it to @MarcoGorelli ?

narwhals/group_by.py
```diff
@@ -23,6 +23,86 @@ def __init__(self, df: DataFrame, *keys: str | Iterable[str]) -> None:
...
+            We define library agnostic functions:
+
+            >>> def func(df_any):
+            ...     df = nw.from_native(df_any)
+            ...     df = df.group_by("a").agg(nw.col("b").sum()).sort("a")
+            ...     return nw.to_native(df)
+
+            >>> def func_mult_col(df_any):
+            ...     df = nw.from_native(df_any)
+            ...     df = df.group_by(["a", "b"]).agg(nw.sum("c")).sort("a", "b")
```
MarcoGorelli, line 66:
```suggestion
            ...     df = df.group_by("a", "b").agg(nw.sum("c")).sort("a", "b")
```

MarcoGorelli (approved):
nice one, thanks @natmokval !

narwhals/group_by.py
MarcoGorelli: this is the same, just a little shorter

mkdocs.yml
natmokval: thanks for the comment. I sorted alphabetically `API Reference` and submitted PR #245

PR #227, opened 2024-05-29
added max method docstring

brentomagic:


narwhals/expression.py
```diff
@@ -2290,7 +2290,43 @@ def min(*columns: str) -> Expr:
...
 def max(*columns: str) -> Expr:
     """
-    Instantiate an expression representing the maximum of one or more columns, similar to `polars.max`.
+    Return the maximum value.
+
+    Note:
+       Syntactic sugar for nw.col(names).max().
+
+    Parameters:
+        *names
```
MarcoGorelli, line 2299: the function signature says `columns`, but here you have `names` - can you update that to also use `names`?

(we should do the same with other such functions)

narwhals/expression.py
```diff
@@ -2290,7 +2290,43 @@ def min(*columns: str) -> Expr:
 
 def max(*columns: str) -> Expr:
     """
-    Instantiate an expression representing the maximum of one or more columns, similar to `polars.max`.
+    Return the maximum value.
+
+    Note:
+       Syntactic sugar for nw.col(names).max().
```
MarcoGorelli, line 2296:
```suggestion
       Syntactic sugar for ``nw.col(names).max()``.
```

MarcoGorelli (commented):
thanks @brentomagic ! this looks close

MarcoGorelli (approved):
thanks @brentomagic !

PR #228, opened 2024-05-29
added docstring for min method

brentomagic:


MarcoGorelli (commented):
nice one - same comment as in the `max` pr

MarcoGorelli (approved):
thanks @brentomagic !

Issue #204, opened 2024-05-23
Research / POC: would Narwhals work with Fairlearn?


MarcoGorelli:
Here's the project: https://github.com/fairlearn/fairlearn

I'd be curious to see if it could be Narwhalified

**Important note**: please do **not** make any PRs to fairlearn itself. This is just intended as an exercise to find missing things in Narwhals which we might want to add in order to make it more useful. We have not discussed anything with Fairlearn devs, so we should not make unsolicited PRs to them.

The task here is:
- Check out FairLearn, look at their contributor guide: https://fairlearn.org/v0.10/contributor_guide/index.html for how to get set up
- Take a look at where they use pandas internally, and see if those parts could be Narwhalified
- It's very likely that you'll come across something which Narwhals doesn't (yet) support. In which case, please open an issue here, and we can add that feature

Once again, just to make sure there's no misunderstandings, please do not make any PRs to Fairlearn itself. This is just intended as an exploratory piece to see what we're missing in Narwhals - let's not disturb Fairlearn devs

EdAbati:
Hey @MarcoGorelli :) this sounds fun, I'd like to help.

I'll go through `fairlearn`'s modules and update you here

MarcoGorelli:
Woohoo that's great! Multiple people can investigate this together and then we can share findings

EdAbati:
Please see below the scripts where `pandas` is used (based on `grep -rl --include=\*.py "import pandas" fairlearn/*`)

I'll slowly update this comment with my findings. If anyone else is interested, feel free to help (maybe priotising the scripts with no comments?)  🙂

✅ means the script can be "narwhalified"

- [x] fairlearn/datasets/_fetch_acs_income.py ✅
   https://github.com/EdAbati/fairlearn/commit/adab8b30ebe9ff30d51393acb69083cbcca8c7c1
   _notes_:
   -  should `.to_pandas()` return a Series or a DataFrame when  there is only one column?
   - is there a `.iloc` alternative in narwhals?
- [ ] fairlearn/metrics/_bootstrap.py
- [ ] fairlearn/metrics/_disaggregated_result.py
- [ ] fairlearn/metrics/_plotter.py
- [ ] fairlearn/metrics/_group_feature.py
- [x] fairlearn/metrics/_annotated_metric_function.py ✅
   https://github.com/EdAbati/fairlearn/commit/f9180b1cd457117285ed6a45b7e5061597617cfa
[...]

PR #238, opened 2024-05-31
Keep Pandas-like Series original name

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

As discussed in #204, @MarcoGorelli is it what you had in mind?

## Checklist

- [x] Tests added 
   is it the correct place? Or should we create a `pandas_like/test_series.py` file?
- [ ] Documented the changes

Should we also add a line in the docs that say that certain operations assume `str` column names and may not work (e.g. `DataFrame.__getitem__`) ?
Or maybe we can have a warning that is triggered when the column name is not `str` (both in `Series` and `DataFrame`)

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -78,7 +78,7 @@ class PandasSeries:
             DataFrame this column originates from.
         """
 
-        self._name = str(series.name) if series.name is not None else ""
+        self._name = series.name
         self._series = series
         self._implementation = implementation
         self._use_copy_false = False
@@ -119,7 +119,7 @@ class PandasSeries:
 
     @property
     def name(self) -> str:
-        return self._name
+        return self._name  # type: ignore[no-any-return]
 
     @property
     def shape(self) -> tuple[int]:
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
thanks @EdAbati ! could you run `pytest narwhals --doctest-modules` too? there's a doctest that needs fixing up

MarcoGorelli:
> Should we also add a line in the docs that say that certain operations assume str column names and may not work (e.g. DataFrame.__getitem__) ?
Or maybe we can have a warning that is triggered when the column name is not str (both in Series and DataFrame)

tbh i think it's ok as-is, a warning risks being annoying. people using narwhals will probably pass in series/dataframe names taken from objects their users pass anyway, so using whatever the user passes should be fine

MarcoGorelli (approved):
thanks @EdAbati !

PR #245, opened 2024-06-01
DOC: sort API Reference

natmokval:
xref #215

sorted alphabetically API Reference

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -17,19 +17,19 @@ nav:
   - Related projects: related.md
   - API Reference:
     - api-reference/dataframe.md
-    - api-reference/dependencies.md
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -16,20 +16,20 @@ nav:
   - Roadmap: roadmap.md
   - Related projects: related.md
   - API Reference:
-    - api-reference/index.md
```
MarcoGorelli, line 19: thanks - i think all the upper-case ones should go first, and then the lowercase ones?

```
In [10]: sorted('DLdl')
Out[10]: ['D', 'L', 'd', 'l']
```

MarcoGorelli (approved):
thanks @natmokval !

PR #265, opened 2024-06-06
feat: implement narwhals.maybe_convert_dtypes

### LINKED ISSUE #241: Add `narwhals.maybe_convert_dtypes`
pandas-like libraries have `convert_dtypes`: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html

People use this to "promote" types - e.g., to promote all types to use the pyarrow-backed ones.

Polars doesn't have anything like this, because each type is just self-standing. E.g. `pl.Int64` is just `pl.Int64`, there aren't any variants of it

So, the task here is to add a method `maybe_convert_dtypes`, which for pandas-like calls `convert_dtypes`, and for the rest is a no-op, just like how Narwhals already has other `maybe_*` funcs: https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_set_index

DeaMariaLeon:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #241 

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -216,42 +216,13 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
     return df
 
 
-def maybe_convert_dtypes(  # noqa: PLR0913
-    df: T,
-    infer_objects: bool = True,  # noqa: FBT001, FBT002
-    convert_string: bool = True,  # noqa: FBT001, FBT002
-    convert_integer: bool = True,  # noqa: FBT001, FBT002
-    convert_boolean: bool = True,  # noqa: FBT001, FBT002
-    convert_floating: bool = True,  # noqa: FBT001, FBT002
-    dtype_backend: str = "numpy_nullable",
-) -> T:
+def maybe_convert_dtypes(df: T, *args: bool, **kwargs: bool | str) -> T:
     """
     Convert columns to the best possible dtypes using dtypes supporting ``pd.NA``, if df is pandas-like.
 
     Notes:
         For non-pandas-like inputs, this is a no-op.
-
-    Arguments:
-        infer_objects : bool, default True
-            Whether object dtypes should be converted to the best possible types.
-        convert_string : bool, default True
-            Whether object dtypes should be converted to ``StringDtype()``.
-        convert_integer : bool, default True
-            Whether, if possible, conversion can be done to integer extension types.
-        convert_boolean : bool, defaults True
-            Whether object dtypes should be converted to ``BooleanDtypes()``.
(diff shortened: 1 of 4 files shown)


narwhals/utils.py
```diff
@@ -214,3 +214,85 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
             )
         )
     return df
+
+
+def maybe_convert_dtypes(  # noqa: PLR0913
+    df: T,
+    infer_objects: bool = True,  # noqa: FBT001, FBT002
```
MarcoGorelli, line 221:
```suggestion
    df: T,
    *,
    infer_objects: bool = True,  # noqa: FBT001, FBT002
```
(and then remove the noqas)

MarcoGorelli: though actually, all of these _might_ get renamed in pandas, and then we'd be stuck

DeaMariaLeon: Done.
I wonder why it doesn't fail locally.

tests/utils_test.py
```diff
@@ -60,3 +60,20 @@ def test_maybe_set_index_polars() -> None:
...
     result = nw.maybe_set_index(df, "b")
     assert result is df
+
+
+def test_maybe_convert_dtypes_pandas() -> None:
+    import numpy as np
+
+    df = nw.from_native(pd.DataFrame({"a": [1.1, np.nan]}, dtype=np.dtype("float64")))
+    result = nw.to_native(nw.maybe_convert_dtypes(df))
+    expected = pd.DataFrame({"a": [1.1, pd.NA]}, dtype=pd.Float64Dtype())
```
MarcoGorelli, line 70: this fails on the minimum pandas version, could you use `Int64Dtype`?

DeaMariaLeon: Ups, I didn't see this comment before I tried the last commit. I'll update it.

MarcoGorelli:
looks like pandas' series repr was a bit different in older versions (you can reproduce this locally by pip-installing the pandas version from the minimum-versions ci job) for string dtype

maybe only include bool and int in the example?

DeaMariaLeon:
Thanks @MarcoGorelli

MarcoGorelli (approved):
thanks @DeaMariaLeon !

PR #266, opened 2024-06-06
feat: add `narwhalify` decorator

MarcoGorelli:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -103,6 +103,17 @@ def from_native(
 def from_native(
     native_dataframe: Any,
     *,
+    strict: bool = ...,
+    eager_only: None = ...,
+    series_only: None = ...,
+    allow_series: None = ...,
+) -> DataFrame | LazyFrame: ...
+
+
+@overload
+def from_native(
+    native_dataframe: Any,
+    *,
     strict: bool,
     eager_only: bool | None,
     series_only: bool | None,
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -302,7 +302,7 @@ def narwhalify(
             else:
                 kwargs[first_key] = df
                 result = func(**kwargs)
-            return to_native(result)
+            return to_native(result, strict=strict)
 
         return wrapper
 
@@ -379,7 +379,7 @@ def narwhalify_method(
             else:
                 kwargs[first_key] = df
                 result = func(self, **kwargs)
-            return to_native(result)
+            return to_native(result, strict=strict)
 
         return wrapper
 
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 from functools import wraps
 from typing import TYPE_CHECKING
 from typing import Any
@@ -294,7 +295,13 @@ def narwhalify(
             if args:
                 df_any = args[0]
             elif kwargs:
-                first_key = next(iter(kwargs.keys()))
+                params = list(inspect.signature(func).parameters.keys())
+                first_key = params[0]
+                if first_key == "self":
+                    raise TypeError(
+                        "It looks like you're trying to call `@nw.narwhalify` on a class method - please "
+                        "use `@nw.narwhalify_method` instead."
+                    )
                 df_any = kwargs[first_key]
             else:
                 raise TypeError("Expected function which takes at least one argument.")
@@ -379,7 +386,15 @@ def narwhalify_method(
             if args:
                 df_any = args[0]
             elif kwargs:
-                first_key = next(iter(kwargs.keys()))
+                params = list(inspect.signature(func).parameters.keys())
+                if params[0] not in ("cls", "self"):
+                    msg = (
+                        "`@nw.narwhalify_method` is meant to be called on class methods, "
+                        "where the first argument is typically `cls` or `self` - however, yours "
+                        f"is: {params[0]}."
+                    )
+                    raise TypeError(msg)
+                first_key = params[1]
                 df_any = kwargs[first_key]
             else:
                 raise TypeError("Expected function which takes at least one argument.")
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -277,7 +277,7 @@ def narwhalify(
         allow_series: Whether to allow series (default is only dataframe / lazyframe).
 
     See Also:
-        narwhalify_decorator: If you want narwhalify a class method, use that instead.
+        narwhalify_decorator: If you want to narwhalify a class method, use that instead.
     """
 
     def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -245,7 +245,7 @@ def narwhalify(
     allow_series: bool | None = None,
 ) -> Callable[..., Any]:
     """
-    Decorate a function so that the first argument gets narwhalified.
+    Decorate function so it becomes dataframe-agnostic.
 
     Instead of writing
 
@@ -267,6 +267,14 @@ def narwhalify(
     def func(df):
         return df.group_by('a').agg(nw.col('b').sum())
     ```
+
+    You can also pass in extra arguments, e.g.
+
+    ```python
+    @nw.narhwalify(eager_only=True)
+    ```
+
+    that will get passed down to `nw.from_native`.
 
     Arguments:
         func: Function to wrap in a `from_native`-`to_native` block.
@@ -323,7 +331,7 @@ def narwhalify_method(
     allow_series: bool | None = None,
 ) -> Callable[..., Any]:
     """
-    Decorate a function so that the first argument (other than `self`) gets narwhalified.
+    Decorate method so it becomes dataframe-agnostic.
 
     Instead of writing
 
@@ -347,6 +355,14 @@ def narwhalify_method(
         def func(self, df):
             return df.group_by('a').agg(nw.col('b').sum())
     ```
+
+    You can also pass in extra arguments, e.g.
+
+    ```python
+    @nw.narhwalify_method(eager_only=True)
+    ```
+
+    that will get passed down to `nw.from_native`.
 
     Arguments:
         func: Function to wrap in a `from_native`-`to_native` block.
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -297,11 +297,6 @@ def narwhalify(
             elif kwargs:
                 params = list(inspect.signature(func).parameters.keys())
                 first_key = params[0]
-                if first_key == "self":
-                    raise TypeError(
-                        "It looks like you're trying to call `@nw.narwhalify` on a class method - please "
-                        "use `@nw.narwhalify_method` instead."
-                    )
                 df_any = kwargs[first_key]
             else:
                 raise TypeError("Expected function which takes at least one argument.")
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -233,9 +233,58 @@ def narwhalify(
     series_only: bool | None = None,
     allow_series: bool | None = None,
 ) -> Callable[..., Any]:
+    """
+    Decorate a function so that the first argument gets narwhalified with `strict=False`.
+
+    Using
+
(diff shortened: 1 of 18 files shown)


narwhals/translate.py
```diff
@@ -223,11 +225,32 @@ def get_native_namespace(obj: Any) -> Any:
     return obj.__native_namespace__()
 
 
+def narwhalify(
+    func: Callable[..., Any] | None = None, *, eager_only: bool = False
```
MarcoGorelli, line 229: todo - add the other args, not just `eager_only`. they should probably be explicitly listed for tab-completion purposes. there's not many anyway

MarcoGorelli:
looks like pandas' series repr was a bit different in older versions (you can reproduce this locally by pip-installing the pandas version from the minimum-versions ci job) for string dtype

maybe only include bool and int in the example?

baggiponte:
Ciao Marco! Cool feature. Unfortunately you missed my lightning talk about this, but it was modelled after [this video](https://www.youtube.com/watch?v=_QXlbwRmqgI) by the excellent Anthony. TLDR: you should consider implementing the decorator using `contextlib.contextmanager`. At first glance, this decorator only returns a context manager... but it actually also makes the function you decorated available as a decorator. What do you think?

MarcoGorelli:
thanks! that seems like a different application though, not sure how it work here, as I need to pass `df` though - could you show an example please?

MarcoGorelli:
🤔 I'm not 100% about this

If we only narwhalify the first argument, then it means that
```
class Foo:
    @nw.narwhalify
    def transform(self, df):
        return df.select(self.columns)
```
would raise, because the first argument to be passed to `narwhalify` would be `self`, not `df`!

On the other hand, if we put _all_ arguments through `from_native` with `strict=False`, then we're encouraging non-strict patterns... 🤔 

Dunno. Gonna think about it a bit more

FBruzzesi:
@baggiponte I gave it a try and while the conversion to narwhals is seamless, the `to_native` is not obvious to me how is could work in a context manager. The context would require to receive something, and this is not trivial in my experience. The one time I had to accomplish this, I had to fallback to implementing the context manager as a class so that it is stateful.

@MarcoGorelli 

> maybe we just need to have both @nw.narwhalify and @nw.narwhalify_method?

There could be a more general (maybe internal only) function, which then can be _partial_ -ized for specific use cases and exposed these decorators. I have a somewhat relevant example in mind on how to do this ([`multi_shape_tracker`](https://github.com/FBruzzesi/deczoo/blob/main/deczoo/decorators.py#L661C5-L661C24) - please no hard judgment on this  😂)

narwhals/translate.py
```diff
@@ -223,11 +236,180 @@ def get_native_namespace(obj: Any) -> Any:
...
+        native_dataframe: Raw dataframe from user.
+            Depending on the other arguments, input object can be:
+
+            - pandas.DataFrame
+            - polars.DataFrame
+            - polars.LazyFrame
+            - anything with a `__narwhals_dataframe__` or `__narwhals_lazyframe__` method
+            - pandas.Series
+            - polars.Series
+            - anything with a `__narwhals_series__` method
```
FBruzzesi, line 281: Shouldn't this be: 

```suggestion
        func : Function to wrap in a `from_native`, `to_native` block.
```

Am I missing something here?

narwhals/translate.py
```diff
@@ -223,11 +236,180 @@ def get_native_namespace(obj: Any) -> Any:
...
+    def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
+        @wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            if args:
+                df_any = args[0]
+            elif kwargs:
+                first_key = next(iter(kwargs.keys()))
+                df_any = kwargs[first_key]
+            else:
+                raise TypeError("Expected function which takes at least one argument.")
```
FBruzzesi, line 298: Maybe this detail should be explained in the docstring? i.e. best practice is to wrap a function that takes a dataframe as first argument (or otherwise explicitly pass it as first keyword argument)

narwhals/translate.py
```diff
@@ -223,11 +236,180 @@ def get_native_namespace(obj: Any) -> Any:
...
+def narwhalify_method(
+    func: Callable[..., Any] | None = None,
+    *,
+    strict: bool = True,
+    eager_only: bool | None = None,
+    series_only: bool | None = None,
+    allow_series: bool | None = None,
+) -> Callable[..., Any]:
+    """
+    Decorate a function so that all arguments get narwhalified with `strict=False`.
```
FBruzzesi, line 332: Same as above

narwhals/translate.py
```diff
@@ -223,11 +236,180 @@ def get_native_namespace(obj: Any) -> Any:
...
+    import narwhals as nw
+
+    def func(df_any):
+        df = nw.from_native(df_any, strict=False)
+        df = df.group_by('a').agg(nw.col('b').sum())
+        return nw.to_native(df)
+    ```
+
+    Arguments:
+        native_dataframe: Raw dataframe from user.
```
FBruzzesi, line 356: Same as above

FBruzzesi (commented):
Looks great, incredible effort to replace all the expression examples already!

I left a few considerations for the docstrings of the `narwhalify` decorators.

One thing I am not biggest fan of is how the code for `narwhalify` and `narwhalify_method` is almost one to one 🙈 
[My previous comment](https://github.com/narwhals-dev/narwhals/pull/266#issuecomment-2156018451) addresses it, but we can iterate on this after merging

MarcoGorelli:
thanks for your comments 🙏 will address

> incredible effort to replace all the expression examples already!

😄 that's mostly vim macros (@baggiponte ya proud??)

EdAbati (commented):
Got curious when I saw this on Discord. It looks clean! 😊

narwhals/translate.py
MarcoGorelli: my copy-pasta 🍝

MarcoGorelli:
thanks a tonne both!

> One thing I am not biggest fan of is how the code for narwhalify and narwhalify_method is almost one to one 🙈
https://github.com/narwhals-dev/narwhals/pull/266#issuecomment-2156018451 addresses it, but we can iterate on this after merging

agree, I'm not proud of it 🙈 will try addressing, and if I'm too noob to figure it out, will open an issue so it's not forgotten

FBruzzesi:
@MarcoGorelli one idea could be to use [`inspect.ismethod`](https://docs.python.org/3/library/inspect.html#inspect.ismethod) vs [`inspect.isfunction`](https://docs.python.org/3/library/inspect.html#inspect.isfunction) to check which case we receive.

I also double checked how they behave for staticmethods (False for method, True for function) - I didn't check the internals but I would not be surprise to see if they check for `self` or `cls` as first argument name 😂

MarcoGorelli:
🤔 thanks, but I feel uneasy about using `inspect`...feels quite brittle if we're recommending this for use in prod applications, I'd suggest just accepting a little copy-and-pasting

MarcoGorelli:
🤔 however, this currently doesn't give users the freedom to pass in arguments in which order they like:
```python
@nw.narwhalify
def func(df, a):
    return df.select(nw.sum('a')+a)

def func(df, a):
    df = nw.from_native(df)
    df = df.select(nw.sum('a')+a)
    return nw.to_native(df)

print(func(df=pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]}), a=3))
print(func(a=3, df=pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})))
```
uncommenting the second `func`, this fails

Maybe we do need `inspect` to get the first positional argument then

narwhals/translate.py
MarcoGorelli: > otherwise explicitly pass it as first keyword argument

yeah this is the part i'm not keen on...it's unexpected to users if they _have_ to pass arguments in a certain order...will see what we can do about this

MarcoGorelli:
Just tried this out with tpc-h q1, and there doesn't seem to be any detectable overhead from this: https://www.kaggle.com/code/marcogorelli/narwhals-tpch-q1-s2/output?scriptVersionId=182410736

```
{5 items
"pandas[pyarrow][native]":float8.593971141999987
"pandas[pyarrow]":float8.490190933000008
}
```

(pandas[pyarrow] is running pandas via Narwhals, pandas[pyarrow][native] is running pandas directly)

It got _marginally_ faster, but I think that's just statistical fluke, there's no mechanism (yet!) in Narwhals that would make the code run faster. But it's encouraging that the overhead isn't detectable

MarcoGorelli:
Overall, perf looks good, Narwhal's overhead seems to not be detectable:

image

Sometimes things get a little faster, sometimes a little slower...I don't think the differences are meaningful. I'll add a "overhead" page to the docs, I think this is good to show people

> @MarcoGorelli one idea could be to use [inspect.ismethod](https://docs.python.org/3/library/inspect.html#inspect.ismethod) vs [inspect.isfunction](https://docs.python.org/3/library/inspect.html#inspect.isfunction) to check which case we receive.

in for a penny, in for a pound...if we're already using `inspect.signature`, might as well use `ismethod` / `isfunction` - thanks!

MarcoGorelli:
🤔 the inspect functions aren't working the way I thought they would (at least, once we inspect what gets passed to the decorator)
```python
(Pdb) p func
<function test_narwhalify_method_called.<locals>.Foo.func at 0x7f2adf1c8680>
(Pdb) p inspect.ismethod(func)
False
```

I'll leave that as a follow-up - but this is already, IMO, a nice ergonomic improvement for users. I'd say, let's ship it! 🚀 
Thanks all for comments and suggestions!

narwhals/translate.py
```diff
@@ -223,11 +237,192 @@ def get_native_namespace(obj: Any) -> Any:
...
+                eager_only=eager_only,
+                series_only=series_only,
+                allow_series=allow_series,
+            )
+            if args:
+                result = func(df, *args[1:], **kwargs)
+            else:
+                kwargs[first_key] = df
+                result = func(**kwargs)
+            return to_native(result, strict=strict)
```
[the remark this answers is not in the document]
FBruzzesi: Hey @hameerabbasi, thanks for the feedback and reporting this. 

That is totally correct, and even more generally one could return multiple values, dataframes or both, and it is hard to know how we should behave with each one of these.

For now `narwhalify` is intended as a shortcut for most common use case(s) of:

- One dataframe in
- Do operations with it
- Same dataframe out


I believe that for other cases it is better for the users to be _explicit_ in when, how and what to convert.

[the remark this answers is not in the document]
MarcoGorelli: thanks @hameerabbasi for your input!

indeed, as Francesco says, this is just intended as a shortcut for what's probably the most common case, it's not meant to work for every scenario

maybe we should add other shortcuts, like:
- narwhalify_all
- narwhalify_some
- narwhalify_series
- ...

I really want to obsess over user-friendliness with Narwhals

Some people sometimes claim that developer-facing APIs need to be maximally explicit rather than ergonomic, and I completely (but respectfully) disagree with them 😄

skrawcz:
@MarcoGorelli good stuff -- inspired by our conversation?

Some quick API thoughts:

```
@nw.narwhalify(parameters=None) # all that are a Series or DF type, -- default
@nw.narwhalify(parameters=["param1", "param3"]) # only named parameters are translated
```

then you do or don't do things depending on the return type:

```
@nw.narwhalify(convert_return=True) # default
@nw.narwhalify(convert_return=False) # don't convert return type
```

MarcoGorelli:
thanks @skrawcz !

hmm I like that! been thinking about this one...might be a good topic for our next community call

I don't expect the overhead to be too big here, it's just a few isinstance checks, and people shouldn't be calling Python function in hot for-loops anyway

PR #283, opened 2024-06-11
docs: add overhead section

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/docs/related.md
+++ b/docs/related.md
@@ -7,7 +7,7 @@ Standardised way of interchanging data between libraries, see
 
 ## DataFrame API Standard
 
-Project which aims to "provide a standard interface that encapsulates implementation details of dataframe libraries. This will allow users and third-party libraries to write code that interacts and operates with a standard dataframe, and not with specific implementations.", see [here](https://data-apis.org/dataframe-api/draft/).
+Now-discontinued project which aimed to "provide a standard interface that encapsulates implementation details of dataframe libraries. This will allow users and third-party libraries to write code that interacts and operates with a standard dataframe, and not with specific implementations.", see [here](https://data-apis.org/dataframe-api/draft/).
 
 The Narwhals author was originally involved, but left due to irreconcilable differences in vision.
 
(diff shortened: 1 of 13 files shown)


docs/related.md
```diff
@@ -7,7 +7,7 @@ Standardised way of interchanging data between libraries, see
 
 ## DataFrame API Standard
 
-Project which aims to "provide a standard interface that encapsulates implementation details of dataframe libraries. This will allow users and third-party libraries to write code that interacts and operates with a standard dataframe, and not with specific implementations.", see [here](https://data-apis.org/dataframe-api/draft/).
+Now-discontinued project which aimed to "provide a standard interface that encapsulates implementation details of dataframe libraries. This will allow users and third-party libraries to write code that interacts and operates with a standard dataframe, and not with specific implementations.", see [here](https://data-apis.org/dataframe-api/draft/).
```
FBruzzesi, line 10: > Now-discontinued project

Just out of curiosity: is that official?

docs/related.md
```diff
@@ -28,6 +28,8 @@ dispatches to 20+ backends. Some differences with Narwhals are:
   from non-pandas environments, the difference in the relative size increase is ~1000x
 - Narwhals only supports 4 backends, Ibis more than 20
 - Narwhals is focused on fundamental dataframe operations, Ibis on SQL backends
+- Narwhals has negligible overhead for dataframe backends, while
+  [Ibis' overhead is significant](https://github.com/ibis-project/ibis/issues/9345)
```
FBruzzesi, line 32: Maybe 
```suggestion
  [Ibis' overhead can be significant](https://github.com/ibis-project/ibis/issues/9345)
```
😇

Not sure if that's the case for all dataframe libraries

docs/related.md
MarcoGorelli: that's what they've told me

docs/related.md
MarcoGorelli: it's about 2x for Polars, 4x for pandas

thanks, addressed in b41b267

I'm honestly not sure whether to publish a comparison or whether that'd be perceive as "too mean"

MarcoGorelli: update: the Polars overhead isn't consistent, sometimes it's on-par, sometimes it's slower, sometimes slightly faster

i haven't checked, but I presume it's because they leverage Polars' SQL engine, whereas for pandas they do a direct translation?

Regarding their pandas backend, they don't recommend using it at all, and say it's mostly there for historical reasons. so I don't think a public comparison is warranted. But I think this is useful to answer the question "why not just use Ibis"

PR #291, opened 2024-06-11
feat: `strftime` for datetime namespace

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

-  `# noqa: A002` is because ruff complain about the "format" naming, but that's what polars uses.
- other comments in the code itself

### DIFF
--- a/tests/test_dt.py
+++ b/tests/test_dt.py
@@ -173,8 +173,8 @@ def test_total_minutes(timedeltas: timedelta) -> None:
     "fmt", ["%Y-%m-%d", "%Y-%m-%d %H:%M:%S", "%Y/%m/%d %H:%M:%S", "%G-W%V-%u", "%G-W%V"]
 )
 def test_dt_to_string(constructor: Any, fmt: str) -> None:
-    input_frame = nw.from_native(constructor(data))
-    input_series: nw.Series = input_frame["a"]  # type: ignore[assignment]
+    input_frame = nw.from_native(constructor(data), eager_only=True)
+    input_series = input_frame["a"]
 
     expected_col = [datetime.strftime(d, fmt) for d in data["a"]]
 
--- a/tests/test_dt.py
+++ b/tests/test_dt.py
@@ -172,14 +172,14 @@ def test_total_minutes(timedeltas: timedelta) -> None:
 @pytest.mark.parametrize(
     "fmt", ["%Y-%m-%d", "%Y-%m-%d %H:%M:%S", "%Y/%m/%d %H:%M:%S", "%G-W%V-%u", "%G-W%V"]
 )
-def test_dt_strftime(constructor: Any, fmt: str) -> None:
+def test_dt_to_string(constructor: Any, fmt: str) -> None:
     input_frame = nw.from_native(constructor(data))
     input_series: nw.Series = input_frame["a"]  # type: ignore[assignment]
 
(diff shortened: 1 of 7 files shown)


tests/test_dt.py
```diff
@@ -166,3 +166,20 @@ def test_total_minutes(timedeltas: timedelta) -> None:
...
     assert result_pdns == result_pl
+
+
+@pytest.mark.parametrize("constructor", [pd.DataFrame, pl.DataFrame])
+@pytest.mark.parametrize(
+    "fmt", ["%Y-%m-%d", "%Y-%m-%d %H:%M:%S", "%Y/%m/%d %H:%M:%S", "%G-W%V-%u", "%G-W%V"]
+)
+def test_dt_strftime(constructor: Any, fmt: str) -> None:
+    input_frame = nw.from_native(constructor(data))
+    input_series: nw.Series = input_frame["a"]  # type: ignore[assignment]
```
FBruzzesi, line 177: If I don't specify this, then line 181 complains that `input_series` may be series _or_ dataframe, and dataframe does not have a `.dt` attribute

tests/test_dt.py
```diff
@@ -166,3 +166,20 @@ def test_total_minutes(timedeltas: timedelta) -> None:
     assert result_pda == result_pl
     assert result_pdn == result_pl
     assert result_pdns == result_pl
+
+
+@pytest.mark.parametrize("constructor", [pd.DataFrame, pl.DataFrame])
+@pytest.mark.parametrize(
+    "fmt", ["%Y-%m-%d", "%Y-%m-%d %H:%M:%S", "%Y/%m/%d %H:%M:%S", "%G-W%V-%u", "%G-W%V"]
```
FBruzzesi, line 173: Polars has a more exhaustive suite. Let me know if you want to replicate or this is good enough.

P.s. to replicate, it would be good to have `.item()`

FBruzzesi:
Regarding CI failing:

```diff
-  "2020/03/01 00:00:00"
-  "2020/04/01 00:00:00"
-  "2020/05/01 00:00:00"
+  "2020/03/01 00:…
+  "2020/04/01 00:…
+  "2020/05/01 00:…
```
Not sure if... should I just trim the values?!

tests/test_dt.py
MarcoGorelli: `item` would be good 👍 I think this is fine anyway

MarcoGorelli (commented):
amazing, thanks!

MarcoGorelli:
In Polars, there's `.dt.to_string`, which does the same thing. Polars also has `.dt.strftime` as an alias...it's one of the very few aliases in Polars.

I think here (Narwhals), because we're stricter, we should probably just pick one

For the reverse operation, we have `.str.to_datetime`, not `.str.strptime`. So, I think for this one, we should probably use `.dt.to_string`?

MarcoGorelli (approved):
legend, thanks @FBruzzesi !

FBruzzesi:
I felt left out without a "ship it" gif 😂

SadGIF

MarcoGorelli:
😄 not sure why it's not showing

PR #300, opened 2024-06-14
Expression String Method - Contains

ugohuche:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -308,6 +308,15 @@ class PandasExprStringNamespace:
             "ends_with",
             suffix,
         )
+    
+    def contains(self, pat: str, regex: bool = True) -> PandasExpr:
+        return reuse_series_namespace_implementation(
+            self._expr,
+            "str",
+            "contains",
+            pat,
+            regex,
+        )
 
     def head(self, n: int = 5) -> PandasExpr:
         return reuse_series_namespace_implementation(
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -309,13 +309,13 @@ class PandasExprStringNamespace:
             suffix,
         )
 
-    def contains(self, pat: str, regex: bool = True) -> PandasExpr:
+    def contains(self, pattern: str, literal: bool) -> PandasExpr:
         return reuse_series_namespace_implementation(
             self._expr,
             "str",
             "contains",
-            pat,
-            regex,
+            pattern,
+            literal,
         )
 
     def head(self, n: int = 5) -> PandasExpr:
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -314,8 +314,8 @@ class PandasExprStringNamespace:
             self._expr,
             "str",
             "contains",
-            pat=pattern,
-            regex=not literal,
+            pattern,
+            literal,
         )
 
     def head(self, n: int = 5) -> PandasExpr:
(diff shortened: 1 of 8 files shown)


narwhals/_pandas_like/series.py
```diff
@@ -504,6 +504,9 @@ def ends_with(self, suffix: str) -> PandasSeries:
             self._series._series.str.endswith(suffix),
         )
 
+    def contains(self, pat: str, regex: bool = True) -> PandasSeries:
+        return self._series._from_series(self._series._series.str.contains(pat, regex))
```
FBruzzesi, line 508: This is possibly an opinionated comment, but rather than mapping parameters dynamically in `narwhals/expression.py`, I would do it in here. Namely:
```suggestion
    def contains(self, pattern: str, literal: bool) -> PandasSeries:
        return self._series._from_series(self._series._series.str.contains(pat=pattern, regex=not literal))
```

With that in place, the other implementations should follow almost seamlessly by reusing this one directly.
What do you think?

FBruzzesi:
Very cool feature! I was looking forward to this one! Thanks for implementing it ✨

narwhals/_pandas_like/series.py
FBruzzesi: Also, `regex = ...` should be explicitly specified, as otherwise the regex variable will be passed to the second argument of `str.contains` which is `case` and not `regex` itself ([docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.contains.html))

ugohuche: Yeah, this is much cleaner

ugohuche:
> Very cool feature! I was looking forward to this one! Thanks for implementing it ✨

I was more than happy to do it. 😊

ugohuche:
> thanks @ugohuche ! can you add a test please? it should go into `tests/expr/str/contains_test`

Will do

narwhals/_pandas_like/series.py
```diff
@@ -504,6 +504,11 @@ def ends_with(self, suffix: str) -> PandasSeries:
             self._series._series.str.endswith(suffix),
         )
 
+    def contains(self, pattern: str, literal: bool) -> PandasSeries:
```
MarcoGorelli, line 507: can you make `literal` keyword-only please?

MarcoGorelli (commented):
thanks, should be close

narwhals/_pandas_like/expr.py
```diff
@@ -309,6 +309,15 @@ def ends_with(self, suffix: str) -> PandasExpr:
             suffix,
         )
 
+    def contains(self, pattern: str, *, literal: bool = False) -> PandasExpr:
```
MarcoGorelli, line 312: let's try to not set defaults in `_pandas_like`, so we make sure they get passed down explicitly from above
```suggestion
    def contains(self, pattern: str, *, literal: bool) -> PandasExpr:
```

MarcoGorelli (approved):
seriously good work here, well done @ugohuche ! and thanks @FBruzzesi for reviewing

narwhals/_pandas_like/expr.py
ugohuche: Alright

ugohuche:
Thanks @MarcoGorelli. I learned some new things working on this one

PR #305, opened 2024-06-16
enh: `slice`, `starts_with`, `tail` methods for `str` namespace

FBruzzesi:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

None

## Checklist

- [x] Tests added 
- [x] Documented the changes

Implements `slice`, `starts_with` and `tail` methods in the str namespace for Expr and Series.

It is a quite large one, but slice implementation allowed to refactor head and tail, however using slice does not allow for negative indexes as polars does. Since they were not supported anyway in the first place, I moved to this implementation but happy to discuss how to work around it.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1,5 +1,7 @@
 site_name: Narwhals
 repo_url: https://github.com/narwhals-dev/narwhals.git
+watch:
+  - narwhals
 nav:
   - Home: index.md
   - Why: why.md
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -1809,7 +1809,7 @@ class SeriesStringNamespace:
             literal: If True, treats the pattern as a literal string.
                      If False, assumes the pattern is a regular expression.
 
-        Example:
+        Examples:
             >>> import pandas as pd
             >>> import polars as pl
             >>> import narwhals as nw
@@ -1832,6 +1832,7 @@ class SeriesStringNamespace:
             3     True
             4     None
             dtype: object
+
             >>> func(s_pl)  # doctest: +NORMALIZE_WHITESPACE
             shape: (5,)
             Series: '' [bool]
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -1847,12 +1847,86 @@ class SeriesStringNamespace:
             self._series._series.str.contains(pattern, literal=literal)
         )
 
+    def slice(self, offset: int, length: int | None = None) -> Series:
+        r"""
+        Create subslices of the string values of a Series.
+
+        Arguments:
+            offset: Start index. Negative indexing is supported.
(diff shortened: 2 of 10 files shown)


mkdocs.yml
```diff
@@ -1,5 +1,7 @@
 site_name: Narwhals
 repo_url: https://github.com/narwhals-dev/narwhals.git
+watch:
+  - narwhals
```
MarcoGorelli, line 4: ooh, what's this?

narwhals/series.py
```diff
@@ -1810,9 +1961,53 @@ def head(self, n: int = 5) -> Series:
                "zukky"
             ]
         """
-        if self._series._is_polars:
-            return self._series.__class__(self._series._series.str.slice(0, n))
-        return self._series.__class__(self._series._series.str.head(n))
+        return self._series.__class__(self._series._series.str.slice(0, n))
```
MarcoGorelli, line 1964: can we just call the underlying library directly here? so, keep it as it is, but in `_pandas_like`, you make `head` call `slice`

narwhals/series.py
```diff
@@ -1810,9 +1961,53 @@ def head(self, n: int = 5) -> Series:
...
+            shape: (4,)
+            Series: '' [str]
+            [
+               "atata"
+               "taata"
+               "atata"
+               "kkyun"
+            ]
+        """
+        return self._series.__class__(self._series._series.str.slice(-n))
```
MarcoGorelli, line 2010: same

MarcoGorelli (commented):
awesome, thanks @FBruzzesi ! 🏆

mkdocs.yml
FBruzzesi: Sorry it slipped in, it is for dynamic reloading if something is in the folder changes

narwhals/series.py
FBruzzesi: The min version of Polars we support does not have `str.head`. If you look at the diff, it was already implemented using slice for Polars.

I thought to implement `str.slice` for pandas as well and simplify the API.

WDYT?

FBruzzesi: I think the problem is more on the Polars side, that our min supported version does not implement `str.head`, hence we use slice.

I thought that implementing pandas `str.slice` would bring both a feature and unify the API call in here.

WDYT?

MarcoGorelli: aaah, I see, thanks, that's why we were using `.str[:n]` in the first place 😄 cool, looks fine then, thanks for explaining!

MarcoGorelli (approved):
nice one, feel free to merge!

PR #306, opened 2024-06-16
enh: series and expr round method

### LINKED ISSUE #302: [Enh]: Support for nw.round()

_No response_

Came up when trying TPC-H query Q11. Issue https://github.com/narwhals-dev/narwhals/issues/282

_No response_

_No response_

None

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #302 

## Checklist

- [x] Tests added 
- [x] Documented the changes

I just realized polars does not have a `DataFrame.round`, I guess `df.select(pl.all().round())` would do the trick :)

### DIFF
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -294,6 +294,9 @@ class PandasExpr:
     def tail(self, n: int) -> Self:
         return reuse_series_implementation(self, "tail", n)
 
+    def round(self: Self, decimals: int = 0) -> Self:
+        return reuse_series_implementation(self, "round", decimals)
+
     @property
     def str(self) -> PandasExprStringNamespace:
         return PandasExprStringNamespace(self)
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -294,7 +294,7 @@ class PandasExpr:
     def tail(self, n: int) -> Self:
         return reuse_series_implementation(self, "tail", n)
 
-    def round(self: Self, decimals: int = 0) -> Self:
+    def round(self: Self, decimals: int) -> Self:
         return reuse_series_implementation(self, "round", decimals)
 
     @property
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -500,6 +500,9 @@ class PandasSeries:
     def tail(self: Self, n: int) -> Self:
         return self._from_series(self._series.tail(n))
 
+    def round(self: Self, decimals: int = 0) -> Self:
+        return self._from_series(self._series.round(decimals=decimals))
+
     @property
     def str(self) -> PandasSeriesStringNamespace:
         return PandasSeriesStringNamespace(self)
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -500,7 +500,7 @@ class PandasSeries:
     def tail(self: Self, n: int) -> Self:
         return self._from_series(self._series.tail(n))
 
-    def round(self: Self, decimals: int = 0) -> Self:
+    def round(self: Self, decimals: int) -> Self:
         return self._from_series(self._series.round(decimals=decimals))
 
     @property
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -1519,6 +1519,49 @@ class Expr:
 
         return self.__class__(lambda plx: self._call(plx).tail(n))
 
+    def round(self, decimals: int = 0) -> Expr:
+        r"""
+        Round underlying floating point data by `decimals` digits.
+
+        Arguments
+            decimals: Number of decimals to round by.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> import pandas as pd
+            >>> import polars as pl
+            >>> data = {"a": [1.12345, 2.56789, 3.901234]}
+            >>> df_pd = pd.DataFrame(data)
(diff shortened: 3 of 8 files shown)


narwhals/_pandas_like/expr.py
```diff
@@ -294,6 +294,9 @@ def head(self, n: int) -> Self:
     def tail(self, n: int) -> Self:
         return reuse_series_implementation(self, "tail", n)
 
+    def round(self: Self, decimals: int = 0) -> Self:
```
MarcoGorelli, line 297: i should add a pre-commit rule here, but i think we should avoid default args in _pandas_like
```suggestion
    def round(self: Self, decimals: int) -> Self:
```

narwhals/_pandas_like/series.py
```diff
@@ -500,6 +500,9 @@ def head(self: Self, n: int) -> Self:
     def tail(self: Self, n: int) -> Self:
         return self._from_series(self._series.tail(n))
 
+    def round(self: Self, decimals: int = 0) -> Self:
```
MarcoGorelli, line 503:
```suggestion
    def round(self: Self, decimals: int) -> Self:
```

narwhals/expression.py
```diff
@@ -1519,6 +1519,49 @@ def tail(self, n: int = 10) -> Expr:
 
         return self.__class__(lambda plx: self._call(plx).tail(n))
 
+    def round(self, decimals: int = 0) -> Expr:
+        r"""
+        Round underlying floating point data by `decimals` digits.
```
MarcoGorelli, line 1524: i think pandas and polars may differ in how they round from midway (`.5`, `1.5`), shall we add a note about it?

tests/test_common.py
```diff
@@ -765,3 +765,19 @@ def test_item_value_error(
 ) -> None:
     with pytest.raises(ValueError, match=err_msg):
         nw.from_native(df_raw, eager_only=True).item(row, column)
+
+
+@pytest.mark.parametrize("constructor", [pd.DataFrame, pl.DataFrame])
+@pytest.mark.parametrize("decimals", [0, 1, 2])
+def test_round(constructor: Any, decimals: int) -> None:
```
MarcoGorelli, line 772: shall we move this to `tests/expr/round_test`? i think going forwards it's good to aim for one test per method

MarcoGorelli (commented):
awesome! just some very minor comments

narwhals/expression.py
FBruzzesi: Do you know if the difference only applies in the range `(-.5, .5)`? 

> shall we add a note about it?
If we know how to correct it manually, I think we should. Steering pandas behavior to polars' one

MarcoGorelli: i think pandas rounds to the nearest even, polars towards zero? something like that

not sure about how to correct it reliably - maybe we can just document it for now?

FBruzzesi: I added a note, but I would rather have a strategy to address cases like this one in the future 🤔

narwhals/expression.py
```diff
@@ -1519,6 +1519,58 @@ def tail(self, n: int = 10) -> Expr:
...
+        Arguments:
+            decimals: Number of decimals to round by.
+
+        Notes:
+            For values exactly halfway between rounded decimal values pandas and Polars behave differently.
+
+            pandas rounds to the nearest even value (e.g. -0.5 and 0.5 round to 0.0, 1.5 and 2.5 round to 2.0, 3.5 and
+            4.5 to 4.0, etc..).
+
+            Polars rounds towards the larger integer (e.g. -0.5 to -1.0, 0.5 to 1.0, 1.5 to 2.0, 2.5 to 3.0, etc..).
```
MarcoGorelli, line 1535:
```suggestion
            Polars rounds away from 0 (e.g. -0.5 to -1.0, 0.5 to 1.0, 1.5 to 2.0, 2.5 to 3.0, etc..).
```

narwhals/series.py
```diff
@@ -1707,6 +1707,54 @@ def tail(self: Self, n: int = 10) -> Self:
...
+        Arguments
+            decimals: Number of decimals to round by.
+
+        Notes:
+            For values exactly halfway between rounded decimal values pandas and Polars behave differently.
+
+            pandas rounds to the nearest even value (e.g. -0.5 and 0.5 round to 0.0, 1.5 and 2.5 round to 2.0, 3.5 and
+            4.5 to 4.0, etc..).
+
+            Polars rounds towards the larger integer (e.g. -0.5 to -1.0, 0.5 to 1.0, 1.5 to 2.0, 2.5 to 3.0, etc..).
```
MarcoGorelli, line 1723:
```suggestion
            Polars rounds away from 0 (e.g. -0.5 to -1.0, 0.5 to 1.0, 1.5 to 2.0, 2.5 to 3.0, etc..).
```

MarcoGorelli (approved):
awesome, thanks @FBruzzesi ! feel free to merge

I just made a tiny edit, as I think "away from 0" is more accurate than "towards the larger integer", as 0 is larger than -1 even though its magnitude is smaller. nitpicking here, sorry 😛 I think the examples you added explain it perfectly - the Narwhals docs may even end up being a useful reference of pandas vs Polars differences 😄

PR #307, opened 2024-06-17
perf: use fastpath in with_columns

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -7,6 +7,7 @@ from typing import Iterable
 from typing import Literal
 from typing import overload
 
+from narwhals._pandas_like.expr import PandasExpr
 from narwhals._pandas_like.utils import create_native_series
 from narwhals._pandas_like.utils import evaluate_into_exprs
 from narwhals._pandas_like.utils import horizontal_concat
@@ -170,8 +171,6 @@ class PandasDataFrame:
         *exprs: IntoPandasExpr,
         **named_exprs: IntoPandasExpr,
     ) -> Self:
-        from narwhals._pandas_like.expr import PandasExpr
-
         index = self._dataframe.index
         new_series = evaluate_into_exprs(self, *exprs, **named_exprs)
         # If the inputs are all Expressions which return full columns
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -170,12 +170,20 @@ class PandasDataFrame:
         *exprs: IntoPandasExpr,
         **named_exprs: IntoPandasExpr,
     ) -> Self:
+        from narwhals._pandas_like.expr import PandasExpr
+
         index = self._dataframe.index
         fast_path = True
         new_series = evaluate_into_exprs(self, *exprs, **named_exprs)
         # If the inputs are all Series (and not scalars), we can use
-        # a fast path (concat, instead of assign)
-        fast_path = all(s.len() > 1 for s in new_series)
+        # a fast path (concat, instead of assign).
+        # Only use fastpath if all inputs are expressions
+        # (todo: reproduce scikit-learn failure)
+        fast_path = (
+            all(s.len() > 1 for s in new_series)
+            and all(isinstance(x, PandasExpr) for x in exprs)
+            and all(isinstance(x, PandasExpr) for (_, x) in named_exprs.items())
+        )
 
         if fast_path:
             new_names = {s.name: s for s in new_series}
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -167,16 +167,32 @@ class PandasDataFrame:
 
     def with_columns(
         self,
-        *exprs: IntoPandasExpr | Iterable[IntoPandasExpr],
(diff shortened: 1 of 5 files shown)


MarcoGorelli:
this makes things ~2x as fast, for 100_000 rows, on the `add_lags` function from scikit-lego

tests/utils_test.py
```diff
@@ -24,6 +24,15 @@ def test_maybe_align_index_pandas() -> None:
     assert_frame_equal(nw.to_native(result), expected)
 
 
+def test_with_columns_sort() -> None:
```
MarcoGorelli, line 27: extra test for safety

narwhals/_pandas_like/dataframe.py
```diff
@@ -167,16 +167,33 @@ def filter(
...
+        if fast_path:
+            new_names = [s.name for s in new_series]
+            df = horizontal_concat(
+                [
+                    *[
+                        self._dataframe.loc[:, s]
+                        for s in self._dataframe.columns
+                        if s not in new_names
+                    ],
+                    *[validate_dataframe_comparand(index, s) for s in new_series],
```
MarcoGorelli, line 189: this doesn't preserve order...one sec

MarcoGorelli:
✅ tested this out with scikit-lego, and all the tests still pass

narwhals/_pandas_like/dataframe.py
```diff
@@ -167,16 +167,47 @@ def filter(
 
     def with_columns(
         self,
-        *exprs: IntoPandasExpr | Iterable[IntoPandasExpr],
+        *exprs: IntoPandasExpr,
         **named_exprs: IntoPandasExpr,
     ) -> Self:
+        from narwhals._pandas_like.expr import PandasExpr
```
FBruzzesi, line 173: Very nit: I wonder if it's worth to import it on top level. I would expect `with_columns` to be a pretty common method

narwhals/_pandas_like/dataframe.py
```diff
@@ -167,16 +167,47 @@ def filter(
...
         )
+
+        if fast_path:
+            new_names = {s.name: s for s in new_series}
+            to_concat = []
+            # Make sure to preserve column order
+            for s in self._dataframe.columns:
+                if s in new_names:
+                    to_concat.append(
+                        validate_dataframe_comparand(index, new_names.pop(s))
```
FBruzzesi, line 195: The "modifying while looping" scares the s**t out of me 🥲

narwhals/_pandas_like/dataframe.py
```diff
@@ -167,16 +167,47 @@ def filter(
...
-        df = self._dataframe.assign(
-            **{
-                series.name: validate_dataframe_comparand(self._dataframe.index, series)
-                for series in new_series
-            }
+        # If the inputs are all Expressions which return full columns
+        # (as opposed to scalars), we can use a fast path (concat, instead of assign).
+        # We can't use the fastpath if any input is not an expression (e.g.
+        # if it's a Series) because then we might be changing its flags.
+        # See `test_memmap` for an example of where this is necessary.
```
FBruzzesi, line 181: Holy moly that's a great catch! 

I am getting a bit confused on the wording "e.g. if it's a Series", doesn't `validate_dataframe_comparand` check if other is a Series? (I am reading such function for the first time though)

FBruzzesi (approved):
Looks like an amazing feature and performance booster.

The code checks out, I left comments more for my understanding than anything else :)

narwhals/_pandas_like/dataframe.py
MarcoGorelli: yup, this works, thanks! for some reason i thought i was gonna get a circular import

in general these imports should be quite cheap, if we can reduce the circular imports that's probably a good thing

narwhals/_pandas_like/dataframe.py
MarcoGorelli: 😄 I see what you mean, but we're not modifying what we're iterating over, but another variable

we iterate over `self._dataframe.columns`, and modify `new_names`

narwhals/_pandas_like/dataframe.py
MarcoGorelli: great question! so

- expr: IntoPandasExpr. this is something that can be converted to an expression, like:
  - a string (with a column name)
  - a PandasSeries
  - a PandasExpr
- evaluate_into_expr evaluates each input `expr` into a `PandasSeries`
- `validate_dataframe_comparand` returns the underlying `pandas.Series` from each `PandasSeries`, possibly replacing the `Index`, see https://narwhals-dev.github.io/narwhals/other/pandas_index/

At this stage, we're checking whether the input `exprs` are `PandasExpr`, before they have been converted to `PandasSeries`

MarcoGorelli:
thanks a tonne for your review!

I checked again, and this passes the scikit-lego test suite (but I'll check once more before releasing 😄 )

shipping then 🚀

Issue #230, opened 2024-05-30
Add narwhals.show_version


DeaMariaLeon:
Useful for bug reports as commented here:  https://github.com/narwhals-dev/narwhals/pull/226#discussion_r1620101833

PR #316, opened 2024-06-18
ci: Added Version

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues 

- Related issue #https://github.com/narwhals-dev/narwhals/issues/230
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Adapted from sklearn


DeaMariaLeon:
@MarcoGorelli Is it OK to have the dependencies hard-coded on [narwhals/functions.py](https://github.com/narwhals-dev/narwhals/pull/316/files#diff-0ef4e551729612f2a086037e5a418a025125491ef2dd6e1bbb0a7a2ed333ea81) lines 69-79?
```
deps = [
        "covdefaults",
        "pandas",
        "polars",
        "pre-commit",
        "pyarrow",
        "pytest",
        "pytest-cov",
        "hypothesis",
        "scikit-learn",
    ]
```
The coverage test is failing. Apparently the missing lines (that should be tested) are the ones that are executed when there is a `PackageNotFoundError`.  I wonder if  I have to add a smart way to check the dependencies in `_get_deps_info` (lines above). If that's not the case, I haven't found yet how to test for a package not found. 

```
for modname in deps:
        try:
            deps_info[modname] = version(modname)
        except PackageNotFoundError:  # noqa: PERF203
            deps_info[modname] = ""
    return deps_info

```

FBruzzesi:
Hey @DeaMariaLeon, I took a quick look.

For the failing test, I think adding `cudf` to the list of libraries would fix it as ci/cd does not installs cudf, hence it should enter such except block :)

MarcoGorelli:
thanks for doing this!

I'd suggest just reporting `__version__` for each package. I think the only ones we need are:
- pandas
- polars
- cudf
- modin
- pyarrow
- numpy
- narwhals

These should all have `__version__`, if we come across a package later which doesn't, we deal with that then

DeaMariaLeon:
@MarcoGorelli do you mean using `__version__` instead of `from importlib.metadata import version`?
To use __version__ I then should use your functions `get_polars()`, `get_pandas()`, right? 
If not, then the packages need to be imported, don't they?  -  I have not updated with your comment since I'm not sure. Just added the right packages. 

Thanks @FBruzzesi, adding `cudf` seem to do the trick.

MarcoGorelli (approved):
this is probably fine, thanks! i remembered there being some issue with importlib-metadata, but it might be been pre-Python3.8, and this passes CI, so is probably all good!

feel free to merge if you're happy with it

PR #317, opened 2024-06-18
feat: add DataFrame.rows

### LINKED ISSUE #285: [Enh]: Add support for DataFrame.rows

_No response_

I have a library which accepts a pandas dataframe and serializes it into a list of lists (`df.values.tolist()`) before processing it further. I would like to make the library work with polars dataframes as well and would like to use narwhals for the same.

For polars, `nw.DataFrame.rows()` would be the same as `pl.DataFrame.rows()`.
For pandas, `nw.DataFrame.rows()` would be `pd.DataFrame.values.tolist()`.

_No response_

yes

_No response_

Priyansh121096:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #285

## Checklist

- [x] Tests added 
- [x] Documented the changes

I've matched the API with https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.rows.html

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -628,6 +628,42 @@ class DataFrame(BaseFrame):
         """
         return super().columns
 
+    def rows(
+        self, *, named: bool = False
+    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
+        """
+        Returns all data in the DataFrame as a list of rows of python-native values.
+
+        By default, each row is returned as a tuple of values given in the same order
+        as the frame columns. Setting named=True will return rows of dictionaries instead.
+
+        Examples:
+            >>> import pandas as pd
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> df = {"foo": [1, 2, 3], "bar": [6.0, 7.0, 8.0], "ham": ["a", "b", "c"]}
+            >>> df_pd = pd.DataFrame(df)
+            >>> df_pl = pl.DataFrame(df)
+
+            We define a library agnostic function:
+
+            >>> def func(df_any, named):
+            ...     df = nw.from_native(df_any)
+            ...     return df.rows(named=named)
+
+            We can then pass either pandas or Polars to `func`:
+
+            >>> func(df_pd, False)
+            [(1, 6.0, 'a'), (2, 7.0, 'b'), (3, 8.0, 'c')]
+            >>> func(df_pd, True)
+            [{'foo': 1, 'bar': 6.0, 'ham': 'a'}, {'foo': 2, 'bar': 7.0, 'ham': 'b'}, {'foo': 3, 'bar': 8.0, 'ham': 'c'}]
+            >>> func(df_pl, False)
+            [(1, 6.0, 'a'), (2, 7.0, 'b'), (3, 8.0, 'c')]
+            >>> func(df_pl, True)
+            [{'foo': 1, 'bar': 6.0, 'ham': 'a'}, {'foo': 2, 'bar': 7.0, 'ham': 'b'}, {'foo': 3, 'bar': 8.0, 'ham': 'c'}]
+        """
+        return self._dataframe.rows(named=named)  # type: ignore[no-any-return]
+
     def with_columns(
         self, *exprs: IntoExpr | Iterable[IntoExpr], **named_exprs: IntoExpr
     ) -> Self:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -631,17 +631,17 @@ class DataFrame(BaseFrame):
 
     @overload
     def iter_rows(
-        self, *, named: Literal[False], buffer_size: int
+        self, *, named: Literal[False], buffer_size: int = ...
     ) -> Iterator[tuple[Any, ...]]: ...
 
     @overload
     def iter_rows(
-        self, *, named: Literal[True], buffer_size: int
+        self, *, named: Literal[True], buffer_size: int = ...
     ) -> Iterator[dict[str, Any]]: ...
 
     @overload
     def iter_rows(
-        self, *, named: bool, buffer_size: int
+        self, *, named: bool, buffer_size: int = ...
     ) -> Iterator[tuple[Any, ...]] | Iterator[dict[str, Any]]: ...
 
     def iter_rows(
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -647,19 +647,19 @@ class DataFrame(BaseFrame):
 
             We define a library agnostic function:
 
-            >>> def func(df_any, named):
+            >>> def func(df_any, *, named):
             ...     df = nw.from_native(df_any)
             ...     return df.rows(named=named)
 
             We can then pass either pandas or Polars to `func`:
 
-            >>> func(df_pd, False)
+            >>> func(df_pd, named=False)
             [(1, 6.0, 'a'), (2, 7.0, 'b'), (3, 8.0, 'c')]
-            >>> func(df_pd, True)
+            >>> func(df_pd, named=True)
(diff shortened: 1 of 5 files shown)


narwhals/dataframe.py
```diff
@@ -628,6 +628,42 @@ def columns(self) -> list[str]:
...
+            >>> import pandas as pd
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> df = {"foo": [1, 2, 3], "bar": [6.0, 7.0, 8.0], "ham": ["a", "b", "c"]}
+            >>> df_pd = pd.DataFrame(df)
+            >>> df_pl = pl.DataFrame(df)
+
+            We define a library agnostic function:
+
+            >>> def func(df_any, named):
```
MarcoGorelli, line 650: nitpick, but this is a "boolean trap" - can we make `named` keyword-only for the example please?

i.e.
```
func(df_any, *, named):
```

then, when you call it - `func(df_pd, named=False)`, etc

tests/test_common.py
```diff
@@ -314,6 +314,39 @@ def test_columns(df_raw: Any) -> None:
...
+            True,
+            [
+                {"a": 1, "b": 4, "z": 7.0},
+                {"a": 3, "b": 4, "z": 8.0},
+                {"a": 2, "b": 6, "z": 9.0},
+            ],
+        ),
+    ],
+)
+@pytest.mark.filterwarnings("ignore:Determining|Resolving.*")
```
MarcoGorelli, line 334: do we need this in this test?

tests/test_common.py
```diff
@@ -314,6 +314,39 @@ def test_columns(df_raw: Any) -> None:
...
+def test_rows(
```
MarcoGorelli, line 335: could you make a new file `tests/frame/rows_test.py` for this one please?

MarcoGorelli (commented):
this is amazing, well @Priyansh121096 for figuring it all out! 🙌 

I just have some minor comments really - do you have time / interest to address them? if not no worries, I can take it from here, lmk your preference

narwhals/dataframe.py
```diff
@@ -628,6 +628,42 @@ def columns(self) -> list[str]:
         """
         return super().columns
 
+    def rows(
+        self, *, named: bool = False
+    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
```
FBruzzesi, line 633: I think it would be a great addition to add type hinting for the two cases:

```suggestion     
    @overload
    def rows(
        self, *, named: Literal[True]
    ) -> list[dict[str, Any]]:

    @overload
    def rows(
        self, *, named: Literal[False]
    ) -> list[tuple[Any, ...]]:
    
    @overload
    def rows(
        self, *, named: bool = False
    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
    
    def rows(
        self, *, named: bool = False
    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
```

FBruzzesi (commented):
That is great! Thanks for the PR! I left a comment to add type hinting 😁 
P.s. it would require some additional imports

Priyansh121096:
Thanks for your comments. I'll address them soon.

narwhals/dataframe.py
Priyansh121096: Addressed.

narwhals/dataframe.py
Priyansh121096: Addressed.

tests/test_common.py
Priyansh121096: Removed.

tests/test_common.py
Priyansh121096: Addressed.

narwhals/dataframe.py
Priyansh121096: I was wondering if we want to put these overloads in an `if TYPE_CHECKING:` block. What do you think @FBruzzesi ?

FBruzzesi: Thanks for addressing it!

> What do you think @FBruzzesi ?

This is personal taste grey zone! I will let @MarcoGorelli make the final call 😁

MarcoGorelli (commented):
amazing stuff! thanks a tonne @Priyansh121096 

Just got a question about this functionality - instead of `DataFrame.rows`, couldn't we just add `DataFrame.iter_rows`?

Because then, if someone wants a list, then can do
```python
rows = list(df.iter_rows())
```

However, if someone is just going to loop over the rows, then
```
for row in df.iter_rows():
    # do something with `row`
```
would be more efficient than
```
for row in `df.rows()`:
    # do something with `row`
```

Priyansh121096:
> Just got a question about this functionality - instead of DataFrame.rows, couldn't we just add DataFrame.iter_rows?

@MarcoGorelli I was going to propose something similar (you beat me to it 😄). I was thinking we could expose both `rows` and `iter_rows` since the polars API does as well. 

One issue with this one is I'm not sure if pandas has a convenient equivalent to this when `named=True`. Rest of the cases are fine:
1. polars + `named=False` = `df.iter_rows(named=False)`
2. polars + `named=True` = `df.iter_rows(named=True)`
3. pandas + `named=False` = `df.itertuples(index=False, name=None)`
4. pandas + `named=True` = ?

Please let me know if you're aware of a pandas API which returns an iterator for iterating over rows as dictionaries.

FBruzzesi:
@Priyansh121096 I tried to play with the `into` argument of `.to_dict()` method, sadly with no success.

I guess that 4. can become `iter(df.to_dict("records"))`?!

MarcoGorelli:
How about using https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict (which , despite the underscore, is public):

something like (simplified)
```
def iter_rows(df):
    yield from (row._asdict() for row in df.itertuples(index=False))
```

Priyansh121096:
> I guess that 4. can become iter(df.to_dict("records"))?!

I feel like this defeats the purpose of using iter_rows over rows though.

> How about using https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict (which , despite the underscore, is public):

Amazing! This should work.

Priyansh121096:
@MarcoGorelli I'll raise another PR for iter_rows soon. Can we merge this one?

MarcoGorelli:
thanks - tbh I'm not really sure about DataFrame.rows, I might even suggest deprecating it altogether in Polars itself

could we just repurpose this one for DataFrame.iter_rows please? sorry for not having thought about this straight away

Priyansh121096:
> could we just repurpose this one for DataFrame.iter_rows please?

@MarcoGorelli pushed a change for this.

narwhals/dataframe.py
```diff
@@ -629,22 +630,32 @@ def columns(self) -> list[str]:
...
+        Returns an iterator over the DataFrame of rows of python-native values.
 
-        By default, each row is returned as a tuple of values given in the same order
-        as the frame columns. Setting named=True will return rows of dictionaries instead.
+        :param named:
+            By default, each row is returned as a tuple of values given in the same order as the frame columns.
+            Setting named=True will return rows of dictionaries instead.
+        :param buffer_size:
+            Determines the number of rows that are buffered internally while iterating over the data.
+        :see: https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.iter_rows.html
```
MarcoGorelli, line 658:
```suggestion
            :see: https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.iter_rows.html
```

MarcoGorelli (approved):
thanks @Priyansh121096 !

MarcoGorelli:
@pre-commit.ci autofix

PR #321, opened 2024-06-18
feat: add `narwhals.lit`

### LINKED ISSUE #309: [Enh]: support creating column with literal value

I'm currently looking at #204 , I found a place where this could be helpful

Similar to [`polars.lit`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.lit.html), it would be nice to have the capability to add a column with a constant value.

I think we could have a `narwhals.lit` function with a similar signature as `polars.lit`.

_No response_

Happy to make a PR!

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #309 

## Checklist

- [x] Tests added 
- [x] Documented the changes

- I've not implemented yet the `allow_object` param that `polars.lit` have. I'm not 100% sure on what it does and how to translate it in the pandas world. In which cases does polars use 'object'? 🤔
- What do you think about implementing something like `PandasSeries.from_iterable`? I saw that there are multple `PandasSeries(series_from_iterable(...), ...)` calls
- Do you think we should also test for more "complex" types like `list`? Polars supports it but I haven't seen a `list` dytpe in narwhals


FBruzzesi (commented):
PR looks great! Thank you so much for it, `nw.lit` is absolutely a needed feature.

The only unexpected behavior I can think of is what happens then passing a numpy array. Namely in polars the following happens:

```py
import numpy as np
import polars as pl

(
  pl.DataFrame({"a": [0, 1]})
  .with_columns(
      b = pl.lit(["x", "y"]),
      c = pl.lit(np.array([1, 2]))
  )
)
```
```
shape: (2, 3)
┌─────┬────────────┬─────┐
│ a   ┆ b          ┆ c   │
│ --- ┆ ---        ┆ --- │
│ i64 ┆ list[str]  ┆ i64 │
╞═════╪════════════╪═════╡
│ 0   ┆ ["x", "y"] ┆ 1   │
│ 1   ┆ ["x", "y"] ┆ 2   │
└─────┴────────────┴─────┘
```

MarcoGorelli (commented):
I'm _seriously_ impressed you figured this all out. Well done!

will do careful check for edge cases

FBruzzesi:
Additional note: when creating a new column without using `lit`, narwhals already handles numpy vs other well. 

```py
import narwhals as nw
import numpy as np
import polars as pl

(
  pl.DataFrame({"a": [0, 1]})
  .pipe(nw.from_native)
  .with_columns(
      b = ["x", "y"],
      c = np.array([1, 2]),
  )
  .pipe(nw.to_native)
)
```
```terminal
shape: (2, 3)
┌─────┬────────────┬─────┐
│ a   ┆ b          ┆ c   │
│ --- ┆ ---        ┆ --- │
│ i64 ┆ list[str]  ┆ i64 │
╞═════╪════════════╪═════╡
│ 0   ┆ ["x", "y"] ┆ 1   │
│ 1   ┆ ["x", "y"] ┆ 2   │
└─────┴────────────┴─────┘
```
@MarcoGorelli  I remember we discussed this since it was needed in lego, but I can't find the conversation right now

narwhals/_pandas_like/namespace.py
```diff
@@ -108,6 +108,30 @@ def all(self) -> PandasExpr:
             implementation=self._implementation,
         )
 
+    def lit(self, value: Any, dtype: dtypes.DType | None) -> PandasExpr:
+        def _lit_pandas_series(df: PandasDataFrame) -> PandasSeries:
+            pandas_series = PandasSeries(
+                series_from_iterable(
+                    [value for _ in range(len(df._dataframe.index))],
+                    name="lit",
+                    index=df._dataframe.index,
```
MarcoGorelli, line 117: can you just do
```suggestion
                    [value],
                    name="lit",
                    index=df._dataframe.index[0:1],
```
so the `validate_series_comparand` / `validate_dataframe_comparand` can take care of broadcasting? I'd expect it to be more efficient than allocating `len(df)` copies of `value`

EdAbati:
Thank you both for the comments! :)

a. Passing a `list` into `lit` seems to work.

```python
import polars as pl
import pandas as pd
import narwhals as nw

df = pd.DataFrame({"a": [0, 1]})
df_pl = pl.DataFrame(df)

@nw.narwhalify
def func(df):
    return df.with_columns(c=nw.lit([1, 2]))
```

```
>>> func(df_pl)
shape: (2, 2)
┌─────┬───────────┐
│ a   ┆ c         │
│ --- ┆ ---       │
│ i64 ┆ list[i64] │
╞═════╪═══════════╡
│ 0   ┆ [1, 2]    │
│ 1   ┆ [1, 2]    │
└─────┴───────────┘

>>> func(df)
   a       c
0  0  [1, 2]
1  1  [1, 2]
```

The only "issue" is that we cannot specify the correct `DType` (i.e. `nw.lit([1, 2], dtype=...)` just yet, because we don't have a list dtype. Or am I missing something?

b. Passing a `np.array` makes pandas and polars behave differently at the moment. 🤔

```python

@nw.narwhalify
def func(df):
    import numpy as np

    return df.with_columns(c=nw.lit(np.array([1, 2])))
```

```
>>> func(df_pl)
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ c   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 0   ┆ 1   │
│ 1   ┆ 2   │
└─────┴─────┘

>>> func(df)
   a       c
0  0  [1, 2]
1  1  [1, 2]
```

narwhals/_pandas_like/namespace.py
EdAbati: Thanks! I was expeciting there was a more efficient way to do this. 😅

Stupid question: are these `validate_series_comparand` / `validate_dataframe_comparand` implemented by the various libraries?

MarcoGorelli:
thanks for checking that carefully - I think we should disallow passing numpy arrays to `lit`? at least for now - if we just raise an error for now, we can always loosen the behaviour later if we want to, whilst preserving backwards compatibility 😇 

>  just yet, because we don't have a list dtype. Or am I missing something?

you're right, we don't :) largely because potential consumers of Narwhals probably wouldn't be using it, given that in pandas pre-pyarrow it would just be `object` dtype ... but we should implement it properly in Narwhals eventually, it's definitely worthwhile

narwhals/_pandas_like/namespace.py
EdAbati: had to revert this for "list" literal value 😕 please let me know if there is a better way

narwhals/_pandas_like/namespace.py
```diff
@@ -108,6 +108,37 @@ def all(self) -> PandasExpr:
...
         )
 
+    def lit(self, value: Any, dtype: dtypes.DType | None) -> PandasExpr:
+        def _lit_pandas_series(df: PandasDataFrame) -> PandasSeries:
+            if isinstance(value, (list, tuple)):
+                data = [value for _ in range(len(df._dataframe))]
+                index = df._dataframe.index
+            else:
+                data = [value]
+                index = [0]
```
MarcoGorelli, line 118: we don't have nested datatypes in Narwhals yet, shall we save on some complexity (until we have a use case) and just do
```
if isinstance(value, (list, tuple)):
    raise NotImplementedError(<some informative error message>)
data = [value]
index = df._dataframe.index[0:1]
```
?

tests/test_common.py
```diff
@@ -226,6 +227,59 @@ def test_sum_all(df_raw: Any) -> None:
...
+        else [(1, 2) for _ in range(3)]  # type: ignore[misc]
+    )
+    expected = {
+        "a": [1, 3, 2],
+        "b": [4, 4, 6],
+        "z": [7.0, 8.0, 9.0],
+        "list": [[1, 2] for _ in range(3)],
+        "tuple": expected_tuple,
+    }
+    compare_dicts(result_native, expected)
```
MarcoGorelli, line 269: same, shall we just combine this with the test below it?

tests/utils.py
```diff
@@ -21,7 +21,11 @@ def compare_dicts(result: Any, expected: dict[str, Any]) -> None:
         for key in result.columns:
             assert key in expected
     for key in expected:
-        for lhs, rhs in zip_longest(result[key], expected[key]):
+        result_key = (
+            result[key].to_list() if hasattr(result[key], "to_list") else result[key]
+        )
+
+        for lhs, rhs in zip_longest(result_key, expected[key]):
```
MarcoGorelli, line 28: then we don't need to modify this

MarcoGorelli (commented):
thanks, just some minor comments (I think we can already get this in, and worry about the complexity of nested data types later?)

just one question - how did you go about figuring this all out? did you read the "how it works" page? did you use a debugger / read the code very carefully?
seeing as you probably spent some time on it - any tips on how to make it clearer? personally, any extra developer-facing documentation / refactoring / better variable names / etc. is 110% welcome. I think Narwhals is fairly easy to use, but I'd like to get to the point where it also fairly easy to contribute to

narwhals/_pandas_like/namespace.py
EdAbati: Sounds good! Let's keep it simple :)

Out of curiosity, is there a difference between `index = [0]` and `index = df._dataframe.index[0:1]`? Both seems to work fine but I guess I am missing something. When is the `index` used?

EdAbati:
> how did you go about figuring this all out? did you read the "how it works" page? did you use a debugger / read the code very carefully?

Kind of both. I read the "how it work works" page out of curiosity, when I found out about the project. To contribute to this I read the code, searching for similar implementations and used the debugger.

I found the code easy to follow, also thanks to the typing. So 👏 👏  to you and the rest of the team

The contributing guide and process (with the template issues and PRs) is also nice!

Maybe we can add some documentation about the structure of the unit tests. I am used to find the same module/submodule structure as in the library. Here it seems the tests cases in the same submodule are in more scripts. I was a bit unsure where to add mine 😅

Regarding variable names, the only thing I noticed is the one I mentioned above. I was unsure why we needed to use `PandasSeries`, before realising that `series_from_iterables` was returning a ‘native series’.

MarcoGorelli (approved):
thanks a tonne @EdAbati !

MarcoGorelli:
> Maybe we can add some documentation about the structure of the unit tests. I am used to find the same module/submodule structure as in the library. Here it seems the tests cases in the same submodule are in more scripts. I was a bit unsure where to add mine 😅

most of the structure of the unit tests is quite odd. probably because I did it in a hurry as a weekend project. if you fancy improving the tests structure, that's very welcome - i've added you to the team so feel free to submit anything you think is an improvement and merge anything you think is ready

EdAbati:
Thank you for the feedback and for adding me to the team! This is "lit" 🔥 (sorry had to make the joke)

narwhals/_pandas_like/namespace.py
MarcoGorelli: when aligning different objects (in the `validate_`) functions. the difference is just that if `df` is already length 1, then using `df._dataframe.index[0:1]` won't cause an unnecessary `set_axis` call

Issue #320, opened 2024-06-18
Extend `narwhalify` decorator


MarcoGorelli:
As discussed in this PR https://github.com/narwhals-dev/narwhals/pull/266, we should probably extend `narwhalify` with extra capabilities

Maybe - only maybe - we should aim for it to be the primary way that people interact with Narwhals? Not totally sure on this one

FBruzzesi:
It may be worth discussing a few strategy that we can have. I see the following:

1. (Try to) convert every argument to narwhals with `strict=False` flag (both in input and output)
2. Allow user to specify which arguments and output to convert to and back from narwhals.
    - Leaves more freedom to the user, if some very custom operation need to be performed.
3. Use type checks to convert dataframe/series only objects 

I personally prefer option 1., it is possibly opinionated but also a very clear promise that everything that can be converted, will be converted.

Extra: in each case we should check that the dataframe objects come with the same backend

WDYT?

FBruzzesi:
[WIP attempt ](https://github.com/narwhals-dev/narwhals/compare/main...enh/extend-narwhalify) for 1. 

- Implementation is 90% there I would say
- Mypy needs some love
- Docs need to be updated
- Tests need to be updated

Issue #274, opened 2024-06-10
clean: refactor `narwhalify` and `narwhalify_method` decorators


MarcoGorelli:
There was a suggestion here to reduce copy-and-pasting: https://github.com/narwhals-dev/narwhals/pull/266#issuecomment-2156018451

PR #328, opened 2024-06-21
Enh: extend `narwhalify`

FBruzzesi:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #320 and #274

## Checklist

- [x] Tests added 
- [x] Documented the changes

There was some _heated_ discussion on how we should behave on the conversion for the return part:

The risk I see with this change is that:
- `strict=False` will allow for any input and any output
- while this is ok for the input part, not all output are created equal
- the main risk is to have someone returning something like `df1, df2` which will be returned as `nw.DataFrame`s

Two other notes for the changes:

- The moving from `from narwhals.translate import get_cudf` to `from narwhals.dependencies import get_cudf` (and similars), is due to the fact that as I was experimenting another approach I was getting circular imports. 
- Series file has a big diff, but now we can use `@narwhalify` in all operations, so why not :)


docs/basics/complete_example.md
```diff
@@ -20,11 +20,15 @@ We're going to take in a dataframe, and return a dataframe of the same type.
 Therefore, we use `@nw.narwhalify_method` (the counterpart to `@nw.narwhalify` which is
 meant to be used for methods):
 
+!!! info
+    As of version 0.y.z, `@nw.narwhalify_method` has been deprecated, and `nw.narwhalify`
```
FBruzzesi, line 24: Reminder for future self: Change version

I didn't want to pin it as we ship very fast, and this PR could stay here for longer than expected

docs/basics/complete_example.md
```diff
@@ -20,11 +20,15 @@ We're going to take in a dataframe, and return a dataframe of the same type.
...
+    can be used on methods as well.
+
 ```python
 import narwhals as nw
 
 class StandardScaler:
-    @nw.narwhalify_method
+    @nw.narwhalify_method  # since v0.y.z, this is deprecated in favor of @nw.narwhalify
```
FBruzzesi, line 31: Reminder for future self: Change version

docs/basics/dataframe.md
```diff
@@ -150,3 +149,42 @@ Let's try it out:
     df = pl.LazyFrame({'a': [1, 1, 2], 'b': [4, 5, 6]})
     print(func(df).collect())
     ```
+
+## Example 4: multiple inputs
+
+As of narwhals v0.y.z, `nw.narwhalify` can be used to decorate functions that take
```
FBruzzesi, line 155: Reminder for future self: Change version

docs/basics/dataframe.md
```diff
@@ -150,3 +149,42 @@ Let's try it out:
     df = pl.LazyFrame({'a': [1, 1, 2], 'b': [4, 5, 6]})
     print(func(df).collect())
     ```
+
+## Example 4: multiple inputs
```
FBruzzesi, line 153: I hope this section exemplify the new behavior, if anyone has more ideas I am all ears

narwhals/translate.py
```diff
@@ -294,131 +304,64 @@ def func(df):
...
+            kwargs = {
+                name: from_native(
+                    value,
+                    strict=strict,
+                    eager_only=eager_only,
+                    series_only=series_only,
+                    allow_series=allow_series,
+                )
+                for name, value in kwargs.items()
+            }
```
FBruzzesi, line 327: @MarcoGorelli should we check that passed dataframe/series are coming from the same backend?
It may be not so trivial, but on the other hand, it could speed up parsing by avoiding loading backend libraries N times and some number of checks.

MarcoGorelli: I think we have a check in `join` that objects are from the same library

not sure if it's worth it though, maybe it can be removed, and we just assume it's all from the same library?

FBruzzesi: > not sure if it's worth it though, maybe it can be removed, and we just assume it's all from the same library?

I would be fine with giving responsibility to the users for this. It is also significantly easier to validate once one knows which arguments are dataframe/series, rather than dataframe/series or anything else.

> I think we have a check in join that objects are from the same library

I think it is significantly more complex as:

- non-dataframe elements can be passed
- once the first dataframe is found, need to check all other arguments to have the same backend, but only if dataframes, otherwise just pass them along

I didn't put much time on this yet though. I will give it a try soon-ish 👌

MarcoGorelli: right, thanks - if anyone's mixing different libraries together they'll likely need to do some data conversion anyway, so narwhals probably isn't the library for their use case anyway, OK to just document the assumption

FBruzzesi: One way of doing it is:

- parse everything (as current implementation), then
- validate backends

```py
backends = {
    b() for v in [*args, *kwargs.values()] if (b:=getattr(v, "__native_namespace__", None))
}

if len(backends) > 1:
    msg = "Found multiple backends. Make sure that all inputs come from the same one."
    raise ValueError(msg)
```

The issue I have with this is:

- It adds some overhead
- It fails with Series, as they do not implement `__native_namespace__`

WDYT?

MarcoGorelli: I think this should be fine - the overhead should be minimal, as we'd just be loading modules which the user has already imported. We can also add `__native_namespace__` to `Series`

We can time it, but my intuition is that it's probably low enough to be acceptable

FBruzzesi: Just pushed such change, which includes adding `__native_namespace__` to `Series` 😎

docs/basics/column.md
```diff
@@ -148,8 +148,8 @@ def my_func(df):
     print(my_func(df))
     ```
 
-Note that, this time, we couldn't use `@nw.narwhalify`, as the final step in
-our function wasn't `nw.to_native`, so we had to explicitly use `nw.from_native`
-as the first step. In general, we recommend using the decorator where possible,
-as it looks a lot cleaner, and only using `nw.from_native` / `nw.to_native` explicitly
-when you need them.
+Note that, even though the output of our function is not a dataframe nor a series, we can
+still use `narwhalify` with `strict=False` flag (which is actually the default behavior).
```
MarcoGorelli, line 152: maybe we should just show the default? this is probably one of the first pages newcomers will see, maybe it's better to keep things simple?

docs/basics/complete_example.md
MarcoGorelli: from https://github.com/search?q=narwhalify_method&type=code, it looks like nobody's using it

maybe we can just get rid of it right away

narwhals/dataframe.py
```diff
@@ -10,11 +10,11 @@
...
 
 from narwhals._pandas_like.dataframe import PandasDataFrame
+from narwhals.dependencies import get_cudf
+from narwhals.dependencies import get_modin
+from narwhals.dependencies import get_pandas
 from narwhals.dependencies import get_polars
 from narwhals.dtypes import to_narwhals_dtype
-from narwhals.translate import get_cudf
-from narwhals.translate import get_modin
-from narwhals.translate import get_pandas
```
MarcoGorelli, line 17: nice one, been meaning to do this

narwhals/translate.py
MarcoGorelli: I like what you've done here

pyproject.toml
```diff
@@ -17,6 +17,10 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 
+dependencies = [
+    "typing-extensions>=4.4.0; python_version < '3.10'",
+]
```
MarcoGorelli, line 22: why do we need this? atm I think it's only ever used within `if TYPE_CHECKING` blocks, so we shouldn't need it

narwhals/series.py
```diff
@@ -86,9 +91,9 @@ def shape(self) -> tuple[int]:
 
             We define a library agnostic function:
 
-            >>> def func(s_any):
-            ...     s = nw.from_native(s_any, series_only=True)
-            ...     return s.shape
+            >>> @nw.narwhalify
+            ... def func(s_any):
+            ...     return s_any.shape
```
MarcoGorelli, line 97: are we sure about `s_any` here?

- from the function's body perspective, `s: nw.Series`
- but, from the point of view of anyone reading the function, `s_any` is not a Narwhals series, it's their native object

Maybe we need a `IntoNarwhalsSeries` / `IntoNarwhalsDataFrame` / `IntoNarwhals/LazyFrame` / `IntoNarwhals` type aliases in `nw.typing`, so that people can annotate functions in a way that's legible to the user (i.e. "this function accepts anything which is convertible to Narwhals), but in the function body, their IDE can help them with type completion and all

Anyway, one for a separate PR, out-of-scope for this one, just thinking about it

MarcoGorelli (commented):
Wonderful, thanks! Really on board with the general idea, I think this will make it much easier to adopt Narwhals!

Just have minor comments

docs/basics/complete_example.md
FBruzzesi: It looks as if it never existed now 😂

pyproject.toml
FBruzzesi: I think I added when trying another implementation. Got rid of it! Thanks for spotting

narwhals/series.py
FBruzzesi: > from the function's body perspective, s: nw.Series

yes

> but, from the point of view of anyone reading the function, s_any is not a Narwhals series, it's their native object

more yes

Type hinting will need some more love all around I think. Here is exactly as you mention, from the user point of view, any native object can be passed, while the function will work with the narwhals objects - but that's taken care of by the decorator

docs/basics/complete_example.md
```diff
@@ -47,7 +46,7 @@ doesn't either.
 
 Note how here, we're not returning a dataframe to the user - we just take a dataframe in, and
 store some internal state. Therefore, we use `nw.from_native` explicitly, as opposed to using the
-utility `@nw.narwhalify_method` decorator.
+utility `@nw.narwhalify` decorator.
```
MarcoGorelli, line 49: we can now use `narwhalify` for this example, right? if it returns `None`, and the default is `strict=False`, then I think it'll still work? it might be easier to teach to use narwhalify everywhere?

MarcoGorelli (commented):
thanks!

one comment on the tutorial + the version number in the example needs filling in

MarcoGorelli (approved):
nice, thanks @FBruzzesi !

MarcoGorelli:
feel free to merge if happy - also, as a follow-up, we maybe this should be the suggested "how to" on the readme, as opposed to `from_native` / `to_native`?

PR #331, opened 2024-06-22
feat: add nw.StableAPI, and perfect-backwards-compatibility policy

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

~it might be a bit tricky to get this to work for `Expr`...gonna try though~

~Still need to add all the top-level `narwhals.*` methods...~

still need to fill in the docstrings


narwhals/series.py
```diff
@@ -1835,11 +1847,14 @@ def get_categories(self) -> Series:
                "mango"
             ]
         """
-        return self._series.__class__(self._series._series.cat.get_categories())
+        return self._series.__class__(
+            self._series._series.cat.get_categories(), api_version=self._api_version
+        )
```
FBruzzesi, line 1852: Is it worth using 
```suggestion
        return self._from_series(self._series._series.cat.get_categories())
```
here and in many other `self._series.__class__` calls? It would avoid passing `self._api_version` all the time

narwhals/translate.py
```diff
@@ -108,27 +121,30 @@ def from_native(
...
 
-def from_native(
+def from_native(  # noqa: PLR0913
     native_dataframe: Any,
     *,
     strict: bool = True,
     eager_only: bool | None = None,
     series_only: bool | None = None,
     allow_series: bool | None = None,
+    api_version: str | None = None,
```
FBruzzesi, line 147: I know it is still WIP but commenting just in case:

- The argument is not documented in the docstring
- Is there a reason to be `None` instead of the default which is used otherwise?
- Maybe most important: I am not expecting to support any single version, I think we should limit and validate the value passed through.
 
Example combining the latter two points:

```py
Versions = Literal["0.20", "1.0"]
...
    
def from_native(
    ...
    api_version: Versions  = "0.20",
):

    if api_version not in get_args(Versions): Raise(...)
```

FBruzzesi (commented):
Trying to keep up with this one and left a couple of comments along the way 😇

narwhals/series.py
MarcoGorelli: yup, good one thanks!

narwhals/translate.py
MarcoGorelli: yup, thanks! this is just a start, but it works well enough that the scikit-lego / timebased-cv tests suites pass

I'll post the diff in the linked issue to solicit feedback

MarcoGorelli:
Found a blocker - this breaks static typing:
```python
import narwhals

nw = narwhals.StableAPI('0.20')


def func(df: nw.DataFrame) -> nw.DataFrame:
    return df
```
```
t.py:20: error: Name "nw.DataFrame" is not defined  [name-defined]
```

narwhals/_translate.py
```diff
@@ -0,0 +1,1323 @@
...
+        We follow development of pandas and Polars _very_ closely, and so we expect to be
+        able to make compatible Narwhals releases well in advance of pandas/Polars ones.
+    """
+
+    api_version: API_VERSION
+
+    DataFrame = DataFrame
+    LazyFrame = LazyFrame
+    Series = Series
+    Expr = Expr
```
FBruzzesi, line 542: Does this solves the static typing issues?

narwhals/_translate.py
```diff
@@ -0,0 +1,1323 @@
...
+    Int8 = dtypes.Int8
+    UInt64 = dtypes.UInt64
+    UInt32 = dtypes.UInt32
+    UInt16 = dtypes.UInt16
+    UInt8 = dtypes.UInt8
+    Float64 = dtypes.Float64
+    Float32 = dtypes.Float32
+    String = dtypes.String
+    Categorical = dtypes.Categorical
+    Datetime = dtypes.Datetime
```
FBruzzesi, line 1311: Should we move this up at the start of the class?

FBruzzesi (commented):
That's an insane effort! I am slightly concerned with maintainability  🙈

Also left a few comments here and there 😇

narwhals/_translate.py
MarcoGorelli: nope 😭

narwhals/_expression.py
```diff
@@ -28,9 +32,13 @@ def extract_native(expr: Expr, other: Any) -> Any:
 
 
 class Expr:
-    def __init__(self, call: Callable[[Any], Any]) -> None:
+    def __init__(self, call: Callable[[Any], Any], api_version: API_VERSION) -> None:
         # callable from namespace to expr
         self._call = call
+        self._api_version = api_version
+
+    def _from_expr(self, call: Callable[[Any], Any]) -> Self:
```
EdAbati, line 40: what do you think about changing the name to something like
```suggestion
    def _from_native_expr(self, call: Callable[[Any], Any]) -> Self:
```

narwhals/_expression.py
```diff
@@ -28,9 +32,13 @@ def extract_native(expr: Expr, other: Any) -> Any:
 
 
 class Expr:
-    def __init__(self, call: Callable[[Any], Any]) -> None:
+    def __init__(self, call: Callable[[Any], Any], api_version: API_VERSION) -> None:
         # callable from namespace to expr
         self._call = call
+        self._api_version = api_version
```
EdAbati, line 38: I see that when `Expr` is instantiated below, we do many `api_version or DEFAULT_API_VERSION`.
Would it make sense to move the logic here?
like
```python
class Expr
    def __init__(self, call: Callable[[Any], Any], api_version: API_VERSION|None = None) -> None:
        ...
        self._api_version = api_version or DEFAULT_API_VERSION
```

EdAbati (commented):
This is very very impressive! @MarcoGorelli 🤩

I left some small suggestions mostly around naming.

I think some parts of these PRs can also be merged indipendently from the `StableAPI` design. For example, the renaming of `dataframe.py`, `series.py`, `translate.py` to `_` private methods, or the `LazyFrame` refactor that doesn't try to map to pandas-like df anymore, or the move from calling directly `nw.DataFrame` etc to using `nw.from_native` in tests and docs.

It may be an idea that helps with merging this quicker and to have the `StableApi` PR diff a bit smaller

narwhals/_expression.py
MarcoGorelli: nice!

narwhals/_expression.py
MarcoGorelli: yeah probably

MarcoGorelli:
thanks @EdAbati !

I think this StableAPI class is probably not the way to go unfortunately, and that `narwhals.stable.v0_20` / `narwhals.stable.v1_0` would work a lot better

fancy making a separate PR to incorporate the good parts of this PR which you picked up on?

EdAbati:
ah 😕 is it because of the typing problem? yep I can do do some PRs, coming in the next days

MarcoGorelli:
thanks!

btw, if anything I suggest doesn't interest you, please do feel free to say 'no' (not saying that that's necessarily the case here, just in general)

MarcoGorelli:
Closing in favour of https://github.com/narwhals-dev/narwhals/pull/410, which is working a lot better

thanks for having taken a look here!

PR #336, opened 2024-06-25
refactor: Refactor and reorganise tests

EdAbati:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [ ] Documented the changes

As mentioned in #321 , this is a proposal to clean up a bit the structure of the tests. I tried to keep the PR as small and manageble as possible, let me know if you think I should split it further. :)

The logic:
- follow the structure of the modules in the library: naming convention `test_<modulename>`. If a module has multiple test files it becomes a folder. 
- anything else has `<name of test>_test.py`. With the exception of `test_common` and `test_invalid`.
- grouped all expression related tests in the same `expr` folder. Here there are subfolders for each `expr` context. e.g. `cat`, `str`. `dt` has every tests into `test_dt.py`.
- move to `utils.py` the code that was repeated: `maybe_get_modin_df`

Possible extra cleaning (that may make the diff bigger):
- rename `expr` -> `expression`, `frame` -> `dataframe` to match the module names
- create a folder for tpch unittest as hypothesis have (there is just 1 for now though)
- move `frame/test_invalid` into `frame/test_common`?
- anything else?


tests/test_group_by.py
```diff
@@ -1,14 +1,19 @@
...
 import polars as pl
 import pytest
 
 import narwhals as nw
 from tests.utils import compare_dicts
 
-df_pandas = pd.DataFrame({"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]})
-df_lazy = pl.LazyFrame({"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]})
+data_1 = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+data_2 = {"a": [1, 1, 3], "b": [4, 4, 6], "c": [7, 8, 9]}
```
EdAbati, line 13: any better names for these? 🤔

`data_2` is the data used for the tests that were in `group_by_test.py`

tests/utils.py
```diff
@@ -1,12 +1,18 @@
...
 from typing import Any
 from typing import Iterator
 from typing import Sequence
 
+if TYPE_CHECKING:
+    import pandas as pd
 
-def zip_longest(left: Sequence[Any], right: Sequence[Any]) -> Iterator[Any]:
+
+def zip_strict(left: Sequence[Any], right: Sequence[Any]) -> Iterator[Any]:
```
FBruzzesi, line 15: Nice naming 👌

tests/utils.py
```diff
@@ -21,10 +27,25 @@ def compare_dicts(result: Any, expected: dict[str, Any]) -> None:
...
+        for lhs, rhs in zip_strict(result[key], expected[key]):
             if isinstance(lhs, float) and not math.isnan(lhs):
                 assert math.isclose(lhs, rhs, rel_tol=0, abs_tol=1e-6), (lhs, rhs)
             elif isinstance(lhs, float) and math.isnan(lhs):
                 assert math.isnan(rhs), (lhs, rhs)  # pragma: no cover
             else:
                 assert lhs == rhs, (lhs, rhs)
+
+
+def maybe_get_modin_df(df_pandas: pd.DataFrame) -> Any:
```
FBruzzesi, line 39: Neat! Thanks a lot

tests/test_group_by.py
FBruzzesi: Can't we just keep `data_1` for both tests and change `expected` of `test_group_by_iter` accordingly? Am I missing something?

FBruzzesi (commented):
Thanks a lot for the refactor. I share the same feeling of test structure being a bit messy and often end up writing a test in the wrong place 😂

MarcoGorelli:
thanks for doing this! much better indeed

happy to leave this to you two, feel free to merge when you're happy with it

tests/test_group_by.py
EdAbati: Yes I think it makes sense!
I am now using the previous `data_2` for all. It had repeating values in 2 columns, I think that is why `test_group_by_iter`  was using it. What do you think?

FBruzzesi: Oh I see, thanks for pointing that out! Anyway, keeping the one dataset that works for both tests is all we need! Thanks for the adjustment

FBruzzesi (approved):
That's great @EdAbati, let's merge this and iterate on it if needed! 

I believe that size of this PR was really on point, easy and fast to review :)

NarwhalChubbiwhalGIF (2)

P.s. I know that by Marco standards, cats are the official "ship it" gifs to use, but I would like to propose more narwhals 😂

EdAbati:
"merge gifs" is one of the main reasons I contribute to this project 😄

PR #344, opened 2024-06-28
feat: add nw.Object dtype

### LINKED ISSUE #343: add nw.Object dtype
This shouldn't do anything clever

MarcoGorelli:
closes #343 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -373,9 +373,10 @@ def set_axis(obj: T, index: Any, implementation: str) -> T:
         return obj.set_axis(index, axis=0)  # type: ignore[no-any-return, attr-defined]
 
 
-def translate_dtype(dtype: Any) -> DType:
+def translate_dtype(column: Any) -> DType:
     from narwhals import dtypes
 
+    dtype = column.dtype
     if dtype in ("int64", "Int64", "Int64[pyarrow]"):
         return dtypes.Int64()
     if dtype in ("int32", "Int32", "Int32[pyarrow]"):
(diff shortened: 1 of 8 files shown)


tests/frame/schema_test.py
```diff
@@ -12,3 +17,25 @@
...
     assert {"a": nw.Int32()} == {"a": nw.Int32()}
+
+
+def test_object() -> None:
+    df = pd.DataFrame({"a": [1, 2, 3]}).astype(object)
+    result = nw.from_native(df).schema
+    assert result["a"] == nw.Object
+
+
+def test_string_disguised_as_object() -> None:
```
FBruzzesi, line 28: Love the naming!

narwhals/_pandas_like/utils.py
```diff
@@ -412,7 +413,15 @@ def translate_dtype(dtype: Any) -> DType:
...
-        return dtypes.String()
+        if (idx := column.first_valid_index()) is not None and isinstance(
+            column.loc[idx], str
+        ):
+            # Infer based on first non-missing value.
+            # For pandas pre 3.0, this isn't perfect.
+            # After pandas 3.0, pandas has a dedicated string dtype
+            # which is inferred by default.
+            return dtypes.String()
+        return dtypes.Object()
```
FBruzzesi, line 424: Not sure if the comment is related, but shouldn't all non-null/nan elements be str?

FBruzzesi (commented):
Just noticed it is already merged! I left one comment 🤔

narwhals/_pandas_like/utils.py
MarcoGorelli: you can store anything in object dtype, it doesn't have to be strings

FBruzzesi: That's exactly my point, _all_ elements should be str, not just the first one.

Reproducible example with the above implementation:

```py
class Bar:
    ...

df = pd.DataFrame({"a": ["foo", Bar()]}).astype(object)
result = nw.from_native(df).schema
result["a"] == nw.String
```

```terminal
True
```

MarcoGorelli: ah sure, I see what you mean

true. this is just an approximation, we're just matching what Polars does here https://docs.pola.rs/api/python/stable/reference/series/index.html

PR #347, opened 2024-06-28
feat: cross join

### LINKED ISSUE #301: [Enh]: Support for `how="cross" in DataFrame.join()

_No response_

Needed for TPC-H query Q11. Issue https://github.com/narwhals-dev/narwhals/issues/282

_No response_

_No response_

None

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #301

## Checklist

- [X] Tests added 
- [X] Documented the changes

I am concerned about cudf - I was not able to find a clear doc regarding join's (better called merge)  🤔

### DIFF
--- a/docs/basics/complete_example.md
+++ b/docs/basics/complete_example.md
@@ -65,8 +65,7 @@ import narwhals as nw
 
 class StandardScaler:
     @nw.narwhalify(eager_only=True)
-    def fit(self, df_any):
-        df = nw.from_native(df_any, eager_only=True)
+    def fit(self, df):
         self._means = {col: df[col].mean() for col in df.columns}
         self._std_devs = {col: df[col].std() for col in df.columns}
 
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -284,10 +284,18 @@ class PandasDataFrame:
                 and (pd := get_pandas()) is not None
                 and parse_version(pd.__version__) < parse_version("1.2.0")
             ):
-                import secrets
 
-                n_bytes = max(len(c.encode("utf-8")) for c in self.columns) + 1
-                key_token = secrets.token_hex(n_bytes)
+                def generate_unique_token(
+                    n_bytes: int, columns: list[str]
+                ) -> str:  # pragma: no cover
+                    import secrets
+
+                    while True:
+                        token = secrets.token_hex(n_bytes)
+                        if token not in columns:
+                            return token
+
+                key_token = generate_unique_token(8, self.columns)
 
                 return self._from_dataframe(
                     self._dataframe.assign(**{key_token: 0}).merge(
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -284,20 +284,25 @@ class PandasDataFrame:
                 and (pd := get_pandas()) is not None
                 and parse_version(pd.__version__) < parse_version("1.2.0")
             ):
+                import secrets
+
+                n_bytes = max(len(c.encode("utf-8")) for c in self.columns) + 1
+                key_token = secrets.token_hex(n_bytes)
+
                 return self._from_dataframe(
-                    self._dataframe.assign(__cross_join_key__=0).merge(
-                        other._dataframe.assign(__cross_join_key__=0),
+                    self._dataframe.assign(**{key_token: 0}).merge(
+                        other._dataframe.assign(**{key_token: 0}),
                         how="inner",
-                        left_on="__cross_join_key__",
-                        right_on="__cross_join_key__",
+                        left_on=key_token,
+                        right_on=key_token,
                         suffixes=("", "_right"),
                     ),
-                ).drop("__cross_join_key__")
+                ).drop(key_token)
             else:
                 return self._from_dataframe(
                     self._dataframe.merge(
                         other._dataframe,
-                        how=how,
+                        how="cross",
                         suffixes=("", "_right"),
                     ),
                 )
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -268,7 +268,7 @@ class PandasDataFrame:
         self,
         other: Self,
         *,
-        how: Literal["left", "inner", "outer"] = "inner",
+        how: Literal["left", "inner", "outer", "cross"] = "inner",
         left_on: str | list[str],
         right_on: str | list[str],
     ) -> Self:
@@ -276,6 +276,15 @@ class PandasDataFrame:
             left_on = [left_on]
         if isinstance(right_on, str):
             right_on = [right_on]
+
+        if how == "cross":
+            return self._from_dataframe(
+                self._dataframe.merge(
+                    other._dataframe,
+                    how=how,
+                    suffixes=("", "_right"),
+                ),
+            )
 
         return self._from_dataframe(
             self._dataframe.merge(
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -270,8 +270,8 @@ class PandasDataFrame:
         other: Self,
         *,
         how: Literal["left", "inner", "outer", "cross"] = "inner",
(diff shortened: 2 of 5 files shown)


docs/basics/complete_example.md
```diff
@@ -65,8 +65,7 @@ import narwhals as nw
 
 class StandardScaler:
     @nw.narwhalify(eager_only=True)
-    def fit(self, df_any):
-        df = nw.from_native(df_any, eager_only=True)
+    def fit(self, df):
         self._means = {col: df[col].mean() for col in df.columns}
```
FBruzzesi, line 69: This should have been addresses in #328, my bad!

narwhals/_pandas_like/dataframe.py
```diff
@@ -277,6 +277,15 @@ def join(
...
             right_on = [right_on]
 
+        if how == "cross":
+            return self._from_dataframe(
+                self._dataframe.merge(
+                    other._dataframe,
+                    how=how,
+                    suffixes=("", "_right"),
+                ),
+            )
```
FBruzzesi, line 287: Reason for this is that pandas raises an error if keys are even passed

narwhals/dataframe.py
```diff
@@ -1383,11 +1386,10 @@ def join(
...
             other: DataFrame to join with.
 
-            how: {'inner'}
-                  Join strategy.
+            how: Join strategy.
 
-                  * *inner*: Returns rows that have matching values in both
-                              tables
+                  * *inner*: Returns rows that have matching values in both tables
+                  * *cross*: Returns the Cartesian product of rows from both tables
```
FBruzzesi, line 1392: Renders as follow:

image

tests/frame/test_common.py
```diff
@@ -312,6 +313,30 @@ def test_join(df_raw: Any) -> None:
...
+    rows = [
+        {**left_row, **{k + "_right": v for k, v in right_row.items()}}
+        for left_row in df_.iter_rows(named=True)
+        for right_row in df_.iter_rows(named=True)
+    ]
+    # Pivot to single dict of values: {"a": [1, 1, ...], "z": [7., 8., ...], ...}
+    expected = defaultdict(list)
+    for item in rows:
+        for key, value in item.items():
+            expected[key].append(value)
```
FBruzzesi, line 335: In my first attempt I just hardcoded some values, then I wanted to test for frames with NA values.

There could be a better way, but it did not come to my mind.

First step (`rows = .. `) generates the row values as list of dicts.
Second step, pivots those values in a single dict as expected from `compare_dicts`

FBruzzesi:
Ok, so tests are failing for Modin, and I would expect the same issue with CuDF 😭

MarcoGorelli:
that's fine IMO, we can just xfail them and open an issue their repos

they're the ones claiming to follow the pandas api anyway 😄

FBruzzesi:
> that's fine IMO, we can just xfail them and open an issue their repos
> 
> they're the ones claiming to follow the pandas api anyway 😄

I was already addressing with a workaround needed also for pandas version < 1.2 (min supported is 1.1)

narwhals/_pandas_like/dataframe.py
```diff
@@ -277,6 +278,30 @@ def join(
...
+            ):
+                return self._from_dataframe(
+                    self._dataframe.assign(__cross_join_key__=0).merge(
+                        other._dataframe.assign(__cross_join_key__=0),
+                        how="inner",
+                        left_on="__cross_join_key__",
+                        right_on="__cross_join_key__",
+                        suffixes=("", "_right"),
+                    ),
+                ).drop("__cross_join_key__")
```
FBruzzesi, line 295: @MarcoGorelli this is the workaround. Is it nice? Not really!

If someone comes with a dataframe with a column called `__cross_join_key__` I will get very sad. Maybe we can do even one more step further with `__narwhals_cross_join_key__`?!

MarcoGorelli: How about generating a random token until we find a string which isn't already a column name?

I did something [here](https://github.com/psf/black/pull/2357/files) once, check `get_token` in `src/black/handle_ipynb_magics.py`

narwhals/_pandas_like/dataframe.py
```diff
@@ -277,6 +278,35 @@ def join(
...
 
+        if how == "cross":
+            if self._implementation in {"modin", "cudf"} or (
+                self._implementation == "pandas"
+                and (pd := get_pandas()) is not None
+                and parse_version(pd.__version__) < parse_version("1.2.0")
+            ):
+                import secrets
+
+                n_bytes = max(len(c.encode("utf-8")) for c in self.columns) + 1
```
FBruzzesi, line 289: The +1 should guarantee that it's not already in the columns

FBruzzesi:
Just noticed that in polars, join keys can be empty as they are not used anyway in cross join. 
In narwhals that would raise an error. Should I address it here?

narwhals/_pandas_like/dataframe.py
FBruzzesi: mmmh wondering if we can have issues as pandas column names do not need to be strings

MarcoGorelli:
> Just noticed that in polars, join keys can be empty as they are not used anyway in cross join. In narwhals that would raise an error. Should I address it here?

probably, thanks!

so, Polars doesn't validate that they are empty? I think Polars should probably throw an error rather than silently ignoring join keys in cross-joins then? this might a good issue to open in the polars repo if you fancy it

narwhals/_pandas_like/dataframe.py
```diff
@@ -268,15 +269,52 @@ def join(
...
         if isinstance(left_on, str):
             left_on = [left_on]
         if isinstance(right_on, str):
             right_on = [right_on]
 
+        if how == "cross":
+            if self._implementation in {"modin", "cudf"} or (
+                self._implementation == "pandas"
+                and (pd := get_pandas()) is not None
+                and parse_version(pd.__version__) < parse_version("1.2.0")
```
MarcoGorelli, line 285: is this right? the pandas docs say it's new in 1.2.0, but I guess it's a typo, the 1.3 docs don't have it https://pandas.pydata.org/pandas-docs/version/1.3/reference/api/pandas.DataFrame.join.html, and it only appears in 1.4

narwhals/_pandas_like/dataframe.py
```diff
@@ -268,15 +269,52 @@ def join(
...
+
+                def generate_unique_token(
+                    n_bytes: int, columns: list[str]
+                ) -> str:  # pragma: no cover
+                    import secrets
+
+                    while True:
+                        token = secrets.token_hex(n_bytes)
+                        if token not in columns:
+                            return token
```
MarcoGorelli, line 296: this should be unlikely/impossible, but still, could we raise an assertion error if for some reason the loop doesn't exit after 100 iterations?

tests/frame/test_common.py
```diff
@@ -312,6 +313,41 @@ def test_join(df_raw: Any) -> None:
...
+    df_ = df.collect() if isinstance(df, nw.LazyFrame) else df
+
+    # List of dicts: [{"a": 1, "z": 7., ...}, {"a": 1, "z": 8., ...}, ...]
+    rows = [
+        {**left_row, **{k + "_right": v for k, v in right_row.items()}}
+        for left_row in df_.iter_rows(named=True)
+        for right_row in df_.iter_rows(named=True)
+    ]
+    # Pivot to single dict of values: {"a": [1, 1, ...], "z": [7., 8., ...], ...}
+    expected = defaultdict(list)
```
MarcoGorelli, line 340: could we also construct `expected` explicitly please? else we risk having a bug in `iter_rows` which hides this another bug in cross-join

MarcoGorelli (commented):
awesome effort here!

MarcoGorelli (approved):
I tried to break it, but couldn't

Let's ship it then - amazing work @FBruzzesi !

PR #357, opened 2024-06-30
test: Add pylint checks

ELC:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

@MarcoGorelli

This PR introduces PyLint checks, ideally ruff should be sufficient but there are several checks that are [only available in pylint](https://github.com/astral-sh/ruff/issues/970). Pylint generates a score based on the failing tests, but that score weights the checks in different ways so I configured it like ruff, either everything passes or it fails (fail-under=10).

There are some sections for the design, format and similarity which provide sensitive defaults that can be customized if need be. The checks contain some optional checkers as well and I believe these will greatly help increasing the code quality and readability.

The rationale is that before doing refactors which could be subjective, trusting unopinionated checks is easier. All the failing checks have been disabled so that new PRs are not affected. The idea is fixing and re-enabling checks one by one, leveraging issues for detail discussion.

If you agree with including pylint as part of the pre-commit checks I would:

1. Open a follow-up PR fixing the most trivial checks
2. Create one issue per non-trivial check to have a separate discussion

I'll be looking forward to hear your thoughts

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,39 +1,49 @@
 repos:
-- repo: https://github.com/astral-sh/ruff-pre-commit
-  # Ruff version.
-  rev: 'v0.4.8'
-  hooks:
-    # Run the formatter.
-    - id: ruff-format
-    # Run the linter.
-    - id: ruff
-- repo: https://github.com/pycqa/pylint
-  # Cannot be replaced by ruff until https://github.com/astral-sh/ruff/issues/970 is closed
-  rev: "v3.2.5"
-  hooks:
-    - id: pylint
-- repo: https://github.com/pre-commit/mirrors-mypy
-  rev: 'v1.10.0'
-  hooks:
-    - id: mypy
-      additional_dependencies: ['polars==0.20.10', 'pytest==8.0.1']
-      exclude: utils|tpch
-- repo: https://github.com/codespell-project/codespell
-  rev: 'v2.3.0'
-  hooks:
-    - id: codespell
-      files: \.(py|rst|md)$
-      args: [--ignore-words-list=ser]
-- repo: local
-  hooks:
-    - id: check-api-reference
-      name: check-api-reference
-      pass_filenames: false
-      entry: python -m utils.check_api_reference
-      language: python
-      additional_dependencies: [polars]
-    - id: imports-are-banned
-      name: import are banned (use `get_pandas` instead of `import pandas`)
-      entry: (?<!>>> )import (pandas|polars|modin|cudf)
-      language: pygrep
-      files: ^narwhals/
+  - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -1,35 +1,56 @@
...
+    rev: "v3.2.5"
+    hooks:
+      - id: pylint
+        additional_dependencies:
+          - "polars"
+          - "pandas"
+          - "pytest"
+          - "scikit-learn"
+          - "hypothesis"
+          - "nox"
```
MarcoGorelli, line 23: this is making the environment too big for pre-commit ci

MarcoGorelli (commented):
thanks for your PR!

do any of these checks find any issues in the codebase? if so, could you please enable some that do and fix the associated issues in the PR? else I'm a bit wary of adding extra checks

ELC:
All of the checks disabled with the TODO in the pyproject.toml found issues
in he codebase

Some of them are trivial like "unnecessary else in if/elif/else" but other
are more complex like "redifing built-ins" or "having functions with oo
many branches"

Regarding the size of environment, we can disable checks on imported
modules to make the env smaller, however the only compromise is time,
pre-commit can handle arbitrarily large envs

A similar argument could be made about the mypy hook, to do proper type
checking, it needs all dependencies installed in the pre-commit, otherwise
only locally defined types are checked

ELC:
I saw what you meant, the env is to big for the free service of pre-commit to run it. Github Actions and local setup could handle it just fine though.

I made the env smaller but this is indeed a compromise in the checks done.

This will become an issue soon enough after we introduce new backends, for MyPy we are already installing polars and pandas and if we would like to add pyspark for example, the env would be too big for the `pre-commit.ci` runner to build it. Maybe we should consider moving this step to Github Actions and running `pre-commit run --all-files` there.

MarcoGorelli:
thanks for explaining

> Some of them are trivial like "unnecessary else in if/elif/else" but other
are more complex like "redifing built-ins" or "having functions with oo
many branches"

I think ruff also catches these, and I turned some of them off - tbh I'd rather keep linting simple, lightweight, and fast

ELC:
I saw some of the ruff checks were disabled, are you open to merge PR
fixing those? Even though they may stay disabled

If you are willing, we can consider fixing what Ruff and PyLint suggests
without necessarily making it part of the pre commits

MarcoGorelli:
> I saw some of the ruff checks were disabled, are you open to merge PR
fixing those?

sure, thanks! we could even consider enabling them then

MarcoGorelli:
thanks for your pr but this seems to have gone out of date, closing then

PR #358, opened 2024-06-30
refactor: always recommend `from_native`

EdAbati:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

As implemented in #331, I like the idea of recommending to use `nw.from_native` instead of the classes directly. Moving this to a separate PR as @MarcoGorelli asked. 
The recommendation was already mentioned in some docstrings, but not applied everywhere. This also allows to simplify the `LazyFrame.__init__`.

### DIFF
--- a/tests/hypothesis/test_concat.py
+++ b/tests/hypothesis/test_concat.py
@@ -44,27 +44,29 @@ def test_concat(  # pragma: no cover
 
     if how == "horizontal":
         df_pl = (
-            nw.LazyFrame(df_polars)
+            nw.from_native(df_polars)
+            .lazy()
             .collect()
             .rename({"a": "d", "b": "e"})
             .lazy()
             .drop("c")
         )
         df_pd = (
-            nw.LazyFrame(df_pandas)
+            nw.from_native(df_pandas)
+            .lazy()
             .collect()
             .rename({"a": "d", "b": "e"})
             .lazy()
             .drop("c")
         )
     else:
-        df_pl = nw.LazyFrame(df_polars)
-        df_pd = nw.LazyFrame(df_pandas)
+        df_pl = nw.from_native(df_polars).lazy()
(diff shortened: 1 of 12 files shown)


tests/hypothesis/test_concat.py
```diff
@@ -44,27 +44,29 @@ def test_concat(  # pragma: no cover
 
     if how == "horizontal":
         df_pl = (
-            nw.LazyFrame(df_polars)
+            nw.from_native(df_polars)
+            .lazy()
```
EdAbati, line 48: I kept this like it was #331 for now, but I'm not 100% sure why we are testing here (and other tests below) the `LazyFrame` interface, when both outputs are not lazy.

What would we miss it we just had `from_native()`?

MarcoGorelli: 😄 this looks like a mistake tbh, I think `from_native` should be fine, thanks for noticiing

MarcoGorelli: I _might_ have been chasing coverage with some of these...but this `.lazy().collect()` definitely feels useless

EdAbati: > I might have been chasing coverage with some of these

I'll try to delete some of them and see if coverage is happy 😅

EdAbati:
Deleted some redundant `.lazy()` calls in test here: https://github.com/narwhals-dev/narwhals/pull/358/commits/a4ee7e9f0981db6fffe5dbb05f05476f9a32c76f

Coverage is 100% happy

MarcoGorelli (approved):
thanks @EdAbati !

PR #363, opened 2024-06-30
Use enums for implementations

ELC:

- [X] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [N/A] Tests added 
- [N/A] Documented the changes

## Rationale

By using enums instead of strings we achieve:

1. Single source of truth: now to add a new backend, one only needs to modify the `narwhals/dependencies.py` file
2. Simplify contribution: by having enums in a single location and using `assert_never`, once a new value is added to the Enum, MyPy will flag all the places where modifications are needed, this is much easier to navigate than finding failed tests
3. For multi-backend implementations (such as pandas), a file called `implementations.py` is added showing the subset of implementations supported for that specific platform
4. Reduce boilerplate: with Enums and a couple of mappings, there is no longer a need for many if/elif/else logic, thus making the code easier to extend. There are several cases not yet addressed in this PR but now the only function used from `dependencies` is the `get_backend` rather than the individual `get_pandas`, etc.
5. Less error prone: by using enums, MyPy becomes much more helpful by actually detecting missing cases in the logic when new backends are added.

## Open questions:

This PR assumes `typing_extensions` is installed, this should be a relatively common case as most projects use this library, however, to keep narwhals dependency free, we would need to add `assert_never` to the utils. 

```python
import enum
from typing_extensions import NoReturn

def assert_never(arg: NoReturn) -> NoReturn:
    raise AssertionError("Expected code to be unreachable")
```
[Source](https://typing.readthedocs.io/en/latest/source/unreachable.html#assert-never-and-exhaustiveness-checking)

The `assert_never` and the `Never` type are included in Python 3.11 so this functionality should be explicitly made available to keep the library compatible with 3.8+

An interesting read on this topic is the following article by [Adam Johnson](https://adamj.eu/tech/2022/10/14/python-type-hints-exhuastiveness-checking/)

---

Would like to hear your thoughts on these changes @MarcoGorelli / @EdAbati

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -11,6 +11,7 @@ from typing import overload
 
 from narwhals._expression_parsing import evaluate_into_exprs
 from narwhals._pandas_like.expr import PandasExpr
+from narwhals._pandas_like.utils import Implementation
 from narwhals._pandas_like.utils import create_native_series
 from narwhals._pandas_like.utils import generate_unique_token
 from narwhals._pandas_like.utils import horizontal_concat
@@ -39,7 +40,7 @@ class PandasDataFrame:
         self,
         native_dataframe: Any,
         *,
-        implementation: str,
+        implementation: Implementation,
         backend_version: tuple[int, ...],
     ) -> None:
         self._validate_columns(native_dataframe.columns)
@@ -59,11 +60,11 @@ class PandasDataFrame:
         return PandasNamespace(self._implementation, self._backend_version)
 
     def __native_namespace__(self) -> Any:
-        if self._implementation == "pandas":
+        if self._implementation is Implementation.PANDAS:
             return get_pandas()
-        if self._implementation == "modin":  # pragma: no cover
+        if self._implementation is Implementation.MODIN:  # pragma: no cover
             return get_modin()
-        if self._implementation == "cudf":  # pragma: no cover
+        if self._implementation is Implementation.CUDF:  # pragma: no cover
             return get_cudf()
         msg = f"Expected pandas/modin/cudf, got: {type(self._implementation)}"  # pragma: no cover
         raise AssertionError(msg)
@@ -304,8 +305,12 @@ class PandasDataFrame:
             right_on = [right_on]
 
         if how == "cross":
-            if self._implementation in {"modin", "cudf"} or (
-                self._implementation == "pandas" and self._backend_version < (1, 4)
+            if (
+                self._implementation is Implementation.MODIN
+                or self._implementation is Implementation.CUDF
+            ) or (
+                self._implementation is Implementation.PANDAS
+                and self._backend_version < (1, 4)
             ):
                 key_token = generate_unique_token(
                     n_bytes=8, columns=[*self.columns, *other.columns]
@@ -416,9 +421,9 @@ class PandasDataFrame:
         return self._native_dataframe.to_numpy()
 
     def to_pandas(self) -> Any:
-        if self._implementation == "pandas":
+        if self._implementation is Implementation.PANDAS:
             return self._native_dataframe
-        if self._implementation == "modin":  # pragma: no cover
+        if self._implementation is Implementation.MODIN:  # pragma: no cover
             return self._native_dataframe._to_pandas()
         return self._native_dataframe.to_pandas()  # pragma: no cover
 
--- a/narwhals/_pandas_like/group_by.py
+++ b/narwhals/_pandas_like/group_by.py
@@ -9,6 +9,7 @@ from typing import Callable
 from typing import Iterator
 
 from narwhals._expression_parsing import parse_into_exprs
+from narwhals._pandas_like.utils import Implementation
 from narwhals._pandas_like.utils import is_simple_aggregation
 from narwhals._pandas_like.utils import native_series_from_iterable
 from narwhals.utils import remove_prefix
@@ -43,7 +44,7 @@ class PandasGroupBy:
             namespace=self._df.__narwhals_namespace__(),
             **named_aggs,
         )
-        implementation: str = self._df._implementation
+        implementation: Implementation = self._df._implementation
         output_names: list[str] = copy(self._keys)
         for expr in exprs:
             if expr._output_names is None:
@@ -189,7 +190,7 @@ def agg_pandas(  # noqa: PLR0913
             implementation=implementation,
         )
 
-    if implementation == "pandas" and backend_version >= (2, 2):
+    if implementation is Implementation.PANDAS and backend_version >= (2, 2):
         result_complex = grouped.apply(func, include_groups=False)
     else:  # pragma: no cover
         result_complex = grouped.apply(func)
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import secrets
+from enum import Enum
+from enum import auto
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
@@ -19,6 +21,12 @@ if TYPE_CHECKING:
     from narwhals.dtypes import DType
 
     ExprT = TypeVar("ExprT", bound=PandasExpr)
+
+
+class Implementation(Enum):
+    PANDAS = auto()
+    MODIN = auto()
+    CUDF = auto()
 
 
 def validate_column_comparand(index: Any, other: Any) -> Any:
@@ -86,18 +94,18 @@ def create_native_series(
     iterable: Any,
     index: Any = None,
     *,
-    implementation: str,
+    implementation: Implementation,
     backend_version: tuple[int, ...],
 ) -> PandasSeries:
     from narwhals._pandas_like.series import PandasSeries
 
-    if implementation == "pandas":
+    if implementation is Implementation.PANDAS:
         pd = get_pandas()
         series = pd.Series(iterable, index=index, name="")
-    elif implementation == "modin":
+    elif implementation is Implementation.MODIN:
         mpd = get_modin()
         series = mpd.Series(iterable, index=index, name="")
-    elif implementation == "cudf":
+    elif implementation is Implementation.CUDF:
         cudf = get_cudf()
         series = cudf.Series(iterable, index=index, name="")
     return PandasSeries(
@@ -123,24 +131,24 @@ def is_simple_aggregation(expr: PandasExpr) -> bool:
 
 
 def horizontal_concat(
-    dfs: list[Any], *, implementation: str, backend_version: tuple[int, ...]
+    dfs: list[Any], *, implementation: Implementation, backend_version: tuple[int, ...]
 ) -> Any:
     """
     Concatenate (native) DataFrames horizontally.
 
     Should be in namespace.
     """
-    if implementation == "pandas":
+    if implementation is Implementation.PANDAS:
         pd = get_pandas()
 
         if backend_version < (3,):
             return pd.concat(dfs, axis=1, copy=False)
         return pd.concat(dfs, axis=1)  # pragma: no cover
-    if implementation == "cudf":  # pragma: no cover
+    if implementation is Implementation.CUDF:  # pragma: no cover
         cudf = get_cudf()
 
         return cudf.concat(dfs, axis=1)
-    if implementation == "modin":  # pragma: no cover
+    if implementation is Implementation.MODIN:  # pragma: no cover
         mpd = get_modin()
 
         return mpd.concat(dfs, axis=1)
@@ -149,7 +157,7 @@ def horizontal_concat(
 
 
 def vertical_concat(
-    dfs: list[Any], *, implementation: str, backend_version: tuple[int, ...]
+    dfs: list[Any], *, implementation: Implementation, backend_version: tuple[int, ...]
 ) -> Any:
     """
     Concatenate (native) DataFrames vertically.
@@ -165,17 +173,17 @@ def vertical_concat(
         if cols_current != cols:
             msg = "unable to vstack, column names don't match"
             raise TypeError(msg)
-    if implementation == "pandas":
+    if implementation is Implementation.PANDAS:
         pd = get_pandas()
 
         if backend_version < (3,):
             return pd.concat(dfs, axis=0, copy=False)
         return pd.concat(dfs, axis=0)  # pragma: no cover
-    if implementation == "cudf":  # pragma: no cover
+    if implementation is Implementation.CUDF:  # pragma: no cover
         cudf = get_cudf()
 
         return cudf.concat(dfs, axis=0)
-    if implementation == "modin":  # pragma: no cover
+    if implementation is Implementation.MODIN:  # pragma: no cover
         mpd = get_modin()
 
         return mpd.concat(dfs, axis=0)
@@ -187,18 +195,18 @@ def native_series_from_iterable(
     data: Iterable[Any],
     name: str,
     index: Any,
-    implementation: str,
+    implementation: Implementation,
 ) -> Any:
     """Return native series."""
-    if implementation == "pandas":
+    if implementation is Implementation.PANDAS:
         pd = get_pandas()
 
         return pd.Series(data, name=name, index=index, copy=False)
-    if implementation == "cudf":  # pragma: no cover
+    if implementation is Implementation.CUDF:  # pragma: no cover
         cudf = get_cudf()
 
         return cudf.Series(data, name=name, index=index)
-    if implementation == "modin":  # pragma: no cover
+    if implementation is Implementation.MODIN:  # pragma: no cover
         mpd = get_modin()
 
         return mpd.Series(data, name=name, index=index)
@@ -207,13 +215,22 @@ def native_series_from_iterable(
 
 
 def set_axis(
-    obj: T, index: Any, *, implementation: str, backend_version: tuple[int, ...]
+    obj: T,
+    index: Any,
+    *,
+    implementation: Implementation,
+    backend_version: tuple[int, ...],
 ) -> T:
-    if implementation == "pandas" and backend_version < (1,):  # pragma: no cover
+    if implementation is Implementation.PANDAS and backend_version < (
+        1,
+    ):  # pragma: no cover
         kwargs = {"inplace": False}
     else:
         kwargs = {}
-    if implementation == "pandas" and backend_version >= (1, 5):  # pragma: no cover
+    if implementation is Implementation.PANDAS and backend_version >= (
+        1,
+        5,
+    ):  # pragma: no cover
         kwargs["copy"] = False
     else:  # pragma: no cover
         pass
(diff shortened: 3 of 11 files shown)


MarcoGorelli (commented):
thanks for your PR! on board with the idea

narwhals/_pandas_like/dataframe.py
```diff
@@ -8,16 +8,17 @@
 from typing import Literal
 from typing import overload
 
+from typing_extensions import assert_never
```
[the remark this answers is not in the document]
ELC: Done

narwhals/_arrow/namespace.py
```diff
@@ -53,7 +53,9 @@ def _create_expr_from_callable(  # noqa: PLR0913
             output_names=output_names,
         )
 
-    def _create_expr_from_series(self, series: ArrowSeries) -> ArrowExpr:
+    def _create_expr_from_series(
+        self, series: ArrowSeries
+    ) -> ArrowExpr:  # pragma: no cover
```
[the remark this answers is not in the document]
ELC: This was unexpected, I rebased to the latest main and this function was not 100% covered. None of this PR's change should have affected `_arrow`

I will investigate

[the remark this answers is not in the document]
francocastano: Which test is supposed to cover this function? @MarcoGorelli

ELC:
@MarcoGorelli - This should be now ready to merge, however, there is a file without 100% coverage (not related to this PR):

image

narwhals/_pandas_like/group_by.py
```diff
@@ -160,15 +160,16 @@ def func(df: Any) -> Any:
...
-    if implementation == "pandas":
-        pd = get_pandas()
+    extra_args: dict[str, Any] = {}
+    if implementation is Implementation.PANDAS:  # pragma: no cover
+        backend = get_backend(implementation)
+        backend_version = parse_version(backend.__version__)
+        minimum_version = parse_version("2.2.0")
+
+        if backend_version >= minimum_version:
+            extra_args["include_groups"] = False
```
EdAbati, line 170: (nitpicking, sorry) suggesting variable names and usage that I'd find more immediate/readable

```suggestion
    apply_kwargs: dict[str, Any] = {}
    if implementation is Implementation.PANDAS:  # pragma: no cover
        backend = get_backend(implementation)

        if parse_version(backend.__version__) >= parse_version("2.2.0"):
            apply_kwargs["include_groups"] = False
```

narwhals/_pandas_like/utils.py
```diff
@@ -325,57 +321,46 @@ def vertical_concat(dfs: list[Any], implementation: str) -> Any:
...
 
-def set_axis(obj: T, index: Any, implementation: str) -> T:
-    if implementation == "pandas" and parse_version(
-        get_pandas().__version__
-    ) >= parse_version("1.5.0"):
+def set_axis(obj: T, index: Any, implementation: PANDAS_IMPLEMENTATIONS) -> T:
+    backend = get_backend(implementation)
+    backend_verison = parse_version(backend.__version__)
+    minimum_version = parse_version("1.5.0")
+    if implementation is Implementation.PANDAS and backend_verison >= minimum_version:
```
EdAbati, line 360: (similar nitpicking)

```suggestion
    if implementation is Implementation.PANDAS and (
        backend_verison >= parse_version("1.5.0")
    ):
```

EdAbati (commented):
Thanks for the PR! I really like the use of Enum here.

To solve the coverage issue , can you try to update `ArrowSeries` and `ArrowDataFrame` to use the `Implementation` enum? They still use the string

I also left 2 very minor comments

ELC:
All the tests are passing now, however seems like there is some coverage issues in my local setup using ubuntu latest on a Github Codespace.

This is the main branch:
image

image

This is related to the recently merged cross join PR (#347)

Could we merge this anyway or should we wait until the coverage is 100% in main?

MarcoGorelli:
> Could we merge this anyway or should we wait until the coverage is 100% in main?

sorry about that - it's 100% in CI, but some tests don't run locally. I've addressed that in https://github.com/narwhals-dev/narwhals/pull/379

.pre-commit-config.yaml
```diff
@@ -12,7 +12,10 @@ repos:
   rev: 'v1.10.0'
   hooks:
     - id: mypy
-      additional_dependencies: ['polars==0.20.10', 'pytest==8.0.1']
+      additional_dependencies:
+        - "polars==0.20.10"
+        - "pytest==8.0.1"
+        - "typing-extensions==4.12.2"
```
MarcoGorelli, line 18: isn't this already a dependency of mypy?

https://github.com/python/mypy/blob/d1d3c780c7a2d30b2a038903289ea7487303a218/pyproject.toml#L1-L15

narwhals/_pandas_like/dataframe.py
```diff
@@ -290,10 +284,15 @@ def join(
...
 
         if how == "cross":
-            if self._implementation in {"modin", "cudf"} or (
-                self._implementation == "pandas"
-                and (pd := get_pandas()) is not None
-                and parse_version(pd.__version__) < parse_version("1.4.0")
+            backend = get_backend(self._implementation)
+            if (
+                self._implementation is Implementation.MODIN
+                or self._implementation is Implementation.CUDF
```
MarcoGorelli, line 290: does `self._implementation in {Implementation.MODIN, Implementation.CUDF}` work?

narwhals/_pandas_like/implementations.py
```diff
@@ -0,0 +1,22 @@
...
+from narwhals.dependencies import Implementation
+from narwhals.dependencies import get_backend
+from narwhals.typing import assert_never
+
+PANDAS_IMPLEMENTATIONS = Literal[
+    Implementation.PANDAS, Implementation.MODIN, Implementation.CUDF
+]
+
+
+def get_series_implementation(implementation: PANDAS_IMPLEMENTATIONS) -> Any:
```
MarcoGorelli, line 14: not sure about this function name tbh

narwhals/_pandas_like/utils.py
```diff
@@ -325,57 +321,47 @@ def vertical_concat(dfs: list[Any], implementation: str) -> Any:
...
+    return series_implementation(data, name=name, index=index)
 
 
-def set_axis(obj: T, index: Any, implementation: str) -> T:
-    if implementation == "pandas" and parse_version(
-        get_pandas().__version__
-    ) >= parse_version("1.5.0"):
+def set_axis(obj: T, index: Any, implementation: PANDAS_IMPLEMENTATIONS) -> T:
+    backend = get_backend(implementation)
+    backend_verison = parse_version(backend.__version__)
```
MarcoGorelli, line 358: we don't need to parse the version for non-pandas implementations here?

narwhals/dependencies.py
```diff
@@ -15,6 +19,15 @@
     import pandas
 
 
+class Implementation(Enum):
+    POLARS = auto()
+    PANDAS = auto()
+    MODIN = auto()
+    CUDF = auto()
+    PYARROW = auto()
+    NUMPY = auto()
```
MarcoGorelli, line 28: what's this?

narwhals/typing.py
```diff
@@ -32,4 +33,9 @@ def join(self, *args: Any, **kwargs: Any) -> Any: ...
 # Anything which can be converted to a Narwhals DataFrame.
 IntoDataFrame: TypeAlias = Union["NativeDataFrame", "DataFrame"]
 
-__all__ = ["IntoExpr", "IntoDataFrame"]
+
+def assert_never(_: NoReturn) -> NoReturn:
+    raise AssertionError("Expected code to be unreachable")
+
+
+__all__ = ["IntoExpr", "IntoDataFrame", "assert_never"]
```
MarcoGorelli, line 41: not sure `assert_never` should be a public type, it's only used internally here

ELC: If we put a leading underscore, it will make I private *to the typing module*, it is internal, but there are no naming conventions for internal members. That is, it is used in other modules of the package however not intended to be used outside the package

narwhals/dependencies.py
ELC: This is the enum where we register the different implementations, whenever we add a new one, we need to add it here and MyPy will flag all the necessary changes

narwhals/_pandas_like/utils.py
ELC: It was written this way to avoid using `get_pandas`, the idea with the enum is that now we only expose and use `get_backend`, making he function agnostic on how that backend is loaded

narwhals/_pandas_like/implementations.py
ELC: Me neither, feel free to suggest any name

narwhals/_pandas_like/dataframe.py
ELC: Unfortunately MyPy only supports type narrowing in Enums with the `is` operator. Runtime behaviour is identical but MyPy will not narrow the type unless done this way

I believe this is because `in` uses `==` and not `is` internally but I am not totally sure

.pre-commit-config.yaml
ELC: It is but, since we are explicitly using it in some imports, I believe it is better to have pinned it.

Just because *explicit is better than implicit*

ELC:
I will resolve the conflicts later today. They all are about the same theme, the files need to use the `get_backend` facade instead of the concrete `get_*`

narwhals/dependencies.py
MarcoGorelli: Sorry I mean, why numpy?

narwhals/_pandas_like/utils.py
MarcoGorelli: sure but you don't need to call `parse_version` for non-pandas implementations here

narwhals/dependencies.py
ELC: Because there are a couple of places where we call  `get_backend` with `Implementation.Numpy`. The idea behind the enums is to always access the compute backend through `get_backend`.

This is not fully done yet, but after this PR is merge, I can work on replacing all the intances of `get_*` with the `get_backend`. At the moment there is only two places, which are the entrypoints.

The rationale behind this is to make the code fully compute backend agnostic

narwhals/_pandas_like/utils.py
ELC: It was just to make it easier on the eyes, we can move the `parse_version` into the condition expression

ELC:
Finished with all modifications, tests are passing with 100% coverage and I simplify some of the functions as well to leverage the enums.

Let me know your thoughts

@MarcoGorelli / @EdAbati

narwhals/dependencies.py
MarcoGorelli: sure, but we don't have a numpy backend?

I mean, pandas uses it under-the-hood, but as far as narwhals is concerned, the dataframe backend is pandas, not numpy

MarcoGorelli (commented):
thanks for your PR

I don't think we can get rid of the `get_*` functions, they're part of the public API, and are already used

https://github.com/FBruzzesi/timebasedcv/blob/ce0205173c883574372168dfd4b63e46ccfb621a/timebasedcv/splitstate.py#L17

internal cleanups are fine, but the public api should remain stable (we can make exceptions in some cases, but not if some part of the API is already being used)

ELC:
This then becomes a design question, I can revert back to continue exposing the `get_*` functions but internally we should only use the `get_backend`

The value of narwhals is to provide an abstraction over the dataframe engine, if one can access individual engines, it is leaking the abstraction. 

For the internal package use case, adding a new engine should not require changing if/elif/else in many places, there should be abstractions in place so that only one place (like `dependencies.py`) needs to be modified, in addition to the concrete folder of that implementation. 

If a consumer code needs a particular engine, they should import it directly, not through narwhals, specially because narwhals cannot guarantee the engine will be available.

I don't know exactly what is the vision but I assumed that someone using narwhals will like to remove most of the engine specifics and just use the engine for I/O and narwhals for expressions, hence using something like `get_pandas` from consumer code seemed unexpected to me

narwhals/dependencies.py
ELC: That is true, I just wanted to abstract the way we interact with dependencies from within the package to avoid calling `get_numpy` directly and use the `get_backend` function. This may be a little misleading as Numpy is not a backend pre se. It was just for convenience.

However to further limit the scope of what is and what is not a dataframe engine, I created two types below `DATAFRAME_IMPLEMENTATIONS` and `SERIES_IMPLEMENTATIONS` which do not include Numpy. They are both thought to be subset of this `Implementation` enum. That means though that maybe `Implementation` is not the best name for this.

I wanted the enum to capture everything that was external, be it an auxiliary library like numpy or pyarrow_compute or a dataframe engine like pandas

ELC:
@MarcoGorelli is this ready to merge or are there other changes you would like to include?

narwhals/dependencies.py
MarcoGorelli: I don't think numpy and pyarrow.compute really belong here

I'd suggest just having `Implementation` here, and then `PandasImplementation`

ELC: And to make sure that both `pyarrow.compute` and `numpy` are loadable by the `get_backend` function, how should we name that Enum? It should be different than `Implementation`

MarcoGorelli: we can just use `get_numpy` / `get_pyarrow_compute`

noxfile.py
```diff
@@ -33,7 +33,7 @@ def pytest_coverage(session: Session) -> None:
 @nox.session(python=PYTHON_VERSIONS[0])  # type: ignore[misc]
 def minimum_versions(session: Session) -> None:
     session.install(
-        "pandas==0.25.3",
+        "pandas==1.1.5",
```
ELC, line 36: @MarcoGorelli I made this change because tests are not passing. Pyarrow does not support `pandas< 1.0.0`.

Tests were broken by merging #404

narwhals/dependencies.py
ELC: Done

noxfile.py
MarcoGorelli: could you clarify please? looks like that PR (and the associated merge to main) pass CI https://github.com/narwhals-dev/narwhals/commits/main/ ?

ELC: These are some of the errors I see when running test locally on latest version of main:

```text
UNEXPECTED EXCEPTION: AttributeError("'DataFrame' object has no attribute 'convert_dtypes'")
  File "/usr/lib/python3.8/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest narwhals.utils.maybe_convert_dtypes[6]>", line 1, in <module>
  File "/workspaces/narwhals/narwhals/utils.py", line 261, in maybe_convert_dtypes
    df_any._dataframe._dataframe.convert_dtypes(*args, **kwargs)
  File "/workspaces/narwhals/.nox/minimum_versions/lib/python3.8/site-packages/pandas/core/generic.py", line 5179, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'DataFrame' object has no attribute 'convert_dtypes'
/workspaces/narwhals/narwhals/utils.py:249: UnexpectedException
```


```text
UNEXPECTED EXCEPTION: ImportError('pyarrow requires pandas 1.0.0 or above, pandas 0.25.3 is installed')
  File "/usr/lib/python3.8/doctest.py", line 1336, in __run
    exec(compile(example.source, filename, "single",
  File "<doctest narwhals.series.Series.to_pandas[8]>", line 1, in <module>
  File "/workspaces/narwhals/narwhals/translate.py", line 400, in wrapper
    result = func(*args, **kwargs)
  File "<doctest narwhals.series.Series.to_pandas[6]>", line 3, in func
  File "/workspaces/narwhals/narwhals/series.py", line 1209, in to_pandas
    return self._series.to_pandas()
  File "/workspaces/narwhals/.nox/minimum_versions/lib/python3.8/site-packages/polars/series/series.py", line 4434, in to_pandas
    else self.to_arrow().to_pandas(**kwargs)
  File "pyarrow/array.pxi", line 830, in pyarrow.lib._PandasConvertible.to_pandas
  File "pyarrow/array.pxi", line 1426, in pyarrow.lib.Array._to_pandas
  File "pyarrow/array.pxi", line 1649, in pyarrow.lib._array_like_to_pandas
  File "pyarrow/pandas-shim.pxi", line 103, in pyarrow.lib._PandasAPIShim.series
  File "pyarrow/pandas-shim.pxi", line 96, in pyarrow.lib._PandasAPIShim._check_import
  File "pyarrow/pandas-shim.pxi", line 65, in pyarrow.lib._PandasAPIShim._import_pandas
ImportError: pyarrow requires pandas 1.0.0 or above, pandas 0.25.3 is installed
/workspaces/narwhals/narwhals/series.py:1203: UnexpectedException
```

```text
______________________________________________ [doctest] narwhals.expression.ExprDateTimeNamespace.total_seconds ______________________________________________
2635 
2636             We define a dataframe-agnostic function:
2637 
2638             >>> @nw.narwhalify
2639             ... def func(df):
2640             ...     return df.with_columns(a_total_seconds=nw.col("a").dt.total_seconds())
2641 
2642             We can then pass either pandas or Polars to `func`:
2643 
2644             >>> func(df_pd)
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -                       a  a_total_seconds
    -0        0 days 00:00:10               10
    -1 0 days 00:00:20.040000               20
    +                a  a_total_seconds
    +0        00:00:10               10
    +1 00:00:20.040000               20

/workspaces/narwhals/narwhals/expression.py:2644: DocTestFailure
```

They all appear when running doctest with minimum versions. However, CICD passes just fine, so I am concerned there is a misconfiguration somewhere, probably missing a couple of `skipif`s. The error clearly states that `Pyarrow` requires `pandas >=1.0.0`

ELC: @MarcoGorelli I believe I found the causes:

1. The `minimum_versions` is running `pandas==0.25.3` but wihtout the `--doctest-modules` - [Source](https://github.com/narwhals-dev/narwhals/blob/cc2c1f11983f48122d102d22382468192c13737e/.github/workflows/extremes.yml#L9-L34)
2. The `pretty_old_versions` is running `--doctest-modules` but without `pandas==0.25.3`, it is using `pandas==1.1.5` - [Source](https://github.com/narwhals-dev/narwhals/blob/cc2c1f11983f48122d102d22382468192c13737e/.github/workflows/extremes.yml#L36-L63)

Was this made purposely? If so, the noxfile and the tests do not accurately reflect what the CICD does

MarcoGorelli: ah thanks - I don't use nox myself, another dev added it, it should probably be updated

MarcoGorelli:
thanks for your pr 🙏  there's some changes I agree with, others not so much - I'll add a commit later before merging

ELC:
Great this got merged!

I don't see the `get_backend` functionality in the final set of changes, are you suggesting I made a separete PR for those or was that accidentally lost in the last force push?

PR #366, opened 2024-07-01
Re-enable Ruff Checks 

ELC:

- [X] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Pre-commit execution time before this PR on whole codebase:
image

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -378,75 +378,93 @@ def set_axis(obj: T, index: Any, implementation: str) -> T:
         return obj.set_axis(index, axis=0)  # type: ignore[no-any-return, attr-defined]
 
 
-def translate_dtype(column: Any) -> DType:
+def __translate_primitive_dtype(dtype: str) -> DType | None:
     from narwhals import dtypes
 
-    dtype = column.dtype
-    if str(dtype) in ("int64", "Int64", "Int64[pyarrow]", "int64[pyarrow]"):
-        return dtypes.Int64()
-    if str(dtype) in ("int32", "Int32", "Int32[pyarrow]", "int32[pyarrow]"):
-        return dtypes.Int32()
-    if str(dtype) in ("int16", "Int16", "Int16[pyarrow]", "int16[pyarrow]"):
-        return dtypes.Int16()
-    if str(dtype) in ("int8", "Int8", "Int8[pyarrow]", "int8[pyarrow]"):
-        return dtypes.Int8()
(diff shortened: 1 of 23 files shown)


narwhals/_pandas_like/utils.py
```diff
@@ -378,75 +378,93 @@ def set_axis(obj: T, index: Any, implementation: str) -> T:
...
+        for pandas_dtype, dtype_factory in dtype_mappers.items()
+        if dtype == pandas_dtype
+    )
+    return dtype_factory()
+
+
+def __translate_datetime_dtype(dtype: str) -> DType | None:
+    from narwhals import dtypes
+
+    if dtype in ("category",) or dtype.startswith("dictionary<"):
```
MarcoGorelli, line 420: I have a couple of questions

- why is category here in `datetime_dtype`?
- why 2 leading underscores? isn't one enough?

narwhals/_pandas_like/utils.py
```diff
@@ -468,116 +486,104 @@ def get_dtype_backend(dtype: Any, implementation: str) -> str:
...
-        msg = "Date dtype only supported for pyarrow-backed data types in pandas"
-        raise NotImplementedError(msg)
+
+    for dtype_type, dtype_to_backend_mapping in dtypes_mapping.items():
+        if not isinstance_or_issubclass(dtype, dtype_type):
+            continue
+
+        if dtype_backend not in dtype_to_backend_mapping:
+            msg = f"Type {dtype_type} not supported for backend {dtype_backend}"
+            raise NotImplementedError(msg)
```
MarcoGorelli, line 583: not sure we should raise here - we just want to preserve the dtype backend if possible

ELC: this will return the dtype backend, this raise is for the case of Date that only exists in PyArrow and not Pandas or Numpy

narwhals/_pandas_like/utils.py
ELC: 1. That is a mistake, it should not be there
2. To enforce it is not going to be used outside: https://realpython.com/python-double-underscore/#double-leading-underscore-in-classes-pythons-name-mangling

ELC:
Following authors suggestions I will divide this in atomic PRs, one per check

PR #346, opened 2024-06-28
Working draft of the plotting tutorial

mistShard:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #201 
- Closes #

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added 
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below.
I added two files to aid in creating plotting tutorials. The tutorial shows a single graph now.

Issue #201, opened 2024-05-23
Add plotting tutorial


MarcoGorelli:
Currently, we have one complete example in which we create a standard scaler: https://narwhals-dev.github.io/narwhals/basics/complete_example/

It would be nice to add a second example, in which we make a plot of some kind. I don't know what exactly, feel free to let your imagination run wild. But the resulting plot must be pretty and colourful

If anyone wants to add a second complete example, that'd be great

then, we could:
- rename "complete example" to "Complete Example I: scikit-learn-like transformer"
- add "Complete Example II: plotting function"

TylerEnglish:
I’ll work on this

ChristianLG2:
@MarcoGorelli Should we only add the code along a png example it in the documentation or should we add it somewhere else? 
(Working with Benji on this)

MarcoGorelli:
you might be able to get the code to generate the plot, not sure. otherwise exporting a png sounds fine, thanks!

MarcoGorelli:
You'll probably want to add a file `docs/basics/plotting.md`

then, in `mkdocs.yml`, add that file to "`Tutorial`", so it shows:
```
    - basics/complete_example.md
    - basics/plotting.md
```

MarcoGorelli:
you might also want to install https://pypi.org/project/mkdocs-matplotlib/

mistShard:
### Bar plot

As I worked on the bar plot (using matplotlib), I wrote the following code attempting to use the ouput of `nw.from_native()` in the same way that a Pandas or Polars dataframe would be used:

```python
import narwhals as nw
import matplotlib.pyplot as plt
import pandas as pd
import polars as pl

data2 = {'Player': ['David De Gea', 'Paul Pogba', 'Edinson Cavani', 'Anthony Martial', 'Marcus Rashford', 'Cristiano Ronaldo'], 'Annual Salary': [19500000, 15080000, 13000000, 13000000, 10400000, 26800000]}

pd_df = pl.DataFrame(data2)

def plot_bar(df_any):
    df = nw.from_native(df_any)

    # Create the bar plot
    plt.figure(figsize=(10, 6))
    colors = ['#1f77b4', '#ff7f0e', '#2ca02c', '#d62728', '#9467bd', '#8c564b']
    plt.bar(df['Player'], df['Annual Salary'].to_numpy(), color=colors, width=0.8)
```
[...]

PR #402, opened 2024-07-04
Made functions agnostic. Added Bar plot

mistShard:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #201 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

I implemented the feedback I received on my previous draft pull request #346 . I also added a Bar plot tutorial.


mistShard:
@MarcoGorelli I'm working on the assumption that the more in-depth code explanations you suggested I add should focus on Narwhals and not necessarily the plotting logic. Is this accurate?

ChristianLG2 (commented):
Hey Benji great job you have here, lets just work on some improvements:

1. Lets modify the bar_plot to include the currency formatting and modify y axis to be representative of your population (Just like we discussed).
2. Lets add the pie chart
3. Lets describe the Narwhals functions we are using.

Let me know if you have any questions!!! Keep it up!

MarcoGorelli (commented):
Thanks for doing this, solid effort!

A few comments:
- right now, it reads more like an example than a tutorial. Perhaps we can add a bit more explanation about what you're doing and how it works? In particular, something that I think needs mentioning is that matplotlib calls `np.array` on its inputs, which defers to `__array__`, which `narwhals.Series` supports, and that's why we're able to pass Narwhals Series directly to matplotlib
- are football players' salaries the best topic? It just feels a bit...common. Maybe we could do something related to animals?
- I think a single plot is enough to demonstrate the topic, we shouldn't need more than that
- If possible, it would be good to create the data in the tutorial itself, rather than storing and reading a csv file. For example:
```python
import polars as pl
import pandas as pd

data = {...}  # put your data here
pl_df = pl.DataFrame(data)
pd_df = pd.DataFrame(data)
```

It doesn't have to be a big dataset, we just need to show that it's possible to visualise it and pass `narwhals.Series` to matplotlib. `plot_population_line` is pretty much as complex as it needs to get, I don't think we need the complexity of showing group-by / sort / filter

Issue #372, opened 2024-07-01
feat: Add missing methods to PyArrow backend


MarcoGorelli:
Narwhals methods are generally available for pandas-like APIs and Polars

We should make sure that PyArrow table has the same amount of coverage. Nearly anything can be converted to PyArrow, so this would really be worth it.

The task here is:
1. Choose a method from the `MISSING` list in [`https://github.com/narwhals-dev/narwhals/blob/main/utils/check_backend_completeness.py#L14`](https://github.com/narwhals-dev/narwhals/blob/main/utils/check_backend_completeness.py#L14)
2. Implement that method in `narwhals/_arrow/dataframe.py` or `narwhals/_arrow/series.py`
3. Make sure that there are tests to cover the new function you've added
4. Once you've got 100% coverage and all tests pass, open a PR! 🙌 

It might be a good idea to leave a comment indicating which method(s) you're working on, so we don't duplicate work

brentomagic:
i'll start with series.min.

MarcoGorelli:
cool @brentomagic ! it might be good to first address part of https://github.com/narwhals-dev/narwhals/issues/401 so that there's a test for `Series.min` which uses `constructor_series`, so you can then swap out `constructor_series_with_pyarrow` in its place

or to start with a method where the tests already use `constructor_series`, such as `Series.to_frame`

lucianosrp:
Just adding here that https://github.com/narwhals-dev/narwhals/pull/455 adds the StringNamespace in the PyArrow implementation (`narwhals/_arrow`)

anopsy:
Hi there! If it's not already taken I'll work on   "Series.is_in", "Series.len"

MarcoGorelli:
yeeesss @anopsy ! 🙌 !

MarcoGorelli:
> Edit: Now I see, len is on the MISSING list, but there's already _ len _ in _arrow.series.py 🤔

yeah we need both, as `Expr.len()` is a (lazy) method, but Python requires us to return an integer from `__len__`

FBruzzesi:
Not in the list but possible to support we have some dunder methods as well:

- [ ] `__rand__`
- [ ] `__ror__`
- [ ] `__radd__`
- [ ] `__rsub__`
- [ ] `__rmul__`
- [ ] `__rpow__`
- [ ] `__rmod__`
- [ ] `__truediv__`
- [ ] `__rtruediv__`
- [ ] `__floordiv__`
[...]

PR #404, opened 2024-07-04
feat: some pyarrow dataframe methods

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Introduces the following methods for pyarrow dataframe:

- collect
- drop
- drop_nulls
- lazy
- sort
- to_pandas


FBruzzesi:
@MarcoGorelli CI fails due to `pandas<v1.0.0`.
In general I believe cases like this one can happen. One way of dealing with interoperability constraint could be to decorate pyarrow `to_pandas` method with something similar to [scikit-learn available_if](https://scikit-learn.org/stable/modules/generated/sklearn.utils.metaestimators.available_if.html) and support it only for certain versions installed?! WDYT?

MarcoGorelli:
this makes me so happy!

MarcoGorelli:
> One way of dealing with interoperability constraint could be to decorate pyarrow to_pandas method with something similar to [scikit-learn available_if](https://scikit-learn.org/stable/modules/generated/sklearn.utils.metaestimators.available_if.html) and support it only for certain versions installed?! WDYT?

Nice, I didn't know about that. sounds good! or just xfail some tests for now

narwhals/_arrow/dataframe.py
```diff
@@ -112,3 +117,37 @@ def select(
         pa = get_pyarrow()
         df = pa.Table.from_arrays([s._series for s in new_series], names=names)
         return self._from_dataframe(df)
+
+    def drop(self, *columns: str | Iterable[str]) -> Self:
+        return self._from_dataframe(self._dataframe.drop(list(flatten(columns))))
```
FBruzzesi, line 122: pyarrow `Table.drop` has been renamed to `drop_columns` from version 12.0.0, however for now `drop`  is kept for backwards compatibility.

MarcoGorelli (approved):
thanks so much, being able to run pyarrow with polars syntax would be 🔥 I think/hope this'll attract a few people to the project

Issue #355, opened 2024-06-29
Research: how would Narwhals work in scikit-learn?


MarcoGorelli:
Scikit-learn mentioned Narwhals here https://github.com/scikit-learn/scikit-learn/pull/28513#issuecomment-2131226993

Regardless of whether they decide to use it or not, it would be beneficial to check whether Narwhals would at least work in scikit-learn, and whether it could in principle solve the linked issue

@EdAbati - you've contributed to scikit-learn, and you know Narwhals well - perhaps this issue might be of interest to you?

EdAbati:
Uuuh this can be fun! I contributed to features related to the Array API compatibility lately. But I'll gladly also try to figure out this part of scikit-learn

>  it would be beneficial to check whether Narwhals would at least work in scikit-learn, and whether it could in principle solve the linked issue

Agreed, let's start having a look at these:

- [ ] `sklearn.inspection.permutation_importance`
   @EdAbati working on it
- [ ] `sklearn.model_selection. cross_val_score`
- [ ] `sklearn.model_selection. cross_validate`


If anyone wants to help, please feel free to comment.

EdAbati:
Some early thoughts:

- what could be the cleanest way to select by row index? the pandas equivalent would be `X.iloc[shuffling_idx, col_idx]`, I guess polars also works with `X[shuffling_idx, col_idx]`, right? what do you think, is it a missing feature?
- what about `.copy()`/`.clone()`? FYI I am looking at [this](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/inspection/_permutation_importance.py#L42-L60)

MarcoGorelli:
> what do you think, is it a missing feature?

I'd say  so, yes! I think we need this one in Altair too. Does implementing it interest you? I think it just requires an extra branch `DataFrame__getitem__`

> what about .copy()/.clone()?

Sure, doesn't hurt to add `DataFrame.clone` 👍

EdAbati:
> Does implementing it interest you?

Most of the time my answer to the question is "yes". My only problem is time 😅 I will create a issue in case someone else is able to pick it up before me

MarcoGorelli:
😄 I'll take the getitem one on then so we can propose it to Altair too

EdAbati:
Another useful feature could be something like polars' `DataFrame.replace_column`.
Or do we have already the capability to overwrite a column without changing its position?

MarcoGorelli:
[...]

Issue #401, opened 2024-07-04
clean: factor out tests from `test_common.py`


MarcoGorelli:
A lot of tests are in this file: https://github.com/narwhals-dev/narwhals/blob/main/tests/frame/test_common.py

This has a couple of issues:
- it's harder to find tests for some specific functions
- the tests rely on global `df_pandas` / `df_polars` / etc. variables

I think we should factor these out into dedicated functions, in which the input is a dictionary, and we use the `constructor` fixture (which parametrises over the different libraries)

Then, if we want to test extra backends, we should just be able to add the new library to the `constructor` fixture

See https://github.com/narwhals-dev/narwhals/pull/400/files for an example of how to do this

Another example: https://github.com/narwhals-dev/narwhals/pull/416 (note that `constructor_with_pyarrow` might not necessarily work yet - try it, and if it doesn't, just use `constructor`)

FBruzzesi:
It should not be hard, but there are still a few tests in [`series_only/test_common.py`](https://github.com/narwhals-dev/narwhals/blob/main/tests/series_only/test_common.py)

PR #406, opened 2024-07-04
feat: add `clone` for dataframes

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

 Mentioned in #355 

## Checklist

- [x] Tests added 
- [x] Documented the changes

I am not very familiar with pyarrow Dataframes. What could be a clean way to implement copy/clone?
I added `NotImplementedError` as a placeholder, if we keep the error I will add a test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -52,6 +52,10 @@ def polars_constructor(obj: Any) -> IntoDataFrame:
     return pl.DataFrame(obj)
 
 
+def polars_lazy_constructor(obj: Any) -> pl.LazyFrame:
+    return pl.LazyFrame(obj)
+
+
 if parse_version(pd.__version__) >= parse_version("2.0.0"):
     params = [pandas_constructor, pandas_nullable_constructor, pandas_pyarrow_constructor]
 else:  # pragma: no cover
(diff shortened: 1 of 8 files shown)


MarcoGorelli (commented):
Nice, thanks @EdAbati !

I just have two comments:
- should we add LazyFrame.clone too?
- seeing as this is a new test, shall we follow the practices from https://github.com/narwhals-dev/narwhals/issues/401, and use a new `tests/frame/clone_test.py` file, and use the `constructor` fixture? i.e. the one without pyarrow

tests/conftest.py
```diff
@@ -52,6 +52,10 @@ def polars_constructor(obj: Any) -> IntoDataFrame:
     return pl.DataFrame(obj)
 
 
+def polars_lazy_constructor(obj: Any) -> pl.LazyFrame:
```
EdAbati, line 55: I think I couldn't use `IntoDataFrame` here. What would be the correct way to type this?

Should we add a `IntoLazyFrame` type too?

EdAbati:
(BTW the test are looking very nice now with the new `pytest.fixture`s)

tests/conftest.py
MarcoGorelli: yup, probably! and maybe we will also need `IntoFrame` for when you don't know?

on the typing front, I'd like to see if we can make `DataFrame` generic

MarcoGorelli (approved):
graaaandee @EdAbati 

if there's anything you need for sklearn lmk, thanks for doing this!

Issue #282, opened 2024-06-11
bench: implement TPC-H queries using Narwhals


MarcoGorelli:
In https://github.com/narwhals-dev/narwhals/tree/main/tpch/notebooks, we implement several TPC-H queries, and run them via Narwhals

We currently have the first 7 there, and should work on adding more

To track progress:
- [ ] Q8: blocked by #47 . Skip this one for now
- [x] Q9
- [x] Q10
- [x] Q11
- [ ] Q12
- [ ] Q13
- [ ] Q14
- [x] Q15
- [ ] Q16
- [x] Q17
- [x] Q18
- [x] Q19
- [x] Q20
- [x] Q21
- [ ] Q22

You can find reference implementations for these queries here: https://github.com/pola-rs/tpch/tree/main/queries/polars

The task here is:
1. Make an account on kaggle.com
2. choose one of the above queries (e.g. Q9 - don't 
3. Make a fork of https://www.kaggle.com/code/marcogorelli/narwhals-tpch-q5-s2 (you should see a "copy and edit" button on Kaggle)
4. Remove `q5_pandas_native`. We don't currently have reference implementations for them https://github.com/pola-rs/tpch/tree/main/queries/pandas, so we skip them
5. replace `def q5` with something like the query you find in https://github.com/pola-rs/tpch/tree/main/queries/polars . You will need to modify it a bit, check the current notebooks and try to follow how they do it. You'll also need to get it to work for Narwhals, so that means:
  - using `nw.col` instead of `pl.col`
  - using `nw.from_native` at the start of the function for each input, and `nw.to_native` at the end

Once you have a notebook which executes, please share it here!

---

**Note**: if you see `q5_pandas` or `q5_ibis`, you can ignore them for the purpose of this issue, no need to copy them. Let's only focus on making sure that the Narwhals API is extensive enough for these queries

ugohuche:
I'll implement Q9 and Q10

ugohuche:
@MarcoGorelli I noticed that Narwhals doesn't yet have 'contains' in the ExprStringNamespace, which is used in query 9.

MarcoGorelli:
ah, thanks @ugohuche ! looks like we've found the next issue :) would you like to work on implementing `str.contains`? If you look at `str.ends_with`, I think it might be kind of similar, feel free to ask if you get stuck

once that's done, we can get back to this

ugohuche:
Yeah, I'd love to give it a try

[...]

PR #407, opened 2024-07-04
TPCH Queries 9 and 10

ugohuche:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues 

- Related issue #282
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

TPCH Benchmarking Queries 9 and 10


tpch/notebooks/q9/execute.ipynb
EdAbati, line 1: Comparing to [this](https://github.com/pola-rs/tpch/blob/main/queries/polars/q9.py#L36), the `.round(2)` is missing after the sum in line 40 of the 3rd cell

In case it was previously discussed to omit it, I think it should be added.

tpch/notebooks/q10/execute.ipynb
EdAbati, line 1: As in the other comment, I'd add the `round(2)` after `nw.sum`.

Out of curiosity, did we get an error if we don't create the `revenue` column before the group by?

EdAbati (commented):
Thanks for your PR :)

It looks good, I have just added a couple of comments. Not sure if they were discussed somewhere else.

tpch/notebooks/q10/execute.ipynb
ugohuche: Initially, the revenue column was created inside the ```agg``` with the ```round(2)```, but it gave a UserWarning of a "complex group-by expression" from the pandas API and it slows it down significantly for pandas.

MarcoGorelli (approved):
thanks @ugohuche , well done! and thanks @EdAbati for the review - I think it's fine to skip `round` as it would otherwise give the userwarning for the pandas-like dataframes

PR #410, opened 2024-07-05
RFC: Stable api

### LINKED ISSUE #259: RFC Perfect backwards compatibility policy
Curious to hear thoughts on this one - especially from @baggiponte 

Narwhals follows a strict subset of the Polars API. What happens if/when some part of this subset gets changed in Polars itself?

For example:
- pre 1.0, Polars coalesces overlapping columns in left joins by default
- post 1.0, it won't by default

So, what do we do in Narwhals? I'd suggest allowing people to specify which API version they want to follow in `pyproject.toml`, e.g.:
```toml
[tool.narwhals]
api_version = "0.20"
```

Then:
- if somebody specifies `api_version = "0.20"` in pyproject.toml, then we promise to keep giving them the pre 1.0 behaviour indefinitely (coalesce by default in left joins)
- if somebody specifies `api_version = "1.0"` in pyproject.toml (or if they don't specify it), then we give them the post 1.0 Polars behaviour (don't coalesce by default in left joins)

The idea is that if you implement your library using Narwhals, then so long as you specify an `api_version` in `pyproject.toml`, then your code should preserve the same behaviour indefinitely. You may occasionally need to bump your minimum Narwhals package version (so we can deal with working around upstream changes in pandas/Polars/etc), but you should never need to rewrite your dataframe logic, it'll be perfectly backwards-compatible.

---

Very keen to also get @FBruzzesi and @koaning 's take on this, as consumers. Would you appreciate / be open to specifying a Narwhals `api_version` in your `pyproject.toml`, in exchange for the promise that the Narwhals API you use won't change?

---

So long as we stick to fundamental dataframe operations, and a strict subset of Polars (rather than trying to reimplement _everything_...) then I think this should be doable!

MarcoGorelli:
closes #259 

The policy is laid out in `docs/backcompat.md`, and is:
- If you write code using `import narwhals.stable.v1 as nw`, then we promise to
  never change or remove any public function you're using.
- If we need to make a backwards-incompatible change, it will be pushed into
  `narwhals.stable.v2`, leaving `narwhals.stable.v1` unaffected.
- We will maintain `narwhals.stable.v1` indefinitely, even as `narwhals.stable.v2` and other
  stable APIs come out. For example, Narwhals version 1.0.0 will offer
  `narwhals.stable.v1`, whereas Narwhals 2.0.0 will offer both `narwhals.stable.v1` and
  `narwhals.stable.v2`.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

some todos:
- [x] factor out what can be factored out
- [x] add stable api document
- [x] override typing for methods that don't return Self
- [x] check typing for group-by - maybe make groupby generic?
- [x] test docstring consistency for overrided methods

diff:
- most of the changes are in tests, where I've done
  ```diff
  - import narwhals as nw
  + import narwhals.stable.v1 as nw
  ```
- the non-trivial changes are in:
  - docs/backcompat.md
  - narwhals/stable/v1.py
  - tests/stable_api_test.py
  - tests/translate/stable_narwhalify_test.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,15 @@ Seamlessly support all, without depending on any!
 
 - ✅ **Just use** a subset of **the Polars API**, no need to learn anything new
 - ✅ **Zero dependencies**, Narwhals only uses what
-  the user passes in, so you can keep your library lightweight
+  the user passes in so your library can stay lightweight
 - ✅ Separate **lazy** and eager APIs, use **expressions**
 - ✅ Support pandas' complicated type system and index, without
   either getting in the way
 - ✅ **100% branch coverage**, tested against pandas and Polars nightly builds
 - ✅ **Negligible overhead**, see [overhead](https://narwhals-dev.github.io/narwhals/overhead/)
 - ✅ Let your IDE help you thanks to **full static typing**, see [typing](https://narwhals-dev.github.io/narwhals/typing/)
+- ✅ **Perfect backwards compatibility policy**,
+  see [stable api](https://narwhals-dev.github.io/narwhals/backcompat/) for how to opt-in
 
 ## Used by / integrates with
 
--- a/docs/backcompat.md
+++ b/docs/backcompat.md
@@ -1,0 +1,92 @@
+# Perfect backwards compatibility policy
+
+Narwhals is primarily aimed at library maintainers rather than end users. As such,
+we need to take stability and backwards compatibility extra-seriously. Our policy is:
+
+- If you write code using `import narwhals.stable.v1 as nw`, then we promise to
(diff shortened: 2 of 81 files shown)


MarcoGorelli:
let's do this! 🚀 🎸 🍾

PR #420, opened 2024-07-05
clean: factor tests out of test_common

shriyakalakata:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #401 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


tests/frame/double_selected_test.py
```diff
@@ -0,0 +1,21 @@
...
+    expected = {"a": [2, 6, 4], "b": [8, 8, 12]}
+    compare_dicts(result_native, expected)
+    result = df.select("z", nw.col("a", "b") * 2)
+    result_native = nw.to_native(result)
+    expected = {"z": [7, 8, 9], "a": [2, 6, 4], "b": [8, 8, 12]}
+    compare_dicts(result_native, expected)
+    result = df.select("a").select(nw.col("a") + nw.all())
+    result_native = nw.to_native(result)
+    expected = {"a": [2, 6, 4]}
+    compare_dicts(result_native, expected)
```
MarcoGorelli, line 21: in this test we're calling `nw.col('a', 'b')`, so perhaps it's better placed in `tests/expr`?

same for the `double_test` from the previous PR actually, sorry for not having picked up on that straight away

tests/frame/rename_test.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any
+
+import narwhals as nw
+from tests.utils import compare_dicts
+
+
+def test_rename(constructor: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df = nw.from_native(constructor(data), eager_only=True)
+    result = df.rename({"a": "x", "b": "y"})
```
MarcoGorelli, line 10: nice! here, we're calling `df.rename`, so `tests/frame` is the right location

MarcoGorelli (commented):
Awesome! this is going to make it much easier to test different backends

Just one comment on test locations, which is applicable to all the added tests except for `rename`

shriyakalakata:
> Just one comment on test locations

I'll move them around! In general, when would you put a test for a function in `tests/frame` vs `tests/expr`?

MarcoGorelli:
thanks!
if the test is checking a dataframe method (e.g. `df.rename`), then `tests/frame`

if it's testing an expression method (e.g. `nw.col('a') * 2`), then `tests/expr`

shriyakalakata:
I see, thanks for the clarification!

tests/expr/sumh_literal_test.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any
+
+import narwhals as nw
+from tests.utils import compare_dicts
+
+
+def test_sumh_literal(constructor: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df = nw.from_native(constructor(data), eager_only=True)
+    result = df.with_columns(horizonal_sum=nw.sum_horizontal("a", nw.col("b")))
```
MarcoGorelli, line 10: this and the other one (`test_sumh`) are both testing `Expr.sum_horizontal` - could we put them both in the same file `tests/expr/sum_horizontal_test`?

MarcoGorelli (commented):
thanks for updating nearly there

tests/expr/sumh_literal_test.py
shriyakalakata: good idea!

MarcoGorelli (approved):
thanks @shriyakalakata !

PR #427, opened 2024-07-06
clean: factor `filter_test` and `filter_test_series` out of `test_common`

condekind:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue: https://github.com/narwhals-dev/narwhals/issues/401
- Closes: None

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Before the change, 6 tests were executed when I ran:
```
pytest tests/frame/test_common.py::test_filter tests/frame/test_common.py::test_filter_series
```
Now, 8 are executed with:
```
pytest tests/frame/filter_test.py
```
Is that expected?  

---

Also, I tried using `constructor_with_pyarrow` but two tests failed with errors like:
```
...
E       AttributeError: 'ArrowDataFrame' object has no attribute 'filter'
...
E       AttributeError: 'ArrowDataFrame' object has no attribute 'with_columns'
...
```

Any suggestions? Maybe there's an easy fix, I'm not too familiar with this repo yet.  

---

**One more question**: why aren't factored out tests using `nw.to_native(result)` and comparing `result_native` anymore? (just out of curiosity, I don't know why they were being converted to native to begin with)

### DIFF
--- a/tests/frame/filter_test.py
+++ b/tests/frame/filter_test.py
@@ -1,0 +1,22 @@
+from typing import Any
+
+import narwhals as nw
+from tests.utils import compare_dicts
+
+
+def test_filter(constructor: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df = nw.from_native(constructor(data), eager_only=True)
+    result = df.filter(nw.col("a") > 1)
+    expected = {"a": [3, 2], "b": [4, 6], "z": [8.0, 9.0]}
+    compare_dicts(result, expected)
+
+
+def test_filter_series(constructor: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df = nw.from_native(constructor(data), eager_only=True).with_columns(
+        mask=nw.col("a") > 1
+    )
+    result = df.filter(df["mask"]).drop("mask")
+    expected = {"a": [3, 2], "b": [4, 6], "z": [8.0, 9.0]}
+    compare_dicts(result, expected)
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
thanks @condekind !

MarcoGorelli:
> Now, 8 are executed with:

`constructor` parametrises over more backends - that's one more good reason to use it 👍

MarcoGorelli:
> Also, I tried using constructor_with_pyarrow but two tests failed with errors like:

that's almost certainly because we just haven't implemented some functionality for pyarrow yet

condekind:
> > Now, 8 are executed with:
> 
> `constructor` parametrises over more backends - that's one more good reason to use it 👍

Makes sense, thanks! I thought `constructor_with_pyarrow` was preferred since https://github.com/narwhals-dev/narwhals/issues/401 says at the end:

> note that constructor_with_pyarrow might not necessarily work yet - try it, and if it doesn't, just use constructor

If that's not the case I'll use `constructor` for others to come

MarcoGorelli:
thanks! we probably need to get a lot further with https://github.com/narwhals-dev/narwhals/issues/372 before we can use `constructor_with_pyarrow` more

would be curious to hear your thoughts on https://github.com/narwhals-dev/narwhals/pull/410 btw (only if it interests you though! I understand this is outside the scope of the internship you're doing, I don't want to put pressure on you to do anything that's outside your interests / time 🤗 )

condekind:
I'll take a look. It sounds very nice for users to have that stable API :thinking:

PR #445, opened 2024-07-09
Refactor: Re enable and fix Ruff's TRY Checks

ELC:

- [X] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/group_by.py
+++ b/narwhals/_pandas_like/group_by.py
@@ -146,9 +146,10 @@ def agg_pandas(  # noqa: PLR0913
         try:
             result_simple = grouped.agg(aggs)
         except AttributeError as exc:
-            raise RuntimeError(
+            error_message = (
                 "Failed to aggregated - does your aggregation function return a scalar?"
-            ) from exc
+            )
+            raise RuntimeError(error_message) from exc
         result_simple.columns = [f"{a}_{b}" for a, b in result_simple.columns]
         result_simple = result_simple.rename(columns=name_mapping).reset_index()
         return from_dataframe(result_simple.loc[:, output_names])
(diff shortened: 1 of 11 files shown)


FBruzzesi (commented):
Thanks for the PR!

In the code blocks where TRY003 and TRY004 are satisfied we currently use `msg = ... `. 
I would rather aim to keep it the same variable name for the entire codebase. 

Personally `error_message` is a bit lengthy, while `msg` may be too generic.

Maybe `err_msg` is the compromise we can settle with?

Edit: I just noticed #447 that renames all the other instances

MarcoGorelli:
yeah let's get the other one in first

narwhals/_pandas_like/group_by.py
```diff
@@ -146,9 +146,10 @@ def agg_pandas(  # noqa: PLR0913
         try:
             result_simple = grouped.agg(aggs)
         except AttributeError as exc:
-            raise RuntimeError(
+            error_message = (
                 "Failed to aggregated - does your aggregation function return a scalar?"
-            ) from exc
+            )
+            raise RuntimeError(error_message) from exc
```
MarcoGorelli, line 152: this is fine, happy to add this (with `msg` instead of `error_message`)

ELC:
Changed all instances to `msg`

I do agree with FBruzzesi and would avoid variable names that are too short, but this is not critical

FBruzzesi (approved):
Thanks for adjusting 👌

KittyKittenGIF

PR #447, opened 2024-07-09
Refactor: Re enable and fix Ruff's EM101 Check

ELC:

- [X] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import platform
 import sys
+from typing import Any
 from typing import Iterable
 from typing import Literal
 from typing import TypeVar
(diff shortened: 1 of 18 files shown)


narwhals/functions.py
```diff
@@ -21,14 +22,14 @@
 def concat(
     items: Iterable[FrameT],
     *,
-    how: Literal["horizontal", "vertical"] = "vertical",
+    how: Literal["horizontal", "vertical"] | Any = "vertical",
```
MarcoGorelli, line 25: why `Any`?

MarcoGorelli:
OK with this check, but please explicitly call out any changes you've made which are not simple linter-driven ones (such as changing type annotations)

MarcoGorelli:
let's stick with `msg`, thanks anyway for your PR

narwhals/functions.py
ELC: Because how else would it enter into the if statement? If it is typed as `Literal["vertical", "horizontal"]` there should be no code assuming is something else like the if block at the beginning

PR #455, opened 2024-07-09
feat: add `to_lowercase` and `to_uppercase` to `Expr.str` namespace 

lucianosrp:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

This feature allows to convert strings to upper and lower case variants, similar to what is already possible with polars' [`Expr.str.to_lowercase()`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.to_lowercase.html#polars.Expr.str.to_lowercase)  and pandas' [`Series.str.lower()`](https://pandas.pydata.org/docs/reference/api/pandas.Series.str.lower.html)
### Unicode caveat
There is a caveat for PyArrow's backend that converts the character 'ß' to uppercase 'ẞ' instead of 'SS'. There are probably few others Unicode edgecases but I think the vast majority of users won't be affected. This has been documented in the `to_uppercase` method and linked the issue: **https://github.com/apache/arrow/issues/34599** 

This was my first contribution here, feedback is always welcomed 🙏

### DIFF
--- a/docs/api-reference/expressions_str.md
+++ b/docs/api-reference/expressions_str.md
@@ -11,5 +11,7 @@
         - starts_with
         - tail
         - to_datetime
+        - to_uppercase
+        - to_lowercase
       show_source: false
       show_bases: false
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -2081,6 +2081,12 @@ class ExprStringNamespace:
             lambda plx: self._expr._call(plx).str.to_datetime(format=format)
         )
 
+    def to_uppercase(self) -> Expr:
+        return self._expr.__class__(lambda plx: self._expr._call(plx).str.to_uppercase())
+
+    def to_lowercase(self) -> Expr:
+        return self._expr.__class__(lambda plx: self._expr._call(plx).str.to_lowercase())
+
 
 class ExprDateTimeNamespace:
     def __init__(self, expr: Expr) -> None:
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -2087,7 +2087,7 @@ class ExprStringNamespace:
 
         Notes:
             The PyArrow backend will convert 'ß' to 'ẞ' instead of 'SS'.
-            For more info see: https://github.com/apache/arrow/issues/34599
+            For more info see [the related issue](https://github.com/apache/arrow/issues/34599).
             Other Unicode edge cases may be present and are not tested.
 
         Examples:
--- a/narwhals/expression.py
+++ b/narwhals/expression.py
@@ -2082,9 +2082,88 @@ class ExprStringNamespace:
         )
 
     def to_uppercase(self) -> Expr:
+        r"""
+        Transform string to uppercase variant.
+
+        Notes:
+            The PyArrow backend will convert 'ß' to 'ẞ' instead of 'SS'.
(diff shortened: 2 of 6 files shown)


FBruzzesi (approved):
Thanks for the PR! This is neat 🙌🏼

CatCatJumpingGIF

lucianosrp:
Will fix the doctests ! 🫠

docs/api-reference/expressions_str.md
```diff
@@ -11,5 +11,7 @@
         - starts_with
         - tail
         - to_datetime
+        - to_uppercase
+        - to_lowercase
```
MarcoGorelli, line 15:
```suggestion
        - to_datetime
        - to_lowercase
        - to_uppercase
```

MarcoGorelli (commented):
Amazing! Well done @lucianosrp for figuring it all out!

I just have some really minor comments

I think we can probably also support this for pyarrow.Table, right? Fancy changing `constructor` to `constructor_with_pyarrow` in the test, and implementing this in `narwhals/_arrow/series.py` and `narwhals/_arrow/expr.py`? It should be quite similar to what you've already done

Again, seriously good work here, great to have you on board 🙌

lucianosrp:
@MarcoGorelli 
> I think we can probably also support this for pyarrow.Table, right? Fancy changing `constructor` to `constructor_with_pyarrow` in the test, and implementing this in `narwhals/_arrow/series.py` and `narwhals/_arrow/expr.py`? It should be quite similar to what you've already done

Ah, right! I probably have missed this..I will add that too

narwhals/expression.py
```diff
@@ -2081,6 +2081,92 @@ def to_datetime(self, format: str) -> Expr:  # noqa: A002
...
         )
 
+    def to_uppercase(self) -> Expr:
+        r"""
+        Transform string to uppercase variant.
+
+        Notes:
+            The PyArrow backend will convert 'ß' to 'ẞ' instead of 'SS'.
+            For more info see [the related issue](https://github.com/apache/arrow/issues/34599).
+            There may be other unicode-edge-case-related variations across implementations.
```
FBruzzesi, line 2091: Not sure how much we want to expand on this topic, but I would consider adding some additional information, such as mentioning that pyarrow is based on [utf8proc](https://juliastrings.github.io/utf8proc/) for sake of completeness.

WDYT?

MarcoGorelli (approved):
amazing, thanks @lucianosrp !

we're really keen to expand pyarrow functionality so any PRs in that direction are extremely welcome!

narwhals/expression.py
MarcoGorelli: sure, happy to take this as a follow-up! 👍 for now I'll merge as-is

lucianosrp:
My pleasure, @MarcoGorelli 
Thanks also for the feedback @FBruzzesi !

This is an awesome tool! 🔥 I am happy to help expanding it!

PR #462, opened 2024-07-09
bug: fix `__getitem__` for tuple of row and col keys

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

Currently this will break:

```python
import polars as pl
import pandas as pd
import narwhals as nw

df = pd.DataFrame({"a": [0, 1,2], "b": [0, 1,2]})
df_pl = pl.DataFrame(df)

@nw.narwhalify
def func(df):
    return df[[0,2],0]

func(df_pl)
func(df)
```

~The code I added only works for `polars`, I still have to figure out how to make pandas-like and pyarrow happy.~

This functionality is needed for #355 :)

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -478,7 +478,9 @@ class DataFrame(BaseFrame[FrameT]):
 
                 - str: extract column
                 - slice or Sequence of integers: slice rows from dataframe.
-
+                - tuple of Sequence of integers and str or int: slice rows and extract column at the same time.
+                  If the second element of the tuple is an integer, it is interpreted as the column index. Otherwise,
+                  it is interpreted as the column name.
         Notes:
             In contrast with Polars, pandas allows non-string column names.
             If you don't know whether the column name you're trying to extract


narwhals/dataframe.py
```diff
@@ -408,16 +408,21 @@ def shape(self) -> tuple[int, int]:
...
     @overload
     def __getitem__(self, item: str) -> Series: ...
 
     @overload
     def __getitem__(self, item: slice) -> Self: ...
 
-    def __getitem__(self, item: str | slice | Sequence[int]) -> Series | Self:
-        if isinstance(item, str):
+    def __getitem__(
+        self, item: str | slice | Sequence[int] | tuple[Sequence[int], str | int]
```
EdAbati, line 423: Typing here is getting a bit crowded, should we have [something similar to `polars`](https://github.com/pola-rs/polars/blob/main/py-polars/polars/_typing.py#L256)?

narwhals/_pandas_like/dataframe.py
```diff
@@ -106,6 +114,20 @@ def __getitem__(self, item: str | slice) -> PandasLikeSeries | PandasLikeDataFra
                 backend_version=self._backend_version,
             )
 
+        elif isinstance(item, tuple) and len(item) == 2:
+            from narwhals._pandas_like.series import PandasLikeSeries
+
+            if item[1] in self.columns:
+                native_series = self._native_dataframe.loc[item]
+            else:
+                native_series = self._native_dataframe.iloc[item]
```
EdAbati, line 123: my understanding is that the the first item in the tuple is always a sequence of `int`s and the second item can be `str` or `int`. Given that we could have `int` column names in pandas, I thought that this `if item[1] in self.columns` should be ok to decide whether to use `iloc` or `loc`. 

but I feel like I am missing some edge case here 🤔 what do you think?

MarcoGorelli:
thanks for your PR!

the "it might be a column by position, it might be a column by name" part kind of scares me - are scikit-learn using it like this? or are they always using it by position?

narwhals/_pandas_like/dataframe.py
EdAbati: I agree I didn't like myself what I committed. :D I really like this solution, thank you for adding `get_column`

FYI what I was trying to support is [this use case](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/inspection/_permutation_importance.py#L67).
Here both rows and columns are accessed via position.

EdAbati:
I am also happy to close this PR if you think we should not support this type of args in `__getitem__` 

`X[row_ids, col_idx]` can now be achieved with 
```python
col_name = X.columns[col_idx]
X.get_column(col_name)[row_ids]
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -115,6 +123,23 @@ def __getitem__(self, item: str | slice) -> PandasLikeSeries | PandasLikeDataFra
                 backend_version=self._backend_version,
             )
 
+        elif isinstance(item, tuple) and len(item) == 2:
+            from narwhals._pandas_like.series import PandasLikeSeries
+
+            if isinstance(item[1], str):
+                native_series = self._native_dataframe.loc[item]
+            elif isinstance(item[1], int):
+                native_series = self._native_dataframe.iloc[item]
```
MarcoGorelli, line 132: nice - simple and predictable

narwhals/dataframe.py
MarcoGorelli: sure (not a blocker though)

MarcoGorelli (approved):
thanks @EdAbati !

the test reminds that me we need some better testing utils in Narwhals 😄

PR #480, opened 2024-07-10
Docs/attribute docstrings

lucianosrp:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Another small one here ! I promise I will get back to more Pyarrow implementations after this one! 😂

While annotating functions with Narwhals' custom types, I wanted to double check the "definition" of those.

The explanation for each type was previously given as as comment in the source code itself, which is usually fine. But this additional information does not propagate in the "definition" view of some (most?) text editors. 

By adding a string litteral immediately after the value assignment (from [PEP-258](https://peps.python.org/pep-0258/#attribute-docstrings)), the information is then made available in the "definition/documentation views", see example below:

**Before**
Screenshot 2024-07-10 113310

**Code Change**
```python
IntoExpr: TypeAlias = Union["Expr", str, "Series"]
"""Anything which can be converted to an expression."""
```

**After**
Screenshot 2024-07-10 113350

### DIFF
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -29,14 +29,17 @@ if TYPE_CHECKING:
         def join(self, *args: Any, **kwargs: Any) -> Any: ...
 
 
-# Anything which can be converted to an expression.
 IntoExpr: TypeAlias = Union["Expr", str, "Series"]
-# Anything which can be converted to a Narwhals DataFrame.
+"""Anything which can be converted to an expression."""
+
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
didn't know that this would work, thanks @lucianosrp !

PR #482, opened 2024-07-10
clean: (partial) factor `test_common`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #401 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Unsure how to handle `to_pandas` and `to_numpy` failures


tests/frame/to_pandas_test.py
```diff
@@ -0,0 +1,27 @@
...
+)
+def test_convert_pandas(constructor_with_pyarrow: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df_raw = constructor_with_pyarrow(data)
+    result = nw.from_native(df_raw).to_pandas()  # type: ignore[union-attr]
+
+    if constructor_with_pyarrow.__name__.startswith("pandas"):
+        expected = constructor_with_pyarrow(data)
+    else:
+        expected = pd.DataFrame(data)
```
FBruzzesi, line 25: I am not too happy with this, but `asserting_frame_equal` clearly takes the backend into consideration. 
Alternative would be to just check for values

tests/frame/to_numpy_test.py
```diff
@@ -0,0 +1,21 @@
...
+import narwhals.stable.v1 as nw
+
+
+def test_convert_numpy(constructor: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.1, 8, 9]}
+    df_raw = constructor(data)
+    result = nw.from_native(df_raw, eager_only=True).to_numpy()
+
+    expected = np.array([[1, 3, 2], [4, 4, 6], [7.1, 8, 9]]).T
+    np.testing.assert_array_equal(result, expected)
```
FBruzzesi, line 16: `pandas_nullable` and `pandas_pyarrow` will convert `7.0` to integer, therefore the type check fails. Using `7.1` fixes that

tests/frame/to_pandas_test.py
```diff
@@ -0,0 +1,27 @@
...
+
+import narwhals.stable.v1 as nw
+from narwhals.utils import parse_version
+
+
+@pytest.mark.filterwarnings("ignore:.*Passing a BlockManager.*:DeprecationWarning")
+@pytest.mark.skipif(
+    parse_version(pd.__version__) < parse_version("2.0.0"),
+    reason="too old for pandas-pyarrow",
+)
```
MarcoGorelli, line 16: maybe we should be using `request` and only skipping the elements of `constructor` where this fails...but ok for a follow-up

tests/frame/to_pandas_test.py
MarcoGorelli: looks good, not sure what else we can do here

MarcoGorelli (approved):
solid effort, thanks so much for doing this 💪

PR #484, opened 2024-07-10
feat: Dask Support Implementation

### LINKED ISSUE #396: [Enh]: Add support for dask

_No response_

Dask is a dataframe library built on top of pandas to allow distributed computation of dataframes (like pyspark for example). It's API is a strict subset of the pandas API so in theory integrating it would be feasible. I'd be really down for contributing / looking into how support might work, but wanted to check if it's within the scope/aims of Narwhals or if it's been considered before? (Can't see anything on the issue tracker so I guess no?)

Potentially a lot of pandas code could be reused although I haven't looked in depth to say that which much confidence.

_No response_

I'd love to put in some contributions if it's within scope, I'm pretty experienced with Dask (have made some minor constibutions to dask-dataframe) but not so much with Narwhals. Would be great to get sense of interest for this as a feature, I realise adding in support for a whole new dataframe library is a big undertaking.

benrutter:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #396
- Closes #396

## Checklist

- [x] Tests added 
- [ ] Documented the changes (I don't think so, is there a changelog to add to?)

Hello,

This is still a draft PR (although everything is in a working state, there's a very unperformance step in it I've put in that I do need to take out) but I'd love to get some feedback around making integration as sleek as possible!

I've implemented a bunch of logic with something along the lines of:

```python
if self._integration == "dask":
    return do_some_bespoke_thing()
return the_happy_path()
```
Which I *think* for the most part is fine because that's generally how other implementations like modin etc are done? Not sure if there's any guidance around another way of doing this? (or maybe I should just put in some friendly comments where that happens to explain why dask is implemented differently). I actually found it pretty clear to follow along, but the alternative I could think of woud be subclassing or something.

The other thing I've done (mentioned over on the Narwhals discord) is implement `.collect()` to `.compute()` a dask dataframe, which seems like good behaviour. Dask dataframes are a little like polars lazyframes, and computing them forces all data into a single noded dataframe.

There are also a few things where something just isn't possible with dask, for instance cross-joins. Which dask doesn't support natively. Theoretically you could hack something in by doing something like adding two columns that are always true, joining on that, then dropping the column, but I thought it was a better move to maintain the implementations of dask as mark as "not implemented" (particularly as my understanding is dask intentionally doesn't implement cross joins for performance reasons).

I was thinking it might be nice to put in a decorator along the lines of:

```
@not_implemented_in(["dask", "modin"])
def self.some_method(x: Any) -> Any:
   ...
```

To nicely raise a NotImplementedError if that method gets called with one of those backends? Any thoughts?

It's my first time working with the Narwhal's codebase - I've had a blast running through it! I'm probably not familiar with all the conventions etc, so please do let me know if I've missed something major.


narwhals/_pandas_like/series.py
```diff
@@ -138,6 +141,8 @@ def _from_iterable(
         )
 
     def __len__(self) -> int:
+        if self._implementation == "dask":
+            return len(self._series)
         return self.shape[0]
```
MarcoGorelli, line 146: we might as well just return `len(self._series)` for all of them, thus avoiding the if statement?

MarcoGorelli (commented):
Awesome effort, thanks @benrutter !

a few comments:
- I think a subclass or a decorator should be fine. I think there's not that many if/then statements you had to add? maybe a decorator is easier then. for `shape` we can unify the implementations under `len(self._native_series)`
- merge conflicts: there are some merge conflicts, but it's mostly down to some things having been renamed:
  - PandasDataFrame -> PandasLikeDataFrame
  - ._series -> ._native_series
  - implementation == "dask" => implementation is Implementation.DASK

The major refactors are over (for pandas-like), I promise not to do this again 😄  😳 Fancy fixing up the merge conflicts?

narwhals/_pandas_like/dataframe.py
```diff
@@ -165,7 +167,8 @@ def select(
         if not new_series:
             # return empty dataframe, like Polars does
             return self._from_dataframe(self._dataframe.__class__())
-        new_series = validate_indices(new_series)
+        if self._implementation != "dask":
+            new_series = validate_indices(new_series)
```
MarcoGorelli, line 171: rather than this, could we just always validate indices, and then, if indices aren't aligned, then for dask, we raise an error, whereas for other implementations we use `set_axis` and continue?

benrutter:
@MarcoGorelli absolutely! I'll do some "tidy up" changes and then work through the merge conflicts.

benrutter:
Ok, I *thought* I'd sorted out the merge conflicts, and at least `test_common` is working for both. But looks like I've introduced some kind of syntax error somewhere else or something. There's a lot of failing tests, oh dear. . .

I'll have a bit more of a look some time soon.

MarcoGorelli:
awesome stuff! love the decorator

a lot of `narwhals/_pandas_like/utils.py` was moved to `narwhals/_expression_parsing.py`, as it's reused between pandas-like APIs and pyarrow

benrutter:
Ah thanks! That probably explains it then- I might just need to implement a few small bits over there then.

benrutter:
Ok, all tests passing bar one which I'll mention in a sec (and sorted out the latest conflicts).

Looks like one of the tests which checks the API docs for stable.v1 is passing, but I'm not 100% sure of the intention behind it so don't want to mess too much without checking.

Is the idea that I should go in and update the documentation so that v1 matches elsewhere?

MarcoGorelli:
> Is the idea that I should go in and update the documentation so that v1 matches elsewhere?

yup, if you update the main namespace docs, you should also update the v1 docs

benrutter:
Awesome - thanks @MarcoGorelli for all the guidance! Looks like all the tests are passing now (I spotted they're failing in 3.8 python but that just looks like the old type hinting API not being supported rather than anything new here)

narwhals/_pandas_like/dataframe.py
```diff
@@ -66,6 +67,8 @@ def __native_namespace__(self) -> Any:
             return get_modin()
         if self._implementation is Implementation.CUDF:  # pragma: no cover
             return get_cudf()
+        if self._implementation == "dask":  # pragma: no cover
```
MarcoGorelli, line 70: this should be `is Implementation.DASK`

narwhals/_pandas_like/group_by.py
```diff
@@ -28,10 +28,13 @@ class PandasLikeGroupBy:
     def __init__(self, df: PandasLikeDataFrame, keys: list[str]) -> None:
         self._df = df
         self._keys = list(keys)
+        keywords = {}
+        if df._implementation != "dask":
```
MarcoGorelli, line 32: same

narwhals/_pandas_like/group_by.py
```diff
@@ -56,13 +59,18 @@ def agg(
                 raise ValueError(msg)
             output_names.extend(expr._output_names)
 
+        dataframe_is_empty = (
+            self._df._native_dataframe.empty
+            if self._df._implementation != Implementation.DASK
+            else len(self._df._native_dataframe) == 0
+        )
```
MarcoGorelli, line 66: can we just use `self._df.is_empty()`?

narwhals/_pandas_like/series.py
```diff
@@ -107,12 +109,15 @@ def __native_namespace__(self) -> Any:
             return get_modin()
         if self._implementation is Implementation.CUDF:  # pragma: no cover
             return get_cudf()
+        if self._implementation == "dask":  # pragma: no cover
```
MarcoGorelli, line 112: same

narwhals/_pandas_like/utils.py
```diff
@@ -53,7 +57,10 @@ def validate_column_comparand(index: Any, other: Any) -> Any:
         if other.len() == 1:
             # broadcast
             return other.item()
-        if other._native_series.index is not index:
+        if (
+            other._native_series.index is not index
+            and other._implementation != Implementation.DASK
+        ):
```
MarcoGorelli, line 63: what if `other._native_series.index is not index` and `other._implementation is Implementation.DASK`? I think we need to raise an error message in that case

narwhals/_pandas_like/utils.py
```diff
@@ -79,15 +86,65 @@ def validate_dataframe_comparand(index: Any, other: Any) -> Any:
...
                 implementation=other._implementation,
                 backend_version=other._backend_version,
             )
         return other._native_series
-    msg = "Please report a bug"  # pragma: no cover
+        return other._series
+    raise AssertionError("Please report a bug")
+
+
+def maybe_evaluate_expr(df: PandasDataFrame, expr: Any) -> Any:
```
MarcoGorelli, line 104: this, and the function below, are now in `narwhals/_expression_parsing.py`, please remove them from here and use those

narwhals/_pandas_like/utils.py
```diff
@@ -171,6 +233,10 @@ def vertical_concat(
         mpd = get_modin()
 
         return mpd.concat(dfs, axis=0)
+    if implementation == "dask":  # pragma: no cover
```
MarcoGorelli, line 236: same

narwhals/_pandas_like/utils.py
```diff
@@ -194,6 +260,20 @@ def native_series_from_iterable(
         mpd = get_modin()
 
         return mpd.Series(data, name=name, index=index)
+    if implementation == "arrow":
+        pa = get_pyarrow()
+        return pa.chunked_array([data])
+    if implementation is Implementation.ARROW:  # pragma: no cover
```
MarcoGorelli, line 266: typo?

narwhals/_pandas_like/utils.py
```diff
@@ -218,6 +298,9 @@ def set_axis(
         kwargs["copy"] = False
     else:  # pragma: no cover
         pass
+    if implementation == "dask":
```
MarcoGorelli, line 301: same

narwhals/_pandas_like/utils.py
```diff
@@ -449,6 +532,8 @@ def to_datetime(implementation: Implementation) -> Any:
         return get_modin().to_datetime
     if implementation is Implementation.CUDF:
         return get_cudf().to_datetime
+    if implementation == "dask":
```
MarcoGorelli, line 535: same

narwhals/dependencies.py
```diff
@@ -32,6 +32,11 @@ def get_modin() -> Any:  # pragma: no cover
     return None
 
 
+def get_dask() -> Any:
+    """Get dask.dataframe module (if already improted - else return None)."""
```
MarcoGorelli, line 36: typo improted

tests/utils.py
```diff
@@ -51,6 +57,17 @@ def maybe_get_modin_df(df_pandas: pd.DataFrame) -> Any:
             return mpd.DataFrame(df_pandas.to_dict(orient="list"))
 
 
+def maybe_get_dask_df(df_pandas: pd.DataFrame) -> Any:
```
MarcoGorelli, line 60: where is this used?

MarcoGorelli (commented):
thanks @benrutter !

left some comments, there's some pre-commit checks to fixup too, but awesome work here

benrutter:
Ah quick update on this, I actually spotted that "get_dask()" (and get_modin()) obviously only imports dask *if it's already imported*. Which makes total sense, but also meant that dask wasn't being included in the new tests at all, since I hadn't added an initial import statement anywhere.

I've put one + an ImportError suppression in the conftest which has worked nicely (yaaay! 🙌) but that's also turned up a whole bunch of errors in the test.

I suspect its a mix of:
- Stuff that's not supported in dask so the test just needs to be skippped
- Stuff where I've not implemented something correctly / at all yet and just hadn't realised till now
- Stuff where I've borked the  implementation in the merge conflict

I'll work through, and the failing tests actually give me better confidence that the final integration should be reliable and nicely tested. Might take a little time though, let me know if it'd be easier for me to close off the PR while I work on it and reopen another later or not.

MarcoGorelli:
thanks @benrutter for the update! up to you if you want to keep this PR or open a new one, whichever's easiest for you

MarcoGorelli:
i'll work on updating this, would quite like to see it, but I think it should be back lazyframe, not dataframe?

benrutter:
Oh yikes! Thanks for taking a look, I think it *might* look worse than it is because some of the testing stuff like .to_dict isn't supported. Maybe I'm being naively optimistic 😅

I haven't taken a look at this for a little but but I'm hoping to spend some time hacking on it again soon.

benrutter:
Closing off for now so that dask functionality can be implemented in line with #566 later.

I'll keep playing around with the fork though.

PR #489, opened 2024-07-11
feat: Added Expr and Series Count method

ugohuche:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes


MarcoGorelli (commented):
amazing @ugohuche !

seeing as you've done it for arrow as well, does it work to use `constructor_with_pyarrow` in the test?

we might not yet have `nw.all` working for arrow, so you may need to use `nw.col('a', 'b', 'z')` instead

MarcoGorelli:
awesome @ugohuche ! now you just need `count` to `docs/api-reference/series.md` and `docs/api-reference/expr.md`

ugohuche:
> awesome @ugohuche ! now you just need `count` to add `docs/api-reference/series.md` and `docs/api-reference/expr.md`

I just added it now, but I have one last failing check and I don't quite understand it 🥲

narwhals/series.py
```diff
@@ -354,6 +354,34 @@ def mean(self) -> Any:
...
+
+            We define a library agnostic function:
+
+            >>> @nw.narwhalify
+            ... def func(s_any):
+            ...     return s_any.count()
+
+            We can then pass either pandas or Polars to `func`:
+
+            >>> func(s_pd)  # doctest:+SKIP
```
MarcoGorelli, line 377: why do we have to doctest skip this one? does it work without the skip?

MarcoGorelli (commented):
wonderful stuff! 

I've just left a question

the CI failure was unrelated, due to an internal warning from pandas which we can't do anything about - I've silenced it in https://github.com/narwhals-dev/narwhals/pull/497

narwhals/series.py
ugohuche: It's because of the numpy changes, as the new version causes pandas to output differently i.e np.int64(3), and causes the newer python tests to pass and older one's to fail.

MarcoGorelli: aah I see thanks - OK, tests are passing like this, let's ship it 🚀

MarcoGorelli (approved):
amazing thanks @ugohuche !

PR #491, opened 2024-07-11
feat: pyarrow `filter`, comparand and logic operators

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372

## Checklist

- [x] Tests added 
- [x] Documented the changes

So the story goes as follow: I wanted to implement `.filter(...)` method, which was ok-ish to achieve. However when running tests, operations were not supported, so I started to take care of some dunder methods.

I am opening as a draft as:

- Not sure about `__r<operator>__` methods, but I would like to add them if possible (maybe separate PR?)
- Some operations may need to be double checked (see code comments)


narwhals/_arrow/namespace.py
```diff
@@ -116,3 +119,10 @@ def all(self) -> ArrowExpr:
             output_names=None,
             backend_version=self._backend_version,
         )
+
+    def all_horizontal(self, *exprs: IntoArrowExpr) -> ArrowExpr:
+        # Why is this showing up as uncovered? It defo is?
```
FBruzzesi, line 124: aaaaaaaaah copy pasting!

narwhals/_arrow/series.py
```diff
@@ -70,6 +118,35 @@ def __mul__(self, other: Any) -> Self:
         other = validate_column_comparand(other)
         return self._from_native_series(pc.multiply(self._native_series, other))
 
+    def __truediv__(self, other: Any) -> Self:
+        pc = get_pyarrow_compute()
+        ser = self._native_series
+        other = validate_column_comparand(other)
+        return self._from_native_series(pc.divide(ser, other))
```
FBruzzesi, line 125: Notice that `pc.divide` maintain datatype (see comment in the test)

narwhals/_arrow/series.py
```diff
@@ -70,6 +118,35 @@ def __mul__(self, other: Any) -> Self:
...
+
+    def __floordiv__(self, other: Any) -> Self:
+        pc = get_pyarrow_compute()
+        ser = self._native_series
+        other = validate_column_comparand(other)
+        return self._from_native_series(pc.floor(pc.divide(ser, other)))
```
FBruzzesi, line 131: Need to double check behavior for negative numbers

narwhals/_arrow/series.py
```diff
@@ -70,6 +118,35 @@ def __mul__(self, other: Any) -> Self:
...
+        pc = get_pyarrow_compute()
+        ser = self._native_series
+        other = validate_column_comparand(other)
+        return self._from_native_series(pc.power(ser, other))
+
+    def __mod__(self, other: Any) -> Self:
+        pc = get_pyarrow_compute()
+        ser = self._native_series
+        other = validate_column_comparand(other)
+        res = pc.subtract(ser, pc.multiply(pc.divide(ser, other), other))
```
FBruzzesi, line 143: As `__mod__` should work among integers, I am exploiting the fact that `pc.divide` will yield the integer division
(namely the equivalent of `s - s//other * other`)

tests/expr/arithmetic_test.py
```diff
@@ -0,0 +1,32 @@
...
+def test_arithmetic(
+    attr: str, rhs: Any, expected: list[Any], constructor_with_pyarrow: Any, request: Any
+) -> None:
+    if "pandas_pyarrow" in str(constructor_with_pyarrow) and attr == "__mod__":
+        request.applymarker(pytest.mark.xfail)
+
+    data = {"a": [1, 2, 3]}
+    df = nw.from_native(constructor_with_pyarrow(data))
+    result = df.select(getattr(nw.col("a"), attr)(rhs))
+    compare_dicts(result, {"a": expected})
```
FBruzzesi, line 32: Not including right operations as polars do not support them either?!

tests/expr/filter_test.py
```diff
@@ -11,8 +11,8 @@
 }
 
 
-def test_filter(constructor: Any) -> None:
-    df = nw.from_native(constructor(data), eager_only=True)
+def test_filter(constructor_with_pyarrow: Any) -> None:
+    df = nw.from_native(constructor_with_pyarrow(data), eager_only=True)
```
FBruzzesi, line 15: Yeeey 🚀

tests/series/arithmetic_test.py
```diff
@@ -15,18 +15,25 @@
         ("__add__", 1, [2, 3, 4]),
         ("__sub__", 1, [0, 1, 2]),
         ("__mul__", 2, [2, 4, 6]),
-        ("__truediv__", 2, [0.5, 1.0, 1.5]),
+        ("__truediv__", 2.0, [0.5, 1.0, 1.5]),
```
FBruzzesi, line 18: Related to the comment above, since the data are all integers, if other is also an integer, then the data type is maintained. This is different from how polars and pandas handle it

tests/series/to_frame_test.py
```diff
@@ -6,6 +6,6 @@
 data = [1, 2, 3]
 
 
-def test_eq_ne(constructor_series: Any) -> None:
+def test_to_frame(constructor_series: Any) -> None:
```
FBruzzesi, line 9: I guess it was just some copy pasting in the past :)

MarcoGorelli:
thanks @FBruzzesi !

regarding division, I think we need to make an api decision here. I think `pc.divide` matches `__floordiv__` in polars/pandas, but for `__truediv__`, we should probably cast to float?

should we check if the numerator is int, and if so, cast it to float, in arrow `__truediv__`?

FBruzzesi:
> should we check if the numerator is int, and if so, cast it to float, in arrow __truediv__?

I need to double/triple check this, but if I am not mistaken, if we convert to float and the result is safe to be casted to int e.g. `[2, 4, 6] /2`, when creating a new array then it will be automatically casted to int.

I am saying this as consequence of [this comment](https://github.com/narwhals-dev/narwhals/pull/482/files#r1672489693)

narwhals/_arrow/series.py
```diff
@@ -70,6 +118,53 @@ def __mul__(self, other: Any) -> Self:
...
+
+        if original_type_is_int:
+            ser = ser.cast(pa.float64())
+
+        other = validate_column_comparand(other)
+        result = pc.divide(ser, other)
+
+        if original_type_is_int:
+            with contextlib.suppress(pa.ArrowInvalid):
+                result = result.cast(native_type)
```
FBruzzesi, line 141: TIL: this is same of

```py
try:
    result = result.cast(native_type)
except pa.ArrowInvalid:
    pass
```

MarcoGorelli: I'm not sure we should do this, perhaps we should follow Polars here?
```
In [1]: import polars as pl

In [2]: pl.Series([4, 6]) / pl.Series([2, 3])
Out[2]:
shape: (2,)
Series: '' [f64]
[
        2.0
        2.0
]
```

FBruzzesi: Oh I see! Thanks for pointing that out! Adjusting it now!

narwhals/_arrow/series.py
MarcoGorelli: pandas does some crazy stuff here

https://github.com/pandas-dev/pandas/blob/262fcfbffcee5c3116e86a951d8b693f90411e68/pandas/core/arrays/arrow/array.py#L124-L154

narwhals/_arrow/series.py
```diff
@@ -70,6 +118,45 @@ def __mul__(self, other: Any) -> Self:
...
+        pc = get_pyarrow_compute()
+
+        ser = self._native_series
+        native_type = ser.type
+        original_type_is_int = pa.types.is_integer(
+            native_type
+        ) or pa.types.is_unsigned_integer(native_type)
+
+        if original_type_is_int:
+            ser = ser.cast(pa.float64())
```
MarcoGorelli, line 132: this is what pandas does for pyarrow dtypes

https://github.com/pandas-dev/pandas/blob/262fcfbffcee5c3116e86a951d8b693f90411e68/pandas/core/arrays/arrow/array.py#L108-L122

I'm wondering why they use `safe=False` and also cast the denominator...will check this

MarcoGorelli (commented):
legend for adding filter and the rest! thanks so much

I think `__truediv__` and `__floordiv__` might require more attention, to avoid running into the same issues which pandas has already run into - should we split those two off into a separate PR, so we can quickly merge the rest of the awesome improvements here?

MarcoGorelli:
thanks!

are we sure about `__mod__`? I tried it out, and get some failures (just for pyarrow) with this hypothesis test:
```
from hypothesis import given, assume
import hypothesis.strategies as st
import pyarrow as pa
import pandas as pd
import polars as pl

@nw.narwhalify
def mod_func(x: nw.Series, y: nw.Series) -> nw.Series:
    return x % y

@given(
    left= st.integers(min_value=-100, max_value=100),
    right= st.integers(min_value=-100, max_value=100)
)
def test_mod(left: int, right: int):
    assume(right != 0)
    result = mod_func(pd.Series([left]), pd.Series([right]))
    expected = left % right
    assert result.item() == expected
    result = mod_func(pl.Series([left]), pl.Series([right]))
    assert result.item() == expected
    result = mod_func(pa.chunked_array([[left]]), pa.chunked_array([[right]]))
    assert result[0].as_py() == expected
```

maybe we can leave it out for now? pyarrow-backed pandas raises not implemented error for this operation anyway

FBruzzesi:
> are we sure about __mod__? I tried it out, and get some failures (just for pyarrow) with this hypothesis test:

I think it is again related to int divided by int. When casting to float in `pc.divide` I am getting the correct result.

Happy to open another PR for it

MarcoGorelli:
I think it's to do with signs. doing `20 % -42`, pandas Polars and Python all return `-22`, whereas the calculation here returns `20`
```
(Pdb) p result
20.0
(Pdb) p expected
-22
(Pdb) left
20
(Pdb) right
-42
```

MarcoGorelli (approved):
thanks @FBruzzesi , awesome stuff!

Issue #461, opened 2024-07-09
Refactor: factor out PolarsDataFrame


MarcoGorelli:
The current structure is:

```
.
└── narwhals
    ├── _arrow
    ├── _pandas_like
    └── stable
```

and differences between different Polars versions are handled directly in `narwhals.dataframe.py` and `narwhals.series.py`. I think it'd be cleaner if there was an extra `_polars` directory, i.e.:
```
.
└── narwhals
    ├── _arrow
    ├── _pandas_like
    ├── _polars
    └── stable
```
and that differences between Polars versions could all be handled in there

We could probably then remove `_is_polars` and `_backend_version` from `narwhals.DataFrame`

FBruzzesi:
Closed by #561

PR #516, opened 2024-07-13
feat: `collect_schema` method

### LINKED ISSUE #313: [Enh]: DataFrame.collect_schema and LazyFrame.collect_schema

Polars 1.0 is about to come out, and it will have `collect_schema`. We should add it here too

https://docs.pola.rs/api/python/dev/reference/lazyframe/api/polars.LazyFrame.collect_schema.html#polars.LazyFrame.collect_schema

Polars 1.0 compatibility

_No response_

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #313 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Arguably complies with the current pattern, but could get cleaner if #461 finds its way in

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -737,10 +737,7 @@ class DataFrame(BaseFrame[FrameT]):
     @property
     def schema(self) -> Schema:
         r"""
-        Get a dict[column name, DataType].
-
-        Notes:
-            Remark that Polars returns its own `Schema` type instead of a dictionary.
+        Get an ordered mapping of column names to their data type.
 
         Examples:
             >>> import polars as pl
@@ -774,10 +771,7 @@ class DataFrame(BaseFrame[FrameT]):
 
     def collect_schema(self: Self) -> Schema:
         r"""
-        Get a dict[column name, DataType].
-
-        Notes:
-            Remark that Polars returns its own `Schema` type instead of a dictionary.
+        Get an ordered mapping of column names to their data type.
 
         Examples:
             >>> import polars as pl
@@ -2178,10 +2172,7 @@ class LazyFrame(BaseFrame[FrameT]):
     @property
     def schema(self) -> Schema:
         r"""
-        Get a dict[column name, DType].
-
-        Notes:
-            Remark that Polars returns its own `Schema` type instead of a dictionary.
+        Get an ordered mapping of column names to their data type.
 
         Examples:
             >>> import polars as pl
@@ -2201,10 +2192,7 @@ class LazyFrame(BaseFrame[FrameT]):
 
     def collect_schema(self: Self) -> Schema:
         r"""
-        Get a dict[column name, DType].
-
-        Notes:
-            Remark that Polars returns its own `Schema` type instead of a dictionary.
+        Get an ordered mapping of column names to their data type.
 
         Examples:
             >>> import polars as pl
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -96,10 +96,15 @@ class BaseFrame(Generic[FrameT]):
         )
 
     def collect_schema(self) -> Schema:
+        if self._is_polars and self._backend_version < (1, 0, 0):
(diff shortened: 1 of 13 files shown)


narwhals/dataframe.py
```diff
@@ -112,6 +119,9 @@ def drop_nulls(self) -> Self:
 
     @property
     def columns(self) -> list[str]:
+        if self._is_polars and self._backend_version >= (1, 0, 0):
+            return self._compliant_frame.collect_schema().names()  # type: ignore[no-any-return]
```
MarcoGorelli, line 123: not sure about this - like this we'd be hiding that `.columns` isn't a free operation for lazyframes

shouldn't we also define a `Schema` object and return that from `collect_schema`, and let Polars issue its warning for `.columns` / `.schema`?

FBruzzesi: Oh I see your point! Of course that makes total sense - I will give it a try in the next few days 👌

narwhals/schema.py
```diff
@@ -0,0 +1,70 @@
+from __future__ import annotations
```
FBruzzesi, line 1: Gently stolen from polars, let me mention that

FBruzzesi:
@MarcoGorelli Just added the `Schema` class, it's 1 to 1 with [what polars does](https://github.com/pola-rs/polars/blob/main/py-polars/polars/schema.py).

I don't think we need a schema per backend, but happy to hear your thoughts on it

narwhals/dataframe.py
```diff
@@ -720,10 +735,13 @@ def with_row_index(self, name: str = "index") -> Self:
         return super().with_row_index(name)
 
     @property
-    def schema(self) -> dict[str, DType]:
+    def schema(self) -> Schema:
         r"""
         Get a dict[column name, DataType].
 
+        Notes:
+            Remark that Polars returns its own `Schema` type instead of a dictionary.
```
MarcoGorelli, line 743: is this still relevant?

narwhals/dataframe.py
```diff
@@ -745,16 +763,52 @@ def schema(self) -> dict[str, DType]:
             You can pass either pandas or Polars to `func`:
 
             >>> df_pd_schema = func(df_pd)
-            >>> df_pd_schema
-            {'foo': Int64, 'bar': Float64, 'ham': String}
+            >>> df_pd_schema  # doctest:+SKIP
+            Schema({'foo': Int64, 'bar': Float64, 'ham', String})
```
MarcoGorelli, line 767: it could be argued that this is a breaking change...although `Schema` is a subclass of `OrderedDict`, and seems to be behaved the same way

I don't think there's any code written currently which would break under this change? If so, then I think this fits in with the stable API spirit, and we can do it

narwhals/dataframe.py
FBruzzesi: Whops!

narwhals/dataframe.py
FBruzzesi: > it could be argued that this is a breaking change...although Schema is a subclass of OrderedDict, and seems to be behaved the same way

`Schema` adds methods to an `OrderedDict`, I think we are fairly fine

> I don't think there's any code written currently which would break under this change? If so, then I think this fits in with the stable API spirit, and we can do it

So... I was wondering how the stable api would work. We can _only_ add, but never change or delete?!

MarcoGorelli: we need to make sure that any code written with `v1` doesn't break - so, we can remove from `narwhals`, but keep it around in `v1`

I think this case is fine anyway, as ordereddict is still a dict 😎

MarcoGorelli (approved):
amazing, thanks @FBruzzesi !

PR #520, opened 2024-07-14
feat: pyarrow `item`, `is_empty`, `rename`, `write_parquet`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Can this be considered addiction 😂🙈?


narwhals/_arrow/dataframe.py
```diff
@@ -324,3 +329,55 @@ def collect(self) -> ArrowDataFrame:
...
+        df = self._native_dataframe
+
+        num_rows = df.num_rows
+        columns = df.column_names
+
+        for row in range(num_rows):
+            if not named:
+                yield tuple(df[_col][row].as_py() for _col in columns)
+            else:
+                yield {_col: df[_col][row].as_py() for _col in columns}
```
FBruzzesi, line 374: that's a lot of `.as_py` calls, should we be concern about performance here?

narwhals/dependencies.py
```diff
@@ -51,6 +51,15 @@ def get_pyarrow_compute() -> Any:  # pragma: no cover
     return None
 
 
+def get_pyarrow_parquet() -> Any:  # pragma: no cover
+    """Get pyarrow.parquet module (if pyarrow has already been imported - else return None)."""
+    if "pyarrow" in sys.modules:
+        import pyarrow.parquet as pp
+
+        return pp
+    return None
```
FBruzzesi, line 60: Not really necessary... thoughts?

narwhals/_arrow/dataframe.py
MarcoGorelli: hmm maybe...when I asked about iter_rows vs rows to polars devs, they said that doing it in buffers made a massive difference

if there's not a way to do that in pyarrow, maybe we can leave this one out for now?

narwhals/dependencies.py
MarcoGorelli: it's fine 😄

narwhals/_arrow/dataframe.py
FBruzzesi: I was not able to find any reference on how to get an entire row performantly, especially as a python object.

> we can leave this one out for now?

What about adding a performance warning? After all, one should know that arrow has columnar data format, and we are asking for row iteration 🙈

FBruzzesi: Ok I probably need to investigate further then. Let me leave `iter_rows` out of this PR

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #524, opened 2024-07-14
feat: pyarrow `lit`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372

## Checklist

- [x] Tests added 
- [x] Documented the changes

Assumes that `lit` takes a scalar value

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -7,6 +7,7 @@ from typing import Iterable
 
 from narwhals import dtypes
 from narwhals._arrow.expr import ArrowExpr
+from narwhals._arrow.series import ArrowSeries
 from narwhals._expression_parsing import parse_into_exprs
 from narwhals.dependencies import get_pyarrow
 from narwhals.utils import flatten
@@ -15,8 +16,6 @@ if TYPE_CHECKING:
     from typing import Callable
 
     from narwhals._arrow.dataframe import ArrowDataFrame
-    from narwhals._arrow.expr import ArrowExpr
-    from narwhals._arrow.series import ArrowSeries
     from narwhals._arrow.typing import IntoArrowExpr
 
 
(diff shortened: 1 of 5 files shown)


narwhals/_arrow/namespace.py
```diff
@@ -126,6 +125,26 @@ def all(self) -> ArrowExpr:
             backend_version=self._backend_version,
         )
 
+    def lit(self, value: Any, dtype: dtypes.DType | None) -> ArrowExpr:
+        def _lit_arrow_series(df: ArrowDataFrame) -> ArrowSeries:
+            arrow_series = ArrowSeries._from_iterable(
+                data=[value] * len(df),
```
FBruzzesi, line 131: It assumes `value` to be a scalar

narwhals/_arrow/utils.py
```diff
@@ -123,7 +123,7 @@ def validate_column_comparand(other: Any) -> Any:
     return other
 
 
-def validate_dataframe_comparand(other: Any) -> Any:
+def validate_dataframe_comparand(length: int, other: Any) -> Any:
```
FBruzzesi, line 126: pandas has its own differences anyway, so I felt like it's ok to be different 😂

narwhals/_arrow/utils.py
```diff
@@ -136,9 +136,8 @@ def validate_dataframe_comparand(other: Any) -> Any:
         return NotImplemented
     if isinstance(other, ArrowSeries):
         if len(other) == 1:
-            # broadcast
-            msg = "not implemented yet"  # pragma: no cover
-            raise NotImplementedError(msg)
+            pa = get_pyarrow()
+            return pa.chunked_array([[other.item()] * length])
```
MarcoGorelli, line 140: does it work to just return `other.item()`?

FBruzzesi: Since it has len 1 and Series implements `.item` yes 🙌🏼 but I feel like I didn't get the question?!

MarcoGorelli: as in, it should be more efficient to just return `other.item()` and let pyarrow handle the broadcasting, rather than allocating a new chunkedarray here?

FBruzzesi: >  I feel like I didn't get the question

Yep I knew there was something 🙈

> let pyarrow handle the broadcasting

From what I tried this does not work, here what I tried and the respective result

- `return other.item()` -> `TypeError: '<type>' object is not iterable`
- `return [other.item()]` -> `Tpyarrow.lib.ArrowInvalid: Column 3 named lit expected length X but got length 1`
- `return pa.chunked_array([[other.item()]])` -> `Tpyarrow.lib.ArrowInvalid: Column 3 named lit expected length X but got length 1`

MarcoGorelli: ah right, thanks! if we ever support something like `df > scalar` then we may want to reconsider and return `other.item()` here and do the broadcasting somewhere else, but as we're only using `validate_dataframe_comparand` for `with_columns`, I think this is ok 👍

MarcoGorelli (approved):
thanks @FBruzzesi !

love this addiction 😄 I think we're not too far off being able to execute the first few tpch with pyarrow table

PR #533, opened 2024-07-15
ci: add downstream tests for altair + scikit-lego

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


.github/workflows/downstream_tests.yml
```diff
@@ -0,0 +1,46 @@
+name: ci
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
FBruzzesi, line 6: Maybe instead of every PR we can make it both as workflow dispatch (i.e. manual trigger) and dependency of publish to pypi?
I am not sure if you intend to add scikit-lego as well, but test there typically take between 2 and 5 mins

MarcoGorelli: yeah that might be better

FBruzzesi: > I am not sure if you intend to add scikit-lego as well

I will take it as a _yes_ 😂
image

MarcoGorelli: altair actually takes less than our own tests 😄 I think we might as well keep that on in CI

and for scikit-lego, perhaps some tests these could be marked "slow", and we just skip those here?

FBruzzesi: Agree! I think it is enough to run everything without extra dependencies as in [our scheduled ci](https://github.com/koaning/scikit-lego/blob/main/.github/workflows/schedule-dependencies.yml#L30C12-L30C104)

TL;DR

```
run: pytest -n auto --disable-warnings --cov=sklego -m "not cvxpy and not formulaic and not umap"
```

MarcoGorelli:
thanks @FBruzzesi ! this adds a 2-minute job for altair, and a 3-minute one for scikit-lego. for now, I'd suggest we play it safe and just run them, they're not too onerous

PR #537, opened 2024-07-16
Implemented round function for pyarrow

brentomagic:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -163,6 +163,10 @@ class ArrowSeries:
         pc = get_pyarrow_compute()
         return pc.sum(self._native_series)  # type: ignore[no-any-return]
 
+    def round(self) -> int:
+        pc = get_pyarrow_compute()
+        return pc.round(self._native_series)  # type: ignore[no-any-return]
+
     def std(self, ddof: int = 1) -> int:
         pc = get_pyarrow_compute()
         return pc.stddev(self._native_series, ddof=ddof)  # type: ignore[no-any-return]
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -141,6 +141,9 @@ class ArrowExpr:
     def count(self) -> Self:
         return reuse_series_implementation(self, "count", returns_scalar=True)
 
+    def round(self) -> Self:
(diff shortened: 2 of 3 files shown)


MarcoGorelli:
thanks @brentomagic - could you remove the `xfail` from `tests/expr/round_test` to check it passes with pyarrow?

tests/expr/round_test.py
```diff
@@ -11,7 +11,7 @@
 @pytest.mark.parametrize("decimals", [0, 1, 2])
 def test_round(request: Any, constructor: Any, decimals: int) -> None:
     if "pyarrow_table" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
+        request.applymarker(pytest)
```
MarcoGorelli, line 14: you can remove these lines completely 😉

narwhals/_arrow/series.py
```diff
@@ -163,6 +163,10 @@ def sum(self) -> int:
         pc = get_pyarrow_compute()
         return pc.sum(self._native_series)  # type: ignore[no-any-return]
 
+    def round(self) -> int:
+        pc = get_pyarrow_compute()
+        return pc.round(self._native_series)  # type: ignore[no-any-return]
```
FBruzzesi, line 168: `round` method should take number of digits as well

```suggestion
    def round(self, decimals: int) -> int:
        pc = get_pyarrow_compute()
        return pc.round(self._native_series, n_digits=decimals)  # type: ignore[no-any-return]
```

FBruzzesi (commented):
I left a comment seems it looks like a critical piece is missing in the function definition.

This should be adjusted in `ArrowExpr` as well.

Also could you take a look at the [discussion on how to round and differences between polars and pandas](https://github.com/narwhals-dev/narwhals/pull/306/files#r1641848166).

[Pyarrow round](https://arrow.apache.org/docs/python/generated/pyarrow.compute.round.html) allows for different rounding modes, so potentially there is one matching polars API out of the box

PR #546, opened 2024-07-17
feat: add `DataFrame.pivot` for pandas like and Polars backend

### LINKED ISSUE #392: [Enh]: Support for `DataFrame.pivot`

_No response_

_pivot_-al feature to add!

_No response_

_No response_

Challenges:
- Regarding StableAPI, polars pivot would bring the first challenge in supporting two different APIs [v1.0](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html) vs [v0.20](https://docs.pola.rs/api/python/version/0.20/reference/dataframe/api/polars.DataFrame.pivot.html) 
- I am not sure that pandas-like `pivot_table` does support all the `aggregate_function`s out of the box

FBruzzesi:
closes #392 

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #392 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Opening to start getting some feedbacks. Implementation is for pandas-like only for now. Arrow may work partially (need to check how pandas arrow does it I guess)


narwhals/dataframe.py
```diff
@@ -1968,6 +1968,92 @@ def clone(self) -> Self:
...
+    def pivot(
+        self: Self,
+        on: str | list[str] | None,
+        *,
+        index: str | list[str] | None = None,
+        values: str | list[str] | None = None,
+        aggregate_function: Literal[
+            "min", "max", "first", "last", "sum", "mean", "median", "len"
+        ]
+        | None = None,
```
FBruzzesi, line 1980: For now these are supported 🙈

FBruzzesi:
@MarcoGorelli I am considering closing this one for now.
 
There a couple of things that strongly bug me:

1. Multiple columns in polars have a very custom layout, hard to reproduce in practice from pandas
    ```py
    import polars as pl

    data = {
        "ix": [1, 2, 1, 1, 2, 2],
        "col": ["b", "b", "a", "a", "a", "a"],
        "col_b": ["x", "y", "x", "y", "x", "y"],
        "foo": [7, 1, 0, 1, 2, 2],
        "bar": [9, 4, 0, 2, 0, 0],
    }

    df= pl.DataFrame(data)
    df_pl.pivot(on=["col", "col_b"], index="ix", values=["foo", "bar"], aggregate_function="len").columns
    ```
    > ['ix', 'foo_{"b","x"}', 'foo_{"b","y"}', 'foo_{"a","x"}', 'foo_{"a","y"}', 'bar_{"b","x"}', 'bar_{"b","y"}', 'bar_{"a","x"}', 'bar_{"a","y"}']
2. The other issue is that polars behavior changed a lot both from 0.20 and 1.x, as well as from minor 0.20.x versions.

While for point 1 it would be ok to just say we have partial support and so on, for point 2 something like #461 feels like a stronger requirement. For now I implemented the logic in narhwals DataFrame just to make an example of how it would work.

WDYT?

narwhals/dataframe.py
```diff
@@ -1968,6 +1968,115 @@ def clone(self) -> Self:
...
+
+            result = self._compliant_frame.pivot(columns=on, **kwargs)
+            cols = result.columns
+
+            if isinstance(on, str):
+                on = [on]
+
+            combined_on = r"|".join(on)
+            result = result.rename(
+                {c: re.sub(rf"_{combined_on}_", "_", c, count=1) for c in cols}
```
FBruzzesi, line 2072: TODO: "_" should be `separator`

FBruzzesi:
Having some troubles with regression on polars pre 1.0 on output column names 🙈

FBruzzesi:
I am closing this for now.

Supporting polars<1.0 is quite hard as the API changed so significantly that mapping from one to another for all the edge cases is quite too complex.

A doable proposal could be to only allow for `on` and `index` to not take multiple values.

Additional, modin is now failing while it wasn't before 🥲

MarcoGorelli:
I've made some updates, it's green now 🥦 Do the changes look good to you?

FBruzzesi:
Thanks @MarcoGorelli , I think it is good enough for now. I can take a look for how to do it in pyarrow. I have a script to do it _fairly_ manually somewhere 😂

MarcoGorelli:
thanks for your effort here, this is great!

PR #551, opened 2024-07-18
feat: arrow string namespace(s)

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372 (but hidden)

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -361,16 +361,50 @@ class ArrowSeriesCatNamespace:
 
 
 class ArrowSeriesStringNamespace:
-    def __init__(self, series: ArrowSeries) -> None:
+    def __init__(self: Self, series: ArrowSeries) -> None:
         self._arrow_series = series
 
-    def to_uppercase(self) -> ArrowSeries:
+    def starts_with(self: Self, prefix: str) -> ArrowSeries:
+        pc = get_pyarrow_compute()
+        return self._arrow_series._from_native_series(
+            pc.equal(self.slice(0, len(prefix))._native_series, prefix)
+        )
+
+    def ends_with(self: Self, suffix: str) -> ArrowSeries:
+        pc = get_pyarrow_compute()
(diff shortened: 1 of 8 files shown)


narwhals/_arrow/series.py
```diff
@@ -361,16 +361,50 @@ def get_categories(self) -> ArrowSeries:
...
+        return self._arrow_series._from_native_series(
+            pc.utf8_slice_codeunits(
+                self._arrow_series._native_series, start=offset, stop=stop
+            ),
+        )
+
+    def to_datetime(self: Self, format: str | None = None) -> ArrowSeries:  # noqa: A002
+        pc = get_pyarrow_compute()
+        return self._arrow_series._from_native_series(
+            pc.strptime(self._arrow_series._native_series, format=format, unit="ns")
```
FBruzzesi, line 398: Regarding `unit` argument: it is required by [`pc.strptime`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.strptime.html), lowest granularity should be safe?!

MarcoGorelli: Polars defaults to `'us'`, shall we follow that?

MarcoGorelli, line 398:
```suggestion
            pc.strptime(self._arrow_series._native_series, format=format, unit="us")
```

MarcoGorelli (approved):
thanks @FBruzzesi

narwhals/_arrow/series.py
FBruzzesi: I just noticed we don't support explicit time unit and other arguments because of pandas.
But yes let's default to the same time unit as polars until we will be able to explicitly pass it as an argument

PR #576, opened 2024-07-22
feat: add arrow value_counts 

anopsy:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues #

- Related issue #372 
- Closes #

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -340,6 +340,28 @@ class ArrowSeries:
             return self._native_series[0].as_py()
         return self._native_series[index].as_py()
 
+    def value_counts(self: Self, *, sort: bool = False, parallel: bool = False) -> Any:  # noqa: ARG002
+        """Parallel is unused, exists for compatibility"""
+        from narwhals._arrow.dataframe import ArrowDataFrame
+
+        pc = get_pyarrow_compute()
+        pa = get_pyarrow()
+
+        name_ = (
+            "index" if self._native_series._name is None else self._native_series._name
+        )
+
+        val_count = pc.value_counts(self._native_series)
+        val_count = pa.Table.from_struct_array(val_count).rename_columns([name_, "count"])
+
+        if sort:
+            val_count = val_count.sort_by([("count", "descending")])
+
+        return ArrowDataFrame(
+            val_count,
+            backend_version=self._backend_version,
+        )
+
     def zip_with(self: Self, mask: Self, other: Self) -> Self:
         pc = get_pyarrow_compute()
 
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -352,7 +352,9 @@ class ArrowSeries:
         )
 
         val_count = pc.value_counts(self._native_series)
-        val_count = pa.Table.from_struct_array(val_count).rename_columns([name_, "count"])
+        val_count = pa.Table.from_arrays(
+            [val_count.field("values"), val_count.field("counts")], names=[name_, "count"]
+        )
 
         if sort:
             val_count = val_count.sort_by([("count", "descending")])
--- a/tests/series_only/value_counts_test.py
+++ b/tests/series_only/value_counts_test.py
@@ -10,7 +10,7 @@ data = {"a": [4, 4, 6, 4, 1, 1]}
 
 
 def test_value_counts(request: Any, constructor: Any) -> None:
-    if "pandas_series_nullable" in str(constructor) or sys.version_info < (
+    if "pandas_series_nullable" in str(constructor) and sys.version_info < (
         3,
         9,
     ):  # fails for py3.8
--- a/tests/series_only/value_counts_test.py
+++ b/tests/series_only/value_counts_test.py
@@ -1,31 +1,26 @@
 import sys
 from typing import Any
 
-import numpy as np
 import pytest
 
 import narwhals.stable.v1 as nw
+from tests.utils import compare_dicts
 
-data = {"a": [4, 4, 6]}
+data = {"a": [4, 4, 6, 4, 1, 1]}
 
 
 def test_value_counts(request: Any, constructor: Any) -> None:
-    if "pandas_series_nullable" in str(constructor) and sys.version_info < (
+    if "pandas_series_nullable" in str(constructor) or sys.version_info < (
         3,
         9,
     ):  # fails for py3.8
         request.applymarker(pytest.mark.xfail)
 
-    s = nw.from_native(constructor(data), eager_only=True)["a"]
+    series = nw.from_native(constructor(data), eager_only=True)["a"]
 
-    sorted_result = s.value_counts(sort=True)
-    assert sorted_result.columns == ["a", "count"]
+    sorted_result = series.value_counts(sort=True)
+    expected = {"a": [4, 1, 6], "count": [3, 2, 1]}
+    compare_dicts(sorted_result, expected)
 
-    expected = np.array([[4, 2], [6, 1]])
-    assert (sorted_result.to_numpy() == expected).all()
-
-    unsorted_result = s.value_counts(sort=False)
-    assert unsorted_result.columns == ["a", "count"]
-
-    a = unsorted_result.to_numpy()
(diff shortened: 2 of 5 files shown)


narwhals/_arrow/series.py
anopsy, line 1: I have a question here, I sort by count, do I see and understand correctly that the _pandas_like value_counts sorts by index/value?

FBruzzesi: Whops! I believe that is a mistake!

anopsy: in arrow or in pandas? according to the polars/pandas docs though it should sort by frequency so "count"?

anopsy:
I checked the failing checks and I see in some of them ```FAILED tests/series_only/value_counts_test.py::test_value_counts[pyarrow_table_constructor] - AttributeError: type object 'pyarrow.lib.Table' has no attribute 'from_struct_array'```
gives an error, while normal test don't raise this error, really curious if I could do anything to fix it

narwhals/_arrow/series.py
FBruzzesi: Sorry, yes I meant we are wrong in sorting by index. Polars sorts by count, so we should do the same

anopsy: Should I make a PR? and maybe we could extend the tests to cover for it as well? I can do both tomorrow/Wednesday

tests/series_only/value_counts_test.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+
+import narwhals.stable.v1 as nw
+
+data = {"a": [4, 4, 6]}
+
+
+def test_value_counts(request: Any, constructor: Any) -> None:
+    if "pandas_series_nullable" in str(constructor):  # fails for py3.8
+        request.applymarker(pytest.mark.xfail)
```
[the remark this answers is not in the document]
MarcoGorelli: I think we may have inadvertently been skipping the rest of the test because of the `pytest.skip` 🙈  taking a look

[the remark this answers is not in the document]
anopsy: image
I thought that a dot means a passed test, and "s" a skipped one. But it makes sense Anyway I'll take care of sys.version_info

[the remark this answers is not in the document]
MarcoGorelli: is this on your branch? I think you're doing it correctly, but on `main`, maybe not?

tests/series_only/value_counts_test.py
```diff
@@ -0,0 +1,27 @@
...
+    assert sorted_result.columns == ["a", "count"]
+
+    expected = np.array([[4, 2], [6, 1]])
+    assert (sorted_result.to_numpy() == expected).all()
+
+    unsorted_result = s.value_counts(sort=False)
+    assert unsorted_result.columns == ["a", "count"]
+
+    a = unsorted_result.to_numpy()
+    assert (a[a[:, 0].argsort()] == expected).all()
```
MarcoGorelli, line 27: I think this is an old test, seeing as we're updating things, can we write it using `compare_dict`?

So
```python
    sorted_result = series.value_counts(sort=True)
    expected = {'b': [4, 6], 'count': [2, 1]}
    compare_dicts(sorted_result, expected)

    unsorted_result = series.value_counts(sort=False).sort('count', descending=True)
    compare_dicts(unsorted_result, expected)
```

for `sort=False`, the output is meant to be random, so I'm not sure we can make a fully deterministic test case - but we can at least check that if we then sort outselves, the result is correct

narwhals/_arrow/series.py
MarcoGorelli: cool, thanks @anopsy ! well done going so much into depth and spotting existing issues 🥳

tests/series_only/value_counts_test.py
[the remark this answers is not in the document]
anopsy: I'm not sure if I understand correctly. But yes I did run the test on my branch with changes(arrow-valcounts)
image
when I switch to main I cannot run the test, because it doesn't exist

tests/series_only/value_counts_test.py
anopsy: yass! maybe I'll also extend the test a bit by adding more values, so it also checks the sort arg properly

tests/series_only/value_counts_test.py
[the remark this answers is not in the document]
MarcoGorelli: sorry I meant, the corresponding test:
```
pytest tests/series_only/test_common.py -k value_counts
```
still, looks like only one test gets skipped, so we're all good 😇

tests/series_only/value_counts_test.py
anopsy: should it be  if "pandas_series_nullable" in str(constructor) **or** sys.version_info < (3,9 ):  OR **and?**

tests/series_only/value_counts_test.py
```diff
@@ -0,0 +1,26 @@
...
+import pytest
+
+import narwhals.stable.v1 as nw
+from tests.utils import compare_dicts
+
+data = {"a": [4, 4, 6, 4, 1, 1]}
+
+
+def test_value_counts(request: Any, constructor: Any) -> None:
+    if "pandas_series_nullable" in str(constructor) or sys.version_info < (
```
MarcoGorelli, line 13: I think this should be an `and`? It's a bit in an old version of pandas for pandas-nullable series, but the rest of the libraries should run fine on Python 3.8

narwhals/_arrow/series.py
```diff
@@ -340,6 +340,28 @@ def item(self: Self, index: int | None = None) -> Any:
...
+
+        pc = get_pyarrow_compute()
+        pa = get_pyarrow()
+
+        name_ = (
+            "index" if self._native_series._name is None else self._native_series._name
+        )
+
+        val_count = pc.value_counts(self._native_series)
+        val_count = pa.Table.from_struct_array(val_count).rename_columns([name_, "count"])
```
MarcoGorelli, line 355: looks like old versions of pyarrow don't have `from_struct_array`, seems it was only added in PyArrow 15.0.0 https://arrow.apache.org/release/15.0.0.html

maybe here we need to do a check on `self._backend_version` and raise `NotImplementedError` if `self._backend_version` is older than 15.0.0?

anopsy: @FBruzzesi  also suggested we could unpack the struct manually
```
value_counts = pc.value_counts(series)

pa.Table.from_arrays([
    value_counts.field("values"),
    value_counts.field("counts")
    ], names=["name_", "count])
```

WDYT?

MarcoGorelli: ooh that'd be better if possible, nice!

anopsy: okay, now I realised that I use ```
    if **"pandas_series_nullable"** in str(constructor) and sys.version_info < (
        3,
        9,
    ):  # fails for py3.8
```
while using a dataframe constructor, could this be the reason for failing 3.8v checks?
If that's the case should I change it to "pandas_nullable_constructor"?

FBruzzesi: Yes! It is not marking anything to fail, thus pytest does not expect any test to be failing!

> If that's the case should I change it to "pandas_nullable_constructor"?

More yes!

anopsy: all green 🥦🥦🥦🥦🥦

MarcoGorelli: 😆 best "all green" emoji ever

MarcoGorelli (approved):
absolute legend @anopsy , and thanks @FBruzzesi for reviewing

@anopsy I think you've got a talent for finding (and then solving 🙌 ) issues which end up being more complicated than they initially look 😄

anopsy:
> you've got a talent for finding (and then solving 🙌 )
and also for creating new issues 😅 
Thanks for support

Issue #47, opened 2024-04-11
add when-then-otherwise


MarcoGorelli:
https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.when.html

FBruzzesi:
Should we close this in favor of #668?

MarcoGorelli:
yup, thanks!

PR #588, opened 2024-07-23
feat: add when-then-otherwise expression

aivanoved:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #47 

## Checklist

- [x] Tests added 
- [x] Documented the changes

This PR so far adds a simple `when-then-otherwise` expression with no chaining

~The PR does not aim to be complete as of now, the current aim is to get an initial review as to if this approach is consistent with the project's quality and style of code~

TODO:
- [x] Allow `Iterable[IntoPandasExpr]` to the pandas API for `predicates`
- [x] ~Allow `when-then-otherwise` chaining~ obsolete: see comments below and #668
- [x] Address noqa ignores

~Note: Temporary converted back to draft due to a git rebase issue~ fixed

Note: due to an unfortunate rebase gone wrong, the commit history is a mess, we could open a new pr from a new branch or when merging do a squash merge. Further advice on this would be much appreciated.


narwhals/expr.py
```diff
@@ -3605,6 +3606,96 @@ def sum_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
...
+        >>> df_pd = pd.DataFrame({"a": [1, 2, 3], "b": [5, 10, 15]})
+
+        We define a dataframe-agnostic function:
+
+        >>> @nw.narwhalify
+        ... def func(df_any):
+        ...     from narwhals.expr import when
+        ...
+        ...     return df_any.with_columns(
+        ...         when(nw.col("a") < 3).then(5).otherwise(6).alias("a_when")
```
MarcoGorelli, line 3674: I think we should export it from the top-level, so then here it becomes `nw.when` (just like in Polars you'd do `pl.when`)

MarcoGorelli (commented):
Amazing effort here, well done @aivanoved ! I'm seriously impressed that you picked up what's probably the hardest issue on the issue tracker and worked through it all

I've left a comment - also, if you run `pytest tests --cov=narwhals`, it'll show you which paths aren't yet covered by tests

Overall, this looks very positive!

narwhals/_pandas_like/namespace.py
```diff
@@ -335,10 +336,117 @@ def __init__(
...
+        self._backend_version = backend_version
+        self._condition = condition
+        self._above_when = above_when
+        self._then_value = then_value
+        self._otherwise_value = otherise_value
+
+        # TODO @aivanoved: this is way slow as during computation time this takes
+        # quadratic time need to improve this to linear time
+        self._condition = self._condition & (~self._above_when._already_set)  # type: ignore[has-type]
+        self._already_set = self._above_when._already_set | self._condition  # type: ignore[has-type]
```
MarcoGorelli, line 379: not sure I follow - rather than aiming for a complete solution from the start with a todo, could we start with the simplest possible case? this would already be enough to unblock running the rest of the tpc-h queries

i.e. to support `nw.when(...).then(...)` and `nw.when(...).then(...).otherwise(...)`, without chaining multiple of these together

it would also be easier to review like this, and we can then leave the chaining part for a follow-up

MarcoGorelli (commented):
thanks for updating! 🙏

narwhals/_pandas_like/namespace.py
aivanoved: @MarcoGorelli

this (commit)[https://github.com/narwhals-dev/narwhals/pull/588/commits/2f03bd0dbb4979bbec1e24ef8c4fcc504752e778] already implements the simple `nw.when(...).then(...)` and `nw.when(...).then(...).otherwise(...)`

And as of now other than the todo, we have the chaining as well

I wouldn't mind to revert to the above mentioned commit if we want to keep thing simple

narwhals/expr.py
aivanoved: I have added it to the top level api, however is not yet stabalised, so there are some errors in the test suit

narwhals/_pandas_like/namespace.py
aivanoved: I can update the comment to be more detailed for sure

the reason this introduces a quadratic complexity for this operation is because I assign the expression to all previous masks to the current `ChainedWhen` mask and then call this expression in every `ChainedWhen` going down.

example:

```py
import pandas as pd

import narwhals as nw

data = {
    "a": [1, 2, 3, 4, 5],
}
df = nw.from_native(pd.DataFrame(data))

where_then_1 = nw.when(nw.col("a") == 1).then(7)
where_then_2 = where_then_1.when(nw.when(nw.col("a") == 3)).then(9).otherwise(19)

df.with_columns(where_then_2.alias("a_chained_when"))
```

When computing the mask for `when_then_2` it holds the expression for the mask of itself and `when_then_1`, `when_then_1` holds its own mask

When computing we call to see what the result of `when_then_1` is so we compute its mask with one call, cost of `1` for masks
When computing the result for `when_then_2`, we compute its condition's mask and the collective mask of all previous calls, which in this case is just the mask of `when_then_1`, so a cost of `1+1=2` for masks

The total cost for two chained calls is `3 = 2*3/2`

One could follow a similar logic to get that for `n`-chained calls we would compute masks a total of `n*(n+1)/2` times

MarcoGorelli: cool, thanks for explaining! sure, if you don't mind reverting and then keeping the additional complexity to a follow-up, I think we can probably merge this quite quickly

aivanoved:
@MarcoGorelli can you take another look at this?

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +262,97 @@ def concat(
...
+        plx = self.__class__(self._implementation, self._backend_version)
+        if predicates:
+            condition = plx.all_horizontal(*flatten(predicates))
+        elif constraints:
+            condition = plx.all_horizontal(
+                *flatten([plx.col(key) == value for key, value in constraints.items()])
+            )
+        else:
+            msg = "at least one predicate or constraint must be provided"
+            raise TypeError(msg)
```
MarcoGorelli, line 280: what if both `predicates` and `constraints` are passed?

I'd suggest we leave `constraints` out from the initial implementation, we don't support it in `filter` either

MarcoGorelli (commented):
thanks @aivanoved !

one comment on `constraints`, but I think it adds unnecessary complexity so the simplest fix is probably to just not support it for now

narwhals/_pandas_like/namespace.py
aivanoved: Sounds good, makes things much simpler

I believe in polars as well passing both is disallowed, can check as am not sure

aivanoved: @MarcoGorelli

It should be removed now

aivanoved:
@MarcoGorelli 

Note that I have added support for users to be able to provide `Series` or `Expr` as `then` and `otherwise` values, not just scalars, which I think is a crucial behaviour

Currently not all code paths are tested as according to the coverage test, am looking into this

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +251,130 @@ def concat(
...
+    np = get_numpy()
+
+    if not np:
+        msg = "numpy is required for this function"
+        raise ImportError(msg)
+    if isinstance(value, PandasLikeExpr):
+        return value
+    elif isinstance(value, PandasLikeSeries):
+        return plx._create_expr_from_series(value)
+    elif isinstance(value, np.ndarray):
```
aivanoved, line 287: Interestingly enough `polars` considers python arrays as scalars for `when-then-otherwise`, but numpy arrays and series as iterables to zip with

aivanoved: Also note that the pandas code does not guarantee type safety

@aivanoved TODO for me

MarcoGorelli:
legend, thanks!

cool, once code coverage is complete we can ship it 🚀

narwhals/_pandas_like/namespace.py
aivanoved: This is out of the scope of this pr as seems to have to do with `PandasLikeSeries.zip_with`

aivanoved:
@MarcoGorelli

can you take another look at this?

MarcoGorelli:
thanks @aivanoved ! 

looks like there's a warning in the docs build
```
WARNING -  griffe: narwhals/expr.py:3676: Failed to get 'name: description' pair from 'predicates'
```
and test failures on python 3.8
```
FAILED tests/expr_and_series/when_test.py::test_no_arg_when_fail[polars_eager_constructor] - ValueError: No predicates or constraints provided to `when`.
FAILED tests/expr_and_series/when_test.py::test_no_arg_when_fail[polars_lazy_constructor] - ValueError: No predicates or constraints provided to `when`.
```

aivanoved:
The doc test I will address this

However, for the python 3.8 issue - the issue comes from that starting from `polars` version `0.20.4` onwards the error thrown when there are no `predicates` or `constraints` is a `TypeError`, for which we test
For versions below this, meaning `polars<=0.20.3`, the error thrown is a `ValueError`, in the ci the installed version of `polars` is `0.20.3`

I can change the error type depending on the installed `polars` version, however this seems unsatisfactory, please advise @MarcoGorelli

For reference [here](https://github.com/pola-rs/polars/blob/58b0da8bc31d71d21be8782c5013747516652f81/py-polars/polars/utils/_parse_expr_input.py#L148) is the error thrown in version `0.20.3`
And [here](https://github.com/pola-rs/polars/blob/cdef09b98f5164271ac1a489b395accf4a363549/py-polars/polars/utils/_parse_expr_input.py#L184) is the same error in version `0.20.4`

> thanks @aivanoved !
> 
> looks like there's a warning in the docs build
> 
> ```
> WARNING -  griffe: narwhals/expr.py:3676: Failed to get 'name: description' pair from 'predicates'
> ```
> 
> and test failures on python 3.8
> 
> ```
> FAILED tests/expr_and_series/when_test.py::test_no_arg_when_fail[polars_eager_constructor] - ValueError: No predicates or constraints provided to `when`.
> FAILED tests/expr_and_series/when_test.py::test_no_arg_when_fail[polars_lazy_constructor] - ValueError: No predicates or constraints provided to `when`.
> ```

MarcoGorelli:
I think you can do `with pytest.raises((ValueError, TypeError)):`

any reason why you closed the PR?

aivanoved:
No, I guess I misclicked and canceled it by accident

aivanoved:
> I think you can do `with pytest.raises((ValueError, TypeError)):`
> 
> 
> 
> any reason why you closed the PR?

As long as we are fine with having `narwhals` and `polars` throwing different types of error for `polars<=0.20.3` this works, just want to make sure we are happy with  the this decision

aivanoved:
@MarcoGorelli

Everything should be working now I think

aivanoved:
gentle ping @MarcoGorelli

MarcoGorelli:
thanks for updating - hoping to merge this soon

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +251,118 @@ def concat(
...
```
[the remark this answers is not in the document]
MarcoGorelli: indeed, if `value` is a numpy array, it implies that numpy was already imported

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +250,116 @@ def concat(
...
+    from narwhals.dependencies import get_numpy
+
+    # NumPy is a required dependency of pandas
+    np = get_numpy()
+
+    if isinstance(value, PandasLikeExpr):
+        return value
+    elif isinstance(value, PandasLikeSeries):
+        return plx._create_expr_from_series(value)
+    elif (np := get_numpy()) is not None and isinstance(value, np.ndarray):
```
FBruzzesi, line 283: As numpy is already available from line 277, there is no need for it here:

```suggestion
    elif isinstance(value, np.ndarray):
```

FBruzzesi (commented):
I left a few comments on the test cases, but overall I am incredibly happy with this PR, and could not spot anything wrong 🙌🏼 I hope to merge it soon 🚀

On a sidenote, I wonder if we should factor out the classes in a different module as polars does, namely:

```terminal
narwhals/_<backend>/whenthen.py
narwhals/whenthen.py
```
but absolutely not a blocker as we will be able to move it in the future anyway without issues

narwhals/expr.py
```diff
@@ -3643,6 +3643,76 @@ def sum_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
...
+    def otherwise(self, value: Any) -> Expr:
+        return Expr(lambda plx: self._call(plx).otherwise(extract_compliant(plx, value)))
+
+
+def when(*predicates: IntoExpr | Iterable[IntoExpr]) -> When:
+    """
+    Start a `when-then-otherwise` expression.
+    Expression similar to an `if-else` statement in Python. Always initiated by a `pl.when(<condition>).then(<value if condition>)`., and optionally followed by chaining one or more `.when(<condition>).then(<value>)` statements.
+    Chained when-then operations should be read as Python `if, elif, ... elif` blocks, not as `if, if, ... if`, i.e. the first condition that evaluates to `True` will be picked.
+    If none of the conditions are `True`, an optional `.otherwise(<value if all statements are false>)` can be appended at the end. If not appended, and none of the conditions are `True`, `None` will be returned.
```
FBruzzesi, line 3674: Suggesting to add a few newlines to get a less compact rendering:

```suggestion
    r"""
    Start a `when-then-otherwise` expression.

    Expression similar to an `if-else` statement in Python. Always initiated by a `pl.when(<condition>).then(<value if condition>)`, and optionally followed by chaining one or more `.when(<condition>).then(<value>)` statements.

    Chained when-then operations should be read as Python `if, elif, ... elif` blocks, not as `if, if, ... if`, i.e. the first condition that evaluates to `True` will be picked.

    If none of the conditions are `True`, an optional `.otherwise(<value if all statements are false>)` can be appended at the end. If not appended, and none of the conditions are `True`, `None` will be returned.
```

Currently:

image

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +250,116 @@ def concat(
...
+
+
+class PandasWhen:
+    def __init__(
+        self,
+        condition: PandasLikeExpr,
+        implementation: Implementation,
+        backend_version: tuple[int, ...],
+        then_value: Any = None,
+        otherise_value: Any = None,
```
FBruzzesi, line 298: A few typos

```suggestion
        otherwise_value: Any = None,
```

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +250,116 @@ def concat(
...
+    ) -> None:
+        self._implementation = implementation
+        self._backend_version = backend_version
+        self._condition = condition
+        self._then_value = then_value
+        self._otherwise_value = otherise_value
```
FBruzzesi, line 304:
```suggestion
        self._otherwise_value = otherwise_value
```

narwhals/_pandas_like/namespace.py
```diff
@@ -250,3 +250,116 @@ def concat(
...
+        condition = self._condition._call(df)[0]
+
+        value_series = _when_then_value_arg_process(
+            plx, self._then_value, shape=condition.shape
+        )._call(df)[0]
+        otherwise_series = _when_then_value_arg_process(
+            plx, self._otherwise_value, shape=condition.shape
+        )._call(df)[0]
+
+        return [value_series.zip_with(condition, otherwise_series)]
```
MarcoGorelli, line 322: this would create an `object` dtype series, which we need to avoid

the dtype of `value_series` needs to be preserved

MarcoGorelli (changes requested):
thanks for your PR, just took a closer look and found a blocker

aivanoved:
I'm currently on vacation will take a closer look into this again, but will be slower to respond, apologies for the delay

narwhals/_pandas_like/namespace.py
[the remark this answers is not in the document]
aivanoved: good point, didn't look into cudf much actually, will have to check it out in more detail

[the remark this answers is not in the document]
aivanoved: I put the error throw path just in case that someone or something somehow messed with the python installation path and `numpy` was unavailable at runtime. This is quite an esoteric scenario but still put the code there just in case. Can delete the failure
path and simplify complexity.

`numpy` is indeed a required dependency of pandas, which js why in the tests I had to mock the `get_numpy` function

If we delete the failure path we can also simplify the test module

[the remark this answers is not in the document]
aivanoved: I actually just saw the changes @MarcoGorelli that you just did what I mentioned in the previous comment

oops on my part for not checking the last commit

narwhals/_pandas_like/namespace.py
aivanoved: I assume we are talking about the `None` possibility for `otherwise_value`?

MarcoGorelli: yup!

MarcoGorelli:
cool, no hurry, enjoy your holiday, thanks @aivanoved !

narwhals/_pandas_like/namespace.py
aivanoved: ok, looking deeper into this

we have found an issue with `zip_with` in the `pandas` implementation as in the `polars` documentation of `zip_with`, [here](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.zip_with.html), `other` series already has to be of the same type of the series it is zipped with. No such guarantees are made in the `pandas` implementation in narwhals, which leads to this issue.

aivanoved: I have added a fix for this in this pr for now, if you want we can move this fix into a separate pr

tests/expr_and_series/when_test.py
```diff
@@ -0,0 +1,127 @@
+from __future__ import annotations
+
+from typing import Any
+
+import pytest
+
+import narwhals.stable.v1 as nw
+from narwhals.stable.v1 import when
```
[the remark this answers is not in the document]
aivanoved: yes, I believe so, because I need to wrap the intermediate classes for `When` and `Then` in a stable vrsion

tests/expr_and_series/when_test.py
```diff
@@ -0,0 +1,127 @@
...
+        "a_when": [3, None, None],
+    }
+    compare_dicts(result, expected)
+
+
+def test_value_expression(request: Any, constructor: Any) -> None:
+    if "pyarrow_table" in str(constructor) or "dask" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
+    df = nw.from_native(constructor(data))
+    result = df.with_columns(when(nw.col("a") == 1).then(nw.col("a") + 9).alias("a_when"))
```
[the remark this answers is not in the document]
aivanoved: added

aivanoved:
Current ci fails with the new `zip_with`, due to a know, issue in `modin`, [here](https://github.com/modin-project/modin/issues/7364)

narwhals/_pandas_like/namespace.py
aivanoved: Also note that polars also does some sneaky casting sometimes, example

```python
import polars as pl

s = pl.Series([1, 2, 3])
df = pl.DataFrame({'s': s})

print(df)
# shape: (3, 1)
# ┌─────┐
# │ s   │
# │ --- │
# │ i64 │
# ╞═════╡
# │ 1   │
# │ 2   │
# │ 3   │
# └─────┘





print(df.select(pl.when(pl.col('s') == 1).then(-1).alias('s_cast')))
# shape: (3, 1)
# ┌────────┐
# │ s_cast │
# │ ---    │
# │ i32    │
# ╞════════╡
# │ -1     │
# │ null   │
# │ null   │
# └────────┘
```

aivanoved: > ok, looking deeper into this

> we have found an issue with zip_with in the pandas implementation as in the polars documentation of zip_with, [here](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.zip_with.html), other series already has to be of the same type of the series it is zipped with. No such guarantees are made in the pandas implementation in narwhals, which leads to this issue.

I have removed this fix now, though the issue still stands, I have replaced it with an expansion on the `maybe_cast_dtypes` functionality

aivanoved: Now the functionality is as follows

```python
import narwhals as nw
import pandas as pd
import polars as pl


@nw.narwhalify
def function(df):
    return df.with_columns(nw.when(nw.col('s') == 1).then(-1).alias('s_when'))

s = [1, 2, 3, 4, 5]
df_pd = pd.DataFrame({'s': s})
df_pl = pl.DataFrame({'s': s})





print(function(df_pd))
# s  s_when
# 0  1      -1
# 1  2    <NA>
# 2  3    <NA>
# 3  4    <NA>
# 4  5    <NA>





print(function(df_pd).dtypes)
# s         int64
# s_when    Int64





print(function(df_pl))
# dtype: object
# shape: (5, 2)
# ┌─────┬────────┐
# │ s   ┆ s_when │
# │ --- ┆ ---    │
# │ i64 ┆ i32    │
# ╞═════╪════════╡
# │ 1   ┆ -1     │
# │ 2   ┆ null   │
# │ 3   ┆ null   │
# │ 4   ┆ null   │
# │ 5   ┆ null   │
# └─────┴────────┘
```

MarcoGorelli: thanks for looking into this so deep!

i'd like to spend a little more time on this - nullable dtypes aren't generally supported by the data ecosystem unfortunately, so i'm not sure we should convert to them on behalf of the user

we'll get this in 💪 let's just take our time to avoid having to make breaking changes later

aivanoved: I had other thoughts on how to solve this for the type issue, which again mean changing the type of the column in pandas

- changing to a `float` type
- use masked arrays for  the`integer` case

MarcoGorelli (approved):
thanks @aivanoved ! I just pushed some slight changes to avoid using nullable dtypes when the user didn't ask for them

EdAbati:
Thank you @aivanoved for working this! This is a great addition

PR #590, opened 2024-07-23
Feat: `Series.shift` Pyarrow Backend Implementation 

mfonekpo:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


narwhals/_arrow/expr.py
```diff
@@ -192,6 +192,9 @@ def sum(self) -> Self:
     def drop_nulls(self) -> Self:
         return reuse_series_implementation(self, "drop_nulls")
 
+    def series_shift(self, n: int) -> Self:
+        return reuse_series_implementation(self, "series_shift", n, returns_scalar=True)
```
MarcoGorelli, line 196: are you sure this returns a scalar?

MarcoGorelli:
thanks @mfonekpo  - could you put a descriptive title please?

mfonekpo:
> thanks @mfonekpo  - could you put a descriptive title please?

Yes Marco. Done.

MarcoGorelli:
thanks @mfonekpo - could you name the test so it ends with `_test.py` please? else it won't run

mfonekpo:
> thanks @mfonekpo - could you name the test so it ends with `_test.py` please? else it won't run

Ohhhh....how did I miss that.

narwhals/_arrow/expr.py
```diff
@@ -194,6 +194,9 @@ def sum(self) -> Self:
     def drop_nulls(self) -> Self:
         return reuse_series_implementation(self, "drop_nulls")
 
+    def series_shift(self, n: int) -> Self:
+        return reuse_series_implementation(self, "series_shift", n)
```
EdAbati, line 198:
```suggestion
    def shift(self, n: int) -> Self:
        return reuse_series_implementation(self, "shift", n)
```

these should have the same name otherwise `narwhals` will not be able to find it

tests/expr_and_series/series_shift_test.py
EdAbati, line 1: not sure if this was already discussed in discord or somewhere else.

Since we have already the test of `shift`, I think we should just use that and removing the `xfail` there

EdAbati (commented):
Thanks for the PR :)

narwhals/_arrow/series.py
```diff
@@ -209,6 +209,10 @@ def drop_nulls(self) -> ArrowSeries:
         pc = get_pyarrow_compute()
         return self._from_native_series(pc.drop_null(self._native_series))
 
+    def shift(self, n: int) -> Self:
+        pc = get_pyarrow_compute()
+        return self._from_native_series(pc.shift(self._native_series, n))
```
FBruzzesi, line 214: This might end up being not so straightforward after all. We should _manually_ slice the original array and build the new one, this will also be dependent on the sign of `n`

mfonekpo:
Alright then.

narwhals/_arrow/series.py
```diff
@@ -209,6 +209,30 @@ def drop_nulls(self) -> ArrowSeries:
...
+                chunk.slice(0, len(chunk) - n) for chunk in chunks
+            ]
+        elif n < 0:
+            # If n is negative, append nulls and truncate the start
+            shifted_chunks = [chunk.slice(-n, len(chunk) + n) for chunk in chunks] + [
+                pa.array([None] * -n, type=dtype)
+            ]
+        else:
+            # If n is 0, return the original series
+            shifted_chunks = chunks
```
MarcoGorelli, line 230: out of interest, where did you get this from?

tests/expr_and_series/round_test.py
```diff
@@ -1,25 +0,0 @@
-from __future__ import annotations
```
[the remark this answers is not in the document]
MarcoGorelli: yup - please don't delete this test @mfonekpo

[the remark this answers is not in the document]
mfonekpo: Hi Marco. I did not touch any file `round.py`.  I did a reset and resolved some conflicts. Must have happened during that. Let me look through the code and revert.

narwhals/_arrow/series.py
mfonekpo: Well, according to FG's last comment, I looked up steps I can use to resolve it. I wrote the code, taking note of n's polarity in his comment. I don't know Marco, how do you suggest I go about it?

MarcoGorelli: cool, thanks - I'm just a bit surprised, as it looks a bit more advanced than other work you've submitted. if you figured this all out, then well done 🙌

mfonekpo:
There is room for growth, right? Thank you. Is it ok? because you said it
is out of interest. What do I do?

MarcoGorelli:
yup, definitely room for growth :) it's OK, if you add the test I suggested and it passes, then it should be fine

mfonekpo:
Thanks, I will!

tests/expr_and_series/shift_test.py
```diff
@@ -28,11 +23,8 @@ def test_shift(request: Any, constructor: Any) -> None:
     compare_dicts(result, expected)
 
 
-def test_shift_series(request: Any, constructor_eager: Any) -> None:
-    if "pyarrow_table" in str(constructor_eager):
-        request.applymarker(pytest.mark.xfail)
-
-    df = nw.from_native(constructor_eager(data), eager_only=True)
+def test_shift_series(constructor: Any) -> None:
+    df = nw.from_native(constructor(data), eager_only=True)
```
MarcoGorelli, line 27: there's been a little renaming going on recently, so we need to change this:
```suggestion
def test_shift_series(constructor_eager: Any) -> None:
    df = nw.from_native(constructor_eager(data), eager_only=True)
```
It needs to be eager because this test checks Series, which is eager-only

utils/check_backend_completeness.py
```diff
@@ -0,0 +1,95 @@
+"""
```
MarcoGorelli, line 1: can you remove this file please?

tests/expr_and_series/shift_test.py
```diff
@@ -13,11 +11,8 @@
 }
 
 
-def test_shift(request: Any, constructor: Any) -> None:
-    if "pyarrow_table" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
-
-    df = nw.from_native(constructor(data))
+def test_shift(constructor_eager: Any) -> None:
+    df = nw.from_native(constructor_eager(data), eager_only=True)
```
MarcoGorelli, line 15: these don't need to change in this test - only in `test_shift_series` did you need to do that, `constructor` is fine here, because we're testing `Expr.shift`
```suggestion
def test_shift(constructor: Any) -> None:
    df = nw.from_native(constructor(data), eager_only=True)
```

tests/expr_and_series/series_shift_test.py
```diff
@@ -0,0 +1,48 @@
...
+    compare_dicts(result_a, expected_a)
+    compare_dicts(result_b, expected_b)
+    compare_dicts(result_c, expected_c)
+    compare_dicts(result_d, expected_d)
+
+
+def test_shift_multi_chunk_pyarrow() -> None:
+    tbl = pa.table({"a": [1, 2, 3]})
+    tbl = pa.concat_tables([tbl, tbl, tbl])
+    tbl = tbl.combine_chunks()
```
MarcoGorelli, line 40: can you remove this line please? the purpose of this test is to check that it works correctly when you have multiple chunks

narwhals/_arrow/series.py
```diff
@@ -221,6 +221,30 @@ def drop_nulls(self) -> ArrowSeries:
...
```
MarcoGorelli, line 242: this is showing up as uncovered, could you add a test which covers it please? i.e. shift by `0`

MarcoGorelli (commented):
thanks for updating 🙏 ! have left a few comments

In general, I'd suggest that before asking for a review, you click on "files changed" and look through your changes yourself - it happens quite often that extra changes end up there which shouldn't be there. This fine is fine whilst learning 🤠 I'm just pointing it out so you know how to improve

tests/expr_and_series/series_shift_test.py
mfonekpo: line 40 correct?

MarcoGorelli: yeah, the `combine_chunks` part

mfonekpo: wow. When I was debugging, I noticed that the `result` returned **[null,1,2,1,2,1,2]** which was kind of weird, and totally different from  `expected`. And combining the chunks happened to pass.

mfonekpo: Does it mean that the method is supposed to affect every chunk individually? and not as a whole?

mfonekpo: Alright...I will take a look

tests/expr_and_series/shift_test.py
MarcoGorelli: could you address this comment please?

MarcoGorelli: This still needs addressing, could you read over the suggested change more carefully and address it please?

MarcoGorelli (approved):
I've added a commit to get this through

A bit of general feedback:
- in tests, you should construct `expected` explicitly, without any logic. The test case I handed you was like this, there was no need to modify it
- the implementation for `shift` wasn't quite right. I'd pointed you to the solution (https://stackoverflow.com/a/78805975/4451315), why not try to just use that?

Anyway, thanks for your work here 💪 let's ship it

mfonekpo:
Thank you Marco, I have taken note of the points you listed. should I
refactor it to meet the standard?

PR #602, opened 2024-07-23
feat: add from_dict

### LINKED ISSUE #572: feat: `nw.from_dict`?
In `formulaic` they do something like this:
- they have some dataframe `self.data`
- they then create a pandas dataframe from a dict

I _think_ we need a way to do the same here. For example:
```python
import numpy as np
import pandas as pd
import narwhals as nw
df = pd.DataFrame({'b': [1, 2]})
nw_df = nw.from_native(df, eager_only=True)

other = nw.from_dict({'a': [1,2,3]}, module=nw.get_native_namespace(nw_df))
```

So like this, you pass `nw_df.__native_namespace__()` as an argument in the `from_dict` call, and Narwhals knows to create a `nw.DataFrame` backed by the right object (here, a pandas dataframe)

I dunno. We don't have the ability to just create dicts from nothing, so we need to _slightly_ depart from the Polars signature, although the top-level Narwhals functions already depart from Polars anyway

anyone got a better API in mind?

---

one alternative which comes to mind:
```
other = nw.__compliant_namespace__().from_dict({'a': [1,2,3]})
```
which doesn't exactly thrill me

MarcoGorelli:
closes #572 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -59,7 +59,7 @@ def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
     Examples:
         >>> import pandas as pd
         >>> import polars as pl
-        >>> import narwhals.stable.v1 as nw
+        >>> import narwhals as nw
         >>> data = {"a": [1, 2, 3], "b": [4, 5, 6]}
 
         Let's define a dataframe-agnostic function:
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -88,8 +88,8 @@ def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
         └─────┴─────┘
     """
     if native_namespace is get_polars():
-        return from_native(native_namespace.from_dict(data), eager_only=True)
-    if (
+        native_frame = native_namespace.from_dict(data)
+    elif (
         native_namespace is get_cudf()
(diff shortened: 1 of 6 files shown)


narwhals/functions.py
```diff
@@ -42,6 +48,61 @@ def concat(
     )
 
 
+def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
```
FBruzzesi, line 51: Shall we add just the schema argument? We should be able to convert it to the native via `reverse_translate_dtype`

MarcoGorelli: yes! not as required though, right?

FBruzzesi: yep, polars default is None, it will infer

MarcoGorelli: If so, OK to leave it as a follow-up? This is all I need for Formulaic at the moment 😄

https://github.com/narwhals-dev/narwhals/issues/603

PR #606, opened 2024-07-23
feat: add `schema` argument to `from_dict` function

### LINKED ISSUE #603: feat: add `schema` argument to `from_dict`
See https://github.com/narwhals-dev/narwhals/pull/602#discussion_r1688654218 for context

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #603 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Code is ugly :|

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -25,6 +25,8 @@ from narwhals.utils import validate_laziness
 FrameT = TypeVar("FrameT", bound=Union[DataFrame, LazyFrame])  # type: ignore[type-arg]
 
 if TYPE_CHECKING:
+    from narwhals.dtypes import DType
+    from narwhals.schema import Schema
     from narwhals.series import Series
 
 
@@ -48,12 +50,18 @@ def concat(
     )
 
 
-def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
+def from_dict(
+    data: dict[str, Any],
+    schema: dict[str, DType] | Schema | None = None,
+    *,
+    native_namespace: Any,
+) -> DataFrame[Any]:
     """
     Instantiate DataFrame from dictionary.
 
     Arguments:
         data: Dictionary to create DataFrame from.
+        schema: The DataFrame schema as dict of {name: type}, unused for pandas-like DataFrame's.
         native_namespace: The native library to use for DataFrame creation.
 
     Examples:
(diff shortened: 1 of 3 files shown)


narwhals/functions.py
```diff
@@ -88,15 +96,60 @@ def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
         └─────┴─────┘
     """
     if native_namespace is get_polars():
-        native_frame = native_namespace.from_dict(data)
+        if schema:
+            from narwhals._polars.utils import (
+                reverse_translate_dtype as polars_reverse_translate_dtype,
```
FBruzzesi, line 101: Aliasing is for mypy complaining that pandas one takes different number of arguments

narwhals/functions.py
```diff
@@ -88,15 +96,60 @@ def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
...
+                reverse_translate_dtype as pandas_like_reverse_translate_dtype,
+            )
+
+            implementation = (
+                Implementation.PANDAS
+                if native_namespace is get_pandas()
+                else Implementation.MODIN
+                if native_namespace is get_modin()
+                else Implementation.CUDF
+            )
```
FBruzzesi, line 129: I may have a tiny refactor in mind for this

narwhals/functions.py
```diff
@@ -48,12 +50,18 @@ def concat(
...
+    schema: dict[str, DType] | Schema | None = None,
+    *,
+    native_namespace: Any,
+) -> DataFrame[Any]:
     """
     Instantiate DataFrame from dictionary.
 
     Arguments:
         data: Dictionary to create DataFrame from.
+        schema: The DataFrame schema as dict of {name: type}, unused for pandas-like DataFrame's.
```
FBruzzesi, line 64: Schema has to be complete if provided (meaning all columns should be specified, and only supports Schema, dicts or None for now. Polars has more "variety"

narwhals/functions.py
```diff
@@ -88,15 +100,58 @@ def from_dict(data: dict[str, Any], *, native_namespace: Any) -> DataFrame[Any]:
...
+                if native_namespace is get_modin()
+                else Implementation.CUDF
+            )
+            schema = {
+                name: pandas_like_reverse_translate_dtype(
+                    schema[name], native_type, implementation
+                )
+                for name, native_type in native_frame.dtypes.items()
+            }
+            native_frame = native_frame.astype(schema)
```
MarcoGorelli, line 140: could we do the `astype` on each column before creating `native_frame`? it might save pandas from consolidating columns into 2D blocks which we might then need to split up if `schema` assigns them different dtypes

narwhals/functions.py
MarcoGorelli: Nice, I like that!

Maybe we should even expose `Implementation`, and `from_dict` could take `implementation` as an argument?

We haven't made a release yet so we can still make this change 😎

narwhals/functions.py
FBruzzesi: How would this work in practice? Meaning, I would expect that `data` is a dictionary which values can be something like python lists and I would not know how to cast those to (dataframe) native types before having the dataframe. What am I missing?

narwhals/functions.py
FBruzzesi: Let me open the PR to refactor implementation in the first place then 😇

narwhals/functions.py
MarcoGorelli: could you call `pd.Series(..., copy=False, dtype=dtype)` on each one?

`copy=False` should be fine as we don't have any in-place operations in the Narwhals API

MarcoGorelli: actually, what you've done is fine

MarcoGorelli (approved):
thanks @FBruzzesi 

looks like pandas would still copy when creating a dataframe from a dict of Series, and realistically I suspect people would use this method whilst relying on automatic dtype inference

MarcoGorelli:
> Maybe we should even expose Implementation, and from_dict could take implementation as an argument?

let's leave that to a follow-up, this looks good!

PR #608, opened 2024-07-24
chore: extend `Implementation` for all backends

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -15,6 +15,7 @@ from narwhals.dependencies import get_numpy
 from narwhals.dependencies import get_pyarrow
 from narwhals.dependencies import get_pyarrow_compute
 from narwhals.dependencies import get_pyarrow_parquet
+from narwhals.utils import Implementation
 from narwhals.utils import flatten
 from narwhals.utils import generate_unique_token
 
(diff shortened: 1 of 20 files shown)


MarcoGorelli (approved):
nice, thanks @FBruzzesi !

FBruzzesi:
>  just a mypy issue

How dare?!

MarcoGorelli:
really nice, thanks @FBruzzesi !

Maybe the signature is fine with `native_namespace` rather than `Implementation`? Because then, for anyone extending Narwhals, we give them a possibility of including `.from_dict` in their native namespace, whereas if they need to pass `Implementation`, then it's not clear how they can modify one of our enums

FBruzzesi:
> Maybe the signature is fine with `native_namespace` rather than `Implementation`? 

We may add it in the signature and one of the two has to be not None. 

> Because then, for anyone extending Narwhals, we give them a possibility of including `.from_dict` in their native namespace, whereas if they need to pass `Implementation`, then it's not clear how they can modify one of our enums

It would break in any case right? As it would branch to the block:

```py
 else:  # pragma: no cover
        msg = f"Expected library supported by Narwhals, got: {native_namespace}"
        raise ValueError(msg)
```

MarcoGorelli:
> It would break in any case right? As it would branch to the block:

true, but we could just do a blanket
```
else:
    native_namespace.from_dict(data
```
and note in the "extending.md" docs that the native namespace is expected to implement `from_dict`

FBruzzesi:
I like that :)

FBruzzesi:
Taking a look at this. We would need to expose an `implementation` attribute/property in Narhwals BaseFrame as well. Any concern?

MarcoGorelli:
i'm not sure I see why, but i'm probably missing something, if you open draft pr with what you have i'll take a look

PR #609, opened 2024-07-24
feat : added arrow `is_between` and `is_between_test.py`

anopsy:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # #372 
- Closes #

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/tests/series_only/is_between_test.py
+++ b/tests/series_only/is_between_test.py
@@ -2,11 +2,9 @@ from __future__ import annotations
 
 from typing import Any
 
-import pandas as pd
 import pytest
 
 import narwhals.stable.v1 as nw
-from narwhals.utils import parse_version
 from tests.utils import compare_dicts
 
 data = [1, 4, 2, 5]
@@ -24,11 +22,9 @@ data = [1, 4, 2, 5]
 def test_is_between(
     request: Any, constructor_series: Any, closed: str, expected: list[bool]
 ) -> None:
-    if "pandas_series_nullable_constructor" in str(constructor_series) and parse_version(
-        pd.__version__
-    ) < (2, 2):
-        # bug in old pandas
+    if "pandas_series_nullable_constructor" in str(constructor_series):
         request.applymarker(pytest.mark.xfail)
+
     ser = nw.from_native(constructor_series(data), series_only=True)
     result = ser.is_between(1, 5, closed=closed)
     compare_dicts({"a": result}, {"a": expected})
--- a/tests/series_only/is_between_test.py
+++ b/tests/series_only/is_between_test.py
@@ -1,0 +1,34 @@
+from __future__ import annotations
+
+from typing import Any
+
+import pandas as pd
+import pytest
+
+import narwhals.stable.v1 as nw
+from narwhals.utils import parse_version
+from tests.utils import compare_dicts
+
+data = [1, 4, 2, 5]
+
+
+@pytest.mark.parametrize(
+    ("closed", "expected"),
+    [
+        ("left", [True, True, True, False]),
+        ("right", [False, True, True, True]),
+        ("both", [True, True, True, True]),
+        ("none", [False, True, True, False]),
+    ],
+)
+def test_is_between(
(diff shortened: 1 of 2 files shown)


anopsy:
Okay so the  [XPASS(strict)]  issue seems to reappear in 2 checks, really curious about this one

tests/series_only/is_between_test.py
```diff
@@ -0,0 +1,34 @@
...
+    ],
+)
+def test_is_between(
+    request: Any, constructor_series: Any, closed: str, expected: list[bool]
+) -> None:
+    if "pandas_series_nullable_constructor" in str(constructor_series) and parse_version(
+        pd.__version__
+    ) < (2, 2):
+        # bug in old pandas
+        request.applymarker(pytest.mark.xfail)
```
MarcoGorelli, line 31: I think you can just remove this - I suspect you got it from the `value_counts` tests, but that was for a very specific bug in pandas for `value_counts`. in general, we don't need it

anopsy: yes, excatly it was from value_counts and I didn't understand its purpose that's why I asked if I should leave it there.

MarcoGorelli: ah I see sorry, I thought you were asking something like "it's already in an existing `test_is_between` test, should I leave it there?" - all good then 🙌

tests/series_only/is_between_test.py
```diff
@@ -0,0 +1,30 @@
...
+        ("both", [True, True, True, True]),
+        ("none", [False, True, True, False]),
+    ],
+)
+def test_is_between(
+    request: Any, constructor_series: Any, closed: str, expected: list[bool]
+) -> None:
+    if "pandas_series_nullable_constructor" in str(constructor_series):
+        request.applymarker(pytest.mark.xfail)
+
```
MarcoGorelli, line 27: sorry I meant - you can just remove this completely

anopsy: aaaah okay, sorry for the misunderstanding. Really curious why more checks are failing without it now.

MarcoGorelli: because now you're telling pytest "check that this fails for pandas nullable series" - however, the implementation passes for all libraries and all versions, so we don't need any `pytest.mark.xfail` 😎

anopsy: Now I see 🙈, I didn't notice that you marked the entire if stetement, but just removed it and noticed it worked. Thank you why though!

MarcoGorelli (approved):
thanks @anopsy !

tests/series_only/is_between_test.py
```diff
@@ -0,0 +1,25 @@
...
+        ("left", [True, True, True, False]),
+        ("right", [False, True, True, True]),
+        ("both", [True, True, True, True]),
+        ("none", [False, True, True, False]),
+    ],
+)
+def test_is_between(constructor_series: Any, closed: str, expected: list[bool]) -> None:
+    ser = nw.from_native(constructor_series(data), series_only=True)
+    result = ser.is_between(1, 5, closed=closed)
+    compare_dicts({"a": result}, {"a": expected})
```
FBruzzesi, line 25: Could you check what happens with different dtypes i.e. lower_bound and/or upper_bound being float instead of int?
I hope not, but we may need to do some casting?!

FBruzzesi: Never mind it is merged 😂

MarcoGorelli: 😄 still a good topic for a follow-up, I probably shouldn't merge so quickly some times

anopsy: I just checked with 1.0 and 5.0 and the test pass. Btw I forgot to remove series.is_between from backend completness, doing it now

PR #616, opened 2024-07-24
feat: arrow datetime namespace

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #372 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Edit: I am surprised that the min version of pyarrow doesn't break 😂🙌🏼


narwhals/_arrow/series.py
```diff
@@ -584,6 +584,162 @@ def to_string(self, format: str) -> ArrowSeries:  # noqa: A002
...
+        )
+        return self._arrow_series._from_native_series(result)
+
+    def ordinal_day(self: Self) -> ArrowSeries:
+        pc = get_pyarrow_compute()
+        return self._arrow_series._from_native_series(
+            pc.day_of_year(self._arrow_series._native_series)
+        )
+
+    def total_minutes(self: Self) -> ArrowSeries:
```
FBruzzesi, line 651: For all the `total_*` methods, I am using the following strategy:

- Extract array time unit (question, should we add more?)
- Convert time unit to factor of `total_*` unit
- Divide or multiply by such scalar: notice that duration cannot be multiplied/divided by float/double, hence we need to distinguish for when to multiply and when to divide
- Finally cast to int: this is because polars returns int dtype

narwhals/_arrow/series.py
```diff
@@ -584,6 +584,162 @@ def to_string(self, format: str) -> ArrowSeries:  # noqa: A002
...
+
+        unit_to_micro_factor = {
+            "ms": 1e3,  # milli
+            "us": 1,  # micro
+            "ns": 1e3,  # nano
+        }
+
+        factor = pa.scalar(unit_to_micro_factor[unit], type=pa.int64())
+
+        if unit == "ms":  # pragma: no cover
```
MarcoGorelli, line 717: is it possible to make a test which hits this? maybe it could be a separate test, in which you specifically instantiate a pyarrow array with 'ms' unit, rather than the parametrised one

MarcoGorelli (commented):
another day, another amazing PR 🎉

narwhals/_arrow/series.py
FBruzzesi: Edit: Possible time units for pyarrow duration are "s", "ms", "us", "ns". Now all these are covered :)

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #621, opened 2024-07-25
docs: replace `df_any`/`s_any` with `df`/`s`

### LINKED ISSUE #620: docs: use `df` instead of `df_any`, and `s` instead of `s_any` in docstring examples
There was some discussion about this on Discord, with opinions being pretty evenly split, but if I counted correctly, the simple `df` / `s` were slightly favoured

My reasoning is:

> `df` is already a generic symbol, not tied to any particular library, and keeping it simple lowers the cognitive load for anyone reading the docs

Let's use that consistently in the docs - so, instead of something like
```python
@nw.narwhalify
def func(df_any):
    return df_any.select(nw.col('a', 'b').min())
```
use
```python
@nw.narwhalify
def func(df):
    return df.select(nw.col('a', 'b').min())
```

aidoskanapyanov:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #620

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/docs/api-reference/typing.md
+++ b/docs/api-reference/typing.md
@@ -119,8 +119,8 @@ from narwhals.typing import IntoDataFrameT, DataFrameT
 
 df = pl.DataFrame({'a': [1,2,3]})
 
-def func(df_any: IntoDataFrameT) -> IntoDataFrameT:
-    df = nw.from_native(df_any, eager_only=True)
+def func(df: IntoDataFrameT) -> IntoDataFrameT:
+    df = nw.from_native(df, eager_only=True)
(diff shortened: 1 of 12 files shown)


MarcoGorelli (approved):
thanks @aidoskanapyanov ! looks like we had some typos to begin with, thanks for fixing

PR #626, opened 2024-07-26
chore: allow unknown/extension namespaces for `from_dict` function

FBruzzesi:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- [Related comment](https://github.com/narwhals-dev/narwhals/pull/608#issuecomment-2247362168)

## Checklist

- [x] Tests added 
- [x] Documented the changes

Opening as draft as I know tests are failing... due to stable api not implementing `Implementation` class. Unlike other classes, this one cannot be inherited from because of Enum, and I am not sure what's the proper way of working with the stable API section

### DIFF
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -43,6 +43,7 @@ from narwhals.translate import from_native
 from narwhals.translate import get_native_namespace
 from narwhals.translate import narwhalify
 from narwhals.translate import to_native
+from narwhals.utils import Implementation
 from narwhals.utils import is_ordered_categorical
 from narwhals.utils import maybe_align_index
 from narwhals.utils import maybe_convert_dtypes
@@ -101,4 +102,5 @@ __all__ = [
     "stable",
     "Schema",
     "from_dict",
+    "Implementation",
 ]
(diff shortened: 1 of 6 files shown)


MarcoGorelli (commented):
thanks for trying this out!

tbh seeing it I feel like perhaps we shouldn't do it, as I don't see how we could get it to work for [extensions](https://narwhals-dev.github.io/narwhals/extending/)

in the end,
```python
native_namespace = nw.get_native_namespace(df)
return nw.from_dict(data, native_namespace)
```
isn't too bad to right, and it lets us handle all cases - maybe it's good enough, and we can just add the error you added
```
            msg = "Unknown namespace is expected to implement `from_dict` function"
            raise AttributeError(msg) from e
```
?

FBruzzesi:
Sure! Let's just add the flexibility for extensions via namespace then :)

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #635, opened 2024-07-27
feat: Add initial (and very minimal) support for Dask.DataFrame

MarcoGorelli:
Alright, let's start stubbing this out 🥳 🚀 🎸 

As a follow-up, we'll need to add nightlies to `extremes.yml`

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/extremes.yml
+++ b/.github/workflows/extremes.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Run doctests
         run: pytest narwhals --doctest-modules
 
-  pandas-nightly:
+  pandas-nightly-and-dask:
     strategy:
       matrix:
         python-version: ["3.12"]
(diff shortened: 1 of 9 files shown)


MarcoGorelli:
@benrutter fancy playing around with this and seeing if you can take it further? I've merged it into main so we don't have a feature branch which lives forever - I'd prefer it if we could merge quite incrementally, else merge conflicts will be a pain - but is this enough to address some issues we'd run into in the other PR?

I think if we could aim to be able to execute TPC-H Q1 with Dask, that'd already be a pretty incredible achievement 🏆

benrutter:
Ah this is immense! Yeah I'll have a play around with it- I think conceotually, putting dask as it's own lazy-only implementation seems like a really good move for maintainability.

I'll have a play around with stuff this week, TPC-H Q1 seems like a really good first goal!

PR #638, opened 2024-07-27
feat: add shift for Dask Dataframe

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -142,3 +142,10 @@ class DaskExpr:
             lambda _input: _input.mean(),
             "mean",
         )
+
+    def shift(self, n: int) -> Self:
+        return self._from_call(
+            lambda _input, n: _input.shift(n),
+            "mean",
+            n,
+        )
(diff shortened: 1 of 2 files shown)


narwhals/_dask/expr.py
anopsy, line 1: I don't get what that "mean" has to do with shift 
```
 lambda _input, n: _input.shift(n),
            "mean",
            n,
```

MarcoGorelli: oooops

it should've been shift 🙈 😳 

it doesn't actually affect the results - the function name is only used for debugging at the moment. but still, I should've got this right in the reference PR

if you make a PR without mistakes (as you probably will) we can make that the reference one in the issue

MarcoGorelli: 😄 so sorry - I linked to a cleaner reference PR now, thank for pointing this out so it'll help other contributors too!

Issue #637, opened 2024-07-27
feat: add methods to Dask backend


MarcoGorelli:
Since #272  we have really minimal support for Dask DataFrame

Expr methods should go in `narwhals/_dask/expr.py`. For tests, you should use existing tests, and just remove the 
```
    if "dask" in str(constructor):
        request.applymarker(pytest.mark.xfail)
```
part. If you can remove that, and the test passes, it means you've done it correctly

## Not too hard

Examples of `Expr` methods which we should add (see [here](https://narwhals-dev.github.io/narwhals/api-reference/expr/) for the full list):
- `Expr.__sub__`
- `Expr.__mul__`
- `Expr.shift`
- `Expr.cum_sum`
- `Expr.is_between`

Note: we should not add anything which modifies the index. So, the following should _not_ be added, even though they appear on the list in the link above:
- Expr.sort
- Expr.head
- Expr.tail

## Harder

- `DataFrame.group_by`
- `DataFrame.filter`
- get things started with namespaces (e.g. `Expr.dt`, `Expr.str`, ...)

## General guidelines:
- please don't ask for the issue to be assigned to you
- please don't ask for permission to work on this issue
- if you're  a first time contribute, please choose 1 method at a time to implement, and leave a comment noting which method you're working on (if you've contributed to Narwhals before, feel free to choose multiple)
- have fun 🥳 

---

- Example pull request: https://github.com/narwhals-dev/narwhals/pull/731/files
- contributing guide: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md

---

An easy way to check if something still needs doing is to look for tests with
```
    if "dask" in str(constructor):
        request.applymarker(pytest.mark.xfail)
```
Then:
1. try removing those two lines
2. run the test, check that it fails
3. implement this functionality for Dask
4. check the test passes

anopsy:
I'll take Expr.__sub__

aidoskanapyanov:
I'll take `Expr.is_between`

anopsy:
Taking `Expr.__mul__`

DeaMariaLeon:
Take `Expr.sum`

FBruzzesi:
@MarcoGorelli I have a couple of questions:

- Could you expand a little bit more, here or somewhere else on the following:
  > we should not add anything which modifies the index
- Should we refer to this issue also for dataframe methods? i.e. not Expr only?

MarcoGorelli:
hey!

[...]

PR #667, opened 2024-07-29
feat: dask `select` method

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below.

Edit: For DaskNamespace, as of now:

- `_create_expr_from_callable`: implemented
- `_create_expr_from_series`, `_create_compliant_series`, `_create_series_from_scalar` raise `NotImplementedError`

feedbacks are welcomed!

PR #670, opened 2024-07-29
feat: dask `columns` property & `filter` method

benrutter:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637

## Checklist

- [x] Tests added 
- [ ] Documented the changes

This is a pretty small PR! I mainly wanted to get a feel for the API 😁 (initially made lots of changes as if it was a regular DataFrame and then remembered it was lazy!)

Adds in a `columns` property for the DaskLazyFrame.

EDIT: Ok, I couldn't resist tinkering. This now applies a fair bit more functionality, getting 'filter' to work (at least for some basic cases since the expressions still aren't fully complete).

I've added in "not-implemented" functions (i.e. empty functions that just raise an error) to appease mypy type checking. I figured this was a better move that `# type: ignore` annotation since they actually will be needed at some point, but let me know if the other approach would be better!

### DIFF
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -52,3 +52,7 @@ class DaskLazyFrame:
             implementation=Implementation.PANDAS,
             backend_version=parse_version(get_pandas().__version__),
         )
+
+    @property
+    def columns(self) -> list[str]:
+        return self._native_dataframe.columns.tolist()  # type: ignore[no-any-return]
--- a/tests/dask_test.py
+++ b/tests/dask_test.py
@@ -260,3 +260,13 @@ def test_str_to_lowercase(
 
     result_frame = df.with_columns(nw.col("a").str.to_lowercase())
     compare_dicts(result_frame, expected)
+
+
+def test_columns() -> None:
+    import dask.dataframe as dd
+
+    dfdd = dd.from_pandas(pd.DataFrame({"a": [1, 2, 3], "b": ["cat", "bat", "mat"]}))
+    df = nw.from_native(dfdd)
+
+    result = df.columns
+    assert set(result) == {"a", "b"}


MarcoGorelli (approved):
legend, mate!

benrutter:
Ah yikes! Sorry @MarcoGorelli but just a heads up that I (*annoyingly*) pushed changes at the exact same time as your approval!

(don't merge without rechecking since you'll get a lot more than you signed up for!)

MarcoGorelli:
😄 thanks Ben! Happy to take the extra stuff too, we just need test coverage for them

benrutter:
Working on it! Any guidance on stuff like the `__lt__` implementation testing? (I'm good to add testing for them alonside the `all_horizontal` and `_create_expr_from_callable` methods but wasn't sure if it's needed since they just call out the dask's own codebase)

MarcoGorelli:
i think just having some comparisons test in which gt/lt/le/ge are all tests together would be good
true that it's identical, but still

maybe we can soon add dask to the constructor fixture anyway, then we can just reuse existing tests

benrutter:
(ok, adding in proper testing has shown me it's actually a little more complex than I'd thought - filter etc is working for simple cases like `col('a') > 43` but something odd is happening when I'm cross comparing `col('a') > col('b')`.

I'll revert the last commit and put it in a bit later as a seperate PR!

FBruzzesi (commented):
This is awesome! I was planning a similar PR as a follow up from #667 !
I think `__and__` implementation is missing to make `all_horizontal` work with multiple statements 😇 but we can keep it as a follow up as well 🔥

MarcoGorelli (approved):

PR #676, opened 2024-07-29
Feat: Dask filter and __and__ method added

benrutter:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637

## Checklist

- [x] Tests added 
- [ ] Documented the changes

Following up from: https://github.com/narwhals-dev/narwhals/pull/670

Introduces filter for dask dataframes, plus a few required methods for comparison to get things working into a testable state. (@FBruzzesi hope it's ok, I went ahead and implemented `__and__` - I know you said you'd probably do this, so I hope I haven't stood on any toes! It seemed a quick add and otherwise testing all horizontal was a little messy)

Note: There's a pretty disgusting hack I've added in that should probably come out before anything is merged. Looks like narwhals `evaulate_into_exprs` has this bit:

```python
series.append(evaluated_expr[0].alias(name))0
```

I'm getting an error here because `evalated_exp[0]` is a dask-exp/dask-series (depending on version) which doesn't have an `alias` method.

I've added this fix:

```python
try:
    to_append = evaluated_expr[0].alias(name)
except AttributeError:
    to_append = evaluated_expr[0].rename(name)  # type: ignore[union-attr]
```

Which works, but the only thing is *should I even have an actual dask expressions here?* The code is running fine seemingly with my above hack, but it seems like this point in the code should have a narwhals abstraction instead?

### DIFF
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -51,7 +51,8 @@ class DaskNamespace:
         )
 
     def all_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
-        return reduce(lambda x, y: x & y, flatten(exprs))  # type: ignore[no-any-return]
+        # coverage shows 55->exit as uncovered?
+        return reduce(lambda x, y: x & y, flatten(exprs))  # type: ignore[no-any-return]  # pragma: no cover
 
     def _create_expr_from_series(self, _: Any) -> NoReturn:
         msg = "`_create_expr_from_series` for DaskNamespace exists only for compatibility"
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -8,7 +8,7 @@ from typing import NoReturn
 
 from narwhals import dtypes
 from narwhals._dask.expr import DaskExpr
-from narwhals._expression_parsing import parse_into_exprs
+from narwhals.utils import flatten
 
 if TYPE_CHECKING:
     from narwhals._dask.dataframe import DaskLazyFrame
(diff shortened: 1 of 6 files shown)


MarcoGorelli:
hey, thanks for working on this

> but it seems like this point in the code should have a narwhals abstraction instead?

i'd originally resisted making `DaskSeries`, as we don't expose it to the user,  but if we want to re-use code in `narwhals._expression_parsing.py`, we may need to...i'll take a look tomorrow

FBruzzesi:
Hey @benrutter, thanks for the PR, I will check in more depth in a bit.
And do not worry you are not stepping on anyone! There are enough tasks for all of us! 🙌🏼

MarcoGorelli:
thanks @benrutter ! Just pushed a commit

I think it might be simpler if we keep `_expression_parsing` separate for the Dask backend, and just use `_dask/utils.py`. This is for a couple of reasons:
- for Dask, we're a lot more constrained in what we accept - for example, we can't just accept a numpy array, or a free-standing Series. We just accept column names or expressions - else we don't know what kind of magic index alignment might happen under out feet (as opposed to in pandas where can just check the index values). I think it's OK to start constrained an possibly loosen later
- it gives a chance to revisit the expression parsing logic. I think `_expression_parsing.py` may well be a bit too complicated for its own good 😄

narwhals/_dask/namespace.py
```diff
@@ -43,6 +50,10 @@ def col(self, *column_names: str) -> DaskExpr:
             backend_version=self._backend_version,
         )
 
+    def all_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
+        # coverage shows 55->exit as uncovered?
```
FBruzzesi, line 54: It should apply when conditions are comma separated, otherwise the "&" is resolved by the expr. Meaning:

```py
.filter(nw.col("a") & nw.col("b") #-> enters __and__ method
.filter(nw.col("a"), nw.col("b") #-> enters all_horizontal function to resolve it with __and__ inside it
```

MarcoGorelli: thanks! any objections to what I've added, or shall we ship it?

FBruzzesi: I didn't go through the code in its entirety. I agree that  Dask won't allow us to achieve something, so it's ok to move forward and eventually try to address corner cases

MarcoGorelli: cool, thanks!

MarcoGorelli (approved):
thanks @benrutter !

PR #684, opened 2024-07-30
docs: explain translation from narwhals api to native api

amol-:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [X] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

The current documentation is unclear about the how objects move from narwhals api to polars compatible backend to native api. This adds a section that tries to make that more clear.

### DIFF
--- a/docs/how_it_works.md
+++ b/docs/how_it_works.md
@@ -152,78 +152,68 @@ and `narwhals.series` coordinate how to dispatch the Narwhals API to each backen
 
 ## Mapping from API to implementations
 
-End user works with narwhals apis, it won't use directly the native dataframe APIS,
-so narwhals will take care itself of translating
-the calls done to the narwhals api to calls to the dataframe wrapper
-( `PandasLikeNamespace`, `PandasLikeDataFrame` or `PandasLikeExpr`) which then
-forwards the calls to the native implementation. Translation of native narwhals
-apis to the dataframe wrapper will usually happen via the namespace.
-Usually you will find the namespace referred as
-as `plx` which stands for *Polars compliant namespace for Library X*
+When end users call Narwhals APIs, how do they get translated to the native Dataframe APIs?
 
-That usually happens in a few different ways:
+Things generally go through a couple of layers:
 
-- `narwhals.DataFrame` -> `PandasLikeDataFrame` is available via the
-  `narwhals.DataFrame._compliant_frame` attribute of the Dataframe.
-  For example in case of a `pyarrow` we would have:
-  ```python-console
-  >>> nwdf = nw.from_native(table)
-  >>> type(nwdf)
-  <class 'narwhals.stable.v1.DataFrame'>
-  >>> nwdf._compliant_frame
-  <narwhals._arrow.dataframe.ArrowDataFrame object at 0xe9a1e3f3b050>
-  ```
-- `narwhals.Expr` -> `PandasLikeExpr` happens via calling `Expr._call` on the
-  target namespace. For example in case of `pyarrow` we would have:
-  ```python-console
-  >>> type(nwdf)
-  <class 'narwhals.stable.v1.DataFrame'>
-  >>> nw.col("b").len()._call(nwdf.__narwhals_namespace__())
-  ArrowExpr(depth=1, function_name=col->len, root_names=['b'], output_names=['b']
-  ```
-- `narwhals.Series` -> `PandasLikeSeries` happens via the
-  `narwhals.Series._compliant_series` attribute of the Series.
-  For example, in case of `pyarrow` we would have:
-  ```python-console
-  >>> nwdf = nw.from_native(table)
-  >>> type(nwdf)
-  <class 'narwhals.stable.v1.DataFrame'>
-  >>> nwseries = nwdf["b"]
-  >>> type(colseries)
-  <class 'narwhals.stable.v1.Series'>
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
thanks @amol- , amazing stuff here! I've just reworded it a bit, and will try to do a bit more tomorrow, but I think this is super-important to be documented, so thanks a tonne for having started it 🙏

amol-:
@MarcoGorelli I think that in the rewording we lost a little bit the message that the last part tried to convey.
In the end I was trying to explain and showcase how `narwhals.Dataframe` and `PandasLikeDataframe` can be used interchangeably in the narwhals api and that thus even if methods return the "Like" version the user doesn't have to care about that.

I'm referring to this part:

After a compliant object is returned, the user will keep using and perform
additional operations on the compliant object itself, as the compliant
object implements the same API as the narwhals objects and is accepted
as an argument by all narwhals functions.

For example `nw.col("b")._call(nwdf.__narwhals_namespace__())` on a
`pyarrow.Table` will return an `ArrowExpr`:

```python
  ArrowExpr(depth=0, function_name=col, root_names=['b'], output_names=['b']
```

but invoking `narwhals.Dataframe.select` on it will work and return
a new `narwhals.Dataframe` as if we passed a `narwhals.Expr` itself:

```python
>>> nwexpr = nw.col("b")
>>> type(nwexpr)
<class 'narwhals.stable.v1.Expr'>
>>> nwr = nwdf.select(nw.col("b"))
>>> type(nwr)
<class 'narwhals.stable.v1.DataFrame'>
>>> arrowexpr = nw.col("b")._call(nwdf.__narwhals_namespace__())
>>> type(arrowexpr)
<class 'narwhals._arrow.expr.ArrowExpr'>
>>> nwr = nwdf.select(arrowexpr)
>>> type(nwr)
<class 'narwhals.stable.v1.DataFrame'>
```

MarcoGorelli:
Thanks - I'll double check but I'm not sure it's intentional that that's supported

MarcoGorelli:
In fact, I just applied the following diff, ran the test suite, and no test failed:
```diff
diff --git a/narwhals/dataframe.py b/narwhals/dataframe.py
index 644db7bc..07c567bd 100644
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -123,6 +123,9 @@ class BaseFrame(Generic[FrameT]):
         *exprs: IntoExpr | Iterable[IntoExpr],
         **named_exprs: IntoExpr,
     ) -> Self:
+        from narwhals._pandas_like.expr import PandasLikeExpr
+        if any(isinstance(x, PandasLikeExpr) for x in exprs):
+            1/0
```

Passing a compliant Expr to `nw.DataFrame.select` doesn't happen internally, and it's not something users are meant to do - could you clarify why it needs mentioning please? I might be missing something though, so sorry if that's the case 🙏

amol-:
> In fact, I just applied the following diff, ran the test suite, and no test failed:
> 
> ```diff
> diff --git a/narwhals/dataframe.py b/narwhals/dataframe.py
> index 644db7bc..07c567bd 100644
> --- a/narwhals/dataframe.py
> +++ b/narwhals/dataframe.py
> @@ -123,6 +123,9 @@ class BaseFrame(Generic[FrameT]):
>          *exprs: IntoExpr | Iterable[IntoExpr],
>          **named_exprs: IntoExpr,
>      ) -> Self:
> +        from narwhals._pandas_like.expr import PandasLikeExpr
> +        if any(isinstance(x, PandasLikeExpr) for x in exprs):
> +            1/0
> ```
> 
> Passing a compliant Expr to `nw.DataFrame.select` doesn't happen internally, and it's not something users are meant to do - could you clarify why it needs mentioning please? I might be missing something though, so sorry if that's the case 🙏

Ah good to know, I thought that the interoperability was by design. So every API always should invoke `_from_compliant_dataframe` (or equivalent) to surface to users only the `narwhals.*` objects? It might make sense to document that.

MarcoGorelli:
Yup! I think the type hints usually make that clear? Most methods return `Self`

Anyway, thanks for your PR, let's ship this

MarcoGorelli (approved):
thanks @amol- !

docs/how_it_works.md
```diff
@@ -150,6 +150,89 @@ Each implementation defines its own objects in subfolders such as `narwhals._pan
...
+then how does that get mapped to the underlying dataframe's native API? Let's walk through
+this example to see.
+
+Things generally go through a couple of layers:
+
+- The user calls some top-level Narwhals API.
+- The Narwhals API forwards the call to a Narwhals-compliant dataframe wrapper, such as
+    - `PandasLikeDataFrame` / `ArrowDataFrame` / `PolarsDataFrame` / ...
+    - `PandasLikeSeries` / `ArrowSeries` / `PolarsSeries` / ...
+    - `PandasLikeExpr` / `ArrowExpr` / `ArrowSeries` / ...
```
anopsy, line 169: I was wondering about that `ArrowSeries` mentioned next to `PandasLikeExpr` / `ArrowExpr`. We do have PolarsExpr so it caught my attention, but then as far as I remember Polars is backed by Arrow Memory Arrays, is this why we see here `ArrowSeries` instead of `PolarsExpr`?

PR #692, opened 2024-07-31
feat: dask `with_row_index` and `rename`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 

## Checklist

- [x] Tests added 
- [ ] Documented the changes


narwhals/_dask/dataframe.py
```diff
@@ -94,3 +94,13 @@ def select(
 
     def drop_nulls(self) -> Self:
         return self._from_native_dataframe(self._native_dataframe.dropna())
+
+    def with_row_index(self: Self, name: str) -> Self:
+        return self._from_native_dataframe(
+            self._native_dataframe.assign(**{name: 1}).assign(
+                **{name: lambda t: t[name].cumsum() - 1}
+            )
```
EdAbati, line 102: (I'm not an expert in Dask, but I worked quite a bit with Spark so my interpretation/way of thinking comes from there)

I am wondering how the performance would be in a distributed setting with partitioned data. Would `cumsum` require to do the calculations in a single node?

Spark with the Pandas API has various ways to set an index: https://spark.apache.org/docs/latest/api/python/user_guide/pandas_on_spark/options.html#default-index-type

Could we maybe use this as an inspiration? Or is it better to create an array and add it as a column as we do for pandas-like dfs? 🤔

FBruzzesi: > I am wondering how the performance would be in a distributed setting with partitioned data. Would `cumsum` require to do the calculations in a single node?

I am not sure, I find dask documentation a bit unexplicit on such topic. I based the implementation on a [TomAugspurger SO answer](https://stackoverflow.com/a/60852409/12411536)

> Spark with the Pandas API has various ways to set an index: https://spark.apache.org/docs/latest/api/python/user_guide/pandas_on_spark/options.html#default-index-type
> 
> Could we maybe use this as an inspiration? Or is it better to create an array and add it as a column as we do for pandas-like dfs? 🤔

Thanks, I will take a closer look and see if that's feasible

EdAbati: Ah I see.

TBH I'm not 100% sure that it is a problem, just wanted to mention it since row index in Spark was a bit tricky. 🙂

We could also decide to investigate this in a follow-up

aidoskanapyanov: Dask has implemented some parallel algorithms for `cumsum` / `cumprod` based on parallel prefix scan algorithms. I don't really know the details, but it's cool stuff :)

Here's a link to PR for reference https://github.com/dask/dask/pull/6675

EdAbati: (sorry for the late reply) Very interesting!

Should we add the `method='blelloch'` to use this fancy algorithm? https://docs.dask.org/en/stable/generated/dask.array.cumsum.html

We can also add a comment to say that the implementation comes from that SO answer

FBruzzesi: I also didn't come back to this!

> Should we add the method='blelloch' to use this fancy algorithm?

The docs state _"More benchmarking is necessary."_, but also the PR was merge almost 4 years ago so I am not sure

> We can also add a comment to say that the implementation comes from that SO answer

Sure! Adding that right away

MarcoGorelli (approved):
thanks @FBruzzesi , and @EdAbati  + @aidoskanapyanov for reviewing!

PR #695, opened 2024-07-31
feat: Add `Expr.round` to Dask

mistShard:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 
- Closes #

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [ ] Documented the changes

PR #696, opened 2024-07-31
feat: add `round` to Arrow

lucianosrp:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 #537 #695
- Closes #

## Checklist

- [x] Tests added 
- [ ] Documented the changes

I was going to the various methods and I found out that `round` was not yet implemented for `arrow`.

So..this one can be a bit problematic as I've seen on #537.
[The discussion on different rounding modes](https://github.com/narwhals-dev/narwhals/pull/306/files#r1641848166) seems to imply that polars is rounding "towards zero". 
~~However,~~ when I tried to specify Pyarrow's `"towards_zero"` rounding mode, the tests were failing:

```shell
E       assert [1.1, 2.5, 3.9] == [1.1, 2.6, 3.9]
E         
E         At index 1 diff: 2.5 != 2.6
E         Use -v to get more diff

tests/expr_and_series/round_test.py:31: AssertionError
```

With the default `"half_to_even"` roundign mode, the tests are passing!

Are we ok to settle it with `"half_to_even"`  for now?
Here is a snippet from [NumPy's docs](https://numpy.org/doc/stable/reference/generated/numpy.round.html) for ref:
>For values exactly halfway between rounded decimal values, NumPy rounds to the nearest even value. Thus 1.5 and 2.5 round to 2.0, -0.5 and 0.5 round to 0.0, etc.

So I understand that polars rounding != (pands, numpy, python,etc.) 
When passing 2.5 in the testing data, polars is the only one to fail but I am not able to repeat this behaviour with the rounding_modes in arrow.


FBruzzesi (approved):
Thanks for the PR! Let's ship it!

Cl00GIF

FBruzzesi:
Sorry I just read the comments on specific rounding.

I think polars does the equivalent of pyarrow `round_mode = "half_towards_infinity"`, shall we try that?

lucianosrp:
@FBruzzesi  - What about now? 

I think it is worth noting that if we were to add 2.5 in the test case, both polars and pyarrow would fail.
image

FBruzzesi:
I think that's because the built-in round has same behaviour of pandas, i.e. round to closest even. If we want to follow Polars then arrow allows for that.

Maybe we need to build expected manually instead of using round in the first place

tests/expr_and_series/round_test.py
```diff
@@ -8,10 +9,23 @@
...
+    return rounded
+
+
+def round_per_backend(backend: str, n: float, decimals: int) -> float:
+    if backend.startswith(("polars", "pyarrow")):
+        return round_half_towards_infinity(n, decimals)
+    else:
+        return round(n, decimals)
+
+
```
lucianosrp, line 26: @FBruzzesi  would this be too overengineered?

FBruzzesi: Maybe a bit, especially because we are ending up testing something which each library is (should be) testing already.

On a practical level, I would keep the tests as they were, and add pyarrow backend in the Notes for both [series](https://github.com/narwhals-dev/narwhals/blob/e3de9a07df2461e3f1fbff7d54fd8f7d037843ca/narwhals/series.py#L1936) and [expr](https://github.com/narwhals-dev/narwhals/blob/e3de9a07df2461e3f1fbff7d54fd8f7d037843ca/narwhals/expr.py#L1640)

WDYT?

lucianosrp: Ah yes, I though so 😝
Yep, all makes sense!

FBruzzesi (approved):
Now we definitely ship it
DolphinGriddyGIF

PR #713, opened 2024-08-04
feat: add `date` to datetime namespace

lucianosrp:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 #372 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes


narwhals/_pandas_like/series.py
```diff
@@ -704,6 +704,11 @@ class PandasLikeSeriesDateTimeNamespace:
     def __init__(self, series: PandasLikeSeries) -> None:
         self._pandas_series = series
 
+    def date(self) -> PandasLikeSeries:
+        return self._pandas_series._from_native_series(
+            self._pandas_series._native_series.dt.date,
+        )
```
FBruzzesi, line 710: I wonder what are the pros and cons of returning a series of type `datetime64[ns]`.

Polars returns a `Date` type which we can continue to chain. Example from the docstring you added:

```py
s_pl.dt.date().dt.year()
```

lucianosrp: Good point! 
Not sure how to go about from this point on... I think this would be the same with  the existing other `dt` methods (?)
Maybe it will require to have a custom class to handle dt chainign on pandas?

MarcoGorelli: thanks @FBruzzesi for observing this!

I think we should only support this if the native series is backed by a pyarrow dtype (e.g. `'timestamp[ns][pyarrow]'`), so that pandas gives us a `date32[day][pyarrow]` output, and otherwise raise notimplementederror

to make the check easier across libraries, maybe assign `.dt.date` to some variable `result`, then check if `result.dtype == object`, and if so, raise not implemented, advising to use pyarrow-backed dtypes?

lucianosrp:
Will work to get this up-to-`**date**`

lucianosrp:
@FBruzzesi , @MarcoGorelli 

Any suggestions on  how we could handle [the failure on pandas min version?](https://github.com/narwhals-dev/narwhals/actions/runs/10290109023/job/28479328500?pr=713)

pyarrow backend on pandas is since v2.0, so the minimum pandas version (`1.1.5`) does not have the `convert_dtypes` argument. 

Here are some thoughts on how to handle that:  
1 - change example to use dask and polars  - this would be a first? 
2 - don't run that line in the doctests for this method - with `# doctest: +SKIP` 
 
 Curious to know your thoughts/comments  :)

lucianosrp:
@FBruzzesi , @MarcoGorelli 

Any suggestions on  how we could handle [the failure on pandas min version?](https://github.com/narwhals-dev/narwhals/actions/runs/10290109023/job/28479328500?pr=713)

pyarrow backend on pandas is since v2.0, so the minimum pandas version (`1.1.5`) does not have the `convert_dtypes` argument. 

Here are some thoughts on how to handle that:  
1 - change example to use dask and polars  - this would be a first? 
2 - don't run that line in the doctests for this method - with `# doctest: +SKIP` 
 
 Curious to know your thoughts/comments  :)

FBruzzesi:
Hey thanks for adjusting so quickly! I really like the solution you found 🙌🏼

> 2 - don't run that line in the doctests for this method - with `# doctest: +SKIP`

I would go with this one, we skip doctest somewhere else for the very same reason. 
It's good to have examples mostly for users, testing on those comes as a byproduct (at least this is my take)

lucianosrp:
We should be all good now🤞
Thanks for the review! Grazie 😉

FBruzzesi (approved):
Awesome! It's a +1 from me 🚀

AwwKittenGIF

MarcoGorelli (approved):

PR #715, opened 2024-08-04
feat: add `strip_chars` to string namespace

lucianosrp:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 #372 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from string import whitespace
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
@@ -794,6 +793,7 @@ class ArrowSeriesStringNamespace:
 
     def strip_chars(self: Self, characters: str | None = None) -> ArrowSeries:
         pc = get_pyarrow_compute()
+        whitespace = " \t\n\r\v\f"
         return self._arrow_series._from_native_series(
             pc.utf8_trim(
                 self._arrow_series._native_series,
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -797,7 +797,7 @@ class ArrowSeriesStringNamespace:
         return self._arrow_series._from_native_series(
             pc.utf8_trim(
                 self._arrow_series._native_series,
-                whitespace + "\n" if characters is None else characters,
+                whitespace if characters is None else characters,
             )
         )
 
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from string import whitespace
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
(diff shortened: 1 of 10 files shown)


narwhals/_arrow/series.py
```diff
@@ -791,6 +791,16 @@ class ArrowSeriesStringNamespace:
     def __init__(self: Self, series: ArrowSeries) -> None:
         self._arrow_series = series
 
+    def strip_chars(self: Self, characters: str | None = None) -> ArrowSeries:
+        pc = get_pyarrow_compute()
+        whitespace = " \t\n\r\v\f"
+        return self._arrow_series._from_native_series(
+            pc.utf8_trim(
+                self._arrow_series._native_series,
+                whitespace if characters is None else characters,
```
FBruzzesi, line 800: Nit suggestion:
```suggestion
                characters or whitespace
```

narwhals/expr.py
```diff
@@ -1833,6 +1833,50 @@ class ExprStringNamespace:
...
+            >>> func(df_pl)
+            shape: (2, 2)
+            ┌────────┬──────────┐
+            │ fruits ┆ stripped │
+            │ ---    ┆ ---      │
+            │ str    ┆ str      │
+            ╞════════╪══════════╡
+            │ apple  ┆ apple    │
+            │        ┆ mango    │
+            │ mango  ┆          │
```
FBruzzesi, line 1873: This rendering hit me so hard 🙈
Shall we use another "whitespace" character instead of a new line?

narwhals/series.py
```diff
@@ -2148,6 +2148,50 @@ class SeriesStringNamespace:
...
+            ┌────────┬──────────┐
+            │ fruits ┆ stripped │
+            │ ---    ┆ ---      │
+            │ str    ┆ str      │
+            ╞════════╪══════════╡
+            │ apple  ┆ apple    │
+            │        ┆ mango    │
+            │ mango  ┆          │
+            └────────┴──────────┘
+        """
```
FBruzzesi, line 2190: Here we should exemplify the behaviour on a series, not on an Expr 🙃

FBruzzesi (commented):
Thanks for the PR, it is very neat 🚀
I left a few considerations that may need to be addressed

narwhals/expr.py
lucianosrp: ikr 😂, I was copying from polar's [Expr.str.strip_chars](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.strip_chars.html#polars.Expr.str.strip_chars) documentation.. I think it's because "\n" is the most _obvious_ trailing whitespace.

See for example how "\t" is not that obvious in the repr 
```pycon
>>> df = pl.DataFrame({"col_a": "mango\t", "col_b":"mango"})
>>> df
shape: (1, 2)
┌────────┬───────┐
│ col_a  ┆ col_b │
│ ---    ┆ ---   │
│ str    ┆ str   │
╞════════╪═══════╡
│ mango	 ┆ mango │
└────────┴───────┘
```
Maybe there is a better way to represent it ? Possibly with a final `to_dict` ?

narwhals/series.py
lucianosrp: Oops ! 😁

narwhals/expr.py
MarcoGorelli: yeah `to_dict(as_series=False)` might be good! and maybe even a good idea for an upstream PR to Polars if you fancy it 😉

FBruzzesi (approved):
Nice one!

CatCatJumpingGIF

PR #702, opened 2024-08-01
patch: `DaskExpr` scalar reduction

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #694 

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below.

This is a first attempt to resolve #694.

1. I am craving for feedbacks on this one
2. There is still a broadcasting "edge case" (left most is scalar) which I would like to try to address before merging 
  - Example is: `df.select(nw.col("a").min(), nw.col("b"))`
  - Main issue is that I can't broadcast that with dask in the first place as of now

Issue #694, opened 2024-07-31
[Bug]: `DaskExpr` reduction to scalar can't be `select`-ed


FBruzzesi:
### Describe the bug

When selecting an expression which ends up in a reduction to a scalar, `AttributeError` is raised:

> AttributeError: 'Scalar' object has no attribute 'name'

### Steps or code to reproduce the bug

```py
import dask.dataframe as dd
import pandas as pd

df = nw.from_native(
    dd.from_pandas(
        pd.DataFrame(data={"a": [1., 3, 2]})
    )
)

for attr in ("min", "max", "mean", "sum"):
    try:
        df.select(getattr(nw.col("a"), attr)())
    except:
        print(f"{attr=} run into exception")
```

### Expected results

No error is thrown

### Actual results

AttributeError get raised

### Please run narwhals.show_version() and enter the output below.

```shell
main branch + 

System:
    python: 3.10.12 (main, Mar 22 2024, 16:50:05) [GCC 11.4.0]
executable: /home/fbruzzesi/open-source/narwhals/.venv/bin/python
   machine: Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 1.1.9
       pandas: 2.2.2
       polars: 1.2.1
         cudf: 
        modin: 0.31.0
      pyarrow: 17.0.0
        numpy: 2.0.1
```

### Relevant log output

<details>
<summary>Entire log</summary>
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/open-source/narwhals/.venv/lib/python3.10/site-packages/dask_expr/_core.py:467, in Expr.__getattr__(self, key)
    466 try:
--> 467     return object.__getattribute__(self, key)
    468 except AttributeError as err:

File ~/open-source/narwhals/.venv/lib/python3.10/site-packages/dask_expr/_expr.py:448, in Expr.name(self)
    446 @property
    447 def name(self):
--> 448     return self._meta.name

AttributeError: 'numpy.float64' object has no attribute 'name'

During handling of the above exception, another exception occurred:

AttributeError                            Traceback (most recent call last)
File ~/open-source/narwhals/.venv/lib/python3.10/site-packages/dask_expr/_collection.py:619, in FrameBase.__getattr__(self, key)
    616 try:
    617     # Fall back to `expr` API
    618     # (Making sure to convert to/from Expr)
--> 619     val = getattr(self.expr, key)
[...]

PR #716, opened 2024-08-04
patch: pandas-like and pyarrow scalar reduction fix

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #694, as mentioned in [comment](https://github.com/narwhals-dev/narwhals/pull/702#discussion_r1703230227) there are cases in which the issue in 694 doesn't match polars behavior for pandas and pyarrow as well.

Compared to dask, this is easier to fix as we know series length due to eagerness

## Checklist

- [x] Tests added 
- [ ] Documented the changes


narwhals/_pandas_like/utils.py
```diff
@@ -417,9 +417,16 @@ def reverse_translate_dtype(  # noqa: PLR0915
 
 
 def validate_indices(series: list[PandasLikeSeries]) -> list[Any]:
-    idx = series[0]._native_series.index
-    reindexed = [series[0]._native_series]
-    for s in series[1:]:
+    lengths = [len(s) for s in series]
+    max_length = max(lengths)
+
+    idx = next(s for s in series if len(s) == max_length)._native_series.index
```
FBruzzesi, line 423: This is the first index that has max length, hence the left-most non scalar index will still guide the final index

narwhals/_arrow/utils.py
```diff
@@ -246,3 +250,32 @@ def cast_for_truediv(arrow_array: Any, pa_object: Any) -> tuple[Any, Any]:
...
+
+    for s, length in zip(series, lengths):
+        if max_length > 1 and length == 1:
+            value = s._native_series[0]
+
+            if backend_version < (13,) and hasattr(value, "as_py"):  # pragma: no cover
+                value = value.as_py()
+
+            reshaped.append(
+                pa.chunked_array([[value] * max_length], type=s._native_series.type)
```
FBruzzesi, line 276: Should we create the array with numpy instead of a python list?

narwhals/_pandas_like/utils.py
```diff
@@ -417,9 +417,16 @@ def reverse_translate_dtype(  # noqa: PLR0915
 
 
 def validate_indices(series: list[PandasLikeSeries]) -> list[Any]:
-    idx = series[0]._native_series.index
-    reindexed = [series[0]._native_series]
-    for s in series[1:]:
+    lengths = [len(s) for s in series]
+    max_length = max(lengths)
+
+    idx = series[lengths.index(max_length)]._native_series.index
```
FBruzzesi, line 423: This is the first index that has max length, hence the left-most non scalar index will still guide the resulting index

tests/expr_and_series/reduction_test.py
```diff
@@ -0,0 +1,80 @@
...
+    df = nw.from_native(constructor(data))
+    result = df.with_columns(
+        nw.col("a").min().alias("min"),
+        nw.col("b").max().alias("max"),
+        nw.col("a", "b").mean(),
+    )
+    expected = {"min": [1, 1, 1], "max": [6, 6, 6], "a": [2, 2, 2], "b": [5, 5, 5]}
+    compare_dicts(result, expected)
+
+    df = nw.from_native(constructor(data))
```
FBruzzesi, line 60: I am recreating the dataframe each time because modin was resulting is a weird weird bug, or at least I think so.

I will try to re-factor later on

narwhals/_arrow/utils.py
MarcoGorelli: i think it doesn't matter, and what you've done should be faster

narwhals/_pandas_like/utils.py
```diff
@@ -417,9 +417,16 @@ def reverse_translate_dtype(  # noqa: PLR0915
...
+    reindexed = []
+
+    for s, length in zip(series, lengths):
+        if max_length > 1 and length == 1:
+            s = s._from_native_series(s._native_series.repeat(max_length))  # noqa: PLW2901
+
```
MarcoGorelli, line 429: like this, we're setting an index, and then overwriting it in the `set_axis` call below

can we directly do
```python
if max_length > 1 and length == 1:
    native_series = make_native_series([item(s._native_series)]*max_length, index=idx, name=s.name)
    reindexed.append(s._from_native_series(native_series))
else:
        if s._native_series.index is not idx:
```
?

narwhals/_pandas_like/utils.py
```diff
@@ -417,9 +417,16 @@ def reverse_translate_dtype(  # noqa: PLR0915
 
 
 def validate_indices(series: list[PandasLikeSeries]) -> list[Any]:
```
MarcoGorelli, line 419: both here, and in the new function for pyarrow, maybe let's call them `broadcast_series`?

MarcoGorelli (commented):
nice one! thanks for noticing plus fixing

narwhals/_arrow/utils.py
FBruzzesi: I somehow am afraid that time is similar, but `np.full(max_length, value)` is more memory efficient, especially if we had a way to map arrow to numpy datatypes

narwhals/_pandas_like/utils.py
FBruzzesi: Nice catch! Let me adjust

narwhals/_arrow/utils.py
MarcoGorelli: sure let's do that

narwhals/_arrow/utils.py
```diff
@@ -246,3 +251,95 @@ def cast_for_truediv(arrow_array: Any, pa_object: Any) -> tuple[Any, Any]:
...
+                    type=pa_dtype,
+                )
+            )
+        else:
+            reshaped.append(s_native)
+
+    return reshaped
+
+
+def pyarrow_to_numpy_dtype(pa_dtype: Any) -> Any:  # pragma: no cover
```
FBruzzesi, line 291: Shall we test this? 🙈

narwhals/_arrow/utils.py
```diff
@@ -246,3 +251,95 @@ def cast_for_truediv(arrow_array: Any, pa_object: Any) -> tuple[Any, Any]:
...
+
+            value = s_native[0]
+
+            if hasattr(value, "as_py"):  # pragma: no cover
+                value = value.as_py()
+
+            reshaped.append(
+                pa.array(
+                    np.full(shape=max_length, fill_value=value, dtype=np_dtype),
+                    type=pa_dtype,
```
FBruzzesi, line 282: Using `pa.array` directly because of [pyarrow-numpy integration docs](https://arrow.apache.org/docs/python/numpy.html#numpy-to-arrow)

narwhals/_arrow/utils.py
FBruzzesi: Alright, if this doesn't deserve or require a test, PR should be ready to review

tests/expr_and_series/reduction_test.py
```diff
@@ -0,0 +1,66 @@
...
+            [nw.col("a"), nw.col("b").min().alias("min")],
+            {"a": [1, 2, 3], "min": [4, 4, 4]},
+        ),
+    ],
+    ids=range(5),
+)
+def test_scalar_reduction_select(
+    request: Any, constructor: Any, expr: list[Any], expected: dict[str, list[Any]]
+) -> None:
+    if "dask" in str(constructor) and int(request.node.callspec.id[-1]) != 1:
```
FBruzzesi, line 31: Don't panic (just yet): this friendly `request.node.callspec.id` will look like the following: `<constructor_name>-<id>`. The `id` is specified in `@pytest.mark.parametrize` and ranges from 0 to 4 (number of test cases).

As dask passes the second test which corresponds to `request.node.callspec.id="dask_lazy_constructor-1"`, I am extracting the id value



polars_lazy_constructor-4

narwhals/_arrow/utils.py
FBruzzesi: Rollback:

RollbackGIF

FBruzzesi: eventually we can split numeric vs rest

narwhals/_arrow/utils.py
```diff
@@ -246,3 +250,27 @@ def cast_for_truediv(arrow_array: Any, pa_object: Any) -> tuple[Any, Any]:
...
+
+    pa = get_pyarrow()
+
+    reshaped = []
+    for s, length in zip(series, lengths):
+        s_native = s._native_series
+        if max_length > 1 and length == 1:
+            value = s_native[0]
+            if hasattr(value, "as_py"):  # pragma: no cover
+                value = value.as_py()
```
MarcoGorelli, line 271: does this need doing everywhere or just for `self._backend_version < (13,)`?

MarcoGorelli (approved):
nice one, thanks @FBruzzesi !

narwhals/_arrow/utils.py
FBruzzesi: Need to check, numpy was certainly failing to create an array with a value of arrow scalar

PR #723, opened 2024-08-05
feat: support reductions in lazyframe.select for Dask

MarcoGorelli:
🤔 getting closer... ~not totally happy with this though as `df.select(nw.col('a').sum().alias('a_sum'))` wouldn't work~

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added 
- [ ] Documented the changes

PR #727, opened 2024-08-06
feat: Added `Expr.all` and `Expr.any` to Dask

mistShard:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

I used `all_horizontal_test.py` to run tests. All tests pass.

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -4,6 +4,7 @@ from copy import copy
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
+from typing import NoReturn
 
 from narwhals.dependencies import get_dask
 
@@ -419,6 +420,16 @@ class DaskExpr:
             decimals,
             returns_scalar=False,
         )
+
+    def drop_nulls(self) -> NoReturn:
+        # We can't (yet?) allow methods which modify the index
+        msg = "`Expr.drop_nulls` is not supported for the Dask backend. Please use `LazyFrame.drop_nulls` instead."
+        raise NotImplementedError(msg)
+
+    def sort(self, *, descending: bool = False, nulls_last: bool = False) -> NoReturn:
+        # We can't (yet?) allow methods which modify the index
+        msg = "`Expr.sort` is not supported for the Dask backend. Please use `LazyFrame.sort` instead."
+        raise NotImplementedError(msg)
 
     def abs(self) -> Self:
         return self._from_call(


narwhals/_dask/expr.py
```diff
@@ -272,6 +272,19 @@ def round(self, decimals: int) -> Self:
...
+        *,
+        ignore_nulls: bool = True,
+    ) -> Self:
+        return self._from_call(
+            lambda _input, ignore_nulls: _input.all(
+                axis=None, skipna=ignore_nulls, split_every=False, out=None
+            ),
+            "all",
+            ignore_nulls,
+        )
```
FBruzzesi, line 286: 👌👌

FBruzzesi (commented):
Implementation is on point 🙌🏼 

However the test you modified doesn't test `Expr.all`, and now that I am checking we don't have a dedicated test which does not require also the namespace `all` selector.

I am referring to `tests/expr_and_series/any_all_test.py`.
Do you mind editing such test to manually select columns with `nw.col("a", "b", "c").all()` instead of using `nw.all().all()`?

narwhals/_dask/expr.py
mistShard: After making the adjustments to the test as you described, I ran the tests and received the error:
`FAILED tests/expr_and_series/any_all_test.py::test_any_all[dask_lazy_constructor] - AttributeError: 'Scalar' object has no attribute 'name'`

I had thought it may have something to do with the fact that `Expr.all()` might not work for series, but on viewing re-reading the error I don't think that is the case. Please, what am I missing? I do not understand which object it is referring to as a `scalar` object.

FBruzzesi: Oh I see, this is related to #694 and Marco is working on a partial fix in #723. 

The scalar is a [Dask native Scalar object](https://github.com/dask/dask-expr/blob/1c646712b3c74eb9bed52bec59e442ce24d165c8/dask_expr/_collection.py#L4761). 

Maybe this needs to wait until #723 is merged (or we need to find another way of testing, but I would not suggest it)

FBruzzesi: @mistShard heads up #723 has been merged! By introducing `returns_scalar=True` you should be able to make this work 🙌🏼

mistShard:
@FBruzzesi `any_all_test.py` now runs successfully without xfails.

tests/expr_and_series/abs_test.py
```diff
@@ -1,14 +1,10 @@
...
-import pytest
-
 import narwhals.stable.v1 as nw
 from tests.utils import compare_dicts
 
 
-def test_abs(constructor: Any, request: Any) -> None:
-    if "dask" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
+def test_abs(constructor: Any) -> None:
```
FBruzzesi, line 7: If you roll back these changes, as they are addressed in the other PR, all CI should get green and we will be able to merge

FBruzzesi (commented):
Thanks for updating! I left a comment to make sure everything is successful in CI

tests/expr_and_series/abs_test.py
MarcoGorelli: yup - or alternatively, fetch upstream and rebase against `upstream/main`

PR #743, opened 2024-08-08
feat: "carefully" allow for dask Expr that modify index

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Pretty dangerous stuff to workaround the dask index.

To assess that the implementation is working as expected, I implemented both `sort` (different index but same length) and `drop_nulls` (different index due to different length)


tests/expr_and_series/drop_nulls_test.py
```diff
@@ -7,52 +7,60 @@
...
-        "B": [5, 6, 7, 8],
-        "C": [None, None, None, None],
-        "D": [9, 10, 11, 12],
-    }
+def test_drop_nulls_broadcast(constructor: Any, request: Any) -> None:
+    if "dask" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.col("a").drop_nulls(), nw.col("d").drop_nulls())
+    expected = {"a": [1.0, 2.0], "d": [6, 6]}
```
FBruzzesi, line 42: Sadly this broadcast is not working as drop_nulls does not return a scalar. I would consider this an edge case and focus on the broader support

anopsy:
image

MarcoGorelli:
thanks for trying this - i'll test it out and see if there's a perf impact

narwhals/_dask/dataframe.py
```diff
@@ -106,8 +109,21 @@ def select(
...
             return self._from_native_dataframe(df)
 
-        df = self._native_dataframe.assign(**new_series).loc[:, list(new_series.keys())]
-        return self._from_native_dataframe(df)
+        pd = get_pandas()
+        de = get_dask_expr()
+
+        col_order = list(new_series.keys())
+
+        left_most_series = next(  # pragma: no cover
```
FBruzzesi, line 117: this is guaranteed to not end up in `StopIteration` error as if everything was a scalar the previous block would have been entered and returned

MarcoGorelli:
we've got the notebooks in `tpch/notebooks`, the first two support Dask - fancy running them with this branch and seeing if there's any perf impact?

FBruzzesi:
Hey @MarcoGorelli, I am giving another thought on this feature (which I would still love to see), here is a simple idea to have partial support without loss of performance:

- `sort` is the only method that changes the index and result in an output with the same length. Instead of changing the index to each series, we can do that specifically in `sort`, namely:
   ```py
   def sort(self: Self, *, descending: bool = False, nulls_last: bool = False) -> Self:
        na_position = "last" if nulls_last else "first"

        def func(_input: Any, ascending: bool, na_position: bool) -> Any:  # noqa: FBT001
            name = _input.name

            result =_input.to_frame(name=name).sort_values(
                by=name, ascending=ascending, na_position=na_position
            )[name]
            return de._expr.AssignIndex(result, _input.index)

        return self._from_call(
            func,
            "sort",
            not descending,
            na_position,
            returns_scalar=False,
        )
    ```
- All the other methods that change the index, do so by reducing the length of the series. In my working experience and in TPCH queries they are mostly used before a reduction or in isolation, therefore we should not worry of changing their index
- What is left and unsupported you may ask? Multiple reductions ending up with the same length won't be possible. Example:
    ```py
    df.select(
        head=pl.col("a").head(),
        tail=pl.col("a").tail()
    )
    ```

What do you think?

narwhals/_dask/dataframe.py
```diff
@@ -96,17 +103,25 @@ def select(
...
             # This is a simple slice => fastpath!
             return self._from_native_frame(self._native_frame.loc[:, exprs])
 
+        n_modifies_index = sum(
+            getattr(e, "_modifies_index", 0)
+            for e in list(exprs) + list(named_exprs.values())
+        )
+        if n_modifies_index > 1:
+            msg = "Found multiple expressions that modify the index"
+            raise ValueError(msg)
```
FBruzzesi, line 119: TODO: Logic should be more complex:

- If total number of exprs is 1, then we can allow to modify the index
- If more than 1 exprs are present, then this should be banned as in `with_columns`

FBruzzesi:
@MarcoGorelli I am tagging this as ready for review as I re-worked it a bit more.

The TL;DR is:

- `sort` is kind of special, as it modifies the index but returns a Series of the same length of the original one, therefore in such specific case I am manually re-assigning the index
- for all other methods, I added a boolean flag to `DaskExpr` called `modifies_index` and:
  - that is not allowed in `with_columns`
  - in `select` it should be allowed only if there are no other exprs or there is a reduction following (I need to address both these cases actually).

Yet before developing further, I would like some feedback on how likable this approach is and if we want to move forward with it 🙏🏼

narwhals/_dask/expr.py
```diff
@@ -641,9 +674,65 @@ def func(df: DaskLazyFrame) -> list[Any]:
...
+            lambda _input, _n: _input.head(_n, compute=False),
+            "head",
+            n,
+            returns_scalar=False,
+            modifies_index=True,
+        )
+
+    def tail(self: Self, n: int) -> Self:
+        return self._from_call(
+            lambda _input, _n: _input.tail(_n, compute=False),
```
FBruzzesi, line 724: So, `head` has a `npartitions` param which can be set to -1 and scan all partitions, while `tail` does not.

MarcoGorelli:
thanks @FBruzzesi !

to be honest I don't know about using such private methods, it makes me feel slightly uneasy - @phofl do you have time/interest in taking a look? specifically the `de._collection.Series(de._expr.AssignIndex(result, _input.index))` part in `narwhals/_dask/expr.py`

I think that for sql engines (like duckdb, which hopefully we can get to eventually) operations like `df.select(nw.col('a').sort(), nw.col('b'))` would be problematic anyway, so I don't think it'd be an issue to leave them out of the Narwhals area of support

narwhals/_dask/dataframe.py
```diff
@@ -119,8 +132,15 @@ def select(
             )
             return self._from_native_frame(df)
 
-        df = self._native_frame.assign(**new_series).loc[:, list(new_series.keys())]
-        return self._from_native_frame(df)
+        col_order = list(new_series.keys())
+
+        left_most_series = next(  # pragma: no cover
+            s for s in new_series.values() if not isinstance(s, de._collection.Scalar)
```
phofl, line 138: note to self: we should expose this in Dask

narwhals/_dask/expr.py
phofl: I haven't seen cases where users actually want more than 1 partition if they call head or tail tbh, Yes this is technically an issue, but not something I've encountered in the wild

narwhals/_dask/expr.py
```diff
@@ -641,9 +674,65 @@ def func(df: DaskLazyFrame) -> list[Any]:
...
+
+        def func(_input: Any, ascending: bool, na_position: bool) -> Any:  # noqa: FBT001
+            import dask_expr as de  # ignore-banned-import
+
+            name = _input.name
+
+            result = _input.to_frame(name=name).sort_values(
+                by=name, ascending=ascending, na_position=na_position
+            )[name]
+            return de._collection.Series(de._expr.AssignIndex(result, _input.index))
```
phofl, line 692: Yeah I can second @MarcoGorelli here, please don't do this. We are sorting a Series (i.e. a single column from the df), correct?

I would:

```
tmp = _input.reset_index().sort_values(...)
result = tmp[_input.name]
result.index = tmp["use the index name of _input"]
```

phofl: Or do you want to keep the Index of the input?

If yes, this is fundamentally a bad idea in Dask, it will shoot you in the foot all over the place. You have zero guarantees that the partitions keep their lengths when sorting (it is a lot more likely that they do not), so this is bound to fail in all kinds of places

FBruzzesi: Hey @phofl thanks for taking the time.

> We are sorting a Series (i.e. a single column from the df), correct?

Yes indeed, but with the final goal to potentially add it as a new column to the original dataframe, and that's where the index misalignment gets into the game.

I will try the approach you are suggesting, which is not too far off from what is already implemented, and see if everything else falls into place 🙏🏼

FBruzzesi:
Hi @phofl, apologies to call you in the mix once more.

I have a few questions in order to make this work and guarantee that we don't end up with a

> fundamentally a bad idea in Dask, it will shoot you in the foot all over the place. 

- How can we test for when Dask will shoot us in the foot if we do something _bad_?
- The latest approach TL;DR is that if a method changes the index, then it either has to be followed by a reduction or be a single selection. Examples: 
    - Reductions:
      ```py
      df.select(
          head_sum=pl.col("a").head().sum(),
          tail_mean=pl.col("a").tail().mean(),
      )
      ```
      which would translate to something like `dd.concat([df["a"].head().sum(), df["a"].tail().mean()])`
     - Single selection:
       ```py
        df.select(
           head=pl.col("a").head(),
        )
        ```
   In sight of the first question, what do you think about this approach? Is it a fundamentally bad idea?

MarcoGorelli:
thanks for having explored this!

tbh i don't think we should do it - it adds complexity which we must then maintain, and then we'd need to do it for duckdb / pyspark / ... . I'd prefer to keep the lazy-only layer simpler

closing then, but thanks again for your investigation here! 🙏

FBruzzesi:
Thanks @MarcoGorelli! As you mentioned in the past, "no" might be temporary, we can always come back to this if needed

PR #750, opened 2024-08-08
feat: add  `str.replace` and  `str.replace_all`

lucianosrp:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 #372 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

In polars, `Expr.replace` and `Expr.replace_all` also accept and `Expr` for the pattern and value arguments. 
But this seems only to be true when using `pl.lit` as arguments.  
I just implemented the type `str` for this PR.

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -517,7 +517,7 @@ class DaskExprStringNamespace:
         *,
         literal: bool = False,
     ) -> DaskExpr:
-        return self._expr._from_call(  # pragma: no cover , coverage bug?
+        return self._expr._from_call(
             lambda _input, _pattern, _value, _literal: _input.str.replace(
                 _pattern, _value, n=-1, regex=not literal
             ),
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -516,9 +516,17 @@ class DaskExprStringNamespace:
         value: str,
         *,
         literal: bool = False,
-        n: int = 1,
     ) -> DaskExpr:
-        return self.replace(pattern=pattern, value=value, literal=literal, n=-1)
+        return self._expr._from_call(  # pragma: no cover , coverage bug?
+            lambda _input, _pattern, _value, _literal: _input.str.replace(
+                _pattern, _value, n=-1, regex=not literal
+            ),
+            "replace",
+            pattern,
+            value,
+            literal,
+            returns_scalar=False,
+        )
 
     def strip_chars(self, characters: str | None = None) -> DaskExpr:
         return self._expr._from_call(
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -490,6 +490,21 @@ class DaskExprStringNamespace:
     def __init__(self, expr: DaskExpr) -> None:
         self._expr = expr
 
+    def replace(
+        self, pattern: str, value: str, *, literal: bool = False, n: int = 1
+    ) -> DaskExpr:
+        return self._expr._from_call(
+            lambda _input, _pattern, _value, _literal, _n: _input.str.replace(
+                _pattern, _value, n=n, regex=not literal
+            ),
+            "replace",
+            pat=pattern,
+            repl=value,
+            n=n,
+            regex=not literal,
+            returns_scalar=False,
+        )
+
     def strip_chars(self, characters: str | None = None) -> DaskExpr:
         return self._expr._from_call(
             lambda _input, characters: _input.str.strip(characters),
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -500,13 +500,13 @@ class DaskExprStringNamespace:
     ) -> DaskExpr:
         return self._expr._from_call(
             lambda _input, _pattern, _value, _literal, _n: _input.str.replace(
-                _pattern, _value, n=n, regex=not literal
+                _pattern, _value, regex=not _literal, n=_n
             ),
             "replace",
             pattern,
             value,
-            n,
             literal,
+            n,
             returns_scalar=False,
         )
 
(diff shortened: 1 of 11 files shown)


narwhals/_dask/expr.py
```diff
@@ -490,6 +490,44 @@ class DaskExprStringNamespace:
...
+        )
+
+    def replace_all(
+        self,
+        pattern: str,
+        value: str,
+        *,
+        literal: bool = False,
+    ) -> DaskExpr:
+        return self._expr._from_call(  # pragma: no cover , coverage bug?
```
lucianosrp, line 520: coverage seems not to be hitting this line somehow?

narwhals/_pandas_like/expr.py
```diff
@@ -371,6 +371,29 @@ class PandasLikeExprStringNamespace:
...
+        )
+
+    def replace_all(
+        self,
+        pattern: str,
+        value: str,
+        *,
+        literal: bool = False,
+    ) -> PandasLikeExpr:
+        return reuse_series_namespace_implementation(  # pragma: no cover , coverage bug?
```
lucianosrp, line 393: same for this line

tests/utils.py
```diff
@@ -20,7 +20,10 @@ def zip_strict(left: Sequence[Any], right: Sequence[Any]) -> Iterator[Any]:
 def compare_dicts(result: Any, expected: dict[str, Any]) -> None:
     if hasattr(result, "collect"):
         result = result.collect()
-    if hasattr(result, "columns"):
+    if hasattr(result, "column_names"):  # To handle Pyarrow ChuckedArray:
```
lucianosrp, line 23: `result.columns` returned a `ChuckedArray` for this one which was throwing me an error.

narwhals/_dask/expr.py
FBruzzesi: It should be because in `narwhals.Expr`, `replace_all` is defined by calling `replace`, hence it doesn't trigger the call to the compliant `replace_all` but to the compliante `replace`

lucianosrp: aahh, of course!! Should I make it call `replace_all` just for consistency? even though it is just one paramter change.

FBruzzesi: In my opinion yes: I always hope that under the hood there is some optimization going on for (at least some) backends

tests/utils.py
FBruzzesi: I think the idea is to compare with narwhals dataframe directly, not converting back to native :)

tests/expr_and_series/str/replace_test.py
```diff
@@ -0,0 +1,130 @@
...
+    expected: dict[str, list[str]],
+) -> None:
+    df = nw.from_native(constructor(data))
+
+    result_df = nw.to_native(
+        df.select(
+            nw.col("a").str.replace_all(pattern=pattern, value=value, literal=literal)
+        )
+    )
+    compare_dicts(result_df, expected)
```
FBruzzesi, line 130: Following the comment on utils, this would be:

```suggestion
    result= df.select(
            nw.col("a").str.replace_all(pattern=pattern, value=value, literal=literal)
        )
    compare_dicts(result, expected)
```

narwhals/_dask/expr.py
```diff
@@ -490,6 +490,44 @@ class DaskExprStringNamespace:
...
+        return self._expr._from_call(
+            lambda _input, _pattern, _value, _literal, _n: _input.str.replace(
+                _pattern, _value, n=n, regex=not literal
+            ),
+            "replace",
+            pattern,
+            value,
+            n,
+            literal,
+            returns_scalar=False,
```
FBruzzesi, line 510: As a general principle, we are not binding the arguments in the lambda, but passing those over:

```suggestion
            lambda _input, _pattern, _value, _literal, _n: _input.str.replace(
                _pattern, _value, regex=not _literal, n=_n
            ),
            "replace",
            pattern,
            value,
            literal,
            n,
            returns_scalar=False,
```

I caught this because the order of the args below were different from the lambda signature yet tests didn't fail

narwhals/_dask/expr.py
```diff
@@ -490,6 +490,44 @@ class DaskExprStringNamespace:
...
+        literal: bool = False,
+    ) -> DaskExpr:
+        return self._expr._from_call(
+            lambda _input, _pattern, _value, _literal: _input.str.replace(
+                _pattern, _value, n=-1, regex=not literal
+            ),
+            "replace",
+            pattern,
+            value,
+            literal,
```
FBruzzesi, line 527: Same as above:

```suggestion
            lambda _input, _pattern, _value, _literal: _input.str.replace(
                _pattern, _value, n=-1, regex=not _literal
            ),
            "replace",
            pattern,
            value,
            literal,
```

FBruzzesi (commented):
One final concern to address 😇

narwhals/_dask/expr.py
lucianosrp: Thaks for checking this 🙏

FBruzzesi (approved):
Thanks for all the quick adjustments 👌

HiThereGIF

MarcoGorelli:
nice one! this is probably fine (i'll do another check in the morning, then will merge)

MarcoGorelli (approved):
thanks @lucianosrp , and thanks @FBruzzesi for reviewing!

lucianosrp:
aah more gifs ! 🤣
Thanks @FBruzzesi for the attentive review and comments 🙏

PR #760, opened 2024-08-09
feat: allow `from_dict` to not have native_namespace arg if all inputs are already narwhals Series

### LINKED ISSUE #756: Don't require passing native_namespace to `nw.from_dict` if inputs are all nw.Series already
The following should be supported:
```
data: dict[str, nw.Series]
nw.from_dict(data)
```

MarcoGorelli:
closes #756 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


MarcoGorelli:
@FBruzzesi fancy taking a look if you have interest/time?

narwhals/functions.py
```diff
@@ -97,6 +98,19 @@ def from_dict(
...
+    if not data:
+        msg = "from_dict cannot be called with empty dictionary"
+        raise ValueError(msg)
+    if native_namespace is None:
+        val = next(iter(data.values()))
+        if not isinstance(val, Series):
+            msg = "Calling `from_dict` without `native_namespace` is only supported if all input values are already Narwhals Series"
+            raise TypeError(msg)
+        native_namespace = val.__native_namespace__()
+        data = {key: to_native(value) for key, value in data.items()}
```
FBruzzesi, line 113: Only addition I would consider is:

- either a test to break if values have different backends
- explicitly raise if that's the case

narwhals/functions.py
```diff
@@ -52,7 +52,7 @@ def from_dict(
     data: dict[str, Any],
     schema: dict[str, DType] | Schema | None = None,
     *,
-    native_namespace: ModuleType,
+    native_namespace: ModuleType | None = None,
```
FBruzzesi, line 55: I am somewhat scared to provide a default, but also it should break if incompatible, so maybe no harm in that

FBruzzesi (commented):
Left a couple of consideration on "safety". 

Everything else looks good :) 

KittenKittyGIF

narwhals/functions.py
MarcoGorelli: why scared? would you prefer if people pass `nw.from_dict(data, native_namespace=None)` explicitly?

FBruzzesi: Fair point 👌

MarcoGorelli:
merging for now so we can release, and as far as I can tell your comment can be addressed as a follow-up if necessary

thanks for your review! 🙏

PR #762, opened 2024-08-09
feat: dask expr `len` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/tests/expr_and_series/len_test.py
+++ b/tests/expr_and_series/len_test.py
@@ -6,7 +6,18 @@ import narwhals.stable.v1 as nw
 from tests.utils import compare_dicts
 
 
-def test_len(constructor: Any, request: Any) -> None:
+def test_len_no_filter(constructor: Any) -> None:
+    data = {"a": list("xyz"), "b": [1, 2, 1]}
+    expected = {"l": [3], "l2": [6]}
+    df = nw.from_native(constructor(data)).select(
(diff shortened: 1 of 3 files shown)


tests/expr_and_series/len_test.py
```diff
@@ -6,7 +6,18 @@
...
+    expected = {"l": [3], "l2": [6]}
+    df = nw.from_native(constructor(data)).select(
+        nw.col("a").len().alias("l"),
+        (nw.col("a").len() * 2).alias("l2"),
+    )
+
+    compare_dicts(df, expected)
+
+
+def test_len_len_chaining(constructor: Any, request: Any) -> None:
```
FBruzzesi, line 20: Renamed this test, and keep failing for dask as `.filter` is not implemented. Wondering if that's ok to implement even in the sad case that we do not move forward with #743 . 

At the end, I think `expr.filter` is mainly use to filter within a context and then aggregate (exactly as in the test here)

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #764, opened 2024-08-09
feat: dask expr `is_first|last_distinct` methods

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Dask does not implement Series/Expr `duplicated` as pandas. 
This implementation is very much inspired from the work on the pyarrow backend - I assume not very cheap though


narwhals/_dask/expr.py
```diff
@@ -473,6 +473,50 @@ def clip(
...
+    def is_first_distinct(self: Self) -> Self:
+        def func(_input: Any) -> Any:
+            _name = _input.name
+            col_token = generate_unique_token(n_bytes=8, columns=[_name])
+            _input = (
+                _input.to_frame()
+                .assign(**{col_token: 1})
+                .assign(
+                    **{col_token: lambda t: t[col_token].cumsum(method="blelloch") - 1}
+                )
```
MarcoGorelli, line 485: is there a way to do this without passing a lambda to dask? been told it blocks some optimisations

MarcoGorelli (commented):
looks good, just a comment (which might be applicable in other places)

narwhals/_dask/expr.py
FBruzzesi: I will look into it, thanks for the heads up!

MarcoGorelli: thanks!

this is what I meant: https://github.com/narwhals-dev/narwhals/pull/765

It's fine if we have some lambdas on the Narwhals side which we evaluate before passing the result to Dask. But the input to Dask should ideally not be a lambda

FBruzzesi: Added and factored it out (3 times is enough to make a tiny function)

MarcoGorelli (approved):
legeeeend

PR #766, opened 2024-08-10
chore: tests series_only refactor

### LINKED ISSUE #401: clean: factor out tests from `test_common.py`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #401 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Refactor a bunch of code both out of `tests/series_only/test_common.py` and from `tests/series_only/*` to `tests/expr_and_series/*` whenever needed


narwhals/_arrow/series.py
```diff
@@ -185,7 +185,7 @@ def __rtruediv__(self, other: Any) -> Self:
         pc = get_pyarrow_compute()
         ser = self._native_series
         other = validate_column_comparand(other)
-        if not isinstance(other, (pa.Array, pa.ChunkedArray)):
+        if not isinstance(other, (pa.Array, pa.ChunkedArray)):  # pragma: no cover
```
MarcoGorelli, line 188: thanks for doing this!

just a question about this line, why is it no longer covered?

MarcoGorelli (approved):
nice, this is really pleasing 😋

Issue #742, opened 2024-08-08
[Enh]: Extend signature of supported methods


FBruzzesi:
### Please describe the purpose of the new feature or describe the problem to solve.

There are some methods for which we have partial support, meaning that not the full polars api is supported.

However this should be possible for some methods and/or some arguments (examples below)

### Additional information that may help us understand your needs.

Here a (partial) list of methods and arguments missing:

| method | missing argument(s) | note |
| --| --| -- |
| `DataFrame.drop` | `strict` | ✅ Fixed in #773 |
| `DataFrame.drop_nulls` | `subset` | ✅ Fixed in #779 |
| `DataFrame.join` | `suffix`| ✅ |
| `DataFrame.join_asof` | `suffix`, `tolerance` | |

MarcoGorelli:
cool, happy to expand where it makes sense - if it's possible to get support across pandas/polars/pyarrow then it's probably already worth doing!

FBruzzesi:
Closing this for now, we can come back to `join_asof` and other methods when and if required for adoption

PR #773, opened 2024-08-10
feat: extend dataframe `drop` method

FBruzzesi:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #742 

## Checklist

- [x] Tests added 
- [x] Documented the changes


narwhals/_arrow/dataframe.py
```diff
@@ -285,8 +286,18 @@ def join(
...
-    def drop(self, *columns: str) -> Self:
-        return self._from_native_dataframe(self._native_dataframe.drop(list(columns)))
+    def drop(self: Self, *columns: str, strict: bool = True) -> Self:
+        cols = set(self.columns)
+        to_drop = list(columns)
+        if strict:
+            for d in to_drop:
+                if d not in cols:
+                    msg = f'"{d}" not found'
+                    raise ColumnNotFoundError(msg)
```
FBruzzesi, line 296: The logic is same everywhere. I wonder if we should move it directly to `BaseFrame`.

Also, I noticed that polars is "greedy" in the raise, meaning that if 2 columns are in the drop list but missing, the error will only mention the first one

narwhals/_arrow/dataframe.py
```diff
@@ -285,8 +286,18 @@ def join(
             ),
         )
 
-    def drop(self, *columns: str) -> Self:
-        return self._from_native_dataframe(self._native_dataframe.drop(list(columns)))
+    def drop(self: Self, *columns: str, strict: bool = True) -> Self:
```
FBruzzesi, line 289: Ok hear me out, I know that we don't add defaults in here, but we _sometimes_ used compliant dataframe drop method (e.g. in `join`). I can add it in the call there

narwhals/_arrow/dataframe.py
MarcoGorelli: I think Polars changed behaviour here in 1.0 too (before it would just ignore missing columns)

so, fine by me to just implement it in BaseFrame and let everyone benefit

FBruzzesi: Cool, I just noticed CI failing for polars pre 1. Refactoring into base now :)

FBruzzesi: Somehow I keep f*cking up parsing `*columns` - it's probably just late 🙈

FBruzzesi: Ok it is somehow fixed but worth taking a double look at it as the expected types along the chain of calls are a bit different

FBruzzesi:
Aside: I noticed that polars allows selectors other than column names. I think it should not be too complex to integrate. 
Thoughts?

narwhals/_pandas_like/dataframe.py
```diff
@@ -318,10 +318,8 @@ def with_columns(
     def rename(self, mapping: dict[str, str]) -> Self:
         return self._from_native_dataframe(self._native_dataframe.rename(columns=mapping))
 
-    def drop(self, *columns: str) -> Self:
-        return self._from_native_dataframe(
-            self._native_dataframe.drop(columns=list(columns))
-        )
+    def drop(self: Self, columns: str | list[str]) -> Self:
```
MarcoGorelli, line 321: does it need to be `str | list[str]`? didn't it get flattened one level above?

FBruzzesi: I will check with mypy for internal use

MarcoGorelli:
> Aside: I noticed that polars allows selectors other than column names. I think it should not be too complex to integrate.
Thoughts?

if it's not too complex - sure, I do like these 👍

narwhals/_arrow/dataframe.py
```diff
@@ -266,14 +266,16 @@ def join(
...
+                ._native_dataframe.join(
                     other.with_columns(**{key_token: plx.lit(0, None)})._native_dataframe,
                     keys=key_token,
                     right_keys=key_token,
                     join_type="inner",
                     right_suffix="_right",
-                ),
-            ).drop(key_token)
+                )
+                .drop(key_token),
```
FBruzzesi, line 277: Avoids a few functions and attributes calls

FBruzzesi:
> > Aside: I noticed that polars allows selectors other than column names. I think it should not be too complex to integrate.
> > Thoughts?
> 
> if it's not too complex - sure, I do like these 👍

Happy to keep it as a follow up honestly. Other methods would benefit from that as well and it could be worth to factor out some sort of "parser" for selectors and column names

narwhals/dataframe.py
```diff
@@ -137,10 +138,18 @@ def head(self, n: int) -> Self:
     def tail(self, n: int) -> Self:
         return self._from_compliant_dataframe(self._compliant_frame.tail(n))
 
-    def drop(self, *columns: str | Iterable[str]) -> Self:
-        return self._from_compliant_dataframe(
-            self._compliant_frame.drop(*flatten(columns))
-        )
+    def drop(self, *columns: Iterable[str], strict: bool) -> Self:
+        cols = set(self.collect_schema().names())
```
MarcoGorelli, line 142: this is the part I'm a little unsure about

I'm not sure we should be triggering `collect_schema` for Polars LazyFrame - I'll check what they're doing

MarcoGorelli (commented):
looks good, just got a comment on the polars lazyframe case, will need to think about it

narwhals/dataframe.py
FBruzzesi: We can let polars do what polars does natively.
I can factor out the parsing into a utils function and call it downstream instead of in `BaseFrame`.

MarcoGorelli: thanks! sure, but at the moment we still reach this `collect_schema` call for Polars, right? that's the part i think we should avoid

FBruzzesi: yes at the moment we do. I will rework this PR a bit to avoid that in the versions that polars natively supports `.drop(..., strict=...)`

MarcoGorelli: maybe we can just avoid it completely for Polars, and note in the docstring that for Polars<1 dropping non-existent columns silently passes?

FBruzzesi: Whops! I was doing the refactor and just read this.

In polars pre v1 passing a column not present in the data just ends up doing nothing. I wonder how v1 does it as it has to know at least the column names, so I would expect implicitly to collect those?!

MarcoGorelli: > so I would expect implicitly to collect those?!

yup, but only a `LazyFrame.collect` time, not at `LazyFrame.drop` time
```
In [8]: df = pl.LazyFrame({'a': [1,2,3]})

In [9]: df.drop('b')
Out[9]: <LazyFrame at 0x7FE016F33510>

In [10]: df.drop('b').collect()
ColumnNotFoundError                       Traceback (most recent call last)
Cell In[10], line 1
----> 1 df.drop('b').collect()

File ~/scratch/.venv/lib/python3.11/site-packages/polars/lazyframe/frame.py:1942, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, no_optimization, streaming, background, _eager, **_kwargs)
   1939 # Only for testing purposes atm.
   1940 callback = _kwargs.get("post_opt_callback")
-> 1942 return wrap_df(ldf.collect(callback))

ColumnNotFoundError: b
```

FBruzzesi: Just to make sure to be on the same page before making other changes:

- Eager and lazy post v1 we let polars do its thing
- Eager pre v1, we can explicitly check column names as per other backend
- Lazy pre v1, ignore the strict argument and raise a warning saying something along the line of "please go eager here" if passed as True?

MarcoGorelli: thanks for checking, i'd say:


- _Eager and lazy post v1 we let polars do its thing_: agree
- _Eager pre v1, we can explicitly check column names as per other backend_ agree
- _Lazy pre v1, ignore the strict argument and raise a warning saying something along the line of "please go eager here" if passed as True?_ I was thinking more: just let Polars do its thinking, but leave a note in the docs about the pre-v1 behaviour difference. So, a docs note, not a runtime warning (which might be what you meant, just emphasising to be sure)

FBruzzesi: I meant a runtime warning, especially because `strict=True` is the default, and that's the thing we ignore

MarcoGorelli: Reckon a runtime warning risks being annoying, as it may be unnecessary? And going eager might represent a significant performance degradation?

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #780, opened 2024-08-12
chore: rename internal dtype functions

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -30,7 +30,7 @@ jobs:
   publish-to-pypi:
     name: >-
       Publish Python 🐍 distribution 📦 to PyPI
-    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    if: startsWith(github.ref, 'refs/tags/v')  # only publish to PyPI on tag pushes
     needs:
     - build
     runs-on: ubuntu-latest
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -1,9 +1,6 @@
 from __future__ import annotations
 
 from typing import TYPE_CHECKING
-from typing import Any
-
-from narwhals.utils import isinstance_or_issubclass
 
 if TYPE_CHECKING:
     from typing_extensions import Self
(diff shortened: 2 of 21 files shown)


.github/workflows/publish_to_pypi.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   publish-to-pypi:
     name: >-
       Publish Python 🐍 distribution 📦 to PyPI
-    if: startsWith(github.ref, 'refs/tags/')  # only publish to PyPI on tag pushes
+    if: startsWith(github.ref, 'refs/tags/v')  # only publish to PyPI on tag pushes
```
MarcoGorelli, line 33: drive-by - tags should start with `v` if we're releasing

narwhals/dtypes.py
```diff
@@ -85,51 +84,3 @@ class Enum(DType): ...
 
 
 class Date(TemporalType): ...
-
-
-def translate_dtype(plx: Any, dtype: DType) -> Any:
```
MarcoGorelli, line 90: turns out we didn't need this function 😄 😭

FBruzzesi:
That's a nice cleanup indeed! How come that namespaces are not covered now? Aren't we using them somewhere? Or maybe we initialize them directly in other methods without passing through `__narwhals_namespace__`

MarcoGorelli:
thanks!

> Or maybe we initialize them directly in other methods without passing through __narwhals_namespace__

i've done this in a couple of other places, but I think `Series.__narwhals_namespace__` is unreachable - might as well getrid of it for now then? can always add it back later if it helps somewhere

FBruzzesi:
> i've done this in a couple of other places

Thanks! I think it makes complete sense!

> but I think `Series.__narwhals_namespace__` is unreachable - might as well getrid of it for now then? can always add it back later if it helps somewhere

I am very indifferent if keeping with no cover or removing it 😂

MarcoGorelli:
thanks for your review!

PR #783, opened 2024-08-13
feat: improve typing for `to_pandas` / `to_numpy` / `to_arrow`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import contextlib
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
@@ -12,7 +13,6 @@ from narwhals._arrow.utils import floordiv_compat
 from narwhals._arrow.utils import narwhals_to_native_dtype
 from narwhals._arrow.utils import translate_dtype
 from narwhals._arrow.utils import validate_column_comparand
-from narwhals.dependencies import get_numpy
 from narwhals.dependencies import get_pandas
 from narwhals.dependencies import get_pyarrow
 from narwhals.dependencies import get_pyarrow_compute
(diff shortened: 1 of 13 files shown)


narwhals/dataframe.py
```diff
@@ -21,6 +21,8 @@
     from pathlib import Path
 
     import numpy as np
+    import pandas as pd
+    import pyarrow as pa
```
FBruzzesi, line 25: How would this work if I am in a dev env without one of these libraries?

MarcoGorelli: IIUC it would just be reported as `Any`

narwhals/_arrow/series.py
```diff
@@ -393,7 +392,8 @@ def is_in(self, other: Any) -> Self:
         return self._from_native_series(pc.is_in(ser, value_set=value_set))
 
     def arg_true(self) -> Self:
-        np = get_numpy()
+        import numpy as np  # ignore-banned-import
```
MarcoGorelli, line 395: currently, `import pyarrow` also imports NumPy

but I don't know if that's something we can rely on never changing in the future

so I think we should explicitly import numpy here

i'll add some tests regarding imports, that's probably better than relying on these `get_` functions

MarcoGorelli:
merging for now so I can use `# ignore-banned-import` in the other pr

narwhals/dataframe.py
MarcoGorelli: @FBruzzesi , just to give a quick demo with the latest Narwhals installed in a new venv, in which I have Polars but not Modin:
```
import narwhals as nw
from narwhals.dependencies import is_modin_dataframe, is_polars_dataframe
from narwhals.typing import IntoDataFrame

def func(df: IntoDataFrame) -> None:
    if is_modin_dataframe(df):
        reveal_type(df)
    elif is_polars_dataframe(df):
        reveal_type(df)
```
outputs
```
$ mypy t.py 
t.py:7: note: Revealed type is "Any"
t.py:9: note: Revealed type is "polars.dataframe.frame.DataFrame"
```

PR #786, opened 2024-08-13
feat: Support Arrow PyCapsule Interface for export

### LINKED ISSUE #784: [Enh]: Pass through the Arrow PyCapsule Interface

The [Arrow PyCapsule Protocol](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) makes it possible to share Arrow data between libraries without requiring a dependency on pyarrow. Many of the DataFrame libraries that narwhals wraps already support it (See https://github.com/apache/arrow/issues/39195 for ecosystem summary), so my thinking is that it wouldn't be too much work to support the protocol on the Narwhals DataFrame itself.

Vega-Altair recently adopted Narwhals as a way to support polars without a pyarrow dependency, and I'm looking at how I could do the same in VegaFusion. The next VegaFusion release is going to support ingesting DataFrames that support the PyCapsule Interface (WIP by @kylebarron in https://github.com/vega/vegafusion/pull/501), so it would be nice to be able to pass the Narwhals DataFrame directly to VegaFusion and process it using this protocol.

_No response_

_No response_

_No response_

MarcoGorelli:
closes #784 

@kylebarron fancy taking a look to see if this is what needs doing / if I've understood the assignment?

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -7,6 +7,8 @@ from typing import Literal
 from typing import Sequence
 from typing import overload
 
+from narwhals.dependencies import get_pyarrow
+
 if TYPE_CHECKING:
     import numpy as np
     from typing_extensions import Self
@@ -59,14 +61,19 @@ class Series:
         """
         Export a Series via the Arrow PyCapsule Interface.
 
-        Narwhals doesn't implement anything itself here - if the underlying series
-        implements the interface, it'll return that, else you'll get an `AttributeError`.
+        Narwhals doesn't implement anything itself here:
+
+        - if the underlying series implements the interface, it'll return that
+        - else, it'll call `to_arrow` and then defer to PyArrow's implementation
 
         https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html
         """
-        return self._compliant_series._native_series.__arrow_c_stream__(
-            requested_schema=requested_schema
-        )
+        native_series = self._compliant_series._native_series
+        if hasattr(native_series, "__arrow_c_stream__"):
+            return native_series.__arrow_c_stream__(requested_schema=requested_schema)
+        pa = get_pyarrow()
+        ca = pa.chunked_array([self.to_arrow()])
+        return ca.__arrow_c_stream__(requested_schema=requested_schema)
 
     @property
     def shape(self) -> tuple[int]:
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -68,7 +68,8 @@ class Series:
         - if the underlying series implements the interface, it'll return that
         - else, it'll call `to_arrow` and then defer to PyArrow's implementation
 
-        https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html
+        See [PyCapsule Interface](https://arrow.apache.org/docs/dev/format/CDataInterface/PyCapsuleInterface.html)
+        for more.
         """
         native_series = self._compliant_series._native_series
         if hasattr(native_series, "__arrow_c_stream__"):
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -74,7 +74,7 @@ class Series:
         if hasattr(native_series, "__arrow_c_stream__"):
             return native_series.__arrow_c_stream__(requested_schema=requested_schema)
         try:
-            import pyarrow as pa
+            import pyarrow as pa  # ignore-banned-import
         except ModuleNotFoundError as exc:  # pragma: no cover
             msg = f"PyArrow>=14.0.0 is required for `__arrow_c_stream__` for object of type {type(native_series)}"
             raise ModuleNotFoundError(msg) from exc
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -76,10 +76,10 @@ class Series:
         try:
             import pyarrow as pa  # ignore-banned-import
         except ModuleNotFoundError as exc:  # pragma: no cover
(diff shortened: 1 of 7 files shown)


kylebarron (approved):
Yeah I think that makes sense! There's not much you can do if the underlying dataframe doesn't support it. If you had some way to access an arrow table from the underlying dataframe you could do something more, but I think this is good

MarcoGorelli:
cool, thanks!

> If you had some way to access an arrow table from the underlying dataframe

we do have `narwhals.DataFrame.to_arrow`, which returns a pyarrow table - is that what you meant? If so, what else would you suggest adding?

kylebarron:
In that case I'd suggest
```py
def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
    try:
        return self._compliant_frame._native_frame.__arrow_c_stream__(
            requested_schema=requested_schema
        )
    except AttributeError:
        pyarrow_table = self.to_arrow()
        return pyarrow_table.__arrow_c_stream__(requested_schema=requested_schema)
```
So in the first case, we can use the source's implementation of the pycapsule interface and pyarrow doesn't need to be installed, while in the second case we can still ensure the method never raises an exception.

MarcoGorelli:
thanks! though I think the user could call this themselves? e.g.
```python
df: nw.DataFrame
try:
    result = df.__arrow_c_stream__(requested_schema=requested_schema)
except AttributeError:
    result = df.to_arrow().__arrow_c_stream__(requested_schema=requested_schema)
```
?
I think that'd be more explicit, not totally sure we should be calling `to_arrow` on behalf of the user. By 'user', I mean the library developer using Narwhals, I'd be inclined to leave it up to them whether or not to fallback to PyArrow

kylebarron:
Well, one of my primary arguments for the pycapsule interface is that it allows an ecosystem of data producers and consumers to interoperate without any knowledge of the other, solely by looking for an `__arrow_c_stream__` dunder method. Calling `.to_arrow()` would indeed be more explicit, but it would require library consumers to _know about narwhals_, which I'd argue in the general case is not true. E.g. `pyarrow.table()` only knows to check for `__arrow_c_stream__`.

MarcoGorelli:
I was thinking more of the Vegafusion case - I think it's better for them if they choose to explicitly call `to_arrow`. Otherwise they might be calling `__arrow_c_stream__` all over the places, thinking it's cheap, whereas it would've been better to do a single `to_arrow` upfront 😇 

I think any library developer using Narwhals (e.g. vegafusion) would and should know about Narwhals, whereas lower-level libraries like PyArrow shouldn't:
- PyArrow can just check for `__arrow_c_stream__` (as it currently does)
- Vegafusion can choose whether and when to call `to_arrow` before accessing `__arrow_c_stream__`

MarcoGorelli:
In any case, as they often say, "in open source, 'no' is temporary but 'yes' is forever" - doubly so with our stable api policy 😆 

So, as the current implementation looks good to you, I'd say - let's start with that, we can always loosen it later if necessary

Thanks for your review and input, much appreciated 🙏 !

kylebarron:
> Otherwise they might be calling `__arrow_c_stream__` all over the places, thinking it's cheap, whereas it would've been better to do a single `to_arrow` upfront

As a general note, consumers can't do this because consumers don't have a way to know whether the source is a _table_ that already exists in memory or whether it's a _stream_ that can only be called once. E.g. a pyarrow `RecordBatchReader` is a stream and you can only call `__arrow_c_stream__` once. So for a consumer like vegafusion, it would be important for it to import all the data once and then operate as it needs to on it.

MarcoGorelli:
Ah nice, thanks for explaining!

In that case, I'm leaning more towards your suggestion - it would also mean being able to support this for versions of pandas prior to 2.2 but which support converting to pyarrow table

narwhals/series.py
```diff
@@ -71,7 +71,14 @@ def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
...
-        pa = get_pyarrow()
+        try:
+            import pyarrow as pa
+        except ModuleNotFoundError as exc:  # pragma: no cover
+            msg = f"PyArrow>=14.0.0 is required for `__arrow_c_stream__` for object of type {type(native_series)}"
+            raise ModuleNotFoundError(msg) from exc
+        if parse_version(pa.__version__) < (14, 0):  # pragma: no cover
+            msg = f"PyArrow>=14.0.0 is required for `__arrow_c_stream__` for object of type {type(native_series)}"
+            raise ModuleNotFoundError(msg)
         ca = pa.chunked_array([self.to_arrow()])
```
kylebarron, line 82: I think this might require pyarrow 15

MarcoGorelli: in pandas the requirement is PyArrow 14+ (I also just ran the tests with pyarrow 13 and 14 - the former fails, the latter passes)

MarcoGorelli: ah sorry, that's for DataFrame

kylebarron: It was added to pa.chunked_array in a later release, yes. I think it was here: https://github.com/apache/arrow/pull/40818

MarcoGorelli:
thanks Kyle for your help!

cool, let's ship this 🚢

PR #788, opened 2024-08-13
chore: import overhaul

MarcoGorelli:
I'd originally introduced these `get_` functions (`get_numpy`, `get_pandas`, etc...) to make sure we didn't import anything which the user hadn't already imported

This is fine for short-circuiting `isinstance` checks. However, we're also importing 3rd party libraries within functions, not just for `isinstance` checks, and my concern is that we're assuming that some imports are automatically available, but that this might not always be the case in the future. For example, currently, `import pyarrow` automatically imports `numpy` as well, and so `np = get_numpy()` works without issues. But I don't know if we can assume that that'll always be the case, so we may need to import numpy explicitly to be safe

So, I've added explicit `is_pandas_dataframe` / `is_polars_series` / etc. methods which we can use for `isinstance` checks, and in other places, we can just import the library we need. This does require a `#  ignore-banned-imports`, which I hope gets people to double-check what they're doing

I've also added tests to check that extra dependencies aren't imported

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/dependencies.py
+++ b/narwhals/dependencies.py
@@ -125,7 +125,7 @@ def is_pyarrow_chunked_array(ser: Any) -> TypeGuard[pa.ChunkedArray]:
 
 
 def is_pyarrow_table(df: Any) -> TypeGuard[pa.Table]:
-    """Check whether `ser` is a PyArrow Table without importing PyArrow."""
+    """Check whether `df` is a PyArrow Table without importing PyArrow."""
     return bool((pa := get_pyarrow()) is not None and isinstance(df, pa.Table))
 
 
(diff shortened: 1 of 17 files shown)


narwhals/dependencies.py
```diff
@@ -45,26 +49,6 @@ def get_pyarrow() -> Any:  # pragma: no cover
...
-    # TODO(marco): remove this one, as it's at odds with the others, as it imports
-    # something new
-    if "pyarrow" in sys.modules:
-        import pyarrow.compute as pc
-
-        return pc
-    return None
-
-
-def get_pyarrow_parquet() -> Any:  # pragma: no cover
```
MarcoGorelli, line 59: these weren't part of the public API, so we're still in time to remove them 😎

PR #789, opened 2024-08-14
feat: allow selectors in `drop` and `drop_nulls`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [ ] Documented the changes

Opening as draft because Dask is failing for selectors with the following error:

> E           KeyError: 'Cannot index with non-boolean dask Series. Try passing computed values instead (e.g. ``ddf.loc[iindexer.compute()]``)'

~and I am not sure if it is fixable 🙃~
Fixed ✅

### DIFF
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -130,7 +130,7 @@ class DaskLazyFrame:
 
         column_names = [c for c in subset_ if isinstance(c, str)]
         selector_names = self.select(
-            *[c.is_null()._call(plx) for c in subset_ if isinstance(c, Selector)]
+            *[c._call(plx) for c in subset_ if isinstance(c, Selector)]
         ).columns
 
         return self._from_native_frame(
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -25,6 +25,7 @@ if TYPE_CHECKING:
     from narwhals._dask.namespace import DaskNamespace
     from narwhals._dask.typing import IntoDaskExpr
     from narwhals.dtypes import DType
+    from narwhals.selectors import Selector
 
 
 class DaskLazyFrame:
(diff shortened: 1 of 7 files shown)


FBruzzesi:
Took me a freaking hour to figure out the issue with scikit-lego was that pandas can have columns with int names. That seems to not stuck into my mind 🙈

narwhals/_dask/dataframe.py
```diff
@@ -113,12 +115,27 @@ def select(
...
         plx = self.__narwhals_namespace__()
-        return self.filter(~plx.any_horizontal(plx.col(*subset).is_null()))
+        subset_: Iterable[str | Selector] = (
+            [subset] if isinstance(subset, (str, Selector)) else subset
+        )
+
+        column_names = [c for c in subset_ if isinstance(c, str)]
+        selector_names = self.select(
+            *[c.is_null()._call(plx) for c in subset_ if isinstance(c, Selector)]
+        ).columns
```
FBruzzesi, line 134: I am _giving for granted_ that the optimizer knows we are only getting the column names at the end

MarcoGorelli:
nice! will take a careful look, but on board with the idea, selectors are really ergonomic

narwhals/_pandas_like/dataframe.py
```diff
@@ -233,12 +236,28 @@ def select(
...
+        from narwhals.selectors import Selector
+
         plx = self.__narwhals_namespace__()
-        return self.filter(~plx.any_horizontal(plx.col(*subset).is_null()))
+        subset_: Iterable[str | Selector] = (
+            [subset] if isinstance(subset, (str, Selector)) else subset
+        )
+
+        column_names = [c for c in subset_ if isinstance(c, str)]
+        selectors = [c.is_null()._call(plx) for c in subset_ if isinstance(c, Selector)]
```
MarcoGorelli, line 253: do we need to call `is_null` here if we're not using these `selectors` in `any_horizontal`?

FBruzzesi: I am not sure I fully get the question, this part:

> if we're not using these selectors in any_horizontal?

how so?

My reasoning is as follows:

- If a selector is present, we evaluate it, else this will end up being an empty list anyway.
- `any_horizontal` is called only if at least one (evaluated) selector ends up being there

MarcoGorelli: oops sorry, I meant to comment this on the `_dask` one - if you only need to get the column names, out, does it work to just do `*[c._call(plx) for c in subset_ if isinstance(c, Selector)]` there?

FBruzzesi: Oh that seems significantly better! Let me give it a try

MarcoGorelli, line 253: at the moment non-string columns are supported here
```python
In [18]: @nw.narwhalify
    ...: def func(df):
    ...:     return df.drop_nulls(subset=[0])
    ...:

In [19]: func(pd.DataFrame({0: [1,2,None]}))
Out[19]:
     0
0  1.0
1  2.0
```
so if I'm reading this right we'd risk regressing

MarcoGorelli (commented):
thanks for your pr and for the effort you've put in here!

tbh the implementation does feel a bit complex - is there a use-case in scikit-lego that would need this? if not i feel a bit hesistant about complicating things to this level

FBruzzesi:
Hey Marco! 

> tbh the implementation does feel a bit complex - is there a use-case in scikit-lego that would need this?

I don't have a direct use case just yet, part of why I like polars so much is this ergonomic way of doing a bunch of stuff.

>  if not i feel a bit hesistant about complicating things to this level

That's alright, we don't need to merge any idea 😂 I am ok with giving it a try a see how complex it ends up being and decide whether or not we want to maintain it. We can always come back to it if it serves some actual use case

narwhals/_pandas_like/dataframe.py
FBruzzesi: I think I only maintained the support for dask, but that would be an easy fix as in `_dask/dataframe.py` line 154

PR #798, opened 2024-08-16
test: Improve `duration_attribute_tests`

luke396:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

- Add constructor_lazy and ensure it fails as expected.
- Combine two separate tests into one.

### DIFF
--- a/tests/expr_and_series/dt/datetime_duration_test.py
+++ b/tests/expr_and_series/dt/datetime_duration_test.py
@@ -40,9 +40,9 @@ def test_duration_attributes(
     request: Any,
     constructor: Any,
     attribute: str,
-    expected_a: list[int] | None,
-    expected_b: list[int] | None,
-    expected_c: list[int] | None,
+    expected_a: list[int],
+    expected_b: list[int],
+    expected_c: list[int],
 ) -> None:
     if "lazy" in str(constructor) or (
         parse_version(pd.__version__) < (2, 2) and "pandas_pyarrow" in str(constructor)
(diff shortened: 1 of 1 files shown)


tests/expr_and_series/dt/datetime_duration_test.py
```diff
@@ -23,79 +23,54 @@
         timedelta(milliseconds=1, microseconds=300),
     ],
     "c": np.array([None, 20], dtype="timedelta64[ns]"),
-    "d": [
```
luke396, line 26: This appears to be unused, so it is removed.

luke396:
The combination of two tests is somewhat aggressive; if it's not suitable, I would like to revert them.

FBruzzesi (commented):
Thanks for your PR, in particular for cleaning up dead code!

The idea of moving to `constructor` instead of `constructor_eager` is to be able to assess lazy  instances for expressions, but not for series.

If you check other tests that do this, they split into:

- one test for `constructor` where expression are tested
- one test for `constructor_eager` where series are tested

[Example `.abs()`](https://github.com/narwhals-dev/narwhals/blob/main/tests/expr_and_series/abs_test.py).

I would suggest to follow a similar approach to make sure that polars lazyframe (and dask lazyframe in the future) can perform those operations.

MarcoGorelli (approved):
thanks @luke396 !

PR #799, opened 2024-08-16
chore: stable api `narwhalify` de-duplication

### LINKED ISSUE #469: [chore]: Simplify logic and deduplicate code in `narwhalify`

See this [snippet](https://github.com/narwhals-dev/narwhals/blob/c904b102bd95dec026ce756cecfda15dabe71f41/narwhals/stable/v1.py#L639-L746)

```python
def narwhalify(
    func: Callable[..., Any] | None = None,
    *,
    strict: bool = False,
    eager_only: bool | None = False,
    series_only: bool | None = False,
    allow_series: bool | None = True,
) -> Callable[..., Any]:
    """
    Decorate function so it becomes dataframe-agnostic.

    `narwhalify` will try to convert any dataframe/series-like object into the narwhal
    respective DataFrame/Series, while leaving the other parameters as they are.

    Similarly, if the output of the function is a narwhals DataFrame or Series, it will be
    converted back to the original dataframe/series type, while if the output is another
    type it will be left as is.

    By setting `strict=True`, then every input and every output will be required to be a
    dataframe/series-like object.

    Instead of writing

    ```python


    def func(df_any):
        df = nw.from_native(df_any, strict=False)
        df = df.group_by("a").agg(nw.col("b").sum())
        return nw.to_native(df)
    ```

    you can just write

    ```python


    @nw.narwhalify
    def func(df):
        return df.group_by("a").agg(nw.col("b").sum())
    ```

    You can also pass in extra arguments, e.g.

    ```python
    @nw.narhwalify(eager_only=True)
    ```

    that will get passed down to `nw.from_native`.

    Arguments:
        func: Function to wrap in a `from_native`-`to_native` block.
        strict: Whether to raise if object can't be converted or to just leave it as-is
            (default).
        eager_only: Whether to only allow eager objects.
        series_only: Whether to only allow series.
        allow_series: Whether to allow series (default is only dataframe / lazyframe).
    """

    # TODO(Unassigned): do we have a way to de-dupe this a bit?
    def decorator(func: Callable[..., Any]) -> Callable[..., Any]:
        @wraps(func)
        def wrapper(*args: Any, **kwargs: Any) -> Any:
            args = [
                from_native(
                    arg,
                    strict=strict,
                    eager_only=eager_only,
                    series_only=series_only,
                    allow_series=allow_series,
                )
                for arg in args
            ]  # type: ignore[assignment]

            kwargs = {
                name: from_native(
                    value,
                    strict=strict,
                    eager_only=eager_only,
                    series_only=series_only,
                    allow_series=allow_series,
                )
                for name, value in kwargs.items()
            }

            backends = {
                b()
                for v in [*args, *kwargs.values()]
                if (b := getattr(v, "__native_namespace__", None))
            }

            if backends.__len__() > 1:
                msg = "Found multiple backends. Make sure that all dataframe/series inputs come from the same backend."
                raise ValueError(msg)

            result = func(*args, **kwargs)

            return to_native(result, strict=strict)

        return wrapper

    if func is None:
        return decorator
    else:
        # If func is not None, it means the decorator is used without arguments
        return decorator(func)
```

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #469 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Not been fully discussed how to approach this, but this is how I would go about it

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -133,7 +133,7 @@ def new_series(
         raise NotImplementedError(msg)
     else:  # pragma: no cover
         try:
-            # implementation is UNKNOWN, Narhwals extension using this feature should
+            # implementation is UNKNOWN, Narwhals extension using this feature should
             # implement `from_dict` function in the top-level namespace.
             native_series = native_namespace.new_series(name, values, dtype)
         except AttributeError as e:
@@ -256,7 +256,7 @@ def from_dict(
         native_frame = native_namespace.table(data, schema=schema)
     else:  # pragma: no cover
         try:
-            # implementation is UNKNOWN, Narhwals extension using this feature should
+            # implementation is UNKNOWN, Narwhals extension using this feature should
             # implement `from_dict` function in the top-level namespace.
             native_frame = native_namespace.from_dict(data)
         except AttributeError as e:
--- a/narwhals/stable/v1.py
+++ b/narwhals/stable/v1.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from functools import wraps
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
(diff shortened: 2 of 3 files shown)


MarcoGorelli (approved):
well that was simple 🤦 

thanks so much @FBruzzesi !

PR #835, opened 2024-08-21
dask-quantile-draft

anopsy:
Hello, that's my attempt at solving Dask `Expr.quantile`- the method works as expected (according to my own tests). However I've been trying to find a way to include dask-case in our tests, since dask can work only with interpolation="nearest" and not the other ones.
What I tried: 
- adding is_dask param to the parameters in tests and running the tests using nullcontext, but compare dict doesn't want to cooperate with error messages,
- writing separate test for dask cases and using `if "dask" in str(contructor)` to call that test but also failed here 
I was thinking about including every single constructor type in the parameters, but that seems a bit too verbose and the number of possible combinations is humongous.
At the moment I'm exploring other options, but feel free to 🛟 *send help* 🛟 if you know how to approach it

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -4,6 +4,7 @@ from copy import copy
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
+from typing import Literal
 from typing import NoReturn
 
 from narwhals._dask.utils import add_row_index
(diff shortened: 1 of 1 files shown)


anopsy:
I hope you enjoy your holiday :smiley: Thank you so much I knew I'm missing some really simple solution :heart_hands:

FBruzzesi:
I think it is only missing an alert/note in the exposed method and then it should be good to go?!

anopsy:
like that?

FBruzzesi:
> like that? ` Note: dask has its own method to approximate quantile and it doesn't implement 'nearest', 'higher', 'lower', 'midpoint' as interpolation method - use 'linear' instead`

Yes, you could add it in the already existing note and create bullet points. If possible please consider adding an embedded link to the dask method as well.

MarcoGorelli (approved):
thanks @anopsy (and @FBruzzesi for reviewing!)

content looks good, merge when ready

anopsy:
cool, thanks for suggestion, was wondering how to arrange those two notes, also what do you mean by an embedded link to the dask method? A link to the dask docs?

FBruzzesi:
We end up using [Series.quantile](https://docs.dask.org/en/stable/generated/dask.dataframe.Series.quantile.html)

PR #839, opened 2024-08-21
feat: support `__getitem__` with slices for columns

MarcoGorelli:
I kinda wish we didn't have to go overboard with `__getitem__` overloads, but scikit-learn are using this, so I think we should

My general thinking on this is: make sure that Narwhals is easily usable by scikit-learn. Then, it's their decision whether or not they adopt - but we should at least be ready

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -160,7 +160,7 @@ class ArrowDataFrame:
                             columns[item[1].start : item[1].stop : item[1].step]
                         )
                     )
-                msg = f"Expected slice of integers of strings, got: {type(item[1])}"  # pragma: no cover
+                msg = f"Expected slice of integers or strings, got: {type(item[1])}"  # pragma: no cover
                 raise TypeError(msg)  # pragma: no cover
 
             from narwhals._arrow.series import ArrowSeries
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -137,6 +137,32 @@ class ArrowDataFrame:
             )
 
         elif isinstance(item, tuple) and len(item) == 2:
+            if isinstance(item[1], slice):
+                columns = self.columns
+                if isinstance(item[1].start, str) or isinstance(item[1].stop, str):
+                    start = (
+                        columns.index(item[1].start)
+                        if item[1].start is not None
+                        else None
+                    )
+                    stop = (
+                        columns.index(item[1].stop) + 1
+                        if item[1].stop is not None
+                        else None
+                    )
+                    step = item[1].step
+                    return self._from_native_frame(
+                        self._native_frame.take(item[0]).select(columns[start:stop:step])
+                    )
+                if isinstance(item[1].start, int) or isinstance(item[1].stop, int):
+                    return self._from_native_frame(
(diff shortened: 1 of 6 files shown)


narwhals/_arrow/dataframe.py
```diff
@@ -137,6 +137,32 @@ def __getitem__(
...
+                    return self._from_native_frame(
+                        self._native_frame.take(item[0]).select(columns[start:stop:step])
+                    )
+                if isinstance(item[1].start, int) or isinstance(item[1].stop, int):
+                    return self._from_native_frame(
+                        self._native_frame.take(item[0]).select(
+                            columns[item[1].start : item[1].stop : item[1].step]
+                        )
+                    )
+                msg = f"Expected slice of integers of strings, got: {type(item[1])}"  # pragma: no cover
```
FBruzzesi, line 163: Same typo everywhere 🙈
```suggestion
                msg = f"Expected slice of integers or strings, got: {type(item[1])}"  # pragma: no cover
```

FBruzzesi (commented):
I left a tiny comment on a typo.

I wonder if it could be worth to:
- Add an example in the docstring of the method
- Add a note for the behavior since slicing here is
  - If string: both start and stop are included
  - If int: typical python behavior, start included and stop excluded

narwhals/dataframe.py
```diff
@@ -544,14 +544,27 @@ def __getitem__(
...
+                    - `df[0:2, 0]` extracts the first two rows from the first column and returns
+                        a `Series`.
+                    - `df[[0, 1], [0, 1, 2]]` extracts the first two rows and the first three columns
+                        and returns a `DataFrame`
+                    - `df[0: 2, ['a', 'c']]` extracts the first two rows and columns `'a'` and `'c'` and
+                        returns a `DataFrame`
+                    - `df[:, 0: 2]` extracts all rows from the first two columns and returns a `DataFrame`
+                    - `df[:, 'a': 'c']` extracts all rows and all columns positioned between `'a'` and `'c'`
+                        _inclusive_ and returns a `DataFrame`. For example, if the columns are
+                        `'a', 'b', 'c', 'd'`, then that would extract columns `'a'`, `'b'`, and `'c'`.
```
FBruzzesi, line 563: This is amazing ✨🚀

MarcoGorelli:
good points, thanks! I think this is easier to explain with examples

narwhals/dataframe.py
MarcoGorelli: 😄 thanks, maybe we should upstream this to Polars itself, there's zero docs on getitem there

FBruzzesi: That's a good idea, I had never seen `df[:, "a":"c"]` in polars code 😂

FBruzzesi: Rendering may need one less indentation:

image

PR #840, opened 2024-08-21
feat: dask namespace `concat` method 

benrutter:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #637

## Checklist

- [x] Tests added 
- [x] Documented the changes

I've added in a `concat` method returning a `DaskLazyFrame`. One thing that *wasn't* tested for the `concat_test` was whether an error gets thrown trying to horizontally concat frames of different length. I'm not 100% sure what happens within polars/pandas when you do this, but dask will throw a confusing assertion error when you try to access anything within the dataframe, so I thought putting in a handy check here was worthwhile.

### DIFF
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -156,7 +156,7 @@ class DaskNamespace:
                 tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
             ):
                 msg = "unable to vstack with non-matching columns"
-                raise TypeError(msg)
+                raise AssertionError(msg)
             axis = 0
             join = "inner"
         elif how == "horizontal":
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -169,7 +169,7 @@ class DaskNamespace:
                     f"Columns with name(s): {', '.join(duplicates)} "
                     "have more than one occurrence"
                 )
-                raise TypeError(msg)
+                raise AssertionError(msg)
             axis = 1
             join = "outer"
         else:
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -1,18 +1,20 @@
 from __future__ import annotations
 
 from functools import reduce
+from itertools import chain
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
+from typing import Iterable
 from typing import NoReturn
 
 from narwhals import dtypes
+from narwhals._dask.dataframe import DaskLazyFrame
 from narwhals._dask.expr import DaskExpr
 from narwhals._dask.selectors import DaskSelectorNamespace
 from narwhals._expression_parsing import parse_into_exprs
 
 if TYPE_CHECKING:
-    from narwhals._dask.dataframe import DaskLazyFrame
     from narwhals._dask.typing import IntoDaskExpr
 
 
@@ -134,6 +136,52 @@ class DaskNamespace:
 
     def sum_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
         return reduce(lambda x, y: x + y, parse_into_exprs(*exprs, namespace=self))
+
+    def concat(
+        self,
+        items: Iterable[DaskLazyFrame],
+        *,
+        how: str = "vertical",
+    ) -> DaskLazyFrame:
+        import dask.dataframe as dd  # ignore-banned-import
+
+        if len(list(items)) == 0:
+            msg = "No items to concatenate"
+            raise ValueError(msg)
+        native_frames = [i._native_frame for i in items]
+        axis: int
+        join: str
+        if how == "vertical":
+            if not all(
+                tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
+            ):
+                msg = "unable to vstack with non-matching columns"
+                raise TypeError(msg)
+            axis = 0
+            join = "inner"
+        elif how == "horizontal":
+            all_column_names: list[str] = list(chain(*[i.columns for i in native_frames]))
+            if len(all_column_names) != len(set(all_column_names)):
+                duplicates = [
+                    i for i in all_column_names if all_column_names.count(i) > 1
+                ]
+                msg = (
+                    f"Columns with name(s): {', '.join(duplicates)} "
+                    "have more than one occurrence"
+                )
+                raise TypeError(msg)
+            axis = 1
+            join = "outer"
+        else:
+            msg = (
+                "Only valid options for concat are 'vertical' and 'horizontal' "
+                f"({how} not recognised)"
+            )
+            raise NotImplementedError(msg)
+        return DaskLazyFrame(
+            dd.concat(native_frames, axis=axis, join=join),
+            backend_version=self._backend_version,
+        )
 
     def _create_expr_from_series(self, _: Any) -> NoReturn:
         msg = "`_create_expr_from_series` for DaskNamespace exists only for compatibility"
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -18,7 +18,6 @@ from narwhals._expression_parsing import parse_into_exprs
 if TYPE_CHECKING:
     import dask_expr
 
-    from narwhals._dask.dataframe import DaskLazyFrame
     from narwhals._dask.typing import IntoDaskExpr
 
 
@@ -153,7 +152,7 @@ class DaskNamespace:
         import dask.dataframe as dd  # ignore-banned-import
 
         if len(list(items)) == 0:
-            msg = "No items to concatenate"
+            msg = "No items to concatenate"  # pragma: no cover
             raise AssertionError(msg)
         native_frames = [i._native_frame for i in items]
         if how == "vertical":
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -147,7 +147,7 @@ class DaskNamespace:
 
         if len(list(items)) == 0:
             msg = "No items to concatenate"
-            raise ValueError(msg)
+            raise AssertionError(msg)
         native_frames = [i._native_frame for i in items]
         axis: int
         join: str
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -160,7 +160,7 @@ class DaskNamespace:
             axis = 0
             join = "inner"
         elif how == "horizontal":
-            all_column_names: list[str] = list(chain(*[i.columns for i in native_frames]))
+            all_column_names: list[str] = [column for frame in native_frames for column in frame.columns]
             if len(all_column_names) != len(set(all_column_names)):
                 duplicates = [
                     i for i in all_column_names if all_column_names.count(i) > 1
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 from functools import reduce
-from itertools import chain
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
@@ -160,7 +159,9 @@ class DaskNamespace:
             axis = 0
             join = "inner"
         elif how == "horizontal":
-            all_column_names: list[str] = [column for frame in native_frames for column in frame.columns]
+            all_column_names: list[str] = [
+                column for frame in native_frames for column in frame.columns
+            ]
             if len(all_column_names) != len(set(all_column_names)):
                 duplicates = [
                     i for i in all_column_names if all_column_names.count(i) > 1
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -148,21 +148,21 @@ class DaskNamespace:
             msg = "No items to concatenate"
             raise AssertionError(msg)
         native_frames = [i._native_frame for i in items]
-        axis: int
-        join: str
         if how == "vertical":
             if not all(
                 tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
             ):  # pragma: no cover
                 msg = "unable to vstack with non-matching columns"
                 raise AssertionError(msg)
-            axis = 0
-            join = "inner"
-        elif how == "horizontal":
+            return DaskLazyFrame(
+                dd.concat(native_frames, axis=0, join="inner"),
+                backend_version=self._backend_version,
+            )
+        if how == "horizontal":
             all_column_names: list[str] = [
                 column for frame in native_frames for column in frame.columns
             ]
-            if len(all_column_names) != len(set(all_column_names)):
+            if len(all_column_names) != len(set(all_column_names)):  # pragma: no cover
                 duplicates = [
                     i for i in all_column_names if all_column_names.count(i) > 1
                 ]
@@ -171,18 +171,11 @@ class DaskNamespace:
                     "have more than one occurrence"
                 )
                 raise AssertionError(msg)
-            axis = 1
-            join = "outer"
-        else:
-            msg = (
-                "Only valid options for concat are 'vertical' and 'horizontal' "
-                f"({how} not recognised)"
+            return DaskLazyFrame(
+                dd.concat(native_frames, axis=1, join="outer"),
+                backend_version=self._backend_version,
             )
-            raise NotImplementedError(msg)
-        return DaskLazyFrame(
-            dd.concat(native_frames, axis=axis, join=join),
-            backend_version=self._backend_version,
-        )
+        raise NotImplementedError
 
     def mean_horizontal(self, *exprs: IntoDaskExpr) -> IntoDaskExpr:
         dask_exprs = parse_into_exprs(*exprs, namespace=self)
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -154,7 +154,7 @@ class DaskNamespace:
         if how == "vertical":
             if not all(
                 tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
-            ):
+            ):  # pragma: no cover
                 msg = "unable to vstack with non-matching columns"
                 raise AssertionError(msg)
             axis = 0
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -6,9 +6,7 @@ import narwhals.stable.v1 as nw
 from tests.utils import compare_dicts
 
 
-def test_concat_horizontal(constructor: Any, request: Any) -> None:
-    if "dask" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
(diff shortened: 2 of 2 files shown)


narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,41 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        native_frames = [i._native_frame for i in items]
+        axis: int
+        if how == "vertical":
+            all_columns = set(chain(*[i.columns for i in native_frames]))
+            if not all(set(i.columns) == all_columns for i in native_frames):
+                msg = "unable to vstack with non-matching columns"
+                raise TypeError(msg)
+            axis = 0
+        elif how == "horizontal":
+            if len({len(i) for i in native_frames}) != 1:
```
FBruzzesi, line 160: `len` will trigger computations and we would rather avoid that.
I am not sure if `dask_expr._expr.are_co_aligned` applied to dataframes, but if it does, I think that the following check could be all we need:

```suggestion
            if not get_dask_expr()._expr.are_co_aligned(*native_frames):
```

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,41 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+                raise ValueError(msg)
+            axis = 1
+        else:
+            msg = (
+                "Only valid options for concat are 'vertical' and 'horizontal' "
+                f"({how} not recognised)"
+            )
+            raise NotImplementedError(msg)
+        return DaskLazyFrame(
+            dd.concat([i._native_frame for i in items], axis=axis),
```
FBruzzesi, line 171: You can use `native_frames` already created:

```suggestion
            dd.concat(native_frames, axis=axis),
```

FBruzzesi (commented):
Hey thanks for the PR, that was fast :) 

I left a couple of comments, the main concern is about triggering computation on behalf of the user, which we would like to avoid as much as possible.

I would also double check the behaviour of the default `join="outer"` argument for [dd.concat](https://docs.dask.org/en/latest/generated/dask.dataframe.concat.html).

Finally, regarding the missing test, feel free to add it to get higher/better coverage of edge cases.

narwhals/_dask/namespace.py
benrutter: Looks like `are_co_aligned` doesn't work (throws an error when it shouldn't in testing). I think its because `are_co_aligned` is checking a bit more than things just having the same length, which actually makes me wonder what horizontal concatenation would do if dataframes had the same length but different partitioning (I think it would probably error but not get caught with my check).

I added in the check as a convenience though, it's probably fine to take out (dask will just raise an error for invalid vertical concatenation later on) - what do you think?

benrutter:
Cheers @FBruzzesi!

Think you're right around the behaviour of "outer" by default. In theory, I think it only comes into play on "vertical" merges (otherwise, mismatching indexes throw errors) where columns are checked to be matching already, but that sounds a little like an error waiting to happen.

Just looked at adding another test, but looks like its totally fine to horizontally concatenate different length frames outside of dask. For now, I've taken out the check for matching lengths- it'll avoid any computing for figuring out length, and pass over handling the error to dask, which is probably preferable? (let me know if I'm missing a trick on how to check concats are likely to be valid though)

narwhals/_dask/namespace.py
benrutter: I've taken *out* the check for now, which I've convinced myself it a good move because:
- I don't think my `len` comparison is really covering all non-valid cases
- It avoids the additional compute

FBruzzesi:
Hey sorry I am from mobile, so this could be a sloppy comment. I will try to do my best.

> Think you're right around the behaviour of "outer" by default. In theory, I think it only comes into play on "vertical" merges (otherwise, mismatching indexes throw errors) where columns are checked to be matching already, but that sounds a little like an error waiting to happen.

Alright, to me the join keyword resonates more on a horizontal concat but apparently not. I am happy to keep the check on the column names.

> Just looked at adding another test, but looks like its totally fine to horizontally concatenate different length frames outside of dask. For now, I've taken out the check for matching lengths- it'll avoid any computing for figuring out length, and pass over handling the error to dask, which is probably preferable? (let me know if I'm missing a trick on how to check concats are likely to be valid though)

Maybe we can just let dask run its behavior without manually checking for dataframe lengths. This is what we are doing for pandas, yet I would try to see how Polars LazyFrame behaves (i.e. if it throws an error, and if it does it most likely happens only at collect time).

benrutter:
Thanks @FBruzzesi, comment is really helpful! At the moment, the changes I've made since my initial commit (have squashed the commits together for neatness):

- keyword argument of "inner" now passed into `dd.concat` - I think this is the desired behaviour?
- I've taken out length checks to avoid unnecessary compute and just let dask do its own thing with checks
- Extra tweak: I also swapped the `import dask.dataframe as dd  # ignore-banned-import` for `dd = get_dask_dataframe()` (I think that's better practice maybe?)

On the `join="inner"` vs `join="outer"` behaviour, where you thinking around any specific cases? I can add some extra tests in if there's an example that you have in mind.

FBruzzesi:
> * keyword argument of "inner" now passed into `dd.concat` - I think this is the desired behaviour?

Thanks! I am still a bit unsure about the difference. I can run some test later today or tomorrow, unless you have some examples ready to show

> * I've taken out length checks to avoid unnecessary compute and just let dask do its own thing with checks

I can still see the check using `len`?!

> * Extra tweak: I also swapped the `import dask.dataframe as dd  # ignore-banned-import` for `dd = get_dask_dataframe()` (I think that's better practice maybe?)

We recently changed all those (check #788), so I would suggest to keep it as before:

```diff
- dd = get_dask_dataframe()
+ import dask.dataframe as dd  # ignore-banned-import
```

> On the `join="inner"` vs `join="outer"` behaviour, where you thinking around any specific cases? I can add some extra tests in if there's an example that you have in mind.

Back to the first point

benrutter:
Well that's embarrassing! Looks like my "neat squashing" was just a "neat deleting a commit" for the validation check removal 🫠 All taken out now.

Ok, I've done some playing around and actually looks like dask *can* concatenate on axis 1 with different lengths anyway (I don't know if this works with expression based columns though, definitely was getting an assertion error previously). That's pretty handy because it makes the outer/inner behaviour really clear for both!

I'll explain them here, and hopefully you can give me a steer on the desired Narwhals behaviour? (I *think* I know but not 100% sure)

```python
import dask.dataframe

left = dd.from_dict({"a": [1, 2], "b": [1, 2]}, npartitions=1)
right = dd.from_dict({"a": [1, 2, 3], "b": [1, 2, 3], "c": [1, 2, 3]}, npartitions=1)

dd.concat([right, left], axis=0, join="inner").compute()
#   a b
# 0 1 1
# 1 2 2
# 2 3 3
# 0 1 1
# 1 2 2

# inner on axis 0
dd.concat([right, left], axis=0, join="inner").compute()
#   a b
# 0 1 1
# 1 2 2
# 2 3 3
# 0 1 1
# 1 2 2

# outer on axis 0
dd.concat([right, left], axis=0, join="outer").compute()
#   a b c
# 0 1 1 1.0
# 1 2 2 2.0
# 2 3 3 3.0
# 0 1 1 NaN
# 1 2 2 NaN


# inner on axis 1
dd.concat([right, left], axis=1, join="inner").compute()
#   a b c a b
# 0 1 1 1 1 1
# 1 2 2 2 2 2

# outer on axis 1
dd.concat([right, left], axis=1, join="outer").compute()
#   a b c a   b
# 0 1 1 1 1.0 1.0
# 1 2 2 2 2.0 2.0
# 2 3 3 3 NaN Nan
```

Hopefully that makes it a little clearer? Think I muddied the water a little by being confused before on the behaviour for `axis=1`, sorry!

Looking at the other examples in narwhals I *think* that when `axis=0` (vertical concat) we'd want `join=inner` and when 'axis=1' (horizontal concat) we'd want `join=outer`?

I've implemented the above behaviour for now - would appreciate feedback on if that's logical though! 🙏

FBruzzesi:
Awesome, that makes it much more clear and saved me a bunch of time, thanks a lot.

I am checking polars LazyFrame behavior, which is what we should aim to replicate:

Edit:
## TL;DR

- `how="vertical"` behavior should be easy to replicate
- `how="horizontal"` we could end up with a stricter behavior for dask than Polars `LazyFrame`, but I think that's still ok.

## `how="vertical"`

```py
import polars as pl

data1 = {"a": [1, 2, 3], "b": [1, 2, 3], "c": [1, 2, 3]}
data2 = {"a": [1, 2], "b": [1, 2]}

df1 = pl.LazyFrame(data1)
df2 = pl.LazyFrame(data2)

df = pl.concat([df1, df2], how="vertical")  # this doesn't raise just yet, but then when we collect
df.collect()
```
> ShapeError: unable to append to a DataFrame of width 3 with a DataFrame of width 2

So I think we can raise early if columns are not same for vertical, exactly how you are already doing. Once the check is done, join keyword does not matter anymore since the columns will already be the same.

Edit: Actually the column order also matter in polars, i.e. the following fails:

```py
data1 = {"a": [1, 2, 3], "b": [1, 2, 3]}
data2 = {"b": [1, 2], "a": [1, 2]}

df1 = pl.LazyFrame(data1)
df2 = pl.LazyFrame(data2)

df = pl.concat([df1, df2], how="vertical").collect()
```
> ShapeError: unable to vstack, column names don't match: "a" and "b"

## `how="horizontal"`

First check is done on the duplicate column names, before collect time:

```py
data1 = {"a": [1, 2, 3], "b": [1, 2, 3]}
data2 = {"b": [1, 2], "a": [1, 2]}

df1 = pl.LazyFrame(data1)
df2 = pl.LazyFrame(data2)

df = pl.concat([df1, df2], how="horizontal")
```
> DuplicateError: Column with name 'a' has more than one occurrence

This such check should be fairly easy to achieve in dask as well.

If the column names are ok, then, actually the different length does not matter:

```py
data1 = {"a": [1, 2, 3], "b": [1, 2, 3]}
data2 = {"c": [1, 200], "d": [1, 2]}

df1 = pl.LazyFrame(data1)
df2 = pl.LazyFrame(data2)

df = pl.concat([df1, df2], how="horizontal")
df.collect()
```

```
shape: (3, 4)
┌─────┬─────┬──────┬──────┐
│ a   ┆ b   ┆ c    ┆ d    │
│ --- ┆ --- ┆ ---  ┆ ---  │
│ i64 ┆ i64 ┆ i64  ┆ i64  │
╞═════╪═════╪══════╪══════╡
│ 1   ┆ 1   ┆ 1    ┆ 1    │
│ 2   ┆ 2   ┆ 200  ┆ 2    │
│ 3   ┆ 3   ┆ null ┆ null │
└─────┴─────┴──────┴──────┘
```

This behavior I was not able to replicate in dask, meaning that at compute time if the length is different I am getting an `AssertionError` with whatever parameter I am trying to play with.

I think this is ok as we can be more restrictive than polars, but I would like to end up with the same result (cc: @MarcoGorelli ) 

Hope this helps

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,52 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        self,
+        items: Iterable[DaskLazyFrame],
+        *,
+        how: str = "vertical",
+    ) -> DaskLazyFrame:
+        import dask.dataframe as dd  # ignore-banned-import
+
+        if len(list(items)) == 0:
+            msg = "No items to concatenate"
+            raise ValueError(msg)
```
FBruzzesi, line 150:
```suggestion
            raise AssertionError(msg)
```

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,52 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        native_frames = [i._native_frame for i in items]
+        axis: int
+        join: str
+        if how == "vertical":
+            if not all(
+                tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
+            ):
+                msg = "unable to vstack with non-matching columns"
+                raise TypeError(msg)
```
FBruzzesi, line 159:
```suggestion
                raise AssertionError(msg)
```

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,52 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        if len(list(items)) == 0:
+            msg = "No items to concatenate"
+            raise ValueError(msg)
+        native_frames = [i._native_frame for i in items]
+        axis: int
+        join: str
+        if how == "vertical":
+            if not all(
+                tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
+            ):
```
FBruzzesi, line 157: I agree that we are introducing this check just now and for other dataframe can follow later on another PR.

As no test is introduced to check for this case, you can add a `# pragma: no cover` and come back to this later on, otherwise CI will complain regarding coverage not being 100%

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,52 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+            all_column_names: list[str] = list(chain(*[i.columns for i in native_frames]))
+            if len(all_column_names) != len(set(all_column_names)):
+                duplicates = [
+                    i for i in all_column_names if all_column_names.count(i) > 1
+                ]
+                msg = (
+                    f"Columns with name(s): {', '.join(duplicates)} "
+                    "have more than one occurrence"
+                )
+                raise TypeError(msg)
```
FBruzzesi, line 172: Either this:
```suggestion
                raise AssertionError(msg)
```
and maybe we can introduce custom exception later on to emulate what polars does (`DuplicateError`)

FBruzzesi (commented):
Thanks for adjusting on all fronts and already planning some further tests.

I commented on some error raise since:

- in other dataframe we raise AssertionError for empty list of items
- other errors are certainly not type related, so we can keep `AssertionError` for now.

I think we can merge after that :)

narwhals/_dask/namespace.py
benrutter: I never new what `# pragma: no cover` was until this point, and now I do 🙌

benrutter:
Awesome thanks @FBruzzesi - I've made those changes (as an FYI I also raised an issue with over on dask/dask to check around the expected behaviour for concatenating on `axis=1` https://github.com/dask/dask/issues/11343)

narwhals/_dask/namespace.py
```diff
@@ -135,6 +137,52 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        if how == "vertical":
+            if not all(
+                tuple(i.columns) == tuple(native_frames[0].columns) for i in native_frames
+            ):  # pragma: no cover
+                msg = "unable to vstack with non-matching columns"
+                raise AssertionError(msg)
+            axis = 0
+            join = "inner"
+        elif how == "horizontal":
+            all_column_names: list[str] = list(chain(*[i.columns for i in native_frames]))
```
aidoskanapyanov, line 163: Hi @benrutter, I believe this can be "simplified" like this, but feel free to keep it as it is though, just my 2 cents

```suggestion
            all_column_names: list[str] = [column for frame in native_frames for column in frame.columns]
```

benrutter: Thanks @aidoskanapyanov! That definitely is simpler to read - I'm on a mobile at the moment so will merge in later 😁 (think I need to take out then itertools import as well or it'll fail linting).

aidoskanapyanov: I'm guessing pre-commit CI is going to remove the import via autofix 😄

narwhals/_dask/namespace.py
```diff
@@ -135,6 +136,54 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+                )
+                raise AssertionError(msg)
+            axis = 1
+            join = "outer"
+        else:
+            msg = (
+                "Only valid options for concat are 'vertical' and 'horizontal' "
+                f"({how} not recognised)"
+            )
+            raise NotImplementedError(msg)
```
FBruzzesi, line 181: In arrow and pandas like implementation currently we raise `NotImplementedError` without any message.
I agree that it would be a nice addition, but for now we don't have a test for that therefore coverage complains

narwhals/_dask/namespace.py
```diff
@@ -135,6 +136,54 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+            ]
+            if len(all_column_names) != len(set(all_column_names)):
+                duplicates = [
+                    i for i in all_column_names if all_column_names.count(i) > 1
+                ]
+                msg = (
+                    f"Columns with name(s): {', '.join(duplicates)} "
+                    "have more than one occurrence"
+                )
+                raise AssertionError(msg)
```
FBruzzesi, line 173: This is a new check, and without coverage for the moment, feel free to add a pragma no cover until we add it to the other namespaces as well:

```diff
-            if len(all_column_names) != len(set(all_column_names)):
+            if len(all_column_names) != len(set(all_column_names)):  # pragma: no cover
```

narwhals/_dask/namespace.py
```diff
@@ -135,6 +136,54 @@ def any_horizontal(self, *exprs: IntoDaskExpr) -> DaskExpr:
...
+        how: str = "vertical",
+    ) -> DaskLazyFrame:
+        import dask.dataframe as dd  # ignore-banned-import
+
+        if len(list(items)) == 0:
+            msg = "No items to concatenate"
+            raise AssertionError(msg)
+        native_frames = [i._native_frame for i in items]
+        axis: int
+        join: str
```
FBruzzesi, line 152: Coverage complains for these initializations, maybe you can follow the same pattern as the other namespaces:

```py
...
if how == "vertical":
    # the appropriate checks
    return DaskLazyFrame(
        dd.concat(native_frames, axis=0, join="inner"),
    )
elif how == "horizontal"
    # checks
    return DaskLazyFrame(
        dd.concat(native_frames, axis=1, join="outer"),
    )
else:
    raise NotImplementedError
```

I think it would make it easier to _trick_ both mypy and coverage in this way

FBruzzesi (commented):
Logic is 100% in place, there are a couple of issues with coverage for:

- the new checks we discussed
- the variables introduced for mypy type hints

I commented on how to deal with these cases - as soon as all CI is green we are ready to merge

FBruzzesi (approved):
Ready to ship it? 🚀

image

MarcoGorelli:
nice one, thanks all! 🙌

PR #825, opened 2024-08-20
patch: `sum_horizontal` ignore nulls

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #822 

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below.

Not sure how to handle non numeric 🤔 

In horizontal sum, polars concatenates strings and fill nulls with empty strings, but also concatenates strings with numeric:

```py
df = pl.DataFrame({
    "a": [1, 8, 3],
   "c": ["x", "y", "z"],
})

df.select(pl.sum_horizontal("a", "c"))
```

```terminal
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ str │
╞═════╡
│ 1x  │
│ 8y  │
│ 3z  │
└─────┘
```

PR #843, opened 2024-08-22
feat: `mean_horizontal`

### LINKED ISSUE #814: [Enh]: support mean_horizontal

_No response_

It would be great to add support for [`mean_horizontal`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.mean_horizontal.html) from the polars API. Note that it is not the same as `sum_horizontal(cols) / N`, since it excludes null columns from mean.

_No response_

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #814 

## Checklist

- [x] Tests added 
- [x] Documented the changes

As `mean` can only be numeric, unlike #825, it is safe to fill nulls with zeros for the sum, and count the number of non-nulls summands.

### DIFF
--- a/narwhals/_polars/namespace.py
+++ b/narwhals/_polars/namespace.py
@@ -1,11 +1,13 @@
 from __future__ import annotations
 
+from functools import reduce
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Iterable
 from typing import Sequence
 
 from narwhals import dtypes
+from narwhals._expression_parsing import parse_into_exprs
 from narwhals._polars.utils import extract_args_kwargs
 from narwhals._polars.utils import narwhals_to_native_dtype
 from narwhals.dependencies import get_polars
@@ -15,6 +17,7 @@ if TYPE_CHECKING:
     from narwhals._polars.dataframe import PolarsDataFrame
     from narwhals._polars.dataframe import PolarsLazyFrame
     from narwhals._polars.expr import PolarsExpr
+    from narwhals._polars.typing import IntoPolarsExpr
 
 
 class PolarsNamespace:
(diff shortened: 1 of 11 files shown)


FBruzzesi:
Ok, so this ended up being a bit more than initially expected on the polars side 😂

narwhals/_polars/namespace.py
```diff
@@ -85,11 +88,28 @@ def lit(self, value: Any, dtype: dtypes.DType | None = None) -> PolarsExpr:
             return PolarsExpr(pl.lit(value, dtype=narwhals_to_native_dtype(dtype)))
         return PolarsExpr(pl.lit(value))
 
-    def mean(self, *column_names: str) -> Any:
+    def mean(self, *column_names: str) -> PolarsExpr:
```
FBruzzesi, line 91: We should return `PolarsExpr` to be able to keep chaining and interacting with other narwhals exprs if needed

narwhals/_polars/namespace.py
```diff
@@ -85,11 +88,28 @@ def lit(self, value: Any, dtype: dtypes.DType | None = None) -> PolarsExpr:
...
+
+        pl = get_polars()
+        polars_exprs = parse_into_exprs(*exprs, namespace=self)
+
+        if self._backend_version < (0, 20, 8):  # pragma: no cover
+            total = reduce(lambda x, y: x + y, (e.fill_null(0.0) for e in polars_exprs))
+            n_non_zero = reduce(
+                lambda x, y: x + y, ((1 - e.is_null()) for e in polars_exprs)
+            )
+            return PolarsExpr(total._native_expr / n_non_zero._native_expr)
```
FBruzzesi, line 110: `__div__` is not currently implemented and I did not want to make the PR even larger

MarcoGorelli (approved):
this is wonderful, thanks @FBruzzesi !

PR #847, opened 2024-08-22
feat: Support more indexing: boolean lists to DataFrame.filter and Series.filter, add DataFrame.row, 

MarcoGorelli:
A bunch more things which would be useful to scikit-learn

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -711,6 +711,10 @@ class DataFrame(BaseFrame[FrameT]):
         """
         Get values at given row.
 
+        !!!note
+            You should NEVER use this method to iterate over a DataFrame;
+            if you require row-iteration you should strongly prefer use of iter_rows() instead.
+
         Arguments:
             index: Row number.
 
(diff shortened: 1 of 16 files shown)


narwhals/dataframe.py
```diff
@@ -599,6 +604,8 @@ def __getitem__(
                 2
             ]
         """
+        if isinstance(item, int):
+            item = [item]
```
MarcoGorelli, line 608: I'm not wild about using `df[0]` to extract a single row...OK with silently allowing it (and testing it), as scikit-learn use this, but perhaps we can get away with not documenting it to not encourage it 🙈

PR #854, opened 2024-08-23
patch: allow `lit` to broadcast as left operand

### LINKED ISSUE #853: [Bug]: `lit` does not "broadcast" as expected
### Describe the bug

Narwhals implementation of `lit` breaks when it is the left most expr.

Figured this out while doing a fold starting with a lit initial value.

```py
import narwhals as nw
import pandas as pd
import polars as pl
import pyarrow as pa

data = {"a": [1, 2, 3]}

df_pd = pd.DataFrame(data)
df_pl = pl.DataFrame(data)
df_pa = pa.table(data)

@nw.narwhalify
def sum_on_the_right(df):
    return df.select(nw.col("a") + nw.lit(0))

sum_on_the_right(df_pl), sum_on_the_right(df_pd), sum_on_the_right(df_pa)
# This is all good and well

@nw.narwhalify
def sum_on_the_left(df):
    return df.select(nw.lit(0) + nw.col("a"))

sum_on_the_left(df_pl)  # polars has no issues

sum_on_the_left(df_pd)
# ValueError: Length mismatch: Expected axis has 3 elements, new values have 1 elements

sum_on_the_left(df_pa)
# ArrowInvalid: Array arguments must all be the same length
```  

### Expected results

Being able to broadcast value

### Actual results

Broadcast from the left

```shell
System:
    python: 3.10.12 (main, Jul 29 2024, 16:56:48) [GCC 11.4.0]
   machine: Linux-5.15.133.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 1.5.3
       pandas: 2.2.2
       polars: 1.5.0
         cudf: 
        modin: 0.31.0
      pyarrow: 17.0.0
        numpy: 2.0.1
```

### Relevant log output

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #853 

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/namespace.py
+++ b/narwhals/_pandas_like/namespace.py
@@ -15,10 +15,10 @@ from narwhals._pandas_like.series import PandasLikeSeries
 from narwhals._pandas_like.utils import create_native_series
 from narwhals._pandas_like.utils import horizontal_concat
 from narwhals._pandas_like.utils import vertical_concat
+from narwhals.utils import Implementation
 
 if TYPE_CHECKING:
     from narwhals._pandas_like.typing import IntoPandasLikeExpr
-    from narwhals.utils import Implementation
 
 
 class PandasLikeNamespace:
@@ -130,11 +130,17 @@ class PandasLikeNamespace:
         )
 
     def lit(self, value: Any, dtype: dtypes.DType | None) -> PandasLikeExpr:
+        if self._implementation is Implementation.CUDF:
+            import cupy as np  # ignore-banned-import
+        else:
+            import numpy as np  # ignore-banned-import
+
         def _lit_pandas_series(df: PandasLikeDataFrame) -> PandasLikeSeries:
+            native_frame = df._native_frame
             pandas_series = PandasLikeSeries._from_iterable(
-                data=[value],
+                data=np.full(native_frame.shape[0], value),
                 name="lit",
-                index=df._native_frame.index[0:1],
+                index=native_frame.index,
                 implementation=self._implementation,
                 backend_version=self._backend_version,
             )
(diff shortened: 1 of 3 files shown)


narwhals/_pandas_like/namespace.py
```diff
@@ -130,11 +130,17 @@ def all(self) -> PandasLikeExpr:
         )
 
     def lit(self, value: Any, dtype: dtypes.DType | None) -> PandasLikeExpr:
+        if self._implementation is Implementation.CUDF:
+            import cupy as np  # ignore-banned-import
```
FBruzzesi, line 134: cudf depends on cupy, and cupy implements [full](https://docs.cupy.dev/en/stable/reference/generated/cupy.full.html)

MarcoGorelli (approved):
thanks @FBruzzesi - this is probably fine, we need to constructor the whole column anyway in the result, it's not like for `nw.col('a') - nw.col('a').mean()` where we can just pass the scalar through

and 👍 on the cupy thing!

FBruzzesi:
> it's not like for `nw.col('a') - nw.col('a').mean()` where we can just pass the scalar through

Oh this made me think that probably `nw.lit(1) - nw.col("a").mean()` is likely to result of the wrong shape.

I can take a look but most likely _not_ during this weekend

MarcoGorelli:
ah right, thanks for bringing that up

as you say, "we're alright until a release" 😄 

i'll revert this for now then so we can release other things, then we'll get back to this

PR #808, opened 2024-08-18
docs: docstring for `concat`

Sherwin-14:
I wrote the docstring's for concat function. Please have a look into it

PR #869, opened 2024-08-25
patch: stableify `concat` function

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues 

See [comment](https://github.com/narwhals-dev/narwhals/pull/808?notification_referrer_id=NT_kwDOAo1WGLQxMTk5Nzg1MzA4Njo0MjgxNzA0OA#issuecomment-2308934423)

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -10,10 +10,10 @@ def test_concat_horizontal(constructor: Any, request: Any) -> None:
     if "dask" in str(constructor):
         request.applymarker(pytest.mark.xfail)
     data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
-    df_left = nw.from_native(constructor(data))
+    df_left = nw.from_native(constructor(data)).lazy()
 
     data_right = {"c": [6, 12, -1], "d": [0, -4, 2]}
-    df_right = nw.from_native(constructor(data_right))
+    df_right = nw.from_native(constructor(data_right)).lazy()
 
     result = nw.concat([df_left, df_right], how="horizontal")
     expected = {
@@ -34,7 +34,7 @@ def test_concat_vertical(constructor: Any, request: Any) -> None:
         request.applymarker(pytest.mark.xfail)
     data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
     df_left = (
-        nw.from_native(constructor(data)).rename({"a": "c", "b": "d"}).drop("z").lazy()
+        nw.from_native(constructor(data)).lazy().rename({"a": "c", "b": "d"}).drop("z")
     )
 
     data_right = {"c": [6, 12, -1], "d": [0, -4, 2]}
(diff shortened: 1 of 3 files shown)


FBruzzesi:
I don't get how locally it passes. I am getting this some many times 🤔 gotta be some config out of sync, but I am not sure what needs change

MarcoGorelli:
:thinking: how annoying...nothing worse than a ci failure you can't reproduce...will take a look

FBruzzesi:
> 🤔 how annoying...nothing worse than a ci failure you can't reproduce...will take a look

Thanks Marco! I was actually in a bit of a rush and can't put more effort before tomorrow.
The only thing I can think of is that the signature should be:

```diff
def concat(
+    items: Iterable[NwDataFrame[Any] | NwLazyFrame[Any]],
-    items: Iterable[DataFrame[Any] | LazyFrame[Any]],
    *,
    how: Literal["horizontal", "vertical"] = "vertical",
) -> DataFrame[Any] | LazyFrame[Any]:
```

since those stable dataframes are passed in. However then the "inner" concat types would not comply.

Sorry for the quick and dirty solution

tests/frame/test_invalid.py
```diff
@@ -24,7 +24,7 @@ def test_validate_laziness() -> None:
         NotImplementedError,
         match=("The items to concatenate should either all be eager, or all lazy"),
     ):
-        nw.concat([nw.from_native(df, eager_only=True), nw.from_native(df).lazy()])
+        nw.concat([nw.from_native(df, eager_only=True), nw.from_native(df).lazy()])  # type: ignore[list-item]
```
FBruzzesi, line 27: I am arguing that this is correct for mypy to complain, as it is supposed to raise an error and that's what we are checking

tests/frame/concat_test.py
```diff
@@ -10,10 +10,10 @@ def test_concat_horizontal(constructor: Any, request: Any) -> None:
     if "dask" in str(constructor):
         request.applymarker(pytest.mark.xfail)
     data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
-    df_left = nw.from_native(constructor(data))
+    df_left = nw.from_native(constructor(data)).lazy()
 
     data_right = {"c": [6, 12, -1], "d": [0, -4, 2]}
-    df_right = nw.from_native(constructor(data_right))
+    df_right = nw.from_native(constructor(data_right)).lazy()
```
FBruzzesi, line 16: Solution was to cast both to lazy so that mypy knows that it's not either eager or lazy.

tests/frame/test_invalid.py
MarcoGorelli: yup, and this is quite satisfying 😋

tests/frame/concat_test.py
MarcoGorelli: ah I remember noticing this some time ago, then forgot about it...

MarcoGorelli (approved):
thanks for doing this @FBruzzesi !

PR #873, opened 2024-08-26
docs: Recommend `uv` in contributing guide

MarcoGorelli:
Maybe we should just be opinionated and recommend `uv` in the contributing guide? My experience with it is that it "just works" in a way no other tool does

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,8 @@ git clone git@github.com:YOUR-USERNAME/narwhals.git
 
 Here's how you can set up your local development environment to contribute.
 
+#### Option 1: Use UV (recommended)
+
 1. Make sure you have Python3.8+ installed (for example, Python 3.11), create a virtual environment,
    and activate it. If you're new to this, here's one way that we recommend:
    1. Install uv: https://github.com/astral-sh/uv?tab=readme-ov-file#getting-started
@@ -72,6 +74,14 @@ uv pip install pre-commit
 pre-commit install
 ```
 This will automatically format and lint your code before each commit, and it will block the commit if any issues are found.
+
+#### Option 2: use python3-venv
+
+1. Make sure you have Python 3.8+ installed. If you don't, you can check [install Python](https://realpython.com/installing-python/)
+   to learn how. Then, [create and activate](https://realpython.com/python-virtual-environments-a-primer/)
+   a virtual environment.
+
+Then, follow steps 2-4 from above but using `pip install` instead of `uv pip install`.
 
 ### 5. Working on your issue
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,8 +80,7 @@ This will automatically format and lint your code before each commit, and it wil
 1. Make sure you have Python 3.8+ installed. If you don't, you can check [install Python](https://realpython.com/installing-python/)
    to learn how. Then, [create and activate](https://realpython.com/python-virtual-environments-a-primer/)
    a virtual environment.
-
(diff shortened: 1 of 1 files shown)


lucianosrp:
Nice one! We could eventually also set up the project with `uv sync` for contributors

lucianosrp:
With `uv sync` I get this at the moment fyi
```
❯ uv sync
Using Python 3.12.2
Creating virtualenv at: .venv
  × No solution found when resolving dependencies:
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy Python>=3.10 and the requested Python version (>=3.8) does not satisfy
      Python>=3.9,<3.10, we can conclude that Python>=3.9 is incompatible.
      And because only the following versions of dask[dataframe] are available:
          dask[dataframe]<=2024.7.0
          dask[dataframe]==2024.7.1
          dask[dataframe]==2024.8.0
          dask[dataframe]==2024.8.1
      and the requested Python version (>=3.8) does not satisfy Python>=3.10, we can conclude that dask[dataframe]>=2024.7.0 is incompatible.
      And because narwhals[dask] depends on dask[dataframe]>=2024.7 and your project requires narwhals[dask], we can conclude that your
      projects's requirements are unsatisfiable.
```

EdAbati:
I think it is great that we recommend `uv`, I'm also liking it more and more. :)

I'd still keep an "alternative" guide with plain `pip`/`venv`. In some cases, people may prefer to stick with the tool they already know (e.g., beginners, or someone looking to get a quick PR). Requiring to install (and know) a tool may be a (small) barrier of entry, what do you think?

I quite like the [sklearn guide](https://scikit-learn.org/dev/developers/advanced_installation.html#building-from-source): it recommends conda but also have a "classic" pip alternative guide

MarcoGorelli:
thanks for your inputs!

going to think about this one / ask around. because conversely, showing multiple options look confusing to newcomers too. but maybe that's ok, we can show uv / pip / conda, and be opinionated in marking the first one as "recommended" so that a total newcomer doesn't feel paralysed by choice

CONTRIBUTING.md
```diff
@@ -47,22 +47,41 @@ git clone git@github.com:YOUR-USERNAME/narwhals.git
...
-
-1. Make sure you have Python3.8+ installed (for example, Python 3.11)
-2. Create a new virtual environment with `python3.11 -m venv .venv` (or whichever version of Python3.9+ you prefer)
-3. Activate it: `. .venv/bin/activate`
-4. Install Narwhals: `pip install -e .`
-5. Install test requirements: `pip install -r requirements-dev.txt`
-6. Install docs requirements: `pip install -r docs/requirements-docs.txt`
+Here's how you can set up your local development environment to contribute.
+
+#### Option 1: Use UV (recommended)
```
lucianosrp, line 52: What about using [content tabs](https://squidfunk.github.io/mkdocs-material/reference/content-tabs/#linked-content-tabs-feature-enabled) here ? 

Something like
```
=== "uv (recommended)"
    ```uv pip install```
=== "pip"
    ```pip install```
```

etc

EdAbati (approved):
nice! :)

MarcoGorelli:
thanks for your reviews! ok shipping as-is for now so that we've got this in, we can then do improvements such as content tabs as follow-ups

PR #888, opened 2024-08-31
docs: Example correction

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -165,13 +165,12 @@ def from_dict(
         >>> import pandas as pd
         >>> import polars as pl
         >>> import narwhals as nw
-        >>> data = {"a": [1, 2, 3], "b": [4, 5, 6]}
+        >>> data = {"c": [5, 2], "d": [1, 4]}
 
         Let's define a dataframe-agnostic function:
 
         >>> @nw.narwhalify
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
thanks @DeaMariaLeon ! I think the current example is technically correct, but probably needs a better explanation. We're trying to show that you can start with a certain dataframe, and then make a new dataframe with new data by using `from_dict` and passing in `native_namespace`

but evidently we didn't explain it very well if you misunderstood it (the blame is definitely on the example not being clear enough, not on you!)

perhaps we could add an extra sentence to the example to clarify what we're doing? like "Let's create a new dataframe of the same class as the dataframe we started with, from a dict of new data"?

DeaMariaLeon:
Thanks for the explanation!

MarcoGorelli (approved):
thanks @DeaMariaLeon

PR #927, opened 2024-09-08
feat: add `on` key to `join`

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -195,7 +195,7 @@ class BaseFrame(Generic[FrameT]):
             raise NotImplementedError(msg)
 
         if how == "cross" and (left_on or right_on or on):
-            msg = "Can not pass left_on, right_on, on for cross join"
+            msg = "Can not pass `left_on`, `right_on` or `on` keys for cross join"
             raise ValueError(msg)
 
         if (how != "cross" and on is None) and (left_on is None or right_on is None):
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -186,6 +186,7 @@ class BaseFrame(Generic[FrameT]):
         how: Literal["inner", "left", "cross", "semi", "anti"] = "inner",
         left_on: str | list[str] | None = None,
         right_on: str | list[str] | None = None,
+        on: str | list[str] | None = None,
     ) -> Self:
         _supported_joins = ("inner", "left", "cross", "anti", "semi")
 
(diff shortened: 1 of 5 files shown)


narwhals/dataframe.py
```diff
@@ -186,17 +186,31 @@ def join(
...
     ) -> Self:
         _supported_joins = ("inner", "left", "cross", "anti", "semi")
 
         if how not in _supported_joins:
             msg = f"Only the following join strategies are supported: {_supported_joins}; found '{how}'."
             raise NotImplementedError(msg)
 
-        if how == "cross" and (left_on or right_on):
-            msg = "Can not pass left_on, right_on for cross join"
+        if how == "cross" and (left_on or right_on or on):
```
MarcoGorelli, line 197: seeing as we're changing this, shall we take the chance to be more explicit and write
```
(left_on is not None or right_on is not None or on is not None)
```
?

narwhals/dataframe.py
```diff
@@ -186,17 +186,31 @@ def join(
...
+            msg = "Can not pass `left_on`, `right_on` or `on` keys for cross join"
             raise ValueError(msg)
 
+        if (how != "cross" and on is None) and (left_on is None or right_on is None):
```
MarcoGorelli, line 201: I can't really make sense of the parentheses like this, maybe it's just me, but how about
```
if how != "cross" and (on is None and (left_on is None or right_on is None)):
```

narwhals/dataframe.py
```diff
@@ -186,17 +186,31 @@ def join(
...
+            msg = f"Either (`left_on` and `right_on`) or `on` keys should be specified for {how}."
+            raise ValueError(msg)
+
+        if (how != "cross" and on is not None) and (
+            left_on is not None or right_on is not None
+        ):
```
MarcoGorelli, line 207: similar to above

narwhals/dataframe.py
```diff
@@ -232,37 +246,24 @@ def join_asof(
...
+        if (on is not None) and (left_on is not None or right_on is not None):
+            msg = "If `on` is specified, `left_on` and `right_on` should be None."
             raise ValueError(msg)
-        if by_left is None and by_right is not None and by is None:
-            msg = "`by_left` can not be None if `by_right` is specified."
-            raise ValueError(msg)
-        if (
+        if (by is None) and (
             (by_left is None and by_right is not None)
             or (by_left is not None and by_right is None)
```
MarcoGorelli, line 257: why not do the same that there is for `on`, i.e.
```
if (by is None) and (by_left is None or by_right is None):
    msg = "Either (`by_left` and `by_right`) or `by` keys should be specified."
```
?

I think it's easier to maintain if the logic for `on/left_on/right_on` is the same as it is for `by/by_left/by_right`

MarcoGorelli (commented):
amazing, thanks @raisadz !

just added some minor nitpicks

MarcoGorelli (approved):
thanks @raisadz !

PR #928, opened 2024-09-09
doc: update drop_nulls for series documentation

LiamConnors:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

It looks like `drop_nans` isn't implemented. 
Was also thinking to update to use an example similar to https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.drop_nulls so it's easy to see the difference between Polars and pandas with `float("nan")`. What do you think @MarcoGorelli? Also, I'm curious - if `drop_nans` was to be implemented, would the expectation be that it would do the same as `drop_nulls` for pandas-like series?

### DIFF
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -783,12 +783,9 @@ class Series:
         """
         Drop all null values.
 
-        See Also:
-          drop_nans
-
         Notes:
-          A null value is not the same as a NaN value.
-          To drop NaN values, use :func:`drop_nans`.
+          pandas and Polars handle null values differently. Polars distinguishes
+          between NaN and Null, whereas pandas doesn't.
 
         Examples:
           >>> import pandas as pd


MarcoGorelli (approved):
well-spotted, thanks @LiamConnors !

> if drop_nans was to be implemented, would the expectation be that it would do the same as drop_nulls for pandas-like series?

I _think_ the expectation is that it would only drop NaN values, and not anything else. So, no `NaT`, no `None`, no `NA`, ...this may require a bit of working around pandas to get it to work

Issue #862, opened 2024-08-24
test: Get all cuDF tests passing


MarcoGorelli:
We don't currently test cuDF in CI (though, hint hint, if anyone from Nvidia is passing by and wants to gift us with GPU minutes in CI...), we just do it manually on a Kaggle notebook: https://www.kaggle.com/code/marcogorelli/testing-cudf-in-narwhals

For a variety of reasons, some tests fail. Some examples:

### Ex. 1 cuDF doesn't implement something
```
FAILED tests/expr_and_series/dt/datetime_attributes_test.py::test_datetime_attributes[cudf_constructor-date-expected0] - AttributeError: 'DatetimeProperties' object has no attribute 'date'
```
Looks like cuDF doesn't implement `.dt.date`. There's not much we can do about this. So, in the test which hits this, we should `xfail` it if the constructor is `cuDF` (see some other test examples for how to xfail using pytest's `request.applymarker`)

### Ex. 2 test just fails
```
tests/expr_and_series/is_duplicated_test.py::test_is_duplicated_expr[cudf_constructor] - AssertionError
```
No details are given here, so this one requires investigation. Note that it also fails for Modin, so we might be using something in the implementation which only works for pandas

EDIT: some (all?) of these might already be resolved

### Ex. 3: cuDF intentionally raises

```
FAILED tests/frame/array_dunder_test.py::test_array_dunder_with_dtype[cudf_constructor] - ValueError: copy=False is not supported because conversion to a numpy array...
```
Like for `Ex. 1` cases, we should just `xfail` these tests

### Ex 4. Runtime Error
```
FAILED tests/expr_and_series/str/contains_test.py::test_contains[cudf_constructor] - RuntimeError: CUDF failure at:/opt/conda/conda-bld/work/cpp/src/strings/reg...
```
This might be a bug in cuDF? It might be worth reproducing it and reporting it to cuDF, and in the meantime, xfailing the test

### Ex 5. `AssertionError: (0.3, 0.3)`

`test_value_counts` shows this. We should investigate why - perhaps cuDF scalars also need porting to Python, like we do with PyArrow (`as_py`)?

LiamConnors:
I'd like to try working through these if that's ok.

MarcoGorelli:
of course, thanks @LiamConnors ! and please do ask if anything trips you up, either here or on Discord

LiamConnors:
[...]

PR #937, opened 2024-09-10
fix: Update `copy` param on `to_numpy` to default to `True` for cuDF

LiamConnors:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue [#862](https://github.com/narwhals-dev/narwhals/issues/862)
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

This PR fixes issue seen in tests:
` ValueError: copy=False is not supported because conversion to a numpy array always copies the data.`

And reduces the number of failing tests by 7.

@MarcoGorelli the cuDF docs show `copy` defaults to `True` on cuDF. So I think that it should set `copy=True` here when it's `None` for cuDF. What do you think - Is that the right approach?

https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.series.to_numpy/
https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.dataframe.to_numpy/

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -590,8 +590,8 @@ class PandasLikeDataFrame:
         from narwhals._pandas_like.series import PANDAS_TO_NUMPY_DTYPE_MISSING
 
         if copy is None:
-            # pandas default differs from Polars
-            copy = False
+            # pandas default differs from Polars, but cuDF default is True
+            copy = self._implementation is Implementation.CUDF
 
         if dtype is not None:
             return self._native_frame.to_numpy(dtype=dtype, copy=copy)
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
thanks @LiamConnors !

agree - we don't expose `copy` in `to_numpy` anyway (https://narwhals-dev.github.io/narwhals/api-reference/series/#narwhals.series.Series.to_numpy) so 👍 to just use the library's default

PR #940, opened 2024-09-10
feat: Dask multiple partitions

### LINKED ISSUE #872: test: run Dask tests with `npartitions=2`
If we use `npartitions=2` in the Dask constructor, some tests fail

I think this is just in cases when Dask may shuffle row order, which we can probably resolve by including an `'i'` column which just enumerates the rows, and then sorting by that at the end

This would also help make the test suite more resilient for future lazy backends

---

I'll add some more details later, sorry if this is unclear

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #872 

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -79,7 +79,9 @@ class DaskLazyFrame:
             and isinstance(predicates[0], list)
             and all(isinstance(x, bool) for x in predicates[0])
         ):
-            msg = "Filtering with boolean mask is not supported for `DaskLazyFrame`"
+            msg = (
+                "`LazyFrame.filter` is not supported for Dask backend with boolean masks."
+            )
             raise NotImplementedError(msg)
 
         from narwhals._dask.namespace import DaskNamespace


narwhals/_dask/dataframe.py
```diff
@@ -79,7 +79,9 @@ def filter(
             and isinstance(predicates[0], list)
             and all(isinstance(x, bool) for x in predicates[0])
         ):
-            msg = "Filtering with boolean mask is not supported for `DaskLazyFrame`"
+            msg = (
+                "`LazyFrame.filter` is not supported for Dask backend with boolean masks."
```
FBruzzesi, line 83: Aligns error message with others. User should not really know about `DaskLazyFrame`

FBruzzesi:
CI Failure is due to a `polars==0.20.31` deprecation warning

PR #942, opened 2024-09-10
feat: add `mode` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

I started to take a very mild look at skrub... and immediately stumbled upon not having `mode` implemented
  
## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -22,6 +22,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from narwhals._arrow.dataframe import ArrowDataFrame
+    from narwhals._arrow.namespace import ArrowNamespace
     from narwhals.dtypes import DType
 
 
@@ -64,6 +65,11 @@ class ArrowSeries:
             name=name,
             backend_version=backend_version,
         )
+
+    def __narwhals_namespace__(self) -> ArrowNamespace:
+        from narwhals._arrow.namespace import ArrowNamespace
+
+        return ArrowNamespace(backend_version=self._backend_version)
 
     def __len__(self) -> int:
         return len(self._native_series)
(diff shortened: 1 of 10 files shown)


narwhals/_arrow/series.py
```diff
@@ -667,6 +673,13 @@ def clip(
     def to_arrow(self: Self) -> pa.Array:
         return self._native_series.combine_chunks()
 
+    def mode(self: Self) -> ArrowSeries:
+        plx = self.__narwhals_namespace__()
+        col_token = generate_unique_token(n_bytes=8, columns=[self.name])
+        return self.value_counts(name=col_token, normalize=False).filter(
+            plx.col(col_token) == plx.col(col_token).max()
+        )[self.name]
```
FBruzzesi, line 681: I was not able to make it work using [`pyarrow.compute.mode`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.mode.html) directly as it requires to know how many distinct most-common values to return.

narwhals/_dask/expr.py
```diff
@@ -644,6 +644,10 @@ def func(df: DaskLazyFrame) -> list[Any]:
             backend_version=self._backend_version,
         )
 
+    def mode(self: Self) -> Self:
+        msg = "`Expr.mode` is not supported for the Dask backend."
+        raise NotImplementedError(msg)
```
FBruzzesi, line 649: Of course, mode changes the index (as it changed the length)

MarcoGorelli (approved):
thanks @FBruzzesi !

FBruzzesi:
Crazy timing! Polars just changed behaviour on mode with multiple columns.

From the docs itself:

```py
df = pl.DataFrame(
    {
        "a": [1, 1, 2, 3],
        "b": [1, 1, 2, 2],
    }
)
df.select(pl.all().mode())
```

- For polars before 1.7, result is as in the test we have
- For polars 1.7 it raises   `InvalidOperationError: Series: a, length 1 doesn't match the DataFrame height of 2`. 

Should we revert this PR?

MarcoGorelli:
i think we can just remove the multi-column test - not sure if it's a bug in Polars, will check

PR #956, opened 2024-09-12
ci: enable uv cache, use astral-sh official action

### LINKED ISSUE #955: [Enh]: Use offical `astral-sh/setup-uv@v2` github action and enable caching in the CI

No response

- We should use an official github action to install `uv` onto the running CI platform
- We can speed up the `uv install` CI step by caching the dependencies, since `requirements-dev.txt` does not change so often.

```diff
...
        with:
          python-version: ${{ matrix.python-version }}
-     - name: Install uv (Unix)
-       if: runner.os != 'Windows'
-       run: curl -LsSf https://astral.sh/uv/install.sh | sh
-     - name: Install uv (Windows)
-       if: runner.os == 'Windows'
-       run: powershell -c "irm https://astral.sh/uv/install.ps1 | iex"
+     - name: Install the latest version of uv
+       uses: astral-sh/setup-uv@v2
+       with:
+         version: "latest"
+     - name: Enable caching
+       id: setup-uv
+       uses: astral-sh/setup-uv@v2
+       with:
+         enable-cache: true
+         cache-suffix: "reqs-cache"
      - name: install-reqs
...
```

_No response_

_No response_

aidoskanapyanov:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #955

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Also added `cache-dependency-glob` pattern to revalidate the cache whenever the `requirements-dev.txt` file is changed. I hope this will work, gotta test it :)

### DIFF
--- a/.github/workflows/check_docs_build.yml
+++ b/.github/workflows/check_docs_build.yml
@@ -18,8 +18,12 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install uv (Unix)
-        run: curl -LsSf https://astral.sh/uv/install.sh | sh
+      - name: Install uv
+        uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "**requirements*.txt"
       - name: install-reqs
         run: uv pip install --upgrade tox virtualenv setuptools pip -r requirements-dev.txt --system
       - name: install-docs-reqs
--- a/.github/workflows/check_tpch_queries.yml
+++ b/.github/workflows/check_tpch_queries.yml
@@ -18,8 +18,12 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install uv 
-        run: curl -LsSf https://astral.sh/uv/install.sh | sh
+      - name: Install uv
+        uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "**requirements*.txt"
       - name: install-reqs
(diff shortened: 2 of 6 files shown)


aidoskanapyanov:
This is how `uv` cached the dependencies, I also added a `python-version` suffix to distinguish between versions:
https://github.com/narwhals-dev/narwhals/actions/caches

Shall we trigger CI rerun and see if it cuts the `install-reqs` step?

MarcoGorelli:
did it work? shall we ship it?

FBruzzesi:
> did it work? shall we ship it?

I honestly could not spot much of the difference, but still would prefer to use the _official_ action rather then a manual installation. 
From the [uv github section](https://docs.astral.sh/uv/guides/integration/github/), it seems we could use uv to install run tests as well (happy to keep it as a followup)

MarcoGorelli (approved):
agree - let's do this then, thanks @aidoskanapyanov !

aidoskanapyanov:
image

I pulled up statistics for `pytest-38 (3.8, ubuntu-latest)` job in `PyTest` workflow for the last  60-ish days via `gh api`.
And it turns out there's really no difference in using cache for installing dependencies. If anything this may have slowed down the run times :/

FBruzzesi:
I guess that uv is faster than building, checking and grabbing from a cache then 😂 Should we revert this one?

aidoskanapyanov:
Yeah, I think we need to revert it, installing from `https://astral.sh/uv/install.sh` looked fine. Seems like a pretty permanent url 😂
Also, cache entries are piling up with each CI run in https://github.com/narwhals-dev/narwhals/actions/caches.
Some of them are being used, but I don't think it makes much of a different in performace.

aidoskanapyanov:
> If anything this may have slowed down the run times 
Could also be that we've added 3 need dependencies since then

PR #957, opened 2024-09-12
Bug: fix an issue with string keys in left and semi joins in Dask, add Dask to tpch benchmarks

raisadz:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [ ] Documented the changes

 - Fix an issue with string keys in left and semi joins in Dask;
 - Added an error message of `Expr.unique` is not supported for the Dask backend as it modifies the index;
 - Added Dask to tpch benchmarks;

### DIFF
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -254,13 +254,20 @@ class DaskLazyFrame:
             )
 
         if how == "semi":
-            other_native = (
-                other._native_frame.loc[:, right_on]
-                .rename(  # rename to avoid creating extra columns in join
-                    columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
+            if isinstance(left_on, str) and isinstance(right_on, str):
+                other_native = (
+                    other._native_frame.loc[:, right_on]
+                    .to_frame(name=left_on)
+                    .drop_duplicates()  # avoids potential rows duplication from inner join
                 )
-                .drop_duplicates()  # avoids potential rows duplication from inner join
-            )
+            else:
+                other_native = (
+                    other._native_frame.loc[:, right_on]
+                    .rename(  # rename to avoid creating extra columns in join
+                        columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
+                    )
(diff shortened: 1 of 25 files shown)


narwhals/_dask/dataframe.py
```diff
@@ -254,13 +254,20 @@ def join(
...
-                other._native_frame.loc[:, right_on]
-                .rename(  # rename to avoid creating extra columns in join
-                    columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
+            if isinstance(left_on, str) and isinstance(right_on, str):
+                other_native = (
+                    other._native_frame.loc[:, right_on]
+                    .to_frame(name=left_on)
+                    .drop_duplicates()  # avoids potential rows duplication from inner join
+                )
+            else:
```
FBruzzesi, line 263: This fails for a case such as:

```py
...
right_on = ["abc"]
left_on = "abc"
```

I would suggest to convert the keys from `str` to `list[str]` at the beginning on the method and reason with them being lists.

Similarly below the check `if isinstance(left_on, str) and isinstance(right_on, str)` would fail for this such case and need to check them separately as we do in [pandas land](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/_pandas_like/dataframe.py#L419)

FBruzzesi (commented):
Thanks for the PR, and adjusting the tpch queries to handle the dask unique case.

I left a major comment that will need to be addressed though 😇

narwhals/_dask/dataframe.py
raisadz: Thanks  for your review @FBruzzesi , I added a conversion from `str` to `list[str]` at the beginning of `join` for Dask

FBruzzesi (approved):
Thanks @raisadz 🙌🏼

NarwhalChubbiwhalGIF

MarcoGorelli:
awesome work here, thanks both! 🙌

PR #960, opened 2024-09-13
feat: `Datetime(time_unit, time_zone)` and `Duration(time_unit)` types

### LINKED ISSUE #467: [Enh]: Add time units and time zone specifics

See this [snippet](https://github.com/narwhals-dev/narwhals/blob/c904b102bd95dec026ce756cecfda15dabe71f41/narwhals/_pandas_like/utils.py#L287-L296)

```python
    if str(dtype).startswith("datetime64"):
        # TODO(Unassigned): different time units and time zones
        return dtypes.Datetime()
    if str(dtype).startswith("timedelta64") or str(dtype).startswith("duration"):
        # TODO(Unassigned): different time units
        return dtypes.Duration()
    if str(dtype).startswith("timestamp["):
        # pyarrow-backed datetime
        # TODO(Unassigned): different time units and time zones
        return dtypes.Datetime()
```

This is related to Ruff's [DTZ001](https://docs.astral.sh/ruff/rules/call-datetime-without-tzinfo/)

To fix this issue, look for ocurrences of `467` in the code

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #467 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Introduces time units and time zones in `Datetime` type.


tests/series_only/cast_test.py
```diff
@@ -75,13 +75,13 @@ def test_cast_date_datetime_pandas() -> None:
     df = df.select(nw.col("a").cast(nw.Datetime))
     result = nw.to_native(df)
     expected = pd.DataFrame({"a": [datetime(2020, 1, 1), datetime(2020, 1, 2)]}).astype(
-        {"a": "timestamp[ns][pyarrow]"}
+        {"a": "timestamp[us][pyarrow]"}
```
FBruzzesi, line 78: Changes the default to the polars one

tests/expr_and_series/cast_test.py
```diff
@@ -179,3 +185,27 @@ class Banana:
...
+    expected = {
+        "date": ["2024-01-01 01:00:00", "2024-01-02 01:00:00", "2024-01-03 01:00:00"]
+    }
+
+    df = nw.from_native(constructor(data))
+    result = df.select(
+        nw.col("date")
+        .cast(nw.Datetime("ms", time_zone="Europe/Rome"))
+        .cast(nw.String())
+        .str.slice(offset=0, length=19)
```
FBruzzesi, line 209: 19: number of characters of `"2024-01-01 01:00:00"`. The format right after that is different for each backend

narwhals/_pandas_like/utils.py
```diff
@@ -221,6 +223,12 @@ def translate_dtype(column: Any) -> DType:
     from narwhals import dtypes
 
     dtype = column.dtype
+
+    pd_datetime_rgx = (
+        r"^datetime64\[(?P<time_unit>ms|us|ns)(?:, (?P<time_zone>[a-zA-Z\/]+))?\]$"
+    )
+    pa_datetime_rgx = r"^timestamp\[(?P<time_unit>ms|us|ns)(?:, tz=(?P<time_zone>[a-zA-Z\/]+))?\]\[pyarrow\]$"
```
FBruzzesi, line 230: Please try to break these 🙈

narwhals/_pandas_like/utils.py
```diff
@@ -425,10 +435,25 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
+        time_zone = getattr(dtype, "time_zone", None)
+
+        # Pandas does not support "ms" or "us" time units before version 1.5.0
+        # Let's overwrite with "ns"
+        if implementation is Implementation.PANDAS and backend_version < (
+            1,
+            5,
+            0,
+        ):  # pragma: no cover
+            time_unit = "ns"
```
FBruzzesi, line 448: I don't think we can do much else here

FBruzzesi:
@MarcoGorelli any experience on how to locate timezone database at "C:\Users\runneradmin\Downloads\tzdata" ? 😂

MarcoGorelli:
wow, nice one! i'll try breaking it a bit, but this is amazing, been wanting to do this for a while 🚀

MarcoGorelli:
I think the dtype comparison isn't quite right:
```python
In [17]: s
Out[17]:
shape: (1,)
Series: '' [datetime[μs, Asia/Kathmandu]]
[
        2019-12-31 18:15:00 +0545
]

In [18]: nw.from_native(s, allow_series=True).dtype == nw.Datetime('us')
Out[18]: True

In [19]: s.dtype == pl.Datetime('us')
Out[19]: False
```

FBruzzesi:
> I think the dtype comparison isn't quite right:
> 
> ```python
> In [17]: s
> Out[17]:
> shape: (1,)
> Series: '' [datetime[μs, Asia/Kathmandu]]
> [
>         2019-12-31 18:15:00 +0545
> ]
> 
> In [18]: nw.from_native(s, allow_series=True).dtype == nw.Datetime('us')
> Out[18]: True
> 
> In [19]: s.dtype == pl.Datetime('us')
> Out[19]: False
> ```

Nice catch! Thanks! Will fix later on 👌

MarcoGorelli:
🤔 interesting, so `elif dtype in {nw.Datetime, nw.Date}:` breaks if we add the time_unit and time_zone attributes

narwhals/dtypes.py
```diff
@@ -71,10 +73,97 @@ class Object(DType): ...
...
+        # allow comparing object instances to class
+        if type(other) is type and issubclass(other, Datetime):
+            return True
+        elif isinstance(other, Datetime):
+            return self.time_unit == other.time_unit and self.time_zone == other.time_zone
+        else:
+            return False
+
+    def __hash__(self: Self) -> int:  # pragma: no cover
+        return hash((self.__class__, self.time_unit, self.time_zone))
```
MarcoGorelli, line 120: do we need to keep this as `return hash(self.__class__)` so that
```
nw.Datetime('us', 'foo') in {nw.Datetime}
```
keeps working?

It's unfortunate that I'd written the code in Altair like that, but I also don't think it's too bad, it's just a little annoying that it departs from what Polars does

FBruzzesi: Down to only one altair test failing 👀

FBruzzesi: Oh I see, this is due to the change of `__repr__`. The following assertion fails in the altair test:

```py
('Field "timedelta" has type "Duration" which is not supported by Altair'
 in 'Field "timedelta" has type "Duration(time_unit=\'ns\')" which is not supported by Altair. Please ...'
).
```

MarcoGorelli: yup - I've opened a PR there to make the test a little laxer https://github.com/vega/altair/pull/3606

FBruzzesi:
> this _might_ actually be a really good use-case for our stable v1 api. keep Altair's code working as-is, but make the `__hash__` behaviour align with Polars' in the main Narwhals namespace and in v2 (when we get there)

I wonder how that would look like concretely. Would we re-define the type in `/stable/v1.py` with the current behavior and _evolve_ it in the main codebase?

FBruzzesi:
As altair CI is failing (`dtype in {nw.Datetime, nw.Date}` is evaluated to false because of the equality method), I think to easiest way for stability is to have the previous implementation of Datetime and Duration in `/stable/v1.py`.

@MarcoGorelli  WDYT?

MarcoGorelli:
i think this isn't super-easy because there's a few places where we do `from narwhals import dtypes`

MarcoGorelli:
ok this is going to take some more effort
```
(Pdb) p type(column)
<class 'narwhals.stable.v1.Series'>
(Pdb) p type(column.dtype)
<class 'narwhals.dtypes.Datetime'>
```
I think the second one should be
```
<class 'narwhals.stable.v1.Datetime'>
```

this is going to be tricky to get right, but I think it'll be worth it. i'll spend some more time on this

FBruzzesi:
Could we use the approach suggested in #1046 ?

MarcoGorelli:
Looks like this might be it 🥳  🍾 can't believe it...this took _hours_...worth it though. It means we can change the main narwhals namespace, with _zero_ impact on Altair users. as it turns out, the stable api was really worth doing... this is so rewarding 🕺 

i think the nightly ci failure is unrelated

i'll check this again tomorrow, then hopefully we can make a release with this in on Tuesday

FBruzzesi:
That's awesome! I will take some time this upcoming week to check what happened in detail!
Also plotly has a test with specific time-zone, thus I am looking forward to the release of this feature 😁

However, to use the "edge" dtypes (or in general features), should we

```diff
+ import narwhals as nw
- import narwhals.stable.v1 as nw
```
?

narwhals/_pandas_like/utils.py
```diff
@@ -213,6 +215,15 @@ def set_axis(
...
 def native_to_narwhals_dtype(column: Any, dtypes: DTypes) -> DType:
     dtype = str(column.dtype)
+
+    pd_datetime_rgx = (
+        r"^datetime64\[(?P<time_unit>ms|us|ns)(?:, (?P<time_zone>[a-zA-Z\/]+))?\]$"
+    )
+    pa_datetime_rgx = r"^timestamp\[(?P<time_unit>ms|us|ns)(?:, tz=(?P<time_zone>[a-zA-Z\/]+))?\]\[pyarrow\]$"
+
+    pd_duration_rgx = r"^timedelta64\[(?P<time_unit>ms|us|ns)\]$"
+    pa_duration_rgx = r"^duration\[(?P<time_unit>ms|us|ns)\]\[pyarrow\]$"
```
MarcoGorelli, line 225: pandas / pyarrow support 'second' time unit, I think that should be allowed to pass through

MarcoGorelli:
> However, to use the "edge" dtypes (or in general features), should we

it's fine to use time_unit / time_zone with stable.v1, it's just probably a good idea to avoid anything which calls `__hash__`. For example, instead of `series.dtype in {nw.Datetime, nw.Date}`, it'd be safer to do `series.dtype == nw.Datetime or series.dtype == nw.Date`

narwhals/_pandas_like/utils.py
FBruzzesi: Just by passing it along or doing manipulation for the user?

MarcoGorelli: i think we can just pass it through - adding a commit soon

narwhals/_pandas_like/utils.py
```diff
@@ -416,15 +433,32 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
-        # TODO(Unassigned): different time units and time zones
-        if dtype_backend == "pyarrow-nullable":
-            return "duration[ns][pyarrow]"
-        return "timedelta64[ns]"
+        du_time_unit = getattr(dtype, "time_unit", "us")
+        return (
+            f"duration[{du_time_unit}][pyarrow]"
+            if dtype_backend == "pyarrow-nullable"
+            else f"timedelta64[{du_time_unit}]"
+        )
```
MarcoGorelli, line 460: we need to do the same pre-1.5.0 check here

narwhals/_pandas_like/utils.py
```diff
@@ -416,15 +433,32 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
+        dt_time_zone = getattr(dtype, "time_zone", None)
+
+        # Pandas does not support "ms" or "us" time units before version 1.5.0
+        # Let's overwrite with "ns"
+        if implementation is Implementation.PANDAS and backend_version < (
+            1,
+            5,
+            0,
+        ):  # pragma: no cover
+            dt_time_unit = "ns"
```
MarcoGorelli, line 446: 🤔 we might want to make this 2.0, the 1.5 support was pretty patchy

MarcoGorelli (approved):
alright, happy with this 🥳 let's ship it?

FBruzzesi:
This is the cutest gif so far! How did you discover it?
FYI: @anopsy

Issue #386, opened 2024-07-02
Use IntoDataFrame in docs


MarcoGorelli:
We now have a `IntoDataFrame` type hint, which means "either a Narwhals dataframe, or something convertible to a Narwhals DataFrame"

I think we should use it in the tutorial / user guide

I'm not totally sure what whether we should use it in docstrings

Given that this tool is aimed at library maintainers, perhaps we should err on the side of including static typing in as many docs as possible?

cc @baggiponte the typing expert for thoughts?

baggiponte:
Ciao Marco, sorry for the late reply. 

> I'm not totally sure what whether we should use it in docstrings

Why so?

MarcoGorelli:
it would require an extra `from narwhals.typing import IntoDataFrame`, and to people who don't like type annotations it might be it look more complicated than it needs to be

dunno, maybe it'd still be better though, people who don't like type hints can just ignore them, if they don't want their IDE to help them it's on them

baggiponte:
> dunno, maybe it'd still be better though, people who don't like type hints can just ignore them, if they don't want their IDE to help them it's on them

I agree with this one. It'd offer a superior experience to the end user. After all, it's just a single type hint, is it not?

luke396:
I'd like to start and work on this.

MarcoGorelli:
I heard from someone recently that `narwhalify` isn't super-clear in the docstrings to a beginner, and that it's not clear what the input and output are of the various `func` examples

I feel like suggesting:
- rename `func` to `my_library_agnostic_function` in all docstrings
- remove `narwhalify` from the docstrings (except for the `narwhalify` docstring), and instead use `from_native` / `to_native`, and add type annotations to each argument

PR #968, opened 2024-09-14
docs: Use `IntoDataFrame` in docs

luke396:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #386
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/docs/basics/dataframe.md
+++ b/docs/basics/dataframe.md
@@ -3,7 +3,8 @@
 To write a dataframe-agnostic function, the steps you'll want to follow are:
 
 1. Initialise a Narwhals DataFrame or LazyFrame by passing your dataframe to `nw.from_native`.
-    All the calculations stay lazy if we start with a lazy dataframe - Narwhals will never automatically trigger computation without you asking it to.
+    All the calculations stay lazy if we start with a lazy dataframe - Narwhals will never
+    automatically trigger computation without you asking it to.
    
     Note: if you need eager execution, make sure to pass `eager_only=True` to `nw.from_native`.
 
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
thanks @luke396 ! I think these ones are fine, I was thinking more about using these type hints in the docstrings

luke396:
> thanks @luke396 ! I think these ones are fine, I was thinking more about using these type hints in the docstrings

Ah, I notice some differences between the user guide/tutorial and the docstring. Are the changes in the tutorial necessary? If so, I'd like to make further modifications; if not, closing the PR is perfectly acceptable.

Regarding the `IntoDataFrame` and `IntoDataFrameT` in the method's docstring, I plan to look into them later.

MarcoGorelli:
I think the tutorial is probably the first thing that a lot of people will see, so improving it is certainly a good idea 👍 If there are places in the tutorial which don't have type annotations, then I think adding them would be beneficial

luke396:
Absolutely, adding type annotations where they're missing is a great idea.

I'm slightly confused about other type annotations, especially those changed in the current PR. To be honest, I'm not entirely sure which ones to use in the tutorial. I've read their definitions, and it seems they have some inclusion relationships or minor differences in certain subtypes. So, which one to choose is a question.

My question is, should we update them in the tutorial based on a specific principle? Or have we already settled on annotations, and I've just missed something?

PR #993, opened 2024-09-18
perf: Impove `pandas`-backend indexing

### LINKED ISSUE #992: [enh]: Pandas indexing is slow

_No response_

As it currently stands the `pandas`-backend indexing operation is very slow in comparison to the native one, that is to say - the difference is `~6-9x` (depending on the run) worse for narwhals

```python
import pandas as pd
df = pd.DataFrame({'a': [1, 2, 3]})
df['a']
```

is much faster than

```python
import pandas as pd
import narwhals as nw
df = pd.DataFrame({'a': [1, 2, 3]})

nw_df = nw.from_native(df)
nw_df['a']
```

for reference am attaching the profiles of both indexing operations

native `pandas`
Screenshot 2024-09-18 at 13 18 01

narwhals `pandas`-backend
Screenshot 2024-09-18 at 13 16 13

as one can see native pandas does some caching, which is not the case in the narwhals case as the backend uses `loc` and `iloc` ubiquitously in the implementation of `__getitem__`, see this [here](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/_pandas_like/dataframe.py#L131-L199), which as evident are slow

Since the difference in `df[col]` and `df.loc[:, col]` are in assignment, and not assignemt is possible in `narwhals` in the form of `nw_df[col] = some_series`, this is a slowdown, which is unnecessary

Similar issues have `PandasLikeDataframe.select` and `PandasLikeDataFrame.with_columns`

`__getitem__` needs to be reworked to use as much as possible the cache in the pandas dataframe object

_No response_

_No response_

aivanoved:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #992 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


MarcoGorelli:
ooh nice one @aivanoved ! if we can get a perf improvement from this and it's safe, then it seems like a no-brainer

aivanoved:
@MarcoGorelli It would be very nice! This is still in draft as am still checking that it is safe

aivanoved:
I don't think there are any safety issues, at the same time thinking about how to test this has been slightly deadend-y. @MarcoGorelli any ideas for tests?

MarcoGorelli:
I think this should be safe. We don't have any in-place operations in the API anyway - the only one is `Series.scatter`, and we really warn users to be careful with that one. And the scatter tests still pass here

FBruzzesi:
Hey @aivanoved thanks for this improvement!
It seems that there are other pandas-like part of the codebase that use `loc[:, ...]`, should we migrate those as well before merging the changes?

aivanoved:
@FBruzzesi I'll take a look at them again, as I was only looking at `__getitem__` specifically

If they are used in assignments, most likely they are meaningful

aivanoved:
I have removed all which are not involved in assignments, also have left the dask backend untouched, as there I am not sure what the implications of the `loc` semantics

MarcoGorelli:
thanks @aivanoved 

tbh i'm worried about trusting pandas here for anything more complicated than `df[column_name]`. for example:
```python
>>> df = pd.DataFrame({True: [1, 2], False: [3, 4]})
>>> df[[True,False]]
   True   False
0      1      3
```

For just taking out a single column into a Series (which is probably what we do most often), then it should be safe - shall we just stick to that for now?

MarcoGorelli (approved):
thanks a tonne @aivanoved for having spotted this and submitted a fix!

PR #995, opened 2024-09-19
test: add specific check for cuDF for test_to_datetime

LiamConnors:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # https://github.com/narwhals-dev/narwhals/issues/862
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Currently this test fails with 
```
FAILED tests/expr_and_series/str/to_datetime_test.py::test_to_datetime[cudf_constructor] - AssertionError: assert '2020-01-01T1...:56.000000000' == '2020-01-01 12:34:56'
```
I created these in the native libraries and I see they have different data types and render differently when converted back to a string.
```
import pandas as pd
import cudf

data = {"a": ["2020-01-01T12:34:56"]}

df_cudf = pd.DataFrame(data)
df_pandas = pd.DataFrame(data)

df_pandas_dt = pd.to_datetime(df_pandas.a, format="%Y-%m-%dT%H:%M:%S")
df_cudf_dt = cudf.to_datetime(df_cudf.a, format="%Y-%m-%dT%H:%M:%S")

print(df_pandas_dt[0])
print(type(df_pandas_dt[0]))
print(df_cudf_dt[0])
print(type(df_cudf_dt[0]))
```

```
2020-01-01 12:34:56
<class 'pandas._libs.tslibs.timestamps.Timestamp'>
2020-01-01T12:34:56.000000000
<class 'numpy.datetime64'>
```

Not sure though if this is an acceptable difference and something to just account for in the tests like this though.

### DIFF
--- a/tests/expr_and_series/str/to_datetime_test.py
+++ b/tests/expr_and_series/str/to_datetime_test.py
@@ -5,6 +5,11 @@ data = {"a": ["2020-01-01T12:34:56"]}
 
 
 def test_to_datetime(constructor: Constructor) -> None:
+    if "cudf" in str(constructor):
+        expected = "2020-01-01T12:34:56.000000000"
+    else:
+        expected = "2020-01-01 12:34:56"
+
     result = (
         nw.from_native(constructor(data))
         .lazy()
@@ -12,4 +17,4 @@ def test_to_datetime(constructor: Constructor) -> None:
         .collect()
         .item(row=0, column="b")
     )
-    assert str(result) == "2020-01-01 12:34:56"
+    assert str(result) == expected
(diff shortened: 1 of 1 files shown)


LiamConnors:
@MarcoGorelli @FBruzzesi wasn't sure if this is the right approach. What do you think? Should we just test for the difference or try and get the output being the same? I see on one of the pages we mention trailing zeroes are possible when it comes to times.

MarcoGorelli (approved):
thanks @LiamConnors ! I think this is probably fine, looks like cudf returns a numpy datetime64 scalar here with 'ns' precision, and you're right, we already mention that `strftime` may vary slightly across implementations

if these differences cause issues to anyone we can look into working around them

PR #1014, opened 2024-09-20
feat: adding to_native to dataframe

### LINKED ISSUE #1001: enh: implement `DataFrame.to_native`
Currently, we can call `nw.to_native`

It would be more user-friendly to have `DataFrame.to_native` though, I keep finding myself reaching for this

Contributing guide: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md

Cheukting:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1001 
- Closes #1001 

## Checklist

- [X] Tests added 
- [X] Documented the changes

As describe in #1001 to_native() method is added to dataframe objects


MarcoGorelli (commented):
awesome, thanks @Cheukting ! just some comments

tests/frame/to_native_test.py
```diff
@@ -0,0 +1,11 @@
...
+
+import narwhals.stable.v1 as nw
+
+
+def test_to_native(constructor_eager: Any) -> None:
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.1, 8, 9]}
+    df_raw = constructor_eager(data)
+    df = nw.from_native(df_raw, eager_only=True)
+
+    assert isinstance(df.to_native(), df_raw.__class__)
```
mikeweltevrede, line 11: I do believe that this does not exactly check what you need it to check because if the type of `df.tonative()` is actually a subclass of `df_raw.__class__`, it will also pass. While usually `isinstance` is preferred over `type()` comparison, I think it does warrant implementation here: `assert type(df.to_native()) == df_raw.__class__`, where you would need to put `# noqa: E721`.

PS: also see https://stackoverflow.com/a/9611031/10367342

tests/frame/to_native_test.py
```diff
@@ -0,0 +1,11 @@
+from typing import Any
+
+import narwhals.stable.v1 as nw
+
+
+def test_to_native(constructor_eager: Any) -> None:
```
mikeweltevrede, line 6: It would be nice to make the test name a bit more descriptive, e.g. `test_to_native_returns_object_of_original_class_type`

MarcoGorelli (approved):
thanks @Cheukting , great to have you as contributor!

MarcoGorelli (approved):
sorry, forgot you wanted a dog

tests/frame/to_native_test.py
MarcoGorelli: true, but none of the dataframes we support are subclasses of each other

PR #1019, opened 2024-09-20
test: use pytest-randomly, recommend pytest-xdist in docs

TNieuwdorp:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [X] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [X] Tests added 
- [X] Documented the changes

After discussion with Francesco he deemed it okay to implement the multithreaded variant to speed up the 3800 unit tests on any system that has more than 1 core. Although pytest is not designed with multithreading in mind, all tests complete succesfully.


MarcoGorelli:
thanks @TNieuwdorp for your pr!

i'm slightly hesistant about enabling this by default, as running pytest multiprocessed means you can't run it with `--pdb` (something I do very often...you can do `-n 0`, but still, it's an extra command to remember when debugging)

maybe we could just enable this one (and `pytest-randomly` while we're here) but only in CI?

@FBruzzesi any thoughts on this?

FBruzzesi:
> thanks @TNieuwdorp for your pr!
> 
> i'm slightly hesistant about enabling this by default, as running pytest multiprocessed means you can't run it with `--pdb` (something I do very often...you can do `-n 0`, but still, it's an extra command to remember when debugging)
> 
> maybe we could just enable this one (and `pytest-randomly` while we're here) but only in CI?
> 
> @FBruzzesi any thoughts on this?

I don't think in CI makes much of a difference as runners have 2 cores and there is some overhead to account for - I am double checking by comparing the numbers with other PRs pytest runtime.

Maybe we can mention in the docs how to speed up testing locally and that's all for now?

MarcoGorelli (approved):
thanks @TNieuwdorp !

Issue #1004, opened 2024-09-20
docs: separate out DataFrame from LazyFrame in "api-completeness"


MarcoGorelli:
This page makes it look like we hardly support anything for Dask: https://narwhals-dev.github.io/narwhals/api-completeness/dataframe/

However, if you scroll down to `LazyFrame`, you'll see that actually, we do 😎 We should make a separate LazyFrame page, distinct from the DataFrame one

jacobInav:
I'd like to check this one out

PR #1021, opened 2024-09-20
docs: Add DataFrame.pipe to API completeness table

### LINKED ISSUE #1006: Mark `pipe` as supported in api-completeness
This page makes it look like we don't support `pipe` for any backend: https://narwhals-dev.github.io/narwhals/api-completeness/dataframe/

However, we do 😎 That's because we implement it directly in `narwhals/dataframe.py` - can we special-case this one so it shows as completed for all backends?

file to look at: https://github.com/narwhals-dev/narwhals/blob/main/utils/generate_backend_completeness.py

contributing guide: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md

Dosenpfand:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 
 
- Closes #1006 

## Checklist

- [N/A] Tests added 
- [X] Documented the changes

### DIFF
--- a/utils/generate_backend_completeness.py
+++ b/utils/generate_backend_completeness.py
@@ -38,6 +38,8 @@ BACKENDS = [
 
 EXCLUDE_CLASSES = {"BaseFrame", "Then", "When"}
 
+DIRECTLY_IMPLEMENTED_CLASS_METHODS = {"DataFrame": ["pipe"]}
+
 
 def get_class_methods(kls: type[Any]) -> list[str]:
     return [m[0] for m in inspect.getmembers(kls) if not m[0].startswith("_")]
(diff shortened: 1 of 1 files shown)


utils/generate_backend_completeness.py
```diff
@@ -38,6 +38,8 @@ class Backend(NamedTuple):
 
 EXCLUDE_CLASSES = {"BaseFrame", "Then", "When"}
 
+DIRECTLY_IMPLEMENTED_CLASS_METHODS = {"DataFrame": ["pipe"]}
```
FBruzzesi, line 41: This should hold also for Series and Expr if possible 😇

Dosenpfand: Got it, I just adapted it.

MarcoGorelli (approved):
thanks @Dosenpfand  and @FBruzzesi for reviewing!

this is great

it still shows `DataFrame.pipe` as being implemented for Dask, which isn't quite correct, but if you fancied addressing https://github.com/narwhals-dev/narwhals/issues/1004, then that would solve it

thanks! 🙏

PR #1041, opened 2024-09-22
fix: add cuDF specific implementation for join how="anti"

LiamConnors:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # https://github.com/narwhals-dev/narwhals/issues/862
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

Right now these tests fail with
`NotImplementedError: Only indicator=False is currently supported`
So the current implementation of `df.join` `how="anti"` won't work.

https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.dataframe.merge/

Seemed like best to mark it as expected to fail for now. @MarcoGorelli what do you think would make sense here. Maybe a note in the docs and a follow up issue to see if we can implement it in another way for cuDF?

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -485,17 +485,27 @@ class PandasLikeDataFrame:
                 )
                 .drop_duplicates()
             )
-            return self._from_native_frame(
-                self._native_frame.merge(
-                    other_native,
-                    how="outer",
-                    indicator=indicator_token,
-                    left_on=left_on,
-                    right_on=left_on,
+            if self._implementation is Implementation.CUDF:
+                return self._from_native_frame(
+                    self._native_frame.merge(
+                        other_native,
+                        how="leftanti",
+                        left_on=left_on,
+                        right_on=left_on,
+                    )
                 )
-                .loc[lambda t: t[indicator_token] == "left_only"]
-                .drop(columns=indicator_token)
-            )
+            else:
+                return self._from_native_frame(
+                    self._native_frame.merge(
+                        other_native,
+                        how="outer",
+                        indicator=indicator_token,
+                        left_on=left_on,
+                        right_on=left_on,
+                    )
+                    .loc[lambda t: t[indicator_token] == "left_only"]
+                    .drop(columns=indicator_token)
+                )
 
         if how == "semi":
             other_native = (
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -485,7 +485,7 @@ class PandasLikeDataFrame:
                 )
                 .drop_duplicates()
             )
-            if self._implementation is Implementation.CUDF:
+            if self._implementation is Implementation.CUDF:  # pragma: no cover
                 return self._from_native_frame(
                     self._native_frame.merge(
                         other_native,
--- a/tests/frame/join_test.py
+++ b/tests/frame/join_test.py
@@ -159,7 +159,11 @@ def test_anti_join(
     join_key: list[str],
     filter_expr: nw.Expr,
     expected: dict[str, list[Any]],
+    request: pytest.FixtureRequest,
 ) -> None:
+    if "cudf" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
+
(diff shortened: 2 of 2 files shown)


FBruzzesi (commented):
Thanks @LiamConnors, I was curious and did some googling.
Maybe this is still possible to achieve: from the docs it seems that [`cudf.DataFrame.merge`](https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.dataframe.merge/#cudf-dataframe-merge) allows for some join types that are not supported in pandas, in particular `how="leftanti"` could be what we are looking for.

Would you mind giving it a spin?

LiamConnors:
Thanks @FBruzzesi! It does sound like it might be what we want. I'll try it out.

LiamConnors:
Thanks @FBruzzesi! This has the tests passing now. I was wondering about this here:
https://github.com/narwhals-dev/narwhals/pull/1041/files#diff-1bb9576cb79de69b2e45bcbc828669f54a6049dbdb76407986b3439ba309369dR481-R487
What situation would cause us to create extra columns in the join? Thanks for your feedback!

FBruzzesi:
> Thanks @FBruzzesi! This has the tests passing now.

Thanks! That's awesome 🚀

>  I was wondering about this here: https://github.com/narwhals-dev/narwhals/pull/1041/files#diff-1bb9576cb79de69b2e45bcbc828669f54a6049dbdb76407986b3439ba309369dR481-R487 What situation would cause us to create extra columns in the join? Thanks for your feedback!

In polars anti join does not add columns in other that match the keys (as it would not know where to add those). However, since the way we replicate that in pandas is via a outer join, columns would be added. To avoid such behavior we are considering solely the column in the keys for `other`. Does this help?

FBruzzesi (approved):
Thanks @LiamConnors ! Time to ship this ✨

CatFunnycatGIF

LiamConnors:
> > Thanks @FBruzzesi! This has the tests passing now.
> 
> 
> 
> Thanks! That's awesome 🚀
> 
> 
> 
> >  I was wondering about this here: https://github.com/narwhals-dev/narwhals/pull/1041/files#diff-1bb9576cb79de69b2e45bcbc828669f54a6049dbdb76407986b3439ba309369dR481-R487 What situation would cause us to create extra columns in the join? Thanks for your feedback!
> 
> 
> 
> In polars anti join does not add columns in other that match the keys (as it would not know where to add those). However, since the way we replicate that in pandas is via a outer join, columns would be added. To avoid such behavior we are considering solely the column in the keys for `other`. Does this help?
> 
> 

I think so. Does that mean it might technically be possible to rework the cuDF implementation to not use it? Just not passing it to the cuDF constructor didn't work, so I was curious.

FBruzzesi:
> I think so. Does that mean it might technically be possible to rework the cuDF implementation to not use it? Just passing the original other data frame to the cuDF constructor didn't work, so I was curious.

From the cudf docs:

```py
leftsemi: similar to inner join, but only returns columns from the left dataframe and ignores all columns from the right dataframe.
leftanti : returns only rows columns from the left dataframe for non-matched records. This is exact opposite to leftsemi join.
```

I would deduce that it is taken care of within the method itself, so possibly the step of slicing the columns and dropping duplicates is relevant for non CUDF implementation only

LiamConnors:
> > I think so. Does that mean it might technically be possible to rework the cuDF implementation to not use it? Just passing the original other data frame to the cuDF constructor didn't work, so I was curious.
> 
> 
> 
> From the cudf docs:
> 
> 
> 
> ```py
> 
> leftsemi : similar to inner join, but only returns columns from the left dataframe and ignores all columns from the right dataframe.
> 
> leftanti : returns only rows columns from the left dataframe for non-matched records. This is exact opposite to leftsemi join.
> 
> ```
> 
> 
> 
> I would deduce that it is taken care of within the method itself, so possibly the step of slicing the columns and dropping duplicates is relevant for non CUDF implementation only

Ah ok I see. Thanks for your help. I'll take another look at that

PR #1044, opened 2024-09-22
feat: Adding nth() for selecting columns

### LINKED ISSUE #1028: [Enh]: adding nth() to dataframe

Like [nth() in polars](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.nth.html), it would be a nice to have as an alternative to select()

To provide more options to select columns

@Cheukting is going to investigate

_No response_

_No response_

Cheukting:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1028 
- Closes #1028 

## Checklist

- [X] Tests added 
- [X] Documented the changes

Adding nth() function so that mimic behaviors of [nth() in polars](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.nth.html)


narwhals/_dask/expr.py
```diff
@@ -71,6 +71,27 @@ def func(df: DaskLazyFrame) -> list[dask_expr.Series]:
...
+            return [
+                df._native_frame.iloc[:, column_index] for column_index in column_indices
+            ]
+
+        return cls(
+            func,
+            depth=0,
+            function_name="nth",
+            root_names=list(map(str, column_indices)),
+            output_names=list(map(str, column_indices)),
```
MarcoGorelli, line 90: I think these could just be `None`, which is what we put when we can't keep track of the names because they're not known in advance (like we do with `nw.all`)

`root_names` / `output_names` are use in `group_by`, there's some explanation here: https://narwhals-dev.github.io/narwhals/how_it_works/#group-by

narwhals/_pandas_like/expr.py
```diff
@@ -81,6 +81,33 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
+                )
+                for column_index in column_indices
+            ]
+
+        return cls(
+            func,
+            depth=0,
+            function_name="nth",
+            root_names=list(map(str, column_indices)),
+            output_names=list(map(str, column_indices)),
```
MarcoGorelli, line 106: same

MarcoGorelli (commented):
Wow, this is amazing 😍 I'm seriously impressed at how much you figured out independently, well done!

For one of the tests to pass, you'll need to add `nth` to the `__all__` of `narwhals/stable/v1.py`

Cheukting:
Seems [top level nth was added in polars since 0.20.26](https://github.com/pola-rs/polars/releases/tag/py-0.20.26), some tests won't pass using older versions (e.g. 0.20.3) is it a way to limit which versions to test?

FBruzzesi:
> Seems [top level nth was added in polars since 0.20.26](https://github.com/pola-rs/polars/releases/tag/py-0.20.26), some tests won't pass using older versions (e.g. 0.20.3) is it a way to limit which versions to test?

That's something we want to address and take care of. In `PolarsNamespace` you could implement `nth` function which dispatch based on polars version:

```py
def nth(self,  *indices: int | Sequence[int]) -> PolarsExpr:
    pl = get_polars()
    if self._backend_version < (0, 20, 26):  # pragma: no cover
        # here goes the workaround
        ... 
     
    else:
        return PolarsExpr(pl.nth(*indices))
```

MarcoGorelli:
> I would propose providing an error message for older versions of Polars to advise users to use col instead.

agreed 👍 that, and perhaps a little note about this in the docstring?

tests/expr_and_series/nth_test.py
```diff
@@ -18,21 +19,43 @@
...
-    else:
-        pl_version = (0, 0, 0)
-    if is_polars and pl_version < (0, 20, 26):
-        with pytest.raises(
-            AttributeError, match="`nth` is only supported for Polars>=0.20.26."
-        ):
-            df.select(nw.nth(idx))
-    else:
-        result = df.select(nw.nth(idx))
-        compare_dicts(result, expected)
```
FBruzzesi: Hey @Cheukting , I would suggest going another direction instead of adding additional fixtures.

You can:

```py
def test_nth_not_polars(
    constructor: Constructor,
    idx: int | list[int],
    expected: dict[str, list[int]],
    request: Any
) -> None:
    pl = get_polars()
    if "polars" in str(constructor) and parse_version(pl) < parse_version(0.20.26):
    
    ...
```

docs/api-reference/narwhals.md
```diff
@@ -10,6 +10,7 @@ Here are the top-level functions available in Narwhals.
         - all_horizontal
         - any_horizontal
         - col
+        - nth
```
FBruzzesi, line 13: Could you move this so that the list is alphabetically ordered?

narwhals/expr.py
```diff
@@ -3741,6 +3742,53 @@ def func(plx: Any) -> Any:
     return Expr(func)
 
 
+def nth(*indices: int | Sequence[int]) -> Expr:
+    """
+    Creates an expression that references one or more columns by their index(es).
+
+    Warning:
+        `nth` is not supported for Polars version < 0.20.26. Please use `col` instead.
```
FBruzzesi, line 3750: You can refer to `col`, but I am afraid we need to add plugin in `mkdocs.yml` file.
For here
```suggestion
        `nth` is not supported for Polars version<0.20.26. Please use [`col`](/api-reference/narwhals/#narwhals.col) instead.
```

In `mkdocs.yml`

```bash
plugins:
- search
- autorefs
- mkdocstrings:
  ...
```

and in `/docs/requirements-doc.txt`

```diff
mkdocs
+ mkdocs-autorefs
mkdocs-material
```

tests/expr_and_series/nth_test.py
```diff
@@ -0,0 +1,48 @@
...
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.nth(idx))
+    compare_dicts(result, expected)
+
+
+@pytest.mark.skipif(
+    parse_version(pl.__version__) >= parse_version("0.20.26"),
+    reason="only riase error for Polars versions < 0.20.26",
+)
+def test_nth_not_supported() -> None:
```
FBruzzesi, line 43: This appear to not be covered as it does not always run, and coveragepy does not pick it up

```suggestion
def test_nth_not_supported() -> None:  # pragma: no cover
```

tests/translate/select_test.py
```diff
@@ -0,0 +1,49 @@
...
+    df = pd.DataFrame(data)
+    result = func(df)
+    pd.testing.assert_frame_equal(result, pd.DataFrame(expected))
+
+
+@pytest.mark.skipif(
+    parse_version(pl.__version__) >= parse_version("0.20.26"),
+    reason="only riase error for Polars versions < 0.20.26",
+)
+def test_nth_old_polars() -> None:
```
FBruzzesi, line 40: Same as for the other test that shows not covered

```suggestion
def test_nth_old_polars() -> None:  # pragma: no cover
```

FBruzzesi (commented):
Hey @Cheukting this looks awesome! I think we will be able to merge soon! 🚀

I have addressed all the issues with the failing CI and the questions you had. Hope it helps  😇

FBruzzesi:
One tiny issue I found while stress testing these changes is the fact that `from_column_indices` return anonymous expressions and therefore `.name` namespace cannot be used with it. Here is what I mean:

```py
import narwhals as nw

import pandas as pd
import polars as pl
import pyarrow as pa

data = {"a": [1, 2], "b": [-1, 0]}

@nw.narwhalify
def func(df):
    return df.select(nw.nth(1).name.suffix("_sfx"))

func(pl.DataFrame(data))
shape: (2, 1)
┌───────┐
│ b_sfx │
│ ---   │
│ i64   │
╞═══════╡
│ -1    │
│ 0     │
└───────┘

func(pd.DataFrame(data))
> ValueError: Anonymous expressions are not supported in `.name.suffix`.
Instead of `nw.all()`, try using a named expression, such as `nw.col('a', 'b')`

func(pa.table(data))
> ValueError: Anonymous expressions are not supported in `.name.suffix`.
Instead of `nw.all()`, try using a named expression, such as `nw.col('a', 'b')`
```

That's because of 
```py
...
    root_names=None,
    output_names=None,
```

I am trying to work around that

narwhals/_dask/expr.py
Cheukting: @FBruzzesi you mean this? Since we do not know the name, shall we use the column indices? (I can also investigate how Polars did it, but I bet they have access to the column names)

FBruzzesi: Whops, I missed this conversation! My apologies!

I agree that they should not be the column indexes, otherwise _very_ unexpected behaviour would follow.
We should at least add a disclaimer in the function docstring or expand the warning with all the `nw.nth` limitations.

MarcoGorelli:
thanks @Cheukting !

@FBruzzesi true, but the whole "root names" / "output names" mechanism is probably ripe for a refactor anyway 😄 OK with leaving that as a potential follow-up?

FBruzzesi:
> whole "root names" / "output names" mechanism is probably ripe for a refactor anyway 😄 

Oh my! Getting worried for this 😂

> OK with leaving that as a potential follow-up?

Sure let's do that, let's just add a disclaimer on limitation in the meanwhile then I am ok with merging it!

Cheukting:
I also agree with @MarcoGorelli I can add a warning message (or something like that) @FBruzzesi I am happy to look at the refactoring

Cheukting:
I think there's no way to express the `names` as is now. For Polars the `root_name` and `output_name` is only evaluated when calling `select`

MarcoGorelli (approved):
love this, and the mkdocstrings-autoref seems to work well

thanks @Cheukting and @FBruzzesi for reviewing!

maybe i'll try refactoring root_names / output_names in a live stream

MarcoGorelli:
looks like the random gif generator didn't come up with an approval dog, so here you go

PR #1049, opened 2024-09-23
feat: add `Series.dtype` return annotation

### LINKED ISSUE #1047: [Enh]: Export `nw.DType` from `Series`'s `.dtype` property

posit-dev/py-shiny

Ex: https://github.com/narwhals-dev/narwhals/blob/3f1619da7fc7f8f389e4d00797d029196b9a807b/narwhals/series.py#L305

Current code has to type the series object before I can ask if it is a numeric column. I'd like to be able to drop the `: nw.DType`. 

```python
def column_is_numeric(nw_data: DataFrame[Any], column_index: int) -> bool:
    series_dtype: DType = nw_data[:, column_index].dtype
    return series_dtype.is_numeric()
```

I'm fine with it staying as is... I'm just curious as to the motivation of the `Any` type given the amazing typing everywhere else.

Return `nw.DType` in the base series class. (Which would require downstream updates)

Type case the `Any` to `nw.DType`

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1047

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -359,7 +359,7 @@ class ArrowSeries:
         )
 
     @property
-    def dtype(self) -> DType:
+    def dtype(self: Self) -> DType:
         return translate_dtype(self._native_series.type)
 
     def abs(self) -> Self:
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from narwhals.dataframe import DataFrame
+    from narwhals.dtypes import DType
 
 
 class Series:
@@ -302,7 +303,7 @@ class Series:
         return len(self._compliant_series)
 
     @property
-    def dtype(self) -> Any:
+    def dtype(self: Self) -> DType:
         """
         Get the data type of the Series.
 
@@ -327,7 +328,7 @@ class Series:
             >>> func(s_pl)
             Int64
         """
-        return self._compliant_series.dtype
+        return self._compliant_series.dtype  # type: ignore[no-any-return]
 
     @property
     def name(self) -> str:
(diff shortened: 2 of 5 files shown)


narwhals/series.py
```diff
@@ -302,7 +303,7 @@ def len(self) -> int:
         return len(self._compliant_series)
 
     @property
-    def dtype(self) -> Any:
+    def dtype(self: Self) -> DType:
```
schloerke, line 306: Thank you!

narwhals/_arrow/series.py
```diff
@@ -359,7 +359,7 @@ def alias(self, name: str) -> Self:
         )
 
     @property
-    def dtype(self) -> DType:
+    def dtype(self: Self) -> DType:
```
schloerke, line 362: I don't believe it's needed

```suggestion
    def dtype(self) -> DType:
```

But if it is... Can you please teach me why? Thank you

FBruzzesi: I took this as opportunity to go a bit banana into hints for the rest of the codebase (uncommitted here). When we know that `self: Self`, then `self._native_series` is known to be a `pa.ChunkedArray`, which has a `type` property.

Probably some editors do not need it anyway 😇

MarcoGorelli: sounds like something which editors should fix 😄 but in the meantime, if this helps the development workflow, then it seems like a good idea to add it 👍

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1057, opened 2024-09-24
feat: add `Series.__iter__`

### LINKED ISSUE #1054: [Enh]: Add `__iter__` support to `Series`

posit-dev/py-shiny . Not required... just a bonus

When trying to perform a loop over the cells in a series to check the cell values, it's nice to use `enumerate(x)` to return both the value and the position.

I would like an `__iter__` method for the `nw.Series` object.

```python
for i, val in enumerate(ser):
    ...
```

This also freely extends to other methods like `zip()`. 

_No response_

I can use a counter on my own

```python
counter = -1
for val in ser:
    counter = counter + 1
    ...
```

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1054 

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -12,7 +12,6 @@ from typing import cast
 from typing import overload
 
 from narwhals.dependencies import is_numpy_array
-from narwhals.utils import flatten
 
 if TYPE_CHECKING:
     from narwhals._arrow.dataframe import ArrowDataFrame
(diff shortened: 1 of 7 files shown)


narwhals/_expression_parsing.py
```diff
@@ -95,7 +94,7 @@ def evaluate_into_exprs(
     """Evaluate each expr into Series."""
     series: ListOfCompliantSeries = [  # type: ignore[assignment]
         item
-        for sublist in (evaluate_into_expr(df, into_expr) for into_expr in flatten(exprs))
+        for sublist in (evaluate_into_expr(df, into_expr) for into_expr in exprs)
```
MarcoGorelli, line 97: ooh nice - was this never needed then?

MarcoGorelli (approved):
nice thanks so much @FBruzzesi !

narwhals/_expression_parsing.py
FBruzzesi: `*expr` should always be a tuple. Not sure why we flatten it?!

PR #1063, opened 2024-09-24
ci: bump version workflow

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

Not an expert, if someone knows better I would be happy to get this reviewed 🙈

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,0 +1,46 @@
+name: Bump version
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_type:     
+        description: 'Type of version bump: patch, minor, or major'
+        required: true
+        default: 'patch'
+        type: choice
+        options: 
+        - patch
+        - minor
+        - major
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    steps:
+
+    - uses: actions/checkout@v4
+
+    - uses: actions/setup-python@v5
+      with:
+        python-version: "3.11"
+
+    - name: bump version
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git checkout -b bump-version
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
nice, thanks for doing this, definitely open to simplifying the current process, new releases should be as easy as possible

my only question is - who can trigger this? if it's "anyone in narwhals-dev", then I think we need to find a way to tighten permissions, I think permissions to make releases should be as few as possible

MarcoGorelli:
nice, thanks for doing this, definitely open to simplifying the current process, new releases should be as easy as possible

my only question is - who can trigger this? if it's "anyone in narwhals-dev", then I think we need to find a way to tighten permissions, I think permissions to make releases should be as few as possible

FBruzzesi:
True! I found a couple of threads and sharing them here for my future self

- https://github.com/orgs/community/discussions/26622
- https://github.com/orgs/community/discussions/49688

MarcoGorelli:
thanks - maybe we can just hardcode some names there for releases then?

.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    - name: bump version
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git checkout -b bump-version
+        python utils/bump_version.py ${{ github.event.inputs.release_type }}
+
+    - name: Create pull request
+      uses: actions/github-script@v6
```
MarcoGorelli, line 35: let's try hard-coding ourselves and see if this works? if it does we can add some others
```suggestion
      uses: actions/github-script@v6
      if: github.actor == 'MarcoGorelli' || github.actor == 'FBruzzesi'
```

PR #1068, opened 2024-09-24
REFACTOR: Unite Your Requirements!

akmalsoliev:
With the arrival of a lightning-fast package manager, mighty UV, the requirements.txt is like yesterday's news. Instead of navigating through a jungle of dependencies, like a mismatched sock drawer, everything can now unite in one glorious spot—just like Italy in 1861!

This marvelous change even makes installing dependencies blazingly fast (except for cudf—polars is better anyways). Now it's as easy as pie with just one command: uv sync --all-extras.  No more typing three commands when you can do it in one! 1 > 3 — gives you more time to enjoy that delicious Italian caffè.

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -50,6 +50,64 @@ jobs:
             cd altair
             mypy altair tests
 
+  marimo:
+    strategy:
+      matrix:
+        python-version: ["3.12"]
+        os: [ubuntu-latest]
+        dependencies: ["core", "core,optional"]
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "**requirements*.txt"
+      - name: clone-marimo
+        run: |
+            git clone https://github.com/marimo-team/marimo.git --depth=1
+            cd marimo
+            git log
+      - name: install-basics
+        run: uv pip install --upgrade tox virtualenv setuptools hatch --system
+      - name: install-marimo-dev
+        run: |
+            cd marimo
+            uv pip install -e ".[dev]" --system
+      - name: install-narwhals-dev
+        run: |
+            uv pip uninstall narwhals --system
+            uv pip install -e . --system
+      - name: show-deps
+        run: uv pip freeze
+      - name: Create assets directory, copy over index.html
+        run: |
+            mkdir -p marimo/marimo/_static/assets
+            cp marimo/frontend/index.html marimo/marimo/_static/index.html
+            cp marimo/frontend/public/favicon.ico marimo/marimo/_static/favicon.ico
+      - name: Run tests with minimal dependencies
+        if: ${{ matrix.dependencies == 'core' }}
+        run: |
+            cd marimo
+            hatch run +py=${{ matrix.python-version }} test:test -v tests/ -k "not test_cli"
+        timeout-minutes: 15
+      - name: Run tests with optional dependencies
+        if: ${{ matrix.dependencies == 'core,optional' }}
+        run: |
+            cd marimo
+            hatch run +py=${{ matrix.python-version }} test-optional:test -v tests/ -k "not test_cli"
+        timeout-minutes: 15
+      - name: Run typechecks
+        run: |
+            cd marimo
+            hatch run typecheck:check
+
   scikit-lego:
     strategy:
       matrix:


akmalsoliev:
P.S. I've updated `pyproject.toml` and banished those pesky `.txt` files. More changes are coming; just testing the waters before diving in!

pyproject.toml
```diff
@@ -31,13 +31,40 @@ exclude = [
...
+    "mkdocstrings[python]>=0.26.1",
+    "polars>=1.0.0",
+]
+#cudf = ["cudf>=23.08.00"]
 modin = ["modin"]
 pandas = ["pandas>=0.25.3"]
 polars = ["polars>=0.20.3"]
 pyarrow = ["pyarrow>=11.0.0"]
-dask = ["dask[dataframe]>=2024.7"]
+dask = ["dask[dataframe]>=2023.5.0"]
```
FBruzzesi, line 67: Maybe Marco recalls why that was the min version

FBruzzesi (commented):
Hey there! Thanks for your PR. We definitly started to be opinionated to suggest uv in our contribution guide. However I have mixed feelings about this one. Let me try to mention a few reason (sorted by dev experience):

- If someone's workflow is not on uv, then how can he/she/they pick up dev-dependencies?
- `uv sync` is _too_ strict for my (but I guess many) development settings and to avoid that  it doesn't uninstall the rest of the libraries in the env, I need to remember to run `uv sync --inexact` instead
- For a few set up based on docker (e.g. devcontainers and codespaces), having a file allows to easily copy and install the dependencies _before_ even copying the python package. This leads to easier caching (at least in my experience).

FBruzzesi:
> Embracing change can indeed be challenging, and that's perfectly normal.

Definitely, and we would like to let people use whatever workflow works for them (not only pip) as well. Thanks for adjusting the pyproject section.

> From what I understand, Docker caching should remain intact whether using `pyproject.toml` or `requirements.txt`. If the current setup is causing issues, I'm keen to explore alternatives that work better for you. Here’s an example to illustrate:

> ```
> FROM python:latest
> 
> COPY ./pyproject.toml /pyproject.toml
> 
> RUN pip install .
> 
> RUN echo installing dev && pip install .[dev]
> 
> RUN echo installing polars && pip install .[polars]
> ```
> 
> Please let me know if there's anything specific you find challenging, and I’d be happy to discuss further.

The above snippet would run from `RUN pip install .` each time something changes in the codebase.
What I usually do is something along the lines:

```dockerfile
...
COPY ./requirements.txt /requirements.txt

RUN uv pip install -r requirements.txt # this will be cached until something changes in requirements.txt

COPY <rest of codebase> <rest of codebase>

RUN uv pip install .
...
```

There should be a way to link requirements.txt dynamically into pyproject, so personally that's how I would do it but happy to discuss

MarcoGorelli:
thanks - you'll need to update the instructions for this to work (e.g. ci, contributing)

fancy making a PR first, so that github automatically runs ci for you, and then coming back to this?

akmalsoliev:
Ciao @FBruzzesi, 
Sorry for late reply, what you think about this setup? 

```docker
FROM ghcr.io/astral-sh/uv:latest

COPY ./pyproject.toml pyproject.toml

RUN uv venv

RUN uv pip compile pyproject.toml --extra dev > requirements.txt
RUN uv pip install -r requirements.txt

RUN echo hello 

RUN uv pip install -r pyproject.toml --extra docs

COPY . .

RUN uv sync --all-extras 
```

When trying on the local machine it felt like there was a caching of 
```
RUN uv pip compile pyproject.toml --extra dev > requirements.txt
RUN uv pip install -r requirements.txt
```

pyproject.toml
FBruzzesi: Not sure why this was resolved, but 2024.7 is indeed the min version we can support, as we need dask_expr.

pyproject.toml
```diff
@@ -32,12 +32,37 @@ exclude = [
...
+    "pyarrow>=17.0.0",
+    "pytest>=8.3.3",
+    "pytest-cov>=5.0.0",
+    "pytest-env>=1.1.5",
+    "pytest-randomly>=3.15.0",
+    "scikit-learn>=1.3.2",
+    "tqdm>=4.66.5",
+    "typing-extensions>=4.12.2",
+    "dask[dataframe]>=2023.5.0"
+]
```
FBruzzesi, line 51: For running in ci/cd we should not pin any of these, as we want to have some old and random versions as well. Also dask requires python 3.9 or above.

pyproject.toml
```diff
@@ -32,12 +32,37 @@ exclude = [
...
+docs = [
+    "jinja2>=3.1.4",
+    "markdown-exec[ansi]>=1.9.3",
+    "mkdocs>=1.6.1",
+    "mkdocs-material>=9.5.36",
+    "mkdocstrings[python]>=0.26.1",
+    "polars>=1.0.0",
+]
+#cudf = ["cudf>=23.08.00"]
```
FBruzzesi, line 60: Any reason for commenting this one out? 🤔

FBruzzesi (commented):
Hey @akmalsoliev thanks again for the effort. The workflow you are proposing seems to work quite well and being able to cache layers properly in docker.

Given that is sorted out, I left some comments that probably require attention and, as @MarcoGorelli mentioned, we need this to be incorporated in the ci/cd workflows properly.
If possible including nox and the contributing documentation as well!

Let me know if I can help in any way forward 😉

pyproject.toml
akmalsoliev: Hi, sorry for the delayed response. If I recall correctly, @MarcoGorelli brought up this issue regarding installing cuDF with conda. The installation process is quite complex, and I'm not sure how to resolve it.
image

pyproject.toml
akmalsoliev: not an issue, was more of demonstration, will be fixed

Issue #1000, opened 2024-09-19
docs: Use pandas, PyArrow, and Polars eager in the DataFrame docstring examples


MarcoGorelli:
In the docstrings example in https://narwhals-dev.github.io/narwhals/api-reference/dataframe/, we always show the output for pandas and Polars

We should instead:
- show pandas, Polars (eager), and Pyarrow
- phrases such as "We can then pass either pandas or Polars:" should be rephrased to "We can then pass any supported library such as pandas, Polars, or PyArrow:"

If this is your first time submitting a pull request, please choose one function to start with, and only submit a pull request for that one function. You've got used to the process, feel free to submit changes for more functions together (e.g. in groups of 3-4 functions in the same PR)

---

Contributing guide: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md

example PR: https://github.com/narwhals-dev/narwhals/pull/1010

Hiderdk:
I can take a stab at making a PR for one of the functions.

Jupetus:
I will try the DataFrame.to_csv()

CarloLepelaars:
I can work do `Expr.mean` in `expr.py`. After this I might be able to do others in `Expr`.

CarloLepelaars:
Glad that the doc update for `Expr.mean` went smooth. Can now also pick up the other functions in `Expr`.

AlessandroMiola:
> In the docstrings example in https://narwhals-dev.github.io/narwhals/api-reference/dataframe/, we always show the output for pandas and Polars
> 
> We should instead:
> 
> * show pandas, Polars (eager), and Pyarrow
> * phrases such as "We can then pass either pandas or Polars:" should be rephrased to "We can then pass any supported library such as pandas, Polars, or PyArrow:"
> 

This is very much personal taste, but what would you think about rendering the different outputs via `mkdocs-material` content tabs? Something like the following

<img width="248" alt="image" src="https://github.com/user-attachments/assets/d59e0c40-210c-4e57-bf6a-79bf29e96d14">

    === "pandas"
        ```python
    
        func(df_pd)
        ```
        Output:
        ```
            a     b
        0  24.5  74.5
        ```
    === "Polars (eager)"
        ```python
    
        func(df_pl)
        ```
        Output:
        ```
        shape: (1, 2)
        ┌──────┬──────┐
        │ a    ┆ b    │
        │ ---  ┆ ---  │
        │ f64  ┆ f64  │
[...]

PR #1073, opened 2024-09-26
feat: add `series.rename`

### LINKED ISSUE #1051: feat: series.rename
I keep reaching for this and realising it doesn't exist

reference: https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.rename.html

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1051

## Checklist

- [x] Tests added 
- [x] Documented the changes


narwhals/series.py
```diff
@@ -1216,6 +1216,47 @@ def alias(self, name: str) -> Self:
...
+            >>> func(s_pl)  # doctest: +NORMALIZE_WHITESPACE
+            shape: (3,)
+            Series: 'bar' [i64]
+            [
+               1
+               2
+               3
+            ]
+        """
+        return self._from_compliant_series(self._compliant_series.alias(name=name))
```
FBruzzesi, line 1258: Should just be the same of

```suggestion
        return self.alias(name=name)
```
?

FBruzzesi (commented):
Thanks @raisadz 🚀

It looks good! I left a comment to simplify a bit the logic. 
I think `./utils/generate_backend_completeness.py` may need some love to render that `rename` is implemented for all series. But I would be ok to leave that as a follow up as well 👌

raisadz:
> Thanks @raisadz 🚀
> 
> It looks good! I left a comment to simplify a bit the logic. I think `./utils/generate_backend_completeness.py` may need some love to render that `rename` is implemented for all series. But I would be ok to leave that as a follow up as well 👌

Thank you for your review @FBruzzesi !

narwhals/series.py
```diff
@@ -1216,6 +1216,47 @@ def alias(self, name: str) -> Self:
...
+            >>> s_pd = pd.Series(s, name="foo")
+            >>> s_pl = pl.Series("foo", s)
+
+            We define a library agnostic function:
+
+            >>> @nw.narwhalify
+            ... def func(s):
+            ...     return s.rename("bar")
+
+            We can then pass either pandas or Polars to `func`:
```
FBruzzesi, line 1242: Hey @raisadz thanks for adjusting!

I have one more tiny request: in #1000 we are trying to add pyarrow to the docstrings. Would you mind adding an example with pyarrow as well? Maybe we don't have an example with series yet, but you can create a pyarrow "series" via `s_pa = pa.chunked_array([s])`.

Then the line here would become:

```suggestion
            We can then pass any supported library such as pandas, Polars, or PyArrow:
```
and finally the rendering below 🙃

raisadz: Thanks for the suggestion! I added an example with PyArrow series to the docstrings

FBruzzesi (approved):
Awesome, thanks @raisadz ! Let's ship it!

CatFunnycatGIF

PR #1090, opened 2024-09-28
fix: `drop_nulls` with subset of length>1

### LINKED ISSUE #1088: [Bug]: Run into safety assertion on `reuse_series_implementation` for drop_nulls
### Describe the bug

When executing `DataFrame.drop_nulls(subset=[...])` we run into safety assertion of `reuse_series_implementation` function

Consider the following tests:

```py
@pytest.mark.parametrize(
    ("subset", "expected"),
    [
        ("a", {"a": [1, 2.0, 4.0], "b": [float("nan"), 3.0, 5.0]}),
        (["a"], {"a": [1, 2.0, 4.0], "b": [float("nan"), 3.0, 5.0]}),
        (["a", "b"], {"a": [2.0, 4.0], "b": [3.0, 5.0]})
    ]
)
def test_drop_nulls_subset(constructor: Constructor, subset: str | list[str], expected: dict[str, float]) -> None:
    result = nw.from_native(constructor(data)).drop_nulls(subset=subset)
```

It will raise for arrow and pandas-like engines, not for polars or dask when subset has more than one column

### Expected results

To not raise

### Actual results

Raise AssertionError

```shell
Latest narwhals
```

### Relevant log output

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1088

## Checklist

- [x] Tests added 
- [ ] Documented the changes

This might not be the permanent fix to the issue, consider it as a hotfix for now

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -373,7 +373,7 @@ class ArrowDataFrame:
             return self._from_native_frame(self._native_frame.drop_null())
         subset = [subset] if isinstance(subset, str) else subset
         plx = self.__narwhals_namespace__()
-        return self.filter(~plx.any_horizontal(*[plx.col(s).is_null() for s in subset]))
+        return self.filter(~plx.any_horizontal(plx.col(*subset).is_null()))
 
     def sort(
         self,
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -373,7 +373,7 @@ class ArrowDataFrame:
             return self._from_native_frame(self._native_frame.drop_null())
         subset = [subset] if isinstance(subset, str) else subset
         plx = self.__narwhals_namespace__()
-        return self.filter(~plx.any_horizontal(plx.col(*subset).is_null()))
+        return self.filter(~plx.any_horizontal(*[plx.col(s).is_null() for s in subset]))
 
     def sort(
         self,
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -16,6 +16,7 @@ from narwhals._arrow.utils import horizontal_concat
 from narwhals._arrow.utils import vertical_concat
 from narwhals._expression_parsing import combine_root_names
 from narwhals._expression_parsing import parse_into_exprs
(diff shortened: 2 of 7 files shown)


MarcoGorelli (changes requested):
thanks for spotting this

i think the issue is just that `any_horizontal` should always only just return a single column, so we could do
```diff
             root_names=combine_root_names(parsed_exprs),
-            output_names=parsed_exprs[0]._output_names,
+            output_names=parsed_exprs[0]._output_names[:1],
         )
 
     def any_horizontal(self, *exprs: IntoPandasLikeExpr) -> PandasLikeExpr:
@@ -262,7 +262,7 @@ class PandasLikeNamespace:
             depth=max(x._depth for x in parsed_exprs) + 1,
             function_name="any_horizontal",
             root_names=combine_root_names(parsed_exprs),
-            output_names=parsed_exprs[0]._output_names,
+            output_names=parsed_exprs[0]._output_names[:1],
         )
```
and that should fix it without needing to do other changes

FBruzzesi:
`parsed_exprs[0]._output_names` can easily be None though.

Do you like: 
```diff
- output_names=parsed_exprs[0]._output_names,
+ output_names=(parsed_exprs[0]._output_names or [None])[:1],
```

MarcoGorelli:
are you're right - let's make a function then, just like we have for `combine_root_names` one line above? in there we can explain that we follow the right-hand-rule like Polars does

FBruzzesi:
> right-hand-rule like Polars does

left-most?

>  let's make a function then

Not sure I follow 100%. `combine_root_names` seems generic-ish.

For this (horizontal) reduction I would expect to have:

- left-most output name if available
- otherwise? left most root name?

MarcoGorelli:
yup, left-most, sorry :)

so `parsed_exprs[0]._output_names[:1] if parsed_exprs[0]._output_names is not None else None`

MarcoGorelli (approved):
love this

thanks @FBruzzesi !

PR #1098, opened 2024-09-30
test: Add a hypothesis test for __getitem__

### LINKED ISSUE #1008: test: add hypothesis test for `DataFrame.__getitem__`
`DataFrame.__getitem__` is...kinda complicated

Let's make a hypothesis test. Something like:

- start with `data = {"a": [1, 2, 3], "b": [4, 5, 6], "c": [7, 8, 9], "d": [1, 4, 2]}`
- generate a row selection, which could be: a list of integers, an integer, a slice of integers, a numpy array of integers
- generate a column selection, which could be: a list of strings, a list of integers, a slice of integers, a slice of strings, a string, an integer
- check that, for PyArrow and pandas, if we slice using `df[row_selection, column_column_selection]` we get the same result we'd get by using Polars
- check that, for PyArrow and pandas, if we slice using `df[row_selection]` we get the same result we'd get by using Polars
- check that, for PyArrow and pandas, if we slice using `df[column_selection]` we get the same result we'd get by using Polars

This would be _really_ valuable to get right

Contributing guide: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md

sjdenny:
Compares polars `__getitem__` calls with pandas & pyarrow.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1008 

## Checklist

- [ ] ~Tests added~
- [ ] ~Documented the changes~

Putting this up for initial feedback - there's a bunch of cases which pyarrow doesn't support (documented in the test).  Do we want to support these cases, or should be instead tighten the test (on pandas as well) and declare these unsupported for now?

Cases:
- pyarrow: doesn't support negative indexes
- pyarrow: pairs of slices unsupported, e.g. `df[0:1, 'a':'b']` (trivial pairs such as `df[:, 'a':'b']` where one slice is `:` are an exception).
- pyarrow: empty edge case, e.g. `df[[], "a":]` fails.
- pyarrow & pandas: `df[..., ::step]` is unsupported, ie `slice(None, None, <something>)`


MarcoGorelli (commented):
this is awesome, thanks @sjdenny !

looks like there's a failure in the "random versions" CI job (specifically, pandas==2.0.3)

if we can address that, then I think we can merge this can then gradually address the rest

sjdenny:
Thanks @MarcoGorelli !

> looks like there's a failure in the "random versions" CI job (specifically, pandas==2.0.3)

Looks like this was a flake instead (at least, I was able to reproduce on pandas==2.2.3 too, by running through more examples).  I've bumped the number of samples up to 10k in db69469: this does (did) reliably find the failing case, but the downside is it's a bit of a slow test now.  Which end of the tradeoff scale do you want to go for?  Related to this, I saw Hypothesis has [settings profiles](https://hypothesis.readthedocs.io/en/latest/settings.html#settings-profiles), which look neat for supporting faster settings locally & more thorough sampling in CI.

sjdenny:
Ah, coverage now.  Looking 👀

MarcoGorelli:
awesome stuff, thanks!

there's a couple of really minor things i wanted to address, but i'll try to get this in shortly, really appreciate your contribution here 🙏

MarcoGorelli (approved):
thanks @sjdenny !

again, so sorry I was really slow with this one - this is really valuable work you've done here, well done 🙌

Issue #1077, opened 2024-09-27
docs: add docstring examples for dtypes


MarcoGorelli:
The dtypes don't currently have docstrings https://narwhals-dev.github.io/narwhals/api-reference/dtypes/

it would be good to add a docstring to each, perhaps even with an example of how to end up with that dtype pandas/polars/pyarrow

anopsy:
I'll take that.
Just found out that Polars documentation doesn't show examples for dtypes. And Pandas documentation has examples that look like: 
```
>>>pd.BooleanDtype()
BooleanDtype
```

anopsy:
hey @MarcoGorelli, I'd like to clarify what do you mean by "example how to end up with that dtype", do you mean - first using cast and then checking the series dtype, or just checking the dtype of series or  a third secret thing?
Btw do we want to implement Dataframe.dtypes ?

MarcoGorelli:
i think any operation which produces that dtype is probably fine...e..g for boolean: `nw.from_native(pd.Series([True, True, False]), series_only=True).dtype` produces `nw.Boolean`

> Dataframe.dtypes

not sure, there's already `DataFrame.schema`

anopsy:
Thank you, that helps a lot!

FBruzzesi:
Only the following are left to be documented:

- [ ] [`Int128`](https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes.Int128)
- [ ] [`UInt128`](https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes.UInt128)
- [ ] [`Field`](https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes.Field)

anopsy:
Working on `Int128` and `UInt128`

PR #1121, opened 2024-10-03
docs : docstrings examples - dtypes

anopsy:
Just a quick check: considering I'll have to use cast to get some of the datatypes, maybe I should wrap the from_native into a func after all?

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1077
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -58,25 +58,194 @@ class Int64(NumericType):
     """
 
 
-class Int32(NumericType): ...
+class Int32(NumericType):
+    """32-bit signed integer type.
+
+    Examples:
+        >>> import pandas as pd
+        >>> import polars as pl
+        >>> import pyarrow as pa
+        >>> import narwhals as nw
+        >>> data = [2, 1, 3, 7]
+        >>> ser_pd = pd.Series(data)
+        >>> ser_pl = pl.Series(data)
+        >>> ser_pa = pa.chunked_array([data])
+
+        >>> def func(ser):
+        ...     ser_nw = nw.from_native(ser, series_only=True)
+        ...     return ser_nw.cast(nw.Int32).dtype
+
+        >>> func(ser_pd)
+        Int32
+        >>> func(ser_pl)
+        Int32
+        >>> func(ser_pa)
+        Int32
+    """
 
 
-class Int16(NumericType): ...
+class Int16(NumericType):
+    """16-bit signed integer type.
+
+    Examples:
+        >>> import pandas as pd
+        >>> import polars as pl
+        >>> import pyarrow as pa
+        >>> import narwhals as nw
+        >>> data = [2, 1, 3, 7]
+        >>> ser_pd = pd.Series(data)
+        >>> ser_pl = pl.Series(data)
+        >>> ser_pa = pa.chunked_array([data])
+
+       >>> def func(ser):
+       ...     ser_nw = nw.from_native(ser, series_only=True)
+       ...     return ser_nw.cast(nw.Int16).dtype
 
 
-class Int8(NumericType): ...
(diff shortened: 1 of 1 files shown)


MarcoGorelli (commented):
thanks @anopsy !

for Categorical, it looks like casting from string is unsupported in old pyarrow versions* - but you can create the categorical directly:
```python
>>> ca = pa.chunked_array(pa.array(['a', 'b'], type=pa.dictionary(pa.uint32(), pa.string())))
>>> nw.from_native(ca, series_only=True)
┌─────────────────────────────────────────┐
| Narwhals Series                         |
| Use `.to_native()` to see native output |
└─────────────────────────────────────────┘
>>> nw.from_native(ca, series_only=True).dtype
Categorical
```

For Struct / List / Array, there is a way to do that in pandas too use `pd.ArrowDtype`:
```
In [11]: pd.Series(data, dtype=pd.ArrowDtype(pa.large_list(pa.large_string())))
Out[11]:
0      ['narwhal' 'orca']
1    ['beluga' 'vaquita']
dtype: large_list<item: large_string>[pyarrow]
```

Similarly, for the struct example, I think you could use
```python
pa.struct({'a': pa.int64(), 'b': pa.large_list(pa.large_string())})
```

---

*maybe we should only run doctests on the latest versions...that would save us a lot of `# doctest: +SKIP`s which would be a good thing

tests/stable_api_test.py
```diff
@@ -12,6 +12,33 @@
 from tests.utils import Constructor
 from tests.utils import assert_equal_data
 
+DTYPES = {
```
MarcoGorelli, line 15: just out of interest, why add this?

narwhals/dtypes.py
```diff
@@ -254,6 +548,22 @@ def to_schema(self) -> OrderedDict[str, DType | type[DType]]:
...
+        >>> import pyarrow as pa
+        >>> import narwhals as nw
+        >>> data = [["narwhal", "orca"], ["beluga", "vaquita"]]
+        >>> ser_pl = pl.Series(data)
+        >>> ser_pa = pa.chunked_array([data])
+
+        >>> nw.from_native(ser_pl, series_only=True).dtype
+        List(String)
+        >>> nw.from_native(ser_pa, series_only=True).dtype
+        List(String)
```
MarcoGorelli, line 564: if you want a pandas example we could add
```
pd.Series(data, dtype=pd.ArrowDtype(pa.large_list(pa.large_string())))
```

MarcoGorelli (commented):
thanks for updating!

tests/stable_api_test.py
anopsy: hey Marco, I wasn't updating, I was just trying to resolve some conflicts and update my fork/local repo. I was working on the dtypes file and also had to adjust the tests and then you did some changes to the test file (among those was adding this DTYPES = {....}-thing, I think you did it during one of the live streams.)

MarcoGorelli: thanks for resolving merge conflicts then 🙌 

i think it might not be necessary any more to have `DTYPES` here then, could you try removing it from this file?

anopsy: Yes, I think I first merged your changes and then committed them back, in the process of updating my local repo. I can see now that you refactored this test 2 weeks ago or so. Yes, I'll try to get it right!

MarcoGorelli (approved):
thanks @anopsy !

PR #1128, opened 2024-10-03
feat: `concat_str`

### LINKED ISSUE #1116: [Enh]: Add `nw.concat_str` to horizontally concatenate string columns

Plotly needs something like that

Being able to horizontally concatenate string columns (see [discussion](https://discord.com/channels/1235257048170762310/1237430779584446546/1291044643240153195))

Implement a dedicate method since `__add__` is not enough for pyarrow backend

Current workaround for other backends is:

```py
reduce(
    lambda x, y: x + y,
    [df_sorted.get_column(c).cast(nw.String()) for c in cols],
    "",
)
```

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1116 

## Checklist

- [x] Tests added 
- [x] Documented the changes

This was a bit harder/more complex than expected to get right. The idea being that polars can switch between:

- `ignore_nulls=False` (default), which means that a single null in the row will be propagated to the final result.
- `ignore_nulls=True`, when this happens, it is not as simple as replacing the null with an empty string, since otherwise the separator will get repeated (I am using a reduce). Also the separator "attached" to the same column and row needs to be nullified to empty string.

All this comes out of the box with polars>=0.20.6 and pyarrow (not sure about min version, CI will tell). On the other hand for pandas, dask and polars<0.20.6, a fair bit of manipulation had to be done, especially for `ignore_nulls=True`. 

The gist of it is using a null_mask, a list of series corresponding to the null values of the original series, and use it to map both the null values and the separators to an empty string.

It has also been a long day, so maybe tomorrow I can take another look with some fresh eyes

### DIFF
--- a/docs/api-reference/narwhals.md
+++ b/docs/api-reference/narwhals.md
@@ -11,6 +11,7 @@ Here are the top-level functions available in Narwhals.
         - any_horizontal
         - col
         - concat
+        - concat_str
         - from_dict
         - from_native
         - get_level
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -30,6 +30,7 @@ from narwhals.expr import all_ as all
 from narwhals.expr import all_horizontal
 from narwhals.expr import any_horizontal
 from narwhals.expr import col
+from narwhals.expr import concat_str
 from narwhals.expr import len_ as len
 from narwhals.expr import lit
 from narwhals.expr import max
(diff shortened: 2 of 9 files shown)


FBruzzesi:
@MarcoGorelli I can really reproduce the CI failure even with the same pinned versions (screenshot attached)

image

MarcoGorelli (approved):
nice one @FBruzzesi, great use of `pl.fold`!

Issue #1139, opened 2024-10-05
[Bug]: unpivot raises for pyarrow with mixed types


FBruzzesi:
### Describe the bug

Pyarrow backend raises an error when running unpivot with mixed types

### Steps or code to reproduce the bug

```py
import narwhals as nw
import pandas as pd
import polars as pl
import pyarrow as pa

data = dict(a=[1, 2], b=[1.1, 2.1], index=[0, 1])

@nw.narwhalify
def func(df):
    return df.unpivot(index="index", on=["a", "b"])

func(pd.DataFrame(data)).dtypes
# index         int64
# variable     object
# value       float64
# dtype: object

func(pl.DataFrame(data)).dtypes
# [Int64, String, Float64]

func(pa.table(data))
```
```terminal
ArrowInvalid: Schema at index 1 was different: 
index: int64
variable: string
value: int64
vs
index: int64
variable: string
value: double
```

### Expected results

Except pyarrow to upcast to datatype

### Actual results

Error is raised

### Please run narwhals.show_version() and enter the output below.

```shell
latest
```

PR #1140, opened 2024-10-05
fix: pyarrow unpivot upcast numeric

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1139

## Checklist

- [x] Tests added 
- [ ] Documented the changes

For plotly, this is enough (as one can't plot string values).

Yet I left a TODO, up for discussion, in case we want to allow numeric to be upcasted to string as well - drawback is that we should do that manually.

### DIFF
--- a/tests/frame/unpivot_test.py
+++ b/tests/frame/unpivot_test.py
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+from typing import Any
+
 import pytest
 
 import narwhals.stable.v1 as nw
 from tests.utils import Constructor
 from tests.utils import compare_dicts
+
+if TYPE_CHECKING:
+    from narwhals.stable.v1.dtypes import DType
 
 data = {
     "a": ["x", "y", "z"],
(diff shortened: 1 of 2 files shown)


tests/frame/unpivot_test.py
```diff
@@ -70,3 +76,26 @@ def test_unpivot_default_var_value_names(constructor: Constructor) -> None:
...
+    ],
+)
+def test_unpivot_mixed_types(
+    request: pytest.FixtureRequest,
+    constructor: Constructor,
+    data: dict[str, Any],
+    expected_dtypes: list[DType],
+) -> None:
+    if "dask" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
```
FBruzzesi, line 97: Dask schema is: `[nw.Int64, nw.Object, nw.Float64]`. If we want to be more flexible in tests like this, we could adopt [dirty-equals](https://github.com/samuelcolvin/dirty-equals) - I really like the tool, I both used it and contributed to it in the past.

MarcoGorelli (approved):
thanks @FBruzzesi !

>         # TODO(Unassigned): Even with promote_options="permissive", pyarrow does not
        # upcast numeric to non-numeric (e.g. string) datatypes

i think this is fine - i'm a little surprised that polars allows this

PR #1145, opened 2024-10-06
feat: allow `format=None` in `str.to_datetime`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

~~Opening as draft since~~ (Ok I misscliked) pyarrow is unsupported and may need some discussion for how to achieve it.

For context, this is the snippet in plotly that requires casting to datetime: [process_dataframe_timeline](https://github.com/plotly/plotly.py/blob/7181c4dbdd86400b2b984434e25128507aa8ea69/packages/python/plotly/plotly/express/_core.py#L1780)

### DIFF
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -540,7 +540,7 @@ class ArrowExprStringNamespace:
             self._expr, "str", "slice", offset, length
         )
 
-    def to_datetime(self, format: str | None = None) -> ArrowExpr:  # noqa: A002
+    def to_datetime(self: Self, format: str | None) -> ArrowExpr:  # noqa: A002
         return reuse_series_namespace_implementation(
             self._expr,
             "str",
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1033,8 +1033,12 @@ class ArrowSeriesStringNamespace:
             ),
         )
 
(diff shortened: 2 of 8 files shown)


MarcoGorelli (commented):
thanks @FBruzzesi !

just a thought - it looks like [here](https://stackoverflow.com/a/76376483/4451315) there's a pure-python solution to inferring a format

maybe, in plotly, they could do something like:
- if it's pandas-like or polars, use `to_datetime` with `format=None`
- else, take the first non-null element, get the format using the solution above, and try to parse using that?

just an idea - not a blocker for this pr anyway!

MarcoGorelli (approved):

FBruzzesi:
> just a thought - it looks like [here](https://stackoverflow.com/a/76376483/4451315) there's a pure-python solution to inferring a format
> 
> * if it's pandas-like or polars, use `to_datetime` with `format=None`
> * else, take the first non-null element, get the format using the solution above, and try to parse using that?

Shall we consider doing that in narwhals for pyarrow backend and null format? I think it can be a quite common use case for anything related with timeseries.

MarcoGorelli:
yeah sure!

PR #1159, opened 2024-10-09
fix: make Series.scatter always return new series

### LINKED ISSUE #1154: [Enh]: `Series.scatter()` should always return a new `Series`

posit-dev/py-shiny

From [docs](https://github.com/narwhals-dev/narwhals/blob/b961d4eb982b945fcd8efb84f896c275086296c0/narwhals/series.py#L145-L150):
> For some libraries (pandas, Polars), this method operates in-place, whereas for others (PyArrow) it doesn't! We recommend being careful with it, and not relying on the in-placeness. For example, a valid use case is when updating a column in an eager dataframe, see the example below.

I'd like for this behavior to be consistent: use copy on write to return a new Series

Pandas is moving towards copy on write by default in pandas v3. 

When using narwhals's .scatter method, it currently displays a warning about enabling copy on write by default for pandas v2.
```
/venv/lib/python3.12/site-packages/narwhals/_pandas_like/series.py:190: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  s.iloc[indices] = values
```
[link to related code](https://github.com/narwhals-dev/narwhals/blob/b961d4eb982b945fcd8efb84f896c275086296c0/narwhals/_pandas_like/series.py#L190)

Change the return type of `.scatter()` to return a new `nw.Series` object.

For pandas, call `pd.options.mode.copy_on_write = True` to enable copy on write today to avoid problems later. 

For polars, update the code to return a new pl.Series object.

Adding `pd.options.mode.copy_on_write = True` line within pandas like `.scatter()` made the warning disappear and gave me expected behavior of a new series. 

I understand this is not a generic solution for pandas-like.

I'm currently getting around it by calling `nw_data = nw_data.clone()` before updating its Series values with `.scatter()`. 

_No response_

Cheukting:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- closes #1154 
- Closes #1154

## Checklist

- [X] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -393,6 +393,7 @@ class PandasLikeDataFrame:
                 backend_version=self._backend_version,
             )
         else:
+            # This is the logic in pandas' DataFrame.assign
             if self._backend_version < (2,):  # pragma: no cover
                 df = self._native_frame.copy(deep=True)
             else:
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -355,7 +355,6 @@ class PandasLikeDataFrame:
     ) -> Self:
         index = self._native_frame.index
         new_columns = evaluate_into_exprs(self, *exprs, **named_exprs)
-
         if not new_columns and len(self) == 0:
             return self
 
@@ -394,7 +393,10 @@ class PandasLikeDataFrame:
                 backend_version=self._backend_version,
             )
         else:
-            df = self._native_frame.copy(deep=False)
+            if self._backend_version < (2,):  # pragma: no cover
+                df = self._native_frame.copy(deep=True)
+            else:
+                df = self._native_frame.copy(deep=False)
             for s in new_columns:
                 df[s.name] = validate_dataframe_comparand(index, s)
         return self._from_native_frame(df)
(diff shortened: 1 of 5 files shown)


Cheukting:
@MarcoGorelli all check passes! This is a joint PR with @marcoala at the sprint

MarcoGorelli (commented):
this is wonderful, thanks @Cheukting and @marcoala!

could you also include a test which checks
```
s = df['a']
s.scatter({'a': [1, 2, 3]})
compare_dicts({'a': s}, expected
```
please? just to check that even with a free-standing series, things don't get modified

Cheukting:
Good catch @MarcoGorelli I have added the test and fix polars behaviour (we were wondering why polars didn't fail the test and now we know)

MarcoGorelli (approved):
thanks so much both!

i just added a commit to also address the other issue you noticed (thanks so much for reporting it!)

note that you can use `extract_native` instead of calling `to_list` - `to_list` creates a Python list, which can be inefficient, whereas `extract_native` extracts the native Polars Series from a Narwhals PolarsSeries object

Issue #1170, opened 2024-10-14
clean: remove `tpch/notebooks` folder


MarcoGorelli:
we migrated all the tpch scripts to https://github.com/narwhals-dev/narwhals/tree/main/tpch/queries, so we can remove the `tpch/notebooks` folder now

PR #1173, opened 2024-10-14
feat: add support for `Series|Expr.skew` method

CarloLepelaars:
This PR adds `skew` to Narwhals. Support is added for Polars, Pandas-like, Arrow and Dask. 

## Checklist

- [x] Tests added 
- [x] Documented the changes


MarcoGorelli (commented):
Awesome effort, thanks @CarloLepelaars , good to have you as contributor! Looks like there's a doctest failure

CarloLepelaars:
Thanks for the kind words! Doctest should be fixed now.

narwhals/_arrow/series.py
```diff
@@ -298,6 +299,17 @@ def std(self, ddof: int = 1) -> int:
 
         return pc.stddev(self._native_series, ddof=ddof)  # type: ignore[no-any-return]
 
+    def skew(self) -> float:
+        values = self._native_series.to_numpy()
+        np = get_numpy()
```
MarcoGorelli, line 304: could we use `import numpy as np  # ignore-banned-import` here? numpy is currently a required dependency of pyarrow, but I don't know if we can count on that always being the case

narwhals/_pandas_like/series.py
```diff
@@ -424,6 +426,23 @@ def std(
...
+        np = get_numpy()
+        pd = get_pandas()
+        ser = self._native_series
+        if hasattr(ser, "skew") and not isinstance(ser.dtype, pd.ArrowDtype):
+            return float(ser.skew())
+        else:
+            values = ser.to_numpy()
+            n = len(values)
+            if n < 3:
+                return float("nan")
```
MarcoGorelli, line 439: i might have misunderstood, but does this set the skew to nan if there's less than 3 values? looks like polars doesn't do that
```
In [2]: pl.Series([1,2]).skew()
Out[2]: 0.0
```

narwhals/expr.py
```diff
@@ -433,6 +433,43 @@ def std(self, *, ddof: int = 1) -> Self:
...
+            0  0.0  2.194964
+            >>> func(df_pl)
+            shape: (1, 2)
+            ┌─────┬──────────┐
+            │ a   ┆ b        │
+            │ --- ┆ ---      │
+            │ f64 ┆ f64      │
+            ╞═════╪══════════╡
+            │ 0.0 ┆ 1.472427 │
+            └─────┴──────────┘
```
MarcoGorelli, line 469: i'm confused, sorry, why do they return different values for `'b'`?

MarcoGorelli (commented):
thanks for updating, just left some comments (i'm a little tired today though so sorry if my comments don't make sense 😅 )

MarcoGorelli:
btw, if you wanted to just fix a typo somewhere in a separate pr (or, say, take https://github.com/narwhals-dev/narwhals/issues/1170), then once you're already a contributor, CI will always run automatically without me having to approve and run - just bringing this up in case it makes it easier for you

narwhals/_arrow/namespace.py
```diff
@@ -325,6 +325,11 @@ def mean(self, *column_names: str) -> ArrowExpr:
             *column_names, backend_version=self._backend_version, dtypes=self._dtypes
         ).mean()
 
+    def skew(self, *column_names: str) -> ArrowExpr:
+        return ArrowExpr.from_column_names(
+            *column_names, backend_version=self._backend_version, dtypes=self._dtypes
+        ).skew()
+
```
FBruzzesi, line 332: Polars does not have a `skew` function in the high level namespace, so I don't think we should either

narwhals/_arrow/series.py
```diff
@@ -298,6 +299,17 @@ def std(self, ddof: int = 1) -> int:
 
         return pc.stddev(self._native_series, ddof=ddof)  # type: ignore[no-any-return]
 
+    def skew(self) -> float:
```
FBruzzesi, line 302: Although it would end up returning a pyarrow scalar, I think we should keep the implementation with native methods, or you can reuse methods implemented, such as all elementary operations

narwhals/_pandas_like/namespace.py
```diff
@@ -177,6 +177,14 @@ def mean(self, *column_names: str) -> PandasLikeExpr:
             dtypes=self._dtypes,
         ).mean()
 
+    def skew(self, *column_names: str) -> PandasLikeExpr:
+        return PandasLikeExpr.from_column_names(
+            *column_names,
+            implementation=self._implementation,
+            backend_version=self._backend_version,
+            dtypes=self._dtypes,
+        ).skew()
```
FBruzzesi, line 186: Same as per arrow namespace comment

narwhals/_pandas_like/series.py
```diff
@@ -424,6 +426,23 @@ def std(
         ser = self._native_series
         return ser.std(ddof=ddof)
 
+    def skew(self) -> Any:
```
FBruzzesi, line 429: As Marco pointed out in the example, polars and pandas implementation seems to differ. We should try to remap to polars behavior here as well

narwhals/_polars/namespace.py
```diff
@@ -98,6 +98,13 @@ def mean(self, *column_names: str) -> PolarsExpr:
             return PolarsExpr(pl.mean([*column_names]), dtypes=self._dtypes)  # type: ignore[arg-type]
         return PolarsExpr(pl.mean(*column_names), dtypes=self._dtypes)
 
+    def skew(self, *column_names: str) -> PolarsExpr:
+        import polars as pl  # ignore-banned-import()
+
+        if self._backend_version < (0, 20, 4):  # pragma: no cover
+            return PolarsExpr(pl.skew([*column_names]), dtypes=self._dtypes)
+        return PolarsExpr(pl.skew(*column_names), dtypes=self._dtypes)
```
FBruzzesi, line 106: `pl.skew` doesn't exist or 🤔

narwhals/expr.py
```diff
@@ -433,6 +433,43 @@ def std(self, *, ddof: int = 1) -> Self:
...
+            >>> df_pd = pd.DataFrame({"a": [1, 2, 3, 4, 5], "b": [1, 1, 2, 10, 100]})
+            >>> df_pl = pl.DataFrame({"a": [1, 2, 3, 4, 5], "b": [1, 1, 2, 10, 100]})
+
+            Let's define a dataframe-agnostic function:
+
+            >>> @nw.narwhalify
+            ... def func(df):
+            ...     return df.select(nw.col("a", "b").skew())
+
+            We can then pass either pandas or Polars to `func`:
```
FBruzzesi, line 456: If possible I would add an example with pyarrow as well (we trying to do so since #1000)

narwhals/expr.py
```diff
@@ -433,6 +433,43 @@ def std(self, *, ddof: int = 1) -> Self:
         """
         return self.__class__(lambda plx: self._call(plx).std(ddof=ddof))
 
+    def skew(self) -> Self:
```
FBruzzesi, line 436: If you fancy going one step further, [polars signature](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.skew.html#polars.Expr.skew) has a `bias` argument - I am also happy to keep it as a follow up if that's too much for one PR

narwhals/_pandas_like/series.py
FBruzzesi: +1 on this

narwhals/series.py
```diff
@@ -519,6 +519,40 @@ def mean(self) -> Any:
         """
         return self._compliant_series.mean()
 
+    def skew(self) -> Any:
```
FBruzzesi, line 522: Same as `Expr.skew`, polars exposes a `bias` parameter

FBruzzesi (commented):
Hey @CarloLepelaars, thanks for the PR!

I left a few comments - the main challenge seems to be how different implementations are between pandas and polars native methods. However polars provide the formula it uses for the computation. It should be possible to reproduce that with native methods or using the series/expr methods that are already implemented in narwhals :)

narwhals/expr.py
CarloLepelaars: Interesting, will check it out. Any suggestions on how to easily expose this `bias` argument for Narwhals as well?

narwhals/_pandas_like/series.py
CarloLepelaars: You're right, this is not necessary. Skew is not that meaningful for 1-2 elements, but now that I think of it no reason for it to be `nan`. Should be resolved now.

narwhals/expr.py
CarloLepelaars: Oops, my bad! Made it consistent now.

narwhals/_arrow/namespace.py
CarloLepelaars: Good one! Was not aware that this was a separate thing. Removed the `skew` functions in the `namespace.py` files.

CarloLepelaars:
> Hey @CarloLepelaars, thanks for the PR!
> 
> I left a few comments - the main challenge seems to be how different implementations are between pandas and polars native methods. However polars provide the formula it uses for the computation. It should be possible to reproduce that with native methods or using the series/expr methods that are already implemented in narwhals :)

This is indeed challenging @FBruzzesi. Currently I've made it so every backend returns the biased population skewness, but we can potentially include an option for the unbiased skewness.

narwhals/expr.py
FBruzzesi: A (maybe) sophisticated example of how such remapping would work is [`value_counts`](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/series.py#L1990) - which however exists only for series, but you should get the gist of it

narwhals/_pandas_like/series.py
CarloLepelaars: This also has the default Polars behavior (i.e. Biased skewness) now. Is that what you mean by remapping?

narwhals/series.py
CarloLepelaars: I'm not sure how to expose this `bias` parameter in Narwhals since its only available in Polars, not in Pandas and Arrow. Do you have ideas?

narwhals/expr.py
CarloLepelaars: Good one! Added a pyarrow example

narwhals/_pandas_like/series.py
FBruzzesi: Maybe this is not explained extensively, but we try to stick to the polars API - which means both signature and behavior matching.

So as mentioned, I am happy to keep the addition of `bias` argument as a follow up, in the meanwhile, everything else should match polars behavior and outcome

CarloLepelaars: Got it! The sticking to the Polars API does not seem to be obvious from `CONTRIBUTING.MD`, but that does clarify why we avoid loading libraries such as `numpy`. 

I've adjusted the function to use native operations.

narwhals/expr.py
CarloLepelaars: I think we should leave this for a separate PR as I don't see a straightforward way to add this. I've matched all frameworks to the Polars behaviour (i.e. Biases Population Skewness).

narwhals/series.py
CarloLepelaars: See conversation in `narwhals/expr.py`

narwhals/series.py
```diff
@@ -519,6 +519,40 @@ def mean(self) -> Any:
...
+        Calculate the sample skewness of the Series.
+
+        Returns:
+            The sample skewness of the Series.
+
+        Examples:
+            >>> import pandas as pd
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> s = [1, 2, 3, 4, 5]
```
FBruzzesi, line 533: Can we use the non trivial example of the expr case?

FBruzzesi (commented):
Hey @CarloLepelaars thanks for adjusting! This looks better now!

I left a comment for the pyarrow case, and I have other two considerations:

- Should we account for the `len(ser) < 3` case and return 0?
- It may be worth checking that the numbers are same even when nulls are present

narwhals/_arrow/series.py
```diff
@@ -298,6 +298,22 @@ def std(self, ddof: int = 1) -> int:
...
+        import pyarrow.compute as pc  # ignore-banned-import()
+
+        ser = self._native_series
+        m = pc.mean(ser)
+        m2 = pc.mean(pc.multiply(pc.subtract(ser, m), pc.subtract(ser, m)))
+        m3 = pc.mean(
+            pc.multiply(
+                pc.multiply(pc.subtract(ser, m), pc.subtract(ser, m)), pc.subtract(ser, m)
+            )
+        )
```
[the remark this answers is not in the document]
CarloLepelaars: Great suggestion, this works! Pushed it

narwhals/series.py
CarloLepelaars: Sure, adjusted this now

CarloLepelaars:
> * Should we account for the `len(ser) < 3` case and return 0?

Let's see, this is where Pandas diverges from the rest. To make it consistent we should only handle the case where `len(data)==2`. In that case Pandas and PyArrow can return `0`. Do you also think that is the way to go?

```
Skewness for 2 elements:
SciPy:  0.000000
Pandas: nan
Polars: 0.000000

Skewness for 1 element:
SciPy:  nan
Pandas: nan
Polars: nan
```

> * It may be worth checking that the numbers are same even when nulls are present

Good one! Can add a case in `unary_test.py` that has nulls.

FBruzzesi:
> Let's see, this is where Pandas diverges from the rest. To make it consistent we should only handle the case where `len(data)==2`. In that case Pandas and PyArrow can return `0`. Do you also think that is the way to go?

Yes, we are trying to stick with polars api and results, so let's manually force that if needed!

> Good one! Can add a case in `unary_test.py` that has nulls.

That would be great - if it is too much though, we can also make it in a follow up PR

FBruzzesi:
> I've covered the cases as discussed and made them consistent with Polars behavior. `unary_test.py` now also covers data with `nan` and cases where there are less than 3 rows.

Thanks for addressing the cases, the CI failure seems unrelated.

However I am still quite unsure that we are matching polars behavior. When counting number of elements for the base cases, we should ignore null values, then (pseudo code):

```py
if n_not_nulls==0:
    return None   # same as pl.Series([]).skew() and pl.Series([None]).skew()
elif n_not_nulls==1:
    return float("nan")  # same as pl.Series([1]).skew() and pl.Series([1, None]).skew()
elif n_not_nulls==2:
    return 0.0  # same as pl.Series([1, 2]).skew() and pl.Series([1, 2, None]).skew()
else:
    return <compute_skew>
```
-

CarloLepelaars:
Implemented your suggestions for nan policy. There is only one edge case left for Dask, where it outputs `nan` instead of 0.0 with 2 non null elements. Not sure how to adjust `_dask/expr.py` to account for that.

FBruzzesi:
Hey @CarloLepelaars, thanks for adjusting. CI is failing because in #1224 , `compare_dicts` was renamed to `assert_equal_data`.

> Implemented your suggestions for nan policy. There is only one edge case left for Dask, where it outputs `nan` instead of 0.0 with 2 non null elements. Not sure how to adjust `_dask/expr.py` to account for that.

Regarding dask, I am not able to try it now, bif could definitly be a tricky one to get right! I am ok with marking it as xfail in tests for now

narwhals/_dask/expr.py
```diff
@@ -426,6 +426,13 @@ def std(self, ddof: int = 1) -> Self:
             returns_scalar=True,
         )
 
+    def skew(self: Self) -> Self:
+        return self._from_call(
+            lambda _input: _input.skew(),
+            "skew",
+            returns_scalar=True,
+        )
```
FBruzzesi, line 434: In case of dask, the behavior is not 100% consistent with polars for length 0, 1, 2.
Honestly, I am ok with that. The majority of use cases, especially if distributed data is needed should not involve those sizes to begin with

tests/expr_and_series/unary_test.py
```diff
@@ -1,49 +1,164 @@
...
+        "c_skew": [float("nan")],
+    }
+    assert_equal_data(result, expected)
+
+
+def test_unary_one_element(constructor: Constructor) -> None:
+    data = {"a": [1], "b": [2], "c": [float("nan")]}
+    # Dask runs into a divide by zero RuntimeWarning for 1 element skew.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
```
MarcoGorelli, line 125: any change we could make this a more targeted warning filter? just in case we accidentally filter out warnings we should pay attention to

i.e. `warnings.filter` with message, category, action

MarcoGorelli (commented):
awesome work, thanks - just got a comment on the warnings

tests/expr_and_series/unary_test.py
FBruzzesi: I used my favorite trick once again 😂

MarcoGorelli:
thanks both! should be good, will do another check but this should make it into the next release

CarloLepelaars:
Awesome, thank you both for working with me on this! Interesting trick to match the warning to Dask only.

MarcoGorelli (approved):
thanks @CarloLepelaars , and @FBruzzesi for review!

in general we're returning native scalars (e.g. numpy scalars for pandas, pyarrow scalars for pyarrow) so I've kept that consistent with the rest of the api here

MarcoGorelli:
just pushed a fix as the `else` part of `m3 / (m2**1.5) if m2 != 0 else 0` wasn't right nor tested

will merge on green and this can enter the next release 🥦

PR #1181, opened 2024-10-15
feat: add from_arrow (which uses the PyCapsule Interface)

### LINKED ISSUE #1158: [Enh]: Construct DataFrame from Arrow PyCapsule object

This request is towards using narwhals to remove the pandas/pyarrow dependencies from VegaFusion 2.0

The flow I'm aiming for with VegaFusion 2.0 is that I'd like to use Narwhals for basic column projection and schema inspection and then use the Arrow PyCapsule API to pass the result to Rust. Then in some cases, the Rust logic will return a new Arrow result in PyCapsule form, and it would be great to be able to use Narwhals to wrap this result using the same backend as the input.

I was picturing perhaps a constructor method in the same family as `from_dict`, accepting an arrow PyCapsule object.

```
nw.from_arrow_capsule(cap, native_namespace=nw.get_native_namespace(input_df))
```

cc @kylebarron for all things Arrow PyCapsule 😄 

_No response_

_No response_

MarcoGorelli:
closes #1158 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes


jonmmease:
Thanks @MarcoGorelli, this looks great!

EdAbati (approved):
It looks 👌👌, just added a minor comment about naming

narwhals/functions.py
```diff
@@ -406,6 +411,100 @@ def _from_dict_impl(
     return from_native(native_frame, eager_only=True)
 
 
+def from_pycapsule(
+    supports_py_capsule: SupportsPyCapsule, *, native_namespace: ModuleType
```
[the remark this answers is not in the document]
MarcoGorelli: 🤔 maybe this should just be `native_frame`?

narwhals/functions.py
```diff
@@ -29,6 +31,9 @@
     from narwhals.series import Series
     from narwhals.typing import DTypes
 
+    class SupportsPyCapsule(Protocol):
+        def __arrow_c_stream__(self) -> Any: ...
```
kylebarron, line 35: This isn't the correct type hint: https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html#protocol-typehints

narwhals/functions.py
```diff
@@ -406,6 +411,100 @@ def _from_dict_impl(
     return from_native(native_frame, eager_only=True)
 
 
+def from_pycapsule(
```
kylebarron, line 414: Personally I'd call this `from_arrow` and under the hood use the PyCapsule Interface. I think the PyCapsule itself is an implementation detail, and not something the user should need to know about.

MarcoGorelli:
thanks @EdAbati and @kylebarron for reviews! 🙏

MarcoGorelli:
gonna ship this so we can make a release, if there's issues we can always address as a follow-up (it's a new feature so it doesn't affect existing users) - thanks all for comments!

narwhals/functions.py
MarcoGorelli: based on https://github.com/pola-rs/polars/issues/12530#issuecomment-2419570358, i think this risks being confused with Polars' `from_arrow`, which does something different

i'd be more inclined to use `from_pycapsule` then

kylebarron: Do you only have top-level functions? Do you export a `DataFrame`-like class? Why not put it on that class?

But there is a difference between the two because your function seems to only convert to a `DataFrame` while the polars function tries to decide whether it should return a `Series` or `DataFrame`.

`from_pycapsule` still doesn't really resolve the ambiguity, but you could change it to `from_arrow_dataframe` or `from_arrow_df` or `from_arrow_table`.

MarcoGorelli: thanks! yeah maybe class constructors are the way to go for this one

PR #1186, opened 2024-10-15
fix: Add df`.rows(named=False)` support for cuDF data frames

schloerke:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # --
- Closes # --

## Checklist

- [x] Tests added 
- [ ] Documented the changes

`posit-dev/py-shiny` specifically relies on `.rows(named=False)` to return unnamed rows. I'd like to be sure that narwhals also supports `.rows(named=False)` for as many types as possible.

`cuDF` currently does not support `.rows(named=False)`, however `.rows(named=True)` does work!
Reprex notebook: https://colab.research.google.com/drive/16AnZzkt0QpJ3NNad8qrmOqPBfUX_jFxa#scrollTo=uM8arylKIF1U

Reprex code:
```python
import narwhals as nw
import cudf

df = cudf.DataFrame(
    {
        "a": list(range(20)),
        "b": list(reversed(range(20))),
        "c": list(range(20)),
    }
)
print(type(df))
print(df)

nw_df = nw.from_native(df, eager_only = True)

try:
  nw_df.rows(named = False)
except Exception as e:
  print("nw_df.rows(named=False)", str(e))

try:
  list(nw_df.iter_rows(named = False))
except Exception as e:
  print("nw_df.iter_rows(named=False)", str(e))

print("manual row calculations",
    [tuple(row.values()) for row in nw_df.rows(named=True)]
)
``` 
```

import narwhals as nw
import cudf

df = cudf.DataFrame(
    {
        "a": list(range(20)),
        "b": list(reversed(range(20))),
        "c": list(range(20)),
    }
)
…try:
  list(nw_df.iter_rows(named = False))
except Exception as e:
  print("nw_df.iter_rows(named=False)", str(e))

print("manual row calculations",
    [tuple(row.values()) for row in nw_df.rows(named=True)]
)

<class 'cudf.core.dataframe.DataFrame'>
     a   b   c
0    0  19   0
1    1  18   1
2    2  17   2
3    3  16   3
4    4  15   4
5    5  14   5
6    6  13   6
7    7  12   7
8    8  11   8
9    9  10   9
10  10   9  10
11  11   8  11
12  12   7  12
13  13   6  13
14  14   5  14
15  15   4  15
16  16   3  16
17  17   2  17
18  18   1  18
19  19   0  19
manual row calculations [(0, 19, 0), (1, 18, 1), (2, 17, 2), (3, 16, 3), (4, 15, 4), (5, 14, 5), (6, 13, 6), (7, 12, 7), (8, 11, 8), (9, 10, 9), (10, 9, 10), (11, 8, 11), (12, 7, 12), (13, 6, 13), (14, 5, 14), (15, 4, 15), (16, 3, 16), (17, 2, 17), (18, 1, 18), (19, 0, 19)]
nw_df.rows(named=False) cuDF does not support iteration of DataFrame via itertuples. Consider using `.to_pandas().itertuples()` if you wish to iterate over namedtuples.
nw_df.iter_rows(named=False) cuDF does not support iteration of DataFrame via itertuples. Consider using `.to_pandas().itertuples()` if you wish to iterate over namedtuples.
```

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -274,7 +274,7 @@ class PandasLikeDataFrame:
     ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
         if not named:
             # cuDF does not support itertuples. But it does support to_dict!
-            if is_cudf_dataframe(self._native_frame):
+            if self._implementation is Implementation.CUDF:  # pragma: no cover
                 # Extract the row values from the named rows
                 return [tuple(row.values()) for row in self.rows(named=True)]
 
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -16,6 +16,7 @@ from narwhals._pandas_like.utils import create_native_series
 from narwhals._pandas_like.utils import horizontal_concat
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals._pandas_like.utils import validate_dataframe_comparand
+from narwhals.dependencies import is_cudf_dataframe
 from narwhals.dependencies import is_numpy_array
 from narwhals.utils import Implementation
 from narwhals.utils import flatten
@@ -141,14 +142,16 @@ class PandasLikeDataFrame:
 
     def __getitem__(
         self,
-        item: str
-        | int
-        | slice
-        | Sequence[int]
-        | Sequence[str]
-        | tuple[Sequence[int], str | int]
-        | tuple[slice | Sequence[int], Sequence[int] | slice]
-        | tuple[slice, slice],
+        item: (
+            str
+            | int
+            | slice
+            | Sequence[int]
+            | Sequence[str]
+            | tuple[Sequence[int], str | int]
+            | tuple[slice | Sequence[int], Sequence[int] | slice]
+            | tuple[slice, slice]
+        ),
     ) -> PandasLikeSeries | PandasLikeDataFrame:
         if isinstance(item, tuple):
             item = tuple(list(i) if is_sequence_but_not_str(i) else i for i in item)
@@ -245,10 +248,29 @@ class PandasLikeDataFrame:
     def columns(self) -> list[str]:
         return self._native_frame.columns.tolist()  # type: ignore[no-any-return]
 
+    @overload
+    def rows(
+        self,
+        *,
+        named: Literal[True],
+    ) -> list[dict[str, Any]]: ...
+
+    @overload
+    def rows(
+        self,
+        *,
+        named: Literal[False] = False,
+    ) -> list[tuple[Any, ...]]: ...
+
     def rows(
         self, *, named: bool = False
     ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
         if not named:
+            # cuDF does not support itertuples. But it does support to_dict!
+            if is_cudf_dataframe(self._native_frame):
+                # Extract the row values from the named rows
+                return [tuple(row.values()) for row in self.rows(named=True)]
+
             return list(self._native_frame.itertuples(index=False, name=None))
 
         return self._native_frame.to_dict(orient="records")  # type: ignore[no-any-return]
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -262,6 +262,13 @@ class PandasLikeDataFrame:
         named: Literal[False] = False,
     ) -> list[tuple[Any, ...]]: ...
 
+    @overload
+    def rows(
+        self,
+        *,
+        named: bool,
+    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]: ...
+
     def rows(
         self, *, named: bool = False
     ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -16,7 +16,6 @@ from narwhals._pandas_like.utils import create_native_series
 from narwhals._pandas_like.utils import horizontal_concat
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals._pandas_like.utils import validate_dataframe_comparand
-from narwhals.dependencies import is_cudf_dataframe
 from narwhals.dependencies import is_numpy_array
 from narwhals.utils import Implementation
 from narwhals.utils import flatten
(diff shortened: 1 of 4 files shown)


schloerke:
I am missing a line in the code for code coverage. I'm hoping `cuDF` is installed within GHA. If not, then I do not know how to remove the code coverage check. 

Happy to hand over the PR or make adjustments 😃

narwhals/_pandas_like/dataframe.py
```diff
@@ -245,10 +248,36 @@ def __getitem__(
...
+    ) -> list[tuple[Any, ...]] | list[dict[str, Any]]: ...
+
     def rows(
         self, *, named: bool = False
     ) -> list[tuple[Any, ...]] | list[dict[str, Any]]:
         if not named:
+            # cuDF does not support itertuples. But it does support to_dict!
+            if is_cudf_dataframe(self._native_frame):
+                # Extract the row values from the named rows
+                return [tuple(row.values()) for row in self.rows(named=True)]
```
FBruzzesi, line 279: You can use `_implementation` attribute instead of directly checking the native frame here. For coverage, let's just tag this branch as no cover:

```suggestion
            if self._implementation is Implementation.CUDF:  # pragma: no cover
                # Extract the row values from the named rows
                return [tuple(row.values()) for row in self.rows(named=True)]
```

Remark that this should affect imports on top of the file

FBruzzesi (commented):
Thanks for the PR @schloerke 🙌🏼! I left a comment addressing the failing CI and some implementation detail as well.

TL;DR 

> I'm hoping cuDF is installed within GHA

It is not :(

> If not, then I do not know how to remove the code coverage check.

We do not cover those specific branches in CI 🙈

MarcoGorelli:
thanks @schloerke ! gpu testing is horrendously manual at the moment unfortunately, we use https://www.kaggle.com/code/marcogorelli/testing-cudf-in-narwhals
i'm in contact with some people from rapids to see what we can do about it

narwhals/_pandas_like/dataframe.py
```diff
@@ -16,6 +16,7 @@
 from narwhals._pandas_like.utils import horizontal_concat
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals._pandas_like.utils import validate_dataframe_comparand
+from narwhals.dependencies import is_cudf_dataframe
```
schloerke:
```suggestion
```

schloerke:
Independent of this PR, do you think `py-shiny` should adopt a safe-guard approach where we convert the data to pandas if an error is found while retrieving the rows?

```python
try:
	row_infos = nw_df.rows(named = False)
except Exception e:
    # Try converting to pandas before getting row infos
	try:
	    row_infos = nw.from_native(nw_df.to_pandas(), eager = True).rows(named = False)
    except Exception:
        # Raise original error
        raise e
```

Conceptually, this would have addressed this `cuDF` bug and `py-arrow` situation.  Just feels like an unnecessary conversion layer.

FBruzzesi:
> Independent of this PR, do you think `py-shiny` should adopt a safe-guard approach where we convert the data to pandas if an error is found while retrieving the rows?

With this PR all eager frames should not support `.rows(named=False)`

> Conceptually, this would have addressed this `cuDF` bug and `py-arrow` situation. Just feels like an unnecessary conversion layer.

I would argue that this should be addressed in Narwhals, and not in the downstream logic.

> ```python
> try:
> 	row_infos = nw_df.rows(named = False)
> except Exception e:
>     # Try converting to pandas before getting row infos
> 	try:
> 	    row_infos = nw.from_native(nw_df.to_pandas(), eager = True).rows(named = False)
>     except Exception:
>         # Raise original error
>         raise e
> ```

With this try-except you get the feature or bug of allowing for interchange level of support dataframe to first be converted to pandas and then get the rows (only if you were allowing for interchange to be provided in the first place)

tests/frame/rows_test.py
```diff
@@ -96,6 +96,32 @@ def test_rows(
...
+)
+def test_rows_eager(
+    constructor_eager: Any,
+    named: bool,  # noqa: FBT001
+    expected: list[tuple[Any, ...]] | list[dict[str, Any]],
+) -> None:
+    # posit-dev/py-shiny relies on `.rows(named=False)` to return unnamed rows
+    data = {"a": [1, 3, 2], "_b": [4, 4, 6], "z": [7.0, 8, 9], "1": [5, 6, 7]}
+    df = nw.from_native(constructor_eager(data), eager_only=True)
+    result = list(df.iter_rows(named=named))
```
MarcoGorelli, line 121: sorry i'm a little confused about this:
- the test is about `rows`
- the comment says that py-shiny relies on `.rows`
- the test tests `.iter_rows`

MarcoGorelli (commented):
awesome, thanks for your PR!

just got a question

tests/frame/rows_test.py
schloerke: 🤦 Thank you!

```suggestion
    # posit-dev/py-shiny relies on `.rows(named=False)` to return unnamed rows
    data = {"a": [1, 3, 2], "_b": [4, 4, 6], "z": [7.0, 8, 9], "1": [5, 6, 7]}
    df = nw.from_native(constructor_eager(data), eager_only=True)
    result = df.rows(named=named)
```

MarcoGorelli (commented):
thanks!

looks like there's a futurewarning bubbling up from pandas to modin, which in narwhals we can't do anything about, we may need to add something like
```python
@pytest.mark.filterwarnings('ignore:.*Starting with pandas version 3.0 all arguments of to_dict')
```
to the test

tests/frame/rows_test.py
```diff
@@ -71,6 +71,32 @@ def test_rows(
     assert result == expected
 
 
+@pytest.mark.parametrize(
```
schloerke, line 74: Ignore warning

```suggestion
@pytest.mark.filterwarnings('ignore:.*Starting with pandas version 3.0 all arguments of to_dict')
@pytest.mark.parametrize(
```

MarcoGorelli (approved):
nice one, thanks @schloerke 

I think it's a good idea to support this for cuDF

it looks like cuDF devs really don't want us iterating over a Series one element at a time (itertuples), presumably because then they'd be repeatedly copying data from gpu to cpu? but doing all the copying in one go with `to_dict` and then iterating over that (as you've done 🙌 ) seems fine

let's ship this then, CI failure is unrelated

PR #1194, opened 2024-10-16
feat: add to_py_scalar

### LINKED ISSUE #1185: enh: add `to_py_scalar`
in the plotly PR () we have

```
def to_py_scalar(scalar_like):
    """If scalar is not python native, tries to convert it to python native."""
    if hasattr(scalar_like, "as_py"):
        return scalar_like.as_py()
    elif hasattr(scalar_like, "item"):
        return scalar_like.item()
    else:
        return scalar_like
```

I think we should have this in Narwhals itself (we _may_ want it to be slightly more robust by doing `isinstance` checks instead of checking for attributes)

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1185 

## Checklist

- [x] Tests added 
- [x] Documented the changes


narwhals/stable/v1/__init__.py
```diff
@@ -948,6 +949,25 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
 
+def to_py_scalar(scalar: Any) -> Any:
+    """If a scalar is not Python native, tries to convert it to Python native.
+
+    Examples:
+        >>> import narwhals.stable.v1 as nw
+        >>> import pandas as pd
+        >>> df = nw.from_native(pd.DataFrame({"a": [1, 2, 3]}))
+        >>> nw.to_py_scalar(df["a"].item(0))
+        np.int64(1)
```
FBruzzesi, line 960: Should we do the same here (numpy arrays of length one)as we do for cupy? I think it would be more consistent

FBruzzesi (commented):
Hey @EdAbati , it already looks good! I left a tiny comment to discuss how we should behave for numpy arrays of length 1 (similar to cupy case)

narwhals/stable/v1/__init__.py
EdAbati: Yeah true, I was a bit unsure since numpy scalars are _basically_ like Python native AFAIK (eg `np.int64(1) == 1`)

But why not? 👌🙂

narwhals/translate.py
```diff
@@ -776,8 +778,43 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    if (  # pragma: no cover
+        cupy and isinstance(scalar_like, cupy.ndarray) and scalar_like.size == 1
+    ):
+        return scalar_like.item()
+
+    np = get_numpy()
+    if np and np.isscalar(scalar_like) and hasattr(scalar_like, "item"):
+        return scalar_like.item()
+
+    return scalar_like
```
FBruzzesi, line 812: Ok this may be _way_ too defensive programming but I am putting the question out there: should we raise for any other type which is not builtin?

```suggestion
    if scalar_like.__class__.__module__ == "builtins":
        return scalar_like
    
    msg = (
        f"Expected object convertible to a scalar, found {type(scalar_like)}. "
        "Please report a bug to https://github.com/narwhals-dev/narwhals/issues"
    )
    raise ValueError(msg)
```

FBruzzesi (approved):
Thanks for the quick adjustment @EdAbati!
You are getting my plus 1 here, yet I also left a consideration/question

CatDanceGIF

MarcoGorelli (commented):
thanks @EdAbati , really nice!

Should we handle pandas-like Timestamp / Timedelta, and return stdlib datetime / timedelta? They're supposed to be mostly interchangeable, but in practice aren't exactly the same and people often end up having to handle both...maybe `to_py_scalar` should just return stdlib datetime / timedelta and save all potential headaches from timestamp / timedelta

narwhals/translate.py
EdAbati: I think you are right! If we don't raise an error we should have called it `maybe_py_scalar`. Added the `ValueError` but restricting on types. Now for example dicts and lists will raise an error. What do you think?

narwhals/translate.py
```diff
@@ -776,8 +781,70 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    all_scalar_types = (
+        int,
+        float,
+        complex,
+        bool,
+        bytes,
+        str,
+        datetime,
+        timedelta,
+        numbers.Number,
```
EdAbati, line 833: this is because also numpy does it: https://numpy.org/doc/2.0/reference/generated/numpy.isscalar.html

narwhals/translate.py
FBruzzesi: I think that's ok for the vast majority of use cases 👌

MarcoGorelli (approved):
awesome, thanks @EdAbati , really nice!

Issue #1199, opened 2024-10-17
test: use `constructor_eager` instead of parametrising over example dataframes


MarcoGorelli:
There's still some cases in the test suite where we're parametrising over dataframes instead of using `constructor` / `constructor_eager`:

The task is:
- look for instances of `df_pandas = ` in the `tests` folder. e.g.: https://github.com/narwhals-dev/narwhals/blob/000c360fef499bc8f4169f0759abd7cd2b6ee0ef/tests/frame/rows_test.py#L13-L41
- rewrite the test using `constructor_eager` instead

PR #1215, opened 2024-10-18
Is into series

LiamConnors:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/dependencies.py
+++ b/narwhals/dependencies.py
@@ -176,6 +176,45 @@ def is_pandas_like_series(arr: Any) -> bool:
     return is_pandas_series(arr) or is_modin_series(arr) or is_cudf_series(arr)
 
 
+def is_into_series(ser: Any) -> bool:
+    """
+    Check whether `ser` can be converted to a Narwhals Series.
+
+    Arguments:
+        ser: The object to check.
+
+    Returns:
+        `True` if `ser` can be converted to a Narwhals Series, `False` otherwise.
+
+    Examples:
+        >>> import pandas as pd
+        >>> import polars as pl
+        >>> import numpy as np
+        >>> import narwhals as nw
+
+        >>> s_pd = pd.Series([1, 2, 3])
+        >>> s_pl = pl.Series([1, 2, 3])
+        >>> np_arr = np.array([1, 2, 3])
+
+        >>> nw.dependencies.is_into_series(s_pd)
+        True
+        >>> nw.dependencies.is_into_series(s_pl)
+        True
+        >>> nw.dependencies.is_into_series(np_arr)
+        False
+    """
+
+    return (
+        (
+            hasattr(ser, "_compliant_series")
+            and hasattr(ser._compliant_series, "__narwhals_series__")
+        )
+        or is_polars_series(ser)
+        or is_pyarrow_chunked_array(ser)
+        or is_pandas_like_series(ser)
+    )
+
+
 __all__ = [
     "get_polars",
     "get_pandas",
@@ -200,4 +239,5 @@ __all__ = [
     "is_dask_dataframe",
     "is_pandas_like_dataframe",
     "is_pandas_like_series",
+    "is_into_series",
 ]
--- a/tests/dependencies/is_into_series_test.py
+++ b/tests/dependencies/is_into_series_test.py
@@ -1,0 +1,18 @@
+from __future__ import annotations
+
+import numpy as np
+import pandas as pd
+import polars as pl
+import pyarrow as pa
+
+import narwhals as nw
+from narwhals.dependencies import is_into_series
+
+
+def test_is_into_series() -> None:
+    assert is_into_series(pa.chunked_array([["a", "b"]]))
+    assert is_into_series(pl.Series([1, 2, 3]))
+    assert is_into_series(pd.Series([1, 2, 3]))
+    assert is_into_series(nw.from_native(pd.Series([1, 2, 3]), allow_series=True))
+    assert not is_into_series(np.array([1, 2, 3]))
+    assert not is_into_series([1, 2, 3])
--- a/docs/api-reference/dependencies.md
+++ b/docs/api-reference/dependencies.md
@@ -14,6 +14,7 @@
         - is_cudf_series
         - is_dask_dataframe
         - is_ibis_table
+        - is_into_series
         - is_modin_dataframe
         - is_modin_series
         - is_numpy_array


narwhals/dependencies.py
```diff
@@ -176,6 +176,45 @@ def is_pandas_like_series(arr: Any) -> bool:
...
+        True
+        >>> nw.dependencies.is_into_series(np_arr)
+        False
+    """
+
+    return (
+        (
+            hasattr(ser, "_compliant_series")
+            and hasattr(ser._compliant_series, "__narwhals_series__")
+        )
```
FBruzzesi, line 211: Would this translate to:

```suggestion
    from narwhals.series import Series
    
    return (
        isinstance(ser, Series)
```

or is there any implication with the stable API?

tests/dependencies/is_into_series_test.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import annotations
```
FBruzzesi, line 1: Commenting here to address entire file. On the flow of #1199 I would suggest to use `constructor_eager` for anything returning True, and manually constructor only numpy array or list that return False. Probably the easiest way of achieving that is by splitting into two different tests

FBruzzesi (commented):
Thanks @LiamConnors ! This is awesome! I left a couple of comments (in the nitpick direction 🙈)

narwhals/dependencies.py
MarcoGorelli: I think we want both, so something like
```
if isinstance(ser, Series)
or hasattr(ser, '__narwhals_series__')
or is_polars_series(ser)
# etc.
```

We're trying to find objects which, if we pass them to `from_native` with `allow_series=True` or `series_only=True`, then they will successfully get converted to a Narwhals Series

And in `from_native`, these are the things we check for

LiamConnors: Thanks. Will take a look and make the required updates.

LiamConnors: I've been looking at the checks in `from_native`.  Just so I make sure I completely understand - is `hasattr(ser, '__narwhals_series__')` to account for passing in a compliant series (number 3 below).

So, you'll get `True` for:

1. a polars, pandas series etc:
```
is_into_series(pl.Series([1, 2, 3]))
```
2. for `nw` series:
```
nw.from_native(pd.Series([1, 2, 3]), allow_series=True)
```
3. And for the underlying `_compliant_series`?
```
nw.from_native(pd.Series([1, 2, 3]), allow_series=True)._compliant_series
```

tests/dependencies/is_into_series_test.py
LiamConnors: Makes sense to me. Thanks for the feedback @FBruzzesi! Taking a look

narwhals/dependencies.py
MarcoGorelli: thanks for asking, definitely good to get clarity on these things!

My hope is that Narwhals can just be glue between compliant implemetations. Currently, all compliant implementations live in the Narwhals codebase itself (`narwhals/_pandas_like`, `narwhals/_polars`, ...), by my hope is that the [extensibility mechanism](https://narwhals-dev.github.io/narwhals/extending/#extending-narwhals) can allow other dataframe libraries to define their own Narwhals-compliant implementation, so that all they expose to us is `__narwhals_dataframe__` / `__narwhals_series__` / `__native_namespace__` (and some others).

Some reasons why people might want to do this (as opposed to having the wrapping code live in Narwhals) are:
- they might want to be in control of the implementation and only hand off to us the bare minimum
- they might want the Narwhals-compatible code to be written in a lower-level language (C, Rust, ...), whereas anything which lives in Narwhals itself must be pure-Python
- they might be a closed-source library which wants to expose the bare minimum

Granted, nobody has yet made such an extension (for a start, the docs don't make it very easy to understand exactly what needs to be done...). But I'm hoping to be able to focus more on it next year, and to provide a reference implementation

Please do ask if this isn't clear, thanks 🙏

LiamConnors: This makes sense to me now. Thanks so much for the detail explanation! 
Really interested to learn more about what it would look like to create a Narwhals-compliant implementation and would love to help in some way with the documentation on the topic.

tests/dependencies/is_into_series_test.py
LiamConnors: I'm having some trouble committing my updates to the tests. Here's what I have, but I'm getting two errors.

```

import numpy as np

import narwhals as nw
from narwhals.dependencies import is_into_series
from tests.utils import ConstructorEager

def test_is_into_series_true(constructor_eager: ConstructorEager) -> None:
    raw_series = constructor_eager({"a": [1, 2, 3]})["a"]
    assert is_into_series(raw_series)
    nw_series = nw.from_native(raw_series, allow_series=True)
    assert is_into_series(nw_series)
```

One error I'm getting with committing is:

```
tests/dependencies/is_into_series_test.py:7:25: TCH001 Move application import `tests.utils.ConstructorEager` into a type-checking block
```
I can't figure this one out. It looks like I'm importing it in the same way as other tests do. 

The other error I get is:

```
tests/dependencies/is_into_series_test.py:11: error: Value of type "NativeFrame | DataFrame[Any] | DataFrameLike" is not indexable  [index]
```
So I guess this is not always expected to work for `ConstructorEager` type?

```
constructor_eager({"a": [1, 2, 3]})["a"]
```
Or there's something I need to add to the types "NativeFrame", "DataFrame[Any]", "DataFrameLike"?

FBruzzesi: Hey @LiamConnors 

> tests/dependencies/is_into_series_test.py:7:25: TCH001 Move application import `tests.utils.ConstructorEager` into a type-checking block

will require to do:

```py

if TYPE_CHECKING:
    from tests.utils import ConstructorEager
```

This is because of `from __future__ import annotations`


> tests/dependencies/is_into_series_test.py:11: error: Value of type "NativeFrame | DataFrame[Any] | DataFrameLike" is not indexable  [index]
> Found 1 error in 1 file (checked 1 source file)

is because lazy frame does not support indexing. It should not complain if you specify `nw.from_native(..., eager_only=True)`

PR #1220, opened 2024-10-18
enh: add dt.timestamp

### LINKED ISSUE #1182: enh: add dt.timestamp
Here's the Polars method: https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.dt.timestamp.html

We should add this to Narwhals for use in Plotly

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1182

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -785,9 +785,9 @@ class ArrowSeriesDateTimeNamespace:
         import pyarrow.compute as pc  # ignore-banned-import
 
         s = self._arrow_series._native_series
-        dtype = s.type
-        if isinstance(dtype, pa.TimestampType):
-            unit = dtype.unit
+        dtype = self._arrow_series.dtype
+        if dtype == self._arrow_series._dtypes.Datetime:
+            unit = dtype.time_unit  # type: ignore[attr-defined]
             s_cast = s.cast(pa.int64())
             if unit == "ns":
                 if time_unit == "ns":
@@ -810,7 +810,7 @@ class ArrowSeriesDateTimeNamespace:
                     result = pc.multiply(s_cast, 1_000)
                 if time_unit == "ms":
                     result = s_cast
-        elif dtype == pa.date32():
+        elif dtype == self._arrow_series._dtypes.Date:
             time_s = pc.multiply(s.cast(pa.int32()), 86400)
             if time_unit == "ns":
                 result = pc.multiply(time_s, 1_000_000_000)
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -810,6 +810,13 @@ class ArrowSeriesDateTimeNamespace:
                     result = pc.multiply(s_cast, 1_000)
                 if time_unit == "ms":
                     result = s_cast
+            if unit == "s":
+                if time_unit == "ns":
+                    result = pc.multiply(s_cast, 1_000_000_000)
+                if time_unit == "us":
+                    result = pc.multiply(s_cast, 1_000_000)
+                if time_unit == "ms":
+                    result = pc.multiply(s_cast, 1_000)
         elif dtype == self._arrow_series._dtypes.Date:
             time_s = pc.multiply(s.cast(pa.int32()), 86400)
             if time_unit == "ns":
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -780,6 +780,49 @@ class ArrowSeriesDateTimeNamespace:
 
         return self._arrow_series._from_native_series(result)
 
+    def timestamp(self: Self, time_unit: Literal["ns", "us", "ms"] = "us") -> ArrowSeries:
+        import pyarrow as pa  # ignore-banned-import
+        import pyarrow.compute as pc  # ignore-banned-import
+
+        s = self._arrow_series._native_series
+        dtype = s.type
+        if isinstance(dtype, pa.TimestampType):
+            unit = dtype.unit
+            s_cast = s.cast(pa.int64())
+            if unit == "ns":
+                if time_unit == "ns":
+                    result = s_cast
+                if time_unit == "us":
+                    result = pc.divide(s_cast, 1_000)
+                if time_unit == "ms":
+                    result = pc.divide(s_cast, 1_000_000)
+            if unit == "us":
+                if time_unit == "ns":
+                    result = pc.multiply(s_cast, 1_000)
+                if time_unit == "us":
+                    result = s_cast
+                if time_unit == "ms":
+                    result = pc.divide(s_cast, 1_000)
+            if unit == "ms":
+                if time_unit == "ns":
+                    result = pc.multiply(s_cast, 1_000_000)
+                if time_unit == "us":
+                    result = pc.multiply(s_cast, 1_000)
+                if time_unit == "ms":
+                    result = s_cast
+        elif dtype == pa.date32():
+            time_s = pc.multiply(s.cast(pa.int32()), 86400)
+            if time_unit == "ns":
+                result = pc.multiply(time_s, 1_000_000_000)
+            if time_unit == "us":
+                result = pc.multiply(time_s, 1_000_000)
+            if time_unit == "ms":
+                result = pc.multiply(time_s, 1_000)
+        else:
+            msg = "Input should be either of TimeStamp or Date type"
+            raise TypeError(msg)
+        return self._arrow_series._from_native_series(result)
+
     def date(self: Self) -> ArrowSeries:
         import pyarrow as pa  # ignore-banned-import()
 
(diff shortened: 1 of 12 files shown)


narwhals/_arrow/series.py
```diff
@@ -780,6 +780,56 @@ def convert_time_zone(self: Self, time_zone: str) -> ArrowSeries:
...
+        if dtype == self._arrow_series._dtypes.Datetime:
+            unit = dtype.time_unit  # type: ignore[attr-defined]
+            s_cast = s.cast(pa.int64())
+            if unit == "ns":
+                if time_unit == "ns":
+                    result = s_cast
+                if time_unit == "us":
+                    result = pc.divide(s_cast, 1_000)
+                if time_unit == "ms":
+                    result = pc.divide(s_cast, 1_000_000)
```
MarcoGorelli, line 798: for all of these, it would be marginally more efficient to use `elif`, else we're doing similar comparisons

we're already validating `time_unit` in `narwhals/expr.py`, so this could just be
```
                if time_unit == "ns":
                    result = pc.multiply(s_cast, 1_000)
                elif time_unit == "us":
                    result = s_cast
                else:
                    result = pc.divide(s_cast, 1_000)
```

MarcoGorelli (commented):
thanks @raisadz , serious effort here, looks really nice, and this is a really valuable feature!

I'd suggest adding in a hypothesis test, something like
```python
from hypothesis import given
import hypothesis.strategies as st

@given(
    inputs=st.datetimes(min_value=datetime(1960,1,1), max_value=datetime(1980,1,1)),
    time_unit=st.sampled_from(['ms', 'us', 'ns']),
    # We keep 'ms' out for now due to an upstream bug: https://github.com/pola-rs/polars/issues/19309
    starting_time_unit=st.sampled_from(['us', 'ns'])
)
def test_me(inputs, time_unit, starting_time_unit) -> None:
    import polars as pl
    import pandas as pd
    import pyarrow as pa
    @nw.narwhalify
    def func(s):
        return s.dt.timestamp(time_unit)
    result_pl = func(pl.Series([inputs], dtype=pl.Datetime(starting_time_unit)))
    result_pd = func(pd.Series([inputs], dtype=f'datetime64[{starting_time_unit}]'))
    result_pdpa = func(pd.Series([inputs], dtype=f'timestamp[{starting_time_unit}][pyarrow]'))
    result_pa = func(pa.chunked_array([[inputs]], type=pa.timestamp(starting_time_unit)))
    assert result_pl[0] == result_pd[0]
    assert result_pl[0] == result_pdpa[0]
    assert result_pl[0] == result_pa[0].as_py()
```
(but with a better name 😅 )

FBruzzesi:
I am catching up with Marco's today's livestream - as he explains how the integer representation works, wouldn't make sense to have something like the following pseudocode (details such as time units, time zones, date vs datetime, etc need to be taken care)

```py
diff = series - datetime(1970, 1, 1, 0, 0, 0, tzinfo=...)
if time_unit == "ns":
    return diff.dt.total_nanoseconds()
elif ...
...
```

Maybe it ends up being similar effort, but feels less redundant than all the nested `if` cases.

Edit: pyarrow also has `nanoseconds_between`, `microseconds_between` and `milliseconds_between` two date/datetimes

tests/expr_and_series/dt/timestamp_test.py
```diff
@@ -0,0 +1,153 @@
...
+    time_unit: Literal["ms", "us", "ns"],
+    starting_time_unit: Literal["ms", "us", "ns"],
+) -> None:
+    import pandas as pd
+    import polars as pl
+    import pyarrow as pa
+
+    @nw.narwhalify
+    def func(s: nw.Series) -> nw.Series:
+        return s.dt.timestamp(time_unit)  # type: ignore[return-value]
```
MarcoGorelli, line 143: this is due to https://github.com/narwhals-dev/narwhals/issues/1229

MarcoGorelli:
thanks @FBruzzesi for taking a look! `total_nanoseconds` is for `Duration`, not `Datetime`, so we can't use it here unfortunately 😅

FBruzzesi:
> thanks @FBruzzesi for taking a look! `total_nanoseconds` is for `Duration`, not `Datetime`, so we can't use it here unfortunately 😅

I am most likely missing something - what I am suggesting is, instead of using the date(time) representation, compute the difference between the series/expr of interest and the scalar 1970-01-01 00:00:00 (unix_epoch). This would result in a Duration object (wouldn't it?) and then use the Duration methods. Sorry if I misunderstood the internal representation explanation

MarcoGorelli:
ah sorry you're right 🤦 maybe that's indeed simpler, will try!

MarcoGorelli:
i think a cast would always be more efficient, from a little timing test:
```
In [50]: %timeit (s - datetime(1970,1,1)).dt.total_seconds()
465 μs ± 64.1 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

In [51]: %timeit s.astype('int64')
52.9 μs ± 10.8 μs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

tests/expr_and_series/dt/timestamp_test.py
```diff
@@ -0,0 +1,221 @@
...
+        ("us", "ms", [978307200000, None, 978480000000]),
+        ("ms", "ns", [978307200000000000, None, 978480000000000000]),
+        ("ms", "us", [978307200000000, None, 978480000000000]),
+        ("ms", "ms", [978307200000, None, 978480000000]),
+        ("s", "ns", [978307200000000000, None, 978480000000000000]),
+        ("s", "us", [978307200000000, None, 978480000000000]),
+        ("s", "ms", [978307200000, None, 978480000000]),
+    ],
+)
+def test_timestamp_datetimes_tz_aware(
```
FBruzzesi, line 84: Nice one 👌

narwhals/series.py
```diff
@@ -4001,3 +4001,63 @@ def convert_time_zone(self, time_zone: str) -> Series:
         return self._narwhals_series._from_compliant_series(
             self._narwhals_series._compliant_series.dt.convert_time_zone(time_zone)
         )
+
+    def timestamp(self, time_unit: Literal["ns", "us", "ms"] = "us") -> Series:
```
FBruzzesi, line 4005: Heads up if merging with main branch:

```suggestion
    def timestamp(self: Self, time_unit: Literal["ns", "us", "ms"] = "us") -> T:
```

(and same in Expr)

narwhals/_pandas_like/series.py
```diff
@@ -944,3 +946,29 @@ def convert_time_zone(self, time_zone: str) -> PandasLikeSeries:
...
             result = self._pandas_series._native_series.dt.tz_convert(time_zone)
         return self._pandas_series._from_native_series(result)
+
+    def timestamp(self, time_unit: Literal["ns", "us", "ms"] = "us") -> PandasLikeSeries:
+        s = self._pandas_series._native_series
+        dtype = self._pandas_series.dtype
+        is_pyarrow_dtype = "pyarrow" in str(self._pandas_series._native_series.dtype)
+        mask_na = s.isna()
+        if dtype == self._pandas_series._dtypes.Date:
+            s_cast = s.astype("Int32[pyarrow]")
```
FBruzzesi, line 956: That's because date can only be pyarrow backed, is that right?

narwhals/_pandas_like/series.py
```diff
@@ -944,3 +946,29 @@ def convert_time_zone(self, time_zone: str) -> PandasLikeSeries:
...
+            original_time_unit = dtype.time_unit  # type: ignore[attr-defined]
+            if (
+                self._pandas_series._implementation is Implementation.PANDAS
+                and self._pandas_series._backend_version < (2,)
+            ):  # pragma: no cover
+                s_cast = s.view("Int64[pyarrow]") if is_pyarrow_dtype else s.view("int64")
+            else:
+                s_cast = (
+                    s.astype("Int64[pyarrow]") if is_pyarrow_dtype else s.astype("int64")
+                )
```
FBruzzesi, line 968: I assume `.view` is more efficient?

FBruzzesi (commented):
Amazing stuff 🔥 I left a few comments for me to learn 🙈

narwhals/_pandas_like/series.py
MarcoGorelli: it's the most ridiculous thing - pandas 1.5 emits a warning telling you to use `view` instead of `astype`, and pandas 2.x emits a warning telling you to use `astype` instead of `view` because `view` is deprecated 🤦 🤦 🤦

narwhals/_pandas_like/series.py
MarcoGorelli: yup! will add a comment

MarcoGorelli (approved):
thanks @raisadz , and @FBruzzesi  for reviewing!

PR #1189, opened 2024-10-16
ci: only check doctests on Python3.12

MarcoGorelli:
It's frustrating to have to skip a lot of doctests because of version differences

I think we can just test them on the latest versions (like Polars does) - they're just intended to demonstrate functionality anyway, the proper testing is done in `tests`

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added 
- [ ] Documented the changes

PR #1222, opened 2024-10-18
chore: nox doctests only in Python 3.12

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

To be consistent with CI #1189 we should just run doctests in Python 3.12 in `nox`

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -582,9 +582,9 @@ class DataFrame(BaseFrame[DataFrameT]):
 
             We can then pass either pandas, Polars or PyArrow to `func`:
 
-            >>> func(df_pd)
-            >>> func(df_pl)
-            >>> func(df_pa)
+            >>> func(df_pd)  # doctest:+SKIP
+            >>> func(df_pl)  # doctest:+SKIP
+            >>> func(df_pa)  # doctest:+SKIP
         """
         self._compliant_frame.write_parquet(file)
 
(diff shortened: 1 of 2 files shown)


narwhals/dataframe.py
```diff
@@ -582,9 +582,9 @@ def write_parquet(self, file: str | Path | BytesIO) -> Any:
 
             We can then pass either pandas, Polars or PyArrow to `func`:
 
-            >>> func(df_pd)
-            >>> func(df_pl)
-            >>> func(df_pa)
+            >>> func(df_pd)  # doctest:+SKIP
+            >>> func(df_pl)  # doctest:+SKIP
+            >>> func(df_pa)  # doctest:+SKIP
```
EdAbati, line 587: these will create a `foo.parquet` locally. If we don't want to skip them, I'll add the file to `.gitignore`

what do you think? :)

FBruzzesi: I would opt for skipping those! Nice catch 👌

FBruzzesi (approved):
Awesome! Thanks @EdAbati 😁

DogHighFiveGIF

MarcoGorelli:
thanks both!

PR #1224, opened 2024-10-18
chore: refactor `compare_dicts`

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

I propose a couple of changes:
- I'd like to have a more verbose assertion error message when a unitest fail. Taking inspiration from `pandas`, I'm printing all the values of the columns that differ and the index where the difference was first noticed. This is particularly helpful for dataframes libraries that don't keep row order (* cough cough * _pyspark_ * cough cough *)

```python
def test_select(constructor: Constructor) -> None:
    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
    result = df.select("a")
    expected = {"a": [1, 3, 3]}
````

Before:
Screenshot 2024-10-18 at 23 50 35

After:
Screenshot 2024-10-18 at 23 49 06

- I think we should rename `compare_dicts` because we are not actually comparing 2 dictionaries. 😅 I propose something like `assert_equal_data`, any better name? This makes the diff of the PR a bit large, happy to revert if you think we should keep the previous name

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -54,14 +54,14 @@ def assert_equal_data(result: Any, expected: dict[str, Any]) -> None:
         expected_key = expected[key]
         for i, (lhs, rhs) in enumerate(zip_strict(result_key, expected_key)):
             if isinstance(lhs, float) and not math.isnan(lhs):
-                are_valid_values = math.isclose(lhs, rhs, rel_tol=0, abs_tol=1e-6)
+                are_equivalent_values = math.isclose(lhs, rhs, rel_tol=0, abs_tol=1e-6)
             elif isinstance(lhs, float) and math.isnan(lhs) and rhs is not None:
-                are_valid_values = math.isnan(rhs)  # pragma: no cover
+                are_equivalent_values = math.isnan(rhs)  # pragma: no cover
             elif pd.isna(lhs):
-                are_valid_values = pd.isna(rhs)
+                are_equivalent_values = pd.isna(rhs)
             else:
-                are_valid_values = lhs == rhs
-            assert are_valid_values, f"Mismatch at index {i}: {lhs} != {rhs}\nExpected: {expected}\nGot: {result}"
+                are_equivalent_values = lhs == rhs
+            assert are_equivalent_values, f"Mismatch at index {i}: {lhs} != {rhs}\nExpected: {expected}\nGot: {result}"
 
 
 def maybe_get_modin_df(df_pandas: pd.DataFrame) -> Any:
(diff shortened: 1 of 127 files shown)

### CODE REVIEW

File: tests/utils.py [hunk 1]
```diff
@@ -30,36 +30,47 @@ def zip_strict(left: Sequence[Any], right: Sequence[Any]) -> Iterator[Any]:
...
 
 
-def compare_dicts(result: Any, expected: dict[str, Any]) -> None:
+def _to_py_object(value: Any) -> Any:
+    # PyArrow: return scalars as Python objects
+    if hasattr(value, "as_py"):  # pragma: no cover
+        return value.as_py()
+    # cuDF: returns cupy scalars as Python objects
+    if hasattr(value, "item"):  # pragma: no cover
+        return value.item()
```
EdAbati, line 39: `_to_py_scalar` could be useful here too

tests/utils.py
```diff
@@ -30,36 +31,37 @@ def zip_strict(left: Sequence[Any], right: Sequence[Any]) -> Iterator[Any]:
...
             elif isinstance(lhs, float) and math.isnan(lhs) and rhs is not None:
-                assert math.isnan(rhs), (lhs, rhs)  # pragma: no cover
+                are_valid_values = math.isnan(rhs)  # pragma: no cover
             elif pd.isna(lhs):
-                assert pd.isna(rhs), (lhs, rhs)
+                are_valid_values = pd.isna(rhs)
             else:
-                assert lhs == rhs, (lhs, rhs)
+                are_valid_values = lhs == rhs
+            assert are_valid_values, f"Mismatch at index {i}: {lhs} != {rhs}\nExpected: {expected}\nGot: {result}"
```
FBruzzesi, line 64: Should we record the full diff instead of stopping at the first unequal encounter?

FBruzzesi (commented):
Thanks @EdAbati I like this a lot more! I left a comment but can also be thought as a follow up.

For the naming, I think this is the closest we have to what in pandas and polars is `assert_frame_equal`, but we are not there quite yet, so I am not sure 🙈

MarcoGorelli (approved):
really nice, thanks @EdAbati ! merging before merge conflicts creep in 😄 

i just renamed `are_valid_values` to `are_equivalent_values`, as I think the first one makes it sound like we're checking that they're not `null`-like values

thanks @FBruzzesi for reviewing - i think showing the full diff could be nice (perhaps, capped at 5 elements? in case we ever export this function publicly the diff ends up being of millions of elements...), let's leave it to a follow-up though

PR #1228, opened 2024-10-19
ci: add vegafusion in ci checks

### LINKED ISSUE #1193: add vegafusion to downstream ci tests
i think we should add vegafusion here: https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml

they haven't made a release yet, but if we're running in ci with the latest commit from their main branch, then I think that's already enough, we don't need to wait for a release

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1193

## Checklist

- [x] Tests added 
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -181,3 +181,38 @@ jobs:
         run: |
             cd py-shiny
             make narwhals-test-integration
+    
+  vegafusion:
+     env:
+       UV_SYSTEM_PYTHON: true
+
+     strategy:
+       matrix:
+         python-version: ["3.12"]
+         os: [ubuntu-latest]
+
+     runs-on: ${{ matrix.os }}
+     steps:
+        - uses: actions/checkout@v4
+        - uses: actions/setup-python@v5
+          with:
+            python-version: ${{ matrix.python-version }}
+        - name: Install uv
+          uses: astral-sh/setup-uv@v3
+          with:
+            enable-cache: "true"
+            cache-suffix: ${{ matrix.python-version }}
+            cache-dependency-glob: "**requirements*.txt"
+        - uses: prefix-dev/setup-pixi@v0.8.1
+          with:
+            pixi-version: v0.27.1
+        - name: clone-vegafusion
+          run: |
+              git clone https://github.com/vega/vegafusion.git --depth=1
+              cd vegafusion
+              git log
+        - name: Build vegafusion wheel
+          run: |
+              cd vegafusion    
+              pixi run build-py-vegafusion
+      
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
thanks @EdAbati - this is awesome, but " Successful in 27m " is probably too much for every commit 😄 

perhaps we can only trigger this one if we apply some label, so then we can just trigger this before releases?

EdAbati:
Ok CI is green, but it takes quite a lot. Using caching, the build time decreases but `pytest` takes ~24m (AFAIK similar time as in the vegafusion CI: https://github.com/vega/vegafusion/actions/runs/11419942913/job/31775120908)

Do you know any tricks to speed this up a bit?

Otherwise I would suggest to treat all these downstream libraries tests as "integration tests" and not run them at every push. For example, we could run them just before merging a PR. Is it by using `auto_merge_enabled` [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)? (Never done this before though)

Thoughts? @MarcoGorelli @FBruzzesi

EdAbati:
> perhaps we can only trigger this one if we apply some label, so then we can just trigger this before releases?

Happy to do that. I don't have strong opinions on what should trigger this or when to test these.

Should we run these test before release and triggered by new tag?

.github/workflows/downstream_tests.yml
```diff
@@ -1,9 +1,8 @@
 name: Test Downstream Libraries
 
-on:
-  pull_request:
-  push:
-    branches: [main]
+on: 
+  workflow_call:
+  workflow_dispatch:
```
EdAbati, line 5: not sure if this is what I meant, but now this can be triggered:
- by other workflows (i.e. in the 1st step of the publish workflow [here](https://github.com/narwhals-dev/narwhals/pull/1228/files#diff-5bdab82401d84fafcce36d5d8c3d9ba4c3682afec172b137defae42bb5f27d43R11)
- manually

MarcoGorelli (approved):
let's give this a go - thanks @EdAbati !

MarcoGorelli:
@EdAbati is it possible to have most downstream tests run on every commit, but only the slow ones (for now, just vegafusion) only run before releases?

EdAbati:
> is it possible to have most downstream tests run on every commit

I will have a look today 👀

PR #1276, opened 2024-10-28
perf: make to_py_scalar 3x faster when argument is str

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

### DIFF
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -46,6 +46,13 @@ if TYPE_CHECKING:
 
 T = TypeVar("T")
 
+NON_TEMPORAL_SCALAR_TYPES = (
+    bool,
+    bytes,
+    str,
+    numbers.Number,
+)
+
 
 @overload
 def to_native(
@@ -843,7 +850,8 @@ def to_py_scalar(scalar_like: Any) -> Any:
         >>> nw.to_py_scalar(1)
         1
     """
-
+    if isinstance(scalar_like, NON_TEMPORAL_SCALAR_TYPES):
+        return scalar_like
     pa = get_pyarrow()
     if pa and isinstance(scalar_like, pa.Scalar):
         return scalar_like.as_py()
@@ -864,18 +872,11 @@ def to_py_scalar(scalar_like: Any) -> Any:
     if pd and isinstance(scalar_like, pd.Timedelta):
         return scalar_like.to_pytimedelta()
 
-    all_scalar_types = (
-        int,
-        float,
-        complex,
-        bool,
-        bytes,
-        str,
-        datetime,
-        timedelta,
-        numbers.Number,
-    )
-    if isinstance(scalar_like, all_scalar_types):
+    # SIM101 suggests merging the tuples, but that would create a new tuple,
+    # so doesn't seem ideal perf-wise?
+    if isinstance(scalar_like, NON_TEMPORAL_SCALAR_TYPES) or isinstance(  # noqa: SIM101
+        scalar_like, (datetime, timedelta)
+    ):
         return scalar_like
 
     msg = (
--- a/tests/stable_api_test.py
+++ b/tests/stable_api_test.py
@@ -87,6 +87,10 @@ def test_stable_api_docstrings() -> None:
         nw_doc = getattr(nw, item).__doc__
         if item == "from_native":
             v1_doc = v1_doc.replace("native_dataframe", "native_object")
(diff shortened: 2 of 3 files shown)


MarcoGorelli:
@EdAbati fancy taking a look?

narwhals/translate.py
```diff
@@ -864,18 +873,11 @@ def to_py_scalar(scalar_like: Any) -> Any:
...
-        str,
-        datetime,
-        timedelta,
-        numbers.Number,
-    )
-    if isinstance(scalar_like, all_scalar_types):
+    # SIM101 suggests merging the tuples, but that would create a new tuple,
+    # so doesn't seem ideal perf-wise?
+    if isinstance(scalar_like, NON_TEMPORAL_SCALAR_TYPES) or isinstance(  # noqa: SIM101
+        scalar_like, (datetime, timedelta)
```
FBruzzesi, line 879: I am ok with removing the check for numbers.Number, yet, was it intended?

Also, what's the reason to keep the check for non temporal types if they return early at the beginning?

FBruzzesi (commented):
Commenting from mobile, just have a tiny question :)

narwhals/translate.py
MarcoGorelli: 😄 sometimes you just gotta review on the go

---

yeah if we have int, float, complex, then I'm not aware of what number.Number adds

furthermore, we have:
```python
In [10]: isinstance(np.int64(1), numbers.Number)
Out[10]: True
```
so checking `number.Number` early on would cause numpy scalars to not be converted to Python scalars

---

`pd.Timestamp` and `pd.Timedelta` would satisfy `isinstance(..., (datetime, timedelta))`, despite not being Python scalars, so if we want to do a cheap early return I think we need to exclude them

FBruzzesi: Yep, super on go.

Got it for numbers! 👌

For the other comment I meant, why is it needed to have `isinstance(scalar_like, NON_TEMPORAL_DTYPES)` before checking for temporal in line above?

MarcoGorelli: ah right, all the branches above just return, I thought something was modifying the input and passing it down, thanks!

FBruzzesi (approved):
Thanks for the quick adjustment @MarcoGorelli 🙌🏼🚀

DogOmwGIF

tests/stable_api_test.py
```diff
@@ -87,6 +87,10 @@ def test_stable_api_docstrings() -> None:
         nw_doc = getattr(nw, item).__doc__
         if item == "from_native":
             v1_doc = v1_doc.replace("native_dataframe", "native_object")
+        if item in {"to_py_scalar"}:
+            # We don't overwrite the docstring for these
```
EdAbati, line 91: Just so I know what to do/review in the future, which kind of methods should be added here?

EdAbati (approved):
Thanks @MarcoGorelli, I like to see/learn about ways to speed up code 

> so checking number.Number early on would cause numpy scalars to not be converted to Python scalars

should we also add `(np.int64(1), 1)` in the tests to capture this? (at the moment that is indirectly tested in the doctests)

and the rest looks great!

narwhals/translate.py
```diff
@@ -46,6 +45,15 @@
 
 T = TypeVar("T")
 
+NON_TEMPORAL_SCALAR_TYPES = (
```
FBruzzesi, line 48: Kind of related, I am running a test, and getting a raise if None enters this function. Should we add it to non temporal scalar types?

tests/stable_api_test.py
MarcoGorelli: I think all of them to be honest, it's probably simpler to just use `import narwhals as nw` in all docstrings and forget about this test 😅 i'll make a separate pr

narwhals/translate.py
MarcoGorelli: _very_ good point, thanks!

MarcoGorelli:
_really_ useful comments here, thanks both for excellent reviews 🙏

Issue #1254, opened 2024-10-25
feat: support rolling / ewm


MarcoGorelli:
these came up in the Plotly PR, and would help get to 100% cross-dataframe compatibility

FBruzzesi:
@DeaMariaLeon I don't really have a plan/schedule on which issues to work on 🙈 Feel free to pick it up or split the work (rolling and ewm).

DeaMariaLeon:
I started working on ewm.

DeaMariaLeon:
Besides `ewm_mean`, Polars has `ewm_mean_by`, `ewm_std`and `ewm_var`. Do we need all of them?  Thanks @MarcoGorelli or @FBruzzesi

MarcoGorelli:
hey - I think this one might be enough for now to see if it's useful to Plotly, i'm a little hesitant to expand the API too much before we have use cases

PR #1290, opened 2024-10-30
feat: add `Expr|Series.rolling_mean` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1254 

## Checklist

- [x] Tests added 
- [x] Documented the changes

Opening as draft because of the following:

- Dask is raising an error
- Pandas does not support weights, we could follow the same approach as per arrow, only if weights are provided
- For arrow I tried to keep a generic approach that can be re-used for ~any~ most other aggregate functions, but would like a feedback on that


MarcoGorelli:
thanks @FBruzzesi ! `weights` might even get removed from the polars ones, shall we keep it out for now?

FBruzzesi:
> thanks @FBruzzesi ! `weights` might even get removed from the polars ones, shall we keep it out for now?

Sure, even better to start simpler from there :)

narwhals/_arrow/utils.py
```diff
@@ -452,3 +455,40 @@ def _parse_time_format(arr: pa.Array) -> str:
         if pc.all(matches.is_valid()).as_py():
             return time_fmt
     return ""
+
+
+def _window_agg(
```
MarcoGorelli, line 460: we're still iterating in Python here...dunno, maybe we should just raise NotImplementedError here?

curious to see a timing comparison of:
- convert to pandas, do rolling mean, back to pyarrow
- do it with this function

MarcoGorelli (commented):
awesome, thanks

pending on the unstable api warning being introduced (https://github.com/narwhals-dev/narwhals/issues/1367), I just have a comment about pyarrow

if we're not sure, i think it's ok to leave it out for now

maybe one day we can get to the point where just having something be part of the Narwhals API is enough to put light pressure on dataframe authors to support a function 😄

narwhals/_arrow/utils.py
FBruzzesi: I will give it a shot and report back :)

FBruzzesi:
Thanks for the feedback Marco
 
> if we're not sure, i think it's ok to leave it out for now

My concern is that for plotly we convert dataframe that support interchange protocol to pyarrow and not to pandas. I am afraid that not supporting it would break _some_ user workflows. If that's how we aim to proceed, I would suggest to convert interchange protocol to pandas as it was done before the narwhals PR

> maybe one day we can get to the point where just having something be part of the Narwhals API is enough to put light pressure on dataframe authors to support a function 😄

That would be nice 😁

narwhals/_arrow/utils.py
FBruzzesi: Yep this is definitly orders of magnitude slower. I will try with a specified implementation

MarcoGorelli:
are we sure it would break a workflow? as far as I can tell, in plotly, you convert to pandas for non-pandas input in trendline

so, if someone was using plotly before and interchanging to pandas, then it means they already have pandas installed, so this shouldn't break their workflow if i understand correctly

FBruzzesi:
> are we sure it would break a workflow? as far as I can tell, in plotly, you convert to pandas for non-pandas input in trendline

Yes correct, this is the current behavior, yet the use case for supporting rolling, expanding, and ewm is to eventually completely remove such conversion in trendline plotly module and use the narwhals functionalities instead.

Therefore, if a user provides a dataframe with interchange protocol that we don't support natively, then it gets [converted to arrow](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_core.py#L1690), and if rolling is not supported, then it would end up breaking.

Does this make sense?

MarcoGorelli:
> and use the narwhals functionalities instead

would it work to use the narwhals functionalities if available, and otherwise just convert to pandas?

MarcoGorelli (approved):
you're too cool for school

narwhals/_arrow/series.py
```diff
@@ -922,6 +922,23 @@ def rolling_sum(
...
+    ) -> Self:
+        rolling_sum = self.rolling_sum(
+            window_size=window_size, min_periods=min_periods, center=center
+        )
+        rolling_count = (
+            (~self.is_null())
+            .cast(self._dtypes.Int32())
+            .rolling_sum(window_size=window_size, min_periods=min_periods, center=center)
+        )
+        return rolling_sum / rolling_count
```
FBruzzesi, line 940: This is a lazy way of doing it. Performances can be enhanced with the same routine of `rolling_sum`, and dividing by the count at the end. I will address it

PR #1298, opened 2024-11-01
feat: Adding ewm_mean

DeaMariaLeon:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

I didn't add anything for Arrow because I'm waiting to see the feedback for https://github.com/narwhals-dev/narwhals/pull/1290


DeaMariaLeon:
The javascript is to add Latex: 

Screenshot 2024-11-01 at 15 41 59

DeaMariaLeon:
Added one test for `adjust`. 

I haven't added anything for Arrow because I'm not sure if we want to add it or not.

tests/expr_and_series/ewm_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+import pandas as pd
+import pytest
+
+import narwhals.stable.v1 as nw
+from tests.utils import Constructor
+from tests.utils import ConstructorEager
+from tests.utils import assert_equal_data
+
+data = {"a": [1, 1, 2], "b": [1, 2, 3]}
```
MarcoGorelli, line 11: can we include a test with nulls too please?

tests/expr_and_series/ewm_test.py
```diff
@@ -0,0 +1,72 @@
...
+    assert_equal_data({"a": result}, expected)
+
+
+@pytest.mark.parametrize("adjust", [True, False])
+def test_ewm_mean_expr_adjust(
+    constructor: Constructor,
+    adjust: bool,  # noqa: FBT001
+) -> None:
+    if "pyarrow_" in str(constructor) or "dask" in str(constructor):  # remove
+        pytest.skip()
```
MarcoGorelli, line 43: we can use `request.applymarker(pytest.mark.xfail)` please? then the test actually runs and we check that it fails, as opposed to being skipped (also, if i remember correctly `pytest.skip` had some undesirable behaviour)

MarcoGorelli (commented):
thanks for updating! i think this is close

narwhals/series.py
```diff
@@ -464,6 +464,33 @@ def ewm_mean(
                2.428571
             ]
 
+            pandas and Polars handle nulls differently. So, calculating ewm over
+            a sequence with null values leads to distinct results:
```
MarcoGorelli, line 468: I think it's that Polars preserves null values, whereas pandas forward-fills

Can we preserve null values for pandas too?

MarcoGorelli: I don't think you need to fill the nulls, but just preserve them - so, if a value was null to start with, it should be null in the result too

we do something like that in `timestamp`

DeaMariaLeon: Doing like you do in `timestamp` solves it for `None`.. but with a series like `[1.0, float("nan"), 4.0]` we still have this:

Pandas:
```
0    1.0
1    NaN
2    3.4
dtype: float64
```

Polars:
```
shape: (3,)
Series: '' [f64]
[
        1.0
        NaN
        NaN
]
```

MarcoGorelli: sure, cause Polars treat 'nan' differently from null - but if we use the null value for both, does the result match? e.g. `[1., None, 4.]`?

DeaMariaLeon: Yes, with a Series like `[1.0, None, float("nan"), 4.0]` we get this:

```
pandas:

0    1.000000
1         NaN
2         NaN
3    3.666667

polars:

dtype: float64
shape: (4,)
Series: '' [f64]
[
        1.0
        null
        NaN
        NaN
]
```

MarcoGorelli: yeah I think just using `None` in the data that you pass to the constructor in the test should be fine

DeaMariaLeon: Test modified.. 
To resume then:  If someone uses a pandas df, they won't get pandas normal results. We match polars behaviour when "None" is in the input, but we don't when there is "NaN". 🤯 .. 😁

CI failing with older versions but I can't reproduce locally.

MarcoGorelli: for Polars 'nan' is only the result of illegal mathematical operations (like 0/0) so it's far rarer to encounter it there

regarding older versions ci - i'd suggest making a separate virtual environment and installing the versions which show up in the `show deps` step of the ci job

DeaMariaLeon: Older versions of polars give similar results to pandas when there is a null (`None`).

So with an input of: `{"a": [2.0, 4.0, None, 3.0]}` Then:
```
Expected: {'a': [2.0, 3.3333333333333335, nan, 3.142857142857143]}
Got: {'a': [2.0, 3.3333333333333335, 3.3333333333333335, 3.142857142857143]}
```

For the moment I'm "xfailing" that test with older versions.  I'm not sure if that's correct.

MarcoGorelli: thanks - is it possible to use `pl.when` to preserve the null values for old versions of Polars?

either that, or raise `NotImplementedError` for old versions of Polars for now, and let's create an issue to track preserving null values in old Polars versions

DeaMariaLeon: Thanks for the review.. I added the "raise" for now. I'll do the follow up PR.

tests/expr_and_series/ewm_test.py
```diff
@@ -0,0 +1,118 @@
...
+) -> None:
+    if any(x in str(constructor) for x in ("pyarrow_table_", "dask", "modin")) or (
+        "polars" in str(constructor) and POLARS_VERSION <= (0, 20, 31)
+    ):
+        request.applymarker(pytest.mark.xfail)
+
+    df = nw.from_native(constructor({"a": [2.0, 4.0, None, 3.0]}))
+    result = df.select(nw.col("a").ewm_mean(com=1, ignore_nulls=ignore_nulls))
+
+    if ignore_nulls:
```
MarcoGorelli, line 99: i'd suggest to include the list as something you parametrise over, rather than inclusiding logic (if/then) in the test. in general, we should use if/then in tests only when necessary, it's something i try to avoid if possible (and sometimes it's not possible unfortunately)

narwhals/_pandas_like/series.py
```diff
@@ -174,6 +174,25 @@ def dtype(self: Self) -> DType:
...
+    def ewm_mean(
+        self,
+        *,
+        com: float | None = None,
+        span: float | None = None,
+        half_life: float | None = None,
+        alpha: float | None = None,
+        adjust: bool = True,
+        min_periods: int = 1,
+        ignore_nulls: bool = False,
```
MarcoGorelli, line 186: there's a lot of parameters here, do we have a test which hits each of them?

tests/expr_and_series/ewm_test.py
```diff
@@ -0,0 +1,118 @@
...
+    if adjust:
+        expected = {
+            "a": [1.0, 1.0, 1.5714285714285714],
+            "b": [1.0, 1.6666666666666667, 2.4285714285714284],
+        }
+    else:
+        expected = {
+            "a": [1.0, 1.0, 1.5],
+            "b": [1.0, 1.5, 2.25],
+        }
```
MarcoGorelli, line 67: same

narwhals/series.py
MarcoGorelli: thanks!

tests/expr_and_series/ewm_test.py
DeaMariaLeon: Changed that.

narwhals/_pandas_like/series.py
DeaMariaLeon: I'll add tests for the parameters then... (I can't hit all the parameters in only one test, at least not the first 4 I think).. Is that what you meant?

DeaMariaLeon:
I added tests for the parameters.

DeaMariaLeon:
Added the `NarwhalsUnstableWarning` 😇

PR #1301, opened 2024-11-01
feat: add `native` property and deprecate `.to_native()` method

### LINKED ISSUE #1295: add `DataFrame.native` / `LazyFrame.native` / `Series.native` properties
`to_native()` hides the fact that it's a practically free operation, it just calls `._compliant_frame._native_frame`

we should probably make `native` a property, but keep `to_native` around in `stable.v1`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1295 

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -2912,7 +2912,7 @@ class LazyFrame(BaseFrame[FrameT]):
             | Use `.to_native` to see native output |
             └───────────────────────────────────────┘
             >>> df = lf.group_by("a").agg(nw.all().sum()).collect()
-            >>> df.to_native().sort("a")
+            >>> df.native.sort("a")
             shape: (3, 3)
             ┌─────┬─────┬─────┐
             │ a   ┆ b   ┆ c   │
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -346,14 +346,14 @@ class DataFrame(BaseFrame[DataFrameT]):
             >>> df_pl = pl.DataFrame(data)
             >>> df_pa = pa.table(data)
 
-            Calling `to_native` on a Narwhals DataFrame returns the native object:
+            Calling `.native` on a Narwhals DataFrame returns the native object:
 
-            >>> nw.from_native(df_pd).to_native()
+            >>> nw.from_native(df_pd).native
                foo  bar ham
             0    1  6.0   a
             1    2  7.0   b
             2    3  8.0   c
-            >>> nw.from_native(df_pl).to_native()
+            >>> nw.from_native(df_pl).native
             shape: (3, 3)
             ┌─────┬─────┬─────┐
             │ foo ┆ bar ┆ ham │
@@ -364,7 +364,7 @@ class DataFrame(BaseFrame[DataFrameT]):
             │ 2   ┆ 7.0 ┆ b   │
             │ 3   ┆ 8.0 ┆ c   │
             └─────┴─────┴─────┘
-            >>> nw.from_native(df_pa).to_native()
+            >>> nw.from_native(df_pa).native
             pyarrow.Table
             foo: int64
             bar: double
@@ -497,12 +497,12 @@ class DataFrame(BaseFrame[DataFrameT]):
 
             Calling `to_native` on a Narwhals DataFrame returns the native object:
 
-            >>> nw.from_native(df_pd).to_native()
+            >>> nw.from_native(df_pd).to_native()  # doctest:+SKIP
                foo  bar ham
             0    1  6.0   a
             1    2  7.0   b
             2    3  8.0   c
-            >>> nw.from_native(df_pl).to_native()
+            >>> nw.from_native(df_pl).to_native()  # doctest:+SKIP
             shape: (3, 3)
             ┌─────┬─────┬─────┐
             │ foo ┆ bar ┆ ham │
@@ -2836,14 +2836,14 @@ class LazyFrame(BaseFrame[FrameT]):
             >>> df_pl = pl.LazyFrame(data)
             >>> df_pa = pa.table(data)
 
-            Calling `to_native` on a Narwhals DataFrame returns the native object:
+            Calling `.native` on a Narwhals DataFrame returns the native object:
 
-            >>> nw.from_native(df_pd).lazy().to_native()
+            >>> nw.from_native(df_pd).lazy().native
                foo  bar ham
             0    1  6.0   a
             1    2  7.0   b
             2    3  8.0   c
-            >>> nw.from_native(df_pl).to_native().collect()
+            >>> nw.from_native(df_pl).native.collect()
             shape: (3, 3)
             ┌─────┬─────┬─────┐
             │ foo ┆ bar ┆ ham │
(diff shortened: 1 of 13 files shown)


narwhals/dataframe.py
```diff
@@ -328,6 +328,11 @@ def _series(self) -> type[Series]:
     def _lazyframe(self) -> type[LazyFrame[Any]]:
         return LazyFrame
 
+    @property
+    def native(self: Self) -> DataFrameT:
+        """Returns native frame underlying Narwhals DataFrame."""
+        return self._compliant_frame._native_frame  # type: ignore[no-any-return]
```
FBruzzesi, line 334: Shall I add examples?

MarcoGorelli:
nice!

- should we use this in all the public docs places where `to_native` currently appears?
- I think we should have it such that in the main namespace, `DataFrame.to_native` / `LazyFrame.to_native` / `Series.to_native` issue deprecation warnings, but not in `stable.v1` (could be a follow-up)
- i think the same docstring as the current `to_native` functions would be good?

FBruzzesi:
I updated accordingly, let me leave a couple of comments:

> * should we use this in all the public docs places where `to_native` currently appears?

I replaced only the cases for `.to_native()` method (one instance), not `nw.to_native(...)`

> * I think we should have it such that in the main namespace, `DataFrame.to_native` / `LazyFrame.to_native` / `Series.to_native` issue deprecation warnings, but not in `stable.v1` (could be a follow-up)

I added this, and tested as well. Yet, since it is a free operation, I don't think it is a big drawback keeping it. I personally like the `to_native()` api better (not sure why).

> * i think the same docstring as the current `to_native` functions would be good?

👍🏼

FBruzzesi:
My understanding is that CI is failing because all the imports in stable v1 docstrings changed

```diff
+ import narwhals as nw
- import narwhals.stable.v1 as nw
```

how's that the case?

MarcoGorelli:
thanks!

> I added this, and tested as well. Yet, since it is a free operation, I don't think it is a big drawback keeping it. I personally like the to_native() api better (not sure why).

🤔 maybe we can bikeshed on this longer then 😄 longterm I think it'd be better to stabilise on a single api...on the one hand, I like the symmetry of to/from, but on the other hand, I do like that `.native` makes it look more like it is (a pratically free operation)

MarcoGorelli:
> My understanding is that CI is failing because all the imports in stable v1 docstrings changed
> 
> ```diff
> + import narwhals as nw
> - import narwhals.stable.v1 as nw
> ```
> 
> how's that the case?

sorry yeah this was in https://github.com/narwhals-dev/narwhals/pull/1278

FBruzzesi:
> sorry yeah this was in #1278

Just to be sure, was this intended?

MarcoGorelli:
CI looks like it's green now, I think the only failure is because of deprecation warnings having changed in py13, looking into it

FBruzzesi:
> CI looks like it's green now, I think the only failure is because of deprecation warnings having changed in py13, looking into it

Found the doctest that was raising, I changed it to call `.native` instead, and I am skipping the `.to_native()` tests in stable api v1. Those would fail unless we change it back to the import `import narwhals.stable.v1 as nw`

MarcoGorelli:
thanks!

> I personally like the to_native() api better (not sure why).

shall we sleep on this one a bit longer then? i don't feel super-strongly about this one tbh, happy to keep `to_native()` if you prefer it

FBruzzesi:
> I personally like the to_native() api better (not sure why).

Would it be bad to introduce `.native` without deprecating `.to_native()`? I understand it may be confusing for a user. Maybe changing the description of `.to_native()` to explicitly mention that's just an alias for `.native` and that's a free operation.

EdAbati:
I also don't have a strong preference.
I see why `.native` hints that it is a 'free' operation but I kind of like how `to_native` reads 😅 (maybe just because I got used to it)

We could keep it as alias for now, and move the discussion to an issue. In the meantime, we can suggest `.native` in every docs to signal it is the 'preferred way'.

MarcoGorelli:
maybe this can be a good topic for discussion for tomorrow's community call?

dangotbanned:
I think adding `Compliant*.native` could be a pretty nice improvement internally.

Shortens paths like this, where lots of places use inconsistent aliasing:

https://github.com/narwhals-dev/narwhals/blob/b03f7160c6c7beed23d83e1daf857d4be7327256/narwhals/_arrow/series.py#L319

https://github.com/narwhals-dev/narwhals/blob/b03f7160c6c7beed23d83e1daf857d4be7327256/narwhals/_arrow/series.py#L771

https://github.com/narwhals-dev/narwhals/blob/b03f7160c6c7beed23d83e1daf857d4be7327256/narwhals/_arrow/series.py#L741


https://github.com/narwhals-dev/narwhals/blob/b03f7160c6c7beed23d83e1daf857d4be7327256/narwhals/_arrow/series.py#L896

https://github.com/narwhals-dev/narwhals/blob/b03f7160c6c7beed23d83e1daf857d4be7327256/narwhals/_arrow/series.py#L832

Being a `@property` is also a good signal that we shouldn't be using in place/mutating operations.

Also for things like https://github.com/narwhals-dev/narwhals/issues/2044#issuecomment-2674262833, there's an obvious place to hint the `TypeVar` and not expose internal attributes

FBruzzesi:
Thanks for reviewing @dangotbanned!

The original PR was not for introducing `native` at the complaint level, rather at Narwhals one.

I think having `Compliant*.native` can be a quick win and using properties we can have more control via its `setter` if we need to

dangotbanned:
@FBruzzesi I've tested this out for some parts of `_arrow` over on https://github.com/narwhals-dev/narwhals/compare/main...stores-native-compliant

Really think it's a nice improvement to readabilty and [DRY](https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself)s things up well 🙂

With a lot of the repetition out of the way, I was able to spot a few more things that could be written more concisely.

Not sure if this is worth opening an issue for - but if we went ahead with it - then it'd probably make sense to try and do it in one or more big chunks to avoid conflicts

FBruzzesi:
> @FBruzzesi I've tested this out for some parts of `_arrow` over on [main...stores-native-compliant](https://github.com/narwhals-dev/narwhals/compare/main...stores-native-compliant)
>
> Really think it's a nice improvement to readabilty and [DRY](https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself)s things up well 🙂
> 
> With a lot of the repetition out of the way, I was able to spot a few more things that could be written more concisely.
> 
> Not sure if this is worth opening an issue for - but if we went ahead with it - then it'd probably make sense to try and do it in one or more big chunks to avoid conflicts

Hey @dangotbanned sorry I had the chance to look at the code only now. It definitly looks a bit different from what this PR was trying to do.
I think adding `native` at the compliant level will have no opposition as it's internal only anyway.

Regarding `native` property at the Narwhals level, I am still not fully sure. I personally like `to_native()` method, yet we should specify in the docs that this is a no-op as users might expect that some particular computation happens.

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/1301#issuecomment-2689320953

Yeah I agree @FBruzzesi wrt the public `to_native()` - I think it also leaves some wiggle-room in the future for adding parameters (if a use-case arises)

Apologies for hijacking the PR w/ (https://github.com/narwhals-dev/narwhals/pull/1301#issuecomment-2686042530) - I'll continue that as it's own thing 👍

FBruzzesi:
Closing this for now

Issue #1239, opened 2024-10-22
[Enh]: allow passing index object directly to `maybe_set_index`


FBruzzesi:
### Please describe the purpose of the new feature or describe the problem to solve.

`maybe_set_index` does not allow to pass the index directly, which could be useful for Series objects

rik-vandervlist-sympower:
I'm looking into this one!

Riik:
I'm looking into this one!

PR #1319, opened 2024-11-04
feat: support passing index object directly into maybe_set_index

Riik:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue #1239
- Closes #

## Checklist

- [x] Tests added 
- [x] Documented the changes

[Pandas set_index docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.set_index.html#pandas-dataframe-set-index)
[Pandas set_index test cases](https://github.com/pandas-dev/pandas/blob/9a015d19514597ad5d1c31f566a0a2bdb17d4cc5/pandas/tests/frame/methods/test_set_index.py#L309)
I've renamed the `column_names: str | list[str]` parameter to `keys: str | Series | list[Series | str]`. I referenced the Pandas docs where which have `keys: label or array-like or list of labels/arrays`. Since this is a backwards-compatability function, I figured it is good for it to be close to the Pandas spec, and the original param name `column` doesn't cover all cases anymore.

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -322,7 +322,7 @@ def maybe_set_index(
                 )
             )
 
-        if index is not None:
+        if index is not None:  # pragma: no cover
             from narwhals.series import Series
 
             if _is_iterable(index):
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -307,11 +307,11 @@ def maybe_set_index(
     native_frame = to_native(df_any)
 
     if column_names is not None and index is not None:
-        msg = "Only one of `column_names` or `keys` should be provided"
+        msg = "Only one of `column_names` or `index` should be provided"
         raise ValueError(msg)
 
     if not column_names and not index:
-        msg = "Either `column_names` or `keys` should be provided"
+        msg = "Either `column_names` or `index` should be provided"
         raise ValueError(msg)
 
     if is_pandas_like_dataframe(native_frame):
@@ -327,7 +327,7 @@ def maybe_set_index(
 
             if _is_iterable(index):
                 index = [
-                    key.to_native() if isinstance(key, Series) else key for key in index
+                    idx.to_native() if isinstance(idx, Series) else idx for idx in index
                 ]
             if isinstance(index, Series):
                 index = index.to_native()
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -333,9 +333,8 @@ def maybe_set_index(
         if column_names:
             msg = "Cannot set index using column names on a Series"
             raise ValueError(msg)
-        native_obj.index = keys
         return df_any._from_compliant_series(  # type: ignore[no-any-return]
-            df_any._compliant_series._from_native_series(native_obj)
+            df_any._compliant_series._from_native_series(native_obj.set_axis(keys))
         )
     else:
         return df_any  # type: ignore[no-any-return]
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -333,9 +333,9 @@ def maybe_set_index(
         if column_names:
             msg = "Cannot set index using column names on a Series"
             raise ValueError(msg)
-
+        native_obj.index = keys
         return df_any._from_compliant_series(  # type: ignore[no-any-return]
-            df_any._compliant_series._from_native_series(native_obj.set_axis(keys))
+            df_any._compliant_series._from_native_series(native_obj)
         )
     else:
         return df_any  # type: ignore[no-any-return]
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -275,16 +275,16 @@ def maybe_get_index(obj: T) -> Any | None:
 
 
 def maybe_set_index(
-    df: T,
+    obj: T,
     column_names: str | list[str] | None = None,
     *,
     index: Series | list[Series] | None = None,
 ) -> T:
     """
-    Set the index of `df`, if `df` is pandas-like, otherwise this is a no-op.
+    Set the index of a DataFrame or a Series, if it's pandas-like.
 
     Arguments:
-        df: object for which maybe set the index (can be either a Narwhals `DataFrame`
+        obj: object for which maybe set the index (can be either a Narwhals `DataFrame`
             or `Series`).
         column_names: name or list of names of the columns to set as index.
             For dataframes, only one of `column_names` and `index` can be specified but
@@ -320,7 +320,7 @@ def maybe_set_index(
         5  2
     """
 
-    df_any = cast(Any, df)
+    df_any = cast(Any, obj)
     native_obj = to_native(df_any)
 
     if column_names is not None and index is not None:
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -282,10 +282,12 @@ def maybe_set_index(
 ) -> T:
     """
     Set the index of `df`, if `df` is pandas-like. The index can either be specified as
-    a existing column name or list of column names with `column_names`, or set directly with a Series or list of Series with `index`.
+    a existing column name or list of column names with `column_names`, or set directly
+    with a Series or list of Series with `index`.
+
     Notes:
-        This is only really intended for backwards-compatibility purposes,
-        for example if your library already aligns indices for users.
+        This is only really intended for backwards-compatibility purposes, for example if
+        your library already aligns indices for users.
         If you're designing a new library, we highly encourage you to not
         rely on the Index.
         For non-pandas-like inputs, this is a no-op.
@@ -304,7 +306,7 @@ def maybe_set_index(
     """
 
     df_any = cast(Any, df)
-    native_frame = to_native(df_any)
+    native_obj = to_native(df_any)
 
     if column_names is not None and index is not None:
         msg = "Only one of `column_names` or `index` should be provided"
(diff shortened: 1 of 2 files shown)


MarcoGorelli:
thanks for your PR!

this would break backwards-compatibility, so I don't think we can just rename to `keys` unfortunately

Riik:
Right, so we preserve the `column_names` for backwards comparability, add index, make both optional, and check that they are never both specified. They new signature will then be 
```
def maybe_set_index(df: T, column_names: str | list[str] | None, index: Series | list[Series] | None) -> T
```
I'll give that a try!

narwhals/utils.py
```diff
@@ -297,14 +302,40 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
...
     df_any = cast(Any, df)
     native_frame = to_native(df_any)
+
+    if column_names is not None and index is not None:
+        msg = "Only one of `column_names` or `keys` should be provided"
+        raise ValueError(msg)
+
+    if not column_names and not index:
+        msg = "Either `column_names` or `keys` should be provided"
+        raise ValueError(msg)
```
FBruzzesi, line 315:
```suggestion
    if column_names is not None and index is not None:
        msg = "Only one of `column_names` or `index` should be provided"
        raise ValueError(msg)

    if not column_names and not index:
        msg = "Either `column_names` or `index` should be provided"
        raise ValueError(msg)
```

narwhals/utils.py
```diff
@@ -297,14 +302,40 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
...
+
+        if index is not None:
+            from narwhals.series import Series
+
+            if _is_iterable(index):
+                index = [
+                    key.to_native() if isinstance(key, Series) else key for key in index
+                ]
+            if isinstance(index, Series):
+                index = index.to_native()
```
FBruzzesi, line 333: How about the following? (will require formatting and maybe ignoring mypy)

```suggestion
        index = [to_native(key, pass_through=True) for key in index] if _is_iterable(index) else [to_native(index, pass_through=True)]
```

FBruzzesi (commented):
Hey @Riik thanks for the effort! I left a minor comment and a suggestion to simplify the logic a bit.

Additionally, could we handle the case for which `df` is actually a series? That's the use case I had in mind when I opened the related issue. I.e. if the native_object is a pandas like series, then I would like to modify its index and clearly that's not possible to do unless another Series/Index/iterable is passed.

Since that's the case, I think we should also raise an error if `native_object` is a series but `column_names` is provided

Riik:
> Additionally, could we handle the case for which df is actually a series? That's the use case I had in mind when I opened the related issue. I.e. if the native_object is a pandas like series, then I would like to modify its index and clearly that's not possible to do unless another Series/Index/iterable is passed.
Ah yes, I see the idea now! I think that should work with the current implementation, let me add a test for it

Riik:
@FBruzzesi I've added a test case and also support for setting the index on a Series. It didn't work out of the box since pd.Series ([docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.index.html)) doesn't have a `set_index()` method, but you can set the index directly by assignment. 

Also, there is the extra case where passing in `column_names` in combination with a Series is now invalid. 

I feel like this function is getting a bit unwieldy, i.e. there a lot of conditionals inside the function. Would it be an option to make a separate function `maybe_set_index_series` instead? That would clean up the logic of this function and keep it simpler. Or maybe there is other ways to clean up this function a bit more. Let me know if you have any ideas!

narwhals/utils.py
```diff
@@ -297,14 +302,48 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
...
+        if isinstance(index, Series):
+            index = index.to_native()
+
+        if is_pandas_like_dataframe(native_frame):
+            return df_any._from_compliant_dataframe(  # type: ignore[no-any-return]
+                df_any._compliant_frame._from_native_frame(native_frame.set_index(index))
+            )
+
+        elif is_pandas_like_series(native_frame):
+            native_frame.index = index
```
FBruzzesi, line 342: @MarcoGorelli pros and cons of setting the index like this vs using [`Series.set_axis`](https://pandas.pydata.org/docs/reference/api/pandas.Series.set_axis.html)?

FBruzzesi:
> @FBruzzesi I've added a test case and also support for setting the index on a Series. It didn't work out of the box since pd.Series ([docs](https://pandas.pydata.org/docs/reference/api/pandas.Series.index.html)) doesn't have a `set_index()` method, but you can set the index directly by assignment.

Thanks! It looks good, I tagged Marco asking if [`Series.set_axis`](https://pandas.pydata.org/docs/reference/api/pandas.Series.set_axis.html) should be prefered instead.
 
> Also, there is the extra case where passing in `column_names` in combination with a Series is now invalid.
> 
> I feel like this function is getting a bit unwieldy, i.e. there a lot of conditionals inside the function. Would it be an option to make a separate function `maybe_set_index_series` instead? That would clean up the logic of this function and keep it simpler. Or maybe there is other ways to clean up this function a bit more. Let me know if you have any ideas!

It still seems pretty reasonable in scope and size. I will take a deeper look tomorrow to see if there is a way to cut some corners :)

FBruzzesi:
@Riik I just merged with main branch and pushed a few changes to:

- parse index and column_names into keys, and avoid having different branches after
- use `set_axis` for series (which I just noticed it makes old version fail - I will get back to `series.index = index`)

Riik:
> @Riik I just merged with main branch and pushed a few changes to parse index and column_names into keys, so that we avoid branching out for dataframe/series in multiple places.

Thank you! I think the function reads quite nicely now, with the `keys` internally it is easier to follow the flow. 

Anything else that still needs to happen on this PR?

FBruzzesi (approved):
I am (biased and) happy with the state of this! I will leave the last word to Marco for the `set_axis` vs `s.index=index` question before merging!
 
DogHighFiveGIF

narwhals/utils.py
```diff
@@ -274,13 +274,20 @@ def maybe_get_index(obj: T) -> Any | None:
...
+    df: T,
+    column_names: str | list[str] | None = None,
+    *,
+    index: Series | list[Series] | None = None,
+) -> T:
     """
-    Set columns `columns` to be the index of `df`, if `df` is pandas-like.
+    Set the index of `df`, if `df` is pandas-like. The index can either be specified as
+    a existing column name or list of column names with `column_names`, or set directly
+    with a Series or list of Series with `index`.
```
MarcoGorelli, line 286: the first line of a docstring should generally be a single sentence (separately, we should enable the ruff rule for this, D400)

narwhals/utils.py
```diff
@@ -297,15 +304,41 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
...
+            df_any._compliant_frame._from_native_frame(native_obj.set_index(keys))
+        )
+    elif is_pandas_like_series(native_obj):
+        if column_names:
+            msg = "Cannot set index using column names on a Series"
+            raise ValueError(msg)
+        native_obj.index = keys
+        return df_any._from_compliant_series(  # type: ignore[no-any-return]
+            df_any._compliant_series._from_native_series(native_obj)
         )
```
MarcoGorelli, line 339: if pandas doesn't have `Series.set_index`, should we be allowing this at all?

narwhals/utils.py
```diff
@@ -297,15 +304,41 @@ def maybe_set_index(df: T, column_names: str | list[str]) -> T:
...
-            df_any._compliant_frame._from_native_frame(
-                native_frame.set_index(column_names)
-            )
+            df_any._compliant_frame._from_native_frame(native_obj.set_index(keys))
+        )
+    elif is_pandas_like_series(native_obj):
+        if column_names:
+            msg = "Cannot set index using column names on a Series"
+            raise ValueError(msg)
+        native_obj.index = keys
```
MarcoGorelli, line 336: isn't this going to modify the original object?

narwhals/utils.py
MarcoGorelli: ah sorry, this is exactly what Francesco had requested, ignore this comment

narwhals/utils.py
```diff
@@ -274,15 +274,37 @@ def maybe_get_index(obj: T) -> Any | None:
...
+    *,
+    index: Series | list[Series] | None = None,
+) -> T:
     """
-    Set columns `columns` to be the index of `df`, if `df` is pandas-like.
+    Set the index of `df`, if `df` is pandas-like, otherwise this is a no-op.
+
+    Arguments:
+        df: object for which maybe set the index (can be either a Narwhals `DataFrame`
+            or `Series`).
```
MarcoGorelli, line 288: is it a bit odd for this to be called `df` when it can be a dataframe or series?

MarcoGorelli (commented):
thanks for updating - implementation looks good, just not sure about the argument name `df`

I did a search on https://github.com/search?q=%22maybe_set_index%28%22+language%3APython&type=code&p=1, and nobody would be affected by such a renaming. I think this is rare enough as a function that we can just change it - the only user I found it scikit-lego, who write
```
nw.maybe_set_index(result, ["fold", "part"])
```
(i.e. they pass the first argument positionally, not by name)

MarcoGorelli (approved):
thanks @Riik  and @FBruzzesi !

rik-vandervlist-sympower:
🙌 Thanks for the feedback and help, it was a fun PR to work on!

PR #1321, opened 2024-11-04
ci: trigger slow downstream libraries tests suites before release

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues 

- Related PR #1228 

## Checklist

- [ ] Tests added 
- [ ] Documented the changes

EDIT: I changed my mind 😄 I think the "slow tests" and the "publish" workflows should all run in the bump-version PR. So we can see if everything went well before merging.

The flow would be:
1. bump version
2. PR with new tag
3. Publish workflow starts along side all other workflows
4. The publish workflow triggers the "slow tests" workflow and wait for completion
5. Library published
6. merge release PR to main and celebration

~The slow tests are triggered when a tag is pushed to a branch that is _not_ `main`. Since "Bump version" creates a tag on a `bump-version` branch, this should also trigger the slow tests when the bump version PR is made. 
I also changed the "Publish Python" workflow to be triggered on tag pushes on `main`.~

~The flow would be:~
~1. bump version~
~2. PR with new tag~
~3. slow tests running along side all tests~
~4. merge PR to main~
~5. Publish workflow is triggered~
~6. Library published and celebrations~

~Currently the "publish" workflow is triggered in the "bump version" PR though. And you can see the results of the workflow before merging to `main`. Is this something that we need to keep?~

### DIFF
--- a/.github/workflows/downstream_tests_slow.yml
+++ b/.github/workflows/downstream_tests_slow.yml
@@ -1,0 +1,62 @@
+name: Test Downstream Libraries - Slow
+
+on: 
+  workflow_call:
+
+jobs:
+  vegafusion:
+     env:
+       UV_SYSTEM_PYTHON: true
+
+     strategy:
+       matrix:
+         python-version: ["3.11"]
+         os: [ubuntu-latest]
+
+     runs-on: ${{ matrix.os }}
+     steps:
+        - uses: actions/checkout@v4
+        - uses: actions/setup-python@v5
+          with:
+            python-version: ${{ matrix.python-version }}
+        - name: Install uv
+          uses: astral-sh/setup-uv@v3
+          with:
+            enable-cache: "true"
+            cache-suffix: ${{ matrix.python-version }}
+            cache-dependency-glob: "**requirements*.txt"
+        - name: clone-vegafusion
+          run: |
+              git clone --single-branch -b v2 https://github.com/vega/vegafusion.git
+              cd vegafusion
+              git log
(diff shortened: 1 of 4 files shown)


.github/workflows/downstream_tests_slow.yml
```diff
@@ -0,0 +1,62 @@
+name: Test Downstream Libraries - Slow
+
+on: 
+  workflow_call:
```
FBruzzesi, line 4: Can we have `workflow_dispatch` as well for a manual trigger, just in case?

```suggestion
  workflow_call:
  workflow_dispatch:
```

FBruzzesi (commented):
Thanks @EdAbati ! I have a tiny comment, everything else looks good to me👌

.github/workflows/downstream_tests_slow.yml
EdAbati: uh yes I forgot :) thanks

FBruzzesi (approved):
Thanks @EdAbati 

NarwhalDolphinGIF

MarcoGorelli (approved):
thanks @EdAbati and @FBruzzesi !

cool, let's see if this works - if we manage to get into Plotly then we should include that too, so it's good that we're setting this up already

PR #1333, opened 2024-11-07
feat: add `maintain_order` to Expr.unique and Series.unique 

### LINKED ISSUE #1328: enh: add `maintain_order` to Expr.unique and Series.unique
As requested here https://github.com/plotly/plotly.py/pull/4790#issuecomment-2459974589

We already have `maintain_order` for `DataFrame.unique`, my concern is that for some libraries this may turn out to not be guaranteed (and for there to not be any option to guarantee it)

However, a workaround like
```python
token = f'{s.name}_tmp'
s.to_frame().with_row_index(token).unique(col).sort(token).get_column(col)
```
should always be possible, were some library to stop guaranteeing it

It's not clear to me whether PyArrow makes any guarantees - we can always just use it for now, and if it stop making such guarantees, we use the workaround above https://github.com/apache/arrow/issues/44663

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1328

## Checklist

- [x] Tests added 
- [x] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -471,7 +471,12 @@ class PandasLikeSeries:
     def cum_sum(self) -> PandasLikeSeries:
         return self._from_native_series(self._native_series.cumsum())
 
-    def unique(self) -> PandasLikeSeries:
+    def unique(self, *, maintain_order: bool = False) -> PandasLikeSeries:
+        """
+        NOTE:
+            The param `maintain_order` is only here for compatibility with the polars API
+            and has no effect on the output.
+        """
         return self._from_native_series(
             self._native_series.__class__(
                 self._native_series.unique(), name=self._native_series.name
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -474,7 +474,7 @@ class PandasLikeSeries:
     def unique(self, *, maintain_order: bool = False) -> PandasLikeSeries:
         """
         NOTE:
-            The param `maintain_order` is only here for compatibility with the polars API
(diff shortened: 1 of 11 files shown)


narwhals/_pandas_like/series.py
```diff
@@ -471,7 +471,12 @@ def abs(self) -> PandasLikeSeries:
     def cum_sum(self) -> PandasLikeSeries:
         return self._from_native_series(self._native_series.cumsum())
 
-    def unique(self) -> PandasLikeSeries:
+    def unique(self, *, maintain_order: bool = False) -> PandasLikeSeries:
+        """
+        NOTE:
+            The param `maintain_order` is only here for compatibility with the Polars API
+            and has no effect on the output.
+        """
```
FBruzzesi, line 479: Maybe for pandas this is possible via `series.drop_duplicates(keep="first")`?

FBruzzesi (commented):
Thanks @raisadz ! I left a comment regarding pandas, which may be doable to guarantee the order to be preserved

narwhals/_pandas_like/series.py
MarcoGorelli: pandas already maintains order, so I don't think we need to change anything

FBruzzesi: Ignore me please 🙈

MarcoGorelli (approved):
thanks @raisadz (and @FBruzzesi for reviewing)!

PR #1341, opened 2024-11-09
feat: add `Series|Expr.is_finite` method

### LINKED ISSUE #1297: [Enh]: Add support for `Series|Expr.is_finite`

Noticed in the [inline comment](https://github.com/lvgig/tubular/pull/339/files#diff-47c010811d911056943033ed83c25936f69e05a7980d3b9855880ab3b9664c6fR188) in a Tubular PR

Add `is_finite` methods

_No response_

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #1297 

## Checklist

- [x] Tests added 
- [x] Documented the changes

As mentioned in the issue itself, pandas and dask treat nan's and null's as same. Actually, even worse, for non nullable backend, `np.isfinite` returns False and for nullable-backends will return `<NA>`. I made the opinionated choice to be consistent across different pandas backends and always return False for nulls and nans. I hope the warning in the docstring is enough


narwhals/_pandas_like/series.py
```diff
@@ -762,6 +762,13 @@ def mode(self: Self) -> Self:
     def __iter__(self: Self) -> Iterator[Any]:
         yield from self._native_series.__iter__()
 
+    def is_finite(self: Self) -> Self:
+        import numpy as np  # ignore-banned-import
+
+        return self._from_native_series(
+            np.isfinite(self._native_series) & ~self._native_series.isna()
+        )
```
FBruzzesi, line 770: Here is a opinionated choice that na is not finite

tests/expr_and_series/is_finite_test.py
```diff
@@ -0,0 +1,45 @@
...
+from tests.utils import ConstructorEager
+from tests.utils import assert_equal_data
+
+data = {"a": [float("nan"), float("inf"), 2.0, None]}
+
+
+@pytest.mark.filterwarnings("ignore:invalid value encountered in cast")
+def test_is_finite_expr(request: pytest.FixtureRequest, constructor: Constructor) -> None:
+    if sys.version_info < (3, 9) and "pandas_pyarrow" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
```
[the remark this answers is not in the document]
FBruzzesi: Edit: insta karma from CI 🙈

EdAbati (approved):
Thanks! Given the difference with NaN/Null across backends, your implementation looks good to me 👌👌


(_backends arguing about nulls_)

narwhals/_pandas_like/series.py
MarcoGorelli: 🤔 no sure, wouldn't we want to preserve null values?

FBruzzesi: Behavior is different for different pandas backend dtype. Let me come back with an example

MarcoGorelli: hmmm actually, for classical pandas types, we wouldn't have the option of returning a nullable boolean (if we want to preserve the dtype backend)

🤔 gonna think about this a little longer

FBruzzesi: These would be the output:

```py
data = [float("nan"), float("inf"), 2.0, None]

s = pd.Series(data)
np.isfinite(s)

0    False
1    False
2     True
3    False
dtype: bool
```

```py
np.isfinite(s.convert_dtypes(dtype_backend="numpy_nullable"))

0     <NA>
1    False
2     True
3     <NA>
dtype: boolean
```

```py
np.isfinite(s.convert_dtypes(dtype_backend="pyarrow"))

0    False
1    False
2     True
3    False
dtype: bool
```

MarcoGorelli:
just noticed that we're already somewhat inconsistent when the resulting column is boolean:
```python
In [6]: nw.from_native(pd.DataFrame(data)).select(nw.col('a')>1).to_native()
Out[6]: 
       a
0  False
1  False
2   True

In [7]: nw.from_native(pl.DataFrame(data)).select(nw.col('a')>1).to_native()
Out[7]: 
shape: (3, 1)
┌───────┐
│ a     │
│ ---   │
│ bool  │
╞═══════╡
│ false │
│ null  │
│ true  │
└───────┘
```

This may be fine, not sure there's too much we can do to work around this, but we probably just need a page alongside https://narwhals-dev.github.io/narwhals/other/column_names/ to explain what to expect from boolean columns

MarcoGorelli:
I'd put up a page about booleans - if we can agree on it, then I think it unblocks this PR https://github.com/narwhals-dev/narwhals/pull/1392

narwhals/_pandas_like/series.py
FBruzzesi: @MarcoGorelli considering the new page we have on booleans, how would you move forward with this?
I am asking just to discuss it, I am ok with keeping the inconsistencies between different pandas dtype backends, and let the use handle those. At the same time, I would aim at some sort of unification towards polars behavior, although in this specific context it seems unfeasible

FBruzzesi: Looks good, I will resolve conflicts with main and adjust

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1342, opened 2024-11-09
feat: add `Series|Expr.rank`

### LINKED ISSUE #1323: [Enh]: Support polars.Expr.rank

I am abstracting a library for computing teaching metrics so that researchers can use their data processing library of choice. Narwhals seems like a good bet (also shout-out to @mikeckennedy for having you on the podcast!). I can't share the specific repository because it contains internal scripts, but this would be supporting [CourseKata](coursekata.org), a low-cost textbook platform dedicated to continuous improvement based on learning science principles.

I would like support for the [`polars.Expr.rank`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.rank.html#polars.Expr.rank) method. One example of how it could be used is to count how often an instructor teaches, given some grouping variable (window). In Polars it might look like this:

```python
df.sort("academic_year").with_columns(
  years_taught=pl.col("academic_year")
    .rank(method="dense")
    .over("instructor_id")
)
```

This would window over `instructor_id` and get the rank by `academic_year`. Essentially, we will get a count of how many academic years an instructor has taught in, and because we are using the "dense" ranking, teaching multiple classes in a year counts as a single year taught.

_No response_

I could probably achieve this by making an intermediate data frame where I filter down `academic_year` using `unique()`, and then make some kind of counter variable based on `instructor_id`, and then `join()` that back to the initial table.

Instead I would rather just go back to using Polars until this feature is supported (if it is on your roadmap!).

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #1323 

## Checklist

- [x] Tests added 
- [x] Documented the changes

This PR provides initial support for rank method. I will start it as a draft due to a bunch of shortcomings:

- **pandas**:
  - there is a (nasty) trick to make it work with nulls/nans and nullable dtypes (related: to [BUG: rank does not respect na_option='keep' for numpy nullable integer dtypes](https://github.com/pandas-dev/pandas/issues/56976), which should be fixed in pandas 3) - I will comment the trick in the code
  - support in `over` context is available
  - support in `group_by` context is not available.
- **pyarrow**:
  - does not support polars default method (namely, "average"), therefore if rank is called without specifying another method, it will end up raising an error
  - does not implement rank in a group by context
  - I am using `pyarrow.compute.rank` which is available but not exposed/documented (?)
- **dask**: it just does not support ranking
- **polars**: 
  - group by context _always_ returns an aggregate, in this case a list of ranks - which is fairly useless as it is a list of increasing/descreasing range until the size of the group


narwhals/_pandas_like/series.py
```diff
@@ -719,6 +719,56 @@ def mode(self: Self) -> Self:
...
+                self._dtypes.UInt64,
+                self._dtypes.UInt32,
+                self._dtypes.UInt16,
+                self._dtypes.UInt8,
+            }
+            and (null_mask := native_series.isna()).any()
+        ):
+            # crazy workaround for the case of `na_option="keep"` and nullable
+            # integer dtypes. This should be supported in pandas > 3.0
+            # https://github.com/pandas-dev/pandas/issues/56976
```
FBruzzesi, line 749: Here is the workaround.

@MarcoGorelli I was not able to properly use the pandas like util function `get_dtype_backend` to figure out the nullable backend. It should not really matter as the non-nullable backend would not result in integer type if the series contains nulls anyway

tests/expr_and_series/rank_test.py
```diff
@@ -0,0 +1,103 @@
...
+        assert_equal_data(result, expected_data)
+
+
+@pytest.mark.parametrize("method", rank_methods)
+def test_rank_expr_in_over_context(
+    request: pytest.FixtureRequest,
+    constructor: Constructor,
+    method: Literal["average", "min", "max", "dense", "ordinal"],
+) -> None:
+    if "polars" not in str(constructor):
```
FBruzzesi, line 93: Well 🥲

FBruzzesi:
Hey @adamblake, this is an initial implementation to support `rank`. In the description I tried to explain all the shortcomings and the challenges I am facing.

FBruzzesi:
I will need to come back to the failing tests: it seems that rank for pandas version 2.0.x with pyarrow backend fails for integer dtype but it works fine for floats

FBruzzesi:
> Nice, thanks for sticking with this!

Of course! Adding the possibility to using it in `over` context is a great addition for the use case reported in the issue in the first place!

> Does it risk being annoying if the default ('average') isn't available in several backends?

For now this is just pyarrow, right?
 
> Could we not set any default, so the user is 'forced' to specify a method (like we do in `quantile`)?

I don't have a strong opinion on this, I am fine to give more responsibility to the user

MarcoGorelli:
> For now this is just pyarrow, right?

you're right sorry, i thought it would be the same for pyarrow-backed pandas. thinking ahead, I think duckdb doesn't support it either

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1345, opened 2024-11-10
chroe: refactor `ArrowDataFrame.with_columns`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

While profiling for plotly, py-spy indicates that we spend a lot of time in `validate_dataframe_comparand` for pyarrow case. This is called only in `with_columns` methods.

This PR proposed two changes:

- Creates constant array via `np.full` in `validate_dataframe_comparand` instead of `[const] * length`
- Changes the logic in `with_columns` to use pyarrow native methods to insert a column value. I expect this to be faster than the current approach of concatenating the already existing columns with the new ones - caveat is if the number of new columns is order(s) of magnitude greater than the existing ones. In _majority_ of scenarios I would expect this to not be the case, but this is the reason I am opening this as **RFC**

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -305,35 +305,30 @@ class ArrowDataFrame:
         *exprs: IntoArrowExpr,
         **named_exprs: IntoArrowExpr,
     ) -> Self:
+        native_frame = self._native_frame
         new_columns = evaluate_into_exprs(self, *exprs, **named_exprs)
-        new_column_name_to_new_column_map = {s.name: s for s in new_columns}
-        to_concat = []
-        output_names = []
-        # Make sure to preserve column order
+
         length = len(self)
-        for name in self.columns:
-            if name in new_column_name_to_new_column_map:
-                to_concat.append(
-                    validate_dataframe_comparand(
-                        length=length,
-                        other=new_column_name_to_new_column_map.pop(name),
-                        backend_version=self._backend_version,
-                    )
-                )
-            else:
-                to_concat.append(self._native_frame[name])
-            output_names.append(name)
-        for s in new_column_name_to_new_column_map:
-            to_concat.append(
-                validate_dataframe_comparand(
-                    length=length,
-                    other=new_column_name_to_new_column_map[s],
-                    backend_version=self._backend_version,
-                )
+        columns = self.columns
+
+        for col_value in new_columns:
+            col_name = col_value.name
+
+            column = validate_dataframe_comparand(
+                length=length,
+                other=col_value,
+                backend_version=self._backend_version,
             )
-            output_names.append(s)
-        df = self._native_frame.__class__.from_arrays(to_concat, names=output_names)
-        return self._from_native_frame(df)
+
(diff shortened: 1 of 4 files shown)


narwhals/_arrow/dataframe.py
```diff
@@ -305,35 +305,30 @@ def with_columns(
...
+                backend_version=self._backend_version,
+            )
+
+            native_frame = (
+                native_frame.set_column(
+                    columns.index(col_name), field_=col_name, column=column
                 )
+                if col_name in columns
+                else native_frame.append_column(field_=col_name, column=column)
             )
```
FBruzzesi, line 329: If column name exists, then we replace the column, else we append at the end

narwhals/_arrow/utils.py
```diff
@@ -174,20 +174,24 @@ def validate_dataframe_comparand(
...
 
-    if isinstance(other, ArrowDataFrame):
-        return NotImplemented
     if isinstance(other, ArrowSeries):
         if len(other) == 1:
+            import numpy as np  # ignore-banned-import
             import pyarrow as pa  # ignore-banned-import
 
-            value = other.item()
+            value = other[0]
```
FBruzzesi, line 184: Avoids `item` to check again the length of the array, we already know it is 1

MarcoGorelli:
nice - there may be some issue on py38, but the rest looks good!

should we wait until altair / marimo's cis are fixed to merge?

FBruzzesi:
Fixed old versions, what's the deal with TPCH taking so long now 🙈?

MarcoGorelli:
🤔 looks like yesterday it went from 2 minutes to 15 minutes

MarcoGorelli:
ok it was dask, i've removed them from the tpch ci and opened an issue on their repo

MarcoGorelli (approved):
really nice, thanks @FBruzzesi  - did you test this against the plotly branch locally?

FBruzzesi:
> did you test this against the plotly branch locally?

Yes, not big of a change as I would expect though - e.g. I would expect that `np.full(1_000_000, 1)` to be visibly faster than `[1] * 1_000_000`.

MarcoGorelli:
yeah where there's `pip install git+https://github.com/narwhals-dev/narwhals` change that to for example `pip install git+https://github.com/narwhals-dev/narwhals@perf/pyarrow-with-columns`

FBruzzesi:
@MarcoGorelli for 1M rows, 50 columns I cannot see any changes in performance for:

- only editing existing columns
- only adding columns
- editing and adding columns in the same `with_columns` statement

both when working with chunked array and scalars. The two approaches seem to be equivalent.
I leave it up to you if this syntax is any better than the previous one

MarcoGorelli:
thanks for checking! i think I prefer this one, if you agree let's ship it

PR #1348, opened 2024-11-10
ci: remove dask from tpch tests

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Related issue # 
- Closes #

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added 
- [ ] Documented the changes

PR #1374, opened 2024-11-13
chore: cleanup tpch execute

EdAbati:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added 
- [x] Documented the changes

Inspired by #1348, I thought it would be handy if adding/removing backends from TPCH was a bit easier. With this change, adding/removing a key in the `BACKEND_READ_FUNC_MAP` dictionary would add/remove a backend from the test.

While refactoring this I found some duplicated (a path was present twice) and unused code (the read functions that I commented out). I hope that with this it would be easier to spot

The diff is big because I removed all the `execute` folder. I know that this is lower priority so don't feel like you have to review straight away, it can wait :)


tpch/execute.py
```diff
@@ -0,0 +1,109 @@
...
+REGION_PATH = DATA_DIR / "region.parquet"
+NATION_PATH = DATA_DIR / "nation.parquet"
+SUPPLIER_PATH = DATA_DIR / "supplier.parquet"
+PART_PATH = DATA_DIR / "part.parquet"
+PARTSUPP_PATH = DATA_DIR / "partsupp.parquet"
+ORDERS_PATH = DATA_DIR / "orders.parquet"
+CUSTOMER_PATH = DATA_DIR / "customer.parquet"
+
+BACKEND_READ_FUNC_MAP = {
+    # "pandas": lambda x: pd.read_parquet(x, engine="pyarrow"), # noqa: ERA001
```
EdAbati, line 25: should we just remove these? I kept them because they were still in the dictionary

tpch/tests/queries_test.py
```diff
@@ -4,20 +4,22 @@
...
-            [sys.executable, "-m", f"execute.{script_path.stem}"],
-            capture_output=True,
-            text=True,
-            check=False,
-        )
-        assert (
-            result.returncode == 0
-        ), f"Script {script_path} failed with error: {result.stderr}"
+
+@pytest.mark.parametrize("query_path", QUERIES_DIR.glob("q[1-9]*.py"))
```
EdAbati, line 14: if this PR doesn't get merged, I think that at least we should have this parametrised :)

tpch/execute.py
```diff
@@ -0,0 +1,109 @@
...
+    "pandas[pyarrow]": lambda x: pd.read_parquet(
+        x, engine="pyarrow", dtype_backend="pyarrow"
+    ),
+    # "polars[eager]": lambda x: pl.read_parquet(x),
+    "polars[lazy]": lambda x: pl.scan_parquet(x),
+    "pyarrow": lambda x: pq.read_table(x),
+    # "dask": lambda x: dd.read_parquet(x, engine="pyarrow", dtype_backend="pyarrow"), # noqa: ERA001
```
FBruzzesi, line 32: Should we enable dask again? They fixed the issue for which we removed it in the first place.

tpch/tests/queries_test.py
FBruzzesi: +1

FBruzzesi (commented):
I love this kind of refactors and simplifications! Thanks @EdAbati !

I left just one comment regarding enabling running dask since they fixed the issue, and from our side we adjusted given their suggestions

tpch/execute.py
EdAbati: you are right. tested locally, it seems to be fine 👀

FBruzzesi (approved):
It's a +10 from me 

DogsClapGIF

MarcoGorelli (approved):
awesome, much better, thanks both!

Issue #1371, opened 2024-11-13
[Enh]: cumulative features


FBruzzesi:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

Cumulative features, together with forward fill and some other checks/hacks, would most likely be enough to enable the equivalent of pandas expanding operations. This is a requirement to complete https://github.com/plotly/plotly.py/issues/4834.

### Please describe the purpose of the new feature or describe the problem to solve.

List of cumulative expressions supported by polars:

- [x] [cum_count](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cum_count.html#polars.Expr.cum_count)
- [x] [cum_min](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cum_min.html#polars.Expr.cum_min)
- [x] [cum_max](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cum_min.html#polars.Expr.cum_max)
- [x] [cum_prod](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cum_prod.html#polars.Expr.cum_prod)
- [x] [cum_sum](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.cum_sum.html#polars.Expr.cum_sum)

With these, we would enable the following additional univariate expanding operations: mean, var, std, skew, kurt.

What's left out is: `median`, `quantile` and `rank` - I don't think we would be able to implement those 🥲 (entire pandas [expanding window function list](https://pandas.pydata.org/docs/reference/window.html#expanding-window-functions)).

## Group by context

Edit: Additionally, we should support these expr in group by's context. This is partially possible:
- for pandas, we can use native methods, only with `reverse=False` (default argument), for `cum_<min|max|sum|prod>` (need to check how [DataFrameGroupBy.cumcount](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.cumcount.html) behaves with nulls.
- for pyarrow, the only way we can enable these are via `__iter__`, which I cannot say how slower it is than native methods

For the moment I would keep these out of the PRs introducing the methods in the first place. Thanks @AlessandroMiola to point that out in #1384

FBruzzesi:
[...]

PR #1380, opened 2024-11-14
feat: add `Series|Expr.cum_count` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1371

## Checklist

- [x] Tests added
- [x] Documented the changes

I have dask skill issues. Replicating what works for pandas raises an error:

>     def _divisions(self):
>        # This is an issue.  In normal Dask we re-divide everything in a step
>        # which combines divisions and graph.
>        # We either have to create a new Align layer (ok) or combine divisions
>        # and graph into a single operation.
>        dependencies = self.dependencies()
>        for arg in dependencies:
>            if not self._broadcast_dep(arg):
>               assert arg.divisions == dependencies[0].divisions
> E               AssertionError

Code:

```py
def cum_count(self, reverse):
    def func(_input: dask_expr.Series, _reverse: bool) -> dask_expr.Series:
        not_na_series = (~_input.isna()).astype(int)
        return (
            not_na_series.cumsum()
            if not _reverse
            else _input.size - not_na_series.cumsum() + not_na_series - 1
       )

return self._from_call(
    func,
    "cum_count",
    reverse,
    returns_scalar=False,
)
```

### DIFF
--- a/docs/api-reference/expr.md
+++ b/docs/api-reference/expr.md
@@ -11,6 +11,7 @@
         - arg_true
         - cast
         - count
+        - cum_count
         - cum_sum
         - diff
         - drop_nulls
--- a/docs/api-reference/series.md
+++ b/docs/api-reference/series.md
@@ -15,6 +15,7 @@
         - cast
         - clip
         - count
+        - cum_count
         - cum_sum
         - diff
         - drop_nulls
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -430,6 +430,9 @@ class ArrowExpr:
             dtypes=self._dtypes,
         )
 
+    def cum_count(self: Self, *, reverse: bool) -> Self:
+        return reuse_series_implementation(self, "cum_count", reverse=reverse)
+
     @property
     def dt(self: Self) -> ArrowExprDateTimeNamespace:
         return ArrowExprDateTimeNamespace(self)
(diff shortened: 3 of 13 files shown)


FBruzzesi:
Betrayed by old polars version itself

MarcoGorelli:
yeah it switched from starting at 0 to starting at 1 at some point...tbh i think it should've been a minor version change rather than a patch one

MarcoGorelli (approved):
wonderful, thanks @FBruzzesi !

agree on keeping Dask out of this for now, as we need to exclude row-order-dependent operations for lazy-only backends

FBruzzesi:
> yeah it switched from starting at 0 to starting at 1 at some point...tbh i think it should've been a minor version change rather than a patch one

Should we check and branch out for other minor polars versions for this?

FBruzzesi:
> agree on keeping Dask out of this for now, as we need to exclude row-order-dependent operations for lazy-only backends

Happy you agree! My issue was `reverse=True` was failing with the error in the description

FBruzzesi:
> > Should we check and branch out for other minor polars versions for this?
> 
> not sure i understand the question / issue, sorry, it looks like your PR already handles the different Polars versions perfectly?

So in polars==0.20.3, `cum_count` does not exist, hence my "workaround" implementation.
However you mentioned that possibly when introduced the count was starting from 0. Since we don't have tests for _every single version_, I am not sure we cover such case.

MarcoGorelli:
ah it changed in 0.20.4, so you already covered it 🙌

PR #1412, opened 2024-11-20
feat: make `Series` generic

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

Last week, I was playing with the idea of having `Series` generic as `DataFrame`s for better typing.

It may be useful since it is related to the current issues/PRs around documentation/typing.


MarcoGorelli (commented):
thanks for doing this!

i think in stable/v1 you may need to do something like
```
from narwhals.typing import Series as NwSeries
Series = NwSeries[Any]
```
to keep it stable for shiny?

EdAbati:
Thanks @MarcoGorelli for the review ✌️

shiny is happy now but `stable.v1.Series` doesn't have the generic typing 😕 If you know any smart way to renable some parts of it let me know.

(the CI error seems unrelated)

narwhals/stable/v1/__init__.py
```diff
@@ -958,15 +958,19 @@ def _stableify(obj: NwDataFrame[IntoFrameT]) -> DataFrame[IntoFrameT]: ...
...
 
 
 def _stableify(
-    obj: NwDataFrame[IntoFrameT] | NwLazyFrame[IntoFrameT] | NwSeries | NwExpr | Any,
+    obj: NwDataFrame[IntoFrameT]
+    | NwLazyFrame[IntoFrameT]
+    | NwSeries[IntoSeriesT]
+    | NwExpr
+    | Any,
 ) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT] | Series | Expr | Any:
```
MarcoGorelli, line 974: i'm a little surprised that mypy allows this, i.e. using a typevar (IntoSeriesT) only in the arguments but not in the return value

If we don't have generic Series in v1, should we make it `NwSeries[Any]`?

MarcoGorelli (commented):
Thanks, nice one! just got a question

Looking over https://narwhals-dev.github.io/narwhals/backcompat/#exceptions, we did write

> we may consider making some type hints more precise.

I feel like we should loop @schloerke into the conversation - would you be OK with us making this type hint more precise? It would only affect Shiny's internal typing and so wouldn't be user-facing, and we would also make the PR to Shiny to get your CI green again

If not, no worries, the stable.v1 mechanism does allow us to only make this change in the main namespace

MarcoGorelli:
i just tried this at it looks like it would break Altair's typing too, so I reverted the extra commit

maybe it's OK to just introduce this in the main namespace and leave `stable.v1` as it is for now

EdAbati:
Thanks both! 
I agree that if it start breaking more than 1 downstream library test, we shouldn't change `stable.v1` for now. :)

MarcoGorelli (approved):
thanks @EdAbati !

been wanting this for a while, just never got round to doing it

I've added a note to `backcompat.md` and have used the typevar in the return values in `to_native`

PR #1421, opened 2024-11-22
docs: explain optional boolean args in from_native

### LINKED ISSUE #1420: [Doc]: Clarify behavior of boolean arguments to `from_native()`

Improvement

While consulting the docs today I noticed that the behavior of the boolean arguments to `from_native()` is not fully explained, particularly since `None` is one of the allowed values, and most of the arguments default to `None`, so it's unclear what is the default behavior./

The explanation for `pass_through` is nearly there, since it clearly explains the behavior in the `True` and `False` cases, but it doesn't explain the behavior in the `None` case. (I notice that the explanation says that `True` is the default even though the value listed under the "Default" column is `None` -- maybe the values in the "Default" column are just incorrect?)

For each boolean argument, docs should explain:

- Behavior if argument is `True`
- Behavior if argument is `False`
- Behavior if argument is `None`

### Screenshot of relevant portion of docs

Screen Shot 2024-11-21 at 6 11 21 PM

MarcoGorelli:
closes #1420

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/docs/requirements-docs.txt
+++ b/docs/requirements-docs.txt
@@ -1,6 +1,8 @@
 black  # required by mkdocstrings_handlers
 jinja2
 duckdb
+# Hopefully temporary until https://github.com/mkdocstrings/mkdocstrings/issues/716
+git+https://github.com/MarcoGorelli/griffe.git@no-overloads
 markdown-exec[ansi]
 mkdocs
 mkdocs-autorefs
--- a/.github/workflows/check_docs_build.yml
+++ b/.github/workflows/check_docs_build.yml
@@ -31,7 +31,9 @@ jobs:
       - name: local-install
         run: uv pip install -e . --system
       - name: check-no-errors
-        run: python -m mkdocs build > output.txt 2>&1
+        run: |
+          python -m mkdocs build > output.txt 2>&1
+          cat output.txt
       - name: assert-no-errors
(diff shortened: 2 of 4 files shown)


FBruzzesi:
Is there a reason to have `None`s in the first place? From a user point of view, I think it would be more clear to have boolean defaults 🤔

docs/requirements-docs.txt
```diff
@@ -1,6 +1,8 @@
 black  # required by mkdocstrings_handlers
 jinja2
 duckdb
+# Hopefully temporary until https://github.com/mkdocstrings/mkdocstrings/issues/716
+git+https://github.com/MarcoGorelli/griffe.git@no-overloads
```
MarcoGorelli, line 5: temporary (hopefully 🤞 ) workaround for https://github.com/mkdocstrings/mkdocstrings/issues/716

MarcoGorelli:
thanks for your review, and thanks @emilykl for helping us improve the docs, really appreciate this and the other suggestions you made!

PR #1451, opened 2024-11-27
feat: add `Series|Expr.rolling_var` and `Series|Expr.rolling_std`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1254 

## Checklist

- [x] Tests added
- [x] Documented the changes

Opening as draft, I still need to figure out how to deal with null/nan's to match polars behavior.

Edit: Scary size diff, yet 25%+ is just the copying of methods to the stable api, and another good 10% are docstrings examples


narwhals/_arrow/utils.py
```diff
@@ -446,3 +446,39 @@ def _parse_time_format(arr: pa.Array) -> str:
         if pc.all(matches.is_valid()).as_py():
             return time_fmt
     return ""
+
+
+def pad_series(
```
FBruzzesi, line 451: Possibly a better name to be found

FBruzzesi:
Checking performances: on 1M floats, pyarrow is ~2x slower than pandas 🤔

tests/expr_and_series/rolling_var_test.py
```diff
@@ -0,0 +1,123 @@
...
+    center=st.booleans(),
+    values=st.lists(st.floats(-10, 10), min_size=5, max_size=10),
+)
+@pytest.mark.skipif(PANDAS_VERSION < (1,), reason="too old for pyarrow")
+@pytest.mark.filterwarnings("ignore:.*:narwhals.exceptions.NarwhalsUnstableWarning")
+def test_rolling_var_hypothesis(center: bool, values: list[float]) -> None:  # noqa: FBT001
+    s = pd.Series(values)
+    window_size = random.randint(2, len(s))  # noqa: S311
+    min_periods = random.randint(2, window_size)  # noqa: S311
+    ddof = random.randint(0, min_periods - 1)  # noqa: S311
```
FBruzzesi, line 98: Ok, it took me a bit, but I figured out that the different behavior appear when `ddof > min_periods` because the denominator can be negative or zero.

Pandas seems to return NaN indistinctly for windows that do not have enough non null elements and these cases, while polars returns inf (and same in pyarrow since we can control that)

narwhals/_arrow/series.py
```diff
@@ -997,9 +966,64 @@ def rolling_mean(
...
+        valid_count = padded_series.cum_count(reverse=False)
+        count_in_window = valid_count - valid_count.shift(window_size).fill_null(0)
+
+        result: Self = self._from_native_series(
+            pc.if_else(
+                (count_in_window >= min_periods)._native_series,
+                (rolling_sum_sq - (rolling_sum**2 / count_in_window))._native_series,
+                None,
+            )
+        ) / max(count_in_window - ddof, 0)  # type: ignore[call-overload]
```
FBruzzesi, line 1009: mypy doesn't really like `max(count_in_window - ddof, 0)` because of 

> No overload variant of "max" matches argument types "ArrowSeries", "int"  [call-overload]

narwhals/_arrow/series.py
```diff
@@ -997,9 +966,64 @@ def rolling_mean(
...
+        rolling_sum_sq = (
+            cum_sum_sq - cum_sum_sq.shift(window_size).fill_null(0)
+            if window_size != 0
+            else cum_sum_sq
+        )
+
+        valid_count = padded_series.cum_count(reverse=False)
+        count_in_window = valid_count - valid_count.shift(window_size).fill_null(0)
+
+        result: Self = self._from_native_series(
```
FBruzzesi, line 1003: again, mypy doesn't really know what happens after the division, so we annotate this, otherwise it yells with:

>  Returning Any from function declared to return "Self"  [no-any-return]

narwhals/expr.py
```diff
@@ -3377,6 +3377,193 @@ def rolling_mean(
...
+            ┌──────┬──────────┐
+            │ a    ┆ b        │
+            │ ---  ┆ ---      │
+            │ f64  ┆ f64      │
+            ╞══════╪══════════╡
+            │ 1.0  ┆ 0.0      │
+            │ 2.0  ┆ 0.707107 │
+            │ null ┆ 0.707107 │
+            │ 4.0  ┆ 1.414214 │
+            └──────┴──────────┘
```
MarcoGorelli, line 3544: 🤔 is this  a bug in Polars?

```python
In [37]: pl.Series([1,2,None,4]).rolling_std(3, min_periods=1)
Out[37]:
shape: (4,)
Series: '' [f64]
[
        0.0
        0.707107
        0.707107
        1.414214
]

In [38]: pl.Series([1,2,1,4]).rolling_std(3, min_periods=1)
Out[38]:
shape: (4,)
Series: '' [f64]
[
        null
        0.707107
        0.57735
        1.527525
]
```

I think the first element of the result should be null, right?

MarcoGorelli: on it https://github.com/pola-rs/polars/pull/20077

MarcoGorelli: merged, so once there's a new Polars release we can update 🚀

FBruzzesi: Thanks Marco! That' awesome! How should we behave for older versions though? One possibility is failing is nulls are present, but I have mixed feeling about it

PR #1456, opened 2024-11-28
feat: create dataframe from 2D numpy array and column names

### LINKED ISSUE #1434: enh: create dataframe from 2D numpy array and column names
For classical numpy dtypes in pandas, this should be more efficient than forcing slices + dicts

Spotted in Prophet

Desired syntax:
```python
import narwhals as nw
import numpy as np
import pandas as pd

arr = np.zeros([[1,2,3], [4,5,6]])
df = nw.from_2d_array(arr, column_names=['a', 'b', 'c'])
```
It should also be possible to pass in `dtype` as an argument - if not passed, it'll be autoinferred
```python
import narwhals as nw
import numpy as np
import pandas as pd

arr = np.zeros([[1,2,3], [4,5,6]])
df = nw.from_2d_array(arr, column_names=['a', 'b', 'c'], dtype=nw.Float64)
```

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1434

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -465,7 +465,7 @@ def _from_dict_impl(
         try:
             # implementation is UNKNOWN, Narwhals extension using this feature should
             # implement `from_dict` function in the top-level namespace.
-            native_frame = native_namespace.from_dict(data)
+            native_frame = native_namespace.from_dict(data, schema=schema)
         except AttributeError as e:
             msg = "Unknown namespace is expected to implement `from_dict` function."
             raise AttributeError(msg) from e
@@ -481,7 +481,7 @@ def from_numpy(
     """Construct a DataFrame from a NumPy ndarray.
 
     Notes:
-        Only the row orientation is currently supported.
+        Only row orientation is currently supported.
 
         For pandas-like dataframes, conversion to schema is applied after dataframe
         creation.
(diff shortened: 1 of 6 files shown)


narwhals/functions.py
```diff
@@ -430,6 +432,163 @@ def _from_dict_impl(
...
+
+    Notes:
+        For pandas-like dataframes, conversion to schema is applied after dataframe
+        creation.
+
+    Arguments:
+        data: Two-dimensional data represented as a NumPy ndarray.
+        schema: The DataFrame schema as Schema or dict of {name: type}.
+        native_namespace: The native library to use for DataFrame creation. Only
+            necessary if inputs are not Narwhals Series.
```
FBruzzesi, line 451:
```suggestion
        native_namespace: The native library to use for DataFrame creation.
```

narwhals/functions.py
```diff
@@ -430,6 +432,163 @@ def _from_dict_impl(
     return from_native(native_frame, eager_only=True)
 
 
+def from_numpy(
+    data: np.ndarray,
+    schema: dict[str, DType] | Schema | None = None,
```
FBruzzesi, line 437: I think it would be nice to support `list[str]` as column names. In my opinion is way more common to specify the names of the columns without their dtypes when creating a dataframe

narwhals/functions.py
```diff
@@ -430,6 +432,163 @@ def _from_dict_impl(
...
+                narwhals_to_native_dtype as polars_narwhals_to_native_dtype,
+            )
+
+            schema_pl = {
+                name: polars_narwhals_to_native_dtype(dtype, dtypes=dtypes)
+                for name, dtype in schema.items()
+            }
+            native_frame = native_namespace.from_numpy(data, schema=schema_pl)
+        else:
+            native_frame = native_namespace.from_numpy(data, orient="col")
```
FBruzzesi, line 536: Why are we defaulting to `orient="col"`? To me this seems the opposite of what polars does if nothing else is provided:

```py
import numpy as np
import polars as pl

data = np.array([[5, 2, 1], [1, 4, 3]])
data
array([[5, 2, 1],
       [1, 4, 3]])

pl.from_numpy(data)
shape: (2, 3)
┌──────────┬──────────┬──────────┐
│ column_0 ┆ column_1 ┆ column_2 │
│ ---      ┆ ---      ┆ ---      │
│ i64      ┆ i64      ┆ i64      │
╞══════════╪══════════╪══════════╡
│ 5        ┆ 2        ┆ 1        │
│ 1        ┆ 4        ┆ 3        │
└──────────┴──────────┴──────────┘
```

while:

```py
pl.from_numpy(data, orient="col")
shape: (3, 2)
┌──────────┬──────────┐
│ column_0 ┆ column_1 │
│ ---      ┆ ---      │
│ i64      ┆ i64      │
╞══════════╪══════════╡
│ 5        ┆ 1        │
│ 2        ┆ 4        │
│ 1        ┆ 3        │
└──────────┴──────────┘
```

This translates to the array transpose in pandas

FBruzzesi (commented):
Hey @raisadz thanks for the PR! I hope it's ok if I am reviewing even if it's in draft 😇

I left a couple of comments, but the main concern is the `orient` behavior. Apologies if you were going to address these 🙈

narwhals/functions.py
raisadz: Hey @FBruzzesi, thanks for your review! I fixed the orientation from columns to rows

narwhals/functions.py
raisadz: Agreed, I added this as an option for `schema`

MarcoGorelli (approved):
thanks @raisadz !

MarcoGorelli:
and thanks @FBruzzesi for reviewing!

Issue #1443, opened 2024-11-25
Question: idiomatic way of elegantly retrieving the underlying DataFrame type


elephaint:
Currently I often have the following code:

```
df_nw = nw.from_native(df)
is_pandas = nw.dependencies.is_pandas_dataframe(df)
.....
if is_pandas:
     do something different
......

```

What is difficult about this, is that I need to keep track of `is_pandas` variables throughout the code, send them in subfunctions, etc. If I have multiple DataFrames, I have multiple such `is_pandas` variables. Ideally, I'd be able to do something such as:

```
is_pandas = df_nw.is_native_pandas
```
i.e., having whether the underlying dataframe is pandas or not simply as a boolean attribute of the Narwhals DataFrame. That would allow me to use `df_nw` everywhere without requiring the auxiliary variables everywhere or first converting to native.

Of course, I know I can also do this everywhere: `nw.dependencies.is_pandas_dataframe(df_nw.to_native()))` but that feels convoluted.

What is the cleanest way to do this?

FBruzzesi:
Hey @elephaint , thanks for your request. This can certainly be a pain point for other libraries trying to adopt narwhals.

I would say that the answer is _it depends_.

We have a set of functionalities, namely `maybe_align_index`, `maybe_get_index` , `maybe_set_index`, `maybe_reset_index`  and `maybe_convert_dtypes`, which are meant to help working with pandas objects without having to manually check all the times.

If that's not enough, `nw.dependencies.is_pandas_dataframe(df_nw.to_native()))` is an option, another one would be `df_nw._compliant_frame._implementation is Implementation.PANDAS` but it doesn't look less convoluted to me.

In [plotly express](https://github.com/plotly/plotly.py/blob/master/packages/python/plotly/plotly/express/_core.py#L1487), I had to do something similar, by adding a flag `is_pd_like` on the first encounter of the dataframe object, and passing that to various functions to branch out the logic.

MarcoGorelli:
Thanks for the request!

I think currently the two documented way would be:
- `if nw.get_native_namespace(df) is nw.dependencies.get_pandas()`
- `if nw.dependencies.is_pandas_dataframe(df.to_native())`

I can see that it would be convenient to have something more ergonomic... 🤔 will think about this one. Thanks for having highlighted this

> another one would be df_nw._compliant_frame._implementation is Implementation.PANDAS but it doesn't look less convoluted to me.

[...]

PR #1480, opened 2024-12-01
feat: add `Series.__contains__`

### LINKED ISSUE #1469: enh: implement `__contains__` for Series
As brought up in https://github.com/narwhals-dev/narwhals/issues/1443#issuecomment-2508932494

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1469 

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -31,22 +31,22 @@ from narwhals.dtypes import UInt32
 from narwhals.dtypes import UInt64
 from narwhals.dtypes import Unknown
 from narwhals.expr import Expr
-from narwhals.expr import all_ as all  # noqa: A004
+from narwhals.expr import all_ as all
 from narwhals.expr import all_horizontal
 from narwhals.expr import any_horizontal
 from narwhals.expr import col
 from narwhals.expr import concat_str
-from narwhals.expr import len_ as len  # noqa: A004
+from narwhals.expr import len_ as len
 from narwhals.expr import lit
-from narwhals.expr import max  # noqa: A004
+from narwhals.expr import max
 from narwhals.expr import max_horizontal
 from narwhals.expr import mean
 from narwhals.expr import mean_horizontal
 from narwhals.expr import median
-from narwhals.expr import min  # noqa: A004
+from narwhals.expr import min
 from narwhals.expr import min_horizontal
 from narwhals.expr import nth
-from narwhals.expr import sum  # noqa: A004
+from narwhals.expr import sum
 from narwhals.expr import sum_horizontal
 from narwhals.expr import when
 from narwhals.functions import concat
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1022,14 +1022,23 @@ class ArrowSeries:
         )
 
     def __contains__(self: Self, other: Any) -> bool:
-        import pyarrow as pa  # ignore-banned-imports
-        import pyarrow.compute as pc  # ignore-banned-imports
+        from pyarrow import ArrowNotImplementedError  # ignore-banned-imports
+        from pyarrow import ArrowTypeError  # ignore-banned-imports
 
-        native_series = self._native_series
-        return maybe_extract_py_scalar(  # type: ignore[no-any-return]
-            pc.is_in(pa.scalar(other, type=native_series.type), native_series),
-            return_py_scalar=True,
-        )
+        try:
+            import pyarrow as pa  # ignore-banned-imports
+            import pyarrow.compute as pc  # ignore-banned-imports
+
+            native_series = self._native_series
+            return maybe_extract_py_scalar(  # type: ignore[no-any-return]
+                pc.is_in(pa.scalar(other), native_series),
+                return_py_scalar=True,
+            )
+        except (ArrowNotImplementedError, ArrowTypeError) as exc:
+            from narwhals.exceptions import InvalidOperationError
+
+            msg = f"Unable to compare other of type {type(other)} with series of type {self.dtype}."
+            raise InvalidOperationError(msg) from exc
 
     @property
     def shape(self: Self) -> tuple[int]:
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1030,8 +1030,13 @@ class ArrowSeries:
             import pyarrow.compute as pc  # ignore-banned-imports
 
             native_series = self._native_series
+            other_ = (
+                pa.scalar(other)
+                if other is not None
+                else pa.scalar(None, type=native_series.type)
+            )
             return maybe_extract_py_scalar(  # type: ignore[no-any-return]
-                pc.is_in(pa.scalar(other), native_series),
+                pc.is_in(other_, native_series),
                 return_py_scalar=True,
             )
         except (ArrowNotImplementedError, ArrowTypeError) as exc:
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1017,6 +1017,13 @@ class ArrowSeries:
             for x in self._native_series.__iter__()
         )
 
+    def __contains__(self: Self, other: Any) -> bool:
+        import pyarrow.compute as pc  # ignore-banned-imports
+
+        return maybe_extract_py_scalar(  # type: ignore[no-any-return]
+            pc.is_in(other, self._native_series), return_py_scalar=True
+        )
+
     @property
     def shape(self: Self) -> tuple[int]:
         return (len(self._native_series),)
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1022,9 +1022,8 @@ class ArrowSeries:
         import pyarrow.compute as pc  # ignore-banned-imports
 
         native_series = self._native_series
-        other_ = pa.null() if other is None else pa.scalar(other, type=native_series.type)
         return maybe_extract_py_scalar(  # type: ignore[no-any-return]
-            pc.is_in(other_, native_series),
+            pc.is_in(pa.scalar(other, type=native_series.type), native_series),
             return_py_scalar=True,
         )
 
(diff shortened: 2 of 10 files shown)


narwhals/_pandas_like/series.py
```diff
@@ -851,6 +851,11 @@ def rolling_mean(
     def __iter__(self: Self) -> Iterator[Any]:
         yield from self._native_series.__iter__()
 
+    def __contains__(self: Self, other: Any) -> bool:
+        if other is None:
+            return self._native_series.isna().any()  # type: ignore[no-any-return]
```
FBruzzesi, line 856: @MarcoGorelli thoughts on this?

MarcoGorelli: i think this seems reasonable, nice

narwhals/_pandas_like/series.py
```diff
@@ -851,6 +851,11 @@ def rolling_mean(
...
+        return other in self._native_series
```
FBruzzesi, line 857: I fell for the example contained in our [why](https://narwhals-dev.github.io/narwhals/why/) section!

There are two ways of doing this:

- `return other in self.to_numpy()`: yet I don't think this is a good idea for cudf
- `return self._native_series.isin({other}).any()`

AlessandroMiola: May I ask a general, possibly silly, question, just for the sake of better understanding? 😇 

I see the point in implementing `__contains__` for _pandas-like_ and _arrow_ and also what you mention here. Instead, I haven't properly got the details behind https://github.com/narwhals-dev/narwhals/issues/1443#issuecomment-2508932494, as I don't see `func(1, pa.chunked_array([data]))` returning `False`.

Should we implement `__contains__` at `narwhals/series.py` to ensure the above behaviour is explicitly triggered via `__contains__`?

FBruzzesi: > as I don't see func(1, pa.chunked_array([data])) returning False.

Damn that's true... but now I don't know why, and I wonder what is being called. The following returns false:

```py
1 in pa.chunked_array([[1,2,3]])
False
```

> Should we implement __contains__ at narwhals/series.py to ensure the above behaviour is explicitly triggered via __contains__?

I am addressing it in the PR

AlessandroMiola: > Damn that's true... but now I don't know why, and I wonder what is being called.

Me neither, had issues in trying to figure it out

> I am addressing it in the PR

yeah, I see it; was just asking to confirm my understanding as I couldn't indeed see where this was coming from, without the explicit reference to `__contains__` 😃

FBruzzesi: Ok, since we implement `__iter__` the check works.

Example:

```py
class A:
    pass

class B:
    def __iter__(self):
        yield from [1,2,3]

0 in B(), 1 in B()
> (False, True)

1 in A()
> TypeError: argument of type 'A' is not iterable
```

AlessandroMiola: nice, thanks for the clarification!:)

narwhals/_arrow/series.py
```diff
@@ -1017,6 +1017,16 @@ def __iter__(self: Self) -> Iterator[Any]:
             for x in self._native_series.__iter__()
         )
 
+    def __contains__(self: Self, other: Any) -> bool:
+        import pyarrow as pa  # ignore-banned-imports
+        import pyarrow.compute as pc  # ignore-banned-imports
+
+        native_series = self._native_series
+        return maybe_extract_py_scalar(  # type: ignore[no-any-return]
+            pc.is_in(pa.scalar(other, type=native_series.type), native_series),
```
MarcoGorelli, line 1026: if `other` is of a type not convertible to `native_series.type`, will this raise? I'm wondering if we should return `False` in such a case?

MarcoGorelli: looks like polars would also raise here:
```
In [2]: 'foo' in pl.Series([1,2,3])
InvalidOperationError                     Traceback (most recent call last)
Cell In[2], line 1
----> 1 'foo' in pl.Series([1,2,3])

File ~/scratch/.venv/lib/python3.12/site-packages/polars/series/series.py:1233, in Series.__contains__(self, item)
   1231 if item is None:
   1232     return self.has_nulls()
-> 1233 return self.implode().list.contains(item).item()

File ~/scratch/.venv/lib/python3.12/site-packages/polars/series/utils.py:106, in call_expr.<locals>.wrapper(self, *args, **kwargs)
    104     expr = getattr(expr, namespace)
    105 f = getattr(expr, func.__name__)
--> 106 return s.to_frame().select_seq(f(*args, **kwargs)).to_series()

File ~/scratch/.venv/lib/python3.12/site-packages/polars/dataframe/frame.py:9138, in DataFrame.select_seq(self, *exprs, **named_exprs)
   9115 def select_seq(
   9116     self, *exprs: IntoExpr | Iterable[IntoExpr], **named_exprs: IntoExpr
   9117 ) -> DataFrame:
   9118     """
   9119     Select columns from this DataFrame.
   9120
   (...)
   9136     select
   9137     """
-> 9138     return self.lazy().select_seq(*exprs, **named_exprs).collect(_eager=True)

File ~/scratch/.venv/lib/python3.12/site-packages/polars/lazyframe/frame.py:2029, in LazyFrame.collect(self, type_coercion, predicate_pushdown, projection_pushdown, simplify_expression, slice_pushdown, comm_subplan_elim, comm_subexpr_elim, cluster_with_columns, collapse_joins, no_optimization, streaming, engine, background, _eager, **_kwargs)
   2027 # Only for testing purposes
   2028 callback = _kwargs.get("post_opt_callback", callback)
-> 2029 return wrap_df(ldf.collect(callback))

InvalidOperationError: is_in operation not supported for dtypes `str` and `list[i64]`
```
probably ok to just follow them here then

narwhals/_pandas_like/series.py
```diff
@@ -851,6 +851,11 @@ def rolling_mean(
     def __iter__(self: Self) -> Iterator[Any]:
         yield from self._native_series.__iter__()
 
+    def __contains__(self: Self, other: Any) -> bool:
+        if other is None:
+            return self._native_series.isna().any()  # type: ignore[no-any-return]
+        return self._native_series.isin({other}).any()  # type: ignore[no-any-return]
```
MarcoGorelli, line 857: why this over `return (self._native_series == other).any()`?

narwhals/_arrow/series.py
MarcoGorelli: still, probably good to have a test to check it raises? we can leave the expression unification for a separate topic

FBruzzesi: I will add a test and make sure that same exception is raised 👌

narwhals/_pandas_like/series.py
FBruzzesi: No strong feelings 😂 I guess I went for `series.isin(...)` as I initially fell for the `other in series`

MarcoGorelli: "fell for" 😄 love that

from a quick test, it looks a lot faster
```
In [14]: %timeit s.isin({0}).any()
6.58 ms ± 687 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [15]: %timeit (s == 0).any()
326 μs ± 45.8 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

narwhals/_arrow/series.py
FBruzzesi: Ok this is definitly wrong behaviour for arrow as the casting would convert from float to int, leading to the following:

```py
import pyarrow as pa
import narwhals as nw

data = [100, 200, None]
s = nw.from_native(pa.table({"a": data}), eager_only=True)["a"]

100.3 in s
True
```

On the Exception side, I would not be sure how to catch the error for pandas as:

```py
(pd.Series([1,2,3]) == "foo").any()
np.False_
```

MarcoGorelli: tbh we could just dispatch do `(s._compliant_series == other).any()` for all backends 😄 that would still be better than the status quo

narwhals/__init__.py
```diff
@@ -31,22 +31,22 @@
 from narwhals.dtypes import UInt64
 from narwhals.dtypes import Unknown
 from narwhals.expr import Expr
-from narwhals.expr import all_ as all  # noqa: A004
+from narwhals.expr import all_ as all
```
FBruzzesi, line 34: Ok something must have gone wrong with the pre-commit sync :|

narwhals/_arrow/series.py
```diff
@@ -1021,6 +1021,31 @@ def __iter__(self: Self) -> Iterator[Any]:
...
+            other_ = (
+                pa.scalar(other)
+                if other is not None
+                else pa.scalar(None, type=native_series.type)
+            )
+            return maybe_extract_py_scalar(  # type: ignore[no-any-return]
+                pc.is_in(other_, native_series),
+                return_py_scalar=True,
+            )
+        except (ArrowInvalid, ArrowNotImplementedError, ArrowTypeError) as exc:
```
FBruzzesi, line 1043: Should it be bare exception?

narwhals/__init__.py
MarcoGorelli: 🤔 i'm so confused 😄

MarcoGorelli (approved):
thanks @FBruzzesi ! and @AlessandroMiola for review

MarcoGorelli:
ah perhaps this comment i left didn't help

> that would still be better than the status quo

by 'status quo' i meant the main branch, not the previous commit in this pr

PR #1533, opened 2024-12-07
feat: allow passing Series to Series.__getitem__

### LINKED ISSUE #1525: enh: allow passing Series to Series.__getitem__
With Polars we can do:
```python
In [8]: spl = pl.Series([1,2,3])

In [9]: spl[spl[0, 1]]
Out[9]:
shape: (2,)
Series: '' [i64]
[
        2
        3
]
```
But Narwhals gives an incomprehensible error message:
```python
In [13]: snw = nw.from_native(spl, series_only=True)

In [14]: snw[snw[0, 1]]
Cell In[14], line 1
----> 1 snw[snw[0, 1]]

File ~/scratch/.venv/lib/python3.12/site-packages/narwhals/series.py:72, in Series.__getitem__(self, idx)
     70 if isinstance(idx, int):
     71     return self._compliant_series[idx]
---> 72 return self._from_compliant_series(self._compliant_series[idx])

File ~/scratch/.venv/lib/python3.12/site-packages/narwhals/_polars/series.py:126, in PolarsSeries.__getitem__(self, item)
    125 def __getitem__(self: Self, item: int | slice | Sequence[int]) -> Any | Self:
--> 126     return self._from_native_object(self._native_series.__getitem__(item))

File ~/scratch/.venv/lib/python3.12/site-packages/polars/series/series.py:1282, in Series.__getitem__(self, key)
   1254 def __getitem__(
   1255     self, key: SingleIndexSelector | MultiIndexSelector
   1256 ) -> Any | Series:
   1257     """
   1258     Get part of the Series as a new Series or scalar.
   1259
   (...)
   1280     ]
   1281     """
-> 1282     return get_series_item_by_key(self, key)

File ~/scratch/.venv/lib/python3.12/site-packages/polars/_utils/getitem.py:90, in get_series_item_by_key(s, key)
     87     return _select_elements_by_index(s, indices)
     89 msg = f"cannot select elements using key of type {type(key).__name__!r}: {key!r}"
---> 90 raise TypeError(msg)

TypeError: cannot select elements using key of type 'Series': ┌─────────────────────────────────────────┐
| Narwhals Series                         |
| Use `.to_native()` to see native output |
└─────────────────────────────────────────┘
```

I think we should:
- allow this if the Series is of integral dtypes
- raise if the Series is boolean, advising users to use `Series.filter` instead

devdanzin:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1525
- Closes #1525

## Checklist

- [x] Tests added
- [ ] Documented the changes

This PR allows passing `Series` to `Series.__getitem__` and adds a test. I'm not sure about how/where to document the changes, any suggestions?

### DIFF
--- a/narwhals/_polars/series.py
+++ b/narwhals/_polars/series.py
@@ -121,10 +121,10 @@ class PolarsSeries:
     def __getitem__(self: Self, item: int) -> Any: ...
 
     @overload
-    def __getitem__(self: Self, item: slice | Sequence[int]) -> Self: ...
+    def __getitem__(self: Self, item: slice | Sequence[int] | NWSeries[Any]) -> Self: ...
 
     def __getitem__(
-            self: Self, item: int | slice | Sequence[int] | NWSeries
+        self: Self, item: int | slice | Sequence[int] | NWSeries[Any]
     ) -> Any | Self:
         if isinstance(item, (int, slice, Sequence)):
             return self._from_native_object(self._native_series.__getitem__(item))
--- a/narwhals/_polars/series.py
+++ b/narwhals/_polars/series.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
 
     from narwhals._polars.dataframe import PolarsDataFrame
     from narwhals.dtypes import DType
+    from narwhals.series import Series as NWSeries
     from narwhals.utils import Version
 
     T = TypeVar("T")
@@ -122,8 +123,12 @@ class PolarsSeries:
     @overload
     def __getitem__(self: Self, item: slice | Sequence[int]) -> Self: ...
 
-    def __getitem__(self: Self, item: int | slice | Sequence[int]) -> Any | Self:
-        return self._from_native_object(self._native_series.__getitem__(item))
+    def __getitem__(
+            self: Self, item: int | slice | Sequence[int] | NWSeries
+    ) -> Any | Self:
+        if isinstance(item, (int, slice, Sequence)):
+            return self._from_native_object(self._native_series.__getitem__(item))
+        return self._from_native_object(self._native_series.__getitem__(item.to_numpy()))
 
     def cast(self: Self, dtype: DType) -> Self:
         ser = self._native_series
--- a/tests/series_only/__getitem___test.py
+++ b/tests/series_only/__getitem___test.py
@@ -1,9 +1,10 @@
 from __future__ import annotations
 
-import pytest
 import polars as pl
+import pytest
 
 import narwhals.stable.v1 as nw
+
 
 def test_getitem() -> None:
     spl = pl.Series([1, 2, 3])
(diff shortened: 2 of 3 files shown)


narwhals/_polars/series.py
```diff
@@ -120,10 +121,14 @@ def dtype(self: Self) -> DType:
...
-    def __getitem__(self: Self, item: int | slice | Sequence[int]) -> Any | Self:
-        return self._from_native_object(self._native_series.__getitem__(item))
+    def __getitem__(self: Self, item: slice | Sequence[int] | NWSeries[Any]) -> Self: ...
+
+    def __getitem__(
+        self: Self, item: int | slice | Sequence[int] | NWSeries[Any]
+    ) -> Any | Self:
+        if isinstance(item, (int, slice, Sequence)):
+            return self._from_native_object(self._native_series.__getitem__(item))
+        return self._from_native_object(self._native_series.__getitem__(item.to_numpy()))
```
MarcoGorelli, line 131: 🤔 not sure we should be converting to numpy here

tests/series_only/__getitem___test.py
```diff
@@ -0,0 +1,18 @@
...
+
+import polars as pl
+import pytest
+
+import narwhals.stable.v1 as nw
+
+
+def test_getitem() -> None:
+    spl = pl.Series([1, 2, 3])
+    assert spl[spl[0, 1]].equals(pl.Series([2, 3]))
```
MarcoGorelli, line 11: can we check what happens when there's missing values too please?

MarcoGorelli (commented):
thanks for your PR!

just left a couple of comments

narwhals/_polars/series.py
devdanzin: Thank you, I had doubts about this too and picked a solution that seemed to work without actually understanding the consequences.

devdanzin: It seems using `item.to_native()` instead works, and also solves an issue with missing values becoming `NaN` with `to_numpy()`. Do you think it's acceptable? Any other corner cases to cover with tests?

tests/series_only/__getitem___test.py
devdanzin: Using `to_native()` instead of `to_numpy()` allows the following tests to pass, but my complete lack of knowledge on numpy, narwhals, polars and related subjects makes me unsure they are correct and show expected behavior:
```python
    spl = pl.Series([1, None, 3])
    assert spl[spl[0, 1]].equals(pl.Series([None, None]))

    snw = nw.from_native(spl, series_only=True)
    assert snw[snw[0, 1]].to_native().equals(pl.Series([None, None]))
```

What's your take on this?

narwhals/_polars/series.py
MarcoGorelli: thanks - i'll check, to be careful we get the details right here

PR #1536, opened 2024-12-07
feat: `ListNamespace.len()` for `Expr` and `Series`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

Introduces `ListNamespace` and only one method (`.len`) for Series and Expr.

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -1167,5 +1167,16 @@ class PandasLikeSeriesListNamespace:
         self._compliant_series = series
 
     def len(self: Self) -> PandasLikeSeries:
+        from narwhals.utils import import_dtypes_module
+
         s = self._compliant_series._native_series
-        return self._compliant_series._from_native_series(s.__class__(s.list.len()))
+
+        native_result = s.__class__(s.list.len())
+        dtype = narwhals_to_native_dtype(
+            dtype=import_dtypes_module(self._compliant_series._version).UInt32(),
+            starting_dtype=native_result.dtype,
+            implementation=self._compliant_series._implementation,
+            backend_version=self._compliant_series._backend_version,
+            version=self._compliant_series._version,
+        )
+        return self._compliant_series._from_native_series(native_result.astype(dtype))
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -864,6 +864,10 @@ class PandasLikeSeries:
     def cat(self) -> PandasLikeSeriesCatNamespace:
         return PandasLikeSeriesCatNamespace(self)
 
+    @property
+    def list(self) -> PandasLikeSeriesListNamespace:
+        return PandasLikeSeriesListNamespace(self)
+
 
 class PandasLikeSeriesCatNamespace:
     def __init__(self, series: PandasLikeSeries) -> None:
@@ -1156,3 +1160,12 @@ class PandasLikeSeriesDateTimeNamespace:
             raise TypeError(msg)
         result[mask_na] = None
         return self._compliant_series._from_native_series(result)
+
+
+class PandasLikeSeriesListNamespace:
+    def __init__(self, series: PandasLikeSeries) -> None:
+        self._compliant_series = series
+
+    def len(self: Self) -> PandasLikeSeries:
+        s = self._compliant_series._native_series
+        return self._compliant_series._from_native_series(s.__class__(s.str.len()))
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -1169,9 +1169,8 @@ class PandasLikeSeriesListNamespace:
     def len(self: Self) -> PandasLikeSeries:
         from narwhals.utils import import_dtypes_module
 
-        s = self._compliant_series._native_series
-
-        native_result = s.__class__(s.list.len())
+        native_series = self._compliant_series._native_series
+        native_result = native_series.list.len().rename(native_series.name, copy=False)
         dtype = narwhals_to_native_dtype(
             dtype=import_dtypes_module(self._compliant_series._version).UInt32(),
             starting_dtype=native_result.dtype,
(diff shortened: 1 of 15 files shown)


tests/expr_and_series/list/len_test.py
```diff
@@ -0,0 +1,45 @@
...
+    constructor: Constructor,
+) -> None:
+    if any(backend in str(constructor) for backend in ("dask", "modin", "cudf")):
+        request.applymarker(pytest.mark.xfail)
+
+    if "pandas" in str(constructor) and PANDAS_VERSION < (2, 2):
+        request.applymarker(pytest.mark.xfail)
+
+    result = nw.from_native(constructor(data)).select(
+        a=nw.col("a").cast(nw.List(nw.Int32())).list.len()
```
FBruzzesi, line 26: This seems like a minor detail, yet it is a big deal, because pandas seems to not preserve the series name. I opened an issue upstream [pandas#60522](https://github.com/pandas-dev/pandas/issues/60522).

I tried a workaround with forcing `root_names` and `output_names` in the `ListNamespace` method but I couldn't manage to make it work properly.

narwhals/_polars/series.py
```diff
@@ -407,3 +411,27 @@ def func(*args: Any, **kwargs: Any) -> Any:
...
+class PolarsSeriesListNamespace:
+    def __init__(self: Self, series: PolarsSeries) -> None:
+        self._series = series
+
+    def len(self: Self) -> PolarsSeries:
+        native_series = self._series._native_series
+        if self._series._backend_version < (1, 16):  # pragma: no cover
+            native_result = native_series.list.len().zip_with(
+                ~native_series.is_null(), native_series.list.first()
+            )
```
FBruzzesi, line 425: Probably there is a better way that using `native_series.list.first()`?

narwhals/_polars/expr.py
```diff
@@ -236,3 +240,29 @@ def func(*args: Any, **kwargs: Any) -> PolarsExpr:
...
+class PolarsExprListNamespace:
+    def __init__(self: Self, expr: PolarsExpr) -> None:
+        self._expr = expr
+
+    def len(self: Self) -> PolarsExpr:
+        native_expr = self._expr._native_expr
+        if self._expr._backend_version < (1, 16):  # pragma: no cover
+            import polars as pl
+
+            native_result: pl.Expr = pl.when(~native_expr.is_null()).then(
```
FBruzzesi, line 254: type annotation is to avoid mypy to raise

narwhals/_pandas_like/utils.py
```diff
@@ -567,9 +570,25 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
     if isinstance_or_issubclass(dtype, dtypes.Enum):
         msg = "Converting to Enum is not (yet) supported"
         raise NotImplementedError(msg)
-    if isinstance_or_issubclass(dtype, dtypes.List):  # pragma: no cover
-        msg = "Converting to List dtype is not supported yet"
-        return NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.List):
+        if implementation is Implementation.PANDAS and backend_version >= (2, 2):
```
FBruzzesi, line 574: Dask and Modin both seem to fail with this type, I assumed cudf as well

FBruzzesi:
@MarcoGorelli test is failing as polars seems to have changed type of `.list.len()` from Int64 to UInt32 (reasonably so). I am always uncertain for what to do in these situation. We could cast to such type for the user in each backend, but I am not sure if we should

narwhals/_polars/series.py
MarcoGorelli: how about
```python
pl.select(pl.when(s.is_not_null()).then(s.list.len()).otherwise(None))[s.name]
```
?

MarcoGorelli:
> We could cast to such type for the user in each backend, but I am not sure if we should

yeah i hear you, unsigned integers in pandas are particularly surprising too...we can probably keep it consistent for each backend though?

narwhals/_polars/series.py
FBruzzesi: TIL about `pl.select` 👌

narwhals/_pandas_like/utils.py
```diff
@@ -567,9 +570,25 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
+            import pandas as pd  # ignore-banned-import
+            import pyarrow as pa  # ignore-banned-import
```
MarcoGorelli, line 576: separately to this PR, but when we do a banned import, we should probably do a try-except and raise a message from narwhals if it fails

narwhals/_polars/expr.py
```diff
@@ -236,3 +240,37 @@ def func(*args: Any, **kwargs: Any) -> PolarsExpr:
...
+        native_expr = self._expr._native_expr
+        native_result = native_expr.list.len()
+
+        if self._expr._backend_version < (1, 16):  # pragma: no cover
+            import polars as pl
+
+            native_result: pl.Expr = (  # type: ignore[no-redef]
+                pl.when(~native_expr.is_null()).then(native_result).cast(pl.UInt32())
+            )
+        if self._expr._backend_version < (1, 17):  # pragma: no cover
```
MarcoGorelli, line 259: does `elif` work here?

narwhals/_pandas_like/series.py
```diff
@@ -1156,3 +1160,23 @@ def timestamp(self, time_unit: Literal["ns", "us", "ms"] = "us") -> PandasLikeSe
...
+class PandasLikeSeriesListNamespace:
+    def __init__(self, series: PandasLikeSeries) -> None:
+        self._compliant_series = series
+
+    def len(self: Self) -> PandasLikeSeries:
+        from narwhals.utils import import_dtypes_module
+
+        s = self._compliant_series._native_series
+
+        native_result = s.__class__(s.list.len())
```
MarcoGorelli, line 1174: can we not set the name here? as in, use `self._compliant_series.name`?

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1542, opened 2024-12-09
feat: add `DataFrame` and `LazyFrame` `explode` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/_arrow/dataframe.py
```diff
@@ -743,3 +744,78 @@ def unpivot(
         )
         # TODO(Unassigned): Even with promote_options="permissive", pyarrow does not
         # upcast numeric to non-numeric (e.g. string) datatypes
+
+    def explode(self: Self, columns: str | Sequence[str], *more_columns: str) -> Self:
```
FBruzzesi, line 748: pyarrow has two paths:

- if nulls or empty lists are not present, then it is enough to:
  1. make sure element counts are same
  2. explode each array individually
- if nulls or empty lists are present, then these are ignore by `pc.list_parent_indices` and `pc.list_flatten`, which _is_ a problem. This implementation falls back to a python list both to flatten the array(s) and to create the corresponding indices .

After flattening, a new table is created by `take`-ing the indices of the non-flattened arrays and the flattened arrays.

narwhals/_pandas_like/dataframe.py
```diff
@@ -937,3 +937,52 @@ def unpivot(
                 value_name=value_name if value_name is not None else "value",
             )
         )
+
+    def explode(self: Self, columns: str | Sequence[str], *more_columns: str) -> Self:
```
FBruzzesi, line 941: If a single column is to be exploded, then we use the pandas native method. If multiple columns, the strategy is to explode the one column with the rest of the dataframe, and the other series individually and finally concatenating them back, plus sorting by original column names order

narwhals/dataframe.py
```diff
@@ -4643,3 +4720,56 @@ def unpivot(
...
+        Notes:
+            It is possible to explode multiple columns only if these columns must have
+            matching element counts.
+
+        Arguments:
+            columns: Column names. The underlying columns being exploded must be of the `List` data type.
+            *more_columns: Additional names of columns to explode, specified as positional arguments.
+
+        Returns:
+            New DataFrame
```
FBruzzesi, line 4736:
```suggestion
            New LazyFrame
```

narwhals/_arrow/dataframe.py
```diff
@@ -749,3 +750,78 @@ def unpivot(
...
+                [
+                    i
+                    for i, count in enumerate(counts.to_pylist())
+                    for _ in range(max(count or 1, 1))
+                ]
+            )
+
+            def explode_null_array(array: pa.ChunkedArray) -> pa.ChunkedArray:
+                exploded_values = []  # type: ignore[var-annotated]
+                for lst_element in array.to_pylist():
```
MarcoGorelli, line 806: i might be missing something but this looks potentially very expensive

FBruzzesi: It definitely is 😐 happy to raise for nullable cases for now and try to figure out a native alternative

FBruzzesi: I did some improvements 🙌🏼 now there is only one `.to_pylist` and no loop even in the case of nulls or empty lists

MarcoGorelli:
thanks - tbh i'm still not too keen on having to_pylist, should be just not support pyarrow and raise this as a feature request with them?

FBruzzesi:
> thanks - tbh i'm still not too keen on having to_pylist, should be just not support pyarrow and raise this as a feature request with them?

Let me take a detour in the rabbit hole before a final decision 😇

FBruzzesi:
@MarcoGorelli thoughts on the following?

The idea being, if consecutive parent indices are the same, their diff will be zero, therefore the counter of such index should increase, otherwise it should not. By how much you ask? the cumulative count starting from 0 (that's why of the -1 after the cumulative sum) - I will definitly add a comment if we decide to move forward with this 

```py
parent_indices = pc.list_parent_indices(native_frame[to_explode[0]])
diff = pc.pairwise_diff(parent_indices.combine_chunks())
is_change_point = pc.or_kleene(pc.equal(diff, 0), pc.is_null(diff))
indices = pc.subtract(
    pc.add(
        parent_indices,
        pc.cumulative_sum(is_change_point.cast(pa.int32()))
    ), 1
)

is_valid_index = pc.is_in(indices, value_set=parent_indices)
exploded_size = len(is_valid_index)

def flatten_func(array: pa.ChunkedArray) -> pa.ChunkedArray:
    dtype = array.type.value_type
    return pc.replace_with_mask(
        pa.array([None] * exploded_size, type=dtype),
        is_valid_index,
        pc.list_flatten(array).combine_chunks(),
    )
```

MarcoGorelli:
nice - imma have to spend some time understanding this 😄 😳

FBruzzesi:
> nice - imma have to spend some time understanding this 😄 😳

I deduce that my explanation was garbage. Let me _explode_ the previous comment

MarcoGorelli:
Not at all, I just hadn't spent enough time on it , will take a look and make sense of it!

narwhals/_pandas_like/dataframe.py
```diff
@@ -949,3 +949,52 @@ def unpivot(
...
+            if isinstance(columns, str)
+            else [*columns, *more_columns]
+        )
+        schema = self.collect_schema()
+        for col_to_explode in to_explode:
+            dtype = schema[col_to_explode]
+
+            if dtype != dtypes.List:
+                msg = f"`explode` operation not supported for dtype `{dtype}`"
+                raise InvalidOperationError(msg)
```
MarcoGorelli, line 969: can we also write which type we expect (List) in the error message?

MarcoGorelli (commented):
OK thanks for explaining - sure, happy to try this out, maybe I could write a hypothesis test to stress test it

would also be open to merging for now without pyarrow (as the rest is quite uncontroversial) and then adding pyarrow later?

narwhals/_arrow/dataframe.py
```diff
@@ -752,3 +753,70 @@ def unpivot(
...
+        if fast_path:
+            indices = pc.list_parent_indices(native_frame[to_explode[0]])
+            flatten_func = pc.list_flatten
+
+        else:
+            msg = (
+                "`DataFrame.explode` is not supported for pyarrow backend and column"
+                "containing null's or empty list elements"
+            )
+            raise NotImplementedError(msg)
```
MarcoGorelli, line 808: do we want a value-dependent error to be raised? would you be opposed to raising completely for pyarrow and raising a feature request on their part?

FBruzzesi: Oh, I see! I am fairly happy with both cases implementations as they are native.
But it's also ok to have a dedicated PR for it. I will adjust

FBruzzesi: Maybe I should also check how pandas does it with pyarrow backend?

MarcoGorelli (approved):
thanks @FBruzzesi ! feel free to merge when ready

PR #1547, opened 2024-12-09
ci: actually test against pandas nightly, don't filter "copy is deprecated" warnings

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -2,7 +2,7 @@ exclude-labels:
   - ignore
   - release
   - dependencies
-name-template: 'Narwhals v$RESOLVED_VERSION'
+name-template: 'Narwhals unreleased'
 version-template: unreleased
 
 change-template: '- $TITLE (#$NUMBER)' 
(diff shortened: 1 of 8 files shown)


MarcoGorelli:
OK so pandas nightly wasn't actually being tested because of the order of installations

Like this we get a bunch of
```python
2024-12-09T15:02:47.8342945Z E           DeprecationWarning: The copy keyword is deprecated and will be removed in a future version. Copy-on-Write is active in pandas since 3.0 which utilizes a lazy copy mechanism that defers copies until necessary. Use .copy() to make an eager copy if necessary.
```
from all the `copy=False` we have

We should only use `copy=False` for pandas<3

We already have `set_axis` which deals with whether or not to pass `copy` depending on the backend and version, we should do the same for `rename` and anything else we call with a deprecated `copy` arg

PR #1551, opened 2024-12-10
feat: `read_csv`

### LINKED ISSUE #1112: feat: read_csv
I was initially hesitant about adding IO methods, the idea being "users provide their own dataframe, we just deal with how to process it", but we already have [from_dict](https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_dict), and https://github.com/ImperialCollegeLondon/pycsvy/issues/83 and Temporian look like good use cases for `read_csv`

pandas and Polars each have dozens of `read_csv` methods...so we may need to careful here about which ones we add, and perhaps only start with the most common ones

The api would be something like
```python
import pandas as pd
nw.read_csv(file, native_namespace=pd)
```
```python
import polars as pl
nw.read_csv(file, native_namespace=pl)
```

We could do:
- `nw.read_csv`: this is eager-only and always returns `nw.DataFrame`
- `nw.scan_csv`: this is the most generic one, and returns `nw.LazyFrame` if possible (e.g. Polars), else `nw.DataFrame`

---

## Alternatives

Keep the status-quo: users are responsible for doing their own IO

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- closes #1112 
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/docs/api-reference/narwhals.md
+++ b/docs/api-reference/narwhals.md
@@ -37,6 +37,7 @@ Here are the top-level functions available in Narwhals.
         - narwhalify
         - new_series
         - nth
+        - read_csv
         - sum
         - sum_horizontal
         - show_versions
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -55,6 +55,7 @@ from narwhals.functions import from_dict
 from narwhals.functions import from_numpy
 from narwhals.functions import get_level
 from narwhals.functions import new_series
+from narwhals.functions import read_csv
 from narwhals.functions import show_versions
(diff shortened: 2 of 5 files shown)


MarcoGorelli (approved):
thanks @raisadz !

Issue #1296, opened 2024-11-01
[Bug]: pandas-wrapped narwhals .over() leads to inconsistent result


mscolnick:
### Describe the bug

```python
import marimo as mo
import narwhals as nw
import polars as pl
import pandas as pd

data = {"item": ["a", "a", "a", "b", "b", "b"], "start": [6, 7, 8, 14, 15, 16]}

df_pandas = pd.DataFrame(data)
df_polars = pl.DataFrame(data)
nw_df_pandas = nw.from_native(df_pandas)
nw_df_polars = nw.from_native(df_polars)

# Correct, creates a new column with the shifted values
nw_df_polars.with_columns(
    (nw.col("start") * 1 - nw.min("start")).over("item").alias("shifted")
).to_native()

# Incorrect, all starts at 0
nw_df_pandas.with_columns(
    (nw.col("start") * 1 - nw.min("start")).over("item").alias("shifted")
).to_native()
```

<img width="1028" alt="image" src="https://github.com/user-attachments/assets/35f4320b-a5a6-4aef-aa73-098c16ff5eb1">

### Steps or code to reproduce the bug

https://static.marimo.app/static/narwhals-bug-bo1z

### Expected results

Expected the results of polars. Values are incrementing.

### Actual results

Values are the same

### Please run narwhals.show_version() and enter the output below.

```shell
System:
    python: 3.12.6 (main, Sep  6 2024, 19:03:47) [Clang 15.0.0 (clang-1500.3.9.4)]
executable: /Users/myles/.cache/uv/archive-v0/sNIzHlLTqqBXxH79AubYO/bin/python
   machine: macOS-14.6-arm64-arm-64bit

Python dependencies:
     narwhals: 1.12.1
       pandas: 2.2.3
       polars: 1.12.0
         cudf: 
        modin: 
      pyarrow: 17.0.0
        numpy: 2.1.2
```
```


### Relevant log output

_No response_

**Comment by mscolnick** [comment]
Ah, I got it. I am probably using the API wrong. I was doing:

```python
(
        pl.col("year") * 12
        + pl.col("month")
        - pl.min("year") * 12
        - pl.min("month")
        + 1
    ).over("project")
```

but since this worked (in polars), I didn't think twice about doing 

```python
(
        pl.col("year") * 12
        + pl.col("month")
        - pl.min("year").over("project") * 12
        - pl.min("month").over("project")
        + 1
    )
```

Makes sense, and fine to close!

[...]

PR #1584, opened 2024-12-13
feat: Improve error message when trying to do unsupported group-by aggregation

### LINKED ISSUE #1575: [Bug]: Pandas Dataframe as native does not work with .shift().over()
### Describe the bug

The below code works if the native dataframe is polars. However, getting an error message using pandas.

```
import polars as pl
import narwhals as nw
from narwhals.typing import FrameT

@nw.narwhalify
def create_lag(df: FrameT) -> FrameT:
    return df.with_columns(
        nw.col("a")
        .shift(1)
        .over("grp")
        .alias("lag_a")
    )

df = pl.DataFrame(
    {
        "a": [1, 2, 3, 4, 5],
        "grp": [1, 1, 2, 2, 2]
    }
)

l = create_lag(df)

df = df.to_pandas()
l2 = create_lag(df)

```

### Expected results

Both the polars and pandas dataframe should add a new column lag_a.

### Actual results

Getting the following error when pandas is the native dataframe

```
narwhals.exceptions.ColumnNotFoundError: The following columns were not found: ['grp']

Hint: Did you mean one of these columns: ['index', 'a']?

```

```shell
Python dependencies:
     narwhals: 1.17.0
       pandas: 2.2.2
       polars: 1.1.0
         cudf: 
        modin: 
      pyarrow: 16.0.0
        numpy: 1.26.4
```

### Relevant log output

_No response_

MarcoGorelli:
closes #1575 
closes #1584

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


MarcoGorelli:
Also, re-reading https://github.com/narwhals-dev/narwhals/issues/1296, it's worse than I thought, the results are actually incorrect. I think we need a release to fix it then

narwhals/_arrow/group_by.py
```diff
@@ -36,7 +41,7 @@ def get_function_name_option(
     import pyarrow.compute as pc
 
     function_name_to_options = {
-        "count": pc.CountOptions(mode="all"),
+        "count": pc.CountOptions(mode="only_valid"),
```
AlessandroMiola, line 44: Hi Marco, I might be wrong, but, while trying to parametrize tests for agg methods in "simple" group_by(s), I've noticed that while this change is coherent with `polars`' `count` behaviour, it might not align with the behaviour of `polars`' `len`.

```
@pytest.mark.parametrize(
    ("agg_expr", "expected"),
    [
        (nw.col("b").len(), {"a": [1, 2], "b": [3, 1]}),
    ],
)
def test_group_by_aggr(
    constructor: Constructor,
    agg_expr: nw.Expr,
    expected: dict[str, list[int | float]],
) -> None:
    data = {"a": [1, 1, 1, 2], "b": [1, None, 2, 3]}
    result = (
        nw.from_native(constructor(data)).group_by("a").agg(agg_expr).sort("a")
    )
```
fails for `pyarrow`.

I understand, though, that the issue is rather on `len` - for which we miss native support in groupby - rather than on `count`.

Just wanted to bring this to your attention, hope it's relevant!

MarcoGorelli: thanks @AlessandroMiola ! you're right - I was thinking that we only supported `agg(nw.len())` (for which we are consistent across backends) but indeed, we also support `agg(nw.col('a').len())`, for which pyarrow skips nulls but shouldn't

were you interested in making a PR to fix this?

AlessandroMiola: Sure, I can try :)

PR #1593, opened 2024-12-14
docs: add tea-tasting to ecosystem

e10v:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

tea-tasting is a Python package for the statistical analysis of A/B tests. Since version 0.3.0 it uses Narwhals to support more dataframes. With this PR I'd like to add tea-tasting to the list of projects in the Narwhals ecosystem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ Join the party!
 - [scikit-lego](https://github.com/koaning/scikit-lego)
 - [scikit-playtime](https://github.com/koaning/scikit-playtime)
 - [tabmat](https://github.com/Quantco/tabmat)
+- [tea-tasting](https://github.com/e10v/tea-tasting)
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
yay, thanks @e10v , and well done with tea-tasting!

I'd also be open to having it in the downstream tests if you fancied adding it there: https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml, then we can have an extra layer of insurance that we won't break things for you

PR #1603, opened 2024-12-16
feat: add `var` to `Expr` and `Series`

### LINKED ISSUE #1586: test: test group_by(...).agg(nw.col(...).var())
Realised today that we don't have any tests for `df.group_by(...).agg(nw.col(...).var())`. Let's add one!

This should be very similar to the std one:

https://github.com/narwhals-dev/narwhals/blob/44cf9e39840f85901d39f8f44b24b6ea39050c3b/tests/group_by_test.py#L130-L136

You could even just parametrise over the aggregation function and `expected`. Or just add it below

Dhanunjaya-Elluri:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1586 
- Closes #1586 

## Checklist

- [x] Tests added
- [x] Documented the changes

### Documentation Updates:
- [docs/api-reference/expr.md](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-e69825f3917ed6bbcde7210918515e093c0583688a5e4a575e078c6ee015d964): Added `var` to the list of functions
- [docs/api-reference/series.md](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-5f45d42af45324bacb9364605e79f506699e0fbd901d25e1ede6e0a56a24f7c2): Added `var` to the list of functions

### Implementation of `var` in the following files:
- [narwhals/_arrow/expr.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-a2caa9f0d49838e83344364b32ff5012fbaa69a71257ff702b11229958d576b3)
- [narwhals/_arrow/series.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-f595265984b673e1c612607ebc3426434ee7162684840b99cd73fcce59fd6a7f)
- [narwhals/_dask/expr.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-faa0f0a1856d70f5a543fca7f7d342dcc3967c65fd55096ec55da0b3afac29a8)
- [narwhals/_pandas_like/expr.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-0ded7e876f1f6f0aa79abda7252f76be22c1a01b1e97e7b65a5da8fa23e35ffd)
- [narwhals/_pandas_like/series.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-59107550a5dea2a4834f7beb8d822b53e59f7abc7f24727dc48cd6d97c076c93)
- [narwhals/expr.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-8dbec323eb770e099734f31eb6dd160668841972f5549ea2d00779ea65853ec5)
- [narwhals/series.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-a05a3e6be8017e9767b6ee214af34fb2425fd0e038e3881165c4a27fe180004d)

### Add / Update of tests
- [tests/expr_and_series/var_test.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-18cf612611a2bd595e4a6fd32045ccbcd9e8a1baf4e8833a7b885fec7dc3c33a): Added new test file for `var`.
- [tests/group_by_test.py](https://github.com/narwhals-dev/narwhals/compare/main...Dhanunjaya-Elluri:feat/var-expr-series?expand=1#diff-c1bff5c041f237e9ed9771bf97d9a94aad9fdd213c328e1467dd1c4fade44b90)

This PR introduces `var` method to `Expr` and `Series`. This function returns variance of the given series. Updated necessary documentation and added necessary tests cases.

### DIFF
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -242,10 +242,10 @@ class ArrowExpr(CompliantExpr[ArrowSeries]):
     def n_unique(self: Self) -> Self:
         return reuse_series_implementation(self, "n_unique", returns_scalar=True)
 
-    def std(self: Self, ddof: int = 1) -> Self:
+    def std(self: Self, ddof: int) -> Self:
         return reuse_series_implementation(self, "std", ddof=ddof, returns_scalar=True)
 
-    def var(self: Self, ddof: int = 1) -> Self:
+    def var(self: Self, ddof: int) -> Self:
         return reuse_series_implementation(self, "var", ddof=ddof, returns_scalar=True)
 
     def skew(self: Self) -> Self:
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -245,6 +245,9 @@ class ArrowExpr(CompliantExpr[ArrowSeries]):
     def std(self: Self, ddof: int = 1) -> Self:
         return reuse_series_implementation(self, "std", ddof=ddof, returns_scalar=True)
 
+    def var(self: Self, ddof: int = 1) -> Self:
+        return reuse_series_implementation(self, "var", ddof=ddof, returns_scalar=True)
+
     def skew(self: Self) -> Self:
         return reuse_series_implementation(self, "skew", returns_scalar=True)
 
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -709,6 +709,10 @@ class Expr:
     def var(self, *, ddof: int = 1) -> Self:
         """Get variance.
 
+        Arguments:
+            ddof: "Delta Degrees of Freedom": the divisor used in the calculation is N - ddof,
+                     where N represents the number of elements. By default ddof is 1.
+
         Returns:
             A new expression.
 
@@ -724,16 +728,16 @@ class Expr:
 
             Let's define a dataframe-agnostic function:
 
-            >>> def my_library_agnostic_function(df_native: IntoFrameT) -> IntoFrameT:
+            >>> def agnostic_var(df_native: IntoFrameT) -> IntoFrameT:
             ...     df = nw.from_native(df_native)
             ...     return df.select(nw.col("a", "b").var(ddof=0)).to_native()
 
             We can then pass any supported library such as Pandas, Polars, or PyArrow to `func`:
 
-            >>> my_library_agnostic_function(df_pd)
+            >>> agnostic_var(df_pd)
                         a         b
             0  316.666667  1.602222
-            >>> my_library_agnostic_function(df_pl)
+            >>> agnostic_var(df_pl)
             shape: (1, 2)
             ┌────────────┬──────────┐
             │ a          ┆ b        │
@@ -742,7 +746,7 @@ class Expr:
             ╞════════════╪══════════╡
             │ 316.666667 ┆ 1.602222 │
             └────────────┴──────────┘
-            >>> my_library_agnostic_function(df_pa)
+            >>> agnostic_var(df_pa)
             pyarrow.Table
             a: double
             b: double
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -655,12 +655,12 @@ class Expr:
         """
         return self.__class__(lambda plx: self._to_compliant_expr(plx).median())
 
-    def std(self, *, ddof: int = 1) -> Self:
+    def std(self, *, ddof: int) -> Self:
         """Get standard deviation.
 
         Arguments:
             ddof: "Delta Degrees of Freedom": the divisor used in the calculation is N - ddof,
-                     where N represents the number of elements. By default ddof is 1.
+                     where N represents the number of elements.
 
         Returns:
             A new expression.
@@ -706,12 +706,12 @@ class Expr:
         """
         return self.__class__(lambda plx: self._to_compliant_expr(plx).std(ddof=ddof))
 
-    def var(self, *, ddof: int = 1) -> Self:
+    def var(self, *, ddof: int) -> Self:
         """Get variance.
 
         Arguments:
             ddof: "Delta Degrees of Freedom": the divisor used in the calculation is N - ddof,
-                     where N represents the number of elements. By default ddof is 1.
+                     where N represents the number of elements.
 
         Returns:
             A new expression.
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -655,12 +655,12 @@ class Expr:
         """
         return self.__class__(lambda plx: self._to_compliant_expr(plx).median())
 
-    def std(self, *, ddof: int) -> Self:
+    def std(self, *, ddof: int = 1) -> Self:
         """Get standard deviation.
 
         Arguments:
             ddof: "Delta Degrees of Freedom": the divisor used in the calculation is N - ddof,
-                     where N represents the number of elements.
+                     where N represents the number of elements. By default ddof is 1.
 
         Returns:
             A new expression.
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -706,6 +706,52 @@ class Expr:
         """
         return self.__class__(lambda plx: self._to_compliant_expr(plx).std(ddof=ddof))
 
+    def var(self, *, ddof: int = 1) -> Self:
+        """Get variance.
+
+        Returns:
+            A new expression.
+
+        Examples:
+            >>> import polars as pl
+            >>> import pandas as pd
+            >>> import pyarrow as pa
+            >>> import narwhals as nw
+            >>> from narwhals.typing import IntoFrameT
+            >>> df_pd = pd.DataFrame({"a": [20, 25, 60], "b": [1.5, 1, -1.4]})
+            >>> df_pl = pl.DataFrame({"a": [20, 25, 60], "b": [1.5, 1, -1.4]})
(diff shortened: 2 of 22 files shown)


narwhals/_arrow/expr.py
```diff
@@ -245,6 +245,9 @@ def n_unique(self: Self) -> Self:
     def std(self: Self, ddof: int = 1) -> Self:
         return reuse_series_implementation(self, "std", ddof=ddof, returns_scalar=True)
 
+    def var(self: Self, ddof: int = 1) -> Self:
```
FBruzzesi, line 248: In here and as well as all other compliant Series/Expr implementation we are now trying to avoid providing default values, could you adjust as:

```suggestion
    def var(self: Self, ddof: int) -> Self:
```

narwhals/expr.py
```diff
@@ -706,6 +706,52 @@ def std(self, *, ddof: int = 1) -> Self:
         """
         return self.__class__(lambda plx: self._to_compliant_expr(plx).std(ddof=ddof))
 
+    def var(self, *, ddof: int = 1) -> Self:
+        """Get variance.
+
+        Returns:
+            A new expression.
```
FBruzzesi, line 713: Could you add `Argument` section:

```suggestion

        Arguments:
            ddof: "Delta Degrees of Freedom": the divisor used in the calculation is N - ddof,
                where N represents the number of elements. By default ddof is 1.
        
        Returns:
            A new expression.
```

narwhals/expr.py
```diff
@@ -706,6 +706,52 @@ def std(self, *, ddof: int = 1) -> Self:
...
+            >>> import pyarrow as pa
+            >>> import narwhals as nw
+            >>> from narwhals.typing import IntoFrameT
+            >>> df_pd = pd.DataFrame({"a": [20, 25, 60], "b": [1.5, 1, -1.4]})
+            >>> df_pl = pl.DataFrame({"a": [20, 25, 60], "b": [1.5, 1, -1.4]})
+            >>> df_pa = pa.table({"a": [20, 25, 60], "b": [1.5, 1, -1.4]})
+
+            Let's define a dataframe-agnostic function:
+
+            >>> def my_library_agnostic_function(df_native: IntoFrameT) -> IntoFrameT:
```
FBruzzesi, line 727: We are trying to have more _meaningful_ function names; something like `agnostic_var` or `agnostic_variance` should do the trick :)

narwhals/series.py
```diff
@@ -1056,6 +1056,37 @@ def std(self, *, ddof: int = 1) -> Any:
...
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> from narwhals.typing import IntoSeries
+            >>> s = [1, 2, 3]
+            >>> s_pd = pd.Series(s)
+            >>> s_pl = pl.Series(s)
+
+            We define a library agnostic function:
+
+            >>> def my_library_agnostic_function(s_native: IntoSeries):
```
FBruzzesi, line 1077: Same as in `Expr`

tests/expr_and_series/var_test.py
```diff
@@ -0,0 +1,46 @@
+from __future__ import annotations
+
+import narwhals.stable.v1 as nw
+from tests.utils import Constructor
+from tests.utils import ConstructorEager
+from tests.utils import assert_equal_data
+
+data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
```
FBruzzesi, line 8: We should consider adding a test with nulls, both here and in `std`. This can be a follow up PR

FBruzzesi (commented):
Thanks @Dhanunjaya-Elluri 🙌🏼 This is a solid effort and also close to be ready for merge!
I left a few comments on some better habits we are trying to pick up ourselves as well when we develop new features - it should be nothing major

tests/group_by_test.py
```diff
@@ -136,6 +136,24 @@ def test_group_by_median(constructor: Constructor) -> None:
...
+    assert_equal_data(result, expected)
+
+
+def test_group_by_var(constructor: Constructor) -> None:
+    data = {"a": [1, 1, 2, 2], "b": [5, 4, 3, 2]}
+    result = (
+        nw.from_native(constructor(data)).group_by("a").agg(nw.col("b").var()).sort("a")
+    )
+    expected = {"a": [1, 2], "b": [0.5] * 2}
+    assert_equal_data(result, expected)
```
AlessandroMiola, line 154: a very tiny comment from me as well :)
we've recently added a test above (`test_group_by_depth_1_agg`) which tries to keep this "family" of tests together; perhaps, we might add the `var` test there (the `std` test is already present instead)

Dhanunjaya-Elluri: Didn't notice this. Was wondering where did test with `std` gone! Done changes

narwhals/_arrow/expr.py
Dhanunjaya-Elluri: Just one question here. Pandas pyarrow, and polars handles variance calculation using `ddof` differently in different versions. How do I handle it?

FBruzzesi: I imagine you are referring to the failing test. Not sure how `std` doesn't have such problem.
Or are there other instances of this happening?

Dhanunjaya-Elluri: Yes, I'm referring to the failing tests. 
For instance, to this `data = {"a": [1, 1, 1, 2], "b": [1, None, 2, 3]}`,
when I run `nox`, 
- Python 3.8 expects `{"a": [1, 2], "b": [0, None]}`, 
- Python >= 3.9 expects `{"a": [1, 2], "b": [0.5, None]}`.

Interestingly `std` does not have this issue!

FBruzzesi: I will take a deeper look later today! Thanks for pointing this out :)

FBruzzesi: It seems to be an issue with pyarrow backend and pandas v2.0.x and the overall dtype in group bys only?

Namely:

- In non-group by context, the result is correct:
    ```py
    pd.Series([1,None, 2]).convert_dtypes(dtype_backend="pyarrow").var()
    0.5
    ```

- if we change the dtype to float, the result is correct:

    ```py
    (
      pd.DataFrame({"a": [1,1,1, 2], "b": [1,None, 2., 100.2]})
      .convert_dtypes(dtype_backend="pyarrow")
      .groupby("a", as_index=False).var()
    )
       a     b
    0  1   0.5
    1  2  <NA>
    ```

- Yet this is incorrect:
    ```py
     (
       pd.DataFrame({"a": [1,1,1, 2], "b": [1,None, 2., 100]})
       .convert_dtypes(dtype_backend="pyarrow")
       .groupby("a", as_index=False).var()
     )
        a     b
    0  1     0
    1  2  <NA>
    ```

@MarcoGorelli I would say that this is ok to xfail for now by checking pandas version and backend WDYT?

We could re-use std implementation, which curiously enough, does not behave in the same way, yet it fails to maintain the backend dtype for such pandas version:

```py
pd.DataFrame({"a": [1,1,1, 2], "b": [1,None, 2., 100]}).convert_dtypes(dtype_backend="pyarrow").groupby("a", as_index=False).std().dtypes
a    int64[pyarrow]
b           float64
dtype: object
```

while it should be 

```py
a    int64[pyarrow]
b    float64[pyarrow]
dtype: object
```

MarcoGorelli: yeah i'd say ok to xfail

tests/group_by_test.py
```diff
@@ -118,8 +119,22 @@ def test_group_by_depth_1_agg(
...
+            pytest.mark.xfail(
+                reason="Known issue with variance calculation in pandas 2.0.x with pyarrow backend in groupby operations"
+            )
+        )
     data = {"a": [1, 1, 1, 2], "b": [1, None, 2, 3]}
-    expr = getattr(nw.col("b"), attr)()
+    if attr in ["var", "std"]:
+        expr = getattr(nw.col("b"), attr)(ddof=1)
+    else:
+        expr = getattr(nw.col("b"), attr)()
```
MarcoGorelli: this doesn't need changing, `ddof=1` is already the public-facing default, it's only in the internal code that we tend to not set defaults
```suggestion
    if attr in ["var", "std"]:
        expr = getattr(nw.col("b"), attr)(ddof=1)
    else:
        expr = getattr(nw.col("b"), attr)()
```

MarcoGorelli (approved):
thanks @Dhanunjaya-Elluri , and @FBruzzesi @AlessandroMiola for reviewing

only the internal functions needed the defaults removing, not the public ones

Issue #1606, opened 2024-12-17
[Bug]: `group_by` context ignores expr arguments


FBruzzesi:
### Describe the bug

For expressions that accept arguments, these are not passed along in `group_by` context (e.g. `ddof` in `std`).

### Steps or code to reproduce the bug

```py
import narwhals as nw
import pandas as pd
import polars as pl

data = {"a": [1, 1, 1, 1, 1, 2, 2, 2, 2, 2], "b": list(range(10))}

nw.from_native(pl.DataFrame(data)).group_by("a").agg(nw.col("b").std(ddof=2)).to_native()
shape: (2, 2)
┌─────┬──────────┐
│ a   ┆ b        │
│ --- ┆ ---      │
│ i64 ┆ f64      │
╞═════╪══════════╡
│ 2   ┆ 1.825742 │
│ 1   ┆ 1.825742 │
└─────┴──────────┘

nw.from_native(pd.DataFrame(data)).group_by("a").agg(nw.col("b").std(ddof=2)).to_native()
   a         b
0  1  1.581139
1  2  1.581139

# which is the same as ddof=1

nw.from_native(pd.DataFrame(data)).group_by("a").agg(nw.col("b").std(ddof=1)).to_native()
   a         b
0  1  1.581139
1  2  1.581139
```

### Expected results

Arguments to be passed along or raise

### Actual results

Arguments get ignored

### Please run narwhals.show_version() and enter the output below.

```shell
System:
    python: 3.12.2 (main, Feb 25 2024, 16:35:05) [GCC 11.4.0]

Python dependencies:
     narwhals: 1.18.4
       pandas: 2.2.3
       polars: 1.17.1
         cudf: 
        modin: 0.32.0
      pyarrow: 18.1.0
        numpy: 2.1.3
```

MarcoGorelli:
thanks for reporting / spotting, this isn't great - do you have a solution in mind, or shall i take a look?

FBruzzesi:
If you have time go ahead. I have scattered slots to take a look or I can tonight 🙃

FBruzzesi:
@MarcoGorelli after a bit of investigation, I was not able to find a way to pass args and kwargs to a function in the pandas group by notation we use for the fast path (namely `output_name = (input_name, aggregation)`).

[...]

PR #1614, opened 2024-12-18
chore: register kwargs for `CompliantExpr`

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

Opening as draft/RFC since the API is not the shiniest we could come up with. I would be more happy to just have to deal with kwargs.
The reason for having both is to reuse `reuse_series_implementation` function, which however we always use with kwargs (happy to refactor this to make it mandatory).

This feature would unblock #1606 (partially) and already address an issue from #1565 and `reverse=True` leading to incorrect result.

### DIFF
--- a/narwhals/_arrow/group_by.py
+++ b/narwhals/_arrow/group_by.py
@@ -169,9 +169,7 @@ def agg_arrow(
         function_name = remove_prefix(expr._function_name, "col->")
 
         if function_name in {"std", "var"}:
-            option = pc.VarianceOptions(
-                ddof=expr._kwargs.get("ddof", 1) if expr._kwargs else 1
-            )
+            option = pc.VarianceOptions(ddof=expr._kwargs.get("ddof", 1))
         elif function_name in {"len", "n_unique"}:
             option = pc.CountOptions(mode="all")
         elif function_name == "count":
--- a/narwhals/_arrow/group_by.py
+++ b/narwhals/_arrow/group_by.py
@@ -137,85 +137,78 @@ def agg_arrow(
             all_simple_aggs = False
             break
 
-    if all_simple_aggs:
-        # Mapping from output name to
-        # (aggregation_args, pyarrow_output_name)  # noqa: ERA001
-        simple_aggregations: dict[str, tuple[tuple[Any, ...], str]] = {}
-        for expr in exprs:
-            if expr._depth == 0:
-                # e.g. agg(nw.len()) # noqa: ERA001
-                if (
-                    expr._output_names is None or expr._function_name != "len"
-                ):  # pragma: no cover
-                    msg = "Safety assertion failed, please report a bug to https://github.com/narwhals-dev/narwhals/issues"
-                    raise AssertionError(msg)
-                simple_aggregations[expr._output_names[0]] = (
-                    (keys[0], "count", pc.CountOptions(mode="all")),
-                    f"{keys[0]}_count",
-                )
-                continue
+    if not all_simple_aggs:
+        msg = (
+            "Non-trivial complex aggregation found.\n\n"
+            "Hint: you were probably trying to apply a non-elementary aggregation with a "
+            "pyarrow table.\n"
+            "Please rewrite your query such that group-by aggregations "
+            "are elementary. For example, instead of:\n\n"
+            "    df.group_by('a').agg(nw.col('b').round(2).mean())\n\n"
+            "use:\n\n"
+            "    df.with_columns(nw.col('b').round(2)).group_by('a').agg(nw.col('b').mean())\n\n"
+        )
+        raise ValueError(msg)
 
-            # e.g. agg(nw.mean('a')) # noqa: ERA001
+    aggs: list[tuple[str, str, pc.FunctionOptions | None]] = []
+    expected_pyarrow_column_names: list[str] = keys.copy()
+    new_column_names: list[str] = keys.copy()
+
+    for expr in exprs:
+        if expr._depth == 0:
+            # e.g. agg(nw.len()) # noqa: ERA001
             if (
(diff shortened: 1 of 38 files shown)


narwhals/_pandas_like/expr.py
```diff
@@ -426,15 +432,27 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
-                    self._root_names
-                ].transform(
-                    CUMULATIVE_FUNCTIONS_TO_PANDAS_EQUIVALENT[self._function_name]
-                )
+                res_native = getattr(
+                    df._native_frame.groupby(list(keys), as_index=False)[
+                        self._root_names
+                    ],
+                    CUMULATIVE_FUNCTIONS_TO_PANDAS_EQUIVALENT[self._function_name],
+                )(skipna=True)
```
FBruzzesi, line 454: I know it is the default, but considering all the versions and the other pandas-like backends, I rather being explicit about it.

narwhals/_pandas_like/expr.py
```diff
@@ -426,15 +432,27 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
                 if self._function_name == "col->cum_count":
                     plx = self.__narwhals_namespace__()
                     df = df.with_columns(~plx.col(*self._root_names).is_null())
 
-                res_native = df._native_frame.groupby(list(keys), as_index=False)[
-                    self._root_names
-                ].transform(
-                    CUMULATIVE_FUNCTIONS_TO_PANDAS_EQUIVALENT[self._function_name]
-                )
+                res_native = getattr(
```
FBruzzesi, line 449: Not a strong preference to use getattr, I realize now that also via transform it is easy to pass kwargs

MarcoGorelli:
thanks! i also tried https://github.com/narwhals-dev/narwhals/pull/1620

I like how this solution is more generic though

FBruzzesi:
> thanks! i also tried #1620
> 
> I like how this solution is more generic though

Maybe it's a good topic for tomorrow community call or we can also jump on a quick brainstorming later today.

I like the flexibility and generality of this approach better - we can (and maybe should) be careful to not bring _too_ many arguments along especially if these are other series/dataframes, hence is some cases we might decide to not register the arguments on purpose. Beside that I would opt for this one.

> The reason for having both is to reuse `reuse_series_implementation `function, which however we always use with kwargs (happy to refactor this to make it mandatory).

I will address this in the meanwhile, it should not be much effort

MarcoGorelli:
yup, agree. and when we get to spark / duckdb etc which can be way more flexible in what they can handle, it'll be better if we have the infra to do something more generic - i'll close the other pr then

FBruzzesi:
@MarcoGorelli this is a new error to me:

```
Using Python 3.13.1 environment at: /opt/hostedtoolcache/Python/3.13.1/x64
Uninstalled 1 package in 5ms
 - polars==1.17.1
WARNING: Requirement '*.whl' looks like a filename, but the file does not exist
ERROR: *.whl is not a valid wheel filename.
Error: Process completed with exit code 1.
```

FBruzzesi:
In terms of completeness:

- `kwargs` is added to all `CompliantExpr`s
- For pyarrow backend #1606 is addressed
- For pandas, dask and pyspark it is not:
  - pandas: this is potentially the most problematic. I have a hacky way to do it (I will comment in #1606)
  - dask: needs investigation if it is possible with a custom `dd.Aggregation` that performs `std` and `var` and if that would result in worse performances
  - pyspark: this should be fairly easy to achieve via expressions instead of `get_spark_function`

MarcoGorelli:
awesome, thanks!

> pandas: this is potentially the most problematic. I have a hacky way to do it (I will comment in https://github.com/narwhals-dev/narwhals/issues/1606)

maybe it's ok, for now, to just send it down the "complex path" if ddof!=1?

FBruzzesi:
> maybe it's ok, for now, to just send it down the "complex path" if ddof!=1?

Yes I think that's better than providing the wrong result for sure

narwhals/_dask/expr.py
FBruzzesi, line 1: Additional changes here were to make the arguments of custom `func` created in many methods to match the signature of the method itself, so that it would be possible to pass them as kwargs to `_from_call`

narwhals/_arrow/group_by.py
```diff
@@ -177,9 +167,17 @@ def agg_arrow(
...
+        if function_name in {"std", "var"}:
+            option = pc.VarianceOptions(ddof=expr._kwargs.get("ddof", 1))
+        elif function_name in {"len", "n_unique"}:
+            option = pc.CountOptions(mode="all")
+        elif function_name == "count":
+            option = pc.CountOptions(mode="only_valid")
+        else:
+            option = None
+
+        function_name = POLARS_TO_ARROW_AGGREGATIONS[function_name]
```
MarcoGorelli, line 180: nice!

narwhals/_dask/expr.py
```diff
@@ -433,20 +438,26 @@ def max(self) -> Self:
...
-        return self._from_call(
+        expr = self._from_call(
             lambda _input, ddof: _input.std(ddof=ddof),
             "std",
-            ddof,
+            ddof=ddof,
             returns_scalar=True,
         )
+        if ddof != 1:
+            expr._depth += 1
```
MarcoGorelli, line 448: 😎

MarcoGorelli (approved):
mate this is flippin' awesome, well done

Issue #1500, opened 2024-12-05
docs: Make docstring examples function names more descriptive, use type hints, use from_native/to_native instaed of narwhalify


MarcoGorelli:
Use descriptive names in docstring examples

Many docstrings have examples such as `def func`, which are non-descriptive

For example:

```python
>>> @nw.narwhalify
... def func(df):
...     return df.select(threshold=nw.col("foo") * 2)
```

We should do two things here:
- use a descriptive function name which starts with `agnostic_`. For example, this one
  could be called `agnostic_select_threshold`
- rewrite it to use `from_native` / `to_native` instead of `narwhalify`
- add type hints: instead of just `(df):`, we should write `(df_native: IntoFrameT) -> IntoFrameT`

So, the docstring example above should become:
```python
>>> def agnostic_select_threshold(df_native: IntoFrameT) -> IntoFrameT:
...     df = nw.from_native(df_native)
...     return df.select(threshold=nw.col("foo") * 2).to_native()
```

Some notes on  the type hints:
- If you're using `IntoFrameT` make sure to include `from narwhals.typing import IntoFrameT`
- If a function is only supported for eager execution (i.e. it's only defined on `narwhals.DataFrame`, not `narwhals.LazyFrame`), then use `IntoDataFrameT` instead of `IntoFrameT`
- make sure the return value of the function is correct. For example, if it returns a Series, use `IntoSeriesT`

---

Before submitting a pull request, please make sure that:
1. linting / formatting checks all pass (`pre-commit run --all-files`)
2. doctests pass (`pytest narwhals --doctest-modules`)

Please choose 1-3 docstrings to work on at a time, and comment here which ones you're working on

Arya-AD:
Hello @MarcoGorelli , 
I think I can handle this issue.

MarcoGorelli:
Awesome thanks! Feel free to let us know if anything trips you up

Arya-AD:
Sure, Thank you!

PR #1625, opened 2024-12-19
feat: add is_nan expression & series method

### LINKED ISSUE #1583: feat: is_nan
Looks like this would be useful to Tubular: https://github.com/lvgig/tubular/pull/349#discussion_r1884006320

There would need to be a caveat in the docstring about how pandas doesn't distinguish between nan and null but other libraries typically do

This requires some attention to check that we get such nan vs null details correct:

- for Polars, there is already is_nan
- for PyArrow, there is https://arrow.apache.org/docs/python/generated/pyarrow.compute.is_nan.html
- for pandas, we need to be careful. We don't want `is_nan` to pick up `NaT` / `None` / anything else which is used for missing values, but instead, specifically, we only want `nan`. So, we should check
  - that the dtype is `Float`
  - that `s != s`

camriddell:
- add support for pandas, arrow, dask
- add to docs
- add tests

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1583 
- Closes #1583 

## Checklist

- [x] Tests added
- [x] Documented the changes

The `is_nan` implementation for pandas and dask currently return True for the following conditions:
1. The backend datatype is float64
2. The NaN value is not equal to itself

This leaves space for false rejections when there is a NaN value inside of a Series with 'object' dtype, though I am uncertain if we care about this edgecase.

Additionally, I was uncertain on how to structure the tests with regards to having NaN values inside of nullable datatypes. I currently dynamically change the expected result but was considering creating separate tests for data that are backed by a nullable datatype vs a non-nullable datatype so some guidance here would be appreciated.


docs/api-reference/expr.md
```diff
@@ -33,6 +33,7 @@
         - is_in
         - is_last_distinct
         - is_null
+        - is_nan
```
MarcoGorelli, line 36: can we keep alphabetical ordering?

narwhals/expr.py
```diff
@@ -1921,6 +1921,78 @@ def is_null(self) -> Self:
...
+            >>> df_pl = pl.DataFrame(
+            ...     {"a": [2, 4, None, 3, 5], "b": [2.0, 4.0, float("nan"), 3.0, 5.0]}
+            ... )
+            >>> df_pa = pa.table(
+            ...     {"a": [2, 4, None, 3, 5], "b": [2.0, 4.0, float("nan"), 3.0, 5.0]}
+            ... )
+
+            Let's define a dataframe-agnostic function:
+
+            >>> def my_library_agnostic_function(df_native: IntoFrameT) -> IntoFrameT:
```
MarcoGorelli, line 1952: https://github.com/narwhals-dev/narwhals/issues/1500 is still in progress, but as we're adding a new function, shall we follow the conventions there?

narwhals/expr.py
```diff
@@ -1921,6 +1921,78 @@ def is_null(self) -> Self:
...
+            ...     df = nw.from_native(df_native)
+            ...     return df.with_columns(
+            ...         a_is_nan=nw.col("a").is_nan(), b_is_nan=nw.col("b").is_nan()
+            ...     ).to_native()
```
MarcoGorelli, line 1956: nice choice of example!

narwhals/series.py
```diff
@@ -1915,6 +1915,46 @@ def is_null(self) -> Self:
...
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> from narwhals.typing import IntoSeriesT
+            >>> s = [1.0, 2.0, float("nan")]
+            >>> s_pd = pd.Series(s, dtype="float64")
+            >>> s_pl = pl.Series(s)
+
+            We define a dataframe-agnostic function:
+
+            >>> def my_library_agnostic_function(s_native: IntoSeriesT) -> IntoSeriesT:
```
MarcoGorelli, line 1936: same

narwhals/series.py
```diff
@@ -1915,6 +1915,46 @@ def is_null(self) -> Self:
...
+            ...     s = nw.from_native(s_native, series_only=True)
+            ...     return s.is_nan().to_native()
+
+            We can then pass either pandas or Polars to `func`:
```
MarcoGorelli, line 1940: it's not called `func` though 😉  (this might be typo elsewhere in the code tbh 😳 )

could we add pyarrow too?

MarcoGorelli (commented):
nice, thanks @camriddell !

in answer to your question on the issue, i think `np.isnan` might be fine too, I'd just check that it doesn't inadvertently end up copying data just to do the operation (whereas `s != s` shouldn't be too expensive)

tests look good to me

camriddell:
@MarcoGorelli I realized Polars raises an InvalidOperationError for checking `is_nan` for non supported dtypes. Is this a behavior we should enforce on all other backends as well?

```python
>>> import polars as pl
>>> pl.Series(["a", "b"]).is_nan()
  File "/home/cameron/repos/opensource/narwhals-dev/.venv/lib/python3.12/site-packages/polars/series/utils.py", line 106, in wrapper
    return s.to_frame().select_seq(f(*args, **kwargs)).to_series()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cameron/repos/opensource/narwhals-dev/.venv/lib/python3.12/site-packages/polars/dataframe/frame.py", line 9138, in select_seq
    return self.lazy().select_seq(*exprs, **named_exprs).collect(_eager=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cameron/repos/opensource/narwhals-dev/.venv/lib/python3.12/site-packages/polars/lazyframe/frame.py", line 2029, in collect
    return wrap_df(ldf.collect(callback))
                   ^^^^^^^^^^^^^^^^^^^^^
polars.exceptions.InvalidOperationError: `is_nan` operation not supported for dtype `str`
```

Should we also raise an exception in these cases? Right now we are returning a Series of False values (though I do need to update this to respect existing Nulls).

---
Some additional testing code.

```python
import polars as pl
from datetime import date

data = [[0], [1.1], ["a"], [date.today()], [[0]], [["a"]], [False]]
cant_check = []
for d in data:
    s = pl.Series(d)
    try:
        s.is_nan()
    except pl.exceptions.InvalidOperationError:
        print(f"\N{cross mark} {s.dtype = }")
    else:
        print(f"\N{white heavy check mark} {s.dtype = }")

# ✅ s.dtype = Int64
# ✅ s.dtype = Float64
# ❌ s.dtype = String
# ❌ s.dtype = Date
# ❌ s.dtype = List(Int64)
# ❌ s.dtype = List(String)
# ❌ s.dtype = Boolean
```

MarcoGorelli:
interesting, thanks - yeah that's probably a good idea. then we avoid the risk that people who are too used to pandas accidentally use `is_nan` in non-float columns in places where they really wanted `is_null`

camriddell:
@MarcoGorelli thoughts on the apparent inconsistency in Polars `is_nan` handling? In Polars, if the underlying dtype is Int64, then a `is_nan` converts the Null value → False. However if the dtypes is Float64 then the Null is preserved.


```python
import pandas as pd
import polars as pl
import pyarrow as pa
import narwhals as nw
data = {"a": [2, None, 4, 5, 6], "b": [2.0, None, float("nan"), 5.0, 6.0]}
df_pd = pd.DataFrame(data).astype({"a": "Int64"})
df_pl = pl.DataFrame(data)
df_pa = pa.table(data)

def agnostic_is_nan_columns(df_native: IntoFrameT) -> IntoFrameT:
    df = nw.from_native(df_native)
    return df.with_columns(
        a_is_nan=nw.col("a").is_nan(), b_is_nan=nw.col("b").is_nan()
    ).to_native()


print(agnostic_is_nan_columns(df_pd))
#       a    b  a_is_nan  b_is_nan
# 0     2  2.0     False     False
# 1  <NA>  NaN      <NA>      True
# 2     4  NaN     False      True
# 3     5  5.0     False     False
# 4     6  6.0     False     False

print(agnostic_is_nan_columns(df_pl))
# shape: (5, 4)
# ┌──────┬──────┬──────────┬──────────┐
# │ a    ┆ b    ┆ a_is_nan ┆ b_is_nan │
# │ ---  ┆ ---  ┆ ---      ┆ ---      │
# │ i64  ┆ f64  ┆ bool     ┆ bool     │
# ╞══════╪══════╪══════════╪══════════╡
# │ 2    ┆ 2.0  ┆ false    ┆ false    │
# │ null ┆ null ┆ false    ┆ null     │
# │ 4    ┆ NaN  ┆ false    ┆ true     │
# │ 5    ┆ 5.0  ┆ false    ┆ false    │
# │ 6    ┆ 6.0  ┆ false    ┆ false    │
# └──────┴──────┴──────────┴──────────┘

print(agnostic_is_nan_columns(df_pa))
# pyarrow.Table
# a: int64
# b: double
# a_is_nan: bool
# b_is_nan: bool
# ----
# a: [[2,null,4,5,6]]
# b: [[2,null,nan,5,6]]
# a_is_nan: [[false,null,false,false,false]]
# b_is_nan: [[false,null,true,false,false]]
```

narwhals/_pandas_like/series.py
```diff
@@ -654,6 +653,13 @@ def is_null(self) -> PandasLikeSeries:
         ser = self._native_series
         return self._from_native_series(ser.isna())
 
+    def is_nan(self) -> PandasLikeSeries:
+        ser = self._native_series
+        if self.dtype.is_numeric():
+            return self._from_native_series(ser != ser)  # noqa: PLR0124
+        msg = f"`is_nan` is not supported for dtype {self.dtype}"
```
camriddell, line 660: Should this error message be extended to include a suggestion? 

"`is_nan` is not supported for dtype {self.dtype}, did you mean to use `is_null`?

MarcoGorelli:
thanks for spotting this! I _think_ this looks like a bug in Polars, shall we report it there? I think either `is_nan` should only be supported on float columns (my preference), or the null preservation should be consistent

narwhals/_pandas_like/series.py
MarcoGorelli: lots of love for kind error messages 💚

camriddell:
Yep- it appears that this bug was noticed in this Polars PR: https://github.com/pola-rs/polars/pull/15889 but not fixed until a broader change (apparently all float-specific operations failed to propagate nulls) was made last week for the 1.18 release: https://github.com/pola-rs/polars/pull/20386.

---

I think only allowing `.is_nan` on the float dtype makes sense here, this will ensure operations are consistent across all backends without us creating a version aware `is_nan` implementation for Polars to smooth out the API differences. I'll go ahead an implement this approach and hopefully wrap this on up!

narwhals/_dask/expr.py
```diff
@@ -702,6 +703,20 @@ def is_null(self: Self) -> Self:
             returns_scalar=False,
         )
 
+    def is_nan(self: Self) -> Self:
+        def func(_input: dask_expr.Series) -> dask_expr.Series:
+            dtype = native_to_narwhals_dtype(_input, self._version, self._implementation)
+            if dtype.is_numeric():
+                return _input != _input  # noqa: PLR0124
+            msg = f"`.is_nan` only supported for numeric dtypes and not {dtype}, did you mean `.is_null`?"
```
MarcoGorelli, line 711: nice err msg!

narwhals/expr.py
```diff
@@ -1925,6 +1925,72 @@ def is_null(self) -> Self:
...
         return self.__class__(lambda plx: self._to_compliant_expr(plx).is_null())
 
+    def is_nan(self) -> Self:
+        """Returns a boolean Series indicating which values are NaN.
+
+        Returns:
+            A new expression.
+
+        Raises:
+            narwhals.InvalidOperationError for non-Float64 dtypes.
```
MarcoGorelli, line 1935: is this outdated?

narwhals/expr.py
```diff
@@ -1925,6 +1925,72 @@ def is_null(self) -> Self:
...
+
+        Notes:
+            pandas, Polars and PyArrow handle null values differently. Polars and PyArrow
+            distinguish between NaN and Null, whereas pandas doesn't.
```
MarcoGorelli, line 1939: we're now linking to the page
```
            pandas handles null values differently from Polars and PyArrow.
            See [null_handling](../pandas_like_concepts/null_handling.md/)
            for reference.
```
shall we do that here (and anywhere else if we forgot any places)?

narwhals/series.py
```diff
@@ -2032,6 +2032,61 @@ def is_null(self) -> Self:
...
         return self._from_compliant_series(self._compliant_series.is_null())
 
+    def is_nan(self) -> Self:
+        """Returns a boolean Series indicating which values are NaN.
+
+        Returns:
+            A new Series.
+
+        Raises:
+            narwhals.InvalidOperationError for non-Float64 dtypes.
```
MarcoGorelli, line 2042: same

narwhals/series.py
```diff
@@ -2032,6 +2032,61 @@ def is_null(self) -> Self:
...
+
+        Notes:
+            pandas and Polars handle NaN values differently. Polars distinguishes
```
MarcoGorelli, line 2045: same

MarcoGorelli (commented):
thanks Cam, awesome stuff! just some minor things

camriddell:
There was a failure for the Modin constructor, which seems to improperly cast from a pyarrow-nullable backend to a numpy backend. Should https://github.com/narwhals-dev/narwhals/blob/7c8c6fa1d3e7a4c31b8bf7577ca80ab9322a1c95/narwhals/_pandas_like/utils.py#L483-L498 also have a clause for if the implementation is modin as well? Perhaps something like this:

```python
def get_dtype_backend(dtype: Any, implementation: Implementation) -> str:
    if any(implementation is  impl for impl in [Implementation.PANDAS, Implementation.MODIN]):
        import pandas as pd

        if hasattr(pd, "ArrowDtype") and isinstance(dtype, pd.ArrowDtype):
            return "pyarrow-nullable"

    if implementation is Implementation.PANDAS:
        try:
            if isinstance(dtype, pd.core.dtypes.dtypes.BaseMaskedDtype):
                return "pandas-nullable"
        except AttributeError:  # pragma: no cover
            # defensive check for old pandas versions
            pass
        return "numpy"
    else:  # pragma: no cover
        return "numpy"
```

Should there be any cases here for CuDF as well?

MarcoGorelli:
well-spotted, thanks! we should probably do the same for modin then

cuDF doesn't seem to have these different dtype backends (it's always nullable arrow dtypes)

camriddell:
> well-spotted, thanks! we should probably do the same for modin then
> 
> cuDF doesn't seem to have these different dtype backends (it's always nullable arrow dtypes)

Should the patch be a part of this PR or a new one since it is a semi-related fix?

MarcoGorelli:
either is fine, whichever workflow works best for you

narwhals/_dask/expr.py
```diff
@@ -706,6 +707,20 @@ def is_null(self: Self) -> Self:
...
+            raise InvalidOperationError(msg)
+
+        return self._from_call(
+            func,
+            "is_null",
+            returns_scalar=False,
```
MarcoGorelli, line 721:
```suggestion
            returns_scalar=self._returns_scalar,
```

MarcoGorelli (approved):
thanks @camriddell !

PR #1627, opened 2024-12-20
feat: Add support for `.shift(n).over('col')` for pandas-like DataFrames

### LINKED ISSUE #1609: enh: support `nw.col('a').shift(n).over('b')`
We now support cumulative operations in `over` for pandas and Polars

We should do the same for `shift`, as pandas also supports that. One challenge will be figuring out how to pass the argument down, but it should be doable

cc @ClaudioSalvatoreArcidiacono in case this interests you

ClaudioSalvatoreArcidiacono:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1609 

## Checklist

- [x] Tests added
- [ ] Documented the changes

See #1609


MarcoGorelli:
nice, thanks @ClaudioSalvatoreArcidiacono !

I think @FBruzzesi 's work in https://github.com/narwhals-dev/narwhals/pull/1614 might make this easier, perhaps we should wait to get that in, then we can update here?

FBruzzesi:
Thanks @ClaudioSalvatoreArcidiacono !

I was about to echo what @MarcoGorelli  just mentioned! Instead of an ad-hoc change, I think waiting for #1614 would make it easier. Possibly it will be ready by end of today.

ClaudioSalvatoreArcidiacono:
Definetly! Nice PR @FBruzzesi

MarcoGorelli:
that PR's merged, if you fancy updating @ClaudioSalvatoreArcidiacono I think we'd love to get this in!

FBruzzesi:
> that PR's merged, if you fancy updating @ClaudioSalvatoreArcidiacono I think we'd love to get this in!

Just a headsup that the PR was shipped with a bug, namely it does not track kwargs in expr composition. You would probably want to wait for #1645 which contains a fix

narwhals/_pandas_like/expr.py
```diff
@@ -443,12 +444,20 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
                     plx = self.__narwhals_namespace__()
                     df = df.with_columns(~plx.col(*self._root_names).is_null())
 
+                if self._function_name == "col->shift":
+                    kwargs = {
+                        "periods": (self._kwargs.get("n", 1) if self._kwargs else 1)
+                    }
```
FBruzzesi, line 450: After #1645 `_kwargs` should always be a dictionary

```suggestion
                    kwargs = {"periods": self._kwargs.get("n", 1)}
```

FBruzzesi (commented):
Pretty happy with this! I left only a tiny comment before merging!

narwhals/_pandas_like/expr.py
ClaudioSalvatoreArcidiacono: Thanks for your suggestion!

FBruzzesi (approved):
Thanks a ton @ClaudioSalvatoreArcidiacono 👌 

DogsHighFiveGIF

PR #1633, opened 2024-12-20
feat: add some `SparkLikeLazyFrame` methods

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Should we start tracking pyspark methods?

## Checklist

- [x] Tests added
- [ ] Documented the changes

It adds:

- `drop_nulls`
- `rename`
- `unique`
- `join`

And fixes `sort`

## TODO

There is _one_ test failing, because of how I implemented join to coalesce columns. The test is `test_left_join_overlapping_column` which tries to map a column called `d` to one called `antananarivo`, which already exist, and would try to rename such one to `antananarivo_right`. Yet fails to do so


narwhals/_spark_like/dataframe.py
```diff
@@ -173,16 +176,87 @@ def sort(
 
         flat_by = flatten([*flatten([by]), *more_by])
         if isinstance(descending, bool):
-            descending = [descending]
+            descending = [descending] * len(flat_by)
```
FBruzzesi, line 179: This took me some mental sanity to figure out. I was like "does pyspark do not know how sorting works?" for a good 5 minutes

tests/spark_like_test.py
```diff
@@ -441,3 +441,393 @@ def test_group_by_multiple_keys(pyspark_constructor: Constructor) -> None:
...
+    result = df.unique(maintain_order=True)
+    assert_equal_data(result, data)
+
+
+def test_inner_join_two_keys(pyspark_constructor: Constructor) -> None:
+    data = {
+        "antananarivo": [1, 3, 2],
+        "bob": [4, 4, 6],
+        "zorro": [7.0, 8, 9],
+        "idx": [0, 1, 2],
```
FBruzzesi, line 550: Something fishing with using `index`, I renamed it here and everywhere else. I was not confident in changing the constructor

narwhals/_spark_like/dataframe.py
```diff
@@ -173,16 +176,87 @@ def sort(
...
 
         sort_cols = [sort_f(col) for col, sort_f in zip(flat_by, sort_funcs)]
         return self._from_native_frame(self._native_frame.sort(*sort_cols))
+
+    def with_row_index(self: Self, name: str) -> Self:
+        import pyspark.sql.functions as F  # noqa: N812
+
+        return self._from_native_frame(
+            self._native_frame.withColumn(name, F.monotonically_increasing_id())
+        )
```
MarcoGorelli, line 198: should we keep this one out for now? i think it would be good to aim to have the lazy layer well-defined, and initially at least to leave out anything that's row-order-dependent

narwhals/_spark_like/dataframe.py
```diff
@@ -173,16 +176,87 @@ def sort(
...
+            msg = "`LazyFrame.unique` with PySpark backend only supports `keep='any'`."
+            raise ValueError(msg)
+
+        if maintain_order is True:
+            warnings.warn(
+                "Argument `maintain_order=True` is unused in `LazyFrame.unique` with "
+                "PySpark backend",
+                UserWarning,
+                stacklevel=find_stacklevel(),
+            )
```
MarcoGorelli, line 225: we should probably raise here, if it's unused?

MarcoGorelli (commented):
cool, thanks!

for tests, maybe after this one we should start adding it to `constructor` and xfailing things?

tests/spark_like_test.py
EdAbati: Background:
In the `_pyspark_constructor_with_session` we make a pyspark dataframe from a pandas dataframe.
When we load the `pandas.DataFrame` in spark the row order may not be maintained.
That's why we do `.reset_index` on the pandas DF and re-sort in spark based on that index (`.sortBy("index").drop("index")`)

If we don't do this, `test_head` could fail, i.e. the "first" rows in the dataframe are always different.

Fix:

Since the default column name in `reset_index` is `"index"`, we can use a special column name that we don't use in the tests in `.reset_index(names=...)` in the constructor. This should solve this 🤞

Long term question:

Can we think of a better way to test `.head` for lazyframes?

narwhals/_spark_like/dataframe.py
EdAbati: `monotonically_increasing_id` will be performant but not deterministic. Do we need `with_row_index` to be deterministic?
I tend to agree with "let's leave this out for now"

tests/spark_like_test.py
FBruzzesi: Thanks for the explanation! When we will move the constructor to the main fixture, we can change the index name so that those tests will not be affected 👌

FBruzzesi:
> for tests, maybe after this one we should start adding it to `constructor` and xfailing things?

We have _a lot_ of expressions that would fail 🙃 I would say to wait a little bit more for now

narwhals/_spark_like/dataframe.py
FBruzzesi: In dask `maintain_order` is unused as well, without any warnings 🤔

FBruzzesi:
I am not sure if the failing test should be considered an edge case, or it is worth to create a temporary name mapping layer/step to have it run successfully 🤔 It would definitly add a fair bit of complexity

MarcoGorelli:
thanks! i'll take a look, but i think we should address the join test

MarcoGorelli:
thanks! i think a loud error like
```
E               pyspark.errors.exceptions.captured.AnalysisException: [COLUMN_ALREADY_EXISTS] The column `antananarivo_right` already exists. Consider to choose another name or rename the existing column.
```
is fine, i think it's ok to assert that for pyspark it raises

narwhals/_spark_like/dataframe.py
```diff
@@ -169,16 +170,75 @@ def sort(
...
+        rename_mapping = {
+            **dict(zip(right_on or [], left_on or [])),
+            **{
+                colname: f"{colname}{suffix}" if colname in left_columns else colname
+                for colname in list(set(right_columns).difference(set(right_on or [])))
+            },
+        }
+        other = other_native.select(
+            [F.col(old).alias(new) for old, new in rename_mapping.items()]
+        )
```
FBruzzesi, line 243: Renaming via select seems to deal with both pyspark 3.4 (3.3 does not support `withColumnsRenamed`) and the issue of renaming to a column which was already existing

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1644, opened 2024-12-22
feat: `ArrowDataFrame.explode`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

I will leave this as draft until we decide how to move forward.

To summarize the discussion(s) in #1542 :

- pyarrow native methods ignore nulls and empty list in explode
- the workaround here is to have a "fast_path" for when nulls or empty lists are not present, and a dedicated path for when they are
- the issue is that we enter python world to create the index via one `.to_pylist()` call


dangotbanned:
@FBruzzesi I feel like this shouldn't have got lost!

`ArrowDataFrame.explode` is one of 3 remaining implementations we need

https://github.com/narwhals-dev/narwhals/blob/2bcc6bbfe188889c15ae7eab544f5ba5e91ba695/narwhals/_arrow/dataframe.py#L350

https://github.com/narwhals-dev/narwhals/blob/2bcc6bbfe188889c15ae7eab544f5ba5e91ba695/narwhals/_arrow/dataframe.py#L466

I might add a PR for `ArrowDataFrame.clone` - since it can just utilize arrow data being immutable

https://github.com/narwhals-dev/narwhals/blob/2bcc6bbfe188889c15ae7eab544f5ba5e91ba695/narwhals/_arrow/dataframe.py#L669

FBruzzesi:
> I feel like this shouldn't have got lost!

Thanks @dangotbanned ♥️ The main concern was a conversion to python object: `filled_counts.to_pylist()` in:

```py
    if fast_path:
            indices = pc.list_parent_indices(native_frame[to_explode[0]])
            flatten_func = pc.list_flatten

        else:
            filled_counts = pc.max_element_wise(counts, 1, skip_nulls=True)
            indices = pa.array(
                [
                    i
                    for i, count in enumerate(filled_counts.to_pylist())
                    for _ in range(count)
                ]
            )
```

dangotbanned:
- [ ] https://github.com/pola-rs/polars/issues/17664

@FBruzzesi @MarcoGorelli 

It seems like `polars` wants to make a breaking change in the next major version - resulting in the behavior as `pyarrow`.

dangotbanned:
@FBruzzesi I'm coming back to this after my recent fiddling with `pyarrow` 😄
There may still be a way to match `polars` - while staying in `pyarrow`-land - but ~~I'm only just starting to look now 🤞~~

**Updated**: Yep, it works! 🥳 (#3347)

---

Regarding: (https://github.com/narwhals-dev/narwhals/pull/1644#issuecomment-2751243106)

3 weeks ago on that issue (https://github.com/pola-rs/polars/issues/17664#issuecomment-3540984248)
> > Note that the option got added to resolve this in [#25289](https://github.com/pola-rs/polars/pull/25289).

- https://github.com/pola-rs/polars/pull/25289


So now [`polars.DataFrame.explode`](https://docs.pola.rs/api/python/dev/reference/dataframe/api/polars.DataFrame.explode.html) has this signature:

```py
    def explode(
        self,
        columns: ColumnNameOrSelector | Iterable[ColumnNameOrSelector],
        *more_columns: ColumnNameOrSelector,
        empty_as_null: bool = True,
        keep_nulls: bool = True,
    ) -> DataFrame:
```

PR #1645, opened 2024-12-22
feat: support std and var with ddof !=1 in pandas-like group by

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1606

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/group_by.py
+++ b/narwhals/_pandas_like/group_by.py
@@ -303,6 +303,10 @@ def agg_pandas(  # noqa: PLR0915
                     "Got two aggregations with the same output name. Please make sure "
                     "that aggregations have unique output names."
                 )
+                for key, value in output_names_counter.items():
+                    if value > 1:
+                        msg += f"\n- '{key}' {value} times"
+                msg = f"Expected unique output names, got:{msg}"
                 raise ValueError(msg)
             result = horizontal_concat(
                 dfs=result_aggs,
--- a/narwhals/_pandas_like/group_by.py
+++ b/narwhals/_pandas_like/group_by.py
@@ -14,6 +14,7 @@ from narwhals._expression_parsing import parse_into_exprs
 from narwhals._pandas_like.utils import horizontal_concat
 from narwhals._pandas_like.utils import native_series_from_iterable
 from narwhals._pandas_like.utils import select_columns_by_name
+from narwhals._pandas_like.utils import set_columns
 from narwhals.utils import Implementation
 from narwhals.utils import find_stacklevel
 from narwhals.utils import remove_prefix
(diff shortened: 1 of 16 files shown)


narwhals/_pandas_like/group_by.py
```diff
@@ -237,43 +250,72 @@ def agg_pandas(  # noqa: PLR0915
...
+        if std_aggs:
+            result_aggs.extend(
+                [
+                    rename(
+                        grouped[list(output_to_root_name_mapping.values())].std(
+                            ddof=ddof
+                        ),
+                        # Invert the dict to have root_name: output_name
+                        # TODO(FBruzzesi): Account for duplicates
+                        columns={v: k for k, v in output_to_root_name_mapping.items()},
```
MarcoGorelli, line 272: 🤔 in case someone does `.agg(nw.col('a').std(ddof=1).alias('b'), nw.col('a').std(ddof=1).alias('c'))`? hmm yeah i guess it's possible

btw, since cuDF have no introduced `cudf.NamedAgg`, we could use that if it makes all this logic a bit easier

FBruzzesi: Yes correct, similarly to simple aggs.
To avoid an explosion in the number of lists to keep track of, maybe there is a better data structure. I will think about it a bit, yet aside that specific case, this PR should be ready

narwhals/_pandas_like/group_by.py
```diff
@@ -237,43 +258,58 @@ def agg_pandas(  # noqa: PLR0915
...
-            ):
+
+            result_aggs.append(result_nunique_aggs)
+
+        if std_aggs:
+            result_aggs.extend(
+                [
+                    grouped[std_root_names]
+                    .std(ddof=ddof)
+                    .set_axis(std_output_names, axis="columns", copy=False)
```
FBruzzesi, line 276: @MarcoGorelli how bad is it to use `set_axis` to rename the columns? What's the alternative? Our `rename` does a mapping, yet here we could do:

```py
grouped[["b", "b"]].std(ddof=2).set_axis(["c", "d"], axis="columns")
```
For once we are exploiting some pandas weirdness 😁

MarcoGorelli: Should be fine let's just use copy=false

MarcoGorelli:
oh nice
```
_______ test_group_by_depth_1_std_var[pandas_pyarrow_constructor-var-2] ________
[XPASS(strict)] 
_______ test_group_by_depth_1_std_var[pandas_pyarrow_constructor-var-0] ________
[XPASS(strict)] 
```

MarcoGorelli (approved):
seriously awesome work here @FBruzzesi !

PR #1671, opened 2024-12-28
chore: flatten `subset` in `unique` one level up

MarcoGorelli:
In general, it's probably good if we simplify the implementation of `CompliantFrame`s by doing these kinds of operations at the Narwhals level

If we want to encourage people to extend Narwhals, it'll be easier if they just have to worry about `subset: list[str] | None` rather than `subset: str | list[str] | None`

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -641,7 +641,7 @@ class ArrowDataFrame(CompliantDataFrame, CompliantLazyFrame):
 
     def unique(
         self: Self,
-        subset: str | list[str] | None,
+        subset: list[str] | None,
         *,
         keep: Literal["any", "first", "last", "none"],
         maintain_order: bool,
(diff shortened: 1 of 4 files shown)


FBruzzesi (approved):
I think that's ok and I agree that for someone extending the API this is easier.

The only good use case to keep signature the same is for other methods to (re)use existing Compliant classes methods, yet it's fairly easy to spot what to pass with the fully annotated internals.

 Ship it!

CatKittenGIF

P.S. consider adding this gif?! 👀

Issue #1546, opened 2024-12-09
test: remove filteredwarnings


MarcoGorelli:
I'm concerned about all these:

https://github.com/narwhals-dev/narwhals/blob/a1a44568ab8d523fa35916c300e98157b9ff4595/pyproject.toml#L112-L131

If we need to tightly scope ignoring a warning to one particular test, then OK, so be it, but most of these shouldn't be globally ignored

It would be very useful to go through the list and see, one by one, which ones can be removed, and where they can be more tightly set (e.g. just using `pytest.mark.filterwarnings` in one particular test case)

PR #1691, opened 2024-12-31
tests: remove unnecessary pytest filterwarnings

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1546
- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Tests pass locally. Am I missing something?

Edit: warnings appear with older versions then (I only tested one old version locally)


MarcoGorelli:
thanks! looks like we may need to catch and filter out some of these during `_pyspark_constructor_with_session`, but the rest are probably good to go

DeaMariaLeon:
I'm not trying to guess here..  it's just that I can't install pyspark on my mac yet. And the thread on stackoverflow about it has too many solutions.

pyproject.toml
```diff
@@ -150,22 +150,8 @@ docstring-code-format = true
...
-  'ignore:distutils Version classes are deprecated:DeprecationWarning',
-  'ignore:In the future `np.bool`',
-  'ignore:make_block is deprecated and will be removed',
-  'ignore:np.find_common_type is deprecated',
-  'ignore:is_sparse is deprecated and will be removed',
-  'ignore:Passing a BlockManager to DataFrame is deprecated',
   'ignore:.*defaulting to pandas implementation',
   'ignore:.*implementation has mismatches with pandas',
-  'ignore:.*Do not use the `random` module inside strategies',
-  'ignore:.*You are using pyarrow version',
```
MarcoGorelli, line 162: we can probably keep this one turned off globally, as it just warns about pyarrow==14 being insecure so it's not going to create issues for us going forwards (unlike, say, silencing a deprecation warning)

DeaMariaLeon: Thanks

MarcoGorelli (approved):
thanks a lot @DeaMariaLeon , this kind of thing can be frustrating but makes a difference in the long run, much appreciated!

PR #1693, opened 2024-12-31
feat: add `sum` and `sum_horizontal` for `SparkLike`

lucas-nelson-uiuc:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Building on #1633

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Implemented methods for PySpark DataFrame such that all non-eager examples run "as expected" in the [DataFrame tutorial](https://narwhals-dev.github.io/narwhals/basics/dataframe/). Specifically, added the following to `narwhals._spark_like`:

- `SparkLikeExpr.sum`
- `SparkLikeNamespace.sum_horizontal`

## TODO

Had two issues putting together this request:

- Running `pytest` failed because Java isn't installed properly on my machine (was hoping Actions would verify this for me but couldn't successfully trigger that either).
- Running `mkdocs serve` also failed for similar reasons I believe.

Also happy to add tests and/or additional methods of interest.


dev.ipynb
FBruzzesi, line 1: This should not be included in the PR 😇

docs/installation.md
FBruzzesi, line 1: Not sure if we want to publicize pyspark support just yet. I think it would be ok to do one once we can at least run TPCH queries 🤔

FBruzzesi (commented):
Thanks for the contribution @lucas-nelson-uiuc! This is great!

Could you please add unit tests for the introduced `sum` and `sum_horizontal` in `./tests/spark_like_test.py` that mirror those in the main test suite?

Other comments are in the respective files :)

dev.ipynb
lucas-nelson-uiuc: Removed

docs/installation.md
lucas-nelson-uiuc: Agreed ^ removed any mentions of PySpark in the mkdocs.

lucas-nelson-uiuc:
Thanks @FBruzzesi

Copied the `sum` and `sum_horizontal` tests and all pass when I run them.

Only consideration is that `test_sumh_nullable` failed on the implementation as of yesterday because PySpark cannot compute sums with NULL values. Added an if-else statement to replace all instances of NULL with zero as the sum is being computed using `pyspark.sql.functions.isnan` to pass the test. Let me know if this works.

FBruzzesi:
Thanks for adjusting @lucas-nelson-uiuc!

I pushed one little change following your suggestion.
The correct behaviour would be to use coalesce to replicate what polars does:

```py
import polars as pl

df = pl.DataFrame({"a": [1, 0, None]}).with_columns(b=pl.col("a")/pl.col("a"))

df
shape: (3, 2)
┌──────┬──────┐
│ a    ┆ b    │
│ ---  ┆ ---  │
│ i64  ┆ f64  │
╞══════╪══════╡
│ 1    ┆ 1.0  │
│ 0    ┆ NaN  │
│ null ┆ null │
└──────┴──────┘

df.select(pl.sum_horizontal("a", "b"))
shape: (3, 1)
┌─────┐
│ a   │
│ --- │
│ f64 │
╞═════╡
│ 2.0 │
│ NaN │
│ 0.0 │
└─────┘
```
As you can see, actual `NaN` is not skipped, but propagated, while null's get skipped, resulting in zeros.

FBruzzesi (approved):
Thanks @lucas-nelson-uiuc 🙌🏼

NarwhalsKittyGIF

lucas-nelson-uiuc:
Happy to help :)

PR #1706, opened 2025-01-02
fix: `DaskLazyFrame` and `ArrowDataFrame` `.with_row_index` column order

### LINKED ISSUE #1703: bug: order of columns after `with_row_index` differs between Dask and Polars.LazyFrame
As spotted here: https://github.com/narwhals-dev/narwhals/pull/1700/files#r1900446885

I'm not sure about this one. My current inclination is to remove `with_row_index` from `narwhals.LazyFrame` because it's row-order dependent (but silently keep it around in dask for backcompat)

### LINKED ISSUE #1707: [Bug]: SparkLike join returns different column order
Noticed while implementing order check in result vs expected columns in #1706 .

In such PR, I adjusted the tests for pyspark to make them pass, yet we should deal with it to make it consistent with other implementations

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1703 
- Closes #1707 

## Checklist

- [ ] Tests added
- [ ] Documented the changes


tests/frame/with_row_index_test.py
```diff
@@ -12,5 +12,5 @@
 
 def test_with_row_index(constructor: Constructor) -> None:
     result = nw.from_native(constructor(data)).with_row_index()
-    expected = {"a": ["foo", "bars"], "ab": ["foo", "bars"], "index": [0, 1]}
+    expected = {"index": [0, 1], "a": ["foo", "bars"], "ab": ["foo", "bars"]}
```
FBruzzesi, line 15: Order is not checked anyway

MarcoGorelli: can we make `assert_equal_data` check the order?

FBruzzesi: Nice one! Turns out pyarrow was also wrong! Should be fixed now

tests/spark_like_test.py
```diff
@@ -892,12 +892,14 @@ def test_left_join_overlapping_column(pyspark_constructor: Constructor) -> None:
...
         "d": [1, 4, 2],
+        "idx": [0, 1, 2],
         "antananarivo_right": [1, 2, 3],
         "d_right": [1, 4, 2],
-        "idx": [0, 1, 2],
     }
     assert_equal_data(result, expected)
 
+    df_left = nw.from_native(pyspark_constructor(data_left))
+    df_right = nw.from_native(pyspark_constructor(data_right))
```
FBruzzesi, line 902: This might need some investigation: If we don't create new dataframes the test fails - which hints to the fact that we are modifying the dataframe?

tests/utils.py
```diff
@@ -82,8 +82,8 @@ def assert_equal_data(result: Any, expected: dict[str, Any]) -> None:
             result = result.collect()
 
     if hasattr(result, "columns"):
-        for key in result.columns:
-            assert key in expected, (key, expected)
+        for idx, (col, key) in enumerate(zip(result.columns, expected.keys())):
+            assert col == key, f"Expected column name {key} at index {idx}, found {col}"
```
FBruzzesi, line 86: Suggestion for a better message is always welcomed :)

MarcoGorelli (approved):
awesome, i'm sure if would've created issues for people without your fix, thanks!

Issue #1714, opened 2025-01-03
[Enh]: Spark Expr missing methods


FBruzzesi:
Methods which are row order dependent are not included.
Methods within a namespace are included only if the namespace at least exists, otherwise it means that all namespace methods are missing.

Methods with one asterisk (*) change the length but don't aggregate - these are deprioritized for now

High priority:

- [x] abs
- [x] all
- [x] any
- [x] clip
- [x] is_between
- [x] is_duplicated
- [x] is_finite
- [x] is_in
- [x] is_nan
- [x] is_unique
- [x] len 
- [x] median
- [x] n_unique
- [x] null_count
- [x] over
- [x] round
- [x] skew
- [ ] drop_nulls (*)
- [ ] fill_null (if strategy is prodived, otherwise it is order dependent)
- [ ] filter (*)
- [ ] mode (*)
- [ ] quantile
- [ ] replace_strict
- [ ] unique (*)

Namespaces:

- [x] name (**)
- [ ] cat (**)
- [x] list (**)
- [ ] dt (**)
  - [ ] to_string
  - [ ] total_microseconds
  - [ ] total_milliseconds
  - [ ] total_minutes
  - [ ] total_nanoseconds
  - [ ] total_seconds
- [ ] str (**)
  - [ ] replace
  - [ ] to_datetime

lucas-nelson-uiuc:
Hey @FBruzzesi ,

Working on implementing scalar methods like `any` and `all` - should be ready to push later today.

Wondering if my thought process for methods like `drop_nulls` and `filter` is "correct". Thinking that all predicate-based methods should make use of PySpark's [filter function](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.filter.html#) - however, this accepts and returns an array of all elements. To properly input/output column objects, thinking we add steps to convert to/from arrays:

```python
def drop_nulls(self) -> Self:
        def _drop_nulls(_input: Column) -> Column:
            from pyspark.sql import functions as F  # noqa: N812

            return F.explode(F.filter(F.array(_input), F.isnotnull)) # alias added by `returns_scalar=False`

        return self._from_call(_drop_nulls, "drop_nulls", returns_scalar=False)
```

Not sure how expensive doing this is or if it collides with future API developments. Lmk what you think

MarcoGorelli:
thanks @lucas-nelson-uiuc for your efforts here

[...]

PR #1721, opened 2025-01-04
Feat(spark): add few missing Expr methods

Dhanunjaya-Elluri:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1714 

## Checklist

- [x] Tests added
- [ ] Documented the changes

Added the following missing methods in `SparkLikeExpr`:
`abs`, `median`, `clip`, `is_between`, `is_duplicated`, `is_finite`, `is_in`, `is_nan`, `is_unique`, `len`, `n_unique`, `round`, `skew`


Dhanunjaya-Elluri:
Hi @FBruzzesi, I've added some missing methods to `SparkLikeExpr` related to #1714. Would appreciate feedback if some changes is required!

narwhals/_spark_like/expr.py
```diff
@@ -183,6 +183,14 @@ def __gt__(self, other: SparkLikeExpr) -> Self:
             returns_scalar=False,
         )
 
+    def abs(self) -> Self:
+        def _abs(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            return F.abs(_input)
+
+        return self._from_call(_abs, "abs", returns_scalar=self._returns_scalar)
```
FBruzzesi, line 192: I am tempted to say that in these "simple" cases we could just do:

```suggestion
    def abs(self) -> Self:
        from pyspark.sql import functions as F  # noqa: N812

        return self._from_call(F.abs, "abs", returns_scalar=self._returns_scalar)
```

narwhals/_spark_like/expr.py
```diff
@@ -272,3 +288,153 @@ def var(self: Self, ddof: int) -> Self:
         )
 
         return self._from_call(func, "var", returns_scalar=True, ddof=ddof)
+
+    def clip(
+        self,
+        lower_bound: Any | None = None,
+        upper_bound: Any | None = None,
+    ) -> Self:
```
FBruzzesi, line 296: We recently introduced support for `lower_bound` and `upper_bound` to be other Expr's.
I am ok to keep it as a follow up, but definitly something we would look forward to.

narwhals/_spark_like/expr.py
```diff
@@ -272,3 +288,153 @@ def var(self: Self, ddof: int) -> Self:
...
+            upper_bound=upper_bound,
+            returns_scalar=self._returns_scalar,
+        )
+
+    def is_between(
+        self,
+        lower_bound: Any,
+        upper_bound: Any,
+        closed: str = "both",
+    ) -> Self:
```
FBruzzesi, line 326: Same as for `clip`, but I imagine this to already work. Only type hints might need to be updated.

Also, we try to avoid default values in the compliant classes, thus it should be:

```diff
- closed: str = "both",
+ closed: str,
```

narwhals/_spark_like/expr.py
```diff
@@ -272,3 +288,153 @@ def var(self: Self, ddof: int) -> Self:
...
+
+    def is_nan(self) -> Self:
+        def _is_nan(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            # Need to handle both NaN and NULL values
+            return F.when(
+                F.isnan(_input) | F.isnull(_input),
+                F.lit(1),
+            ).otherwise(F.lit(0))
```
FBruzzesi, line 390: Couple of consideration here:

- This should return boolean's Column, not sure why we are converting to integer type.
- As spark distinguish between nulls and nan, for `is_nan`, null should remain null.

This should simplify to:

```suggestion
            return F.isnan(_input)
```

narwhals/_spark_like/expr.py
```diff
@@ -272,3 +288,153 @@ def var(self: Self, ddof: int) -> Self:
...
+            _is_unique, "is_unique", returns_scalar=self._returns_scalar
+        )
+
+    def len(self) -> Self:
+        def _len(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            return F.count(_input)
+
+        return self._from_call(_len, "len", returns_scalar=True)
```
FBruzzesi, line 411: Unlike count, `len` should return the length including nulls (see [polars docs](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.len.html))

FBruzzesi (commented):
Serious great effort @Dhanunjaya-Elluri! Thanks a ton for this, I am very excited 🚀

I was in a bit of a rush but manage to leave some feedback.
I think the main critical point for many methods is null's and/or nan's cases where relevant. The best way to catch different behaviour of methods is to test against data containing null's and/or nan's, by adapting the existing tests with the `pyspark_constructor`.

narwhals/_spark_like/expr.py
Dhanunjaya-Elluri: Good point. That being said, should we adopt here too?

https://github.com/narwhals-dev/narwhals/blob/ca445729ccb256c496815847c80ef3dbdff26718/tests/spark_like_test.py#L28

Dhanunjaya-Elluri:
Hi @FBruzzesi, made some changes now. Just one question as requested in the [review](https://github.com/narwhals-dev/narwhals/pull/1721#discussion_r1903234642).

EdAbati (commented):
Thank you very much from me too 🙏🏼

I just had a brief look from mobile for now, it's great work!

CONTRIBUTING.md
```diff
@@ -78,6 +78,29 @@ where `YOUR-GITHUB-USERNAME` will be your GitHub user name.
 
 Here's how you can set up your local development environment to contribute.
 
+#### Prerequisites for PySpark tests
```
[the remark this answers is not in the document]
Dhanunjaya-Elluri: I think it would be simple to just say that pyspark needs java installed and add a link to pyspark documentation.

tests/spark_like_test.py
```diff
@@ -943,3 +955,179 @@ def test_left_join_overlapping_column(pyspark_constructor: Constructor) -> None:
         "c": [4.0, 6.0, None],
     }
     assert_equal_data(result, expected)
+
+
+# Copied from tests/expr_and_series/median_test.py
+@pytest.mark.xfail(
+    sys.version_info < (3, 9),
+    reason="median() not supported on Python 3.8",
+)
```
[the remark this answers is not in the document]
Dhanunjaya-Elluri: Thats a great observation @EdAbati. Just updated! Please review.

tests/spark_like_test.py
```diff
@@ -943,3 +954,175 @@ def test_left_join_overlapping_column(pyspark_constructor: Constructor) -> None:
...
+        "a": [1.0, None, None, 3.0],
+        "b": [1.0, None, 4, 5.0],
+    }
+    df = nw.from_native(pyspark_constructor(data))
+    result = df.select(
+        a=nw.col("a").n_unique(),
+        b=nw.col("b").n_unique(),
+    )
+    expected = {"a": [2], "b": [3]}
+    assert_equal_data(result, expected)
```
FBruzzesi, line 1092: None and nan should also count toward `n_unique`: see [`polars.Expr.n_unique`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.n_unique.html#polars.Expr.n_unique)

tests/spark_like_test.py
```diff
@@ -943,3 +954,175 @@ def test_left_join_overlapping_column(pyspark_constructor: Constructor) -> None:
...
+        "level_0": [0, 1, 2],
+    }
+    assert_equal_data(result, expected)
+
+
+def test_is_nan(pyspark_constructor: Constructor) -> None:
+    data = {"a": [1.0, float("nan"), 2.0, None, 3.0]}
+    df = nw.from_native(pyspark_constructor(data))
+    result = df.select(nan=nw.col("a").is_nan())
+    expected = {"nan": [False, False, False, False, False]}
```
FBruzzesi, line 1027: Shouldn't this result in:

```suggestion
    expected = {"nan": [False, True, False, False, False]}
```
?

tests/spark_like_test.py
```diff
@@ -943,3 +954,175 @@ def test_left_join_overlapping_column(pyspark_constructor: Constructor) -> None:
...
+) -> None:
+    data = {"a": [1, 4, 2, 5]}
+    df = nw.from_native(pyspark_constructor(data))
+    result = df.select(nw.col("a").is_between(1, 5, closed=closed))
+    expected_dict = {"a": expected}
+    assert_equal_data(result, expected_dict)
+
+
+# copied from tests/expr_and_series/is_duplicated_test.py
+def test_is_duplicated(pyspark_constructor: Constructor) -> None:
```
FBruzzesi, line 1007: Can we test this one and `unique` with None's as well?

The expected behaviour is:

```py
data = {"a": [1, 1, 2, None], "b": [1, 2, None, None], "level_0": [0, 1, 2, 3]}
pl.DataFrame(data).select(pl.col("a", "b").is_duplicated())
shape: (4, 2)
┌───────┬───────┐
│ a     ┆ b     │
│ ---   ┆ ---   │
│ bool  ┆ bool  │
╞═══════╪═══════╡
│ true  ┆ false │
│ true  ┆ false │
│ false ┆ true  │
│ false ┆ true  │
└───────┴───────┘
```

FBruzzesi (commented):
Thanks for adjusting @Dhanunjaya-Elluri ! We are getting closer and closer 👌
I just have a couple of comments that need to be addressed

tests/spark_like_test.py
Dhanunjaya-Elluri: I had this question in mind and pointed earlier [here](https://github.com/narwhals-dev/narwhals/pull/1721#discussion_r1903234642). Its because `nan` is replaced with `NULL` here. I guess `nan` should be treated as itself. WDYS?

https://github.com/narwhals-dev/narwhals/blob/2aa436d1de98a2aa0b16d852467969de0b68754d/tests/spark_like_test.py#L42

FBruzzesi: Apologies for missing that!

Let's try to re-create a nan after creation. Maybe something along the following:


```py
    data = {"a": [0, 1.0, None]}
    df = nw.from_native(pyspark_constructor(data))
    result = df.select(nan=(nw.col("a")/nw.col("a")).is_nan())
```
I would expect this to result in `[True, False, False]`

Dhanunjaya-Elluri: Hi @FBruzzesi, I tried this way, but 
```python
df.select(nan=(nw.col("a")/nw.col("a")).is_nan())
```
would return `NoneType`.

Alternatively, I simply tried replacing some value with `nan`, but `SparkLikeNamespace` has no attribute `when`. Any suggestions?

FBruzzesi: I don't understand the pyspark rules. The square root of -1 would generate a NaN, yet I am unable to make it work with narwhals:

```py
frame = session.createDataFrame(pd.DataFrame({"a": [0, 1, None]}).replace({float("nan"): None}))
nw.from_native(frame).select(b = nw.col("a")**0.5).to_native()
```

> InvalidIntoExprError: Expected an object which can be converted into an expression, got <class 'NoneType'>
>
> Hint:
> - if you were trying to select a column which does not have a string
>   column name, then you should explicitly use `nw.col`.
>   For example, `df.select(nw.col(0))` if you have a column named `0`.
> - if you were trying to create a new literal column, then you 
>   should explicitly use `nw.lit`.
>   For example, `df.select(nw.lit(0))` if you want to create a new
>   column with literal value `0`.

Would you be ok to keep `is_nan` method as a follow up? Although it seems quite straightforward, we would like to be able to test it properly

Dhanunjaya-Elluri: Sure, for now I'll push the changes without `is_nan` and followup it later.

tests/spark_like_test.py
Dhanunjaya-Elluri: Done

Dhanunjaya-Elluri:
Hey @FBruzzesi, done changes now. Let me know if there needs to be any changes

narwhals/_spark_like/expr.py
```diff
@@ -249,3 +252,160 @@ def var(self: Self, ddof: int) -> Self:
...
+        def _is_duplicated(_input: Column) -> Column:
+            from pyspark.sql import Window
+            from pyspark.sql import functions as F  # noqa: N812
+
+            # Create a window spec that treats each value separately
+            window = Window.partitionBy(
+                F.when(F.isnull(_input), F.lit("NULL"))
+                .when(F.isnan(_input), F.lit("NAN"))
+                .otherwise(_input)
+            )
```
FBruzzesi, line 318: Kind of wild case but... what if the column is string type and someone has a string with NULL/NAN value already?
I think there was a similar case in which we first check non-nullable case and then apply an extra check on top.
I am from mobile now, it's a bit hard to recover the code snippet.

narwhals/_spark_like/expr.py
```diff
@@ -249,3 +252,160 @@ def var(self: Self, ddof: int) -> Self:
...
+        def _is_unique(_input: Column) -> Column:
+            from pyspark.sql import Window
+            from pyspark.sql import functions as F  # noqa: N812
+
+            # Create a window spec that treats each value separately
+            window = Window.partitionBy(
+                F.when(F.isnull(_input), F.lit("NULL"))
+                .when(F.isnan(_input), F.lit("NAN"))
+                .otherwise(_input)
+            )
```
FBruzzesi, line 364: Same as `is_duplicated`

narwhals/_spark_like/expr.py
```diff
@@ -249,3 +252,160 @@ def var(self: Self, ddof: int) -> Self:
...
+
+    def n_unique(self) -> Self:
+        def _n_unique(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            expr = (
+                F.when(F.isnull(_input), F.lit("NULL"))
+                .when(F.isnan(_input), F.lit("NaN"))
+                .otherwise(_input)
+            )
```
FBruzzesi, line 390: Same as `is_duplicated`

FBruzzesi:
Hey @Dhanunjaya-Elluri  I just took another look at this.

For `is_duplicated`, the following should be all we need:

```py
def is_duplicated(self) -> Self:
        def _is_duplicated(_input: Column) -> Column:
            from pyspark.sql import Window
            from pyspark.sql import functions as F  # noqa: N812

            # Create a window spec that treats each value separately.
            return F.count("*").over(Window.partitionBy(_input)) > 1

        return self._from_call(
            _is_duplicated, "is_duplicated", returns_scalar=self._returns_scalar
        )
```

Similarly for `is_unique`.

However, `n_unique` is definitly more tricky (and we can consider to postpone it for now) as `F.count_distinct` ignores nulls

Dhanunjaya-Elluri:
Hi @FBruzzesi, thanks for taking deeper look. For now I'll push these changes then

FBruzzesi (approved):
Nice one! 🙌🏼

NarwhalSpinGIF

PR #1724, opened 2025-01-04
feat: add `all`, `any` and `null_count` Spark Expressions

lucas-nelson-uiuc:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Builds on #1714

## Checklist

- [x] Tests added
- [ ] Documented the changes

Added the following methods to `SparkLikeExpr` and `SparkLikeNamespace`:

- `any`
- `all`
- `null_count`
- `any_horizontal`

Copied respective tests over - couldn't run them without Java on my machine but running them locally on their respective test datasets worked for me.

Let me know if anything needs to be updated!


tests/spark_like_test.py
EdAbati, line 1: Like you did in `test_any_all`, we always need to use the `pyspark_constructor` to test with the pyspark backend here.
`constructor` will test all the backends [except of pyspark](https://github.com/narwhals-dev/narwhals/blob/main/tests/conftest.py#L192-L200).

Also we can delete the eager tests (i.e. `test_any_all_series`, `test_null_count_series`), since pyspark is a lazy backend. :)

narwhals/_spark_like/expr.py
```diff
@@ -201,6 +201,22 @@ def _alias(df: SparkLikeLazyFrame) -> list[Column]:
             kwargs={**self._kwargs, "name": name},
         )
 
+    def all(self) -> Self:
+        def _all(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            return F.bool_and(_input)
```
EdAbati, line 208: `bool_and` and `bool_or` are only available from pyspark 3.5, while narwhals minimum dependency is 3.3

Is there a way to do write these for older versions of spark?

EdAbati (commented):
Thank you very much for looking into this! 🙏

narwhals/_spark_like/expr.py
MarcoGorelli: this would be new functionality, maybe we could bump the minimum pyspark version?

tests/spark_like_test.py
lucas-nelson-uiuc: Thanks for catching that - updated the constructor and removed all eager tests.

narwhals/_spark_like/expr.py
lucas-nelson-uiuc: not opposed ^ pyspark 3.5 introduced a whole batch of functions that would let us work on other expressions like the ones mentioned in #1714

lucas-nelson-uiuc:
Hey @EdAbati ,

Took an initial swing at implementing the `replace_strict()` method - think I took care of everything except for handling the `test_replace_non_full` test (checks that replacement is exhaustive). Left some thoughts and other questions in my commit - lmk what you think!

narwhals/_spark_like/expr.py
FBruzzesi: This seems the easier way forward. Alternatively, raise for those methods and backend_version pre 3.5. E.g.

```py
if self._backend_version < (3, 5):
    msg = ...

...
```

MarcoGorelli: thanks for the perspective - sure, if it's not too painful to support 3.3+, i'd say that that's ok

EdAbati: moving the discussion to #1744

narwhals/_spark_like/expr.py
```diff
@@ -201,6 +208,22 @@ def _alias(df: SparkLikeLazyFrame) -> list[Column]:
...
+
+        return self._from_call(_all, "all", returns_scalar=True)
```
EdAbati, line 217: We simplified a bit the other methods, we can refactor as

```suggestion
        from pyspark.sql import functions as F  # noqa: N812

        return self._from_call(F.bool_and, "all", returns_scalar=True)
```

narwhals/_spark_like/expr.py
```diff
@@ -201,6 +208,22 @@ def _alias(df: SparkLikeLazyFrame) -> list[Column]:
...
+
+    def any(self) -> Self:
+        def _any(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            return F.bool_or(_input)
+
+        return self._from_call(_any, "any", returns_scalar=True)
```
EdAbati, line 225:
```suggestion
        from pyspark.sql import functions as F  # noqa: N812

        return self._from_call(F.bool_or, "any", returns_scalar=True)
```

same as above

narwhals/_spark_like/expr.py
```diff
@@ -233,6 +256,50 @@ def _min(_input: Column) -> Column:
...
 
+    def null_count(self) -> Self:
+        def _null_count(_input: Column) -> Column:
+            from pyspark.sql import functions as F  # noqa: N812
+
+            return F.count_if(F.isnull(_input))
+
+        return self._from_call(_null_count, "null_count", returns_scalar=True)
+
+    def replace_strict(
```
EdAbati, line 267: I am tempted to say that this should not be implemented for now and just raise a `NotImplementedError`. (as we do in Dask)
We would need to be able to access the dataframe (and collect the results) to get the distinct values of the column.

@FBruzzesi and @MarcoGorelli any thoughts?

EdAbati (commented):
Thanks for updating! :)

I left another couple of small comments

IMO we can just merge `all`, `any` and `null_count` here and worry about the rest in a follow up

narwhals/_spark_like/expr.py
FBruzzesi: > I am tempted to say that this should not be implemented for now and just raise a NotImplementedError. (as we do in Dask)

Sure we can evaluate if and how to support `replace_strict` later on. Super excited to ship the rest for now 🙌🏼

MarcoGorelli: agree!

lucas-nelson-uiuc:
Thanks everyone - committed the changes above. Lmk if anything else needs to be changed

MarcoGorelli:
awesome, thanks! just looks like there's some merge conflicts

lucas-nelson-uiuc:
Might've committed some git crimes resolving that merge conflict :') apologies in advance

FBruzzesi (approved):
Let's ship this🚀 ! Thanks @lucas-nelson-uiuc and apologies for such a long back and forth on this PR 

NarwhalsOceanGIF

PR #1725, opened 2025-01-05
feat: Implement partial "lazy" support for DuckDB (even with this PR, DuckDB support is work-in-progress!)

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #1734, opened 2025-01-06
feat: `LazyFrame.collect` with backend and **kwargs

### LINKED ISSUE #1479: API: `collect` for lazy-only libraries
We currently have `LazyFrame.collect`:
- for Polars.LazyFrame, this goes to Polars.DataFrames
- for dask.DataFrames, it goes to pandas.DataFrame
- when we get to PySpark / DuckDB / etc ... where will it go?

I think one solution could be to add extra keyword arguments to `LazyFrame.collect` to specify such ambiguous cases. Adding extra (non-required) keyword arguments would still be backwards-compatible

The signature could look like this:
```python
def collect(self, duckdb_eager_frame: Literal['pandas', 'polars', 'pyarrow'] = 'pyarrow', pyspark_eager_frame: Literal['pandas'] = 'pandas') -> DataFrame[Any]:
```

This still leave us with the question about what to do for "extension" dataframes, i.e. if someone implements `__narwhals_dataframe__` / `__narwhals_lazyframe__` and extends Narwhals themself. We could require them to also implement a dunder method specifying which eager frame they want their lazy frame collected to, or which ones they support collecting into, and then have an extra kwarg in `collect` for that

@EdAbati any thoughts?

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1479

## Checklist

- [ ] Tests added
- [x] Documented the changes

This is a proposal for #1479.

As it gets more relevant now due to DuckDB support and to decide how we could collect a DuckDB table.

For polars and dask, collect kwargs would follow native collect and compute respectively. For DuckDB we could come up with our own and document it properly. Specifically I would suggest to let the user decide to which dataframe backend to collect to (`return_type`?), with Arrow as default.


MarcoGorelli:
Nice, I like the look of this

it may help to simplify

https://github.com/narwhals-dev/narwhals/blob/8c9525a3c62de8c83f6a8382c94592891e5d2f5c/tests/utils.py#L77-L80

to just `result.collect(engine="gpu")`?

---

> I would suggest to let the user decide to which dataframe backend to collect to (return_type?), with Arrow as default.

agree, I think arrow's a good default for duckdb (also, as far as I can tell, collecting into Polars from duckdb requires pyarrow anyway, suggesting they first collect into pyarrow anyway?). to check my understanding then, this would be compatibe with the `collect` in https://github.com/narwhals-dev/narwhals/pull/1725, and we can add extra kwargs later?

FBruzzesi:
> to just `result.collect(polars_kwargs=dict(engine="gpu"))`?

Yes exactly!

> agree, I think arrow's a good default for duckdb (also, as far as I can tell, collecting into Polars from duckdb requires pyarrow anyway, suggesting they first collect into pyarrow anyway?). 

🤔 now that you mention (and completely unrelated from this PR), we could do the same for pyspark: see [SO Ritchie answer](https://stackoverflow.com/a/73205690/12411536)

> to check my understanding then, this would be compatibe with the `collect` in #1725, and we can add extra kwargs later?

Yes indeed as long as we intend to have the default to be PyArrow

MarcoGorelli:
> we could do the same for pyspark: see [SO Ritchie answer](https://stackoverflow.com/a/73205690/12411536)

yes, nice! 🙌

MarcoGorelli:
I think this looks good, just going to give the chance to others to weigh in

FBruzzesi:
Added `duckdb_kwargs` to specify DuckDB `return_type` in the last commit. Can revert it if we want to sleep on it

narwhals/_duckdb/dataframe.py
```diff
@@ -76,20 +80,67 @@ def __getitem__(self, item: str) -> DuckDBInterchangeSeries:
...
+            )
+
+        elif return_type == "polars":
+            import polars as pl  # ignore-banned-import
+
+            from narwhals._polars.dataframe import PolarsDataFrame
+            from narwhals.utils import Implementation
+
+            return PolarsDataFrame(
+                df=self._native_frame.pl(),
```
FBruzzesi, line 133: Unrelated but.. should we change in `PolarsDataFrame`:

```diff
- df: pl.DataFrame,
+ native_dataframe: pl.DataFrame,
```

narwhals/dataframe.py
```diff
@@ -3610,16 +3610,40 @@ def __getitem__(self, item: str | slice) -> NoReturn:
         msg = "Slicing is not supported on LazyFrame"
         raise TypeError(msg)
 
-    def collect(self) -> DataFrame[Any]:
+    def collect(
+        self: Self,
+        *,
+        polars_kwargs: dict[str, Any] | None = None,
+        dask_kwargs: dict[str, Any] | None = None,
+        duckdb_kwargs: dict[str, str] | None = None,
```
FBruzzesi, line 3618: These could all be `TypedDict`'s 👀

EdAbati:
Niiiice! 🙌🙌

A couple of questions/idea:

- wouldn't `return_type` eventually become an argument for every Lazy backend? e.g. one may want to always collect to `polars` regardless of which Lazy dataframe they are starting from.
  What do you think about changing the signature to:
  ```python
  def collect(self, return_type, ...) -> ...
  ```
- still not 100% sold to the idea of having one specific `kwargs` per Lazy backend, I think it would give us a bit less flexibility. The signature would change if we add more lazy backends (e.g. `pyspark`). Or _I think_ it'd become a problem if:
   - we remove a lazy backend, for example moving it to an separate integration library
   - people wants to create their [own Lazy backends](https://narwhals-dev.github.io/narwhals/extending/#extending-narwhals) and want to pass args to their collect.
   
   What do you think?
   
   Also I wouldn't find particularly ugly something like this (but that's personal preference :D ):
   ```python
    lazy_df = df.lazy().select(nw_v1.all().sum())
    if lazy_df.implementation == Implementation.POLARS:
        eager_df = lazy_df.collect(no_optimization=True)
    elif lazy_df.implementation == Implementation.DASK:
        eager_df = lazy_df.collect(optimize_graph=False)
    elif lazy_df.implementation == Implementation.DUCKDB:
        eager_df = lazy_df.collect(return_type="pyarrow")
   ```

FBruzzesi:
Hey @EdAbati, thanks for your feedback! that's exactly the purpose of a RFC 👌


> * wouldn't `return_type` eventually become an argument for every Lazy backend? e.g. one may want to always collect to `polars` regardless of which Lazy dataframe they are starting from.

Considering the dataframe ecosystem as of today:

- Polars would always collect to polars
- Dask would always collect to pandas
- DuckDB (and maybe Ibis) could collect to polars, arrow and pandas
- Spark natively to pandas (but I see the appeal of collecting to others)

Not sure what we are aim to support in the future 😉 but I would try not to break our head too soon here!

> * still not 100% sold to the idea of having one specific `kwargs` per Lazy backend, I think it would give us a bit less flexibility. The signature would change if we add more lazy backends (e.g. `pyspark`). Or _I think_ it'd become a problem if:
>   
>   * we remove a lazy backend, for example moving it to an separate integration library

Those are definitly fair concerns to think about! Thanks for pointing those out!

>   * people wants to create their [own Lazy backends](https://narwhals-dev.github.io/narwhals/extending/#extending-narwhals) and want to pass args to their collect. (probably a niche use case though)

That's when they can branch out, *or* we can add additional `**kwargs` to be passed to all `.collect` if the implementation is outside of what we cover with the dedicated arguments. In code:

```py
def collect(
    self: Self,
    *,
    polars_kwargs: dict[str, Any] | None = None,
    dask_kwargs: dict[str, Any] | None = None,
    duckdb_kwargs: dict[str, str] | None = None,
    **kwargs: Any,
):

    if self.implementation is Implementation.POLARS and polars_kwargs is not None:
        kwargs_ = polars_kwargs
    elif ...:
        ...
    else:
        kwargs_ = kwargs

    return self._dataframe(
        self._compliant_frame.collect(**kwargs),
        level="full",
    )
```

>   
>   What do you think?
>   Also I wouldn't find particularly ugly something like this (but that's personal preference :D ):
>   ```python
>    lazy_df = df.lazy().select(nw_v1.all().sum())
>    if lazy_df.implementation == Implementation.POLARS:
>        eager_df = lazy_df.collect(no_optimization=True)
>    elif lazy_df.implementation == Implementation.DASK:
>        eager_df = lazy_df.collect(optimize_graph=False)
>    elif lazy_df.implementation == Implementation.DUCKDB:
>        eager_df = lazy_df.collect(return_type="pyarrow")
>   ```

I would be ok deferring the responsibility to the users. But coming back to personal preference, if I were to use an external library to find that I have to do a lot of branching myself, I wouldn't say it is particularly ergonomic 😅

EdAbati:
> Not sure what we are aim to support in the future 😉 but I would try not to break our head too soon here!

Ah! I was under the impression that there was a request to make `LazyFrame`s able to collect to any backend. Similarly to what the methods `to_pandas()`, `to_arrow` etc would do for eager frames. Maybe I misunderstood then :D

Also FYI PySpark 4.0.0 will support [`toArrow`](https://spark.apache.org/docs/4.0.0-preview2/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.toArrow.html)

MarcoGorelli:
there is indeed a request to be able to collect to specific backends (e.g. someone via work asked to be able to collect duckdb-backed lazyframe into polars-backed dataframe), but I think this would still be backend-specific - e.g. not all lazy backends would necessarily have a way to collect to polars

as in, which might not have a way to do `lazy_df.collect(eager_backend='polars')` and know that it will work for all lazy backends...but if we know that say duckdb supports it, we can do `lazy_df.collect(duckdb_kwargs={'eager_backend': 'polars'})`

Having said that, should we:
- use `eager_backend` instead of `return_dtype`? `return_type` kind of sounds like it affects the return type of the function, although that is always `nw.DataFrame`
- should we document a couple of common kwargs for polars / dask backends? e.g. `streaming`, `engine='gpu'`. then for duckdb_kwargs / pyspark_kwargs we can document `'eager_backend'`

> if lazy_df.implementation == Implementation.POLARS:

small note, but this can now be done as `if lazy_df.implementation.is_polars()`

MarcoGorelli:
Been thinking about this a bit more, and in the read functions we have `**kwargs`, perhaps we should be consistent with that? To avoid branching, users could make a collect_kwargs dictionary which maps implementations to kwargs and then do
```python
lf.collect(collect_kwargs[lf.implementation])
```

MarcoGorelli:
TBH, the more I think about this, the more I'd be in favour of just:
```
    def collect(self: Self, backend: ModuleType | Implementation | str | None, **kwargs) -> DataFrame[Any]: ...
    def lazy(self: Self, backend: ModuleType | Implementation| str | None) -> LazyFrame[Any]:
```
where `native_namespace` can accept either namespaces like `pl` / etc..., or strings like `'pandas'`

Because then:

- as a user, I could do
```python
df = (
    lf.group_by('a', 'b').agg(nw.all().mean().name.suffix('_mean'))
    .collect(engine='gpu')  # alternative is `.collect(polars_kwargs={'engine': 'gpu'})`
    .with_columns(pl.col('a').rolling_mean(2))
    .to_native()
```
- as a tool-builder wanting to avoid extensive `if/then` statement, I could do:
```python
collect_kwargs = {
    'polars': {'engine: 'gpu'},
    'dask': {'optimize_graph': False},
    'duckdb': {'backend': 'pyarrow'},
    my_fancy_extension_module: {'my_fancy_kwarg': my_fancy_value},
}
 lazy_df = df.lazy().select(nw.all().sum()).collect(**collect_kwargs)
```
- it's possible to round-trip lazy-eager-lazy, like
```python
lf = lf.group_by('foo').agg(nw.selectors.numeric.mean())
implementation = lf.implementation
df = lf.collect()
train_df, val_df = train_test_split(df)
train_lf = train_df.lazy(implementation)
val_lf = val_df.lazy(implementation)
```

I think we'd be pretty safe to use `backend`, because Polars uses [engine](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.collect.html) and I can't see why they'd add both

Regarding `ModuleType | Implementation | str | None`, I think we can accept all of:
- module (e.g. `polars`, `duckdb`, ...)
- implementation (e.g. `nw.Implementation.POLARS`)
- str (e.g. `'polars'`)
- None: use the default eager/lazy backend for the given module (we can document sensible defaults)

We should probably accept all the above in `from_dict` and similars too, and perhaps deprecate `native_namespace` which is annoyingly long

FBruzzesi:
Thanks @MarcoGorelli, we can do some brainstorming in the community call tomorrow maybe.

> * as a tool-builder wanting to avoid extensive `if/then` statement, I could do:
> 
> ```python
> collect_kwargs = {
>     nw.dependencies.get_polars(): {'engine: 'gpu'},
>     nw.dependencies.get_dask(): {'optimize_graph': False},
>     nw.dependencies.get_pyarrow(): {'backend': 'pyarrow'},
>     my_fancy_extension_module: {'my_fancy_kwarg': my_fancy_value},
> }
>  lazy_df = df.lazy().select(nw.all().sum()).collect(**collect_kwargs[nw.get_native_namespace(df)])
> ```

I am ok with this - it might not be the most ergonomic way but it definitely lowers the effort a lot from our side (we would just need to pass `kwargs` along

> I think we'd be pretty safe to use `backend`, because Polars uses [engine](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.collect.html) and I can't see why they'd add both
> 
> Regarding `ModuleType | Implementation | str | None`, I think we can accept all of:
> 
> * module (e.g. `polars`, `duckdb`, ...)
> * implementation (e.g. `nw.Implementation.POLARS`)
> * str (e.g. `'polars'`)
> * None: use the default eager/lazy backend for the given module (we can document sensible defaults)

Am I correctly assuming this refers to which custom collect backend we want to allow for duckdb/pyspark? Or you would like it to be more general?

Since polars and dask have a native `collect`s backend, I would rather keep that, but I might be missing some use cases.
For pyspark and duckdb I am totally open to customize their available kwargs to allow for `backend` (with pyarrow as default?)

MarcoGorelli:
I was thinking something like:
- `backend` not specified: the default one is used, and we document what that means. This would be:
  - `polars.LazyFrame` -> `polars.DataFrame`
  - `dask.DataFrame` -> `pandas.DataFrame`
  - `duckdb.PyRelation` -> `pyarrow.Table`
  - `pyspark` -> probably `pyarrow.Table`?
- `backend` specified: we collect into the desired eager implementation.

So:
- if someone calls `.collect('pandas')` on a polars.LazyFrame-backed `narwhals.LazyFrame`, then we give them a pandas-backed `narwhals.DataFrame`
- if someone calls `.collect()` on a polars.LazyFrame-backed `narwhals.LazyFrame`, then we give them a `polars.DataFrame`-backed `narwhals.DataFrame`

FBruzzesi:
I am almost finished with the reimplementation. I have just one question left for now.
Since `PandasLikeDataFrame` also implements `.collect()` method, should we return self for `backend=None` or default to pandas?

MarcoGorelli:
I'd say, `self`, so that `from_native(df).lazy().collect()` if starting from cuDF round-trips, right?

narwhals/_spark_like/dataframe.py
```diff
@@ -112,17 +115,53 @@ def _from_native_frame(self: Self, df: DataFrame) -> Self:
...
-            implementation=Implementation.PANDAS,
-            backend_version=parse_version(pd.__version__),
-            version=self._version,
-        )
+            from narwhals._arrow.dataframe import ArrowDataFrame
+
+            return ArrowDataFrame(
+                native_dataframe=pa.Table.from_batches(
+                    self._native_frame._collect_as_arrow()
+                ),
```
FBruzzesi, line 143: I am defaulting to pandas instead of pyarrow because otherwise some tests will fail. Specifically those that result in an empty dataframe - pyarrow `from_batches` raises that RecordBatches cannot be empty. Maybe once pyspark 4.0 is out this could be better integrated?

narwhals/_spark_like/dataframe.py
```diff
@@ -112,17 +115,53 @@ def _from_native_frame(self: Self, df: DataFrame) -> Self:
...
+
+        elif backend in ("polars", Implementation.POLARS, get_polars()):
+            import polars as pl  # ignore-banned-import
+            import pyarrow as pa  # ignore-banned-import
+
+            from narwhals._polars.dataframe import PolarsDataFrame
+
+            return PolarsDataFrame(
+                df=pl.from_arrow(  # type: ignore[arg-type]
+                    pa.Table.from_batches(self._native_frame._collect_as_arrow())
```
FBruzzesi, line 156: Ritchie answer in [stackoverflow](https://stackoverflow.com/a/73205690/12411536)

narwhals/_spark_like/dataframe.py
MarcoGorelli: this would technically be a breaking change - can we catch the exception and return an empty dataframe?

FBruzzesi: Sure I will take a look. Notice that changing to pyarrow now is also a breaking change 🙈

MarcoGorelli: oh right we currently collect into pandas? 

🙈

MarcoGorelli: we've documented pyspark support as "work-in-progress" so we may not need to feel too guilty about changing this 😄

FBruzzesi: yes correct 🙃

narwhals/utils.py
```diff
@@ -110,6 +110,51 @@ def from_native_namespace(
...
+            "modin": Implementation.MODIN,
+            "cudf": Implementation.CUDF,
+            "pyarrow": Implementation.PYARROW,
+            "pyspark": Implementation.PYSPARK,
+            "polars": Implementation.POLARS,
+            "dask": Implementation.DASK,
+            "duckdb": Implementation.DUCKDB,
+            "ibis": Implementation.IBIS,
+        }
+        return mapping.get(backend_name, Implementation.UNKNOWN)
```
dangotbanned, line 136: I was going to suggest this, after adding something similar in (https://github.com/vega/altair/blob/94220be0115e8b13d2ebc686552edf68fd841a54/altair/datasets/_reader.py#L493-L510)

Great to see it in `narhwals` 🎉

MarcoGorelli (approved):
Amazing, thanks so much @FBruzzesi ! I pushed a little commit to simplify backend parsing, to default to PyArrow for PySpark, and to handle the (hopefully rare) empty input cases

I noticed something here regarding collecting to different backends: going from pyspark to pandas loses the timezone-awareness of the dtype, but for pyarrow it's preserved:
```python
(Pdb) p self._native_frame
DataFrame[b: timestamp]
(Pdb) p self._native_frame.toPandas()
                    b
0 2020-01-01 12:34:56
(Pdb) p self._native_frame._collect_as_arrow()
[pyarrow.RecordBatch
b: timestamp[us, tz=UTC]
----
b: [2020-01-01 12:34:56.000000Z]]
```
This helped spot that in `to_datetime` we're not parsing timezone-naive formats as timezone-naive

To me this kinda confirms that PyArrow is probably a better default

narwhals/utils.py
MarcoGorelli: 😄 nice, you're one step ahead

minor comment but I couldn't help spotting the "narhwals" typo in there (line 509) 🙈

dangotbanned: ooh well spotted, thanks @MarcoGorelli

FBruzzesi:
Thanks @MarcoGorelli for all the additional improvement! I am very excited to finally ship this one and enabling `collect` kwargs ✨

MarcoGorelli:
thanks all! let's ship it then

PR #1749, opened 2025-01-07
build: Remove docstring examples before building in `publish_to_pypi.yml`

### LINKED ISSUE #1742: build: remove docstring examples when publishing packages?
Docstring examples account for ~~30%~~ 40% of Narwhals' size

We love docs, but maybe this is a bit excessive 😄

I think it's pretty important to ship docstrings with source code, so that people can inspect docstrings while they're coding. But do we also need the docstring examples? I think they're kind of hard to read just from the IDE's preview anyway

Perhaps only removing the examples from the docstrings when building the wheel strikes a balance between:
- minimising size (completely minify all code)
- good developer experience

The docstring examples would still be visible in the [API Reference](https://narwhals-dev.github.io/narwhals/api-reference/)

----

I just tried this out, and removing all examples from functions is enough to reduce the package size by ~~25%~~ 40%

I also tried removing the "single-line-imports" rule, but that made a barely detectable difference (less than 0.1MB) - sometimes it actually becomes longer to have imports on multiple lines

MarcoGorelli:
closes #1742 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -34,7 +34,6 @@ jobs:
     - name: minify-narwhals
       run: |
         git ls-files narwhals | xargs python utils/remove_docstring_examples.py
-        cat narwhals/expr.py | head -n 100
     - name: Build a binary wheel and a source tarball
       run: python3 -m build
     - name: Store the distribution packages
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
closing as i think that if we do this, it would be better to _add_ the docstrings before buildings docs, than to remove them before publishing the package

PR #1761, opened 2025-01-08
chore: move pyspark tests into main test suite

### LINKED ISSUE #1755: [Enh]: Move spark tests and constructor into main test suite
It might be time to move spark constructor into the main test suite.
I had some hard times trying to accomplish this and re-use the pyspark session fixture 🙈

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #1755 

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_spark_like/dataframe.py
```diff
@@ -106,9 +108,11 @@ def select(
         new_columns_list = [col.alias(col_name) for col_name, col in new_columns.items()]
         return self._from_native_frame(self._native_frame.select(*new_columns_list))
 
-    def filter(self, *predicates: SparkLikeExpr) -> Self:
+    def filter(self, *predicates: SparkLikeExpr, **constraints: Any) -> Self:
         plx = self.__narwhals_namespace__()
-        expr = plx.all_horizontal(*predicates)
+        expr = plx.all_horizontal(
+            *chain(predicates, (plx.col(name) == v for name, v in constraints.items()))
+        )
```
FBruzzesi, line 115: Needed to implement `Expr.__eq__` to get this to work. It overlaps with @EdAbati PR

narwhals/_spark_like/group_by.py
```diff
@@ -146,6 +146,8 @@ def agg_pyspark(
         simple_aggregations.update(
             {
                 output_name: agg_func(root_name)
+                if function_name != "len"
+                else agg_func("*")
```
FBruzzesi, line 150: `F.count` ignores null's. Proposing this workaround.

tests/expr_and_series/reduction_test.py
```diff
@@ -62,7 +62,7 @@ def test_scalar_reduction_with_columns(
     expected: dict[str, list[Any]],
     request: pytest.FixtureRequest,
 ) -> None:
-    if "duckdb" in str(constructor):
+    if "duckdb" in str(constructor) or ("pyspark" in str(constructor)):
```
FBruzzesi, line 65: Only one of the 5 tests is passing (leading to `[XPASS(strict)]`)

MarcoGorelli (commented):
awesome @FBruzzesi - not sure what's happening with tests in running in ci?

FBruzzesi:
> awesome @FBruzzesi - not sure what's happening with tests in running in ci?

We are not installing pyspark at all, therefore but now `--all-cpu-constructors` includes pyspark. However, some python version would not support pyspark at all I believe (3.12 and 3.13)

MarcoGorelli:
ah i see - maybe `--all-cpu-constructors` should only include those which are available?

tests/conftest.py
```diff
@@ -201,7 +231,13 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             constructors.append(EAGER_CONSTRUCTORS[constructor])
             constructors_ids.append(constructor)
         elif constructor in LAZY_CONSTRUCTORS:
-            constructors.append(LAZY_CONSTRUCTORS[constructor])
+            if constructor == "pyspark":
+                if sys.version_info < (3, 12):
+                    constructors.append(pyspark_lazy_constructor())
+                else:
+                    continue
```
FBruzzesi, line 238: @MarcoGorelli maybe this is too much? 🙈

tests/conftest.py
```diff
@@ -155,8 +154,38 @@ def spark_session() -> Generator[SparkSession, None, None]:  # pragma: no cover
...
+                    module="pyspark",
+                    category=DeprecationWarning,
+                )
+                warnings.filterwarnings(
+                    "ignore",
+                    r".*is_datetime64tz_dtype is deprecated and will be removed in a future version.*",
+                    module="pyspark",
+                    category=DeprecationWarning,
+                )
+                pd_df = pd.DataFrame(obj).replace({float("nan"): None}).reset_index()
```
camriddell, line 180: If the objects that come into these constructors are (always?) dictionaries I think we can skip the trip through pandas and construct from a built-in Python object that spark knows how to ingest directly (list of dictionaries). Could be overly cautions, but Spark may infer data types differently if it is handed a pandas DataFrame rather than lists of Python objects.

Since pyspark supports a list of records we could convert dict → list of dicts like so
```
if isinstance(obj, dict):
    obj = [{k: v for k, v in zip(obj, row)} for row in zip(*obj.values())]
```

Or could pass in the rows & schema separately

```python
if isinstance(obj, dict):
    df = ...createDataFrame([*zip(*obj.values())], schema=[*obj.keys()])

```

EdAbati (commented):
Thank you very much for doing this! 🙌

tests/conftest.py
EdAbati: I remember having issues with some tests, where we may need to specify the schema with column type. (but I don't remember exactly what was the problem)

But if we can skip pandas here, it would be 👌👌👌

FBruzzesi: I had the same thought when migrating the codebase, yet I can confirm the data type being an issue for a subset of the tests. I would say to keep it like this for now and eventually address it

tests/conftest.py
MarcoGorelli: i think this is fine 👍

pyproject.toml
```diff
@@ -164,11 +164,16 @@ filterwarnings = [
   'ignore:.*Passing a BlockManager to DataFrame:DeprecationWarning',
   # This warning was temporarily raised by Polars but then reverted.
   'ignore:.*The default coalesce behavior of left join will change:DeprecationWarning',
+  'ignore: unclosed <socket.socket',
+  'ignore:.*The distutils package is deprecated and slated for removal in Python 3.12:DeprecationWarning:pyspark',
+  'ignore:.*distutils Version classes are deprecated. Use packaging.version instead.*:DeprecationWarning:pyspark',
```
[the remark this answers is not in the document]
MarcoGorelli: TIL

nice!

FBruzzesi:
> non-tests changes look good
> 
> i'm about to go out so I didn't finish reading through all the changes in the `tests` folder, but if you checked them and there's no rogue changes feel free to merge
> 
> nice one! 🙌

Thanks Marco! Aside for CI time increasing significantly for when pyspark runs (maybe we could skip the windows one and run pyspark on ubuntu only), I don't see a big risk for merging now.

It is such a better developer experience to add features with tests already there 🙈

EdAbati:
Yeees thank you @FBruzzesi 🥳🥳🥳

MarcoGorelli:
> maybe we could skip the windows one and run pyspark on ubuntu only

yes, 👍 to this, windows is already really slow to run...

PR #1787, opened 2025-01-10
fix root names in pyspark reduction with nw.all()

### LINKED ISSUE #1780: [Bug]: Pyspark reduction does not preserve names root names with `nw.all()`
### Describe the bug

Pyspark reductions do not preserve original root name when applied to `nw.all()`.

I am tempted to say that the issue is in [this line](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/_spark_like/expr.py#L105). However I could not manage to fix it without breaking the code elsewhere so far 🙈

Running:

```py
df.select(nw.all().n_unique())
```

with a lazyframe backed by pyspark DataFrame, will result in a column named `(count(DISTINCT a) + max(CAST((a IS NULL) AS INT)))`.

### Expected results

Original names to be preserved

### Actual results

Pyspark expression names

```shell
Latest
```

### Relevant log output

_No response_

camriddell:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1780 

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Checking if the parent expression is the root and if that expression is "all" to also alias the result seems to alleviate this problem in this particular test. 

This may need to be revisited when the `.name` namespace is created for pyspark as I am unsure whether or not this approach play nicely with multi columnar renaming functions like `prefix`, `suffix` or `map`.

### DIFF
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -102,7 +102,9 @@ class SparkLikeExpr(CompliantExpr["Column"]):
             for _input in inputs:
                 input_col_name = get_column_name(df, _input)
                 column_result = call(_input, **_kwargs)
-                if not returns_scalar:
+                if not returns_scalar or (
+                    (self._depth == 0) and (self._function_name == "all")
+                ):
                     column_result = column_result.alias(input_col_name)
                 results.append(column_result)
             return results
--- a/tests/expr_and_series/n_unique_test.py
+++ b/tests/expr_and_series/n_unique_test.py
@@ -13,7 +13,7 @@ data = {
 
 def test_n_unique(constructor: Constructor) -> None:
     df = nw.from_native(constructor(data))
-    result = df.select(nw.col("a", "b").n_unique())
+    result = df.select(nw.all().n_unique())
     expected = {"a": [3], "b": [4]}
     assert_equal_data(result, expected)
 


narwhals/_spark_like/expr.py
```diff
@@ -102,7 +102,9 @@ def func(df: SparkLikeLazyFrame) -> list[Column]:
             for _input in inputs:
                 input_col_name = get_column_name(df, _input)
                 column_result = call(_input, **_kwargs)
-                if not returns_scalar:
+                if not returns_scalar or (
+                    (self._depth == 0) and (self._function_name == "all")
```
MarcoGorelli, line 106: thanks for looking into this!

I think this might be too specific and that we'll need a more generic solution - will look into it and let you know

MarcoGorelli:
Rough plan of attack could be:
- aggregations should just be carried out as-is. For example, `expr.sum()` just becomes `F.sum(expr)`
- in binary operations between multiple expressions, if one of them aggregates but the other one doesn't, then the one that aggregates gets broadcasted via `.over(Window(F.lit(1)))` and the resulting expression gets `returns_scalar=False`
- in `select`, if everything aggregates, just use `.agg`. Else, those which aggregate get an extra `.over(Window(F.lit(1)))`
- in `group-by` aggregations, everything is expected to aggregate

So then, we can support:

- everything aggregates (e.g. `df.select(nw.col('a').min(), nw.col('b').max())`)
- some aggregate, and some don't (e.g. `df.select('a', nw.col('b').max())`):
```python
   ...: select a, max(b) over ()
   ...: from df
```
- binary operations between some expressions which aggregate and others which don't, e.g. `df.select((nw.col('a') - nw.col('a').mean()).round(2))`:
```python
   ...: select round(a - (mean(a) over ()), 2)
   ...: from rel
```
- group-by, e.g. `df.group_by('a').agg(nw.col('b').mean())
```python
    ...: select a, mean(b)
    ...: from rel
    ...: group by a
```

This would require some refactoring, but it would be well-worth it

PR #1791, opened 2025-01-11
feat: deprecate Expr.head, Expr.tail, Expr.sort, Expr.gather_every, Expr.sample (but keep them in stable.v1)

MarcoGorelli:
One step closer to `narwhals.stable.v2`

These are some operations which we won't be able to support in the lazy layer at all, and I think it would be confusing to have some Expr methods which aren't available at all for lazyframes (which would be different to `Expr.unique`, which could be supported if followed by an aggregation, or `Expr.cum_sum`, which could be supported if `order_by` is passed)

Further, I just don't think any of these expressions are very useful...people call head/tail/sort/sample on dataframes, not typically on individual expressions..

For the cases where people need them, they can use DataFrame or Series operations

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/expr_and_series/gather_every_test.py
+++ b/tests/expr_and_series/gather_every_test.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import pytest
 
+import narwhals as nw_main
 import narwhals.stable.v1 as nw
 from tests.utils import ConstructorEager
 from tests.utils import assert_equal_data
@@ -20,6 +21,9 @@ def test_gather_every_expr(
     expected = {"a": data["a"][offset::n]}
 
     assert_equal_data(result, expected)
+
+    with pytest.deprecated_call():
+        df.select(nw_main.col("a").gather_every(n=n, offset=offset))
 
 
 @pytest.mark.parametrize("n", [1, 2, 3])
--- a/tests/expr_and_series/tail_test.py
+++ b/tests/expr_and_series/tail_test.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 import pytest
 
-import narwhals as nw
+import narwhals as nw_main
+import narwhals.stable.v1 as nw
 from tests.utils import ConstructorEager
 from tests.utils import assert_equal_data
 
 
 @pytest.mark.parametrize("n", [2, -1])
-def test_head(
+def test_tail(
     constructor_eager: ConstructorEager, n: int, request: pytest.FixtureRequest
 ) -> None:
     if "polars" in str(constructor_eager) and n < 0:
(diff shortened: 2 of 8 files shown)


MarcoGorelli:
tubular do:
```python
levels = X.select(nw.col(c).unique())
```
which could be replaced with `levels = X.select(nw.col(c)).unique()`

tests/expr_and_series/tail_test.py
```diff
@@ -2,13 +2,14 @@
...
-import narwhals as nw
+import narwhals as nw_main
+import narwhals.stable.v1 as nw
 from tests.utils import ConstructorEager
 from tests.utils import assert_equal_data
 
 
 @pytest.mark.parametrize("n", [2, -1])
-def test_head(
+def test_tail(
```
MarcoGorelli, line 12: lol

EdAbati (approved):
I think it makes a lot of sense! It looks 👌

tests/expr_and_series/gather_every_test.py
```diff
@@ -21,6 +22,9 @@ def test_gather_every_expr(
 
     assert_equal_data(result, expected)
 
+    with pytest.deprecated_call():
```
FBruzzesi, line 25: TIL!

FBruzzesi (approved):
I am all in favor for deprecating head, tail and gather every. I still think that `sort` and `sample` might have valid use cases, although hard/impossible to implement in some backends

MarcoGorelli:
thanks all for feedback and for taking a look! i think more than implementation difficulties they're just not relational, e.g.
```python
In [16]: df
Out[16]:
shape: (3, 2)
┌───────┬─────┐
│ name  ┆ age │
│ ---   ┆ --- │
│ str   ┆ i64 │
╞═══════╪═════╡
│ Mario ┆ 31  │
│ Luigi ┆ 32  │
│ Wario ┆ 45  │
└───────┴─────┘

In [17]: df.select('name', pl.col('age').sort(descending=True))
Out[17]:
shape: (3, 2)
┌───────┬─────┐
│ name  ┆ age │
│ ---   ┆ --- │
│ str   ┆ i64 │
╞═══════╪═════╡
│ Mario ┆ 45  │
│ Luigi ┆ 32  │
│ Wario ┆ 31  │
└───────┴─────┘
```
you end up with tuples which aren't part of the original object - it's fine for Polars to allow it, but I'm not sure we should be allowing this kind of operation in Narwhals

In any case, happy to bring these back if use cases do come along!

PR #1802, opened 2025-01-12
feat: add `drop_nulls`, `fill_null`, and `filter` Spark Expressions

lucas-nelson-uiuc:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Builds on #1714 

## Checklist

- [x] Tests added
- [ ] Documented the changes

Added support for the following methods - made sure to remove all pyspark constructor checks in the main test suite:
- `drop_nulls`
- `fill_null`
  - noticed the only strategies currently supported in `nw.Expr.fill_null` are "forward" and "backward" - are there plans to add more strategies (possibly in v2)?
- `filter` (_work in progress_)
  - solution works fine in PySpark but doesn't carry over to narwhals (`pyspark.sql.function.filter` expects the predicate to return a `Column`; however, it raises an error since `_from_call` returns `SparkLikeExpr`).
  - lmk if I'm missing something basic

### DIFF
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -316,19 +316,21 @@ class SparkLikeExpr(CompliantExpr["Column"]):
 
             # TODO(unauthored): can only specify `limit` when strategy is set to 'backward' or 'forward'
 
+            fill_value: Column | Any
+
             if strategy is not None:
                 match strategy:
                     case "zero":
-                        fill_value = F.lit(0)
+                        fill_value = 0
                     case "one":
-                        fill_value = F.lit(1)
+                        fill_value = 1
                     case _:
                         msg = f"strategy not supported: {strategy}"
                         raise ValueError(msg)
             else:
                 fill_value = value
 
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
thanks for your pr, really appreciate it!

to be honest i'm not totally sold on having expressions which change length but don't aggregate be supported in the lazy api

we may be able to support them, but only if they're followed by an aggregations, so they're kinda like the equivalent of sql's
```
select sum(a) filter (where b > 1)
from rel
```

Will get back to you on this anyway, but for now we may need to punt on these methods, really sorry

lucas-nelson-uiuc:
No worries - kinda felt like a cheat move anyway so not torn to see it punted.

FBruzzesi:
Hey @lucas-nelson-uiuc thanks for the PR, feel free to split it for `fill_null`, which we definitly want to support (without filling strategies)

PR #1806, opened 2025-01-13
feat: Disallow order-dependent expressions from being passed to nw.LazyFrame

MarcoGorelli:
Another step closer to stable.v2

In a future version, we can add `order_by` to many of these, and then parametrise over all constructors without issues

This is work in progress

Technically this would (breakingly) remove some methods for Dask, so the current behaviour could be preserved for `stable.v1`. ~~I'm just not sure if it's worth it, given we don't seem to have any users for the Dask backend. ~~~~We could just make the change, and if anyone complains we keep these available for Dask in v1~~ and that's exactly what we do in this PR by overwriting `_extract_compliant` in v1

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -3629,7 +3629,7 @@ class LazyFrame(BaseFrame[FrameT]):
         from narwhals.expr import Expr
         from narwhals.series import Series
 
-        if isinstance(arg, BaseFrame):  # pragma: no cover
+        if isinstance(arg, BaseFrame):
             return arg._compliant_frame
         if isinstance(arg, Series):  # pragma: no cover
             msg = "Binary operations between Series and LazyFrame are not supported."
@@ -3655,7 +3655,9 @@ class LazyFrame(BaseFrame[FrameT]):
                 "- Used `pl.col` instead of `nw.col`?"
             )
             raise TypeError(msg)
-        return arg
+        # TODO(unassigned): should this line even be reachable? Should we
+        # be raising here?
+        return arg  # pragma: no cover
 
     @property
     def _dataframe(self) -> type[DataFrame[Any]]:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -3629,10 +3629,11 @@ class LazyFrame(BaseFrame[FrameT]):
         from narwhals.expr import Expr
         from narwhals.series import Series
 
-        if isinstance(arg, BaseFrame):
-            return arg._compliant_frame
+        if isinstance(arg, BaseFrame):  # pragma: no cover
+            msg = "Binary operations between LazyFrames are not supported"
+            raise TypeError(msg)
         if isinstance(arg, Series):  # pragma: no cover
-            msg = "Mixing Series with LazyFrame is not supported."
+            msg = "Binary operations between Series and LazyFrame are not supported."
             raise TypeError(msg)
         if isinstance(arg, Expr):
             if arg._is_order_dependent:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -63,31 +63,14 @@ class BaseFrame(Generic[FrameT]):
             level=self._level,
         )
 
+    def _extract_compliant(self, arg: Any) -> Any:
+        raise NotImplementedError
+
     def _flatten_and_extract(self, *args: Any, **kwargs: Any) -> Any:
         """Process `args` and `kwargs`, extracting underlying objects as we go."""
         args = [self._extract_compliant(v) for v in flatten(args)]  # type: ignore[assignment]
         kwargs = {k: self._extract_compliant(v) for k, v in kwargs.items()}
         return args, kwargs
-
-    def _extract_compliant(self, arg: Any) -> Any:
-        from narwhals.expr import Expr
-        from narwhals.series import Series
-
-        if isinstance(arg, BaseFrame):
-            return arg._compliant_frame
-        if isinstance(arg, Series):
-            return arg._compliant_series
-        if isinstance(arg, Expr):
-            return arg._to_compliant_expr(self.__narwhals_namespace__())
-        if get_polars() is not None and "polars" in str(type(arg)):
-            msg = (
-                f"Expected Narwhals object, got: {type(arg)}.\n\n"
-                "Perhaps you:\n"
-                "- Forgot a `nw.from_native` somewhere?\n"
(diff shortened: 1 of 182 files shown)


MarcoGorelli:
something's gone terribly wrong

narwhals/functions.py
```diff
@@ -1902,7 +1902,8 @@ def sum_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
     return Expr(
         lambda plx: plx.sum_horizontal(
             *[extract_compliant(plx, v) for v in flatten(exprs)]
-        )
+        ),
+        is_order_dependent=False,
```
MarcoGorelli, line 1906: this should probably be "any" of its inputs

narwhals/functions.py
```diff
@@ -1972,7 +1973,8 @@ def min_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
     return Expr(
         lambda plx: plx.min_horizontal(
             *[extract_compliant(plx, v) for v in flatten(exprs)]
-        )
+        ),
+        is_order_dependent=False,
```
MarcoGorelli, line 1977: same

narwhals/functions.py
```diff
@@ -2222,7 +2230,8 @@ def all_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
     return Expr(
         lambda plx: plx.all_horizontal(
             *[extract_compliant(plx, v) for v in flatten(exprs)]
-        )
+        ),
+        is_order_dependent=False,
```
MarcoGorelli, line 2234: any of its inputs

narwhals/functions.py
```diff
@@ -2440,7 +2450,8 @@ def mean_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
     return Expr(
         lambda plx: plx.mean_horizontal(
             *[extract_compliant(plx, v) for v in flatten(exprs)]
-        )
+        ),
+        is_order_dependent=False,
```
MarcoGorelli, line 2454: same

narwhals/functions.py
```diff
@@ -2528,5 +2539,6 @@ def concat_str(
             *[extract_compliant(plx, v) for v in more_exprs],
             separator=separator,
             ignore_nulls=ignore_nulls,
-        )
+        ),
+        is_order_dependent=False,
```
MarcoGorelli, line 2543: same

narwhals/dataframe.py
```diff
@@ -64,32 +65,16 @@ def _from_compliant_dataframe(self, df: Any) -> Self:
...
+    def _extract_compliant(self, arg: Any) -> Any:
+        raise NotImplementedError
+
     def _flatten_and_extract(self, *args: Any, **kwargs: Any) -> Any:
         """Process `args` and `kwargs`, extracting underlying objects as we go."""
         args = [self._extract_compliant(v) for v in flatten(args)]  # type: ignore[assignment]
         kwargs = {k: self._extract_compliant(v) for k, v in kwargs.items()}
         return args, kwargs
 
-    def _extract_compliant(self, arg: Any) -> Any:
```
MarcoGorelli, line 73: instead of defining this in BaseFrame, we define it in DataFrame and LazyFrame so they can have different behaviours

narwhals/dataframe.py
```diff
@@ -3621,6 +3626,40 @@ class LazyFrame(BaseFrame[FrameT]):
...
+                msg = (
+                    "Order-dependent expressions are not supported for use in LazyFrame.\n\n"
+                    "Hints:\n"
+                    "- Instead of `lf.select(nw.col('a').sort())`, use `lf.select('a').sort()\n"
+                    "- Instead of `lf.select(nw.col('a').head())`, use `lf.select('a').head()\n"
+                    "- `Expr.cum_sum`, and other such expressions, are not currently supported.\n"
+                    "  In a future version of Narwhals, a `order_by` argument will be added and \n"
+                    "  they will be supported."
+                )
+                raise TypeError(msg)
```
FBruzzesi, line 3649: What about an `AssertionError` or having a dedicated exception to this instead?

narwhals/expr.py
```diff
@@ -873,7 +963,8 @@ def map_batches(
         return self.__class__(
             lambda plx: self._to_compliant_expr(plx).map_batches(
                 function=function, return_dtype=return_dtype
-            )
+            ),
+            is_order_dependent=True,  # safest assumption
```
FBruzzesi, line 967: what if we add an extra param to the function for the user to determine 🤔?

narwhals/expr.py
```diff
@@ -1715,7 +1837,8 @@ def sort(self, *, descending: bool = False, nulls_last: bool = False) -> Self:
         return self.__class__(
             lambda plx: self._to_compliant_expr(plx).sort(
                 descending=descending, nulls_last=nulls_last
-            )
+            ),
+            is_order_dependent=True,
```
FBruzzesi, line 1841: We removed `sort` anyway. But.. is `sort` order dependent?

FBruzzesi (commented):
Thanks @MarcoGorelli! I have a few tiny comments. In general I am somehow feeling we are restricting the API a lot - but this might be relevant just for lazy polars anyway.

narwhals/dataframe.py
MarcoGorelli: sure, i've introduced an exception

narwhals/expr.py
MarcoGorelli: i think this function is only useful/implemented for eager backends anyway, so for now not sure we should bother

narwhals/expr.py
MarcoGorelli: probably 😄 but yes, if it's being removed it may not matter

MarcoGorelli:
thanks for your review!

> this might be relevant just for lazy polars anyway.

yup, exactly - eager backends are unaffected

and for lazy ones, if we have something for which polars is the only one which supports it, then that makes writing universal code quite challenging

PR #1810, opened 2025-01-14
patch: fix when-then double lit case

### LINKED ISSUE #1722: test: check if we have a test for `nw.when(nw.col('a')>1).then(nw.lit(0)).otherwise(nw.lit(1))`
Do we have a case where both are literals?

It would be good to investigate this, and add a test if necessary

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1722 

## Checklist

- [x] Tests added
- [ ] Documented the changes

Definitly not too proud of this implementation

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -439,8 +439,7 @@ class ArrowWhen:
         try:
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
         except TypeError:
-            # `self._otherwise_value` is a scalar and can't be converted to an expression
-            plx = condition.__narwhals_namespace__()
+            # `self._then_value` is a scalar and can't be converted to an expression
             value_series = plx._create_series_from_scalar(
                 self._then_value, reference_series=condition
             )
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -437,6 +437,14 @@ class ArrowWhen:
         condition = parse_into_expr(self._condition, namespace=plx)(df)[0]
         try:
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
+            if len(value_series) == 1:  # literal case
+                value_series = condition.__class__._from_iterable(
+                    pa.repeat(pa.scalar(value_series[0]), len(condition)),
+                    name="literal",
+                    backend_version=self._backend_version,
+                    version=self._version,
+                )
+
         except TypeError:
             # `self._otherwise_value` is a scalar and can't be converted to an expression
             value_series = condition.__class__._from_iterable(
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -435,27 +435,19 @@ class ArrowWhen:
 
         plx = df.__narwhals_namespace__()
         condition = parse_into_expr(self._condition, namespace=plx)(df)[0]
+
         try:
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
-            if len(value_series) == 1:  # literal or reduction case
-                value_series = condition.__class__._from_iterable(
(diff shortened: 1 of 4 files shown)


narwhals/_arrow/namespace.py
```diff
@@ -437,6 +437,14 @@ def __call__(self: Self, df: ArrowDataFrame) -> Sequence[ArrowSeries]:
         condition = parse_into_expr(self._condition, namespace=plx)(df)[0]
         try:
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
+            if len(value_series) == 1:  # literal case
+                value_series = condition.__class__._from_iterable(
+                    pa.repeat(pa.scalar(value_series[0]), len(condition)),
+                    name="literal",
+                    backend_version=self._backend_version,
+                    version=self._version,
+                )
```
MarcoGorelli, line 446: does it work to use `broadcast_series` like we do below?

FBruzzesi: I think it would look like:

```py
if len(value_series) == 1:  # literal case
    _, value_series_native = broadcast_series(
        [condition, value_series]
    )
    value_series = condition.__class__(
        value_series_native,
        name="literal",
        version=self._version,
    )
```
Not sure how much cleaner this is 🙃

narwhals/_dask/namespace.py
```diff
@@ -398,8 +398,13 @@ def __call__(self, df: DaskLazyFrame) -> Sequence[dx.Series]:
         plx = df.__narwhals_namespace__()
         condition = parse_into_expr(self._condition, namespace=plx)(df)[0]
         condition = cast("dx.Series", condition)
+
         try:
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
+            if getattr(self._then_value, "_function_name", None) == "lit":
```
MarcoGorelli, line 404: i think this should be
```python
            then_expr = parse_into_expr(self._then_value, namespace=plx)
            value_series = then_expr(df)[0]
            if then_expr._returns_scalar:
```
?

narwhals/_dask/namespace.py
```diff
@@ -398,8 +398,15 @@ def __call__(self, df: DaskLazyFrame) -> Sequence[dx.Series]:
...
+
         try:
-            value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
+            then_expr = parse_into_expr(self._then_value, namespace=plx)
+            value_series = then_expr(df)[0]
+            # literal or reduction case
+            if then_expr._returns_scalar:  # type: ignore[attr-defined]
+                _df = condition.to_frame("a")
+                _df["tmp"] = value_series[0]
+                value_series = _df["tmp"]
```
MarcoGorelli, line 409: this is too big to be in a `try` clause, it should either be in `else`, or outside of try-except

narwhals/_pandas_like/namespace.py
```diff
@@ -463,6 +463,15 @@ def __call__(self, df: PandasLikeDataFrame) -> Sequence[PandasLikeSeries]:
...
             value_series = parse_into_expr(self._then_value, namespace=plx)(df)[0]
+            if len(value_series) == 1:  # literal or reduction case
+                value_series = condition.__class__._from_iterable(
+                    [value_series[0]] * len(condition),
+                    name="literal",
+                    index=condition._native_series.index,
+                    implementation=self._implementation,
+                    backend_version=self._backend_version,
+                    version=self._version,
+                )
```
MarcoGorelli, line 474: same as in the pyarrow case, we can simplify here

MarcoGorelli (approved):
thanks @FBruzzesi !

Issue #1743, opened 2025-01-06
[Enh]: cast expr in `SparkLike`


lucas-nelson-uiuc:
### Please describe the purpose of the new feature or describe the problem to solve.

Currently working on implementing the `cast` expression in `SparkLike` - wanted to take this issue to discuss + list out all the key considerations with Spark's data types.

### Suggest a solution if possible.

So far, I'm able to implement a handful of data types. However, I noticed that some types cannot (yet) be implemented or I'm uncertain how they'd be implemented.

**Able to implement (currently testing)**
- `Float64`, `Float32`, `Int64`, `Int32`, `Int16`, `Decimal`
- `String`
- `Boolean`
- `ArrayType` (see comment below)
- `Struct`, `Field`

**Cannot (yet) implement**
- No native support for unsigned integers (`UInt8`, `UInt16`, `UInt32`, `UInt64`)
- No native support for categorical types (`Enum`, `Categorical`)
- No direct parameterization of datetime information (would need to set in/get from [Spark configuration](https://spark.apache.org/docs/latest/api/python/reference/api/pyspark.SparkConf.get.html#pyspark.SparkConf.get))
- No native support for other dtypes (`Object`, `Unknown`)

**Unsure how to implement**
- `pyspark.types.StructField` contains more than just `name` and `dtype` - is it worth updating `narwhals.dtypes.Field` to have these additional (optional) parameters to accommodate PySpark?
  - `nullable`: whether the field can be null (None) or not
  - `metadata`: additional information about the field
- I think PySpark's `ArrayType` functions like Polars' `List` type (at least it doesn't have a width constraint)
  - Should we map `pyspark.types.ArrayType` to `narwhals.dtypes.List`?
  - Is there a way we could implement an `Array` type that aligns with Polars?

Let me know what anyone thinks about the above. Feel free to add onto this with other types we could add in. Thanks!

FBruzzesi:
Hey @lucas-nelson-uiuc thanks for taking the time in this. I would be very happy to see `cast` implementation for pyspark as well.

Coming to your considerations:

> Able to implement (currently testing)

Amazing 🚀👌

> Cannot (yet) implement

I think it is fine to raise an error when casting to these types for now and evaluate later on.


> pyspark.types.StructField contains more than just name and dtype - is it worth updating narwhals.dtypes.Field to have these additional (optional) parameters to accommodate PySpark?

polars should default to nullable, so we can keep such flag on, not sure how to go about having metadata 

[...]

PR #1812, opened 2025-01-15
feat: add `SparkExpr.cast` for basic types

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1743 (just noticed this issue 😕 this PR doesn't close it fully though, we're are still missing more complex dtypes)

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/_spark_like/utils.py
```diff
@@ -35,6 +37,8 @@ def native_to_narwhals_dtype(
         return dtypes.Int32()
     if isinstance(dtype, pyspark_types.ShortType):
         return dtypes.Int16()
+    if isinstance(dtype, pyspark_types.ByteType):
+        return dtypes.Int8()
```
EdAbati, line 41: I remember this came up already @MarcoGorelli . I found the source: https://spark.apache.org/docs/latest/sql-ref-datatypes.html  "`ByteType` is a 1-byte signed integer (within the range of -128 to 127)."  :D also [in the  `spark.pandas` code](https://github.com/apache/spark/blob/master/python/pyspark/pandas/typedef/typehints.py#L258-L259)

narwhals/_spark_like/utils.py
```diff
@@ -58,6 +62,45 @@ def native_to_narwhals_dtype(
...
+    if isinstance_or_issubclass(dtype, dtypes.Int16):
+        return pyspark_types.ShortType()
+    if isinstance_or_issubclass(dtype, dtypes.Int8):
+        return pyspark_types.ByteType()
+    if isinstance_or_issubclass(dtype, dtypes.String):
+        return pyspark_types.StringType()
+    if isinstance_or_issubclass(dtype, dtypes.Boolean):
+        return pyspark_types.BooleanType()
+    if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        return pyspark_types.TimestampType()
```
EdAbati, line 89: a bit unsure if we should add this given that `test_cast_datetime_tz_aware` is not working yet. 

Should we raise a `NotImplementedError`?

narwhals/_spark_like/utils.py
```diff
@@ -58,6 +62,45 @@ def native_to_narwhals_dtype(
...
+        return NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Struct):  # pragma: no cover
+        msg = "Converting to Struct dtype is not supported yet"
+        return NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover
+        msg = "Converting to Array dtype is not supported yet"
+        return NotImplementedError(msg)
+
+    msg = f"Unknown dtype: {dtype}"  # pragma: no cover
+    raise AssertionError(msg)
```
FBruzzesi, line 101: Should we raise a NotImplementedError for unsigned integer as well?
I would prefer that rather than falling into the unknown types

narwhals/_spark_like/utils.py
```diff
@@ -58,6 +62,45 @@ def native_to_narwhals_dtype(
...
+        return pyspark_types.TimestampType()
+    if isinstance_or_issubclass(dtype, dtypes.List):  # pragma: no cover
+        msg = "Converting to List dtype is not supported yet"
+        return NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Struct):  # pragma: no cover
+        msg = "Converting to Struct dtype is not supported yet"
+        return NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover
+        msg = "Converting to Array dtype is not supported yet"
+        return NotImplementedError(msg)
```
FBruzzesi, line 98:
```suggestion
        raise NotImplementedError(msg)
    if isinstance_or_issubclass(dtype, dtypes.Struct):  # pragma: no cover
        msg = "Converting to Struct dtype is not supported yet"
        raise NotImplementedError(msg)
    if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover
        msg = "Converting to Array dtype is not supported yet"
        raise NotImplementedError(msg)
```

narwhals/_spark_like/utils.py
FBruzzesi: I would keep it out for now, as this implementation might lead to the wrong outcome if Datetime with timezone is provided.

tests/expr_and_series/cast_test.py
```diff
@@ -69,28 +71,20 @@ def test_cast(
...
-        nw.col("h").cast(nw.UInt64),
-        nw.col("i").cast(nw.Float32),
-        nw.col("j").cast(nw.Float64),
-        nw.col("k").cast(nw.String),
-        nw.col("l").cast(nw.Datetime),
-        nw.col("m").cast(nw.Int8),
-        nw.col("n").cast(nw.Int8),
-        nw.col("o").cast(nw.String),
-        nw.col("p").cast(nw.Duration),
+        nw.col(col_).cast(dtype) for col_, dtype in schema.items()
```
FBruzzesi, line 84: 🎻🎻

FBruzzesi (commented):
Thanks @EdAbati ! I left a few comments, but really looking forward to merge this 🚀

narwhals/_spark_like/utils.py
EdAbati: lool (let's blame copilot for this one 🥲)

narwhals/_spark_like/utils.py
EdAbati: Sounds good! it is the same thing we do in duckdb for Categorical types.

I think it may be useful to distinguish between:
- error we raise because we haven't implemented the type yet (-> `NotImplementedError`)
- error we raise because the type is not implemented by the backend (-> e.g.`UnsupportedDTypeError`?)

What do you think? Is it something that could be useful?

FBruzzesi: I like the idea, it would definitly be more clear from a user perspective

FBruzzesi (approved):
Thanks @EdAbati 

NarwhalDolphinGIF

PR #1814, opened 2025-01-15
fix: raise on shape mismatch in filter

### LINKED ISSUE #1786: qst - why does `filter(nw.col('temp').max() > 60)` sometimes raise and sometimes not?
from discord:
```python
import pyarrow as pa
import narwhals as nw

t = pa.Table.from_pydict({
  "date": ["2025-01-01", "2025-02-01", "2025-03-01", "2025-04-01"],
  "temperature": [50, 60, 70, 80],
})

nw_f = nw.from_native(t)
column = "date"

print(
    nw_f
    .filter(
        nw.col("temperature").max() > 60,
    )
)
```

I think we should validate the length of the predicates in `filter`, and that this should raise consistently

If Polars allows it, then OK, we can leave that be, but for pandas and pyarrow i think we should raise

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1786 

## Checklist

- [x] Tests added
- [x] Documented the changes

For lazy backends:

- duckdb raises: `duckdb.duckdb.BinderException`
- pyspark raises: `pyspark.errors.exceptions.captured.AnalysisException: [INVALID_WHERE_CONDITION]`
- dask raises `pd.errros.IndexingError`

at compute time. So polars is the only one allowing filtering with a scalar boolean

### DIFF
--- a/narwhals/_arrow/utils.py
+++ b/narwhals/_arrow/utils.py
@@ -6,6 +6,7 @@ from typing import Any
 from typing import Sequence
 from typing import overload
 
+from narwhals.exceptions import ShapeError
 from narwhals.utils import import_dtypes_module
 from narwhals.utils import isinstance_or_issubclass
 
@@ -212,7 +213,12 @@ def broadcast_and_extract_native(
 
 
 def validate_dataframe_comparand(
-    length: int, other: Any, backend_version: tuple[int, ...]
+    length: int,
+    other: Any,
+    backend_version: tuple[int, ...],
+    *,
+    allow_broadcast: bool,
+    method_name: str,
 ) -> Any:
     """Validate RHS of binary operation.
 
@@ -222,7 +228,14 @@ def validate_dataframe_comparand(
     from narwhals._arrow.series import ArrowSeries
 
     if isinstance(other, ArrowSeries):
-        if len(other) == 1:
+        len_other = len(other)
+        if len_other == 1:
+            if length > 1 and not allow_broadcast:
+                msg = (
+                    f"{method_name}'s length: 1 differs from that of the series: {length}"
+                )
+                raise ShapeError(msg)
+
             import numpy as np  # ignore-banned-import
             import pyarrow as pa
 
@@ -230,6 +243,11 @@ def validate_dataframe_comparand(
             if backend_version < (13,) and hasattr(value, "as_py"):
                 value = value.as_py()
             return pa.array(np.full(shape=length, fill_value=value))
+
+        if length != len_other:
+            msg = f"{method_name}'s length: {len_other} differs from that of the series: {length}"
+            raise ShapeError(msg)
+
         return other._native_series
 
     from narwhals._arrow.dataframe import ArrowDataFrame  # pragma: no cover
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -20,6 +20,7 @@ from narwhals._pandas_like.utils import rename
 from narwhals._pandas_like.utils import select_columns_by_name
 from narwhals._pandas_like.utils import validate_dataframe_comparand
 from narwhals.dependencies import is_numpy_array
+from narwhals.exceptions import ShapeError
 from narwhals.utils import Implementation
 from narwhals.utils import check_column_exists
 from narwhals.utils import flatten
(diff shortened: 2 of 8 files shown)


narwhals/_pandas_like/dataframe.py
```diff
@@ -418,7 +419,15 @@ def filter(self, *predicates: IntoPandasLikeExpr, **constraints: Any) -> Self:
...
             # `[0]` is safe as all_horizontal's expression only returns a single column
             mask = expr._call(self)[0]
+            if len(mask) != len(self):
+                msg = (
+                    f"Predicate result has length {len(mask)}, which is incompatible with "
+                    f"DataFrame of length {len(self)}"
+                )
+                raise ShapeError(msg)
+
             _mask = validate_dataframe_comparand(self._native_frame.index, mask)
```
MarcoGorelli, line 429: i think this error should be handled within `validate_dataframe_comparand`, perhaps we could add a `preserve_lhs_length` boolean argument to it?
could we then do the equivalent thing for _arrow?

MarcoGorelli (commented):
thanks! just got a comment on the structure

narwhals/_pandas_like/dataframe.py
FBruzzesi: Just to make you sure I get this suggestion correctly, you would prefer to add an argument that prevents broadcasting in  `validate_dataframe_comparand`. E.g.

```
def validate_dataframe_comparand(..., allow_broadcast: bool):
    ...
    if len(other) == 1 and not allow_broadcast:
         raise ShapeError(...)
```

MarcoGorelli: yup! but only if it's a "length n vs length 1" case, so e.g.
```python
if len(other) ==1 and len(df) > 1 and not allow_broadcast
```
If they're both length-1, then there should be an error

separately to this, we should also do something similar for the series broadcasting cases, so we could raise on user errors in `zip_with` / `when/then/otherwise`

FBruzzesi: Thinking a little more about this while changing `validate_dataframe_comparand`: there would be cases that the shapes are just different, but not necessarily a scalar. I don't know why someone might want to do this but:

```py
df_nw.filter(nw.col("x").unique() > 1)
```

`nw.col("x").unique() > 1` would result to something different than a single boolean, but still with a shape mismatch

MarcoGorelli: yup, agree! we could raise for that too in `validate_dataframe_comparand`, right?
- if `allow_broadcast`, `len(other)` can be `len(df)` or `1`
- else, `len(other)` must be `len(df)`

narwhals/_arrow/utils.py
```diff
@@ -212,7 +213,12 @@ def broadcast_and_extract_native(
 
 
 def validate_dataframe_comparand(
-    length: int, other: Any, backend_version: tuple[int, ...]
+    length: int,
+    other: Any,
+    backend_version: tuple[int, ...],
+    *,
+    allow_broadcast: bool,
+    method_name: str,
```
FBruzzesi, line 221: Needed to align error message with the polar's one

FBruzzesi:
Unrelated scikit-lego 🙈

MarcoGorelli (approved):
love this, thanks @FBruzzesi !

tests/frame/filter_test.py
```diff
@@ -33,3 +35,40 @@ def test_filter_with_boolean_list(
...
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8, 9]}
+    df = nw.from_native(constructor(data))
+
+    context = (
+        pytest.raises(
+            ShapeError,
+            match="filter's length: 1 differs from that of the series: 3",
+        )
+        if any(x in str(constructor) for x in ("pandas", "pyarrow", "modin"))
+        else does_not_raise()
```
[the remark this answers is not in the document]
FBruzzesi: Sure, I will open up an issue

FBruzzesi:
@MarcoGorelli actually this is the change that is causing scikit-lego to break.
I am taking a deeper look now

FBruzzesi:
If `ShapeError` was to inherit from `ValueError` then this would fix the issue. To me however, it would make more sense for this to be an `AssertionError`. WDYT?

PR #1816, opened 2025-01-16
feat: add `AnonymousExprError`

EdAbati:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

I propose to add a error for the (very) common error message about anonlymous expressions.

Ideally I would like to move this a level up (since we raise it for every backend) but I get all sorts of mypy error if I do that. e.g. `mypy` thinks `root_names` could still be `None`

### DIFF
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -15,6 +15,7 @@ from narwhals._arrow.series import ArrowSeries
 from narwhals._expression_parsing import reuse_series_implementation
 from narwhals.dependencies import get_numpy
 from narwhals.dependencies import is_numpy_array
+from narwhals.exceptions import AnonymousExprError
 from narwhals.exceptions import ColumnNotFoundError
 from narwhals.typing import CompliantExpr
 from narwhals.utils import Implementation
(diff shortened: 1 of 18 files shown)


narwhals/_arrow/expr.py
```diff
@@ -390,12 +391,9 @@ def clip(self: Self, lower_bound: Any | None, upper_bound: Any | None) -> Self:
...
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
             if self._output_names is None:
-                msg = (
-                    "Anonymous expressions are not supported in over.\n"
-                    "Instead of `nw.all()`, try using a named expression, such as "
-                    "`nw.col('a', 'b')`\n"
-                )
-                raise ValueError(msg)
+                msg = ".over"
+                raise AnonymousExprError.from_expr_name(msg)
```
EdAbati, line 395: ruff would still complain if we don't move the string to a separate variable

MarcoGorelli:
awesome, thanks! i think it's fine to now move this up - in theory, I don't see why we should be raising this error at all, i think for now it's fine to just raise the error at the compliant level

EdAbati:
> I don't see why we should be raising this error at all

Do you mean that we should support this or that we should let the backends raise their own errors?

> i think for now it's fine to just raise the error at the compliant level

I tried to do something like this:

```python
class ExprNameNamespace:
    ...
    def keep(self: Self) -> ExprT:
         def _func(plx):
            compliant_expr = self._expr._to_compliant_expr(plx)
            if compliant_expr._root_names is None:
                raise AnonymousExprError.from_expr_name(".name.keep")
            return compliant_expr.name.keep()
        
        return self._expr.__class__(_func)
```

but `mypy` was not very happy :D

MarcoGorelli:
> Do you mean that we should start supporting this or that we should let the backends raise their own errors?

I think `df.group_by('a').agg(nw.all().min())` is well-defined, and there's no reason any current backend can't do it, and that we can aim to support it everywhere in the future. In theory, I hope that in the future it'll be possible for us to support it - so, it's OK for now if for some backends we raise `AnonymousExprError`

PR #1822, opened 2025-01-17
feat: add `datetime` selector

### LINKED ISSUE #1813: [Enh]: ncs.datetime selector

use case for this came up in this PR of tubular https://github.com/lvgig/tubular/pull/366.

We want to select all datetime columns and currently have to define all of these:

TIME_UNITS = ["us", "ns", "ms"]
TIME_ZONES = zoneinfo.available_timezones().union({None})

DATETIME_VARIANTS = [
    nw.Datetime(time_unit=time_unit, time_zone=time_zone)
    for time_unit in TIME_UNITS
    for time_zone in TIME_ZONES
]

datetime_columns = list(
            X.select(ncs.by_dtype(*DATETIME_VARIANTS)).columns,)

I expect similar logic could be turned into an ncs.datetime selector function similar to current categorical/numeric.

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1813  for backends that have selectors

## Checklist

- [x] Tests added
- [x] Documented the changes

@MarcoGorelli you might have a better idea for how to test this. I checked how polars does it and it is definitly brute forcing 😂

I took the opportunity to:

- Modernize the selectors examples in the docstrings
- Create a `TimeUnit` alias to use around the codebase

The actual changes are not _too_ large


narwhals/selectors.py
```diff
@@ -234,42 +302,193 @@ def all() -> Expr:
...
+        c: bool
+        ----
+        a: [[1,2]]
+        b: [["x","y"]]
+        c: [[false,true]]
     """
     return Selector(lambda plx: plx.selectors.all())
 
 
+def datetime(
```
FBruzzesi, line 355: Finally, the datetime selector 😂

narwhals/utils.py
```diff
@@ -1067,3 +1072,37 @@ def check_column_exists(columns: list[str], subset: list[str] | None) -> None:
...
+    else:
+        time_zones = (
+            [time_zone] if isinstance(time_zone, (str, timezone)) else list(time_zone)
+        )
+
+    if "*" in time_zones:
+        import zoneinfo
+
+        time_zones.extend(list(zoneinfo.available_timezones()))
+        time_zones.remove("*")
```
FBruzzesi, line 1101: Since we don't allow "*" to be passed to the Datetime constructor as a time zone, I am explicitly creating all the timezones.

narwhals/_dask/selectors.py
```diff
@@ -89,6 +95,18 @@ def func(df: DaskLazyFrame) -> list[Any]:
             kwargs={},
         )
 
+    def datetime(
+        self: Self,
+        time_unit: TimeUnit | Collection[TimeUnit] | None,
+        time_zone: str | timezone | Collection[str | timezone | None] | None,
+    ) -> DaskSelector:  # pragma: no cover
```
FBruzzesi, line 102: For dask, the selector work, but the cast fails with:

> TypeError: Cannot use .astype to convert from timezone-aware dtype to timezone-naive dtype. Use obj.tz_localize(None) or obj.tz_convert('UTC').tz_localize(None) instead.

tests/selectors_test.py
```diff
@@ -72,22 +72,121 @@ def test_categorical(
...
 
 
+def test_datetime(constructor: Constructor, request: pytest.FixtureRequest) -> None:
+    if (
+        "pyspark" in str(constructor)
+        or "duckdb" in str(constructor)
+        or "dask" in str(constructor)
+        or ("pyarrow_table" in str(constructor) and PYARROW_VERSION < (12,))
+        or ("pyarrow" in str(constructor) and is_windows())
+        or ("pandas" in str(constructor) and PANDAS_VERSION < (2,))
```
FBruzzesi, line 86: These are a lot of xfail, but let me go through them:

- pyspark and duckdb: do not implement selectors
- dask: see [comment](https://github.com/narwhals-dev/narwhals/pull/1822#discussion_r1920872606)
- pyarrow < 12, does not implement `replace_time_zone`
- pyarrow in windows: fails to find UTC timezone
- pandas < 2, does not support `time_units!="ns"`

tests/expr_and_series/dt/datetime_attributes_test.py
```diff
@@ -118,7 +118,6 @@ def test_to_date(request: pytest.FixtureRequest, constructor: Constructor) -> No
             "pandas_nullable_constructor",
             "cudf",
             "modin_constructor",
-            "pyspark",
```
FBruzzesi, line 121: Unrelated to this PR

pyproject.toml
```diff
@@ -174,7 +176,8 @@ xfail_strict = true
 markers = ["slow: marks tests as slow (deselect with '-m \"not slow\"')"]
 env = [
   "MODIN_ENGINE=python",
-  "PYARROW_IGNORE_TIMEZONE=1"
+  "PYARROW_IGNORE_TIMEZONE=1",
+  "TZ='UTC'",
```
FBruzzesi, line 180: Useful for local when running pyspark (which we set to run on UTC)

FBruzzesi:
```suggestion
```

narwhals/utils.py
```diff
@@ -1067,3 +1072,44 @@ def check_column_exists(columns: list[str], subset: list[str] | None) -> None:
...
+
+        if sys.version_info >= (3, 9):
+            import zoneinfo
+        else:  # pragma: no cover
+            # This code block is due to a typing issue with backports.zoneinfo package:
+            # https://github.com/pganssle/zoneinfo/issues/125
+            from backports import zoneinfo
+
+        time_zones.extend(list(zoneinfo.available_timezones()))
+        time_zones.remove("*")
```
MarcoGorelli, line 1108: is this taken from Polars? because chrono-tz's time zone identifiers are a bit different in some cases, i'm not sure we should be doing this

FBruzzesi: From [polars docs](https://docs.pola.rs/api/python/stable/reference/selectors.html#polars.selectors.datetime):

```py
"""
...
time_zone: 
    One or more timezone strings, as defined in zoneinfo (to see valid options run `import zoneinfo;
        zoneinfo.available_timezones()` for a full list).
"""
```

MarcoGorelli: sure but it's not in the source code, right?

FBruzzesi: No it is not, polars handles "*" out of the box to match all the timezones and None to handle non aware datetimes (default arg is `time_zone = ("*", None)`)

MarcoGorelli: I still don't think we should be using this, we can now match on `nw.Datetime` to select all datetime columns (tz-aware or not), does that work?

FBruzzesi: It would probably work for the use case in the issue, but this is a bit more general, isn't it?
E.g.:

- If I want to select only datetimes without a timezone I can do `ncs.datetime(time_zone = None)`
- If I want to select only timezone aware datetimes I can do `ncs.datetime(time_zone = "*")`
  - This case would still require to parse "*" or to exclude some dtypes anyway.

I will check if any workaround without explicitly building the set of Datetime's works

MarcoGorelli: thanks! yes it might require some exclude logic, i'd just worry a bit about matching on 598 different Datetime instances (and then still risk missing some)

FBruzzesi: I think the only fast path we could have are the following cases:

```py
if (time_units == {"ms", "us", "ns"}) and (time_zones == {"*", None}):
    return {dtypes.Datetime}

if "*" not in time_zones:
    return {
        dtypes.Datetime(time_unit=tu, time_zone=tz)
        for tu in time_units
        for tz in time_zones
    }
```

If "*" is in time_zones, then we need to manually create combinations I believe. 

The difference I see in polars is this line: https://github.com/pola-rs/polars/blob/main/py-polars/polars/datatypes/classes.py#L463:

> When used to match dtypes, can set this to "*" to check for Datetime columns that have any (non-null) timezone.

However I don't see where this is used - at least in Datetime class the equality doesn't have any special case for "*"

MarcoGorelli: it's probably in the rust code

would something like `by_dtype(nw.Datetime)` - `by_dtype(nw.Datetime(time_zone=None))` work to select all tz-aware columns?

FBruzzesi: > it's probably in the rust code

Yeah I am still not so confident with that, but I could imagine a similar logic as `pl.col("*")` to select all columns

> would something like `by_dtype(nw.Datetime)` - `by_dtype(nw.Datetime(time_zone=None))` work to select all tz-aware columns?

Not really, you would need:

```py
by_dtype([nw.Datetime]) - by_dtype([nw.Datetime("us", time_zone=None), nw.Datetime("ms", time_zone=None), nw.Datetime("ns", time_zone=None)])
```

But now, what if you would like to select all timezones, with time_unit = "ns"? I don't see a clear way of doing it.

Just to make sure I understand correctly, the concern is about potentially missing some timezones?

MarcoGorelli: partially the missing time zones, partially that we're expanding nearly 600 different data types - i can take a look, got a feeling we can probably find a way

FBruzzesi: Regarding the exploding numbers we can have a few "fast path" that return a subset, but I wouldn't worry too much about performances - I am returning a set on purpose here. 

Regarding missing time zones I hear you! Should we consider adding support for wildcard more generally? 👀

MarcoGorelli: i might be missing something but if we look at

https://github.com/narwhals-dev/narwhals/blob/562b213c8e571b690db57b0145f3a0f023d7382e/narwhals/_duckdb/selectors.py#L31-L48

then shouldn't be possible to check that `df.schema[col]` is `Datetime` and that `df.schema[col].time_zone` is not `None`?

narwhals/utils.py
```diff
@@ -1070,3 +1073,40 @@ def check_column_exists(columns: list[str], subset: list[str] | None) -> None:
...
+    time_units: set[str],
+    time_zones: set[str | None],
+) -> bool:
+    return (
+        (dtype == dtypes.Datetime)
+        and (dtype.time_unit in time_units)  # type: ignore[attr-defined]
+        and (
+            dtype.time_zone in time_zones  # type: ignore[attr-defined]
+            or ("*" in time_zones and dtype.time_zone is not None)  # type: ignore[attr-defined]
+        )
```
FBruzzesi, line 1111: @MarcoGorelli this might make the trick you were looking for :)

tests/selectors_test.py
MarcoGorelli: > pyspark and duckdb: do not implement selectors

i think we do have these now, right?

narwhals/utils.py
MarcoGorelli: yup, nice!

narwhals/utils.py
```diff
@@ -1070,3 +1073,40 @@ def check_column_exists(columns: list[str], subset: list[str] | None) -> None:
...
+        if time_unit is None
+        else {time_unit}
+        if isinstance(time_unit, str)
+        else set(time_unit)
+    )
+    time_zones: set[str | None] = (
+        {None}
+        if time_zone is None
+        else {str(time_zone)}
+        if isinstance(time_zone, (str, timezone))
```
MarcoGorelli, line 1093: do we have any test for when it's an instance of `timezone`? can we check that `timezone.utc` and `ZoneInfo.UTC` both work?

MarcoGorelli (commented):
nice, thanks @FBruzzesi ! sorry didn't quite get round to reviewing everything in time for today, we can always include in next week's release (in fact, that's one reason to have regular and short release cycles, so we don't feel like we need to rush things in - if something doesn't make it, the next release is just next week)

tests/selectors_test.py
FBruzzesi: Yes, but also no!
I can add the selector but we have very minimal support for datetime dtype in spark and duckdb (read as: no support for timezones)

narwhals/utils.py
FBruzzesi: Well.. `ZoneInfo` does not work for polars 😱

narwhals/utils.py
```diff
@@ -1170,3 +1173,40 @@ def check_column_names_are_unique(columns: list[str]) -> None:
...
         msg = f"Expected unique column names, got:{msg}"
         raise DuplicateError(msg)
+
+
+def _parse_time_unit_and_time_zone(
+    time_unit: TimeUnit | Iterable[TimeUnit] | None,
+    time_zone: str | timezone | Iterable[str | timezone | None] | None,
+) -> tuple[set[str], set[str | None]]:
+    time_units = (
+        {"ms", "us", "ns"}
```
MarcoGorelli, line 1183: wondering if we should include `'s'` as well

MarcoGorelli (commented):
nice one, implementation looks good!

couple of comments:
- as this is new functionality, shall we make the docstring concise to begin with?
- got a comment about `time_unit`

narwhals/selectors.py
```diff
@@ -474,11 +477,115 @@ def all() -> Selector:
...
+
+            * One or more timezone strings, as defined in zoneinfo (to see valid options
+                run `import zoneinfo; zoneinfo.available_timezones()` for a full list).
+            * Set `None` to select Datetime columns that do not have a timezone.
+            * Set `"*"` to select Datetime columns that have *any* timezone.
+
+    Returns:
+        A new expression.
+
+    Examples:
```
MarcoGorelli, line 499: can we make a short example please?

tests/selectors_test.py
```diff
@@ -85,22 +82,140 @@ def test_categorical(
...
+    assert df.with_columns(
+        nw.col("ts_ms").dt.replace_time_zone("UTC").alias("ts_utc")
+    ).select(
+        ncs.datetime(time_unit=["ms", "us"], time_zone=[None, timezone.utc])
+    ).collect_schema().names() == ["ts_ms", "ts_us", "ts_utc"]
+
+
+def test_datetime_spark_and_duckdb(constructor: Constructor) -> None:
+    if not ("pyspark" in str(constructor) or "duckdb" in str(constructor)):
+        pytest.skip()
```
MarcoGorelli, line 193: these should be supported now?

MarcoGorelli (changes requested):
thanks!

tests/selectors_test.py
FBruzzesi: This is an "easy"-er example test for duckdb and pyspark since we don't support converting with timezones for those backends

narwhals/selectors.py
FBruzzesi: Sure I will do that

tests/selectors_test.py
FBruzzesi: Meaning, all other backends will pass this test anyway, while duckdb and pyspark won't pass the other one. No harm in having all backends here

narwhals/selectors.py
```diff
@@ -474,11 +477,78 @@ def all() -> Selector:
...
+        ...     "tstamp": [
+        ...         datetime(2000, 11, 20, 18, 12, 16, 600000),
+        ...         datetime(2020, 10, 30, 10, 20, 25, 123000),
+        ...     ],
+        ...     "numeric": [3.14, 6.28],
+        ... }
+        >>> df_native = pa.table(data)
+        >>> df_nw = nw.from_native(df_native)
+
+        Let's define a dataframe-agnostic function to select datetime dtypes:
```
MarcoGorelli, line 520: outdated?

MarcoGorelli (approved):
thanks @FBruzzesi

MarcoGorelli:
sorry, just some more merge conflicts

PR #1828, opened 2025-01-19
feat: track whether expressions change length but don't aggregate, and only allow length-changing expressions if they're followed by aggregations in the lazy API

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -342,3 +342,44 @@ def operation_is_order_dependent(*args: IntoExpr | Any) -> bool:
     # it means that it was a scalar (e.g. nw.col('a') + 1) or a column name,
     # neither of which is order-dependent, so we default to `False`.
     return any(getattr(x, "_is_order_dependent", False) for x in args)
+
+
+def operation_changes_length(*args: IntoExpr | Any) -> bool:
+    """Track whether operation changes length.
+
+    n-ary operations between expressions which change length are not
+    allowed. This is because the output might be non-relational. For
+    example:
+        df = pl.LazyFrame({'a': [1,2,None], 'b': [4,None,6]})
+        df.select(pl.col('a', 'b').drop_nulls())
+    Polars does allow this, but in the result we end up with the
+    tuple (2, 6) which wasn't part of the original data.
+
+    Rules are:
+        - in an n-ary operation, if any one of them changes length, then
+          it must be the only expression present
+        - in a comparison between a changes-length expression and a
+          scalar, the output changes length
+    """
+    from narwhals.expr import Expr
+
+    n_length_changing_expressions = len(
+        [x for x in args if isinstance(x, Expr) and x._changes_length]
+    )
+    if n_length_changing_expressions > 1:
+        msg = (
+            "Found multiple expressions which change length. You can only use one "
+            "length-changing expression at a time, unless it is followed by an aggregation."
+        )
+        # TODO(marco): custom error class
+        raise ValueError(msg)
+    return n_length_changing_expressions > 0
+
+
+def operation_aggregates(*args: IntoExpr | Any) -> bool:
+    # If an arg is an Expr, we look at `_aggregates`. If it isn't,
+    # it means that it was a scalar (e.g. nw.col('a').sum() + 1),
+    # which is already length-1, so we default to `True`. If any
+    # expression does not aggregate, then broadcasting will take
+    # place and the result will not be an aggregate.
+    return all(getattr(x, "_aggregates", True) for x in args)
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -363,17 +363,18 @@ def operation_changes_length(*args: IntoExpr | Any) -> bool:
     """
     from narwhals.expr import Expr
 
-    n_length_changing_expressions = len(
-        [x for x in args if isinstance(x, Expr) and x._changes_length]
+    n_exprs = len([x for x in args if isinstance(x, Expr)])
+    changes_length = any(
+        isinstance(x, Expr) and x._changes_length and not x._aggregates for x in args
     )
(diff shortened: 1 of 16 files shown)


tests/expr_and_series/unique_test.py
```diff
@@ -1,20 +1,49 @@
...
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.col("a").unique().sum())
+    expected = {"a": [3]}
     assert_equal_data(result, expected)
 
 
+def test_unique_illegal_combination(constructor: Constructor) -> None:
+    df = nw.from_native(constructor(data))
+    with pytest.raises(LengthChangingExprError):
+        df.select((nw.col("a").unique() + nw.col("b").unique()).sum())
```
MarcoGorelli, line 44: an expression like this is so bizarre that I doubt we'd be cutting off many real use-cases...

narwhals/_expression_parsing.py
```diff
@@ -342,3 +343,44 @@ def operation_is_order_dependent(*args: IntoExpr | Any) -> bool:
...
+
+    n_exprs = len([x for x in args if isinstance(x, Expr)])
+    changes_length = any(
+        isinstance(x, Expr) and x._changes_length and not x._aggregates for x in args
+    )
+    if n_exprs > 1 and changes_length:
+        msg = (
+            "Found multiple expressions at least one of which changes length. You can only use one "
+            "length-changing expression at a time, unless it is followed by an aggregation."
+        )
```
FBruzzesi, line 375: I find this message a bit confusing. I imagine you mean that a length changing expression might be used only if it's the only expression.

Maybe:

```suggestion
        msg = (
            "Found multiple expressions at least one of which changes length. Any length-changing expression "
            "can be used in isolation, unless it is followed by an aggregation."
        )
```

tests/expr_and_series/unique_test.py
FBruzzesi: Agree! 😂

tests/expr_and_series/unique_test.py
```diff
@@ -1,20 +1,49 @@
...
+
+def test_unique_expr_agg(
+    constructor: Constructor, request: pytest.FixtureRequest
+) -> None:
+    if any(x in str(constructor) for x in ("duckdb", "pyspark")):
+        request.applymarker(pytest.mark.xfail)
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.col("a").unique().sum())
+    expected = {"a": [3]}
     assert_equal_data(result, expected)
```
FBruzzesi, line 40: Can we maybe add a test with a length changing and an aggregation together? This should not raise I think.

e.g.

```suggestion


def test_unique_expr_and_agg(
    constructor: Constructor, request: pytest.FixtureRequest
) -> None:
    if any(x in str(constructor) for x in ("duckdb", "pyspark")):
        request.applymarker(pytest.mark.xfail)
    df = nw.from_native(constructor(data))
    result = df.select(nw.col("a").unique().alias("unique"), nw.col("a").sum().alias("sum"))
    expected = {"unique": [1, 2], "sum": [4]}
    
    assert_equal_data(result, expected)


```

WDYT?

FBruzzesi (commented):
Thanks @MarcoGorelli ! This was easier to go through than expected 🚀

I left only a couple of comments but I wonder about other edge cases 🤔

tests/expr_and_series/unique_test.py
MarcoGorelli: i'm confused, sorry - `expected` has keys of different lengths?

narwhals/expr_cat.py
```diff
@@ -64,4 +64,6 @@ def get_categories(self: Self) -> ExprT:
         return self._expr.__class__(
             lambda plx: self._expr._to_compliant_expr(plx).cat.get_categories(),
             self._expr._is_order_dependent,
+            changes_length=self._expr._changes_length,
```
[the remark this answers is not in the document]
MarcoGorelli: great catch, thanks!

tests/expr_and_series/unique_test.py
FBruzzesi: Whops, my bad! Just updated the values. It's a length changing and an aggregated that then get broadcasted in the same select context.

MarcoGorelli: thanks! I think this is the kind of operation that sql / pyspark / wouldn't support, and I'm not sure we need to, it could just as well be done with `df.select('a', nw.col('a').sum().alias('sum')).unique('a')`

trying this out for example raises:
```python
In [14]: rel = duckdb.sql('select * from values (1, 4), (1, 2), (2, 3), (2, 4) df(a, b)')

In [15]: duckdb.sql("""
    ...: select unnest(list(distinct a)), sum(b) over () from rel
    ...: """)
BinderException                           Traceback (most recent call last)
Cell In[15], line 1
----> 1 duckdb.sql("""
      2 select unnest(list(distinct a)), sum(b) over () from rel
      3 """)

BinderException: Binder Error: column "b" must appear in the GROUP BY clause or must be part of an aggregate function.
Either add it to the GROUP BY list, or use "ANY_VALUE(b)" if the exact value of "b" is not important.
```

It would still be backwards-compatible if we later allowed it - but it's easier to start with a simpler and predictable rule, and if necessary relax later

FBruzzesi: Cool! Resolving then!

MarcoGorelli:
thanks for your review!

PR #1830, opened 2025-01-19
chore: rename time zone tests

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes


tests/expr_and_series/dt/datetime_attributes_test.py
```diff
@@ -49,10 +50,15 @@ def test_datetime_attributes(
...
         request.applymarker(pytest.mark.xfail)
-    if "pyspark" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
 
-    df = nw.from_native(constructor(data))
+    if "pyspark" in str(constructor):
+        data_ = {
+            col_name: [v.replace(tzinfo=timezone.utc) for v in col_values]
+            for col_name, col_values in data.items()
+        }
```
FBruzzesi, line 58: This is most likely needed for everyone not based in utc time. Since in `pyspark_lazy_constructor` we set `.config("spark.sql.session.timeZone", "UTC")`, the results will not be the same otherwise. 

Context: it took me a bit to realize why instead of hours 12 and 2 I was getting 11 and 1

tests/expr_and_series/dt/datetime_attributes_test.py
```diff
@@ -120,11 +126,10 @@ def test_to_date(request: pytest.FixtureRequest, constructor: Constructor) -> No
             "pandas_nullable_constructor",
             "cudf",
             "modin_constructor",
-            "pyspark",
         )
     ):
         request.applymarker(pytest.mark.xfail)
-    dates = {"a": [datetime(2001, 1, 1), None, datetime(2001, 1, 3)]}
+    dates = {"a": [datetime(2001, 1, 1), datetime(2001, 1, 3)]}
```
FBruzzesi, line 132: Removal of None is due to the fact that otherwise pyspark ends up creating a column of struct type. We might need to address this in the constructor?

tests/expr_and_series/dt/datetime_attributes_test.py
MarcoGorelli: shall we just put the tzinfo to `timezone.utc` when we define `data=`?

tests/expr_and_series/dt/datetime_attributes_test.py
MarcoGorelli: 🤔 yeah might be good to see if we can address the constructor so it doesn't go via pandas, that might in general be good going forwards

MarcoGorelli (commented):
awesome thanks!

tests/expr_and_series/dt/datetime_attributes_test.py
FBruzzesi: I tried, and all pandas like backends raise warnings/errors sadly

PR #1853, opened 2025-01-22
feat: pyspark and duckdb selectors

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes


tests/selectors_test.py
```diff
@@ -96,7 +88,7 @@ def test_set_ops(
     expected: list[str],
     request: pytest.FixtureRequest,
 ) -> None:
-    if ("pyspark" in str(constructor)) or "duckdb" in str(constructor):
+    if "duckdb" in str(constructor) and not expected:
```
FBruzzesi, line 91: To figure this out due to:

> # TODO(marco): return empty relation with 0 columns?
>            return self._from_native_frame(self._native_frame.limit(0))

took me some sanity 😂

I tried in different ways to create an empty table but duckdb doesn't really like that. Maybe we might just raise an error?

narwhals/_duckdb/selectors.py
```diff
@@ -0,0 +1,221 @@
...
+            returns_scalar=False,
+            version=self._version,
+            kwargs={},
+        )
+
+    def numeric(self: Self) -> DuckDBSelector:
+        dtypes = import_dtypes_module(self._version)
+        return self.by_dtype(
+            [
+                dtypes.Int64,
```
MarcoGorelli, line 52: do we need dtypes.Int128 in here too now?

FBruzzesi: Nice catch!

FBruzzesi: Should it be added everywhere?

MarcoGorelli: yes, probably

MarcoGorelli:
nice, thanks - i'm just in the middle of a big refactor, so i'll finish that, then i'll update these

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #1855, opened 2025-01-23
Track is multi output

MarcoGorelli:
One step closer to #1780.

Next steps:
- in the metadata, store a callable which determines the root names
- only set aliases when necessary, as opposed to renaming after each operation. PySpark / SQL don't let us set as many aliases as we do now, e.g. `'select (a - mean(a) over () as a) as a from rel'` raises

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -7,14 +7,16 @@ from copy import copy
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Sequence
+from typing import TypedDict
 from typing import TypeVar
 from typing import Union
-from typing import cast, TypedDict
+from typing import cast
 from typing import overload
 
 from narwhals.dependencies import is_numpy_array
 from narwhals.exceptions import InvalidIntoExprError
-from narwhals.exceptions import LengthChangingExprError, MultiOutputExprError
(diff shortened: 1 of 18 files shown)


narwhals/_expression_parsing.py
```diff
@@ -338,11 +347,46 @@ def extract_compliant(
...
+    return True
+
+
+def arg_is_order_dependent(arg: IntoExpr | Any) -> bool:
+    from narwhals.expr import Expr
+
+    if isinstance(arg, Expr):
+        return arg._metadata["is_order_dependent"]
+    if isinstance(arg, str):
+        # Column name, e.g. 'a', gets treated as `nw.col('a')`,
```
MarcoGorelli, line 372: sigh...this isn't even always true. darn. we may need to parse column names as expressions a bit earlier

MarcoGorelli:
i think this isn't necessary any more

Issue #1390, opened 2024-11-17
`df.select(0)` treats `0` as literal for Polars, and raises for other backends


MarcoGorelli:
This was brought up in https://github.com/narwhals-dev/narwhals/pull/1379

```python
df = pd.DataFrame({'a': [1,2], 'b': [4,5]})
print(nw.from_native(df).select(0).to_native())
df = pl.DataFrame({'a': [1,2], 'b': [4,5]})
print(nw.from_native(df).select(0).to_native())
```

For Polars, this outputs
```
shape: (1, 1)
┌─────────┐
│ literal │
│ ---     │
│ i32     │
╞═════════╡
│ 0       │
└─────────┘
```
whereas as for pandas and other backends it raises

I don't think this is terrible or too high-priority, as Narwhals users would typically test at least against pandas and Polars

But, it would be nice if we were able to intercept Polars here and make it consistent

EdAbati:
Linked to this, I think we could update the `InvalidIntoExprError` and add a hint mentioning this in case someone is expecting the same behaviour as polars.

Something like:
```
Hint 2: if you were trying to create a literal column, you should explicitly use `nw.lit`.
For example, `df.select(nw.lit(0))` to have a column with a `0` literal value.
```

MarcoGorelli:
yes! want to make a pr to update it? then we can release today

skritsotalakis:
I'm going to tackle this if that's ok

PR #1856, opened 2025-01-23
chore: parse strings as columns names at narwhals-level

MarcoGorelli:
closes https://github.com/narwhals-dev/narwhals/issues/1390 (completely by accident 😳 😄 )

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_interchange/dataframe.py
```diff
@@ -153,25 +153,25 @@ def __getattr__(self: Self, attr: str) -> NoReturn:
...
-
-        frame = self._interchange_frame.select_columns_by_name(exprs)
+    def simple_select(self: Self, *column_names: str) -> Self:
+        frame = self._interchange_frame.select_columns_by_name(list(column_names))
         if not hasattr(frame, "_df"):  # pragma: no cover
             msg = (
                 "Expected interchange object to implement `_df` property to allow for recovering original object.\n"
                 "See https://github.com/data-apis/dataframe-api/issues/360."
             )
-            raise NotImplementedError(frame)
```
FBruzzesi, line 176: 🙈

narwhals/group_by.py
```diff
@@ -30,9 +31,7 @@ def __init__(self: Self, df: DataFrameT, *keys: str, drop_null_keys: bool) -> No
             *self._keys, drop_null_keys=drop_null_keys
         )
 
-    def agg(
-        self: Self, *aggs: IntoExpr | Iterable[IntoExpr], **named_aggs: IntoExpr
-    ) -> DataFrameT:
+    def agg(self: Self, *aggs: Expr | Iterable[Expr], **named_aggs: Expr) -> DataFrameT:
```
FBruzzesi, line 34: I think this should be:

```suggestion
    def agg(self: Self, *aggs: Expr, **named_aggs: Expr) -> DataFrameT:
```
as we don't expect someone to pass `.agg( (nw.sum("a"), nw.sum("b")), (nw.mean("a"), ))`

narwhals/group_by.py
FBruzzesi, line 1: Interesting change here, I am gladly surprised that everything plays out!

FBruzzesi (commented):
Thanks @MarcoGorelli that's quite a refactor!

Not a real concern but maybe something to think about is:

- The asymmetry in some method specs, between Narwhals and Compliant objects
- The expected `simple_select` method (just needs documentation I guess)
- Are we adding any extra overhead?

narwhals/group_by.py
MarcoGorelli: yeah it's allowed to do `.agg([nw.sum('a'), nw.sum('b')])`

narwhals/dataframe.py
```diff
@@ -118,19 +131,38 @@ def columns(self: Self) -> list[str]:
...
+        flat_exprs = tuple(flatten(exprs))
+        if flat_exprs and all(isinstance(x, str) for x in flat_exprs) and not named_exprs:
+            # fast path!
+            try:
+                return self._from_compliant_dataframe(
+                    self._compliant_frame.simple_select(*flat_exprs),
+                )
+            except Exception as e:
+                # Column not found is the only thing that can realistically be raised here.
+                available_columns = self.columns
```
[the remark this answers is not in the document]
MarcoGorelli: true, but it's crashing at this point anyway, might as well let it warn

MarcoGorelli:
thanks for your review!

> The asymmetry in some method specs, between Narwhals and Compliant objects. The expected simple_select method (just needs documentation I guess)

It's true that there's some asymmetry, but I think that's OK - in my head, the separation is:
- compliant level: strict and minimal
- narwhals (user-facing) level: more ergonomic

> Are we adding any extra overhead?

From tests I've done locally it doesn't look like it. Theoretically i'd expect it to be slightly lower, if we're just doing a single `isinstance(expr, str)` check upfront as opposed to repeatedly when parsing expressions at the compliant level

FBruzzesi (approved):
Amazing stuff (as usual!)

PartyDancingGIF

MarcoGorelli:
😄 thanks for your review, much appreciated!

Issue #1561, opened 2024-12-10
[Enh]: Support for `narwhals.Expr.hist`


mscolnick:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

This is for [marimo](https://github.com/marimo-team/marimo)

We already use narhwals, but this would help with performance.

### Please describe the purpose of the new feature or describe the problem to solve.

We would like to move some of our vega charts for `y=count(*), x=bin(column)` that get binned on the frontend to be on the backend. Since they are just a summary per column, we could achieve this with `narwhals.Expr.hist`

### Suggest a solution if possible.

`narwhals.Expr.hist` similar to `polars.Expr.hist`

for us in particular, we are looking to support:

* polars
* pandas
* pyarrow
* ibis (but can convert to pyarrow)

### If you have tried alternatives, please describe them below.

Alternative would be using `vegafusion` but that is an additional dependency that we could potentially avoid.

MarcoGorelli:
thanks for the request - I'd say that this is in-scope. not yet sure how we'd do it for pyarrow but we can explore it

camriddell:
I can work on this

PR #1859, opened 2025-01-23
feat: Series.hist

camriddell:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1561

## Checklist

- [x] Tests added
- [x] Documented the changes

Narwhals Expressions do not yet allow one to return a DataFrame (or a struct), so `.hist` is only implemented at the Series level to be compliant with the rest of the API.

The PyArrow implementation can likely be streamlined a bit more, so a review on that section would be appreciated.


MarcoGorelli:
wow, amazing!

> .hist is only implemented at the Series level to be compliant with the rest of the API.

agree, good design decision here, I think it would be quite awkward to add Expr.hist because:
- struct dtype is not supported by default in pandas (and not at all in pandas pre 2.0. Maybe 1.5. But not before that)
- it's a length-changing expression which it doesn't make sense to aggregate (e.g. `nw.col('a').unique().len()` makes sense but `nw.col('a').hist().struct.field('value').sum()` doesn't seem useful..), so supporting this for pyspark / duckdb / ibis could be a real issue. Ibis does seem to have [bucket](https://ibis-project.org/reference/operations#ibis.expr.operations.histograms.Bucket) but there's no examples and I have no idea what it does

Hi @mscolnick - just wanted to check that Series.hist would still be useful to you?

narwhals/_pandas_like/series.py
```diff
@@ -1031,6 +1031,54 @@ def rank(
...
+    def hist(
+        self: Self,
+        bins: list[float | int] | None = None,
+        *,
+        bin_count: int | None = None,
+        include_category: bool = True,
+        include_breakpoint: bool = True,
+    ) -> PandasLikeDataFrame:
+        from pandas import Categorical
+        from pandas import cut
```
FBruzzesi, line 1043: We should use `self.__native_namespace__()` here as well in place of pandas. I can see that `cudf` has a [`cut` function](https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.cut/), as well as [modin](https://docs.snowflake.com/en/developer-guide/snowpark/reference/python/1.24.0/modin/pandas_api/modin.pandas.cut)

narwhals/_arrow/series.py
```diff
@@ -1004,6 +1005,118 @@ def rank(
...
+        self: Self,
+        bins: list[float | int] | None = None,
+        *,
+        bin_count: int | None = None,
+        include_category: bool = True,
+        include_breakpoint: bool = True,
+    ) -> ArrowDataFrame:
+        import numpy as np  # ignore-banned-import
+        import pyarrow as pa
+        import pyarrow.compute as pc
```
FBruzzesi, line 1018: These are now imported on the top of the file anyway

```suggestion
```

FBruzzesi (commented):
Thanks a ton @camriddell ! I left a couple of comments and will go through the arrow implementation in more details later today 🚀

mscolnick:
awesome work @camriddell! yes, @MarcoGorelli this is definitely valuable for us (marimo) and would use it right away, thank you all for the efforts.

narwhals/series.py
```diff
@@ -4900,6 +4901,46 @@ def rank(
...
+            bin_count: If no bins provided, this will be used to determine the distance of the bins.
+            include_category: Include a column that indicates the upper value of each bin.
+            include_breakpoint: Include a column that shows the intervals as categories.
+
+        Returns:
+            A new DataFrame containing the counts of values that occur within each passed bin.
+        """
+        if bins is None and bin_count is None:
+            msg = "must provide one of `bin_count` or `bins`"
+            raise InvalidOperationError(msg)
```
FBruzzesi, line 4929: I couldn't find in the docs, but if they are both None then 10 bins are used? At least that's what happens if I simply run it with no input

narwhals/series.py
```diff
@@ -4900,6 +4901,46 @@ def rank(
...
+        if bins is not None and bin_count is not None:
+            msg = "can only provide one of `bin_count` or `bins`"
+            raise InvalidOperationError(msg)
```
FBruzzesi, line 4932: Polars raises a `ComputeError` for this case, we still don't have it implemented, but it would be a good opportunity to add it and raise such exception

narwhals/_pandas_like/series.py
```diff
@@ -1031,6 +1031,53 @@ def rank(
 
         return self._from_native_series(ranked_series)
 
+    def hist(
+        self: Self,
+        bins: list[float | int] | None = None,
+        *,
+        bin_count: int | None = None,
+        include_category: bool = True,
+        include_breakpoint: bool = True,
```
FBruzzesi, line 1040: Remove default values in compliant implementations
```suggestion
        bins: list[float | int] | None,
        *,
        bin_count: int | None,
        include_category: bool,
        include_breakpoint: bool,
```

narwhals/_pandas_like/series.py
```diff
@@ -1031,6 +1031,53 @@ def rank(
...
+    ) -> PandasLikeDataFrame:
+        from narwhals._pandas_like.dataframe import PandasLikeDataFrame
+
+        ns = self.__native_namespace__()
+        data: dict[str, Sequence[int | float | str]]
+
+        if bin_count is not None and bin_count == 0:
```
FBruzzesi, line 1047: I think it would be enough to check for equality with zero? (`bins` is implied to be `None`, and `bin_count` to be not `None`)

```suggestion
        if bin_count == 0:
```

narwhals/_pandas_like/series.py
```diff
@@ -1031,6 +1031,53 @@ def rank(
...
+                implementation=self._implementation,
+                backend_version=self._backend_version,
+                version=self._version,
+            )
+
+        result = (
+            ns.cut(self._native_series, bins=bins if bin_count is None else bin_count)
+            .value_counts()
+            .sort_index()
+        )
```
FBruzzesi, line 1066: TIL: we might just be able to use [`Series.value_counts`](https://pandas.pydata.org/docs/reference/api/pandas.Series.value_counts.html) with `bins` and `sort` specification:

```suggestion
        result = (
            self._native_series
            .value_counts(
                bins=bins if bin_count is None else bin_count,
                sort=False
            )
        )
```
and the remaining of the logic to get breakpoint and category if needed.

**Edit** I didn't check if the option is available for cudf and modin. If it isn't, we might have a different path for those with what you currently implemented

narwhals/_arrow/series.py
```diff
@@ -1004,6 +1005,116 @@ def rank(
         result = pc.if_else(null_mask, pa.scalar(None), rank)
         return self._from_native_series(result)
 
+    def hist(
+        self: Self,
+        bins: list[float | int] | None = None,
+        *,
+        bin_count: int | None = None,
+        include_category: bool = True,
+        include_breakpoint: bool = True,
```
FBruzzesi, line 1014:
```suggestion
        bins: list[float | int] | None,
        *,
        bin_count: int | None,
        include_category: bool,
        include_breakpoint: bool,
```

narwhals/_polars/series.py
```diff
@@ -416,6 +416,60 @@ def __contains__(self: Self, other: Any) -> bool:
             msg = f"Unable to compare other of type {type(other)} with series of type {self.dtype}."
             raise InvalidOperationError(msg) from exc
 
+    def hist(
+        self: Self,
+        bins: list[float | int] | None = None,
+        *,
+        bin_count: int | None = None,
+        include_category: bool = True,
+        include_breakpoint: bool = True,
```
FBruzzesi, line 425: And one more:

```suggestion
        bins: list[float | int] | None,
        *,
        bin_count: int | None,
        include_category: bool,
        include_breakpoint: bool,
```

narwhals/_polars/series.py
```diff
@@ -416,6 +416,60 @@ def __contains__(self: Self, other: Any) -> bool:
...
+    ) -> PolarsDataFrame:
+        from narwhals._polars.dataframe import PolarsDataFrame
+        from narwhals.exceptions import InvalidOperationError
+
+        # check for monotonicity, polars<1.0 does not do this.
+        if bins is not None:
+            for i in range(1, len(bins)):
+                if bins[i - 1] >= bins[i]:
+                    msg = "bins must increase monotonically"
+                    raise InvalidOperationError(msg)
```
FBruzzesi, line 435: Should we check this at Narwhals level? Also polars raises a `ComputeError` instead

narwhals/_polars/series.py
```diff
@@ -416,6 +416,60 @@ def __contains__(self: Self, other: Any) -> bool:
...
+
+        # polars<1.0 returned bins -inf to inf in these conditions
+        if (self._backend_version < (1, 0)) and ((len(bins) == 0) or (bin_count == 0)):  # type:ignore[arg-type]
```
FBruzzesi, line 438: 👌 nice catch

FBruzzesi (commented):
Hey @camriddell , I still didn't put my head around the arrow implementation. Yet I left some feedbacks on the other backends!

I hope to make it by end of this week to arrow

narwhals/_pandas_like/series.py
camriddell: Yep, I had this break in Modin when passing `.value_counts(..., sort=False)`. Then I wasn't certain how backwards compatible the `bins=` parameter so I opted to go with using an explicit `cut`. However taking a quick look it seems that `pandas` has allowed `.value_counts(bins=...)` for a while now so this would probably work if you think it is worth changing.

narwhals/series.py
camriddell: Good catch, I made an assumption based on the default function values of `polars...hist(...)`.

narwhals/_polars/series.py
camriddell: I wanted to leave this behavior up to each backend in case polars or some other lib has a fastpath for non-list types (though this is just speculation). 

I am definitely open to moving this logic into a more superficial level so we can guarantee a uniform error for all backends. Would you like me to do this move?

narwhals/_pandas_like/series.py
camriddell: actually, going back through the code and tests- if one directly uses `.value_counts` pandas seems to adjust the passed in bins (which feels like incorrect behavior).

```python
>>> import pandas as pd
>>> pd.__version__
'2.2.3'
>>> pd.Series([0, 1, 2, 3, 4, 5, 6]).value_counts(bins=[0.0, 2.5, 5.0, float('inf')], sort=False)
(-0.001, 2.5]    3 # this should be (0, 2.5] as it was explicitly defined in the `bins` argument
(2.5, 5.0]       3
(5.0, inf]       1
Name: count, dtype: int64
```

camriddell:
Found a new breaking change that is bisected on Polars version 1.5 for `bin_count=...`. I'll manually calculate the bins for backwards compat.

```zsh
❯ python -c 'import polars as pl; print(pl.__version__); print(pl.Series([1,2,3]).hist(bin_count=3))'
1.6.0
shape: (3, 3)
┌────────────┬──────────────────────┬───────┐
│ breakpoint ┆ category             ┆ count │
│ ---        ┆ ---                  ┆ ---   │
│ f64        ┆ cat                  ┆ u32   │
╞════════════╪══════════════════════╪═══════╡
│ 1.666667   ┆ (0.998, 1.666667]    ┆ 1     │
│ 2.333333   ┆ (1.666667, 2.333333] ┆ 1     │
│ 3.0        ┆ (2.333333, 3.0]      ┆ 1     │
└────────────┴──────────────────────┴───────┘
```

```zsh
❯ python -c 'import polars as pl; print(pl.__version__); print(pl.Series([1,2,3]).hist(bin_count=3))'
1.5.0
shape: (4, 3)
┌────────────┬──────────────────────┬───────┐
│ breakpoint ┆ category             ┆ count │
│ ---        ┆ ---                  ┆ ---   │
│ f64        ┆ cat                  ┆ u32   │
╞════════════╪══════════════════════╪═══════╡
│ 0.0        ┆ (-inf, 0.0]          ┆ 0     │
│ 1.333333   ┆ (0.0, 1.333333]      ┆ 1     │
│ 2.666667   ┆ (1.333333, 2.666667] ┆ 1     │
│ inf        ┆ (2.666667, inf]      ┆ 1     │
└────────────┴──────────────────────┴───────┘
```

FBruzzesi:
Amazing discoveries @camriddell , I am so sorry you need to take care of all these edge cases 🙈

MarcoGorelli:
awesome, thanks both!

CI logs show
```
FAILED tests/series_only/hist_test.py::test_hist_count[pyarrow-False-True-params1] - pyarrow.lib.ArrowInvalid: Could not convert <pyarrow.DoubleScalar: -0.006> with type pyarrow.lib.DoubleScalar: did not recognize Python value type when inferring an Arrow data type
```
if it's too much bother to fix, we could just set a minimum pyarrow version for this feature, no big deal

MarcoGorelli (commented):
Very impressive, thanks @camriddell !

I think we just need a little docstring example, then this can be good to go? I can add this later

MarcoGorelli:
I took a closer look, and have a couple of more comments, sorry

For `.hist()` without arguments, it looks like pyarrow produces different results, is this expected?
```
In [25]: nw.from_native(pd.Series([1, 3, 8, 8, 2, 1, 3], name='a'), allow_series=True).hist()
Out[25]: 
┌──────────────────────────────────┐
|        Narwhals DataFrame        |
|----------------------------------|
|   breakpoint      category  count|
|0         1.7  (0.993, 1.7]      2|
|1         2.4    (1.7, 2.4]      1|
|2         3.1    (2.4, 3.1]      2|
|3         3.8    (3.1, 3.8]      0|
|4         4.5    (3.8, 4.5]      0|
|5         5.2    (4.5, 5.2]      0|
|6         5.9    (5.2, 5.9]      0|
|7         6.6    (5.9, 6.6]      0|
|8         7.3    (6.6, 7.3]      0|
|9         8.0    (7.3, 8.0]      2|
└──────────────────────────────────┘

In [26]: nw.from_native(pa.chunked_array([pl.Series(values=[1, 3, 8, 8, 2, 1, 3], name='a').to_arrow()]), allow_series=True).hist().to_polars()
    ...: 
Out[26]: 
shape: (10, 3)
┌────────────┬─────────────────────────────────┬───────┐
│ breakpoint ┆ category                        ┆ count │
│ ---        ┆ ---                             ┆ ---   │
│ f64        ┆ str                             ┆ i64   │
╞════════════╪═════════════════════════════════╪═══════╡
│ 0.7        ┆ (-0.007, 0.7]                   ┆ 2     │
│ 1.4        ┆ (0.7, 1.4]                      ┆ 1     │
│ 2.1        ┆ (1.4, 2.0999999999999996]       ┆ 2     │
│ 2.8        ┆ (2.0999999999999996, 2.8]       ┆ 0     │
│ 3.5        ┆ (2.8, 3.5]                      ┆ 0     │
│ 4.2        ┆ (3.5, 4.2]                      ┆ 0     │
│ 4.9        ┆ (4.199999999999999, 4.89999999… ┆ 0     │
│ 5.6        ┆ (4.8999999999999995, 5.6]       ┆ 0     │
│ 6.3        ┆ (5.6, 6.3]                      ┆ 0     │
│ 7.0        ┆ (6.3, 7.0]                      ┆ 2     │
└────────────┴─────────────────────────────────┴───────┘
```

Second, I'm not totally sure about the categories column being a categorical in Polars in the first place (https://github.com/pola-rs/polars/issues/18645). Do we need it all though? We could always start by just not including that argument and column, and can always add it back later if necessary

narwhals/_arrow/series.py
```diff
@@ -1004,6 +1005,113 @@ def rank(
...
+    ) -> ArrowDataFrame:
+        import numpy as np  # ignore-banned-import
+
+        from narwhals._arrow.dataframe import ArrowDataFrame
+
+        def _hist_from_bin_count(
+            bin_count: int,
+        ) -> tuple[Sequence[int], Sequence[int | float], Sequence[int | float]]:
+            d = pc.min_max(self._native_series)
+            lower, upper = d["min"].as_py(), d["max"].as_py()
```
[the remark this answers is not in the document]
camriddell: Here the `.as_py` was convenient for the algebra that is calculated against a single scalar. Would it be preferable to just use `pyarrow.compute` functions here?

camriddell:
> I took a closer look, and have a couple of more comments, sorry
> 
> For `.hist()` without arguments, it looks like pyarrow produces different results, is this expected?
> 
> ```
> In [25]: nw.from_native(pd.Series([1, 3, 8, 8, 2, 1, 3], name='a'), allow_series=True).hist()
> Out[25]: 
> ┌──────────────────────────────────┐
> |        Narwhals DataFrame        |
> |----------------------------------|
> |   breakpoint      category  count|
> |0         1.7  (0.993, 1.7]      2|
> |1         2.4    (1.7, 2.4]      1|
> |2         3.1    (2.4, 3.1]      2|
> |3         3.8    (3.1, 3.8]      0|
> |4         4.5    (3.8, 4.5]      0|
> |5         5.2    (4.5, 5.2]      0|
> |6         5.9    (5.2, 5.9]      0|
> |7         6.6    (5.9, 6.6]      0|
> |8         7.3    (6.6, 7.3]      0|
> |9         8.0    (7.3, 8.0]      2|
> └──────────────────────────────────┘
> 
> In [26]: nw.from_native(pa.chunked_array([pl.Series(values=[1, 3, 8, 8, 2, 1, 3], name='a').to_arrow()]), allow_series=True).hist().to_polars()
>     ...: 
> Out[26]: 
> shape: (10, 3)
> ┌────────────┬─────────────────────────────────┬───────┐
> │ breakpoint ┆ category                        ┆ count │
> │ ---        ┆ ---                             ┆ ---   │
> │ f64        ┆ str                             ┆ i64   │
> ╞════════════╪═════════════════════════════════╪═══════╡
> │ 0.7        ┆ (-0.007, 0.7]                   ┆ 2     │
> │ 1.4        ┆ (0.7, 1.4]                      ┆ 1     │
> │ 2.1        ┆ (1.4, 2.0999999999999996]       ┆ 2     │
> │ 2.8        ┆ (2.0999999999999996, 2.8]       ┆ 0     │
> │ 3.5        ┆ (2.8, 3.5]                      ┆ 0     │
> │ 4.2        ┆ (3.5, 4.2]                      ┆ 0     │
> │ 4.9        ┆ (4.199999999999999, 4.89999999… ┆ 0     │
> │ 5.6        ┆ (4.8999999999999995, 5.6]       ┆ 0     │
> │ 6.3        ┆ (5.6, 6.3]                      ┆ 0     │
> │ 7.0        ┆ (6.3, 7.0]                      ┆ 2     │
> └────────────┴─────────────────────────────────┴───────┘
> ```
Yep, I know the culprit here (forgot to add the minimum value back in when performing a bin calculation)- I'll add this as a property test as well.


> Second, I'm not totally sure about the categories column being a categorical in Polars in the first place ([pola-rs/polars#18645](https://github.com/pola-rs/polars/issues/18645)). Do we need it all though? We could always start by just not including that argument and column, and can always add it back later if necessary

If I understood some of the discussion going on- aren't we also in trouble by keeping the `"breakpoint"` column around as well? It seems like the new result may just be counts as an integer and the remaining information as a struct.

IMO, the results would be more usable as a struct, but then we lose the closedness information where the string representation encodes 4 pieces of information, left open/closed, right open/closed, left value, right value and a struct of this fashion (or `pandas.IntervalIndex`) loses two of those pieces of information. I'll think on this some more.

narwhals/_arrow/series.py
```diff
@@ -1004,6 +1005,113 @@ def rank(
...
+
+            # extract left/right side of the intervals
+            bin_left = pc.multiply(counts.column("values"), width)
+            bin_right = pc.add(bin_left, width)
+            bin_left = pa.chunked_array(
+                [  # pad lowest bin by 1% of range
+                    [pc.subtract(bin_left[0], (upper - lower) * 0.001).as_py()],
+                    bin_left.to_numpy()[1:],  # pyarrow==0.11.0 needs to infer
+                ]
+            )
```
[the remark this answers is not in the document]
camriddell: Here the `.as_py` was only necessary for old versions (along with the `.to_numpy()`)

MarcoGorelli (approved):
thanks @camriddell !

amazing effort here, and well for uncovering and reporting several upstream bugs!

happy to ship this and see how it goes - thanks for keeping the unstable marker, we can always revisit based on how/if Polars' own API changes

PR #1865, opened 2025-01-25
feat: improving array casting

FBruzzesi:
# What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

There are a few improvements happening in this PR:

- Current implementation of Array is not on par with polars one, and the support of multi dimension arrays is limited
- Introduces:
  - Arrow support for converting to array dtypes
  -  PySpark support for converting to array dtypes
- Fixes
  - polars by supporting newer implementation and backporting for older version
  - duckdb by supporting multidimensional arrays
  - pandas: mostly internal code clean up, the only fix is in `broadcast_and_extract_dataframe_comparand`


narwhals/dtypes.py
```diff
@@ -731,14 +731,37 @@ class Array(DType):
...
 
+    inner: DType | type[DType]
+    size: int
+    shape: tuple[int, ...]
+
     def __init__(
-        self: Self, inner: DType | type[DType], width: int | None = None
+        self: Self,
+        inner: DType | type[DType],
+        shape: int | tuple[int, ...] | None = None,
```
FBruzzesi, line 741: This would be a breaking change :(

narwhals/_arrow/utils.py
```diff
@@ -220,7 +224,7 @@ def broadcast_and_extract_dataframe_comparand(
 
     if isinstance(other, ArrowSeries):
         len_other = len(other)
-        if len_other == 1:
+        if len_other == 1 and length != 1:
```
FBruzzesi, line 227: Otherwise for list/array types we end up getting the first element

narwhals/_duckdb/utils.py
```diff
@@ -111,10 +111,13 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
         )
     if match_ := re.match(r"(.*)\[\]$", duckdb_dtype):
         return dtypes.List(native_to_narwhals_dtype(match_.group(1), version))
-    if match_ := re.match(r"(\w+)\[(\d+)\]", duckdb_dtype):
+    if match_ := re.match(r"(\w+)((?:\[\d+\])+)", duckdb_dtype):
```
FBruzzesi, line 114: Array type in duckdb can also have multiple dimensions, rendering as: `INNER[d1][d2][...]`

narwhals/_duckdb/utils.py
```diff
@@ -171,8 +174,11 @@ def narwhals_to_native_dtype(dtype: DType | type[DType], version: Version) -> st
         )
         return f"STRUCT({inner})"
     if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover
-        msg = "todo"
-        raise NotImplementedError(msg)
+        duckdb_shape_fmt = "".join(f"[{item}]" for item in dtype.shape)  # type: ignore[union-attr]
+        while isinstance(dtype.inner, dtypes.Array):  # type: ignore[union-attr]
+            dtype = dtype.inner  # type: ignore[union-attr]
+        inner = narwhals_to_native_dtype(dtype.inner, version)  # type: ignore[union-attr]
+        return f"{inner}{duckdb_shape_fmt}"
```
FBruzzesi, line 181: First creates the shape `[d1][d2]...` then find the inner type recursively (first being non array)

narwhals/_pandas_like/utils.py
```diff
@@ -160,7 +160,7 @@ def broadcast_and_extract_dataframe_comparand(index: Any, other: Any) -> Any:
     if isinstance(other, PandasLikeSeries):
         len_other = other.len()
 
-        if len_other == 1:
+        if len_other == 1 and len(index) != 1:
```
FBruzzesi, line 163: Similarly as for pyarrow, otherwise for list/array types we end up getting the first element

narwhals/_spark_like/utils.py
```diff
@@ -42,26 +42,24 @@ def native_to_narwhals_dtype(
...
-    if any(isinstance(dtype, t) for t in datetime_types):
+    if isinstance(dtype, (pyspark_types.TimestampType, pyspark_types.TimestampNTZType)):
         return dtypes.Datetime()
     if isinstance(dtype, pyspark_types.DecimalType):  # pragma: no cover
         # TODO(unassigned): cover this in dtypes_test.py
         return dtypes.Decimal()
+    if isinstance(dtype, pyspark_types.ArrayType):  # pragma: no cover
+        return dtypes.List(
+            inner=native_to_narwhals_dtype(dtype.elementType, version=version)
+        )
```
FBruzzesi, line 62: Spark arrays do not have a fixed dimension, so I am converting to a list

narwhals/_pandas_like/utils.py
```diff
@@ -649,68 +647,25 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
-                )
-            )
-        else:  # pragma: no cover
-            msg = (
-                "Converting to List dtype is not supported for implementation "
-                f"{implementation} and version {version}."
-            )
-            return NotImplementedError(msg)
-    if isinstance_or_issubclass(dtype, dtypes.Struct):
+    if isinstance_or_issubclass(dtype, (dtypes.Struct, dtypes.Array, dtypes.List)):
```
FBruzzesi, line 650: This diff is quite nice 😉

narwhals/dtypes.py
```diff
@@ -751,16 +774,24 @@ def __eq__(self: Self, other: DType | type[DType]) -> bool:  # type: ignore[over
...
+    def __repr__(self) -> str:
+        # Get leaf type
+        dtype = self.inner
+        while isinstance(dtype, Array):
+            dtype = dtype.inner
 
-    def __repr__(self: Self) -> str:
         class_name = self.__class__.__name__
-        return f"{class_name}({self.inner!r}, {self.width})"
+        return f"{class_name}({dtype!r}, shape={self.shape})"
```
FBruzzesi, line 794: This diff is that's causing marimo ci to fail

FBruzzesi:
Sporadically coming back to this to avoid falling too far back in conflicts and diffs to solve.

I would love to have as much support as possible for converting between native and narwhals datatypes/schemas. The only issue is marimo CI failing. So far they seemed pretty open to the changes, so we could just provide a fix for them.

Happy to hear feedback on this

MarcoGorelli:
Thanks - sorry didn't get a chance to think about this too much

Is this mainly for ergonomics?

narwhals/dtypes.py
MarcoGorelli: It also wouldn't be too bad to override repr in v1 right?

FBruzzesi:
> Thanks - sorry didn't get a chance to think about this too much

No worries @MarcoGorelli, we have other priorities and there is no rush! We will get everything sooner or later!

> Is this mainly for ergonomics?

I sense that the PR title is completely misleading! Actually it is much more than ergonomics. There are a few improvements happening:

- Current implementation of Array is not on par with polars one, and the support of multi dimension arrays is limited
- Introduces
  - Arrow support for converting to array dtypes
  -  PySpark support for converting to array dtypes
- Fix polars, pandas and duckdb support for arrays
  
I will add this description in the PR body on top

MarcoGorelli:
just took a look at the failing test:
```python
    def test_complex_data_field_types(self) -> None:
        complex_data = self.get_complex_data()
        field_types = complex_data.get_field_types()
>       snapshot("narwhals.field_types.json", json.dumps(field_types))
```
I think that, if we provide a fix, this is probably fine to update, it wouldn't a Marimo-user-facing change

narwhals/dtypes.py
```diff
@@ -765,7 +765,7 @@ class Array(NestedType):
 
     Arguments:
         inner: The datatype of the values within each array.
-        width: the length of each array.
+        shape: the length of each array.
```
MarcoGorelli, line 768: the shape of each array?

narwhals/dtypes.py
```diff
@@ -805,16 +828,24 @@ def __eq__(self: Self, other: DType | type[DType]) -> bool:  # type: ignore[over
...
 
     def __hash__(self: Self) -> int:
-        return hash((self.__class__, self.inner, self.width))
+        return hash((self.__class__, self.inner, self.shape))
+
+    def __repr__(self) -> str:
+        # Get leaf type
+        dtype = self.inner
+        while isinstance(dtype, Array):
+            dtype = dtype.inner
```
MarcoGorelli, line 845: can we put a limit on the depth of this (just to avoid some infinite loop in some unexpected scenario)?
same with the other place where there's the while loop

MarcoGorelli (approved):
thanks, this is probably good to do sooner rather than later (while nobody is explicitly using the array type's shape/width)

just left some minor comments, but then I think that, if combined with a PR to Marimo, this can be OK - for Marimo users it really only affects the repr, makes it more aligned with Polars

MarcoGorelli:
we'll already need to update the Marimo test suite for https://github.com/narwhals-dev/narwhals/pull/1918 anyway, we could bundle this with that

narwhals/dtypes.py
FBruzzesi: This in particular is the same as polars does it: https://github.com/pola-rs/polars/blob/faad12f7277751006e3faebf0fffb1f6bf9aa7e7/py-polars/polars/datatypes/classes.py#L905-L912

and the `__init__` is also recursive:

https://github.com/narwhals-dev/narwhals/blob/981f87c30a7d9284f5730ddd36c5162e111de194/narwhals/dtypes.py#L808-L810

In principle, max depth should be `len(shape)`. Is that a good depth to check? WDYT?

MarcoGorelli: i was thinking something like

https://github.com/psf/black/blob/5f2370170819d282ec14dcda70f963d7574271e2/src/black/handle_ipynb_magics.py#L217-L226

FBruzzesi: I remember the trick, yet I don't feel super comfortable with introducing a fixed number of depth to check for.

Since we know already that the max depth should be `len(shape)`, I think that would be the best limit to introduce, if any.

In practice, if someone generate a datatype to pass which ends up in an infinite recursion, maybe it's good to raise with sure error instead of an `AssertionError`. I am thinking out loud here though

narwhals/dtypes.py
FBruzzesi: Should we maintain `width` in v1 then? My understanding is that we are limited to 1d arrays with the v1 implementation.
Or should we just keep the `__repr__` and show `f"{class_name}({self.inner!r}, {width})"` with `width=self.shape[0]` if shape has only one dimension else the tuple `self.shape`?

narwhals/dtypes.py
MarcoGorelli: Sure, shall we use len shape? 

In theory this shouldn't happen, just trying to guard against implementation oversights

MarcoGorelli:
nice, thanks @FBruzzesi ! going to ship this one then 🚢 

if this causes issues for anyone, please let us know and we can backport some changes to the stable.v1 api

PR #1866, opened 2025-01-25
feat: completely refactor alias tracking and support `nw.all`, `nw.nth`, and selectors across the API

MarcoGorelli:
https://youtu.be/0JGpveomN9A

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


tests/group_by_test.py
FBruzzesi, line 1: 🎻

FBruzzesi (commented):
Amazing stuff! I tried to review one lazy and one eager backend in depth, and then assumed that the same changes apply to all others - I also re-watched the live streaming to get a better high level understanding of it!

I was expecting to see some other tests to XPASS for pyspark, I will try them locally later today!

As a bonus, I think there could be a good opportunity to factor out name namespace and selectors, but definitly let's do it in another PR.

narwhals/_dask/expr.py
```diff
@@ -135,48 +134,51 @@ def _from_call(
         expr_name: str,
         *,
         returns_scalar: bool,
-        **kwargs: Any,
+        **expressibiable_args: Self | Any,
```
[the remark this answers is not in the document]
MarcoGorelli: thanks!

tests/expr_and_series/name/map_test.py
```diff
@@ -20,29 +14,20 @@ def map_func(s: str | None) -> str:
 def test_map(constructor: Constructor) -> None:
     df = nw.from_native(constructor(data))
     result = df.select((nw.col("foo", "BAR") * 2).name.map(function=map_func))
-    expected = {map_func(k): [e * 2 for e in v] for k, v in data.items()}
+    expected = {"oof": [2, 4, 6], "rab": [8, 10, 12]}
```
MarcoGorelli, line 17: @FBruzzesi for the future, i'd much prefer to constructor expected output explicitly (no blame of course! just a consideration for future developments)

FBruzzesi: Of course! I think this is quite old (with respect to narwhals development), after which we started to always explicitly build the expected dictionary 👌

narwhals/_dask/expr.py
```diff
@@ -659,29 +661,28 @@ def null_count(self: Self) -> Self:
...
+            _, aliases = evaluate_output_names_and_aliases(self, df, keys)
             if df._native_frame.npartitions == 1:  # pragma: no cover
                 tmp = df.group_by(*keys, drop_null_keys=False).agg(self)
                 tmp_native = (
                     df.simple_select(*keys)
                     .join(tmp, how="left", left_on=keys, right_on=keys, suffix="_right")
                     ._native_frame
                 )
-                return [tmp_native[name] for name in self._output_names]
+                return [tmp_native[name] for name in aliases]
```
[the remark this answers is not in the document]
MarcoGorelli: we don't need `keys` in the output here - for example, in
```python
nw.col('a').cum_sum().over('b')
```
then `keys` is `['b']`, but the output should only include `'a'`

narwhals/_dask/selectors.py
```diff
@@ -2,15 +2,12 @@
...
+from typing import Sequence
 
 from narwhals._dask.expr import DaskExpr
 from narwhals.utils import import_dtypes_module
 
 if TYPE_CHECKING:
-    try:
-        import dask.dataframe.dask_expr as dx
-    except ModuleNotFoundError:
-        import dask_expr as dx
```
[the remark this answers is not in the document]
MarcoGorelli: 😄 i'll put the annotation back, this was a mistake

MarcoGorelli:
thanks @FBruzzesi !

> I was expecting to see some other tests to XPASS for pyspark, I will try them locally later today!

I think we need a further refactor for that to happen, so that we only use `.over` when absolutely necessary

narwhals/_spark_like/expr.py
```diff
@@ -122,152 +121,153 @@ def _from_call(
...
+            native_results: list[Column] = []
+            native_series_list = self._call(df)
+            other_native_series = {
+                key: maybe_evaluate(df, value)
+                for key, value in expressifiable_args.items()
+            }
+            for native_series in native_series_list:
+                column_result = call(native_series, **other_native_series)
+                native_results.append(column_result)
+            return native_results
```
FBruzzesi, line 136:
```suggestion
            return [
                call(native_series, **other_native_series)
                for native_series in native_series_list
            ]
```

narwhals/_duckdb/expr.py
```diff
@@ -127,44 +125,30 @@ def _from_call(
...
+            native_results: list[duckdb.Expression] = []
+            native_series_list = self._call(df)
+            other_native_series = {
+                key: maybe_evaluate(df, value)
+                for key, value in expressifiable_args.items()
+            }
+            for native_series in native_series_list:
+                column_result = call(native_series, **other_native_series)
+                native_results.append(column_result)
+            return native_results
```
FBruzzesi, line 140:
```suggestion
            return [
                call(native_series, **other_native_series)
                for native_series in native_series_list
            ]
```

FBruzzesi (commented):
Noticed a couple of list comprehensions

PR #1870, opened 2025-01-26
fix: support various reductions in pyspark

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -8,6 +8,7 @@ from typing import Sequence
 
 from narwhals._spark_like.utils import native_to_narwhals_dtype
 from narwhals._spark_like.utils import parse_exprs_and_named_exprs
+from narwhals.typing import CompliantLazyFrame
 from narwhals.utils import Implementation
 from narwhals.utils import check_column_exists
 from narwhals.utils import parse_columns_to_drop
@@ -26,7 +27,6 @@ if TYPE_CHECKING:
     from narwhals._spark_like.namespace import SparkLikeNamespace
     from narwhals.dtypes import DType
     from narwhals.utils import Version
-from narwhals.typing import CompliantLazyFrame
 
 
 class SparkLikeLazyFrame(CompliantLazyFrame):
@@ -94,7 +94,9 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
         *exprs: SparkLikeExpr,
         **named_exprs: SparkLikeExpr,
     ) -> Self:
-        new_columns = parse_exprs_and_named_exprs(self, *exprs, **named_exprs)
+        new_columns = parse_exprs_and_named_exprs(
+            self, *exprs, with_columns_context=False, **named_exprs
+        )
 
         if not new_columns:
             # return empty dataframe, like Polars does
@@ -135,7 +137,9 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
         *exprs: SparkLikeExpr,
         **named_exprs: SparkLikeExpr,
     ) -> Self:
-        new_columns_map = parse_exprs_and_named_exprs(self, *exprs, **named_exprs)
+        new_columns_map = parse_exprs_and_named_exprs(
+            self, *exprs, with_columns_context=True, **named_exprs
+        )
         return self._from_native_frame(self._native_frame.withColumns(new_columns_map))
 
     def drop(self: Self, columns: list[str], strict: bool) -> Self:  # noqa: FBT001
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -182,7 +182,7 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
         from narwhals._spark_like.group_by import SparkLikeLazyGroupBy
 
         return SparkLikeLazyGroupBy(
-            df=self, keys=list(keys), drop_null_keys=drop_null_keys
+            compliant_frame=self, keys=list(keys), drop_null_keys=drop_null_keys
         )
 
     def sort(
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -126,7 +126,7 @@ class SparkLikeExpr(CompliantExpr["Column"]):
         def func(df: SparkLikeLazyFrame) -> list[Column]:
             native_series_list = self._call(df)
             other_native_series = {
-                key: maybe_evaluate(df, value)
+                key: maybe_evaluate(df, value, returns_scalar=returns_scalar)
                 for key, value in expressifiable_args.items()
             }
             return [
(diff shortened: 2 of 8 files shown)


narwhals/_spark_like/utils.py
```diff
@@ -110,9 +111,18 @@ def narwhals_to_native_dtype(
 
 
 def parse_exprs_and_named_exprs(
-    df: SparkLikeLazyFrame, *exprs: SparkLikeExpr, **named_exprs: SparkLikeExpr
+    df: SparkLikeLazyFrame,
+    *exprs: SparkLikeExpr,
+    with_columns_context: bool,
```
FBruzzesi, line 116: I guess one bug could be if someone names a column `with_columns_context` and actually same for `df`

MarcoGorelli: we could address this by making `df` and `with_columns_context` positional-only arguments?

narwhals/_spark_like/utils.py
```diff
@@ -122,16 +132,29 @@ def parse_exprs_and_named_exprs(
...
 
+    if all(returns_scalar) and not with_columns_context:
+        return native_results
+    else:
+        return {
+            col_name: col.over(Window.partitionBy(F.lit(1))) if _returns_scalar else col
+            for (col_name, col), _returns_scalar in zip(
+                native_results.items(), returns_scalar
+            )
+        }
```
MarcoGorelli, line 154: i'm not sure if this is the right place to do this

if we do `select` with all aggregations, we should just call `native_frame.agg`, which is what we do for DuckDB

MarcoGorelli: i think perhaps we should:
- only broadcast (which for pyspark means ".over(Window.partitionBy(F.lit(1))") if there's binary operation
- in `select`, call `native_frame.agg` if all expressions are aggregations
- in `group_by`, just pass the expressions though (like we do for duckdb)

narwhals/_spark_like/utils.py
FBruzzesi: Just read this comment, I still think it would not solve the issue.
I made a fix in #1871

narwhals/_spark_like/utils.py
FBruzzesi: TIL about pyspark `agg`.

It makes sense, I will check a few edge cases and differences to take into account between `select` and `with_columns`. 
For group by, I think the logic is still a bit rudimental and we are not passing the expression itself. We might need to come back to that

narwhals/_spark_like/dataframe.py
```diff
@@ -107,8 +110,38 @@ def select(
...
+                col.over(Window.partitionBy(F.lit(1))).alias(col_name)
+                if _returns_scalar
+                else col.alias(col_name)
+                for (col_name, col), _returns_scalar in zip(
+                    new_columns.items(), returns_scalar
+                )
+            ]
+            return self._from_native_frame(self._native_frame.select(*new_columns_list))
+
+    def with_columns(
```
FBruzzesi, line 129: I just moved this closer to `select` method - it was easier to debug them while in the same screen 🙈

narwhals/_spark_like/expr.py
```diff
@@ -140,7 +140,7 @@ def func(df: SparkLikeLazyFrame) -> list[Column]:
             function_name=f"{self._function_name}->{expr_name}",
             evaluate_output_names=self._evaluate_output_names,
             alias_output_names=self._alias_output_names,
-            returns_scalar=self._returns_scalar or returns_scalar,
+            returns_scalar=returns_scalar,
```
FBruzzesi, line 143: This gave me so much headache before spotting it

narwhals/_spark_like/dataframe.py
```diff
@@ -105,8 +110,38 @@ def select(
...
+            return self._from_native_frame(self._native_frame.agg(*new_columns_list))
+        else:
+            new_columns_list = [
+                col.over(Window.partitionBy(F.lit(1))).alias(col_name)
+                if _returns_scalar
+                else col.alias(col_name)
+                for (col_name, col), _returns_scalar in zip(
+                    new_columns.items(), returns_scalar
+                )
+            ]
```
MarcoGorelli, line 126: i think this is may be too late to set `over`

for example, in `nw.col('a') - nw.col('a').mean()`  - I think it's in the binary operation `__sub__` that `nw.col('a').mean()` needs to become `nw.col('a').mean().over(lit(1))`

as in, we want to translate `nw.col('a') - nw.col('a').mean()` to `F.col('a') - F.col('a').mean().over(F.lit(1))`. the code, however, as far as I can tell, translates it to `(F.col('a') - F.col('a').mean()).over(F.lit(1))`

FBruzzesi: That happens in `maybe_evaluate`, and you can see that now the `reduction_test` are passing.
I know it's not ideal to have it in two places, but I couldn't figure out a unique place in which to handle this as `maybe_evaluate` is called only to evaluate _other_ arguments

MarcoGorelli: ah I see! yes this might be fine then, thanks!

narwhals/_spark_like/utils.py
```diff
@@ -112,9 +113,16 @@ def narwhals_to_native_dtype(
...
-) -> Callable[..., dict[str, Column]]:
-    def func(*exprs: SparkLikeExpr, **named_exprs: SparkLikeExpr) -> dict[str, Column]:
+) -> Callable[..., tuple[dict[str, Column], list[bool]]]:
+    def func(
+        *exprs: SparkLikeExpr, **named_exprs: SparkLikeExpr
+    ) -> tuple[dict[str, Column], list[bool]]:
         native_results: dict[str, list[Column]] = {}
+
+        # `returns_scalar` keeps track if an expression returns a scalar and is not lit.
+        # Notice that lit is quite special case, since it gets broadcasted by pyspark
```
MarcoGorelli, line 123: Out of interest, do we run into any issues if we use `over` anyway with `lit`? No objections to special casing it, just curious

MarcoGorelli (commented):
The group by simplification is wonderful 😻 thanks so much for doing this!

narwhals/_spark_like/utils.py
FBruzzesi: We end up with `tests/expr_and_series/lit_test.py` failing 3 tests due to:

> pyspark.errors.exceptions.captured.AnalysisException: [UNSUPPORTED_EXPR_FOR_WINDOW] Expression "1" not supported within a window function.;

MarcoGorelli: isn't this still going to break for, say
```
df.with_columns(nw.lit(2)+1)
```
?

FBruzzesi: Yes correct

narwhals/_spark_like/utils.py
```diff
@@ -141,8 +161,9 @@ def maybe_evaluate(df: SparkLikeLazyFrame, obj: Any) -> Any:
             msg = "Multi-output expressions (e.g. `nw.all()` or `nw.col('a', 'b')`) not supported in this context"
             raise NotImplementedError(msg)
         column_result = column_results[0]
-        if obj._returns_scalar:
-            # Return scalar, let PySpark do its broadcasting
+        if obj._returns_scalar and obj._function_name != "lit" and not returns_scalar:
```
MarcoGorelli, line 164: do we need the `'->'` split here too?

MarcoGorelli (approved):
nice, well done @FBruzzesi !

just one comment/question

this seems correct! or at least, I couldn't think of a way to break it

narwhals/_spark_like/utils.py
FBruzzesi: I guess it is safe to do so!

MarcoGorelli:
cool, release? 🚀

PR #1887, opened 2025-01-28
feat: spark like `.list` namespace and lazyframe `.explode`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1714 

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -12,6 +12,7 @@ from narwhals._spark_like.utils import parse_exprs_and_named_exprs
 from narwhals.typing import CompliantLazyFrame
 from narwhals.utils import Implementation
 from narwhals.utils import check_column_exists
+from narwhals.utils import import_dtypes_module
 from narwhals.utils import parse_columns_to_drop
 from narwhals.utils import parse_version
 from narwhals.utils import validate_backend_version
@@ -19,6 +20,7 @@ from narwhals.utils import validate_backend_version
 if TYPE_CHECKING:
     from types import ModuleType
 
+    from pyspark.sql import Column
     from pyspark.sql import DataFrame
     from typing_extensions import Self
 
@@ -46,7 +48,7 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
         validate_backend_version(self._implementation, self._backend_version)
 
     @property
-    def _F(self) -> Any:  # noqa: N802
+    def _F(self: Self) -> Any:  # noqa: N802
         if self._implementation is Implementation.SQLFRAME:
             from sqlframe.duckdb import functions
 
@@ -56,7 +58,7 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
         return functions
 
     @property
-    def _native_dtypes(self) -> Any:
+    def _native_dtypes(self: Self) -> Any:
         if self._implementation is Implementation.SQLFRAME:
             from sqlframe.duckdb import types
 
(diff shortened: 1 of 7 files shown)


narwhals/_spark_like/dataframe.py
```diff
@@ -312,3 +314,57 @@ def join(
...
+                        ]
+                    )
+                )
+            )
+
+        msg = (
+            "Exploding on multiple columns is not supported with SparkLike backend since "
+            "we cannot guarantee that the exploded columns have matching element counts."
+        )
+        raise NotImplementedError(msg)
```
FBruzzesi, line 370: In a way... this is so sad since I had a great way of doing it! But cannot verify that the element counts are same without triggering a collect

narwhals/_spark_like/dataframe.py
```diff
@@ -312,3 +314,57 @@ def join(
...
+        if len(to_explode) == 1:
+            return self._from_native_frame(
+                native_frame.select(
+                    *[
+                        self._F.col(col_name).alias(col_name)
+                        if col_name != to_explode[0]
+                        else self._F.explode(col_name).alias(col_name)
+                        for col_name in column_names
+                    ]
+                ).union(
```
FBruzzesi, line 354: Since order is not guaranteed 😇

narwhals/_spark_like/dataframe.py
```diff
@@ -312,3 +314,57 @@ def join(
...
+                )
+                raise InvalidOperationError(msg)
+
+        native_frame = self._native_frame
+        column_names = self.columns
+
+        def null_condition(col_name: str) -> Column:
+            return self._F.isnull(col_name) | (self._F.array_size(col_name) == 0)
+
+        if len(to_explode) == 1:
```
EdAbati, line 345: minor: this could save us 1 indentation level :D

```python
if len(to_explode) != 1:
    msg = (
            "Exploding on multiple columns is not supported with SparkLike backend since "
            "we cannot guarantee that the exploded columns have matching element counts."
        )
```

I'd have enough time to check it. But if you think we cannot find a way to make it work on multiple columns, we could even raise the error before doing the check for dytpes.

narwhals/_spark_like/dataframe.py
```diff
@@ -312,3 +314,57 @@ def join(
...
+            return self._from_native_frame(
+                native_frame.select(
+                    *[
+                        self._F.col(col_name).alias(col_name)
+                        if col_name != to_explode[0]
+                        else self._F.explode(col_name).alias(col_name)
+                        for col_name in column_names
+                    ]
+                ).union(
+                    native_frame.filter(null_condition(to_explode[0])).select(
```
EdAbati, line 355: I think [`explode_outer`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.explode_outer.html) could save us this union 👀
Or am I missing something?

EdAbati (commented):
Thaaank you for this! Just added a couple of comments/questions

narwhals/_spark_like/dataframe.py
FBruzzesi: TIL about this function! Thanks 😁

EdAbati (approved):
This looks 👌👌 to me

PR #1890, opened 2025-01-28
feat: add `LazyFrame.unpivot` for spark and duckdb

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes


narwhals/_duckdb/dataframe.py
```diff
@@ -388,10 +386,57 @@ def sort(
...
+        variable_name = variable_name if variable_name is not None else "variable"
+        value_name = value_name if value_name is not None else "value"
+
+        if variable_name == "":
+            msg = "`variable_name` cannot be empty string for duckdb backend."
+            raise NotImplementedError(msg)
+
+        if value_name == "":
+            msg = "`value_name` cannot be empty string for duckdb backend."
+            raise NotImplementedError(msg)
```
FBruzzesi, line 423: I tried, but it is not of duckdb liking to support these

MarcoGorelli:
thanks, can you rebase please?

narwhals/_duckdb/dataframe.py
```diff
@@ -388,10 +386,57 @@ def sort(
...
+        value_name: str | None,
+    ) -> Self:
+        on_ = [on] if isinstance(on, str) else on
+        index_ = (
+            [index]
+            if isinstance(index, str)
+            else index
+            if isinstance(index, list)
+            else []
+        )
```
MarcoGorelli, line 409: separate to this pr,  but we should be doing this one level above

narwhals/_duckdb/dataframe.py
```diff
@@ -388,10 +386,57 @@ def sort(
...
+
+        if on_ is None:
+            on_ = [c for c in self.columns if c not in index_]
+
+        variable_name = variable_name if variable_name is not None else "variable"
+        value_name = value_name if value_name is not None else "value"
```
MarcoGorelli, line 415: same here probably (one level above)

narwhals/_spark_like/dataframe.py
```diff
@@ -312,3 +312,21 @@ def join(
         return self._from_native_frame(
             self_native.join(other, on=left_on, how=how).select(col_order)
         )
+
+    def unpivot(
```
MarcoGorelli, line 316: wow, so simple!

MarcoGorelli (commented):
awesome work, thanks!

narwhals/dataframe.py
```diff
@@ -330,10 +330,24 @@ def unpivot(
...
+        index_ = (
+            [index]
+            if isinstance(index, str)
+            else index
+            if isinstance(index, list)
+            else []
+        )
+
+        if on_ is None:
+            on_ = [c for c in self.collect_schema().names() if c not in index_]
```
FBruzzesi, line 343: @MarcoGorelli by pushing one level up, we need column names 🙃

MarcoGorelli: ah sorry i don't think we need to push this one, just the `str | list[str]` ones

narwhals/dataframe.py
```diff
@@ -330,10 +330,22 @@ def unpivot(
...
+        index_ = (
+            [index]
+            if isinstance(index, str)
+            else index
+            if isinstance(index, list)
+            else []
+        )
+
+        variable_name = variable_name if variable_name is not None else "variable"
+        value_name = value_name if value_name is not None else "value"
```
MarcoGorelli, line 343: 🤔 I don't understand why the signature isn't `variable_name: str = "variable"` in Polars, am I missing something?

MarcoGorelli (approved):
thanks @FBruzzesi ! feel free to merge on green

got a question which occurred to me, we can always simplify later

narwhals/dataframe.py
```diff
@@ -330,10 +330,22 @@ def unpivot(
...
         value_name: str | None,
     ) -> Self:
+        on_ = [on] if isinstance(on, str) else on
+        index_ = (
+            [index]
+            if isinstance(index, str)
+            else index
+            if isinstance(index, list)
+            else []
+        )
```
MarcoGorelli, line 340: this might be causing issues (see plotly failing test)?

FBruzzesi: Alright, I just pushed the variable and value name parsing on the BaseFrame level. `index` and `on` require parsing only for arrow and duckdb anyway. I think it's fine to keep it at the compliant level

narwhals/dataframe.py
FBruzzesi: Good point! Not sure why they do it like that

PR #1891, opened 2025-01-28
feat: add `LazyFrame.explode` for duckdb

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes


narwhals/_duckdb/dataframe.py
```diff
@@ -395,3 +396,52 @@ def drop_nulls(self: Self, subset: list[str] | None) -> Self:
...
+                select {select_unnest_statement}
+                from rel
+                where {where_condition}
+            )
+            union
+            (
+                select {select_null_statement}
+                from rel
+                where not ({where_condition})
+            );
```
FBruzzesi, line 445: Similar logic to the one mentioned in the pyspark PR, since the order should not matter

FBruzzesi:
Ahhh also duckdb does not raise for un-equal element counts 🥲 so we should dis-allow to pass multiple columns

MarcoGorelli:
> so we should dis-allow to pass multiple columns

sure, looks good (and is what we do for pyspark?) - merge conflict then I think we can ship it?

kinda wondering if we should disallow `explode` with multiple columns globally for `LazyFrame`?

FBruzzesi:
> kinda wondering if we should disallow `explode` with multiple columns globally for `LazyFrame`? or rather, change the signature so it just accepts a single column name, so then people don't even expect multi-column explode to work tbh I think this is ok

I generally love that polars allows these nasty stuff 😂 but I agree that unless for polars, we don't have a way to check it for the other lazy backends

narwhals/_duckdb/dataframe.py
```diff
@@ -427,6 +428,62 @@ def drop_nulls(self: Self, subset: list[str] | None) -> Self:
...
+    def explode(self: Self, columns: str | Sequence[str], *more_columns: str) -> Self:
+        from narwhals.exceptions import InvalidOperationError
+
+        dtypes = import_dtypes_module(self._version)
+
+        to_explode = (
+            [columns, *more_columns]
+            if isinstance(columns, str)
+            else [*columns, *more_columns]
+        )
```
MarcoGorelli, line 440: sorry, one more - this again should be one level up, right?

tests/frame/explode_test.py
```diff
@@ -54,7 +54,11 @@ def test_explode_single_col(
         .select("a", column)
     )
     expected = {"a": ["x", "x", "y", "z", "w"], column: expected_values}
-    assert_equal_data(result, expected)
+    if "duckdb" in str(constructor):
+        for col in result.columns:
+            assert_equal_data(result.select(col), {col: expected[col]})
```
MarcoGorelli, line 59: i'm confused sorry, why are we doing this?

narwhals/_duckdb/dataframe.py
MarcoGorelli: this looks fine, I just wonder if we can use the Python API for this? looks like `duckdb.FunctionExpression('unnest', duckdb.ColumnExpression('a'))` is supported, as well as `rel.union` and `rel.filter`

MarcoGorelli (commented):
thanks @FBruzzesi , nice one! just got a question

tests/frame/explode_test.py
FBruzzesi: because of the `<non-null> union <null>`, we are changing the order of the other columns as well

narwhals/_duckdb/dataframe.py
FBruzzesi: Gave it a try and it would look like the following:


```py
 non_null_rel = rel.filter(
    col_to_explode.isnotnull() & FunctionExpression("len", col_to_explode)>0
    ).select(*[
        FunctionExpression("unnest", col_to_explode).alias(col) if col in columns else col
        for col in original_columns
    ]
)

null_rel = rel.filter(
    ~(col_to_explode.isnotnull() & FunctionExpression("len", col_to_explode)>0)
    ).select(*[
        ConstantExpression(None).alias(col) if col in columns else col
        for col in original_columns
    ]
)

return self._from_native_frame(non_null_rel.union(null_rel))
```

tests/frame/explode_test.py
MarcoGorelli: could we sort by `'a'` or something like that?

MarcoGorelli (approved):
nice, thanks @FBruzzesi !

PR #1892, opened 2025-01-29
feat: spark `date` and `datetime`data types.

osoucy:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [ ] Documented the changes

Implemented `date` and `datetime` data types for pyspark and enabled tests to validate implementation. 

Enabled setting `dtype` on spark `lit` expression. I have seen the some areas where `_native_types` is used, but I did not see how I could make it work in the context of `lit` namespace.

I was working on fixing some of the issues with the `lit` namespace as well, but when I was about the open the PR, I realized fixes for the same issue have already been merged in the main branch. I left an additional parametrization in (`test_lit`) that could be helpful in the future.

Also, was there any specific reason why spark tests were disabled on python 3.12?

### DIFF
--- a/narwhals/_spark_like/namespace.py
+++ b/narwhals/_spark_like/namespace.py
@@ -16,6 +16,7 @@ from narwhals._spark_like.expr import SparkLikeExpr
 from narwhals._spark_like.selectors import SparkLikeSelectorNamespace
 from narwhals._spark_like.utils import ExprKind
 from narwhals._spark_like.utils import n_ary_operation_expr_kind
+from narwhals._spark_like.utils import narwhals_to_native_dtype
 from narwhals.typing import CompliantNamespace
 
 if TYPE_CHECKING:
@@ -80,12 +81,17 @@ class SparkLikeNamespace(CompliantNamespace["Column"]):
         )
 
     def lit(self: Self, value: object, dtype: DType | None) -> SparkLikeExpr:
-        if dtype is not None:
-            msg = "todo"
-            raise NotImplementedError(msg)
-
         def _lit(df: SparkLikeLazyFrame) -> list[Column]:
-            return [df._F.lit(value)]
+            column = df._F.lit(value)
+            if dtype:
+                import pyspark.sql.types as pyspark_types
+
+                native_dtype = narwhals_to_native_dtype(
+                    dtype, self._version, pyspark_types
+                )
+                column = column.cast(native_dtype)
+
+            return [column.alias("literal")]
 
         return SparkLikeExpr(
             call=_lit,
--- a/narwhals/_spark_like/namespace.py
+++ b/narwhals/_spark_like/namespace.py
@@ -91,7 +91,7 @@ class SparkLikeNamespace(CompliantNamespace["Column"]):
                 )
                 column = column.cast(native_dtype)
 
-            return [column.alias("literal")]
+            return [column]
 
         return SparkLikeExpr(
             call=_lit,
--- a/narwhals/_spark_like/namespace.py
+++ b/narwhals/_spark_like/namespace.py
@@ -84,10 +84,8 @@ class SparkLikeNamespace(CompliantNamespace["Column"]):
         def _lit(df: SparkLikeLazyFrame) -> list[Column]:
             column = df._F.lit(value)
             if dtype:
-                import pyspark.sql.types as pyspark_types
-
                 native_dtype = narwhals_to_native_dtype(
-                    dtype, self._version, pyspark_types
+                    dtype, version=self._version, spark_types=df._native_dtypes
                 )
                 column = column.cast(native_dtype)
 
--- a/narwhals/_spark_like/utils.py
+++ b/narwhals/_spark_like/utils.py
@@ -104,6 +104,9 @@ def narwhals_to_native_dtype(
     if isinstance_or_issubclass(dtype, dtypes.Date):
         return spark_types.DateType()
     if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        dt_time_zone = getattr(dtype, "time_zone", None)
+        if dt_time_zone is None:
+            return spark_types.TimestampNTZType()
         return spark_types.TimestampType()
     if isinstance_or_issubclass(dtype, dtypes.List):  # pragma: no cover
         msg = "Converting to List dtype is not supported yet"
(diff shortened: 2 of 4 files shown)


MarcoGorelli:
thanks @osoucy , will take a closer look later!

> Also, was there any specific reason why spark tests were disabled on python 3.12?

this may have just been from when we tried it on an older version and Python 3.12 wasn't supported - if pyspark supports python 3.12, we can re-enable

narwhals/_spark_like/namespace.py
```diff
@@ -80,12 +81,17 @@ def nth(self: Self, *column_indices: int) -> SparkLikeExpr:
...
         def _lit(df: SparkLikeLazyFrame) -> list[Column]:
-            return [df._F.lit(value)]
+            column = df._F.lit(value)
+            if dtype:
+                import pyspark.sql.types as pyspark_types
+
+                native_dtype = narwhals_to_native_dtype(
+                    dtype, self._version, pyspark_types
+                )
+                column = column.cast(native_dtype)
```
FBruzzesi, line 92: This should not load pyspark types directly anymore. You can do:
```suggestion
            if dtype:
                spark_like_types = df._native_dtypes

                native_dtype = narwhals_to_native_dtype(
                    dtype, self._version, spark_like_types
                )
                column = column.cast(native_dtype)
```

narwhals/_spark_like/namespace.py
```diff
@@ -80,12 +81,17 @@ def nth(self: Self, *column_indices: int) -> SparkLikeExpr:
...
+
+            return [column.alias("literal")]
```
FBruzzesi, line 94: We take care of the aliasing via `evaluate_output_names` below - aliasing here was causing some issues when composing with other functions.

We can just return:
```suggestion
            return [column]
```

narwhals/_spark_like/utils.py
```diff
@@ -101,9 +101,10 @@ def narwhals_to_native_dtype(
         return spark_types.StringType()
     if isinstance_or_issubclass(dtype, dtypes.Boolean):
         return spark_types.BooleanType()
-    if isinstance_or_issubclass(dtype, (dtypes.Date, dtypes.Datetime)):
-        msg = "Converting to Date or Datetime dtype is not supported yet"
-        raise NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Date):
+        return spark_types.DateType()
+    if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        return spark_types.TimestampType()
```
FBruzzesi, line 107: I think it would already be good to differentiate between `TimestampType` and `TimestampNTZType`.

My understanding is that it would be quite safe to do:

```py
if isinstance_or_issubclass(dtype, dtypes.Datetime):
    dt_time_zone = getattr(dtype, "time_zone", None)
    if dt_time_zone is None:
        return spark_types.TimestampNTZType()
```
while if the timezone is specified it might not be that simple.

Happy to hear your opinion and expertise on this

tests/conftest.py
```diff
@@ -247,10 +246,7 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
             constructors_ids.append(constructor)
         elif constructor in LAZY_CONSTRUCTORS:
             if constructor == "pyspark":
-                if sys.version_info < (3, 12):  # pragma: no cover
-                    constructors.append(pyspark_lazy_constructor())
-                else:  # pragma: no cover
-                    continue
```
FBruzzesi, line 253: You might need to address the yaml in `.github/workflows/pytest.yml` for the pyspark installation.
Our understanding was that pyspark was not available or fully supported for py 3.12 and 3.13

FBruzzesi (commented):
Hey @osoucy, thanks for your PR - I am very happy to see you here 🙌🏼

I left a few comments, please let us know what's your opinion on those 😇

osoucy:
> Hey @osoucy, thanks for your PR - I am very happy to see you here 🙌🏼
> 
> I left a few comments, please let us know what's your opinion on those 😇

Thanks!

It's my first, but definitely not my last. I really want to use narwhals as the backbone for Laktory, but I really need to improve the current Spark support.

tests/conftest.py
EdAbati: FYI pyspark 3.5 _may_ be compatible with py 3.12 and 3.13 in some system/configuration but those are not officially tested/supported: [source](https://issues.apache.org/jira/browse/SPARK-44120?focusedCommentId=17873589&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17873589)

I had some problems with running our tests on windows for example 😕


and thank you @osoucy for this PR :)

osoucy: Indeed! Python 3.12 seems to works fine on most setups but is not officially supported by pyspark yet. I will revert the changes.

narwhals/_spark_like/utils.py
osoucy: Good point!

Spark always **stores** timestamps as UTC and as such does not store the input timezone. 

Spark **displays** timestamps with the timezone specified in the spark configuration: `spark.conf.set("spark.sql.session.timeZone", "UTC")`

Given this behavior, I think it would be safe to use your suggested implementation. I don't think we would have to worry about the `time_zone` value of `dtype` as this concept in spark does not really exists. It would be converted to UTC anyway.

MarcoGorelli: thanks @osoucy !

> I don't think we would have to worry about the time_zone value of dtype as this concept in spark does not really exists

Well this keeps things simple, nice! Should we be raising then if someone does something like `df.select(nw.col('a').cast(nw.Datetime('us', 'Europe/Paris'))`?

MarcoGorelli:
thanks a tonne for your PR! excellent stuff - will do a final check tomorrow, this should be included in Monday's release

narwhals/_spark_like/utils.py
osoucy: I would assume this would have to be implemented in overriding the `cast` method in _spark_like/expr.py check if its a datetime with a timezone?

MarcoGorelli:
closing then, addressed in https://github.com/narwhals-dev/narwhals/pull/1913 - in the future if you allow edits by maintainers then I can push minor fixups directly and then merge. thanks again for your contribution!

osoucy:
> thanks @osoucy , awesome stuff
> 
> somehow I couldn't push to your branch - did you untick "allow edits by maintainers" by any chance? I pushed to a new branch, but made sure to keep your commits in there #1913

I don't recall unticking or ticking anything when opening the PR. I'll pay closer attention next time.

Anyway, I'm happy with this first contribution!

Issue #1889, opened 2025-01-28
enh: Add `backend` argument to `lazy`


MarcoGorelli:
Similar to https://github.com/narwhals-dev/narwhals/issues/1888

I think `lazy` could have a `backend: ModuleType | Implementation | LazyBackend` argument, so all of the following could be supported:
```python
df: nw.DataFrame
df.lazy('duckdb')
df.lazy(duckdb)
df.lazy(nw.Implementation.DUCKDB)
```

`LazyBackend` could be `Literal['polars', 'dask', 'duckdb', 'pyspark']`. There may be some combinations of eager to lazy backends on which we'd have to raise, but in most cases we can probably manage by going via PyArrow

Ultimately, I think
```python
nw.from_dict(data, backend='pyarrow').lazy('duckdb')
```
would be pretty ergonomic and useful

raisadz:
I would like to take this issue if no one else has started working on it yet

PR #1895, opened 2025-01-29
feat: Add `backend` argument to `lazy`

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1889 
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [ ] Documented the changes


narwhals/dataframe.py
```diff
@@ -498,11 +498,13 @@ def __arrow_c_stream__(self: Self, requested_schema: object | None = None) -> ob
...
         return pa_table.__arrow_c_stream__(requested_schema=requested_schema)
 
-    def lazy(self: Self) -> LazyFrame[Any]:
+    def lazy(self: Self, *, backend: Implementation | None = None) -> LazyFrame[Any]:
         """Lazify the DataFrame (if possible).
 
         If a library does not support lazy execution, then this is a no-op.
 
+        If `backend` is specified some conversions between libraries can occur.
+
```
FBruzzesi, line 507: What about something along the following lines:

```suggestion
        If `backend` is specified, then a conversion between different backends
        might be triggered.
        If a library does not support lazy execution and `backend` is not specified,
        then this is a no-op.
        
        Arguments:
            backend: specifies which Implementation to convert to.
```

Feel free to expand 😇 (also I thought we have a linter to check for documentation of Arguments 🤔

narwhals/dataframe.py
```diff
@@ -545,7 +547,21 @@ def lazy(self: Self) -> LazyFrame[Any]:
             bar: [[6,7,8]]
             ham: [["a","b","c"]]
         """
-        return self._lazyframe(self._compliant_frame.lazy(), level="lazy")
+        supported_backends = (
+            Implementation.DASK,
+            Implementation.DUCKDB,
+            Implementation.POLARS,
+        )
```
FBruzzesi, line 554: RIP Pyspark 😂 Happy to add it later no worries, I am just kidding!

FBruzzesi (commented):
Hey @raisadz thanks a ton! this is a great start 🚀

I have some general concerns on the API/design itself, for pandas and pyarrow specifically, and a suggestion for the docstring - all the comments in the code

narwhals/_pandas_like/dataframe.py
```diff
@@ -709,8 +709,44 @@ def unique(
         )
 
     # --- lazy-only ---
-    def lazy(self: Self) -> Self:
-        return self
+    def lazy(self: Self, *, backend: Implementation | None = None) -> CompliantLazyFrame:
+        from narwhals.utils import parse_version
+
+        if backend is None:
+            return self
```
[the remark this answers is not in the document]
MarcoGorelli: not sure about this - `PandasLikeDataFrame` is a `CompliantLazyFrame` as well as `CompliantDataFrame`, I'd suggest we keep it simple and just keep the current `-> CompliantLazyFrame`

narwhals/_duckdb/dataframe.py
```diff
@@ -142,7 +142,7 @@ def drop(self: Self, columns: list[str], strict: bool) -> Self:  # noqa: FBT001
         selection = (col for col in self.columns if col not in columns_to_drop)
         return self._from_native_frame(self._native_frame.select(*selection))
 
-    def lazy(self: Self) -> Self:
+    def lazy(self: Self, *, backend: Implementation | None = None) -> Self:
         return self
```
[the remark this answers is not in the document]
raisadz: Thank you for the review, @FBruzzesi!

I added the `backend` argument to be able to run the following code:
```
nw_v1.from_native(duckdb_rel).lazy()
```
which otherwise made the tests fail for `V1`. I also now added `ValueError` if `backend` is not `None` for DuckDB.

narwhals/dataframe.py
raisadz: Sure, I will try to look into adding PySpark in a follow-up PR

narwhals/_duckdb/dataframe.py
```diff
@@ -142,7 +142,14 @@ def drop(self: Self, columns: list[str], strict: bool) -> Self:  # noqa: FBT001
         selection = (col for col in self.columns if col not in columns_to_drop)
         return self._from_native_frame(self._native_frame.select(*selection))
 
-    def lazy(self: Self) -> Self:
+    def lazy(self: Self, *, backend: Implementation | None = None) -> Self:
+        # The `backend`` argument has no effect but we keep it here for
+        # backwards compatibility because in `narwhals.stable.v1`
+        # function `.from_native()` will return a DataFrame for DuckDB.
+
+        if backend is not None:  # pragma: no cover
```
FBruzzesi, line 150: I think duckdb should be allowed?

```suggestion
        if backend not in (None, Implementation.DUCKDB):  # pragma: no cover
```

FBruzzesi (commented):
Thanks @raisadz , I left a tiny suggestion for DuckDB only.

In the meanwhile, I adjusted `.collect` to allow backend to take the following types: `str | ModuleType | Implementation | None` - see #1734

We might want to allow the same here? It can be a follow up PR anyway, but I just wanted to make you aware of it.

narwhals/_duckdb/dataframe.py
MarcoGorelli: i don't think it needs to be - if this path only exists for v1 backwards-compatibility, then https://github.com/narwhals-dev/narwhals/pull/1895#discussion_r1936126615 is something somebody could have written, but they wouldn't have passed a backend, so raising for `backend is not None` wouldn't break anyone's code

MarcoGorelli (approved):
thanks @raisadz ! and @FBruzzesi for reviewing

Issue #1726, opened 2025-01-05
test: allow for tests to be run without Polars / PyArrow/ pandas installed


MarcoGorelli:
It would be nice to be able to do say `pytest --constructors=pandas` and not need Polars installed

To do this, we should replace `import polars` with `polars = pytest.importorskip('polars')` only in the places where it's truly necessary to import Polars

MarcoGorelli:
Hi @mgorny !

How would you be looking to run the test suite in Gentoo? If you were able to run the full test suite for say pandas and duckdb, would that be enough? If so, I think it would be quite easy for us to allow
```python
pytest --constructors=pandas,duckdb
```
without needing polars installed

FBruzzesi:
Hey @mgorny, thanks for the feedback! These are very useful to understand what to prioritize!
I think #1843 was close to achieving it. I will spend more time into it or feel free to pick it up

MarcoGorelli:
I think just using `polars = pytest.importorskip('polars')` in the (relatively few) tests which specifically use Polars should already be us pretty close, and would be a relatively small lift?

mgorny:
Ideally, we'd want to skip `duckdb` too, since we don't have it packaged either. `pandas` and `pyarrow` are fine for us. And yes, passing `--constructors=` would be good.

MarcoGorelli:
Cool, I made a start on this in the `importorskip-polars` branch: https://github.com/narwhals-dev/narwhals/tree/importorskip-polars

Seems to work fine:
```console
$ python -c 'import polars'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'polars'
$ pytest tests/group_by_test.py --constructors=pandas
================================================= test session starts =================================================
platform linux -- Python 3.12.6, pytest-8.3.4, pluggy-1.5.0
Using --randomly-seed=176162433
rootdir: /home/marcogorelli/polars-api-compat-dev
configfile: pyproject.toml
plugins: hypothesis-6.123.11, cov-6.0.0, env-1.1.5, randomly-3.16.0
collected 33 items                                                                                                    

tests/group_by_test.py ....................s............                                                        [100%]

============================================ 32 passed
```

There's almost 30 test files left where the `polars` import needs moving inside the test

[...]

PR #1896, opened 2025-01-30
feat: allow for tests to be run without Polars installed, unify exceptions more

mgorny:
CC @MarcoGorelli 

Sending early to get feedback on the approaches used.

I went for:

1. `pytest.importorskip()` if the whole test (or module) seemed to rely on `polars`.
2. Explicit `try/except ImportError:` for tests that partially rely on `polars` — figured it would be weird to have them reported as "skipped" when the other part passed.
3. An explicit tuple variable for expected exceptions for `pytest.raises()`, with appending the `polars` exception when `polars` constructor is used.

Issue #1726

### DIFF
--- a/tests/expr_and_series/dt/ordinal_day_test.py
+++ b/tests/expr_and_series/dt/ordinal_day_test.py
@@ -38,5 +38,7 @@ def test_ordinal_day(dates: datetime) -> None:
     except ImportError:
         pass
     else:
-        result_pl = nw.from_native(pl.Series([dates]), series_only=True).dt.ordinal_day()[0]
+        result_pl = nw.from_native(pl.Series([dates]), series_only=True).dt.ordinal_day()[
+            0
+        ]
         assert result_pl == result_pd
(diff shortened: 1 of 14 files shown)


MarcoGorelli:
thanks @mgorny ! your approach looks good, will take a closer look at the weekend

MarcoGorelli:
ah, i tried unifying exceptions whilst we're here, but that got a bit out of hand - will revert and keep this to the original scope

mgorny:
I'm sorry for not taking it further — it's been a busy week.

MarcoGorelli:
no worries at all! will pick this up again for the next release

mgorny:
@MarcoGorelli, I've just taken a look and it seems that a fair number of merge conflicts have emerged against `main`. Do you happen to have a suggestion how to best proceed here? It's possible that starting over might be the easiest approach. Perhaps we could sync a bit and try to get partial changes in next time, to avoid things getting too much out of sync?

MarcoGorelli:
hey - yes sorry about that

sure, definitely! small PRs are always faster to merge, if we can break this into small steps and just merge those then we can get this in

the "unify exceptions" part has been done separately as well so no need to bother about that 👍

mgorny:
Closing in favor of #2241.

PR #1908, opened 2025-02-01
fix: pandas was raising when index name and column names overlapped in groupby

### LINKED ISSUE #1907: [Bug]: `ValueError: 'Type' is both an index level and a column label` during group-by
### Describe the bug

If a column name overlaps with the index name, then a group-by might raise, as reported here: https://github.com/plotly/plotly.py/issues/5005

```
In [2]: import pandas as pd

In [3]: df = pd.DataFrame({'a': [1,1,2], 'b': [4,5,6]}, index=pd.Index([0,1,2], name='a'))

In [4]: import narwhals as nw

In [5]: nw.from_native(df).group_by('a').agg(nw.col('b').mean())
```
### Expected results
```
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|       a    b     |
|    0  1  4.5     |
|    1  2  6.0     |
└──────────────────┘
```
### Actual results

ValueError: 'a' is both an index level and a column label, which is ambiguous.

```shell
System:
    python: 3.11.10 (main, Sep  9 2024, 22:11:19) [Clang 18.1.8 ]
executable: /home/marcogorelli/plotly-dev/.venv/bin/python
   machine: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 1.24.1
       pandas: 1.5.3
       polars: 1.8.2
         cudf: 
        modin: 
      pyarrow: 17.0.0
        numpy: 1.23.2
```

### Relevant log output

```shell

```

MarcoGorelli:
closes #1907 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 [![PyPI version](https://badge.fury.io/py/narwhals.svg)](https://badge.fury.io/py/narwhals)
 [![Downloads](https://static.pepy.tech/badge/narwhals/month)](https://pepy.tech/project/narwhals)
 [![Trusted publishing](https://img.shields.io/badge/Trusted_publishing-Provides_attestations-bright_green)](https://peps.python.org/pep-0740/)
+[![PYPI - Types](https://img.shields.io/pypi/types/narwhals)](https://pypi.org/project/narwhals)
 
 Extremely lightweight and extensible compatibility layer between dataframe libraries!
 
--- a/narwhals/_pandas_like/group_by.py
+++ b/narwhals/_pandas_like/group_by.py
@@ -18,7 +18,6 @@ from narwhals.utils import Implementation
 from narwhals.utils import find_stacklevel
 
 if TYPE_CHECKING:
-    import pandas as pd
     from typing_extensions import Self
 
     from narwhals._pandas_like.dataframe import PandasLikeDataFrame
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -10,6 +10,7 @@
 [![PyPI version](https://badge.fury.io/py/narwhals.svg)](https://badge.fury.io/py/narwhals)
 [![Downloads](https://static.pepy.tech/badge/narwhals/month)](https://pepy.tech/project/narwhals)
 [![Trusted publishing](https://img.shields.io/badge/Trusted_publishing-Provides_attestations-bright_green)](https://peps.python.org/pep-0740/)
+[![PYPI - Types](https://img.shields.io/pypi/types/narwhals)](https://pypi.org/project/narwhals)
```
MarcoGorelli, line 13: drive-by 🙈

narwhals/_pandas_like/group_by.py
```diff
@@ -37,46 +38,54 @@
...
+) -> pd.DataFrameGroupBy:
+    min_version = (1, 1)
+    if (
+        df._implementation is Implementation.PANDAS and df._backend_version < min_version
+    ):  # pragma: no cover
+        if (
+            not drop_null_keys
+            and df.simple_select(*keys)._native_frame.isna().any().any()
+        ):
+            msg = "Grouping by null values is not supported in pandas < 1.0.0"
```
FBruzzesi, line 52: Should this match `min_version` above?

narwhals/_pandas_like/group_by.py
```diff
@@ -37,46 +38,54 @@
...
+    df: PandasLikeDataFrame, keys: list[str], *, drop_null_keys: bool, reset_index: bool
+) -> pd.DataFrameGroupBy:
+    min_version = (1, 1)
+    if (
+        df._implementation is Implementation.PANDAS and df._backend_version < min_version
+    ):  # pragma: no cover
+        if (
+            not drop_null_keys
+            and df.simple_select(*keys)._native_frame.isna().any().any()
+        ):
```
FBruzzesi, line 51: Unrelated from the PR itself but in a way I am not the biggest fan of this behaviour - this would raise only if there are actually null keys  🤔

FBruzzesi (commented):
I am a bit confused by the fact that we `reset_index` in both cases anyway now 🤔 if that' the correct behaviour do we need a special function for it or we can keep the logic in the `__init__`?

narwhals/_pandas_like/group_by.py
MarcoGorelli: True, it is value-dependent, but it only affects old pandas versions, and null values in the column you're grouping by are hopefully not-too-common

narwhals/_pandas_like/group_by.py
MarcoGorelli: yup, thanks!

narwhals/_pandas_like/group_by.py
```diff
@@ -43,32 +43,27 @@ def __init__(
...
-                and select_columns_by_name(
-                    self._df._native_frame,
-                    self._keys,
-                    self._df._backend_version,
-                    self._df._implementation,
-                )
-                .isna()
-                .any()
-                .any()
+                and self._df.simple_select(*self._keys)._native_frame.isna().any().any()
```
MarcoGorelli, line 55: same, just simpler

narwhals/typing.py
```diff
@@ -60,19 +60,14 @@ def simple_select(
...
+    def __narwhals_lazyframe__(self) -> CompliantLazyFrame: ...
     def __narwhals_namespace__(self) -> Any: ...
     def simple_select(
         self, *column_names: str
     ) -> CompliantLazyFrame: ...  # `select` where all args are column names
 
 
-class CompliantLazyFrame(Protocol):
-    def __narwhals_lazyframe__(self) -> CompliantLazyFrame: ...
-    def __narwhals_namespace__(self) -> Any: ...
```
MarcoGorelli, line 73: another drive-by

MarcoGorelli:
I manually checked, and https://github.com/plotly/plotly.py/issues/4917 is still behaving correctly (yes, there is a test for it, but still, testing plots is really hard, so I checked manually too)

FBruzzesi (approved):
Thanks @MarcoGorelli ! Shall we make a patch release then?
 
NarwhalChubbiwhalGIF

MarcoGorelli:
thanks! sure thing, let's give plotly users a good experience

PR #1914, opened 2025-02-02
feat: enable `ModuleType` and `str` for backend in `.lazy()` method

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1889

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

PR #1917, opened 2025-02-02
chore: `IntoImplementation` type alias

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- See [comment](https://github.com/narwhals-dev/narwhals/pull/1914#issuecomment-2629460801)

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -82,7 +82,7 @@ class DuckDBLazyFrame(CompliantLazyFrame):
 
     def collect(
         self: Self,
-        backend: ModuleType | Implementation | str | None,
+        backend: Implementation | None,
         **kwargs: Any,
     ) -> CompliantDataFrame:
         if backend is None or backend is Implementation.PYARROW:
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -114,7 +114,7 @@ class SparkLikeLazyFrame(CompliantLazyFrame):
 
     def collect(
         self: Self,
-        backend: ModuleType | Implementation | str | None,
+        backend: Implementation | None,
         **kwargs: Any,
     ) -> CompliantDataFrame:
         if backend is Implementation.PANDAS:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
     from narwhals.typing import IntoDataFrame
     from narwhals.typing import IntoExpr
     from narwhals.typing import IntoFrame
+    from narwhals.typing import IntoImplementation
     from narwhals.typing import SizeUnit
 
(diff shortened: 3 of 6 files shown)


MarcoGorelli:
thanks for doing this

not totally sure about this one, some thoughts:
- the docs now show `IntoImplementation`, rather than a more explicit union
- the argument is `backend` but the type hint is `IntoImplementation`
- currently, we convert everything to `nw.Implementation`, but in theory I think we should be able to accept native namespaces which are from extensions

FBruzzesi:
> thanks for doing this
> 
> not totally sure about this one, some thoughts:
> 
> * the docs now show `IntoImplementation`, rather than a more explicit union
> * the argument is `backend` but the type hint is `IntoImplementation`

Makes sense! I am ok with leaving the signature as it is with the union

> * currently, we convert everything to `nw.Implementation`, but in theory I think we should be able to accept native namespaces which are from extensions

This would require to rethink the codebase, not just the signature 🙈

MarcoGorelli:
I thought it would, but when I built this locally then it didn't show up. Though for some reason `Implementation` already doesn't show up https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.lazy 🤔

dangotbanned:
> I thought it would, but when I built this locally then it didn't show up. Though for some reason `Implementation` already doesn't show up [narwhals-dev.github.io/narwhals/api-reference/dataframe#narwhals.dataframe.DataFrame.lazy](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.lazy) 🤔

Oh, is that because there isn't a page for `Implementation`?
https://narwhals-dev.github.io/narwhals/api-reference/


I went looking for one when I wrote this but settled on linking to source:
https://github.com/vega/altair/blob/94220be0115e8b13d2ebc686552edf68fd841a54/altair/datasets/_reader.py#L137-L143

MarcoGorelli:
> Oh, is that because there isn't a page for Implementation?

right, thanks 🤦 

maybe we should make a type alias for `Backend`, rather than `IntoImplementation`?

dangotbanned:
FYI, if we wanna expose it in the future - we now have `IntoBackend`

https://github.com/narwhals-dev/narwhals/blob/62ba1579012584228d85d965887035ae03f5225a/narwhals/_namespace.py#L83

Currently not a runtime object, but it is this idea + **auto-completion**

https://github.com/narwhals-dev/narwhals/blob/62ba1579012584228d85d965887035ae03f5225a/narwhals/_namespace.py#L56-L83

FBruzzesi:
Closing in favor of #2971

Issue #1373, opened 2024-11-13
RFC


adrinjalali:
I was wondering, if exceptions should be unified. I came across this since I was writing a test with this code:

```py
def f(X, cols):
    return nw.from_native(X, eager_only=True).select(cols)
```

with this kinda test:

```py
import pytest
import narwhals as nw

@pytest.mark.parametrize("lib", ["pandas", "polars"])
def test_select_cols(lib):
    lib = pytest.importorskip(lib)

    df = nw.from_native(lib.DataFrame({"a": [1, 2], "b": [10, 20], "c": ["x", "y"]}))
    # In reality it's either a KeyError or a pl.exceptions.ColumnNotFoundError
    with pytest.raises(Exception):
        f(df, ["a", "b", "d"])
```

And I would have expected the same kind of exception here (TBH I would have expected a `KeyError` from `polars`, but I get a `ColumnNotFoundError` which doesn't even inherit from `KeyError`).

Not sure what to think here.

MarcoGorelli:
Thanks for the request! agree, this would be good to aim for

Not sure how feasible it is to do it everywhere, but at least for common kinds of errors - and where it's possible - I agree that this is something worth aiming for

MarcoGorelli:
In this case, this is now addressed, the following passes:
```python
import pytest
import narwhals as nw

def f(X, cols):
    return nw.from_native(X, eager_only=True).select(cols)

@pytest.mark.parametrize("lib", ["pandas", "polars"])
def test_select_cols(lib):
    lib = pytest.importorskip(lib)

    df = nw.from_native(lib.DataFrame({"a": [1, 2], "b": [10, 20], "c": ["x", "y"]}))
    # In reality it's either a KeyError or a pl.exceptions.ColumnNotFoundError
    with pytest.raises(nw.exceptions.ColumnNotFoundError):
        f(df, ["a", "b", "d"])
```
We'll work on unifying and documenting these more

adrinjalali:
legend!

MarcoGorelli:
i think this is mostly addressed, at least for the most common ops, if there's anything outstanding feel free to open a new issue 🙏

PR #1918, opened 2025-02-02
feat: catch Polars exceptions, unify exception raising more

MarcoGorelli:
Unify exceptions, and re-raise Polars ones as Narwhals ones

This doesn't unify _everything_...but I think it's a nice step forwards, and should be ~~totally backwards-compatible~~ mostly backwards-compatible (unless anyone catches Polars exceptions specifically in tests)

Towards #1373

### DIFF
--- a/narwhals/_polars/dataframe.py
+++ b/narwhals/_polars/dataframe.py
@@ -125,7 +125,7 @@ class PolarsDataFrame:
                 msg = f"{e!s}\n\nHint: Did you mean one of these columns: {self.columns}?"
                 raise ColumnNotFoundError(msg) from e
             except Exception as e:  # noqa: BLE001
-                raise catch_polars_exception(e) from None
+                raise catch_polars_exception(e, self._backend_version) from None
 
         return func
 
@@ -368,7 +368,7 @@ class PolarsDataFrame:
                 separator=separator,
             )
         except Exception as e:  # noqa: BLE001
-            raise catch_polars_exception(e) from None
+            raise catch_polars_exception(e, self._backend_version) from None
         return self._from_native_object(result)
 
     def to_polars(self: Self) -> pl.DataFrame:
@@ -453,7 +453,7 @@ class PolarsLazyFrame:
             try:
                 collected_schema = self._native_frame.collect_schema()
             except Exception as e:  # noqa: BLE001
-                raise catch_polars_exception(e) from None
+                raise catch_polars_exception(e, self._backend_version) from None
             return {
                 name: native_to_narwhals_dtype(
                     dtype, self._version, self._backend_version
@@ -469,7 +469,7 @@ class PolarsLazyFrame:
         try:
             result = self._native_frame.collect(**kwargs)
         except Exception as e:  # noqa: BLE001
-            raise catch_polars_exception(e) from None
+            raise catch_polars_exception(e, self._backend_version) from None
 
         if backend is None or backend is Implementation.POLARS:
             from narwhals._polars.dataframe import PolarsDataFrame
--- a/narwhals/_polars/dataframe.py
+++ b/narwhals/_polars/dataframe.py
@@ -139,10 +139,7 @@ class PolarsDataFrame:
                 for name, dtype in self._native_frame.schema.items()
             }
         else:
-            try:
-                collected_schema = self._native_frame.collect_schema()
-            except Exception as e:  # noqa: BLE001
-                raise catch_polars_exception(e) from None
+            collected_schema = self._native_frame.collect_schema()
             return {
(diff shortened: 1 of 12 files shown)


MarcoGorelli:
The only tests failures are from Marimo, where they're checking the Polars error very tightly in their tests
```
        with pytest.raises(pl.exceptions.ComputeError):
            # Polars doesn't support writing nested lists to csv
>           manager.to_csv()

      with pytest.raises(pl.exceptions.ComputeError):
            # Polars doesn't support writing nested lists to csv
>           complex_data.to_csv()
```

MarcoGorelli:
Hey @mscolnick , I just wanted to run this past you

There's been requests to unify the exceptions from Narwhals. We do that somewhat already, but this PR makes another step towards that.

As far as I can't, there'd be no user-facing impact, it would just be matter of updating the expected exception in 2 tests. I'd be happy to make a PR if welcome

Would you be OK with this change? Else no worries, we could always just preserve the current exceptions in the v1 API, I just figured that for something so small it might not be worth it

mscolnick:
Yea that is fine with us and would appreciate the help with the change, thank you!

dangotbanned:
@MarcoGorelli I'd be careful with how this impacts performance, since we only get zero-cost exceptions on >=3.11

Somewhat related:
- https://docs.astral.sh/ruff/rules/try-except-in-loop/
- https://bugs.python.org/issue40222
- https://docs.python.org/3/whatsnew/3.11.html#misc

MarcoGorelli:
Thanks Dan! I'll do some checks before shipping it - I think it shouldn't be an issue because it's only on Dataframe methods, e.g. for
```
df.filter(super_complicated_expression_1, super_complicated_expression_2, super_complicated_expression_3)
```
the try-except would only happen when `filter` is called, and not for each expression

Nonetheless, I would like to time it and be careful with it, so this'll stay of today's release, we can revisit for the next one when we've done more careful checks

MarcoGorelli:
I took a closer look, and we can implement some infallible Polars methods directly outside of the `__getattr__` block, so that they don't go in the `try-except`:
- `head`
- `tail`
- `__len__`

the remaining methods in the try-except are dataframe methods (such as `filter`, `with_columns`, `select`, `sort`) which are fallible and often have their own exception-handling in Polars - I think I'm fine with that and that the benefits of unified exceptions outweigh the `try-except` overhead, it's not like anyone should be running `df.filter` in a tight loop anyway

@dangotbanned just checking if you're happy with this

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/1918#issuecomment-2646573384

All sounds good to me @MarcoGorelli

narwhals/_polars/dataframe.py
```diff
@@ -114,6 +124,8 @@ def func(*args: Any, **kwargs: Any) -> Any:
             except pl.exceptions.ColumnNotFoundError as e:  # pragma: no cover
                 msg = f"{e!s}\n\nHint: Did you mean one of these columns: {self.columns}?"
                 raise ColumnNotFoundError(msg) from e
+            except Exception as e:  # noqa: BLE001
```
dangotbanned, line 127: Could these be narrowed any further than `Exception`?
https://docs.astral.sh/ruff/rules/blind-except/

narwhals/_polars/utils.py
```diff
@@ -217,3 +221,25 @@ def convert_str_slice_to_int_slice(
...
+    if hasattr(pl.exceptions, "PolarsError") and isinstance(
+        exception, pl.exceptions.PolarsError
+    ):
+        # Old versions of Polars didn't have PolarsError.
+        return NarwhalsError(str(exception))
+    elif "polars.exceptions" in str(type(exception)):  # pragma: no cover
+        # Last attempt, for old Polars versions.
+        return NarwhalsError(str(exception))
+    # Just return exception as-is.
+    return exception
```
dangotbanned, line 245: Could you do some kind of conditional block gated on the `polars` version, to determine what goes in:

```py
try:
    ...
except: (here, ... ) as e:
    ...
```

I'm thinking that check could be evaluated *once*, rather than doing this after already checking everything here every call?

Something like https://github.com/vega/altair/blob/49d89551275efb1a0634026af435ec3fa1d4925f/altair/utils/_vegafusion_data.py#L52-L62, but to generate the exception target

narwhals/_polars/dataframe.py
MarcoGorelli: I think this shouldn't be an issue because we re-raise the exception anyway

MarcoGorelli:
thanks for your review!

PR #1924, opened 2025-02-03
feat: add `Schema.to_(arrow|pandas|polars)`

### LINKED ISSUE #1912: [Enh]: `nw.(DType|Schema)` conversion API

We've already adopted `narwhals` in https://github.com/vega/altair
This feature would be helpful for https://github.com/vega/altair/pull/3631

In (https://github.com/vega/altair/pull/3631), I've used serialized dataset schemas to improve consistency between `polars`, `pandas` and `pyarrow` when reading from file.

A challenge I've had is how to incorporate these data types:
- what functions accept a `nw.Schema`?
- when do I need to fall back to native types?

Maybe this is a niche problem, but I'd really appreciate some public API to utilize all the `narwhals` -> *native* type conversion logic.

The `nw.Schema` -> *native* logic for this is already specified in `nw.functions._from_dict_impl`.

<details><summary><code>nw.functions._from_dict_impl</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/functions.py#L444-L546

</p>
</details> 

Additionally, each of these functions for the individual types:

<details><summary><b>6x</b> <code>nw._(.*).utils.narwhals_to_native_dtype</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_arrow/utils.py#L86

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_duckdb/utils.py#L142

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_dask/utils.py#L109

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_pandas_like/utils.py#L518

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_polars/utils.py#L150

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/_spark_like/utils.py#L82

</p>
</details> 

#### Solution 1
Add method(s) on `DType`

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/dtypes.py#L27

#### Solution 2 (Preferred)
Add method(s) on `Schema`

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/schema.py#L27

This doesn't rule out *Solution 1*, but I think it could be the cleaner API if only one were chosen.

Something like this would be pretty ergonomic.
For my use case, I could just pass the `nw.Schema` around and only convert it when needed:

```py
from typing import Any, TYPE_CHECKING

if TYPE_CHECKING:
    from types import ModuleType
    from typing import TypeAlias

    import pandas as pa
    import polars as pl

WhateverPandasIs: TypeAlias = Any

class Schema:
    def to_native(self, native_namespace: ModuleType) -> Any: ...
    def to_polars(self) -> pl.Schema: ...
    def to_arrow(self) -> pa.Schema: ...
    def to_pandas(self) -> dict[str, WhateverPandasIs]: ...
```

This is a short version of what I'm doing currently in (https://github.com/vega/altair/pull/3631#discussion_r1937953187).

It would be great to *not* rely on the `narwhals` internals for this though:

```py

class SchemaCache:
    def schema(self, name: str, /) -> dict[str, nw.dtypes.DType]: ...
    def schema_pyarrow(self, name: str, /):
        schema = self.schema(name)
        if schema:
            from narwhals._arrow.utils import narwhals_to_native_dtype

            m = {k: narwhals_to_native_dtype(v, Version.V1) for k, v in schema.items()}
        else:
            m = {}
        return nw.dependencies.get_pyarrow().schema(m)
```

<details><summary>API overview vega/altair#3631</summary>
<p>

```rst
Load example datasets *remotely* from `vega-datasets`_.

Provides **70+** datasets, used throughout our `Example Gallery`_.

You can learn more about each dataset at `datapackage.md`_.

Examples
--------
Load a dataset as a ``DataFrame``/``Table``::

    from altair.datasets import load

    load("cars")

.. note::
   Requires installation of either `polars`_, `pandas`_, or `pyarrow`_.

Get the remote address of a dataset and use directly in a :class:`altair.Chart`::

    import altair as alt
    from altair.datasets import url

    source = url("co2-concentration")
    alt.Chart(source).mark_line(tooltip=True).encode(x="Date:T", y="CO2:Q")

.. note::
   Works without any additional dependencies.

For greater control over the backend library use::

    from altair.datasets import Loader

    load = Loader.from_backend("polars")
    load("penguins")
    load.url("penguins")

This method also provides *precise* <kbd>Tab</kbd> completions on the returned object::

    load("cars").<Tab>
    #            bottom_k
    #            drop
    #            drop_in_place
    #            drop_nans
    #            dtypes
    #            ...

.. _vega-datasets:
    https://github.com/vega/vega-datasets
.. _Example Gallery:
    https://altair-viz.github.io/gallery/index.html#example-gallery
.. _datapackage.md:
    https://github.com/vega/vega-datasets/blob/main/datapackage.md
.. _polars:
    https://docs.pola.rs/user-guide/installation/
.. _pandas:
    https://pandas.pydata.org/docs/getting_started/install.html
.. _pyarrow:
    https://arrow.apache.org/docs/python/install.html
```

</p>
</details>

- [Some of the more creative ways I've approached IO](https://github.com/vega/altair/blob/7bb6f9e7a8c98152f17a2ec9c57141cc58171148/altair/datasets/_readimpl.py#L328-L443)
- [An example of schema information available](https://github.com/vega/vega-datasets/blob/7a54cc66d1fa30b83b565f9c1de99508badd6c3e/datapackage.json#L2033-L2100)
  - [How that translates to `narwhals`](https://github.com/vega/altair/blob/869d2161bde45d59582c687e574fbfe0f7efe776/altair/datasets/_cache.py#L65-L92)

I'd be happy to help with a PR to implement this if anyone else can see the potential value.

I've really been enjoying using `narwhals` and it has played a **central role** in this pretty long-running PR.
Big thank you to anyone who has contributed 🙏

dangotbanned:
Will close #1912

- [x] ✨ Feature

## Related issues

- Closes #1912

## Checklist

- [x] Tests added
- [x] Documented the changes

- Started with porting `nw.functions._from_dict_impl`
- Extended `Schema` w/ `._version`

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -343,10 +343,11 @@ def _new_series_impl(
             )
 
             backend_version = parse_version(native_namespace.__version__)
-            dtype = pandas_like_narwhals_to_native_dtype(
+            pd_dtype = pandas_like_narwhals_to_native_dtype(
                 dtype, None, implementation, backend_version, version
             )
-        native_series = native_namespace.Series(values, name=name, dtype=dtype)
+            native_series = native_namespace.Series(values, name=name, dtype=pd_dtype)
+        native_series = native_namespace.Series(values, name=name)
 
     elif implementation is Implementation.PYARROW:
         if dtype:
@@ -735,7 +736,7 @@ def _from_numpy_impl(
             )
 
             backend_version = parse_version(native_namespace.__version__)
-            schema = {
+            pd_schema = {
                 name: pandas_like_narwhals_to_native_dtype(
                     dtype=schema[name],
                     dtype_backend=get_dtype_backend(native_type, implementation),
@@ -746,7 +747,7 @@ def _from_numpy_impl(
                 for name, native_type in schema.items()
             }
             native_frame = native_namespace.DataFrame(data, columns=schema.keys()).astype(
-                schema
+                pd_schema
             )
         elif isinstance(schema, list):
             native_frame = native_namespace.DataFrame(data, columns=schema)
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -520,10 +520,8 @@ def _from_dict_impl(
             from narwhals._pandas_like.utils import get_dtype_backend
 
             pd_schema = Schema(schema).to_pandas(
-                dtype_backend=(
-                    get_dtype_backend(native_type, eager_backend)
-                    for native_type in native_frame.dtypes
-                )
+                get_dtype_backend(native_type, eager_backend)
+                for native_type in native_frame.dtypes
             )
             native_frame = native_frame.astype(pd_schema)
 
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -22,6 +22,7 @@ from narwhals.dataframe import LazyFrame
 from narwhals.dependencies import is_numpy_array
 from narwhals.exceptions import ShapeError
 from narwhals.expr import Expr
+from narwhals.schema import Schema
 from narwhals.translate import from_native
 from narwhals.utils import Implementation
 from narwhals.utils import Version
@@ -43,7 +44,6 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from narwhals.dtypes import DType
-    from narwhals.schema import Schema
     from narwhals.series import Series
     from narwhals.typing import IntoDataFrameT
     from narwhals.typing import IntoExpr
@@ -495,18 +495,7 @@ def _from_dict_impl(  # noqa: PLR0915
         msg = f"Unsupported `backend` value.\nExpected one of {supported_eager_backends} or None, got: {eager_backend}."
         raise ValueError(msg)
     if eager_backend is Implementation.POLARS:
-        if schema:
-            from narwhals._polars.utils import (
-                narwhals_to_native_dtype as polars_narwhals_to_native_dtype,
-            )
-
-            schema_pl = {
-                name: polars_narwhals_to_native_dtype(dtype, version=version)
-                for name, dtype in schema.items()
-            }
-        else:
-            schema_pl = None
-
+        schema_pl = Schema(schema).to_polars() if schema else None
         native_frame = native_namespace.from_dict(data, schema=schema_pl)
     elif eager_backend in (
         Implementation.PANDAS,
@@ -552,18 +541,8 @@ def _from_dict_impl(  # noqa: PLR0915
             native_frame = native_frame.astype(schema)
 
     elif eager_backend is Implementation.PYARROW:
-        if schema:
-            from narwhals._arrow.utils import (
-                narwhals_to_native_dtype as arrow_narwhals_to_native_dtype,
-            )
-
-            schema = native_namespace.schema(
-                [
-                    (name, arrow_narwhals_to_native_dtype(dtype, version))
-                    for name, dtype in schema.items()
-                ]
-            )
-        native_frame = native_namespace.table(data, schema=schema)
+        pa_schema = Schema(schema).to_arrow() if schema is not None else schema
+        native_frame = native_namespace.table(data, schema=pa_schema)
     else:  # pragma: no cover
         try:
             # implementation is UNKNOWN, Narwhals extension using this feature should
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -449,12 +449,7 @@ def from_dict(
     backend = validate_native_namespace_and_backend(
         backend, native_namespace, emit_deprecation_warning=True
     )
-    return _from_dict_impl(
-        data,
-        schema,
-        backend=backend,
-        version=Version.MAIN,
-    )
+    return _from_dict_impl(data, schema, backend=backend)
 
 
 def _from_dict_impl(
@@ -462,7 +457,6 @@ def _from_dict_impl(
     schema: dict[str, DType] | Schema | None = None,
     *,
     backend: ModuleType | Implementation | str | None = None,
-    version: Version,  # noqa: ARG001
 ) -> DataFrame[Any]:
     from narwhals.series import Series
     from narwhals.translate import to_native
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -457,12 +457,12 @@ def from_dict(
     )
 
 
-def _from_dict_impl(  # noqa: PLR0915
+def _from_dict_impl(
     data: dict[str, Any],
     schema: dict[str, DType] | Schema | None = None,
     *,
     backend: ModuleType | Implementation | str | None = None,
-    version: Version,
+    version: Version,  # noqa: ARG001
 ) -> DataFrame[Any]:
     from narwhals.series import Series
     from narwhals.translate import to_native
@@ -523,22 +523,14 @@ def _from_dict_impl(  # noqa: PLR0915
 
         if schema:
             from narwhals._pandas_like.utils import get_dtype_backend
-            from narwhals._pandas_like.utils import (
-                narwhals_to_native_dtype as pandas_like_narwhals_to_native_dtype,
-            )
 
-            backend_version = parse_version(native_namespace.__version__)
-            schema = {
-                name: pandas_like_narwhals_to_native_dtype(
-                    dtype=schema[name],
-                    dtype_backend=get_dtype_backend(native_type, eager_backend),
-                    implementation=eager_backend,
-                    backend_version=backend_version,
-                    version=version,
+            pd_schema = Schema(schema).to_pandas(
+                dtype_backend=(
+                    get_dtype_backend(native_type, eager_backend)
+                    for native_type in native_frame.dtypes
                 )
-                for name, native_type in native_frame.dtypes.items()
-            }
-            native_frame = native_frame.astype(schema)
+            )
+            native_frame = native_frame.astype(pd_schema)
 
     elif eager_backend is Implementation.PYARROW:
         pa_schema = Schema(schema).to_arrow() if schema is not None else schema
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -347,7 +347,8 @@ def _new_series_impl(
                 dtype, None, implementation, backend_version, version
             )
             native_series = native_namespace.Series(values, name=name, dtype=pd_dtype)
-        native_series = native_namespace.Series(values, name=name)
+        else:
+            native_series = native_namespace.Series(values, name=name)
 
     elif implementation is Implementation.PYARROW:
         if dtype:
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -99,6 +99,15 @@ class Schema(BaseSchema):
         return len(self)
 
     def to_arrow(self: Self) -> pa.Schema:
+        """Convert Schema to a pyarrow Schema.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> schema = nw.Schema({"a": nw.Int64(), "b": nw.Datetime("ns")})
+            >>> schema.to_arrow()
+            a: int64
+            b: timestamp[ns]
+        """
         import pyarrow as pa  # ignore-banned-import
 
         from narwhals._arrow.utils import narwhals_to_native_dtype
@@ -111,6 +120,24 @@ class Schema(BaseSchema):
     def to_pandas(
         self: Self, *, dtype_backend: str | Iterable[str] | None = None
     ) -> dict[str, Any]:
+        """Convert Schema to an ordered mapping of column names to their pandas data type.
+
+        Arguments:
+            dtype_backend: Backend(s) used for the native types. When providing more than
+                one, the length of the iterable must be equal to the length of the schema.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> schema = nw.Schema({"a": nw.Int64(), "b": nw.Datetime("ns")})
+            >>> schema.to_pandas()
+            {'a': 'int64', 'b': 'datetime64[ns]'}
+
+            >>> schema.to_pandas(dtype_backend="pyarrow-nullable")
+            {'a': 'Int64[pyarrow]', 'b': 'timestamp[ns][pyarrow]'}
+
+            >>> schema.to_pandas(dtype_backend=["pandas-nullable", "pyarrow-nullable"])
+            {'a': 'Int64', 'b': 'timestamp[ns][pyarrow]'}
+        """
         import pandas as pd  # ignore-banned-import
 
         from narwhals._pandas_like.utils import narwhals_to_native_dtype
@@ -155,6 +182,17 @@ class Schema(BaseSchema):
             }
 
     def to_polars(self: Self) -> pl.Schema | Any:
+        """Convert Schema to a polars Schema.
+
+        Returns:
+            A polars schema or plain dict (prior to polars 1.0).
+
+        Examples:
+            >>> import narwhals as nw
+            >>> schema = nw.Schema({"a": nw.Int64(), "b": nw.Datetime("ns")})
+            >>> schema.to_polars()  # doctest:+SKIP
+            Schema([('a', Int64), ('b', Datetime(time_unit='ns', time_zone=None))])
+        """
         import polars as pl  # ignore-banned-import
 
         from narwhals._polars.utils import narwhals_to_native_dtype
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -203,8 +203,12 @@ class Schema(BaseSchema):
 
         from narwhals._polars.utils import narwhals_to_native_dtype
 
-        it = (
+        schema = (
             (name, narwhals_to_native_dtype(dtype, self._version))
             for name, dtype in self.items()
         )
-        return pl.Schema(it) if parse_version(pl.__version__) >= (1, 0, 0) else dict(it)
+        return (
+            pl.Schema(schema)
+            if parse_version(pl.__version__) >= (1, 0, 0)
+            else dict(schema)
+        )
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -98,22 +98,6 @@ class Schema(BaseSchema):
         """
         return len(self)
 
-    def to_native(
-        self: Self,
-        *,
-        backend: ModuleType | Implementation | str,
-        dtype_backend: str | None = None,
-    ) -> dict[str, Any] | pl.Schema | pa.Schema:
-        implementation = Implementation.from_backend(backend)
-        if implementation is Implementation.POLARS:
-            return self.to_polars(backend=implementation)
-        elif implementation.is_pandas_like():
-            return self.to_pandas(backend=implementation, dtype_backend=dtype_backend)
-        elif implementation is Implementation.PYARROW:
-            return self.to_arrow(backend=implementation)
-
-        raise NotImplementedError
-
     def to_arrow(self: Self, *, backend: ModuleType | Implementation | str) -> pa.Schema:
         from narwhals._arrow.utils import narwhals_to_native_dtype
 
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -11,7 +11,16 @@ from typing import TYPE_CHECKING
 from typing import Iterable
 from typing import Mapping
 
+from narwhals.utils import Implementation
+from narwhals.utils import Version
+from narwhals.utils import parse_version
+
 if TYPE_CHECKING:
+    from types import ModuleType
+    from typing import Any
+
+    import polars as pl
+    import pyarrow as pa
     from typing_extensions import Self
 
     from narwhals.dtypes import DType
@@ -85,3 +94,65 @@ class Schema(BaseSchema):
             Number of columns.
         """
         return len(self)
+
+    def to_native(
+        self: Self, *, native_namespace: ModuleType, dtype_backend: str | None = None
+    ) -> dict[str, Any] | pl.Schema | pa.Schema:
+        implementation = Implementation.from_native_namespace(native_namespace)
+        version = Version.MAIN
+        if implementation is Implementation.POLARS:
+            return self.to_polars(backend=implementation, version=version)
+        elif implementation.is_pandas_like():
+            return self.to_pandas(
+                backend=implementation, version=version, dtype_backend=dtype_backend
+            )
+        elif implementation is Implementation.PYARROW:
+            return self.to_arrow(backend=implementation, version=version)
+
+        raise NotImplementedError
+
+    def to_arrow(
+        self: Self, *, backend: ModuleType | Implementation | str, version: Version
+    ) -> pa.Schema:
+        from narwhals._arrow.utils import narwhals_to_native_dtype
+
+        implementation = Implementation.from_backend(backend)
+        schema: pa.Schema = implementation.to_native_namespace().schema(
+            (name, narwhals_to_native_dtype(dtype, version))
+            for name, dtype in self.items()
+        )
+        return schema
+
+    def to_pandas(
+        self: Self,
+        *,
+        backend: ModuleType | Implementation | str,
+        version: Version,
+        dtype_backend: str | None = None,
+    ) -> dict[str, Any]:
+        from narwhals._pandas_like.utils import narwhals_to_native_dtype
+
+        implementation = Implementation.from_backend(backend)
+        backend_version = parse_version(implementation.to_native_namespace().__version__)
+        return {
+            name: narwhals_to_native_dtype(
+                dtype=dtype,
+                dtype_backend=dtype_backend,
+                implementation=implementation,
+                backend_version=backend_version,
+                version=version,
+            )
+            for name, dtype in self.items()
+        }
+
+    def to_polars(
+        self: Self, *, backend: ModuleType | Implementation | str, version: Version
+    ) -> pl.Schema:
+        from narwhals._polars.utils import narwhals_to_native_dtype
+
+        implementation = Implementation.from_backend(backend)
+        schema: pl.Schema = implementation.to_native_namespace().Schema(
+            (name, narwhals_to_native_dtype(dtype, version))
+            for name, dtype in self.items()
+        )
+        return schema
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -185,7 +185,7 @@ class Schema(BaseSchema):
                 for name, dtype, backend in zip(self.keys(), self.values(), backends)
             }
 
-    def to_polars(self: Self) -> pl.Schema | Any:
+    def to_polars(self: Self) -> pl.Schema:
         """Convert Schema to a polars Schema.
 
         Returns:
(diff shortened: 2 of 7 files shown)


dangotbanned:
@MarcoGorelli I've pushed this early to get some thoughts on handling `Version`.

To avoid needing a user-provided arg, I think we'd need to approach it differently than in here with the public/private function pair.

https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/functions.py#L436-L451

If you have something like:

```py
class Schema(BaseSchema):
    _version: ClassVar[Version] = Version.MAIN
```

That could easily be [overridden for v1](https://github.com/narwhals-dev/narwhals/blob/53e780c7e7a2640cca03005a27f2c6913cc40e7e/narwhals/stable/v1/__init__.py#L1032):

```py
class Schema(NwSchema):
    _version = Version.V1
```

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/13135932907/job/36651208874?pr=1924#step:9:336

I am planning to get full-coverage, but seems premature to do so *before* working out the signatures (https://github.com/narwhals-dev/narwhals/pull/1924#issuecomment-2631115464)

MarcoGorelli:
thanks Dan! Yes, having a private `_version` sounds good to me

Some other comments:
- I don't think you need `backend` for `to_pandas`. I think you just need `dtype_backend`, whose signature should match the one in https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html
- I'm not sure about `to_native`, because I might then expect to do:
  ```python
  df= pd.DataFrame({'a': [1,2,3]})
 nw.from_native(df).schema.to_native()
 ```
  and expect to get a pandas schema out the other end. However, that's not what happens, the `nw.Schema` has no knowledge of the underlying implementation. And I think that's OK. Shall we start with just `to_pandas` / `to_pyarrow` / `to_polars`?

dangotbanned:
@MarcoGorelli could you take a look at the markdown for this please? 🤯 (https://github.com/narwhals-dev/narwhals/pull/1924#issuecomment-2634670454)

MarcoGorelli:
yup, thanks!

dangotbanned:
> thanks Dan! Yes, having a private `_version` sounds good to me

Great, I'll get that added soon

> Some other comments:
> 
> * I don't think you need `backend` for `to_pandas`. I think you just need `dtype_backend`, whose signature should match the one in [pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html)
> 
> * I'm not sure about `to_native`, because I might then expect to do:
> ```python
> df= pd.DataFrame({'a': [1,2,3]})
> nw.from_native(df).schema.to_native()
> ```  
> and expect to get a pandas schema out the other end.
> However, that's not what happens, the `nw.Schema` has no knowledge of the underlying implementation. And I think that's OK. 

I think you've stumbled into an interesting topic here @MarcoGorelli.
So the current signatures *deviate* from the [similar methods](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.to_native) for `nw.DataFrame` - as here `backend` is required to account for this lack of knowledge.

Mainly this was from adapting [`nw.from_dict`](https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_dict) - which (prior to #1931) required a `ModuleType` and that acted as a guarantee that we had the import.
Now, the main difference between this and `from_dict` is that `backend` can't be `None` - since there isn't some context to do this trick

https://github.com/narwhals-dev/narwhals/blob/71a5bc5d6848da10e658db9d68a4a139f37c099b/narwhals/functions.py#L473-L480

---

> Shall we start with just `to_pandas` / `to_pyarrow` / `to_polars`?

Yeah definitely! 
Now that I understand whats going on behind the scenes, `to_native` seems less helpful.

### Question
If we drop the `backend` requirement and `to_native()`, do we need handling in `nw.Implementation` for missing imports?

I *can* keep this local to `nw.Schema.to_*`.
But I feel like with the push (#1917, #1931) towards `backend: ModuleType | Implementation | str` - there might be a benefit in some *"infallible"* paths.
If only to provide consistent errors?

MarcoGorelli:
I think in general we need a `import_optional_dependency` utility function, which can unify error messages for missing imports

but `nw.Schema.to_pandas` can import pandas (and let it raise if it's not available), `nw.Schema.to_polars` can import Polars (and let it raise if it's not available), etc...

dangotbanned:
> I think in general we need a `import_optional_dependency` utility function, which can unify error messages for missing imports
> 
> but `nw.Schema.to_pandas` can import pandas (and let it raise if it's not available), `nw.Schema.to_polars` can import Polars (and let it raise if it's not available), etc...

https://results.pre-commit.ci/run/github/760058710/1738698879.vjLSyWjmSYiQy0B_ViLJlA

@MarcoGorelli should this fall under the same *exception-to-the-rule* as the `dtypes` modules?

narwhals/schema.py
```diff
@@ -85,3 +97,47 @@ def len(self: Self) -> int:
...
+            narwhals_to_native_dtype,
+            implementation=Implementation.from_native_namespace(pd),
+            backend_version=parse_version(pd.__version__),
+            version=self._version,
+        )
+        if dtype_backend is None or isinstance(dtype_backend, str):
+            return {
+                name: to_native(dtype=dtype, dtype_backend=dtype_backend)
+                for name, dtype in self.items()
+            }
```
dangotbanned, line 128: This branch is the only bit not covered by existing tests

MarcoGorelli:
thanks, looking good!

you can add a `# ignore-banned-import` where necessary, the check is only there go nudge devs towards double-checking when they perform certain imports that they're really only doing them where necessary

narwhals/functions.py
```diff
@@ -495,18 +489,7 @@ def _from_dict_impl(  # noqa: PLR0915
...
-            )
-
-            schema_pl = {
-                name: polars_narwhals_to_native_dtype(dtype, version=version)
-                for name, dtype in schema.items()
-            }
-        else:
-            schema_pl = None
-
+        schema_pl = Schema(schema).to_polars() if schema else None
```
MarcoGorelli, line 492: very tasty, nice!

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/1924#issuecomment-2635181448

Thanks @MarcoGorell, will give this a try tomorrow

narwhals/schema.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1924/commits/89254450a822ef195d0f2277656cc1e1b2b00e48)

tests/frame/schema_test.py
```diff
@@ -330,3 +332,67 @@ def test_all_nulls_pandas() -> None:
...
+            {
+                "a": "Int64",
+                "b": "string[pyarrow]",
+                "c": "bool",
+                "d": "Float64[pyarrow]",
+                "e": "datetime64[us]",
+            },
+        ),
+    ],
+)
```
dangotbanned, line 385: ### FIXME
Adapt these to consider what is available per-version

https://github.com/narwhals-dev/narwhals/actions/runs/13154862815/job/36709602903?pr=1924

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1924/commits/83949f8da3c038556ab7bd491f686ef8064dbb7c)

dangotbanned:
@MarcoGorelli FYI I've opened (https://github.com/vega/altair/issues/3789) regarding the [downstream failures](https://github.com/narwhals-dev/narwhals/actions/runs/13155711375/job/36712254776?pr=1924)

narwhals/schema.py
```diff
@@ -85,3 +97,48 @@ def len(self: Self) -> int:
...
+        else:
+            return {
+                name: to_native(dtype=dtype, dtype_backend=backend)
+                for name, dtype, backend in zip(self.keys(), self.values(), dtype_backend)
+            }
+
+    def to_polars(self: Self) -> pl.Schema | Any:
```
dangotbanned, line 135: Needs doc

narwhals/schema.py
```diff
@@ -85,3 +97,48 @@ def len(self: Self) -> int:
...
+        import pyarrow as pa  # ignore-banned-import
+
+        from narwhals._arrow.utils import narwhals_to_native_dtype
+
+        return pa.schema(
+            (name, narwhals_to_native_dtype(dtype, self._version))
+            for name, dtype in self.items()
+        )
+
+    def to_pandas(
```
dangotbanned, line 111: Needs doc

narwhals/schema.py
```diff
@@ -85,3 +97,48 @@ def len(self: Self) -> int:
             Number of columns.
         """
         return len(self)
+
+    def to_arrow(self: Self) -> pa.Schema:
```
dangotbanned, line 101: Needs doc

narwhals/schema.py
```diff
@@ -85,3 +97,70 @@ def len(self: Self) -> int:
...
+    def to_polars(self: Self) -> pl.Schema | Any:
+        import polars as pl  # ignore-banned-import
+
+        from narwhals._polars.utils import narwhals_to_native_dtype
+
+        it = (
+            (name, narwhals_to_native_dtype(dtype, self._version))
+            for name, dtype in self.items()
+        )
+        return pl.Schema(it) if parse_version(pl.__version__) >= (1, 0, 0) else dict(it)
```
dangotbanned, line 166: > [!NOTE]
> The `pl.Schema | Any` annotation is the result of this.

I've used `Any` as a pragmatic solution.

other ideas I had that seem like a worse UX:
- `pl.Schema | dict[str, PolarsDataType]` would mean you need to check the result - regardless of version
- adding some kind of flag argument to resolve an overload seems annoying

I don't think there's a clean way to resolve this *statically*, since it is bound to the `polars` version

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/13159096422/job/36723068890?pr=1924

Pretty sure this is unrelated to the PR

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+
+        return pa.schema(
+            (name, narwhals_to_native_dtype(dtype, self._version))
+            for name, dtype in self.items()
+        )
+
+    def to_pandas(
+        self: Self, *, dtype_backend: str | Iterable[str] | None = None
+    ) -> dict[str, Any]:
+        """Convert Schema to an ordered mapping of column names to their pandas data type.
```
dangotbanned, line 126: I'm still not that happy about the verbosity, but didn't see a way to adapt the content of [convert_dtypes](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html) that made sense here

MarcoGorelli:
awesome, thanks Dan! gonna be a bit busy over the next couple of days, but i'll try to review at the weekend and see if we can get it in to Monday's release

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/1924#issuecomment-2637442091 thanks @MarcoGorelli 

Feel free to make any changes you think will help it fit in better w/ `narwhals`

narwhals/schema.py
dangotbanned: Not sure if this was clear, but this also meant adding a `# doctest:+SKIP` here

https://github.com/dangotbanned/narwhals/blob/bce4b20d6d7cbb02345ca4d1d99d566e4d09a2aa/narwhals/schema.py#L199-200

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+            >>> schema.to_pandas(dtype_backend=["pandas-nullable", "pyarrow-nullable"])
+            {'a': 'Int64', 'b': 'timestamp[ns][pyarrow]'}
+        """
+        import pandas as pd  # ignore-banned-import
+
+        from narwhals._pandas_like.utils import narwhals_to_native_dtype
+
+        to_native = partial(
+            narwhals_to_native_dtype,
+            implementation=Implementation.from_native_namespace(pd),
```
MarcoGorelli, line 153: this could just be `Implementation.PANDAS`, right?

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+                msg = (
+                    f"Provided {n_user!r} `dtype_backend`(s), but schema contains {n_actual!r} field(s).\n"
+                    "Hint: instead of\n"
+                    f"    schema.to_pandas(dtype_backend={backends})\n"
+                    "you may want to use\n"
+                    f"    schema.to_pandas(dtype_backend={backends[0]})\n"
+                    f"or\n"
+                    f"    schema.to_pandas(dtype_backend={suggestion})"
+                )
+                raise ValueError(msg)
```
MarcoGorelli, line 184: Do you have a use-case for passing multiple dtype-backends?

If not, I'd suggest we only support `str | None = None`. Like this, we:
- keep complexity down
- match `dtype_backend` in pandas' API (e.g. `read_csv`)

Because if we did want to support multiple backend dtypes, there's an argument that we should also support `Mapping`

Can we start with just the simplest case please?

---

This means we'd need to rename the dtype backend values we use internally in `narwhals/_pandas_like/utils.py`, but that would just be an internal change so we're OK to "just do it"

We should also raise if an unexpected `dtype_backend` value is passed

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+        Examples:
+            >>> import narwhals as nw
+            >>> schema = nw.Schema({"a": nw.Int64(), "b": nw.Datetime("ns")})
+            >>> schema.to_pandas()
+            {'a': 'int64', 'b': 'datetime64[ns]'}
+
+            >>> schema.to_pandas(dtype_backend="pyarrow-nullable")
+            {'a': 'Int64[pyarrow]', 'b': 'timestamp[ns][pyarrow]'}
+
+            >>> schema.to_pandas(dtype_backend=["pandas-nullable", "pyarrow-nullable"])
```
MarcoGorelli, line 144: we're diverging from the pandas names here

should we just use `'pyarrow'` and `'numpy_nullable'`? Like this, they match the current pandas names

If pandas then renames, as is suggested in https://github.com/pandas-dev/pandas/issues/59032, then we can change to use the new names in the docstring but still silently support the old ones so as to not break anyone's code

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+
+        from narwhals._arrow.utils import narwhals_to_native_dtype
+
+        return pa.schema(
+            (name, narwhals_to_native_dtype(dtype, self._version))
+            for name, dtype in self.items()
+        )
+
+    def to_pandas(
+        self: Self, *, dtype_backend: str | Iterable[str] | None = None
```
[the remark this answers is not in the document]
MarcoGorelli: I think it's fine for it to be positional or keyword

narwhals/schema.py
```diff
@@ -85,3 +97,114 @@ def len(self: Self) -> int:
...
+            >>> import narwhals as nw
+            >>> schema = nw.Schema({"a": nw.Int64(), "b": nw.Datetime("ns")})
+            >>> schema.to_polars()  # doctest:+SKIP
+            Schema([('a', Int64), ('b', Datetime(time_unit='ns', time_zone=None))])
+        """
+        import polars as pl  # ignore-banned-import
+
+        from narwhals._polars.utils import narwhals_to_native_dtype
+
+        it = (
```
MarcoGorelli, line 206: could we spell this variable out a bit more please?

narwhals/schema.py
MarcoGorelli: I think it's fine to type as if users are using the latest Polars version

If they're on an old Polars version that doesn't have `pl.Schema`, then their type checker would just resolve it as `Any` anyway (as far as I know)

MarcoGorelli: in CI we only run doctests on the latest versions, so I think it's fine to remove the doctest skip too?

MarcoGorelli (changes requested):
thanks @dangotbanned , awesome feature! can't wait to use this as I find it super-difficult to write pyarrow schemas

i've just left some comments but i think we can ship this quite soon!

narwhals/schema.py
dangotbanned: > we're diverging from the pandas names here
> 
> should we just use `'pyarrow'` and `'numpy_nullable'`? Like this, they match the current pandas names
> 
> If pandas then renames, as is suggested in [pandas-dev/pandas#59032](https://github.com/pandas-dev/pandas/issues/59032), then we can change to use the new names in the docstring but still silently support the old ones so as to not break anyone's code

I did notice this when reading the `pandas` docs as well.
But I was using what seems to already be declared here:

https://github.com/narwhals-dev/narwhals/blob/f33e82ca36695b8d4782f167d2cba058eda5c5d9/narwhals/_pandas_like/utils.py#L517-L540

dangotbanned: Are you asking if those should be aligned with `pandas` or this method only?

narwhals/schema.py
dangotbanned: > this could just be `Implementation.PANDAS`, right?

Interesting 🤔 

So it seems like you can get pretty far into this function before hitting a `pandas` import (517 -> 658)

https://github.com/narwhals-dev/narwhals/blob/f33e82ca36695b8d4782f167d2cba058eda5c5d9/narwhals/_pandas_like/utils.py#L517

I didn't think this would be an option huh

@MarcoGorelli do you think it might be confusing that `pandas` is only required here for nested dtypes?

narwhals/schema.py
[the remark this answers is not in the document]
dangotbanned: > I think it's fine for it to be positional or keyword

Just for my benefit, which of these were you asking for?

```py
    def to_pandas1( # pos-only
        self: Self, dtype_backend: str | Iterable[str] | None = None, /
    ) -> dict[str, Any]: ...
    def to_pandas2( # pos-or-kwd
        self: Self, dtype_backend: str | Iterable[str] | None = None
    ) -> dict[str, Any]: ...
    def to_pandas3( #kwd-only
        self: Self, *, dtype_backend: str | Iterable[str] | None = None
    ) -> dict[str, Any]: ...
```

I'm pretty sure you want `to_pandas2` right?

narwhals/schema.py
MarcoGorelli: aah I see, thanks for explaining! ok with keeping this then

narwhals/schema.py
MarcoGorelli: it's fine to import pandas here, we need it in the `backend_version=` line anyway, but i think we can skip a step in `Implementation.from_native_namespace(pd)` and go straight to `Implementation.PANDAS`

narwhals/schema.py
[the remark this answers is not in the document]
MarcoGorelli: yup, `to_pandas2`, I should've just written it out explicitly

narwhals/schema.py
dangotbanned: Ohh I see, thanks for explaining @MarcoGorelli, will give that a try soon

narwhals/schema.py
dangotbanned: > could we spell this variable out a bit more please?

😭 

```diff
diff --git a/narwhals/schema.py b/narwhals/schema.py
index 0466bbb4..ef29b7f6 100644
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -203,8 +203,12 @@ class Schema(BaseSchema):
 
         from narwhals._polars.utils import narwhals_to_native_dtype
 
-        it = (
+        schema = (
             (name, narwhals_to_native_dtype(dtype, self._version))
             for name, dtype in self.items()
         )
-        return pl.Schema(it) if parse_version(pl.__version__) >= (1, 0, 0) else dict(it)
+        return (
+            pl.Schema(schema)
+            if parse_version(pl.__version__) >= (1, 0, 0)
+            else dict(schema)
+        )

```

But more seriously, yeah this should be fine

narwhals/schema.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1924/commits/95a123a644215683fec0eb9619e25103b915d649)

narwhals/schema.py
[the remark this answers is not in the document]
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1924/commits/4581d95912f36f6fa664576ab3ebd31888afd2cf)

narwhals/schema.py
dangotbanned: > in CI we only run doctests on the latest versions, so I think it's fine to remove the doctest skip too?

Hmm, this could work

https://github.com/narwhals-dev/narwhals/actions/runs/13205078790/job/36866210140?pr=1924#step:10:347

I feel like I just copy/pasted the REPL, maybe the repr changed? (or I'm writing lists of tuples by choice?)

```py
Expected:
    Schema([('a', Int64), ('b', Datetime(time_unit='ns', time_zone=None))])
Got:
    Schema({'a': Int64, 'b': Datetime(time_unit='ns', time_zone=None)})
```

narwhals/schema.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1924/commits/881aae0219482613098725d9a6d1986a78e983ca)

narwhals/schema.py
dangotbanned: > I think it's fine to type as if users are using the latest Polars version
> 
> If they're on an old Polars version that doesn't have `pl.Schema`, then their type checker would just resolve it as `Any` anyway (as far as I know)

@MarcoGorelli this is what `pl.Schema | Any` is hiding right now

image

I'm not sure what you're proposing here

narwhals/schema.py
dangotbanned: > i think we can align both with pandas

Could that be part of a future PR?
I'm not against aligning, but it seems like a bit out of scope personally

dangotbanned: Hang on, @MarcoGorelli 
have you been implying that `'pyarrow'` and `'numpy_nullable'` will produce the same types as `'pyarrow-nullable'` and `'pandas-nullable'`?

MarcoGorelli: this is public-facing, so changing later would be a lot harder - I may not have been clear though, i can add a commit later, this should be very close to ready-to-go and i'd really like to see this feature 🚀

dangotbanned: Happy for you to make changes @MarcoGorelli ♥️

narwhals/schema.py
dangotbanned: @MarcoGorelli if you make a change to resolve this, could you double-check it plays nicely with `pyright` please?

narwhals/typing.py
```diff
@@ -244,6 +245,7 @@ def lit(
     ...     return s.abs().to_native()
 """
 
+DTypeBackend: TypeAlias = Optional[Literal["pyarrow", "numpy_nullable"]]
```
dangotbanned, line 248:
```suggestion
DTypeBackend: TypeAlias = 'Literal["pyarrow", "numpy_nullable"] | None'
```

narwhals/schema.py
MarcoGorelli: sure, looks fine now:
```console
$ pyright narwhals/schema.py 
0 errors, 0 warnings, 0 informations 
```

MarcoGorelli (approved):
thanks @dangotbanned 

just made some minor edits, and got `dtype_backend` to match how it currently is in pandas

happy with this?

dangotbanned:
> thanks @dangotbanned
> 
> just made some minor edits, and got `dtype_backend` to match how it currently is in pandas
> 
> happy with this?
> 

Thanks @MarcoGorelli, just taking a look now

narwhals/schema.py
```diff
@@ -85,3 +98,113 @@ def len(self: Self) -> int:
...
+
+        from narwhals._polars.utils import narwhals_to_native_dtype
+
+        schema = (
+            (name, narwhals_to_native_dtype(dtype, self._version))
+            for name, dtype in self.items()
+        )
+        if parse_version(pl.__version__) < (1, 0, 0):  # pragma: no cover
+            return dict(schema)  # type: ignore[return-value]
+        return pl.Schema(schema)
```
dangotbanned, line 210: @MarcoGorelli why is the order reversed now?

This is just one *small* example from `typing_extensions`:

https://github.com/python/typing_extensions/blob/8184ac61398c187203dad819eb5b9d34005a96ae/src/typing_extensions.py#L547-L552

I find the pattern there is easier to follow:

```py
if parse_version(pl.__version__) >= (1, 0, 0):
    # happy path first
    ...
else:
    # compat path
    ...
```

dangotbanned: The import of `cast` was blocking me adding it as a suggestion (https://github.com/narwhals-dev/narwhals/pull/1924/commits/3bcebc1c2b75eddebf3e88544e0cb312b9361020)

tests/frame/schema_test.py
```diff
@@ -330,3 +334,99 @@ def test_all_nulls_pandas() -> None:
...
+            rf".+3.+but.+schema contains.+5.+field.+Hint.+schema.to_pandas{suggestion}",
+            re.DOTALL,
+        ),
+    ):
+        schema.to_pandas(dtype_backend)
+
+
+def test_schema_to_pandas_invalid() -> None:
+    schema = nw.Schema({"a": nw.Int64()})
+    msg = "Expected one of {None, 'pyarrow', 'numpy_nullable'}, got: 'cabbage'"
```
dangotbanned, line 430: 😆

narwhals/schema.py
dangotbanned: oop I started another thread (sorry) https://github.com/narwhals-dev/narwhals/pull/1924#discussion_r1947895133

narwhals/schema.py
MarcoGorelli: it just became easier to not have `type ignore` and `pragma no cover` on the same line (we measure coverage in the ci job with the latest versions - i've found combining coverage jobs to be too unreliable and flaky)

sure, no objections to https://github.com/narwhals-dev/narwhals/commit/3bcebc1c2b75eddebf3e88544e0cb312b9361020

dangotbanned:
Ready to merge if you're onboard w/ (https://github.com/narwhals-dev/narwhals/pull/1924#discussion_r1947895133) @MarcoGorelli 

Thanks for helping me through this 🎉

MarcoGorelli:
sure, looks good, feel free to apply that and ship it

MarcoGorelli:
changing the title? sure, i think you should have permission to edit

dangotbanned:
> changing the title? sure, i think you should have permission to edit

Oh yeah 🎉

PR #1815, opened 2025-01-16
feat: validate duplicate column names in pyarrow and duckdb

IsaiasGutierrezCruz:
closes #1920 

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1770 
- Closes #1770 

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [] Documented the changes

## If you have comments or can explain your changes, please do so below

To test the validations, I added three methods in the file tests/frame/schema_test.py. @MarcoGorelli, if you think these would be more appropriate in another file, please feel free to let me know. Additionally, I had to add duckdb as a dependency to the extreme GitHub Actions workflow to accommodate these tests.

PR #1933, opened 2025-02-04
fix: raise when join introduces duplicates

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -21,6 +21,7 @@ from narwhals.dependencies import is_numpy_array
 from narwhals.utils import Implementation
 from narwhals.utils import Version
 from narwhals.utils import check_column_exists
+from narwhals.utils import check_column_names_are_unique
 from narwhals.utils import generate_temporary_column_name
 from narwhals.utils import is_sequence_but_not_str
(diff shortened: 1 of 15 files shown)


narwhals/_arrow/dataframe.py
```diff
@@ -90,10 +91,15 @@ def _change_version(self: Self, version: Version) -> Self:
             self._native_frame, backend_version=self._backend_version, version=version
         )
 
-    def _from_native_frame(self: Self, df: pa.Table) -> Self:
-        return self.__class__(
+    def _from_native_frame(
+        self: Self, df: pa.Table, *, validate_column_names: bool = False
```
MarcoGorelli, line 95: 🤔 maybe the internal default should be true, and then we just opt out of the check when we know it's safe to do so

MarcoGorelli:
i think we can just use https://github.com/narwhals-dev/narwhals/pull/1815

PR #1935, opened 2025-02-04
fix(typing): Use correct `TypeVar`/`Union` pairs in `from_native`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1928

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Resolves 4x `pyright` warnings.

I was planning to look at fixing (#1928) but these needed to be addressed first

image

PR #1936, opened 2025-02-04
fix(typing): Use `IntoDataFrameT` in `to_native`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Mentioned in https://github.com/narwhals-dev/narwhals/pull/1935#issuecomment-2634298106

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Resolves a similar issue as (#1935)

<details><summary>Warning (very long message)</summary>
<p>

> Overloaded implementation is not consistent with signature of overload 1
>   Type "(narwhals_object: DataFrame[IntoFrameT@to_native] | LazyFrame[IntoFrameT@to_native] | Series[IntoSeriesT@to_native], *, strict: bool | None = None, pass_through: bool | None = None) -> (IntoFrameT@to_native | Any)" is not assignable to type "(narwhals_object: DataFrame[IntoDataFrameT@to_native], *, pass_through: Literal[False] = ...) -> IntoDataFrameT@to_native"
>     Parameter 1: type "DataFrame[IntoDataFrameT@to_native]" is incompatible with type "DataFrame[IntoFrameT@to_native] | LazyFrame[IntoFrameT@to_native] | Series[IntoSeriesT@to_native]"
>       Type "DataFrame[IntoDataFrameT@to_native]" is not assignable to type "DataFrame[IntoFrameT@to_native] | LazyFrame[IntoFrameT@to_native] | Series[IntoSeriesT@to_native]"
>         "DataFrame[IntoDataFrameT@to_native]" is not assignable to "DataFrame[IntoFrameT@to_native]"
>           Type parameter "DataFrameT@DataFrame" is invariant, but "IntoDataFrameT@to_native" is not the same as "IntoFrameT@to_native"
>         "DataFrame[IntoDataFrameT@to_native]" is not assignable to "LazyFrame[IntoFrameT@to_native]"
>         "DataFrame[IntoDataFrameT@to_native]" is not assignable to "Series[IntoSeriesT@to_native]"PylancereportInconsistentOverload
> translate.py(70, 5): Overload signature is defined here

image

</p>
</details>

Issue #1928, opened 2025-02-04
missing overload for `nw.from_native(obj, eager_only=True, series_only=True)`


MarcoGorelli:

dangotbanned:
Could you give a repro that I can target as solving?

Not sure if (#1935) will impact this

MarcoGorelli:
hey - think just
```
s: Any
reveal_type(nw.from_native(s, eager_only=True, series_only=True))
```
should reproduce it

dangotbanned:
This might take a while to resolve.

I've been playing around with the `@overload`(s) and they're super fragile.
Thought I had a quick fix, but it indirectly broke other ones

dangotbanned:
[`enum.IntFlag`](https://docs.python.org/3/library/enum.html#enum.IntFlag) might be an alternative approach for simpler `@overload`(s).

If all these parameters were reduced to `flags: SomeName`, then you wouldn't need to worry about the combinatorial explosion situation we have here

dangotbanned:
I've just come back to this after a while and decided to try fixing (https://github.com/narwhals-dev/narwhals/issues/1928#issuecomment-2663895487)

Without adding any new `@overload`s, I got **5/6** passing:

<img width="900" alt="Image" src="https://github.com/user-attachments/assets/c1ea7f41-395b-4791-8aa1-5c078e3f184c" />

I think the general issue with `from_native` has been that these `@overload`s are both strict and incomplete.
Most of them are dedicated to `pass_through: Literal[True]`, but would have the same behavior for `pass_through: bool`.

> [!IMPORTANT]
> Simply changing one/two quickly snowballs into other issues 

## Suggestion
We should replace some of the parametric testing in [`tests/translate/from_native_test.py`](https://github.com/narwhals-dev/narwhals/blob/aef26cec50026280f6d3d28a63086e2bd427ff72/tests/translate/from_native_test.py) with something a typechecker can help us with.

### Example
This test produces a  `# type: ignore[call-overload]` in all 4 cases, but only 1 actually raises at runtime


https://github.com/narwhals-dev/narwhals/blob/aef26cec50026280f6d3d28a63086e2bd427ff72/tests/translate/from_native_test.py#L438-L460

With the changes I made for `Series` overloads, I then went back and isolated the failing case (it started giving a different warning, which prompted this):

```py
def test_eager_only_pass_through_main(constructor: Constructor) -> None:
[...]

PR #1953, opened 2025-02-06
fix(typing): Backport `from_native` overload fixes to `v1`

dangotbanned:

- [x] 🐛 Bug Fix

## Related issues

- Related PR #1935

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Aligns with #1935, without using `IntoSeriesT` as `v1.Series` is not generic
https://narwhals-dev.github.io/narwhals/backcompat/#breaking-changes-carried-out-so-far

### DIFF
--- a/narwhals/stable/v1/__init__.py
+++ b/narwhals/stable/v1/__init__.py
@@ -67,7 +67,7 @@ from narwhals.translate import get_native_namespace
 from narwhals.translate import to_py_scalar
 from narwhals.typing import IntoDataFrameT
 from narwhals.typing import IntoFrameT
-from narwhals.typing import IntoSeriesT
+from narwhals.typing import IntoSeriesT as IntoSeriesT  # noqa: PLC0414
 from narwhals.utils import Implementation
 from narwhals.utils import Version
 from narwhals.utils import find_stacklevel
@@ -90,6 +90,7 @@ if TYPE_CHECKING:
     from narwhals.dtypes import DType
     from narwhals.functions import ArrowStreamExportable
     from narwhals.typing import IntoExpr
+    from narwhals.typing import IntoFrame
     from narwhals.typing import IntoSeries
 
 T = TypeVar("T")
@@ -1124,7 +1125,7 @@ def _stableify(
 
 @overload
 def from_native(
-    native_object: IntoDataFrameT | IntoSeriesT,
+    native_object: IntoDataFrameT | IntoSeries,
     *,
     strict: Literal[False],
     eager_only: Literal[False] = ...,
@@ -1136,7 +1137,7 @@ def from_native(
 
 @overload
 def from_native(
-    native_object: IntoDataFrameT | IntoSeriesT,
+    native_object: IntoDataFrameT | IntoSeries,
     *,
     strict: Literal[False],
     eager_only: Literal[True],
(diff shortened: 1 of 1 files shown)


dangotbanned:
While trying out the repro for https://github.com/narwhals-dev/narwhals/issues/1928#issuecomment-2634320499, I realized (#1935, #1936) needed backports

Will follow up with another for (#1936)

narwhals/stable/v1/__init__.py
```diff
@@ -67,7 +67,7 @@
 from narwhals.translate import to_py_scalar
 from narwhals.typing import IntoDataFrameT
 from narwhals.typing import IntoFrameT
-from narwhals.typing import IntoSeriesT
+from narwhals.typing import IntoSeriesT as IntoSeriesT  # noqa: PLC0414
```
FBruzzesi, line 70: I will take this as a learning opportunity: can someone explain to me the reason for this aliasing? 🙈

FBruzzesi (commented):
Thanks @dangotbanned, I have a tiny question 🤔 everything else looks good

narwhals/stable/v1/__init__.py
dangotbanned: This change removed all uses of `IntoSeriesT` in the module.

To prevent another rule/fix removing the import - I had to add a redundant alias
- https://docs.astral.sh/ruff/rules/unused-import/

dangotbanned: IIRC I was getting a warning from a different rule, but `F401` is the one I know

FBruzzesi: Thanks for explaining 👌

I think it's ok to remove the import then if unused. I would expect/suggest user to import from the typing modules and directly from narwhals (either main or v1) only those classes and functionalities that are explicitly in `__all__`.

We might be missing some of them, but typing should not belong there

FBruzzesi: On a separate thought, using a redundant alias and not using it seems like a hack/bug from a linter 🙃

dangotbanned: @FBruzzesi that's the redundant part! 😉 

I'm fine with removing `IntoSeriesT`, I was just trying to avoid making something that *could* be considered a breaking change

No worries if you're using `__all__` for that though

FBruzzesi: > I was just trying to avoid making something that could be considered a breaking change

Ooooh that's a good point actually 😭 Searching on github I can't find any usage of:

```py

nw.IntoSeriesT
```
and actually my editor warns me that `IntoSeriesT` is not explicitly exported.

Let's wait for @MarcoGorelli to weight on this then!

MarcoGorelli: thanks for the attention to detail here!

agree that as `IntoSeriesT` wasn't explicitly exported in `__all__`, and that there's no public usage of it on github, it should be fine to just remove

MarcoGorelli (approved):
thanks @dangotbanned !

i'm super-favourable to any PR to improve static typing so please send more of these if they interest you, would love to improve typing in Narwhals

dangotbanned:
Thanks @MarcoGorelli!

> i'm super-favourable to any PR to improve static typing so please send more of these if they interest you, would love to improve typing in Narwhals

And so begins, the war against red squiggles

MarcoGorelli (approved):
thanks @dangotbanned ! CI failure is unrelated and due to the latest ibis release

PR #1954, opened 2025-02-06
fix(typing): Backport use `IntoDataFrameT` in `to_native`  to `v1`

dangotbanned:

- [x] 🐛 Bug Fix

## Related issues

- Related PR #1936

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Aligns with #1936 and similar change in #1953

### DIFF
--- a/narwhals/stable/v1/__init__.py
+++ b/narwhals/stable/v1/__init__.py
@@ -1603,7 +1603,7 @@ def to_native(narwhals_object: Any, *, pass_through: bool) -> Any: ...
 
 
 def to_native(
-    narwhals_object: DataFrame[IntoFrameT] | LazyFrame[IntoFrameT] | Series,
+    narwhals_object: DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT] | Series,
     *,
     strict: bool | None = None,
     pass_through: bool | None = None,


narwhals/stable/v1/__init__.py
```diff
@@ -1603,7 +1603,7 @@ def to_native(narwhals_object: Any, *, pass_through: bool) -> Any: ...
 
 
 def to_native(
-    narwhals_object: DataFrame[IntoFrameT] | LazyFrame[IntoFrameT] | Series,
+    narwhals_object: DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT] | Series,
     *,
     strict: bool | None = None,
     pass_through: bool | None = None,
```
FBruzzesi, line 1609: Can't comment on other lines so I will do it here:

- Shouldn't lines 1590 and 1600 return annotation be `IntoSeries` (as `strict=True`/`pass_through=False`)
- Then shouldn't the final return annotation be `IntoDataFrameT | IntoFrameT | IntoSeries | Any` (line 1610)

FBruzzesi (approved):
Nice spotting! Thanks @dangotbanned! I left one comment but can be addressed on a separate PR, or just ignored if incorrect (I still struggle with typing)

NarwhalChubbiwhalGIF

narwhals/stable/v1/__init__.py
dangotbanned: > Shouldn't lines 1590 and 1600 return annotation be IntoSeries (as strict=True/pass_through=False)

@FBruzzesi I'd like to take a look at reducing the number of overloads here and for `from_native` *first* and then maybe correct what's left

This particular change was just to support these two signatures

- https://github.com/dangotbanned/narwhals/blob/a582d572e544238cc69eb6b6ea10631c87e93085/narwhals/stable/v1/__init__.py#L1581
- https://github.com/dangotbanned/narwhals/blob/a582d572e544238cc69eb6b6ea10631c87e93085/narwhals/stable/v1/__init__.py#L1593

> Then shouldn't the final return annotation be IntoDataFrameT | IntoFrameT | IntoSeries | Any (line 1610)

Right now, since the final return annotation contains an `Any` - it doesn't matter what else is there

https://github.com/dangotbanned/narwhals/blob/a582d572e544238cc69eb6b6ea10631c87e93085/narwhals/stable/v1/__init__.py#L1610

PR #1955, opened 2025-02-06
chore(typing): add guards `is_compliant_*` (+ friends) 

### LINKED ISSUE #1937: add `is_compliant_*` guards
Resolving `pyright` warnings that look like:
> Attribute `"__narwhals_*__"` is unknown

<details><summary>diff</summary>
<p>

```diff
diff --git a/narwhals/translate.py b/narwhals/translate.py
index 3758ebc6..fd3733e7 100644
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -364,6 +364,7 @@ def _from_native_impl(  # noqa: PLR0915
     from narwhals.dataframe import DataFrame
     from narwhals.dataframe import LazyFrame
     from narwhals.series import Series
+    from narwhals.typing import is_compliant_dataframe
     from narwhals.utils import Implementation
     from narwhals.utils import parse_version
 
@@ -407,7 +408,7 @@ def _from_native_impl(  # noqa: PLR0915
         )
 
     # Extensions
-    elif hasattr(native_object, "__narwhals_dataframe__"):
+    elif is_compliant_dataframe(native_object):
         if series_only:
             if not pass_through:
                 msg = "Cannot only use `series_only` with dataframe"
diff --git a/narwhals/typing.py b/narwhals/typing.py
index 0c284cfa..b051e8ee 100644
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -16,6 +16,10 @@ if TYPE_CHECKING:
     from narwhals.dtypes import DType
     from narwhals.utils import Implementation
 
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
     if sys.version_info >= (3, 10):
         from typing import TypeAlias
     else:
@@ -68,6 +72,10 @@ class CompliantLazyFrame(Protocol):
     ) -> CompliantLazyFrame: ...  # `select` where all args are column names
 
 
+def is_compliant_dataframe(obj: Any) -> TypeIs[CompliantDataFrame]:
+    return hasattr(obj, "__narwhals_dataframe__")
+
+
 CompliantSeriesT_co = TypeVar(
     "CompliantSeriesT_co", bound=CompliantSeries, covariant=True
 )
```

</p>
</details>

Repeating this for the other protocols will resolve **13/16** warnings in `translate.py`

Image

Mentioned in (https://discord.com/channels/1235257048170762310/1272835922924273694/1336364263131975783)

dangotbanned:
Closes #1937

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes  #1937

## Checklist

- [ ] Tests added
- [ ] Documented the changes

`has_native_namespace`, `is_dataframe_like` were needed to resolve all the warnings

### `translate.py` now has 0 red squiggles!
image

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -16,6 +16,7 @@ from narwhals.exceptions import InvalidIntoExprError
 from narwhals.exceptions import LengthChangingExprError
 from narwhals.utils import Implementation
 from narwhals.utils import is_compliant_expr
+from narwhals.utils import is_compliant_series
 
 if TYPE_CHECKING:
     from narwhals._arrow.expr import ArrowExpr
@@ -132,11 +133,11 @@ def parse_into_expr(
     """
     if is_compliant_expr(into_expr):
         return into_expr
-    if hasattr(into_expr, "__narwhals_series__"):
+    if is_compliant_series(into_expr):
         return namespace._create_expr_from_series(into_expr)  # type: ignore[no-any-return, attr-defined]
     if is_numpy_array(into_expr):
-        series = namespace._create_compliant_series(into_expr)  # type: ignore[attr-defined]
-        return namespace._create_expr_from_series(series)  # type: ignore[no-any-return, attr-defined]
+        series = namespace._create_compliant_series(into_expr)
+        return namespace._create_expr_from_series(series)
     raise InvalidIntoExprError.from_invalid_type(type(into_expr))
 
 
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -9,13 +9,13 @@ from typing import Callable
 from typing import Sequence
 from typing import TypeVar
 from typing import Union
-from typing import cast
 from typing import overload
 
 from narwhals.dependencies import is_numpy_array
 from narwhals.exceptions import InvalidIntoExprError
 from narwhals.exceptions import LengthChangingExprError
 from narwhals.utils import Implementation
+from narwhals.utils import is_compliant_expr
 
 if TYPE_CHECKING:
     from narwhals._arrow.expr import ArrowExpr
@@ -82,14 +82,21 @@ def evaluate_into_exprs(
     return series
 
 
+@overload
+def maybe_evaluate_expr(
+    df: CompliantDataFrame, expr: CompliantExpr[CompliantSeriesT_co]
+) -> Sequence[CompliantSeriesT_co]: ...
+
+
+@overload
+def maybe_evaluate_expr(df: CompliantDataFrame, expr: T) -> T: ...
+
+
 def maybe_evaluate_expr(
     df: CompliantDataFrame, expr: CompliantExpr[CompliantSeriesT_co] | T
 ) -> Sequence[CompliantSeriesT_co] | T:
     """Evaluate `expr` if it's an expression, otherwise return it as is."""
-    if hasattr(expr, "__narwhals_expr__"):
-        compliant_expr = cast("CompliantExpr[Any]", expr)
-        return compliant_expr(df)
-    return expr
+    return expr(df) if is_compliant_expr(expr) else expr
 
 
 def parse_into_exprs(
@@ -123,8 +130,8 @@ def parse_into_expr(
     - if it's a string, then convert it to an expression
     - else, raise
     """
-    if hasattr(into_expr, "__narwhals_expr__"):
-        return into_expr  # type: ignore[return-value]
+    if is_compliant_expr(into_expr):
+        return into_expr
     if hasattr(into_expr, "__narwhals_series__"):
         return namespace._create_expr_from_series(into_expr)  # type: ignore[no-any-return, attr-defined]
     if is_numpy_array(into_expr):
@@ -177,7 +184,7 @@ def reuse_series_implementation(
     plx = expr.__narwhals_namespace__()
 
     def func(df: CompliantDataFrame) -> Sequence[CompliantSeries]:
-        _kwargs = {  # type: ignore[var-annotated]
+        _kwargs = {
             arg_name: maybe_evaluate_expr(df, arg_value)
             for arg_name, arg_value in expressifiable_args.items()
         }
@@ -284,7 +291,7 @@ def combine_evaluate_output_names(
     def evaluate_output_names(
         df: CompliantDataFrame | CompliantLazyFrame,
     ) -> Sequence[str]:
-        if not hasattr(exprs[0], "__narwhals_expr__"):  # pragma: no cover
+        if not is_compliant_expr(exprs[0]):  # pragma: no cover
             msg = f"Safety assertion failed, expected expression, got: {type(exprs[0])}. Please report a bug."
             raise AssertionError(msg)
         return exprs[0]._evaluate_output_names(df)[:1]
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -364,11 +364,11 @@ def _from_native_impl(  # noqa: PLR0915
     from narwhals.dataframe import DataFrame
     from narwhals.dataframe import LazyFrame
     from narwhals.series import Series
-    from narwhals.typing import is_compliant_dataframe
-    from narwhals.typing import is_compliant_lazyframe
-    from narwhals.typing import is_compliant_series
-    from narwhals.typing import is_dataframe_like
     from narwhals.utils import Implementation
+    from narwhals.utils import is_compliant_dataframe
+    from narwhals.utils import is_compliant_lazyframe
+    from narwhals.utils import is_compliant_series
+    from narwhals.utils import is_dataframe_like
     from narwhals.utils import parse_version
 
     # Early returns
@@ -829,7 +829,7 @@ def get_native_namespace(
         >>> nw.get_native_namespace(df)
         <module 'polars'...>
     """
-    from narwhals.typing import has_native_namespace
+    from narwhals.utils import has_native_namespace
 
     if has_native_namespace(obj):
         return obj.__native_namespace__()
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -364,6 +364,10 @@ def _from_native_impl(  # noqa: PLR0915
     from narwhals.dataframe import DataFrame
     from narwhals.dataframe import LazyFrame
     from narwhals.series import Series
+    from narwhals.typing import is_compliant_dataframe
+    from narwhals.typing import is_compliant_lazyframe
+    from narwhals.typing import is_compliant_series
+    from narwhals.typing import is_dataframe_like
     from narwhals.utils import Implementation
     from narwhals.utils import parse_version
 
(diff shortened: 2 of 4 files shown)


narwhals/typing.py
```diff
@@ -107,6 +111,30 @@ def lit(
...
+def is_compliant_lazyframe(obj: Any) -> TypeIs[CompliantLazyFrame]:
+    return hasattr(obj, "__narwhals_lazyframe__")
+
+
+def is_compliant_series(obj: Any) -> TypeIs[CompliantSeries]:
+    return hasattr(obj, "__narwhals_series__")
+
+
+def has_native_namespace(obj: Any) -> TypeIs[_NativeNamespace]:
+    return hasattr(obj, "__native_namespace__")
```
dangotbanned, line 131: Wasn't sure on the name for this, so I prefixed the protocol name

The method isn't described by the existing `Compliant*` protocols?

narwhals/translate.py
```diff
@@ -364,6 +364,10 @@ def _from_native_impl(  # noqa: PLR0915
     from narwhals.dataframe import DataFrame
     from narwhals.dataframe import LazyFrame
     from narwhals.series import Series
+    from narwhals.typing import is_compliant_dataframe
+    from narwhals.typing import is_compliant_lazyframe
+    from narwhals.typing import is_compliant_series
+    from narwhals.typing import is_dataframe_like
```
dangotbanned, line 370: I've defined all of them in `typing`, but another home *could* be `dependencies`.

narwhals/typing.py
FBruzzesi: I think it's ok to add `__native_namespace__` to the `CompliantNamespace` protocol in line 107, and call this function `is_compliant_namespace`.

To me _native_ means `pandas`, `polars`, and so on.
WDYT?

narwhals/translate.py
FBruzzesi: I would tend to define functionalities in `dependencies` or `utils` rather than in `typing` module. Let's sleep on this 😁

narwhals/typing.py
```diff
@@ -16,6 +16,10 @@
     from narwhals.dtypes import DType
     from narwhals.utils import Implementation
 
+    if sys.version_info >= (3, 13):
+        from typing import TypeIs
+    else:
+        from typing_extensions import TypeIs
```
FBruzzesi, line 22: Should we just use typing_extensions as we do for `Self`?

FBruzzesi (commented):
Thanks @dangotbanned - I tried to address your questions in the comments 🤞

narwhals/typing.py
dangotbanned: @FBruzzesi this is a bit tricky due to where this function is used.

https://github.com/dangotbanned/narwhals/blob/503c7d3493eda3ff08e31b4fa587067b2fd3d33b/narwhals/translate.py#L832-L849

<details><summary>Code block</summary>
<p>

```py
from narwhals.typing import has_native_namespace

    if has_native_namespace(obj):
        return obj.__native_namespace__()
    if is_pandas_dataframe(obj) or is_pandas_series(obj):
        return get_pandas()
    if is_modin_dataframe(obj) or is_modin_series(obj):  # pragma: no cover
        return get_modin()
    if is_pyarrow_table(obj) or is_pyarrow_chunked_array(obj):
        return get_pyarrow()
    if is_cudf_dataframe(obj) or is_cudf_series(obj):  # pragma: no cover
        return get_cudf()
    if is_dask_dataframe(obj):  # pragma: no cover
        return get_dask()
    if is_polars_dataframe(obj) or is_polars_lazyframe(obj) or is_polars_series(obj):
        return get_polars()
    msg = f"Could not get native namespace from object of type: {type(obj)}"
    raise TypeError(msg)
```

</p>
</details> 

Adding `__native_namespace__` to `CompliantNamespace` would invalidate all the current (`narwhals`) implementations of the protocol.

For example, `PandasLikeNamespace` would need to define `__native_namespace__`.

<details><summary>Code block</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/78f8c0a28f5e5a11ed57a082e57f071398e9f5ef/narwhals/_pandas_like/namespace.py#L35-L433

</p>
</details> 

I assume you'd need to extend the spec to include `_implementation`, since that would be the way you'd get the native package.

Putting that together, this would be what I imagine `CompliantNamespace` would need to be updated to:

```py
class CompliantNamespace(Protocol, Generic[CompliantSeriesT_co]):
    _implementation: Implementation

    def __native_namespace__(self) -> ModuleType:
        return self._implementation.to_native_namespace()

    def col(self, *column_names: str) -> CompliantExpr[CompliantSeriesT_co]: ...
    def lit(
        self, value: Any, dtype: DType | None
    ) -> CompliantExpr[CompliantSeriesT_co]: ...
```

However, this would only work for these backends which store an `Implementation`:
- `ArrowNamespace`
- `SparkLikeNamespace`
- `PandasLikeNamespace`

---

I do think these are points worth discussing, but I would lean towards saying that changes to the protocols seems a bit beyond the scope here.
I'm just trying to make my IDE happy after-all 😅

narwhals/translate.py
dangotbanned: Yeah moving to `utils` sounds better to me than `dependencies`.

Although the *kind* of checks are similar to those in `dependencies`, here we're interested in `narwhals`-specific methods (not external types to import)

narwhals/typing.py
dangotbanned: Sure thing, resolved in (https://github.com/narwhals-dev/narwhals/pull/1955/commits/fd2a72ef0b519bea1778426eda940d8e0165fb21)

narwhals/translate.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1955/commits/fd2a72ef0b519bea1778426eda940d8e0165fb21)

narwhals/typing.py
dangotbanned: > [!IMPORTANT]
> (https://github.com/narwhals-dev/narwhals/pull/1955/commits/fd2a72ef0b519bea1778426eda940d8e0165fb21) has kept the existing names `has_native_namespace`, `_NativeNamespace`
> I'm still very open to changing them

FBruzzesi: Thanks @dangotbanned , I somehow was giving for granted that all compliant namespaces had a `__native_namespace__` method, and I was absolutely wrong - apologies for the mistake.

Additionally, the function would not necessarily be used only on compliant namespaces, but with generic compliant objects. I see that now. Again sorry for the sloppy review!

The only change I would do then is the following:

```diff
- _NativeNamespace
+ SupportsNativeNamespace 
```

Similarly to the protocol naming that is used in the [typing module](https://docs.python.org/3/library/typing.html#protocols)

dangotbanned: No worries @FBruzzesi, I expected the same as well when I started this PR!

> The only change I would do then is the following

Resolved in (https://github.com/narwhals-dev/narwhals/pull/1955/commits/faf4719850dfed47a057431ab14e24add90b73ad)

dangotbanned:
Moving into draft, found some cases that would benefit from this:

```py
def is_compliant_expr(
    obj: CompliantExpr[CompliantSeriesT_co] | Any,
) -> TypeIs[CompliantExpr[CompliantSeriesT_co]]:
    return hasattr(obj, "__narwhals_expr__")
```

Adding that and then resolving those warnings

MarcoGorelli (approved):
thanks @dangotbanned and @FBruzzesi for reviewing!

narwhals/utils.py
```diff
@@ -1170,3 +1178,29 @@ def check_column_names_are_unique(columns: list[str]) -> None:
...
+    obj: CompliantExpr[CompliantSeriesT_co] | Any,
+) -> TypeIs[CompliantExpr[CompliantSeriesT_co]]:
+    return hasattr(obj, "__narwhals_expr__")
+
+
+def has_native_namespace(obj: Any) -> TypeIs[SupportsNativeNamespace]:
+    return hasattr(obj, "__native_namespace__")
+
+
+def is_dataframe_like(obj: Any) -> TypeIs[DataFrameLike]:
```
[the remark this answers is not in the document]
dangotbanned: That is quite the mouthful @MarcoGorelli 

I can change it if you want?

I usually like having guards all prefixed with `is_` so they're discoverable together. 

Not blocking from me though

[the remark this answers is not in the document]
MarcoGorelli: i just think "dataframe-like" risks looking misleading, as not all dataframes support that. e.g. datafusion doesn't

we plan to move away from the dataframe interchange protocol anyway, some motivation for this is given in https://github.com/narwhals-dev/narwhals/discussions/1657

In fact, given that this is only used in one place, maybe it's fine to keep this particular check inlined?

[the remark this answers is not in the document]
dangotbanned: @MarcoGorelli what about renaming it to whatever you prefer - but adding an underscore prefix? 

You still get the benefits of typing, but leave exposing it/finalizing what it represents until `v2`?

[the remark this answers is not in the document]
dangotbanned: Gone ahead and done both-ish w/ (https://github.com/narwhals-dev/narwhals/pull/1955/commits/dbe40a6ebd36e1a055e787249a8c43f403911a61), in case you just wanna merge *something*

dangotbanned:
@FBruzzesi FYI 😞 https://github.com/narwhals-dev/narwhals/actions/runs/13208298072/job/36876551230?pr=1955

narwhals/_expression_parsing.py
```diff
@@ -123,13 +131,13 @@ def parse_into_expr(
...
-    if hasattr(into_expr, "__narwhals_series__"):
+    if is_compliant_expr(into_expr):
+        return into_expr
+    if is_compliant_series(into_expr):
         return namespace._create_expr_from_series(into_expr)  # type: ignore[no-any-return, attr-defined]
     if is_numpy_array(into_expr):
-        series = namespace._create_compliant_series(into_expr)  # type: ignore[attr-defined]
-        return namespace._create_expr_from_series(series)  # type: ignore[no-any-return, attr-defined]
+        series = namespace._create_compliant_series(into_expr)
+        return namespace._create_expr_from_series(series)
```
FBruzzesi, line 140: Nice! 👌

narwhals/typing.py
```diff
@@ -107,6 +107,10 @@ def lit(
     ) -> CompliantExpr[CompliantSeriesT_co]: ...
 
 
+class SupportsNativeNamespace(Protocol):
+    def __native_namespace__(self) -> Any: ...
```
FBruzzesi, line 111: Would it be too pedantic to suggest:

```py
    def __native_namespace__(self) -> ModuleType: ...
```

FBruzzesi (approved):
I am quite happy with the outcome of this! Thanks @dangotbanned for bearing with my sloppy review! You deserve a medal 🎖️ but I will give you a gif:

WooblesTheWooblesGIF

narwhals/typing.py
dangotbanned: @FBruzzesi go for it!

dangotbanned:
@FBruzzesi @MarcoGorelli 

Am I okay to merge this one?

MarcoGorelli:
good one, thanks!

Issue #1961, opened 2025-02-07
CI: get `mypy` passing with `pyarrow-stubs` installed


MarcoGorelli:
The task is to add `pyarrow-stubs` to the `additional_dependencies` of `mypy` in `.pre-commit-config.yaml`, and fix all the reported issues

Then, this'll make it much easier to address #1962

dangotbanned:
Having played around with `_arrow.series` a bit, I'm thinking you might need to add another `TypeVar` for `ArrowSeries`.

Something like this helps with *some* cases:

```py
from __future__ import annotations

from typing import TYPE_CHECKING


import pyarrow as pa

from narwhals.typing import CompliantSeries

if TYPE_CHECKING:
    from typing import TypeVar

    from narwhals.utils import Version

    _Scalar_co = TypeVar("_Scalar_co", bound=pa.Scalar, covariant=True)

class ArrowSeries(CompliantSeries, Generic["_Scalar_co"]):
    def __init__(
        self: Self,
        native_series: pa.ChunkedArray[_Scalar_co],
        *,
        name: str,
        backend_version: tuple[int, ...],
        version: Version,
    ) -> None:
        self._name = name
        self._native_series: pa.ChunkedArray[_Scalar_co] = native_series
        ...
```

Need to investigate some more though. The stubs seem to suggest certain `pyarrow` and `pyarrow.compute` functions would reject parts of the current API. 
Unsure whether the issue is the annotations of: `pyarrow-stubs`, `narwhals` or a combination of both.

<details><summary>Yelling</summary>
<p>

![Image](https://github.com/user-attachments/assets/4176f5fb-78d9-4f9a-a334-c54036080a04)

</p>
</details>

dangotbanned:
@MarcoGorelli I've got pretty far with this today.
Been working on it locally w/ `pyright`, hopefully will have something worth sharing soon 😅

Would you be able to give a mention of `pyarrow-stubs` in https://discuss.python.org/t/call-for-suggestions-nominate-python-packages-for-typing-improvements/80186?

With a project as big as `pyarrow`, it seems quite a weight on the shoulders of [mainly one contributor]( https://github.com/zen-xu/pyarrow-stubs/graphs/contributors)

MarcoGorelli:
Thank you so much! 

Regarding the discussion, I think it would carry more weight if someone other than me made the request - fancy voicing your support for pyarrow stubs there?

dangotbanned:
[...]

PR #1966, opened 2025-02-08
fix: use `mypy` pre-commit in local environment

### LINKED ISSUE #1962: CI / typing: migrate away from pre-commit?
pre-commit is quite fast and nifty, but it runs everything in isolated virtual environments.

This is quite annoying for running type checking with libraries that export types. We can either keep making the `additional_dependencies` field larger and larger, or we can run `mypy` from the main development environment

If the latter, which I'm increasingly leaning towards, then we need to have a way for lint dependencies to stay perfectly in sync

So, we could have a `requirements-lint.txt` file that pins all of them exactly? And then have a Makefile which has a `make lint` command, which does:
```
uv pip install -r requirements-lint.txt
ruff check
ruff format
mypy  # pyright?
```

Some people have also floated the idea of `uv.lock`. I've not really explored that but am open to it as well

A precursor to this would be to address #1961

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issues #1961 
- Closes #1962

## Checklist

- [x] Tests added
- [x] Documented the changes

I did some digging, looking for how other projects run typing and general best practices with `mypy`.
One thing people seem to agree with is that `mirrors-mypy` has its gotchas and may not be ideal for every project.
In our case we would need to add every backend dependencies as `additional-dependencies`. In this case we would have to maintain a script that keep those dependencies up to date.

One alternative would be to run mypy in the local environment where all the dependencies are already installed. This is what this PR is trying to do.

~I have still some open questions, making this draft to test the CI too.~
~I am expecting to see ~300 errors from `mypy` 👀~

Update: I propose to move to this solution where we:
- keep using `pre-commit` for linting but move `mypy` to run in the local env -> this is to avoid having to add many `additional-dependencies`
- Add Makefile to add 1 consistent command to run typing checks: `make typing`
- Run `mypy` in a separate CI pipeline on the entire code

### DIFF
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -29,7 +29,7 @@ jobs:
         # TODO: add more dependencies/backends incrementally
         run: |
           source .venv/bin/activate
-          uv pip install -e ".[typing, core]"
+          uv pip install -e ".[tests, typing, core]"
       - name: show-deps
         run: |
           source .venv/bin/activate
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -14,10 +14,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
       - name: Install uv
         uses: astral-sh/setup-uv@v5
         with:
-          python-version: ${{ matrix.python-version }}
           enable-cache: "true"
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -33,8 +33,7 @@ jobs:
         run: |
           source .venv/bin/activate
           uv pip freeze
-      - name: Run pytest
+      - name: Run mypy
         run: |
           source .venv/bin/activate
-          which mypy
           make typing
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -13,7 +13,7 @@ jobs:
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     env:
-      MYPY_FORCE_COLOR: "1"
+      FORCE_COLOR: "1"
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -25,7 +25,7 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: Create venv
         run: |
-         uv -m venv .venv
+         uv venv .venv
           source .venv/bin/activate
       - name: install-reqs
         # Python3.8 is technically at end-of-life, so we don't test everything
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -1,0 +1,30 @@
+name: Type checking
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+
+jobs:
+  pytest-38:
+    strategy:
+      matrix:
+        python-version: ["3.11"]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "pyproject.toml"
+      - name: install-reqs
+        # Python3.8 is technically at end-of-life, so we don't test everything
+        run: uv pip install -e ".[dev, core]" --system
+      - name: show-deps
+        run: uv pip freeze
+      - name: Run pytest
+        run: make typing
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -26,9 +26,10 @@ jobs:
       - name: Create venv
         run: uv venv .venv
       - name: install-reqs
+        # TODO: add more dependencies/backends incrementally
         run: |
           source .venv/bin/activate
-          uv pip install -e ".[dev, core]"
+          uv pip install -e ".[dev]" polars
       - name: show-deps
         run: |
           source .venv/bin/activate
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -35,5 +35,6 @@ jobs:
           uv pip freeze
       - name: Run pytest
         run: |
+          source .venv/bin/activate
           which mypy
           make typing
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -24,15 +24,15 @@ jobs:
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: Create venv
-        run: |
-         uv venv .venv
-         source .venv/bin/activate
-         echo PATH=$PATH >> $GITHUB_ENV
+        run: uv venv .venv
       - name: install-reqs
-        # Python3.8 is technically at end-of-life, so we don't test everything
-        run: uv pip install -e ".[dev, core]" --system
+        run: |
+          source .venv/bin/activate
+          uv pip install -e ".[dev, core]"
       - name: show-deps
-        run: uv pip freeze
+        run: |
+          source .venv/bin/activate
+          uv pip freeze
       - name: Run pytest
         run: |
           which mypy
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -29,7 +29,7 @@ jobs:
         # TODO: add more dependencies/backends incrementally
         run: |
           source .venv/bin/activate
-          uv pip install -e ".[dev]" polars
+          uv pip install -e ".[dev, core]"
       - name: show-deps
         run: |
           source .venv/bin/activate
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -26,7 +26,8 @@ jobs:
       - name: Create venv
         run: |
          uv venv .venv
-          source .venv/bin/activate
+         source .venv/bin/activate
+         echo PATH=$PATH >> $GITHUB_ENV
       - name: install-reqs
         # Python3.8 is technically at end-of-life, so we don't test everything
         run: uv pip install -e ".[dev, core]" --system
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -29,4 +29,6 @@ jobs:
       - name: show-deps
         run: uv pip freeze
       - name: Run pytest
-        run: make typing
+        run: |
+          which mypy
+          make typing
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -12,6 +12,8 @@ jobs:
         python-version: ["3.11"]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
+    env:
+      MYPY_FORCE_COLOR: "1"
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -29,7 +29,7 @@ jobs:
         # TODO: add more dependencies/backends incrementally
         run: |
           source .venv/bin/activate
-          uv pip install -e ".[dev, core]"
+          uv pip install -e ".[typing, core]"
       - name: show-deps
         run: |
           source .venv/bin/activate
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -12,8 +12,6 @@ jobs:
         python-version: ["3.11"]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
-    env:
-      FORCE_COLOR: "1"
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -37,5 +35,8 @@ jobs:
           uv pip freeze
       - name: Run mypy
         run: |
+          echo "$FORCE_COLOR"
           source .venv/bin/activate
           make typing
+        env:
+          FORCE_COLOR: "1"
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -6,7 +6,7 @@ on:
     branches: [main]
 
 jobs:
-  pytest-38:
+  mypy:
     strategy:
       matrix:
         python-version: ["3.11"]
@@ -23,6 +23,10 @@ jobs:
           enable-cache: "true"
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
+      - name: Create venv
+        run: |
+         uv -m venv .venv
+          source .venv/bin/activate
       - name: install-reqs
         # Python3.8 is technically at end-of-life, so we don't test everything
         run: uv pip install -e ".[dev, core]" --system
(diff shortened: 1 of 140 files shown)


MarcoGorelli:
i like the look of this, thanks for doing it

it looks like this fails in pre-commit-ci, so we may need to find another way of running it in ci? (i'm not averse to removing pre-commit-ci)

EdAbati:
Thanks @MarcoGorelli ! ❤️

This _could_ be a possible approach that doesn't change super much what we currently have:

- we add `mypy` in the dev dependencies
- `mypy` runs in pre-commit locally
- we use `pyproject.toml` to configure mypy's behaviour
- we remove it from the `pre-commit.ci` (since the local env will not be)
- we add a new GH action for typing
- the `Makefile` is to create a command that works the same way in Linux/macOS and Windows (`make typing`)


The errors are now showing up in the CI!
The dependencies we install in the Typing CI will influence which backends we will check for typing error. (just for this example I just added `core`)

Happy to hear your thoughts (also @FBruzzesi and @dangotbanned )

MarcoGorelli:
nice, happy to go with this - if we can get it green and then gradually address things, happy to get this in

dangotbanned:
> nice, happy to go with this - if we can get it green and then gradually address things, happy to get this in

@MarcoGorelli, @EdAbati 

A big chunk of these would be fixable with `np.ndarray` params:
- https://github.com/numpy/numpy/blob/4ab0dc68c3ccdb5d06bfb6a3070893cd99f937eb/numpy/__init__.pyi#L846-L847


IIRC only 1D/2D are supported so something like:

```py
from typing import Any

if TYPE_CHECKING:

    import numpy as np

SomeName: TypeAlias = "np.ndarray[tuple[int] | tuple[int, int], Any]"
```

MarcoGorelli (approved):
well done getting this to pass

happy to get this in then iterate from here?

.github/workflows/typing.yml
```diff
@@ -0,0 +1,40 @@
...
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "pyproject.toml"
+      - name: Create venv
+        run: uv venv .venv
+      - name: install-reqs
+        # TODO: add more dependencies/backends incrementally
+        run: |
+          source .venv/bin/activate
+          uv pip install -e ".[dev]" polars
```
EdAbati, line 32: Ok if we just have polars as dependency, it's all green.

It is in line with what we have in `main` actually, given that the only 2 `additional-dependencies` are `pytest` and `polars`.

It is a bit too early to merge IMO, given that `pre-commit` will stop any commits from environments that have anything more than polars. I'll try to fix or ignore all errors that comes when `core` is installed.

🚧 _work in progress_ 🚧

EdAbati: upsi @MarcoGorelli you were a faster typer. :D What do you think?

MarcoGorelli: not sure what you mean by

>  pre-commit will stop any commits from environments that have anything more than polars.

could you elaborate pleas?

EdAbati: Given that `mypy` is running in `pre-commit` in `local` "mode", it will check the repo using every dependency one has installed locally.

At the moment, the checks will pass only if you install the dev dependencies with `uv pip install -e ".[dev]" polars`

Already adding an extra dependency (e.g. `pyarrow`) will make `mypy` fail -> `pre-commit` complain

I think we should merge this only after `mypy` passes with at least all the libraries in `core`

MarcoGorelli: ah i see, thanks!

can't we ignore the missing imports from pyarrow in pyproject.toml? like

https://github.com/pola-rs/polars/blob/ff994f0502bc5ce886df43ffbbf468a79adcf281/py-polars/pyproject.toml#L132-L137

EdAbati: That's helping reducing a bit the errors, I also disabled `strict = true`

I tried my best to tell `mypy` to calm down a bit, but it still shows a lot of errors. Does anyone know any trick to tell mypy to ignore errors?
I tried [this](https://mypy.readthedocs.io/en/stable/config_file.html#confval-ignore_errors) but didn't have any effect
 
I guess the alternative is to fix them here. 🥲

EdAbati:
Actually if we'd remove `pyarrow-stubs` for a bit we get waaaay less errors 👀 https://github.com/narwhals-dev/narwhals/pull/1966/commits/b301356d7e71c0bafe4ae63af8682d24b5ff0196

dangotbanned:
> Actually if we removed `pyarrow-stubs` for a bit we'd get waaaay less errors 👀 [b301356](https://github.com/narwhals-dev/narwhals/commit/b301356d7e71c0bafe4ae63af8682d24b5ff0196)

@EdAbati and if you uninstall `mypy` we get 0! 😉 

Seriously though, what is the goal here?

Are we trying to resolve the issues or have *something* in CI?

dangotbanned:
Updating the branch (https://github.com/narwhals-dev/narwhals/pull/1966/commits/c22fd06c6d658625bff24009b4728f77e5650766) to get the past few fixes i've done today

MarcoGorelli:
> @EdAbati and if you uninstall mypy we get 0! 😉

😆 true - tbh i think that if we can get it running with an exclusions list (e.g. we initially ignore pyarrow-stubs, pandas, etc...), then we can gradually remove one exclusion at a time

having said that, we may want to pin dependencies in the CI job that runs mypy, else it risks being frustrating for contributors (especially newcomers)

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/1966#issuecomment-2645909580

@MarcoGorelli sounds reasonable 

Maybe this is a dumb question, but what is the issue with running `mypy` without `pre-commit`?
To me it seems the simplest way to get the same errors we'd see locally

MarcoGorelli:
that's fine as well. seeing as we're using pre-commit for other checks, i don't it hurt to have a `make typing` so that the workflow for anyone using pre-commit isn't affected

EdAbati:
> Seriously though, what is the goal here?

My goal here was trying to find a better and more consistent way to run typing tests without disrupting too much the current workflow.
Here we keep `mypy` in `pre-commit` and we run `mypy` on all the relevant code in CI (or manually using `make`) without pre-commit

> Actually if we removed pyarrow-stubs for a bit we'd get waaaay less errors 👀 https://github.com/narwhals-dev/narwhals/commit/b301356d7e71c0bafe4ae63af8682d24b5ff0196

Here I meant that it seems that most of typing errors (we are not already ignoring) come from the `pyarrow` related part of the code. Of course not suggesting to remove this stubs forever 😅
My plan/suggestion is to make the CI green with at least our `core` dependencies installed and then start reintroducing stubs and removing exclusions in follow-up PRs

Still testing a couple of things, hope to have this PR properly ready for review later today 🤞

narwhals/_arrow/series.py
```diff
@@ -339,7 +339,7 @@ def __getitem__(self: Self, idx: int | slice | Sequence[int]) -> Any | Self:
     def scatter(self: Self, indices: int | Sequence[int], values: Any) -> Self:
         import numpy as np  # ignore-banned-import
 
-        mask = np.zeros(self.len(), dtype=bool)
+        mask: np.ndarray = np.zeros(self.len(), dtype=bool)
```
dangotbanned, line 342: @EdAbati maybe this PR could utilize #1977?

EdAbati: definitely!

narwhals/_duckdb/expr.py
```diff
@@ -33,7 +33,7 @@
     from narwhals.utils import Version
 
 
-class DuckDBExpr(CompliantExpr["duckdb.Expression"]):
+class DuckDBExpr(CompliantExpr["duckdb.Expression"]):  # type: ignore[type-var]
```
EdAbati, line 36: This ignore should actually be solved. `duckdb.Expression` is not a `CompatibleSeries`

Is it ok for a follow-up?

narwhals/_duckdb/dataframe.py
```diff
@@ -159,7 +159,7 @@ def select(
         ):
             return self._from_native_frame(
                 self._native_frame.aggregate(
-                    [val.alias(col) for col, val in new_columns_map.items()]
+                    [val.alias(col) for col, val in new_columns_map.items()]  # type: ignore[arg-type]
```
EdAbati, line 162: based on the duckdb stubs, only `str` are allowed in `.aggregate`. 🤔

pyproject.toml
```diff
@@ -37,22 +37,31 @@ polars = ["polars>=0.20.3"]
...
   "typing_extensions",
 ]
+typing = [
+  "typing_extensions",
+  "mypy~=1.15.0",
+]
+dev = [
+  "pre-commit",
+  "narwhals[tests]",
+  "narwhals[typing]",
```
EdAbati, line 56: I think we should separate the dev dependencies into `tests` and `typing`.

The latest pinned `mypy` is available for python >=3.9 and if we simply include it in `dev` our python 3.8 tests will fail.

narwhals/_duckdb/expr_dt.py
EdAbati, line 1: duckdb stubs don't seem to allow these operations. I have a feeling the stubs may be outdated

narwhals/_pandas_like/dataframe.py
```diff
@@ -58,7 +58,7 @@
 from narwhals.typing import CompliantDataFrame
 from narwhals.typing import CompliantLazyFrame
 
-CLASSICAL_NUMPY_DTYPES = frozenset(
+CLASSICAL_NUMPY_DTYPES: frozenset[np.dtype] = frozenset(
```
dangotbanned, line 61: You might be able to get away with this?

Possibly an issue for `3.8` though
```suggestion
CLASSICAL_NUMPY_DTYPES = frozenset["np.dtype"](
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -258,8 +258,8 @@ def __getitem__(
 
         elif isinstance(item, tuple) and len(item) == 2:
             if isinstance(item[1], str):
-                item = (item[0], self._native_frame.columns.get_loc(item[1]))  # pyright: ignore[reportAssignmentType]
```
dangotbanned, line 261: Just want to double check, can you confirm `pyright` is happy here?

I know it isn't on CI and at a glance, this seems reasonable, just wanna be sure.
IIRC I only added the comment today 😅

dangotbanned: All good

narwhals/_duckdb/dataframe.py
MarcoGorelli: looks wrong, it should be `str | list[Expression]`, right? want to make a PR to duckdb?

narwhals/_duckdb/expr.py
MarcoGorelli: yup definitely

dangotbanned:
@MarcoGorelli could you take a look at the merge conflict here please?

Just trying to not break #1918 in the process 🙂

narwhals/_duckdb/expr_dt.py
```diff
@@ -60,23 +60,23 @@ def second(self: Self) -> DuckDBExpr:
     def millisecond(self: Self) -> DuckDBExpr:
         return self._compliant_expr._from_call(
             lambda _input: FunctionExpression("millisecond", _input)
-            - FunctionExpression("second", _input) * 1_000,
```
MarcoGorelli, line 63: I've got a feeling they want to nudge us towards using `* ConstantExpression(1000)`, then mypy accepts that

narwhals/_duckdb/expr_dt.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1966/commits/bf9e0d992cad94e002e8de4bc4d237f4b7a3386c)

narwhals/_duckdb/expr_dt.py
dangotbanned: Yeah that was what I went for in (https://github.com/narwhals-dev/narwhals/pull/1966/commits/bf9e0d992cad94e002e8de4bc4d237f4b7a3386c)

dangotbanned: @MarcoGorelli to clarify, that was only on `_duckdb.expr` (not this module)

narwhals/_duckdb/expr.py
```diff
@@ -440,7 +444,11 @@ def _std(_input: duckdb.Expression, ddof: int) -> duckdb.Expression:
     def var(self: Self, ddof: int) -> Self:
         def _var(_input: duckdb.Expression, ddof: int) -> duckdb.Expression:
             n_samples = FunctionExpression("count", _input)
-            return FunctionExpression("var_pop", _input) * n_samples / (n_samples - ddof)
+            return (
+                FunctionExpression("var_pop", _input)
+                * n_samples
+                / (n_samples - ConstantExpression(ddof))
+            )
```
dangotbanned, line 451: I was just fixing a `pyright` warning here.

If anyone has a meaningful name for the parts of this expression, I think it'd be more readable with an assignment *first*

dangotbanned: Like:
```py

n_samples = n_samples = FunctionExpression("count", _input)
meaningful = (n_samples - ConstantExpression(ddof))
return FunctionExpression("var_pop", _input) * n_samples / meaningful

dangotbanned:
The last warning seems legit to me https://github.com/narwhals-dev/narwhals/actions/runs/13240255809/job/36953842381?pr=1966

Both uses on `bin_count` pass `None` as arguments to functions that don't accept `None`:

image

https://github.com/narwhals-dev/narwhals/blob/4b1d778beeab20c3b5565dba629e320eb23804c8/narwhals/_pandas_like/series.py#L1057-L1067

narwhals/_duckdb/expr_dt.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/1966/commits/5306ce2e31f2d547e9799bf3fa0fe4d2e0f1c12a) and (https://github.com/narwhals-dev/narwhals/pull/1966/commits/6fc845eddfaed5b6cbcae3475e8a3036b5ac61e3)

dangotbanned:
@MarcoGorelli I'm reasonably happy with where the typing is at after (https://github.com/narwhals-dev/narwhals/pull/1966#commits-pushed-5306ce2)

Of these modules, `narwhals._arrow.series` is the only one left with major issues:

<details><summary>pyright shouting at me</summary>
<p>

image


</p>
</details> 

Will try to look into those after this and #1979 have merged

MarcoGorelli (approved):
thanks both! sure, happy to move forwards and address the rest incrementally, much appreciate your efforts here!

camriddell:
@dangotbanned seems like you resolved everything there, thanks!

If it is useful to you in a post-hoc sense

https://github.com/narwhals-dev/narwhals/blob/60e5705d5d219f609de002b14b3d423451c7b924/narwhals/series.py#L2591-L2601

The high-level `narwhals.Series` entrypoint controls over uses runtime determination of `bins` or `bin_count` to determine the behavior of the function. The following rules should be respected: 

1. One cannot pass both `bins` and `bin_count` in as it leads to ambiguous behavior
2. If one is passed in, then it should be treated as its non-`None` type; the other should be set to `None`.
3. If neither are passed it, then it is treated as `bin_count=10` and `bins=None`

I decided to split broad smoke-tests into 2 separate functions:
- `bins=...` is defined, `bin_count=None` https://github.com/narwhals-dev/narwhals/blob/4b1d778beeab20c3b5565dba629e320eb23804c8/tests/series_only/hist_test.py#L84
- `bins=None`, `bin_count=...` is defined https://github.com/narwhals-dev/narwhals/blob/4b1d778beeab20c3b5565dba629e320eb23804c8/tests/series_only/hist_test.py#L166

---

Thanks for resolving all of these typing issue. In the future, it could be worth refactoring these two behaviors into separate functions for each backend and dispatch the correct call from the `narwhals.Series`. I think this would make the internal code make a lot more sense as well.

dangotbanned (approved):

Issue #1848, opened 2025-01-21
refactor: pass `ExprMetadata` down to compliant Exprs from narwhals.Expr


MarcoGorelli:
Currently, we track expression metadata at the compliant level:
- whether it aggregates, changes length, transforms
- the function name

This is quite error-prone, as we repeat logic in quite a few places.

I'd prefer it if:
- we didn't track the function name at all for most backends. It's only really needed for pandas/pyarrow/dask because of limitations in their group-by APIs, so we should only do it there
- one of the metadatas we could track is what kind of column selection we start with (`nth`, `col`, `selector`, `all`). Because in group_by-agg, `selector` and `all` require some special treatment. But this could be done much more simply at the Narwhals level, without tracking function names everywhere in DuckDB / PySpark / Ibis / whatever else we add, which I hope would also have a modern enough syntax

To do this, I haven't tried, but perhaps:
- `_to_compliant_expr` could take a `metadata` argument, which takes a `TypeDict` of metadata which we keep track of at the `narwhals/expr.py` and `narwhals/functions.py` level
- compliant expressions just read from the metadata passed by above to determine how and whether to broadcast

dangotbanned:
Think I've come across this in https://github.com/narwhals-dev/narwhals/pull/2149#discussion_r1986103446

For example, in `Expr.null_count` we know that it is `ExprKind.AGGREGATION`.
But we repeat that as `returns_scalar=True` and then use that as a conditional in multiple places.


https://github.com/narwhals-dev/narwhals/blob/1e45855366fd2a0bb36f17c08e12be89ebc2a125/narwhals/expr.py#L1623-L1652

Gets even trickier with `ArrowExpr` as that turns it into `_return_py_scalar` for `ArrowSeries`

https://github.com/narwhals-dev/narwhals/blob/1e45855366fd2a0bb36f17c08e12be89ebc2a125/narwhals/_arrow/series.py#L532-L534

Eventually resolving it in `maybe_extract_py_scalar`

https://github.com/narwhals-dev/narwhals/blob/1e45855366fd2a0bb36f17c08e12be89ebc2a125/narwhals/_arrow/series.py#L91-L96

dangotbanned:
I know things have changed somewhat since the issue has opened - I think some window functions are already passing `kind` down?

MarcoGorelli:
yup - but you're right, we might be able to use this for the pyarrow scalar stuff too

MarcoGorelli:
[...]

PR #1986, opened 2025-02-10
chore: Expr metadata refactor

MarcoGorelli:
First step towards #1848 

No noticeable difference in perf when running this locally

Theoretically, I'd expect it to make things not-worse, as we're not only looping through `args` once to construct `ExprMetadata`, as opposed to 4 times

But the main reason to do this is to aim towards #1848 , after which, implementations at the compliant level should become much simpler and easier to reason about

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -392,40 +392,45 @@ def combine_metadata(*args: IntoExpr, strings_are_column_names: bool) -> ExprMet
     from narwhals.expr import Expr
 
     n_changes_length = 0
-    n_transforms = 0
-    n_aggregations = 0
-    n_literals = 0
+    has_transforms = False
+    has_aggregations = False
+    has_literals = False
     result_is_order_dependent = False
 
     for arg in args:
         if isinstance(arg, str) and strings_are_column_names:
-            n_transforms += 1
+            has_transforms |= True
         elif isinstance(arg, Expr):
             if arg._metadata["is_order_dependent"]:
                 result_is_order_dependent = True
             kind = arg._metadata["kind"]
             if kind is ExprKind.AGGREGATION:
-                n_aggregations += 1
+                has_aggregations = True
             elif kind is ExprKind.LITERAL:
-                n_literals += 1
+                has_literals = True
             elif kind is ExprKind.CHANGES_LENGTH:
                 n_changes_length += 1
             elif kind is ExprKind.TRANSFORM:
-                n_transforms += 1
+                has_transforms = True
             else:  # pragma: no cover
                 msg = "unreachable code"
                 raise AssertionError(msg)
-    if n_literals and not n_aggregations and not n_transforms and not n_changes_length:
+    if (
+        has_literals
+        and not has_aggregations
+        and not has_transforms
+        and not n_changes_length
+    ):
         result_kind = ExprKind.LITERAL
     elif n_changes_length > 1:
         msg = "Length-changing expressions can only be used in isolation, or followed by an aggregation"
         raise LengthChangingExprError(msg)
-    elif n_changes_length and n_transforms:
+    elif n_changes_length and has_transforms:
         msg = "Cannot combine length-changing expressions with length-preserving ones"
         raise ShapeError(msg)
     elif n_changes_length:
         result_kind = ExprKind.CHANGES_LENGTH
-    elif n_transforms:
+    elif has_transforms:
         result_kind = ExprKind.TRANSFORM
     else:
         result_kind = ExprKind.AGGREGATION
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -322,14 +322,14 @@ def extract_compliant(
     plx: CompliantNamespace[CompliantSeriesT_co],
     other: Any,
     *,
-    parse_column_name_as_expr: bool,
+    strings_are_column_names: bool,
 ) -> CompliantExpr[CompliantSeriesT_co] | CompliantSeriesT_co | Any:
     from narwhals.expr import Expr
     from narwhals.series import Series
 
     if isinstance(other, Expr):
         return other._to_compliant_expr(plx)
-    if parse_column_name_as_expr and isinstance(other, str):
+    if strings_are_column_names and isinstance(other, str):
         return plx.col(other)
     if isinstance(other, Series):
         return other._compliant_series
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -14,6 +14,7 @@ from typing import overload
 from narwhals.dependencies import is_numpy_array
 from narwhals.exceptions import InvalidIntoExprError
 from narwhals.exceptions import LengthChangingExprError
+from narwhals.utils import ExprKind
 from narwhals.utils import Implementation
 from narwhals.utils import is_compliant_expr
 from narwhals.utils import is_compliant_series
@@ -359,7 +360,10 @@ def operation_changes_length(*args: IntoExpr | Any) -> bool:
     from narwhals.expr import Expr
 
     n_exprs = len([x for x in args if isinstance(x, Expr)])
-    changes_length = any(isinstance(x, Expr) and x._changes_length for x in args)
+    changes_length = any(
+        isinstance(x, Expr) and x._metadata["kind"] is ExprKind.CHANGES_LENGTH
+        for x in args
+    )
     if n_exprs > 1 and changes_length:
         msg = (
             "Found multiple expressions at least one of which changes length.\n"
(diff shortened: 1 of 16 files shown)


narwhals/_expression_parsing.py
```diff
@@ -399,3 +357,105 @@ def evaluate_output_names_and_aliases(
...
+    n_aggregations = 0
+    n_literals = 0
+    is_order_dependent = False
+
+    for arg in args:
+        if isinstance(arg, str) and strings_are_column_names:
+            n_transforms += 1
+        elif isinstance(arg, Expr):
+            if arg._metadata["is_order_dependent"]:
+                is_order_dependent = True
```
dangotbanned, line 405: minor nit 😉

```suggestion
            is_order_dependent = arg._metadata["is_order_dependent"]
```

Liking the switch to `Enum` btw

MarcoGorelli: yeah, if any expression there is order-dependent, then the result expression is order-dependent

narwhals/_expression_parsing.py
```diff
@@ -399,3 +357,105 @@ def evaluate_output_names_and_aliases(
...
+                n_aggregations += 1
+            elif kind is ExprKind.LITERAL:
+                n_literals += 1
+            elif kind is ExprKind.CHANGES_LENGTH:
+                n_changes_length += 1
+            elif kind is ExprKind.TRANSFORM:
+                n_transforms += 1
+            else:  # pragma: no cover
+                msg = "unreachable code"
+                raise AssertionError(msg)
```
FBruzzesi, line 417: What do you think about a less branching approach by first partitioning into strings and expressions and then using a counter:

```suggestion
    strings = [arg for arg in args if isinstance(arg, str) and strings_are_column_names]
    exprs = [arg for arg in args if isinstance(arg, Expr)]
    
    result_is_order_dependent = any(expr._metadata["is_order_dependent"] for expr in exprs)
    counter = Counter((expr._metadata["kind"] for expr in exprs))
    
    n_changes_length = counter[ExprKind.CHANGES_LENGTH]
    n_transforms = counter[ExprKind.TRANSFORM] + strings_are_column_names * len(strings)
    n_aggregations = counter[ExprKind.AGGREGATION]
    n_literals = counter[ExprKind.LITERAL]   
```

FBruzzesi (commented):
I love this! I will take a closer look later today! I left one comment because Dan comment draw my attention to such code snippet 🙈

narwhals/_expression_parsing.py
dangotbanned: @FBruzzesi one thing stood out from the description

> Theoretically, I'd expect it to make things not-worse, as we're not only looping through args once to construct ExprMetadata, as opposed to 4 times

I do think it is *visually* nicer not branching so much, but you can probably reduce that without reverting to the 4x loops

FBruzzesi: Fair enough, it can probably go down to a single loop + counter, which would be more than what's already here.
Nevermind then

narwhals/expr.py
```diff
@@ -466,9 +389,10 @@ def any(self: Self) -> Self:
         """
         return self.__class__(
             lambda plx: self._to_compliant_expr(plx).any(),
-            is_order_dependent=self._is_order_dependent,
-            changes_length=False,
-            aggregates=True,
+            ExprMetadata(
+                kind=ExprKind.AGGREGATION,
+                is_order_dependent=self._metadata["is_order_dependent"],
```
FBruzzesi, line 394: Is this expected? Doesn't this aggregation set `is_order_dependent=False`? Similar for other aggregations

narwhals/functions.py
```diff
@@ -1179,12 +1178,7 @@ def sum(*columns: str) -> Expr:
         |    0  3  4.8     |
         └──────────────────┘
     """
-    return Expr(
-        lambda plx: plx.col(*columns).sum(),
-        is_order_dependent=False,
-        changes_length=False,
-        aggregates=True,
-    )
+    return col(*columns).sum()
```
FBruzzesi, line 1181: 🤩

narwhals/_expression_parsing.py
FBruzzesi: Actually, coming back to this and keeping a single loop, we could do:

```py
def combine_metadata(*args: IntoExpr, strings_are_column_names: bool) -> ExprMetadata:
    # combine metadata from `args`

    counts = Counter()
    result_is_order_dependent = False

    for arg in args:
        if isinstance(arg, str) and strings_are_column_names:
            counts[ExprKind.TRANSFORM] += 1
        elif isinstance(arg, Expr):
            if arg._metadata["is_order_dependent"]:
                result_is_order_dependent = True
            counts[arg._metadata["kind"]] += 1

    if (
        counts[ExprKind.LITERAL]
        and not counts[ExprKind.AGGREGATION]
        and not counts[ExprKind.TRANSFORM]
        and not counts[ExprKind.CHANGES_LENGTH]
    ):
        return ExprMetadata(kind=ExprKind.LITERAL, is_order_dependent=result_is_order_dependent)

    if counts[ExprKind.CHANGES_LENGTH] > 1:
        msg = "Length-changing expressions can only be used in isolation, or followed by an aggregation"
        raise LengthChangingExprError(msg)
    if counts[ExprKind.CHANGES_LENGTH] and counts[ExprKind.TRANSFORM]:
        msg = "Cannot combine length-changing expressions with length-preserving ones"
        raise ShapeError(msg)

    if counts[ExprKind.CHANGES_LENGTH]:
        result_kind = ExprKind.CHANGES_LENGTH
    elif counts[ExprKind.TRANSFORM]:
        result_kind = ExprKind.TRANSFORM
    else:
        result_kind = ExprKind.AGGREGATION

    return ExprMetadata(kind=result_kind, is_order_dependent=result_is_order_dependent)
```

FBruzzesi (commented):
Thanks @MarcoGorelli I gave it a deeper look! It seems very promising for the future.

Question: should we re-use `ExprKind` from `_expression_parsing` in spark and duckdb already or is it too premature (We define `ExprKind` in 3 places)?

narwhals/functions.py
MarcoGorelli: 😄 not sure why i didn't write it like this in the first place 😳

narwhals/expr.py
MarcoGorelli: if we do `nw.col('a').cum_sum().mean()`, then the entire expression is order-dependent (as it depends on the order in which you evaluate the cumulative sum)

narwhals/_expression_parsing.py
MarcoGorelli: thanks - it's occurred to me that most of these can just be booleans rather than counts, it's just `changes_length` that needs counting, so i've rewritten to just use booleans

MarcoGorelli:
thanks for your review!

yes the ultimate goal is to remove `ExprKind` from `_pyspark` and `_duckdb`, as well as `aggregate` / `returns_scalar` from `_dask` - we'll get there 😎

narwhals/_expression_parsing.py
```diff
@@ -392,40 +392,45 @@ def combine_metadata(*args: IntoExpr, strings_are_column_names: bool) -> ExprMet
...
-    n_literals = 0
+    has_transforms = False
+    has_aggregations = False
+    has_literals = False
     result_is_order_dependent = False
 
     for arg in args:
         if isinstance(arg, str) and strings_are_column_names:
-            n_transforms += 1
+            has_transforms |= True
```
FBruzzesi, line 402: Possibly nitpicking but I recently discovered that `|` has quite some overhead over `or` for booleans. Also we can probably just set `has_transforms=True` at this point

dangotbanned: @FBruzzesi that's because `or` will short-circuit, but `.__or__` has to evaluate both sides

FBruzzesi: yes exactly👌

MarcoGorelli: thanks!

narwhals/expr.py
```diff
@@ -1792,10 +1691,7 @@ def null_count(self: Self) -> Self:
             └──────────────────┘
         """
         return self.__class__(
-            lambda plx: self._to_compliant_expr(plx).null_count(),
-            self._is_order_dependent,
-            changes_length=self._changes_length,
-            aggregates=self._aggregates,
```
MarcoGorelli, line 1798: this one wasn't right to begin with, `null_count` aggregates

will fix tomorrow

PR #1997, opened 2025-02-11
chore(typing): add overloads to `isinstance_or_issubclass`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #1961

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- I've typed only up to 3-tuple, based on existing usage of the function
- Noticed this was needed to resolve a lot of issues in `_arrow.utils`

### Example (3-tuple)
image

https://github.com/narwhals-dev/narwhals/blob/fd8ccac0c66af1a459c8a1ab0954018f91eb172a/narwhals/_pandas_like/utils.py#L653

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -13,6 +13,7 @@ from typing import Sequence
 from typing import TypeVar
 from typing import Union
 from typing import cast
+from typing import overload
 from warnings import warn
 
 from narwhals.dependencies import get_cudf
@@ -64,6 +65,9 @@ if TYPE_CHECKING:
         "FrameOrSeriesT", bound=Union[LazyFrame[Any], DataFrame[Any], Series[Any]]
     )
     _T = TypeVar("_T")
+    _T1 = TypeVar("_T1")
+    _T2 = TypeVar("_T2")
+    _T3 = TypeVar("_T3")
 
     class _SupportsVersion(Protocol):
         __version__: str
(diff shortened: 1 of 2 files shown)


dangotbanned:
New failures are actually a positive outcome 😉 

https://github.com/narwhals-dev/narwhals/actions/runs/13272633847/job/37055430743?pr=1997

narwhals/utils.py
```diff
@@ -474,7 +478,49 @@ def parse_version(version: str | ModuleType | _SupportsVersion) -> tuple[int, ..
...
+@overload
+def isinstance_or_issubclass(
+    obj_or_cls: Any, cls_or_tuple: tuple[type, ...]
+) -> TypeIs[Any]: ...
+
+
+def isinstance_or_issubclass(obj_or_cls: Any, cls_or_tuple: Any) -> bool:
     from narwhals.dtypes import DType
 
     if isinstance(obj_or_cls, DType):
```
dangotbanned, line 526: I was thinking about adding another `@overload` (as the first in order) for this case.

Would be require *another `TypeVar`:
```py
_DT = TypeVar("_DT", bound=DType)

@overload
def isinstance_or_issubclass(obj_or_cls: _DT, cls_or_tuple: type[_DT]) -> TypeIs[_DT]: ...
```

But AFAICT, there aren't any issues this would resolve.
All cases seem to be `DType | type[DType]`, so it wouldn't get triggered anywhere (statically)

MarcoGorelli (approved):
my brain just exploded 🤯 

thanks @dangotbanned !

dangotbanned:
> my brain just exploded 🤯
> 
> thanks @dangotbanned !

Haha thanks @MarcoGorelli!

For reference, what I've done here is similar to how `builtins` like `map` are typed:
https://github.com/python/typeshed/blob/a410f251ee535950eb4440ad41ae4c00c87e6a67/stdlib/builtins.pyi#L1490-L1534

But you have the additional requirement to type `T | type[T]`.
Also the order of the `@overload`(s) are important because `object` *also* will match `type`

PR #2000, opened 2025-02-12
feat: `IbisLazyFrame` support

### LINKED ISSUE #1886: Request for contributions: Ibis support
I'm quite confident about how we're supporting PySpark DuckDB, and I think it's time to support Ibis too.

I think that it can probably mostly be supported by just copy-and-pasting what's in `narwhals/_duckdb` and making some API adjustments. I don't think any major refactor or anything like that would be required.

Anyone fancy trying this? If so, contributions would be welcome! You can start by reading the [contributing guide](https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md), and reaching out to us here or on Discord if anything trips you up

rwhitten577:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1886 

## Checklist

- [x] Tests added
- [ ] Documented the changes


NickCrews:
I want ibis support in narwhals and can help to get this landed. Ping me if you want a review or help!

Cc for visibility @cpcloud, the maintainer of ibis

MarcoGorelli:
thanks @NickCrews for your help! sure - we're in the middle of some large refactors so it may be prudent to wait a bit to avoid too many merge conflicts, but we will get to this

dangotbanned:
## `ibis.selectors`
I'm interested in how we might be able to take advantage of [`ibis.selectors`](https://github.com/ibis-project/ibis/blob/1045dbea46eacd86cfae5208aa7e6ef05b0312ca/ibis/selectors.py).
I noted in (https://github.com/narwhals-dev/narwhals/pull/2064#issuecomment-2676930181) `ibis` is an outlier because of their [native support](https://ibis-project.org/reference/selectors)

Initial thoughts are that our adaptation *could* work quite differently to the other backends - *maybe* even being more performant

Since (#2064) every backend *besides* `polars` is sharing a lot of code - which recently moved to [`nw._compliant.selectors`](https://github.com/narwhals-dev/narwhals/blob/af20c95c1e35e3dd4183b7d001ea69fd142d718b/narwhals/_compliant/selectors.py).
Most of this is simply performing operations on [`nw.Schema`](https://narwhals-dev.github.io/narwhals/api-reference/schema/)

MarcoGorelli:
Hey @rwhitten577 - we're done with the big refactors, so if it interests you to continue, we'd love to ship this 🚢 🚀 

I'd suggest not worrying about doing anything different for selectors just yet (it looks to me like they call `.schema` anyway?), mirroring what we do for duckdb / sqlframe / pyspark should be fine 👍

MarcoGorelli:
yup,that's right! we'd just translate to the ibis api, so whatever backend is set there would be used

rwhitten577:
Hey @MarcoGorelli, I will try to pick this back up in the next few weeks. I haven't looked through all the refactor changes yet, but do you think I'd be better off rebasing this old branch or are the changes significant enough where I'd want to start over and use duckdb or spark as an example?

MarcoGorelli:
awesome, thanks @rwhitten577 !

i think it might be salvageable to continue from here, but still using `_spark_like` / `_duckdb` as a reference. looks like most of the merge conflicts are in the tests anyway

rwhitten577:
@MarcoGorelli this should be ready for review, all checks passing. 

How do you typically handle backward compatibility with older versions of a library? E.g., the Ibis code used is based on the latest 10.x.x API but Narwhals currently allows >=6.x.x.

dangotbanned:
Thanks for the PR @rwhitten577!

> How do you typically handle backward compatibility with older versions of a library? E.g., the Ibis code used is based on the latest 10.x.x API but Narwhals currently allows >=6.x.x.

This PR has all the current backwards-compat we have per-backend.
I might mark that for review to make this one easier?

- #2371

Do you know which features you've used so far, that are missing in https://github.com/ibis-project/ibis/tree/6.0.0?

pyproject.toml
```diff
@@ -279,7 +280,6 @@ module = [
   "dask.*",
   "dask_expr.*",
   "duckdb.*",
-  "ibis.*",
```
dangotbanned, line 282: Gotta say I'm pretty surprised `ibis` doesn't have a `py.typed` marker 🤔 

- https://github.com/narwhals-dev/narwhals/actions/runs/14491681390/job/40649359983?pr=2000#step:8:33

They use typing pretty heavily all through their API?

narwhals/_ibis/utils.py
```diff
@@ -0,0 +1,180 @@
...
+            for field in dtype.fields
+        ]
+        return ibis_dtypes.Struct.from_tuples(fields)
+    if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover
+        inner = narwhals_to_native_dtype(dtype.inner, version)
+        return ibis_dtypes.Array(value_type=inner, length=dtype.size)
+    if isinstance_or_issubclass(dtype, dtypes.Binary):
+        return ibis_dtypes.Binary()
+    msg = f"Unknown dtype: {dtype}"  # pragma: no cover
+    raise AssertionError(msg)
```
dangotbanned, line 180: It might be worth adding a (*still failing*) branch for `dtypes.Enum` and linking this issue in a comment:

- [ ] https://github.com/ibis-project/ibis/issues/10991

pyproject.toml
dangotbanned: Well, they added one in 

- https://github.com/ibis-project/ibis/issues/5279

And then removed it following
- https://github.com/ibis-project/ibis/issues/6844

I spotted a wild @binste! 😉

dangotbanned: Reverted in (https://github.com/narwhals-dev/narwhals/pull/2000/commits/fdbd3a83f650550198a72a34dd8cf0888fe4cc1c)

MarcoGorelli: yeah, bit disappointing tbh

dangotbanned:
> awesome, thanks @rwhitten577 ! gonna see if we can ship this relatively soon before merge conflicts creep in, and make issue for any follow-ups that might need doing

@MarcoGorelli for my sanity, can we try to reduce these first? 🙏 🙏 🙏
We're at a little over 60 new `pyright` warnings 😨 
- https://github.com/narwhals-dev/narwhals/actions/runs/14493247430/job/40654425110?pr=2000

rwhitten577:
> Thanks for the PR @rwhitten577!
> 
> > How do you typically handle backward compatibility with older versions of a library? E.g., the Ibis code used is based on the latest 10.x.x API but Narwhals currently allows >=6.x.x.
> 
> This PR has all the current backwards-compat we have per-backend. I might mark that for review to make this one easier?
> 
> * [chore: `@requires` for backend method constraints #2371](https://github.com/narwhals-dev/narwhals/pull/2371)
> 
> Do you know which features you've used so far, that are missing in https://github.com/ibis-project/ibis/tree/6.0.0?

Thanks @dangotbanned, I'm sure there are a lot of API changes in their past 4 major releases we'd have to go through. The main thing I ran into going from 9.x to 10.x is they deprecated `ibis.case(...).when...` and switched to `ibis.cases`. I used the latter but could add branches to include `ibis.case` for older versions if needed

rwhitten577:
I can try to address the pyright errors today

MarcoGorelli:
> The main thing I ran into going from 9.x to 10.x is they deprecated ibis.case(...).when... and switched to ibis.cases. I used the latter but could add branches to include ibis.case for older versions if needed

if we're adding initial support, i'd advocate for keeping it simple and just raising if `self._backend_version < (0,10)` for some operations, we can always lower the minimum version later

MarcoGorelli:
yes you could update

https://github.com/narwhals-dev/narwhals/blob/aa48faae5ebaf36d0efd39cc1b0267a4502ee51a/utils/import_check.py#L21-L27

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,707 @@
...
+    from narwhals._ibis.namespace import IbisNamespace
+    from narwhals._ibis.typing import WindowFunction
+    from narwhals.dtypes import DType
+    from narwhals.typing import RankMethod
+    from narwhals.typing import RollingInterpolationMethod
+    from narwhals.utils import Version
+    from narwhals.utils import _FullContext
+
+
+class IbisExpr(LazyExpr["IbisLazyFrame", "ir.Expr"]):
```
dangotbanned, line 42: Using `ir.Expr` seems to be the main source of typing issues.

I was initially thinking of changing the definition of `NativeExpr`

https://github.com/narwhals-dev/narwhals/blob/7059f5b77a0e57b5ec95c688f2072d4134bd4858/narwhals/_compliant/expr.py#L80-L89


But most of the methods being called don't exist on `ir.Expr`

https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/core.py#L67-L920

dangotbanned: E.g. here, all other the methods being called on `expr` are referring to something other than `ir.Expr`?

https://github.com/rwhitten577/narwhals/blob/6e44a8b54c331f6056580a79f2bedf950e78d825/narwhals/_ibis/expr.py#L136-L152

rwhitten577: Yeah I think the Ibis types reported are going to be specific e.g. `IntegerColumn` or `IntegerValue`. Will explore a TypeAlias that covers these

dangotbanned: Seems like we might want to use one (or more) of `Value`, `Scalar`, `Column`:

- https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/generic.py#L39
- https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/generic.py#L1368
- https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/generic.py#L1487

dangotbanned: > Yeah I think the Ibis types reported are going to be specific e.g. `IntegerColumn` or `IntegerValue`. Will explore a TypeAlias that covers these

Right I see, they have things quite a bit more strongly typed.
With methods split up by class - rather than a unified `Expr` or `Column` or `Series` with namespace accessors 🤔

NickCrews: Watch out as well, there is a difference between ir.IntegerScalar and ir.IntegerColumn (both of which inherit from ir.IntegerValue, which is where methods like `.abs()` are implemented): The Column type has aggregation functions such as `.max()`, but the Scalar one does not. Similar for FloatingScalar/FloatingColumn, etc.

dangotbanned: Thanks @NickCrews!

@MarcoGorelli we may need to consider abstracting this out - since IIUC we might run into `AttributeError`s if we aren't careful

dangotbanned: Or at least coming up with some common patterns to follow, where we block certain ops and raise consistent errors

dangotbanned: A basic solution would be like like for `ArrowSeries.median`

https://github.com/narwhals-dev/narwhals/blob/7059f5b77a0e57b5ec95c688f2072d4134bd4858/narwhals/_arrow/series.py#L322-L331

Not sure what kind of overhead this might have if we need to check frequently

MarcoGorelli: > @MarcoGorelli we may need to consider abstracting this out - since IIUC we might run into AttributeErrors if we aren't careful

we already don't allow aggregating scalar-like expressions
```python
In [2]: nw.col('a').mean().mean()
InvalidOperationError                     Traceback (most recent call last)
Cell In[2], line 1
----> 1 nw.col('a').mean().mean()

File ~/scratch/.venv/lib/python3.12/site-packages/narwhals/expr.py:560, in Expr.mean(self)
    541 def mean(self: Self) -> Self:
    542     """Get mean value.
    543
    544     Returns:
   (...)
    558         └──────────────────┘
    559     """
--> 560     return self._with_aggregation(lambda plx: self._to_compliant_expr(plx).mean())

File ~/scratch/.venv/lib/python3.12/site-packages/narwhals/expr.py:83, in Expr._with_aggregation(self, to_compliant_expr)
     81 if self._metadata.kind.is_scalar_like():
     82     msg = "Aggregations can't be applied to scalar-like expressions."
---> 83     raise InvalidOperationError(msg)
     84 return self.__class__(
     85     to_compliant_expr, self._metadata.with_kind(ExprKind.AGGREGATION)
     86 )

InvalidOperationError: Aggregations can't be applied to scalar-like expressions.
```

narwhals/_ibis/utils.py
dangotbanned: Great thanks @rwhitten577 

Resolved in (https://github.com/narwhals-dev/narwhals/pull/2000/commits/af770278af3567ebc4a77fc41c73d52d77d08309)

narwhals/_ibis/expr.py
rwhitten577: Ok pyright _should_ be fixed now. Ibis' type system is funky but did my best to set the correct types or cast to something that made sense

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,732 @@
...
+        return self._with_window_function(func)
+
+    def diff(self) -> Self:
+        def _func(window_inputs: WindowInputs) -> ir.NumericValue:
+            expr = cast("ir.NumericColumn", window_inputs.expr)
+            return expr - cast(
+                "ir.NumericColumn", expr.lag().over(ibis.window(following=0))
+            )
+
+        return self._with_window_function(_func)
```
dangotbanned, line 573: @NickCrews it seems like some of our issues (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2047599047) might be solved by changing the return type of *some* `ibis` methods to `Self`.
Just throwing this out there 🙂 

Picking one at random, but say we start with:

```py
expr: ir.NumericColumn
```

We first call:

https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/generic.py#L2707-L2711

```py
op_1: ir.Column = expr.lag()
```

Then:

https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/generic.py#L817-L125

```py
op_2: ir.Value = op_1.over(ibis.window(following=0))
```

But at that point we can't do the binary op, since the right-hand side isn't an `ir.NumericValue`:

https://github.com/ibis-project/ibis/blob/d7cd846627269b1f4c901cc3f0e380ee9024f796/ibis/expr/types/numeric.py#L646-L648

```py
# Operator "-" not supported for types "NumericColumn" and "Value"Pylance[reportOperatorIssue]
result = expr - op_2
```

narwhals/_ibis/expr.py
dangotbanned: Much appreciated @rwhitten577!

It seems the casts might be unavoidable for now (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2047704276) 😔 

I had to do a lot of that in 
- #2007 

which eventually led to 
- #2124

That is to say, identifying the issues is a big step and could even lead to improvements down the line 😅

narwhals/_ibis/expr.py
rwhitten577: Yep that would help greatly here to return `Self` from funcs like `over`, `name`, etc which currently lose the original column type.

rwhitten577:
> if we're adding initial support, i'd advocate for keeping it simple and just raising if `self._backend_version < (0,10)` for some operations, we can always lower the minimum version later

Just added this guard for the few places `ibis.cases` is used

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2000#issuecomment-2812863347

@rwhitten577 I fumbled big time on the spelling, but I've used (https://github.com/narwhals-dev/narwhals/pull/2000#issuecomment-2809165436) for these in

- [refactor: Use `@require.backend_version`](https://github.com/narwhals-dev/narwhals/pull/2000/commits/3248a50173adedb1de8faf2c101bfa398b7acedb)

narwhals/_ibis/dataframe.py
```diff
@@ -50,11 +50,7 @@ class IbisLazyFrame(CompliantLazyFrame["IbisExpr", "ir.Table"]):
...
     def __init__(
-        self: Self,
-        df: ir.Table,
-        *,
-        backend_version: tuple[int, ...],
-        version: Version,
+        self, df: ir.Table, *, backend_version: tuple[int, ...], version: Version
     ) -> None:
         self._native_frame: ir.Table = df
         self._version = version
```
dangotbanned: FYI @MarcoGorelli 

You might have noticed I've been gradually doing this where it isn't needed.
Now seemed like a good isolated time to remove them all for `ibis` 🙂 

I've been leaving them on the public API, in case there was some benefit I didn't know abour

- https://typing.python.org/en/latest/spec/generics.html#use-in-parameter-types
- (https://github.com/narwhals-dev/narwhals/pull/2000/commits/5df6b651522382e776c25dfaca2bb173908e8aed)

dangotbanned:
> yeah CI check would at least need to be green before merging
> 
> > we can't run this in every environment that we measure coverage on due to upper-bound constraits
> 
> they accepted my pr to remove upper-bound constraints, so we can probably include ibis in the coverage metrics

@MarcoGorelli we can do `ibis` coverage on `3.11` & `3.13` jobs
- (https://github.com/narwhals-dev/narwhals/pull/2000/commits/ae79ca4668055857275cf2885db538c57e0f2639)

I'm not sure how to disable it *only* on the `3.8` & `3.9` jobs 😞
- https://github.com/narwhals-dev/narwhals/actions/runs/14524661536/job/40753630734?pr=2000
- https://github.com/narwhals-dev/narwhals/actions/runs/14524661580/job/40753630959?pr=2000

pyproject.toml
binste: > I spotted a wild @binste! 😉

😄 Funny where paths are crossing... I'll be at PyData Darmstadt next week. Hit me up on Linkedin or via Email in case you're there as well! I think I also saw @MarcoGorelli on the talk schedule

Regarding type hints, I tried to work on this for Ibis with a few PRs but it was just too complex to disentangle for someone not that familiar with the internals of Ibis.

narwhals/_ibis/dataframe.py
MarcoGorelli: thanks - tbh I think we can remove them all separately if you like, IIRC there's some benefit to some IDE but overall I think it increases the noise

pyproject.toml
dangotbanned: @binste ah I'm not there sadly, but would recommend checking out @MarcoGorelli's [talk](https://pretalx.com/pyconde-pydata-2025/talk/CPCNRZ/)!

rwhitten577:
> > yeah CI check would at least need to be green before merging
> > > we can't run this in every environment that we measure coverage on due to upper-bound constraits
> > 
> > 
> > they accepted my pr to remove upper-bound constraints, so we can probably include ibis in the coverage metrics
> 
> @MarcoGorelli we can do `ibis` coverage on `3.11` & `3.13` jobs
> 
> * ([ae79ca4](https://github.com/narwhals-dev/narwhals/commit/ae79ca4668055857275cf2885db538c57e0f2639))
> 
> I'm not sure how to disable it _only_ on the `3.8` & `3.9` jobs 😞
> 
> * https://github.com/narwhals-dev/narwhals/actions/runs/14524661536/job/40753630734?pr=2000
> * https://github.com/narwhals-dev/narwhals/actions/runs/14524661580/job/40753630959?pr=2000

Hey @dangotbanned @MarcoGorelli , anything I can do to help move this PR along? Think it was just the CI coverage on 3.8/3.9 that needed to be resolved

MarcoGorelli:
hey @rwhitten577 ! I think we can lower the target threshold for those jobs in the `.yml` file, say to 75 instead of 80

We already test that coverage is at 100% in the `3.11` and `3.13` jobs anyway 😇

MarcoGorelli:
also, some merge conflicts (so sorry 🙈 )

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2000#issuecomment-2844938786)

Apologies for not following up @rwhitten577! 😔 

I wanted to check in with @MarcoGorelli to make sure everything we *intended* to have coverage for was in place.
I'm still happy to help review once things are green 🙂

narwhals/_ibis/dataframe.py
dangotbanned: Resolved in (#2399)

narwhals/_namespace.py
```diff
@@ -280,6 +294,10 @@ def from_native_object(
     @classmethod
     def from_native_object(cls, native: _NativeDask, /) -> Namespace[DaskNamespace]: ...
 
+    @overload
+    @classmethod
+    def from_native_object(cls, native: _NativeIbis, /) -> Namespace[IbisNamespace]: ...
```
dangotbanned, line 299: It was *almost* painless!

https://github.com/narwhals-dev/narwhals/actions/runs/14779913921/job/41496639991?pr=2000

@MarcoGorelli for the purposes of the new support - does grouping `Ibis` under `LazyOnly` make sense?

We've got that for `DuckDB`, but that also has `Interchange` support in `v1`

dangotbanned: https://github.com/rwhitten577/narwhals/blob/9a8788c7135d40892eff0fcbc206932564c3c6ea/narwhals/_namespace.py#L76-L92

MarcoGorelli: it should the same as what we do for duckdb

altair uses `stable.v1` and supports ibis so the important thing is that that keeps working

rwhitten577: Realized I missed that — adding it now, but it only solves the first mypy error on `from_backend`. The other 2 on `from_native_object` still exist 🤔

dangotbanned: All good - these groupings are *purely* internal for now.
Just needs to be in that hierarchy of aliases for these `@overload` defs

https://github.com/narwhals-dev/narwhals/blob/ea1f5f0c7b73850a97c3f0b4cb92709fbfed2e7b/narwhals/_namespace.py#L58-L88

MarcoGorelli: Given that `ibis.Table` is just detected as `Any` anyway, should we avoid this overload?

dangotbanned: > Realized I missed that — adding it now, but it only solves the first mypy error on `from_backend`. The other 2 on `from_native_object` still exist 🤔

No worries @rwhitten577 I'll check it out locally and see if I can spot the issue more clearly

dangotbanned: I'm pretty sure the issue is `mypy` not understanding `ibis` in some way (e.g. https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2046735198)

I think we'll need to add some minimal protocol for matching an `ibis.Table` - in a similar way that `Dask`, `Modin`, `CuDF` are spec'd

https://github.com/narwhals-dev/narwhals/blob/ea1f5f0c7b73850a97c3f0b4cb92709fbfed2e7b/narwhals/_namespace.py#L99-L116

Those types aren't used at runtime, but help remove the `Unknown`s - which are the issue that produces the false negatives in https://github.com/narwhals-dev/narwhals/actions/runs/14779913921/job/41496639991?pr=2000

rwhitten577: Nice, thanks @dangotbanned! That fixed it & mypy is looking happy now

narwhals/_namespace.py
```diff
@@ -108,7 +107,7 @@ class _NativeCuDF(Protocol):
         def to_pylibcudf(self, *args: Any, **kwds: Any) -> Any: ...
 
     class _NativeIbis(Protocol):
-        def sql(self, query: str, /, *, dialect: str | None = None) -> ibis.Table: ...
+        def sql(self, *args: Any, **kwds: Any) -> Any: ...
```
dangotbanned, line 110: @rwhitten577 just as a follow-up to (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2070605863)

Here I'm trying to reduce the chance of `mypy` looking `ibis.Table` - since they haven't got a `py.typed`.
It seems like there's no issue *for now*, I'm just being careful - since the errors in (https://github.com/narwhals-dev/narwhals/actions/runs/14779913921/job/41496639991?pr=2000) are pretty cryptic 😅

rwhitten577: Ok thanks @dangotbanned, makes sense!

dangotbanned:
Sorry for the last minute curveball!

- #2461

There's a few new things to copy over from `DuckDB`, but nothing major 🙂

- https://github.com/narwhals-dev/narwhals/actions/runs/14781890227/job/41502401824?pr=2000
- https://github.com/narwhals-dev/narwhals/actions/runs/14781890198/job/41502401267?pr=2000

rwhitten577:
> Sorry for the last minute curveball!
> 
> * [chore: Add `Compliant*.to_narwhals` #2461](https://github.com/narwhals-dev/narwhals/pull/2461)
> * https://github.com/narwhals-dev/narwhals/pull/2461/files#diff-7279bb4875393702cb708fa5e16bf232ed9ff3ac45c4734977291696be590a04
> 
> There's a few new things to copy over from `DuckDB`, but nothing major 🙂
> 
> * https://github.com/narwhals-dev/narwhals/actions/runs/14781890227/job/41502401824?pr=2000
> * https://github.com/narwhals-dev/narwhals/actions/runs/14781890198/job/41502401267?pr=2000

No worries @dangotbanned , just updated to include these changes

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,711 @@
...
+            )
+        )
+
+    def fill_null(self, value: Self | Any, strategy: Any, limit: int | None) -> Self:
+        if strategy is not None:
+            msg = "`strategy` is not supported for the Ibis backend"
+            raise NotImplementedError(msg)
+        if limit is not None:
+            msg = "`limit` is not supported for the Ibis backend"  # pragma: no cover
+            raise NotImplementedError(msg)
```
MarcoGorelli, line 631: it should be possible to support these, we do it for `_spark_like` and `_duckdb`, using `last_value`. looks like Ibis expressions have `.last` so it should be possible to use that? 

ok to leave this as a follow-up though 👍

MarcoGorelli (commented):
well done @rwhitten577 ! it so impressive how far you got with this, and how many merge conflicts you navigated through! 🙌 😍 

I haven't reviewed everything yet, just left some minor comments, but I think we should be able to ship this soon!

narwhals/_ibis/expr_dt.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def microsecond(self) -> IbisExpr:
+        return self._compliant_expr._with_callable(
+            lambda _input: _input.microsecond(),
+        )
+
+    def nanosecond(self) -> IbisExpr:
+        return self._compliant_expr._with_callable(
+            lambda _input: _input.microsecond() * 1000,
+        )
```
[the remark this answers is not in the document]
rwhitten577: Good catch, thanks! Updating to `not_implemented`

narwhals/_ibis/expr.py
rwhitten577: Naturally Ibis does not yet support ignoring nulls for `first` and `last` when in the context of a window function: https://github.com/ibis-project/ibis/issues/9539 

I'll add a comment linking to this issue

narwhals/_compliant/group_by.py
```diff
@@ -221,10 +221,13 @@ def _evaluate_expr(self, expr: LazyExprT_contra, /) -> Iterator[NativeExprT_co]:
...
             for native_expr, name, alias in zip(native_exprs, output_names, aliases):
                 if name not in exclude:
-                    yield native_expr.alias(alias)
+                    yield self._alias_native_expr(native_expr, alias)
         else:
             for native_expr, alias in zip(native_exprs, aliases):
-                yield native_expr.alias(alias)
+                yield self._alias_native_expr(native_expr, alias)
+
+    def _alias_native_expr(self, native_expr: Any, alias: str) -> Any:
```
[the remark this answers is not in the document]
rwhitten577: I got it to work with `NativeExprT` once I added `alias` to `NativeExpr`, but had issues returning `NativeExprT_co`. I set up `alias` to return `Self` so it at least recognizes `NativeExprT` returns `NativeExprT`: 

```python
class NativeExpr(Protocol):
    ...

    def alias(self, *args: Any, **kwds: Any) -> Self: ...
```

Is there another way to solve so `NativeExprT_co` works? (Admittedly not as familiar with covariance)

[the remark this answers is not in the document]
dangotbanned: > can `NativeExprT_co` be used as a type here?

Apologies I haven't had a chance to check the PR beyond this 😅 

You can only use covariant `TypeVar`(s) in return annotations.

If it is used in both return and as a parameter - then you need to switch to invariant (`NativeExprT`)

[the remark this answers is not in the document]
rwhitten577: Thanks @dangotbanned , I used the invariant `NativeExprT` and added `alias` to it since that's what was getting called by default. However for Ibis' exprs they use `.name` instead of `.alias` so pyright is not letting the ibis impl pass given that `ir.Value` doesn't match the signature of `NativeExpr` 🤔 

I've been wrestling with it and seems like to get it to work it may require another variant of `NativeExpr` that has `.name`? 

I could also change the functionality here to use `getattr` on the expr and set a class var to control which method gets called to rename the expr, which still can be overridden in subclasses: 

```python
class LazyGroupBy(
    ParseKeysGroupBy[CompliantLazyFrameT, LazyExprT_contra],
    CompliantGroupBy[CompliantLazyFrameT, LazyExprT_contra],
    Protocol38[CompliantLazyFrameT, LazyExprT_contra, NativeExprT_co],
):
    _expr_alias_func_name: str = "alias"
    ...

    def _evaluate_expr(self, expr: LazyExprT_contra, /) -> Iterator[NativeExprT_co]:
        ...
            for native_expr, name, alias in zip(native_exprs, output_names, aliases):
                if name not in exclude:
                    yield getattr(native_expr, self._expr_alias_func_name)(alias)
        else:
            for native_expr, alias in zip(native_exprs, aliases):
                yield getattr(native_expr, self._expr_alias_func_name)(alias)

...

class IbisGroupBy(LazyGroupBy["IbisLazyFrame", "IbisExpr", "ir.Value"]):
    _expr_alias_func_name = "name"
    ...
```

Let me know which would be preferred!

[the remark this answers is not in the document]
dangotbanned: Okay I couldn't help myself 😄 

Resolved in (https://github.com/narwhals-dev/narwhals/pull/2000/commits/f60bdb53e8ac6077107013ccd20dd3907e6a13cb)

narwhals/_compliant/expr.py
```diff
@@ -901,6 +901,9 @@ def _is_expr(cls, obj: Self | Any) -> TypeIs[Self]:
     def _with_callable(self, call: Callable[..., Any], /) -> Self: ...
     def _with_alias_output_names(self, func: AliasNames | None, /) -> Self: ...
 
+    @classmethod
+    def _alias_native(cls, expr: NativeExprT, name: str, /) -> NativeExprT: ...
+
```
dangotbanned, line 906: (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2071693761)

By defining it here, we
- Already have an invariant `TypeVar`
- Don't need to extend the definition of `NativeExpr` protocol
- Can reference the method from outside of `GroupBy` easily

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,717 @@
...
+    def dt(self) -> IbisExprDateTimeNamespace:
+        return IbisExprDateTimeNamespace(self)
+
+    @property
+    def list(self) -> IbisExprListNamespace:
+        return IbisExprListNamespace(self)
+
+    @property
+    def struct(self) -> IbisExprStructNamespace:
+        return IbisExprStructNamespace(self)
```
dangotbanned, line 712: > [!NOTE]
> **(To self)**: Add protocols for `LazyExpr*Namespace` 

Realised while looking through the PR that we're not enforcing the newer-style protocols for `LazyExpr` (except for `LazyExprNameNamespace`)

https://github.com/narwhals-dev/narwhals/blob/3d92e0df1e50e1aa4a7237d039b996c13d890459/narwhals/_compliant/expr.py#L904-L910

Everything else is `Any`, which IIRC was for the benefit of `not_implemented` 🤔 

https://github.com/narwhals-dev/narwhals/blob/3d92e0df1e50e1aa4a7237d039b996c13d890459/narwhals/_compliant/expr.py#L282-L293

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,268 @@
...
+                cast("ir.NumericColumn", e.isnull().ifelse(lit(0), lit(1)))
+                for _expr in exprs
+                for e in _expr(df)
+            )
+            first_expr_name = exprs[0](df)[0].get_name()
+
+            def _name_preserving_sum(
+                e1: ir.NumericColumn, e2: ir.NumericColumn
+            ) -> ir.NumericColumn:
+                return cast("ir.NumericColumn", (e1 + e2).name(e1.get_name()))
```
MarcoGorelli, line 198: these name-preserving operations shouldn't be necessary, as we set the name as late as possible anyway

if this was taken from the dask implementation, it's probably something we forgot to clean up - i'll remove it from there in https://github.com/narwhals-dev/narwhals/pull/2473

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,268 @@
...
+    @property
+    def _lazyframe(self) -> type[IbisLazyFrame]:
+        return IbisLazyFrame
+
+    def concat(
+        self, items: Iterable[IbisLazyFrame], *, how: ConcatMethod
+    ) -> IbisLazyFrame:
+        if how in {"horizontal", "diagonal"}:
+            msg = f"{how!r} concat not supported for Ibis. Please join instead."
+            raise NotImplementedError(msg)
```
MarcoGorelli, line 58: `'horizontal'` is already disallowed before we get here, if we can't support 'diagonal' let's just raise on that

narwhals/_ibis/utils.py
```diff
@@ -0,0 +1,186 @@
...
+    if isinstance_or_issubclass(dtype, dtypes.UInt32):
+        return ibis_dtypes.UInt32()
+    if isinstance_or_issubclass(dtype, dtypes.UInt16):  # pragma: no cover
+        return ibis_dtypes.UInt16()
+    if isinstance_or_issubclass(dtype, dtypes.UInt8):  # pragma: no cover
+        return ibis_dtypes.UInt8()
+    if isinstance_or_issubclass(dtype, dtypes.String):
+        return ibis_dtypes.String()
+    if isinstance_or_issubclass(dtype, dtypes.Boolean):  # pragma: no cover
+        return ibis_dtypes.Boolean()
```
MarcoGorelli, line 154: i think it should be possible to cover these?

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,717 @@
...
+            .name(_input.get_name())
+        )
+
+    def count(self) -> Self:
+        return self._with_callable(lambda _input: _input.count())
+
+    def len(self) -> Self:
+        return self._with_callable(
+            lambda _input: _input.cast("string").fill_null("__NULL__").count()
+        )
```
MarcoGorelli, line 419: this is potentially very expensive

MarcoGorelli (changes requested):
Thanks @rwhitten577 !

Overall looks good, the main blocker from my end is `n_unique` and `len`, in which you cast to `'string'`. That's a very expensive operation which we shouldn't do here

Can we find another way, which doesn't involve casting to strings?

narwhals/_ibis/utils.py
```diff
@@ -0,0 +1,186 @@
...
+    if isinstance_or_issubclass(dtype, dtypes.Categorical):
+        msg = "Categorical not supported by Ibis"
+        raise NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        return ibis_dtypes.Timestamp()
+    if isinstance_or_issubclass(dtype, dtypes.Duration):  # pragma: no cover
+        _time_unit = getattr(dtype, "time_unit", "us")
+        return ibis_dtypes.Interval(_time_unit)
```
dangotbanned, line 162: I'm not sure if this'll work, `pyright` seems to be pointing at the missing constructor

image

It seems you need to do:

```py
ibis_dtypes.Interval(unit=...)
```

Which leads to another issue on the typing from `getattr`
> Argument of type "Any | str" cannot be assigned to parameter "unit" of type "IntervalUnit" in function "__init__"
  Type "Any | str" is not assignable to type "IntervalUnit"
    "str" is not assignable to "IntervalUnit"


That one can be solved with:

```py
ibis_dtypes.Interval(unit=dtype.time_unit)
```

Which then leaves us with a final issue 😅 (that we also have in `polars`, with `"s"`):

> Argument of type "TimeUnit" cannot be assigned to parameter "unit" of type "IntervalUnit" in function "__init__"
  Type "TimeUnit" is not assignable to type "IntervalUnit"
    "Literal['ms']" is not assignable to "IntervalUnit"

That last one can just be ignored like 

https://github.com/narwhals-dev/narwhals/blob/eb1368266cdae6d31ccfe1cde720104241232223/narwhals/_polars/utils.py#L203-L204

dangotbanned: Related to (https://github.com/narwhals-dev/narwhals/pull/2000/commits/5103d3d6996479b23a18651b74c13c9338dc1d75)

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,717 @@
...
+        return self._with_callable(lambda _input: _input.abs())
+
+    def mean(self) -> Self:
+        return self._with_callable(
+            lambda _input: _input.mean(),
+        )
+
+    def skew(self) -> Self:
+        msg = "`skew` is not supported for the Ibis backend"
+        raise NotImplementedError(msg)
```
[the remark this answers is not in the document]
rwhitten577: Done, added link to https://github.com/ibis-project/ibis/issues/11176 as well in comment

narwhals/_ibis/utils.py
```diff
@@ -0,0 +1,186 @@
...
+        return ibis_dtypes.UInt64()
+    if isinstance_or_issubclass(dtype, dtypes.UInt32):
+        return ibis_dtypes.UInt32()
+    if isinstance_or_issubclass(dtype, dtypes.UInt16):  # pragma: no cover
+        return ibis_dtypes.UInt16()
+    if isinstance_or_issubclass(dtype, dtypes.UInt8):  # pragma: no cover
+        return ibis_dtypes.UInt8()
+    if isinstance_or_issubclass(dtype, dtypes.String):
+        return ibis_dtypes.String()
+    if isinstance_or_issubclass(dtype, dtypes.Boolean):  # pragma: no cover
```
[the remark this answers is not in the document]
rwhitten577: Thanks good catch, removed these that were copied over from earlier duckdb impl. Left the Decimal, UInt128 and Int128 uncovered since it didn't seem like any tests were hitting those.

narwhals/_ibis/namespace.py
rwhitten577: Thanks, cleaned up here for Ibis too

narwhals/_ibis/namespace.py
rwhitten577: Updated to only raise on `how == "diagonal"`

narwhals/_ibis/utils.py
rwhitten577: Thanks @dangotbanned! Made these updates but then mypy thought it was an unused `ignore[arg-type]`. Updated to include `# type: ignore[arg-type,unused-ignore]` to satisfy both pyright and mypy.

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,713 @@
...
+        return self._with_callable(
+            lambda _input: _input.nunique() + _input.isnull().any().cast("int8")
+        )
+
+    def count(self) -> Self:
+        return self._with_callable(lambda _input: _input.count())
+
+    def len(self) -> Self:
+        return self._with_callable(
+            lambda _input: _input.count() + _input.isnull().cast("int8").sum()
```
MarcoGorelli, line 411: does it work to do something like
```python
        def func(df: IbisLazyFrame) -> list[ibis.Expression]:
            return [df.count()]

        return self.__class__(
            func,
            evaluate_output_names=self._evaluate_output_names,
            alias_output_names=self._alias_output_names,
            version=self._version,
        )
```
without using `_with_callable`? Looks like Ibis has `table.count` for this, but no expression (🤔 )

rwhitten577: Ah good idea, works like this with `df.native.count()` 👍

narwhals/_ibis/utils.py
```diff
@@ -0,0 +1,221 @@
...
+        return ibis_dtypes.String()
+    if isinstance_or_issubclass(dtype, dtypes.Boolean):
+        return ibis_dtypes.Boolean()
+    if isinstance_or_issubclass(dtype, dtypes.Categorical):
+        msg = "Categorical not supported by Ibis"
+        raise NotImplementedError(msg)
+    if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        return ibis_dtypes.Timestamp()
+    if isinstance_or_issubclass(dtype, dtypes.Duration):
+        return ibis_dtypes.Interval(unit=dtype.time_unit)  # type: ignore[arg-type,unused-ignore]
```
dangotbanned, line 197: Thanks for following up @rwhitten577 (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2072385686)

If you run into any of these, use the `pyright` ignore instead

```suggestion
        return ibis_dtypes.Interval(unit=dtype.time_unit)  # pyright: ignore[reportArgumentType]
```

I suppose this might be more common with `ibis` since `mypy` doesn't seem to know what's happening 😄

rwhitten577: Nice, thanks for the tip! Updated

narwhals/_ibis/expr_dt.py
```diff
@@ -76,19 +82,25 @@ def date(self) -> IbisExpr:
             lambda _input: _input.date(),
         )
 
+    def _bucket(self, kwds: dict[BucketUnit, Any], /) -> Callable[..., ir.TimestampValue]:
+        def fn(_input: ir.TimestampValue) -> ir.TimestampValue:
+            return _input.bucket(**kwds)
+
+        return fn
```
dangotbanned, line 89: There *might* need to be some version gating for this, since its marked with `@util.experimental`

https://github.com/ibis-project/ibis/blob/9ef7ca7d3dc91ddffb54514579792c7c9f9ab6b2/ibis/expr/types/temporal.py#L938-L956

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,263 @@
...
+                    cast("ir.StringValue", ibis.cases((nm, ""), else_=separator))
+                    for nm in null_mask[:-1]
+                )
+                result = reduce(
+                    lambda x, y: x + y,
+                    (s + v for s, v in zip(separators, values)),
+                    init_value,
+                )
+
+            return [result]
```
dangotbanned, line 102: I think this is based on an old version of `DuckDBNamespace.concat_str`?

The refactor we did in (https://github.com/narwhals-dev/narwhals/pull/2321#discussion_r2020949610
) might be applicable here as well?

https://github.com/narwhals-dev/narwhals/blob/b86b5a2b49c5a195f07b63553b771d6652f06cdb/narwhals/_duckdb/namespace.py#L77-L93

dangotbanned: I think that PR generally might be good to look over and see if there are any other bits that are out of sync

- #2321

narwhals/_ibis/expr_dt.py
rwhitten577: Can add. Looks like its been around since 7.1.0 
https://github.com/ibis-project/ibis/commit/ca0f7bc830669288815b074a322f186fdb261dd1#diff-321129912fed3de2b9d062cbe255bc303e1706f96448f939119117926769e977

dangotbanned: @rwhitten577 yeah if that's above the minimum we're targeting.
IIRC that was `6.0.0`, but I'm likely wrong

narwhals/_ibis/namespace.py
rwhitten577: I was able to simplify `concat_str` a bit for the `ignore_nulls` path

narwhals/_ibis/expr.py
```diff
@@ -263,8 +264,8 @@ def _with_window_function(self, window_function: WindowFunction) -> Self:
         return result
 
     @classmethod
-    def _alias_native(cls, expr: ir.Column, name: str, /) -> ir.Value:
-        return expr.name(name)
+    def _alias_native(cls, expr: ExprT, name: str, /) -> ExprT:
+        return cast("ExprT", expr.name(name))
```
dangotbanned, line 268: @rwhitten577 you might be able to use this in more places - where `expr.name` is causing typing issues 

Should mean you get back the same type you started with 🙂

narwhals/_ibis/namespace.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2000/commits/1e02d8908203f938831de80d5bf2c1c8be4320b9), (https://github.com/narwhals-dev/narwhals/pull/2000/commits/7057c1edbc5f31d5095aa8ed039157fd1b2c193c)

MarcoGorelli:
thanks! gonna cc @jcrist @gforsyth  - I recall you said you'd have been interested in this, so if there's anything that stands out that should be improved, please do let us know! 🙏

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,478 @@
...
+                else cast("ir.Column", ibis.asc(by[i], nulls_first=False))
+                for i in range(len(by))
+            ]
+        else:
+            sort_cols = [
+                cast("ir.Column", ibis.desc(by[i], nulls_first=True))
+                if descending[i]
+                else cast("ir.Column", ibis.asc(by[i], nulls_first=True))
+                for i in range(len(by))
+            ]
```
dangotbanned, line 414: Is there any way to write this with less `cast`(s)?

I'm finding it difficult to follow

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,478 @@
...
+                    continue
+
+        return self._with_native(joined)
+
+    def _convert_predicates(
+        self, other: Self, left_on: Sequence[str], right_on: Sequence[str]
+    ) -> list[ir.BooleanColumn] | Sequence[str]:
+        if len(left_on) != len(right_on):  # pragma: no cover
+            msg = "'left_on' and 'right_on' must have the same number of columns."
+            raise ValueError(msg)
```
dangotbanned, line 364: Do we need this if it doesn't hit coverage?

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,478 @@
...
+
+        assert right_on is not None  # noqa: S101
+
+        # Drop duplicate columns from the right table. Ibis keeps them.
+        for right in right_on if isinstance(right_on, list) else [right_on]:
+            to_drop = right + suffix
+            if to_drop in joined.columns:
+                joined = joined.drop(right + suffix)
+            else:  # pragma: no cover
+                continue
```
dangotbanned, line 347: Do we need this if it doesn't hit coverage?

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,478 @@
...
+
+        if by_right is not None:
+            for right in by_right if not isinstance(by_right, str) else [by_right]:
+                to_drop = right + suffix
+                if to_drop in joined.columns:
+                    joined = joined.drop(right + suffix)
+                else:  # pragma: no cover
+                    continue
```
dangotbanned, line 355: Do we need this if it doesn't hit coverage?

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,478 @@
...
+                right = pred_op.right.name
+
+                # If right column is not in the left table, drop it as it will be present in the left column
+                # Mirrors how polars works.
+                if left != right and right not in self.columns:
+                    joined = joined.drop(right)
+
+        return self._with_native(joined)
+
+    def join_asof(  # noqa: PLR0912
```
dangotbanned, line 304: Any efforts towards avoiding these are super appreciated 🙏

- https://docs.astral.sh/ruff/rules/too-many-branches/
- #2485

> [!IMPORTANT]
> Not blocking

dangotbanned (changes requested):
Thanks for working on this and being patient with us @rwhitten577!

I added some stuff in (https://github.com/narwhals-dev/narwhals/pull/2000#commits-pushed-7e5764b), and have some suggestions/questions

Really impressed with how quick you've been to keep this in sync 😍

narwhals/_ibis/dataframe.py
rwhitten577: It does seem to work without this check since the result of `zip(left_on, right_on)` will only keep the number of elements from the shorter list. 

E.g., this input: `df.join(df, how=how, left_on=["antananarivo"], right_on=["antananarivo", "bob"])` 
produces predicates of `["antananarivo" == "antananarivo"]`. `"bob"` is ignored. 

I tested this with the polars constructor and polars does raise an error: 
```
{InvalidOperationError}the number of columns given as join key (left: 1, right:2) should be equal

Resolved plan until failure:

	---> FAILED HERE RESOLVING 'sink' <---
DF ["antananarivo", "bob", "zor ro"]; PROJECT */3 COLUMNS
```

Do you think this length check should be further upstream in the NW API e.g., in [dataframe.py](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/dataframe.py#L237)? Seems like it may be something worth checking for all backend integrations.

rwhitten577: Opened https://github.com/narwhals-dev/narwhals/pull/2504 to solve this

narwhals/_ibis/dataframe.py
rwhitten577: I'll revisit these. I should be able to expand the test case to cover instead of ignoring the 2nd iteration

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,251 @@
...
+                )
+            else:
+                result = cols_casted[0]
+                for col in cols_casted[1:]:
+                    result = result + cast(
+                        "ir.StringValue",
+                        ibis.cases((col.isnull(), ""), else_=separator + col),
+                    )
+
+            return [result]
```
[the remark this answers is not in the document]
rwhitten577: Updated in https://github.com/narwhals-dev/narwhals/pull/2000/commits/630003a5bcc8456af59be39216ad92d48844fc86. 
🤦‍♂️ thanks, I missed `join` when trying to find something like `concat_ws`. That solves the branch for `ignore_nulls` much simpler. Also simplified the other path by essentially using `concat` (same as the `+` operator).

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,251 @@
...
+                for e in _expr(df)
+            )
+            first_expr_name = exprs[0](df)[0].get_name()
+
+            return [
+                (
+                    reduce(lambda x, y: x + y, expr)
+                    / reduce(lambda x, y: x + y, non_null)
+                ).name(first_expr_name)
+            ]
```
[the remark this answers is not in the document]
rwhitten577: Thanks, yeah I think you guys have improved the output name handling over the past few months where I used to have to apply it manually. Cleaning up.

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,251 @@
...
+            evaluate_output_names=combine_evaluate_output_names(*exprs),
+            alias_output_names=combine_alias_output_names(*exprs),
+            backend_version=self._backend_version,
+            version=self._version,
+        )
+
+    def max_horizontal(self, *exprs: IbisExpr) -> IbisExpr:
+        def func(df: IbisLazyFrame) -> list[ir.Value]:
+            cols = list(chain.from_iterable(expr(df) for expr in exprs))
+            return [ibis.greatest(*cols).name(cols[0].get_name())]
```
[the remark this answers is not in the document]
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c8a1296e871ff3534237bf3db317e980117336c4

narwhals/_ibis/namespace.py
```diff
@@ -0,0 +1,251 @@
...
+            evaluate_output_names=combine_evaluate_output_names(*exprs),
+            alias_output_names=combine_alias_output_names(*exprs),
+            backend_version=self._backend_version,
+            version=self._version,
+        )
+
+    def min_horizontal(self, *exprs: IbisExpr) -> IbisExpr:
+        def func(df: IbisLazyFrame) -> list[ir.Value]:
+            cols = list(chain.from_iterable(expr(df) for expr in exprs))
+            return [ibis.least(*cols).name(cols[0].get_name())]
```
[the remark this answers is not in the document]
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c8a1296e871ff3534237bf3db317e980117336c4

narwhals/_ibis/namespace.py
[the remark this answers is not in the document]
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c8a1296e871ff3534237bf3db317e980117336c4 and https://github.com/narwhals-dev/narwhals/pull/2000/commits/efa47c6eb699769953564052579b182c9b3680e8

narwhals/_ibis/dataframe.py
rwhitten577: Simplified to a single branch and single cast in https://github.com/narwhals-dev/narwhals/pull/2000/commits/0fbc0ec3abe6d1f43d22c0b58f191d4a6a8a336d

narwhals/_ibis/dataframe.py
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c6eccc413bc6d52965749d8fadf4d272dc7e4fb1

narwhals/_ibis/dataframe.py
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c6eccc413bc6d52965749d8fadf4d272dc7e4fb1 by simplifying the branches here

narwhals/_ibis/dataframe.py
rwhitten577: Resolved in https://github.com/narwhals-dev/narwhals/pull/2000/commits/c6eccc413bc6d52965749d8fadf4d272dc7e4fb1

narwhals/_ibis/dataframe.py
```diff
@@ -339,20 +339,13 @@ def join_asof(  # noqa: PLR0912
...
 
         # Drop duplicate columns from the right table. Ibis keeps them.
-        for right in right_on if isinstance(right_on, list) else [right_on]:
-            to_drop = right + suffix
-            if to_drop in joined.columns:
-                joined = joined.drop(right + suffix)
-            else:  # pragma: no cover
-                continue
+        to_drop = right_on + suffix
+        joined = joined.drop(to_drop) if to_drop in joined.columns else joined
```
dangotbanned: If you can give this a helpful name, it could be a function you reuse for the loop at

https://github.com/rwhitten577/narwhals/blob/290449d3a0a9d2e0df05aa2849b99580ebdbe967/narwhals/_ibis/dataframe.py#L347-L348

Unless there's some other magic, those lines are the same right?

rwhitten577: Yep good call, just the difference of `right_on` or `by_right` being considered

tests/expr_and_series/len_test.py
```diff
@@ -7,11 +7,12 @@
...
 
 def test_len_no_filter(constructor: Constructor) -> None:
-    data = {"a": list("xyz"), "b": [1, 2, 1]}
-    expected = {"l": [3], "l2": [6]}
+    data = {"a": list("xyz"), "b": [1, 2, None]}
+    expected = {"l": [3], "l2": [6], "l3": [3]}
     df = nw.from_native(constructor(data)).select(
         nw.col("a").len().alias("l"),
         (nw.col("a").len() * 2).alias("l2"),
+        nw.col("b").len().alias("l3"),
```
dangotbanned, line 15: @rwhitten577 just wanted to double check on this test change

It seems to have followed (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2072365853) and the change was in (https://github.com/narwhals-dev/narwhals/commit/465674274b3b0d5a9ae3118ab9b6eb1ac63b1e17)

@MarcoGorelli would you have wanted this split into another test - instead of changing the current case?

dangotbanned: Thanks for explaining @rwhitten577 

I'm fine with the new test as-is 🙂

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,681 @@
...
+            native_dtype = narwhals_to_native_dtype(dtype, self._version)
+            # ibis `cast` overloads do not include DataType, only literals
+            return _input.cast(native_dtype)  # type: ignore[unused-ignore]
+
+        return self._with_callable(_func)
+
+    def is_unique(self) -> Self:
+        return self._with_callable(
+            lambda _input: _input.cast("string")
+            .fill_null("__NULL__")
```
MarcoGorelli, line 616: this needs updating, we can't cast to string here

tests/expr_and_series/len_test.py
MarcoGorelli: thanks for explaining, this looks fine

MarcoGorelli (commented):
Amazing, thanks @rwhitten577 and @dangotbanned for reviewing!

Got some mostly minor comments. I've split out some follow-ups in https://github.com/narwhals-dev/narwhals/issues/2525

The most important comments from my end are:
- not casting to string for numerical ops
- not repeatedly calling `.drop` but instead collecting the items to drop into a list first (in `join`). In theory I think Ibis should be able to optimise this away, but  I don't know if they do, so it's probably safer to just call `drop` once with everything we need to drop

Once this is addressed, we can ship it

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,426 @@
...
+        rname = "{name}" + suffix
+        if other == self:
+            # Ibis does not support self-references unless created as a view
+            other = self._with_native(other.native.view())
+        if how_native == "cross":
+            joined = self.native.join(other.native, how=how_native, rname=rname)
+            return self._with_native(joined)
+        if left_on is None or right_on is None:  # pragma: no cover
+            msg = f"For '{how_native}' joins, both 'left_on' and 'right_on' must be provided."
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
dangotbanned: @MarcoGorelli this is related btw
- https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2083216320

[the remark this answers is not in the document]
rwhitten577: It does look like the incoming values get validated here: https://github.com/narwhals-dev/narwhals/blob/main/narwhals/dataframe.py#L263-L274 

However, the types for `CompliantLazyFrame.join` are `Sequence[str] | None`. By removing this block mypy will raise errors for downstream assumptions about `left_on` or `right_on` since it could be `None`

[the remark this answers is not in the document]
MarcoGorelli: thanks - fine to just use `assert` for this then, like
```
            # help mypy
            assert left_on is not None  # noqa: S101
            assert right_on is not None  # noqa: S101
```

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,426 @@
...
+        joined = self.native.join(other.native, predicates, how=how_native, rname=rname)
+        if how_native == "left":
+            right_names = (n + suffix for n in right_on)
+            joined = self._join_drop_duplicate_columns(joined, right_names)
+            it = (cast("Binary", p.op()) for p in predicates if not isinstance(p, str))
+            for pred in it:
+                right = pred.right.name
+                # Mirrors how polars works.
+                if right not in self.columns and pred.left.name != right:
+                    joined = joined.drop(right)
```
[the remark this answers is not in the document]
rwhitten577: Updated in https://github.com/narwhals-dev/narwhals/pull/2000/commits/b9626920480f546c63466232ad79be2af0829f00

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,681 @@
...
+        # Unused, just for compatibility with PandasLikeExpr
+        from narwhals._ibis.namespace import IbisNamespace
+
+        return IbisNamespace(backend_version=self._backend_version, version=self._version)
+
+    def _cum_window_func(
+        self, *, reverse: bool, func_name: Literal["sum", "max", "min", "count"]
+    ) -> WindowFunction:
+        def func(window_inputs: WindowInputs) -> ir.Value:
+            from ibis import _ as col  # ignore-banned-import
```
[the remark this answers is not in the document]
rwhitten577: Updated this and a few other places in https://github.com/narwhals-dev/narwhals/pull/2000/commits/68e7b8d3a58fb49dd2e4c5e86e53633ceadd9627

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,681 @@
...
+
+        if center:
+            preceding = window_size // 2
+            following = window_size - preceding - 1
+        else:
+            preceding = window_size - 1
+            following = 0
+
+        def func(window_inputs: WindowInputs) -> ir.Value:
+            from ibis import _ as col  # ignore-banned-import
```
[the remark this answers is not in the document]
rwhitten577: Updated in https://github.com/narwhals-dev/narwhals/pull/2000/commits/68e7b8d3a58fb49dd2e4c5e86e53633ceadd9627

narwhals/_ibis/dataframe.py
```diff
@@ -1,161 +1,426 @@
...
+            on: ir.BooleanColumn = op(self.native[left_on], other.native[right_on])
+        else:
+            msg = "Only `backward` and `forward` strategies are currently supported for Ibis"
+            raise NotImplementedError(msg)
+        if by_left is not None and by_right is not None:
+            predicates = self._convert_predicates(other, by_left, by_right)
+        joined = self.native.asof_join(other.native, on, predicates, rname=rname)
+        joined = self._join_drop_duplicate_columns(joined, [right_on + suffix])
+        if by_right is not None:
+            by_right = by_right if not isinstance(by_right, str) else [by_right]
```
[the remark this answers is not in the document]
rwhitten577: Removed in https://github.com/narwhals-dev/narwhals/pull/2000/commits/a1f6a48eb6fe3af109e1f73752448f799fc549a1

narwhals/_ibis/dataframe.py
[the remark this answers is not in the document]
rwhitten577: Added in https://github.com/narwhals-dev/narwhals/pull/2000/commits/a1f6a48eb6fe3af109e1f73752448f799fc549a1

narwhals/_ibis/expr.py
```diff
@@ -0,0 +1,681 @@
...
+    @classmethod
+    def from_column_names(
+        cls: type[Self],
+        evaluate_column_names: EvalNames[IbisLazyFrame],
+        /,
+        *,
+        context: _FullContext,
+    ) -> Self:
+        def func(df: IbisLazyFrame) -> list[ir.Column]:
+            return [df.native[name] for name in evaluate_column_names(df)]
```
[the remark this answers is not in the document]
MarcoGorelli: cool, thanks for explaining

aiming for a full review today, hopefully we can ship it

MarcoGorelli (approved):
thanks @rwhitten577, and @dangotbanned for reviewing

happy with the code now, and we've opened issues with Ibis for missing features on their end

let's ship it! 2 approval gifs here as this was an extra-large one

rwhitten577:
Awesome! Thanks @dangotbanned and @MarcoGorelli for reviewing!

PR #2007, opened 2025-02-13
fix(typing): Resolve all `mypy` & `pyright` errors for `_arrow`

### LINKED ISSUE #1961: CI: get `mypy` passing with `pyarrow-stubs` installed

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #1961

## Checklist

- [ ] Tests added
- [x] Documented the changes

- [~~Planning to finish getting the `mypy` errors to 0~~](https://github.com/narwhals-dev/narwhals/pull/2007#issuecomment-2657341603), then ~~tidying things up~~


dangotbanned:
@MarcoGorelli Almost out of the rabbit hole on this!

I've found some more places where (https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-12054784) would be pretty helpful:

https://github.com/narwhals-dev/narwhals/blob/1d24dec95351126c968922ae44e98aa874c1c1af/narwhals/_arrow/series.py#L126-L160

https://github.com/narwhals-dev/narwhals/blob/1d24dec95351126c968922ae44e98aa874c1c1af/narwhals/_arrow/series.py#L549-L552

Essentially anywhere that `narwhals` object would "change" its `TypeVar`, the current `self.__class__(...)` route breaks the typing.

So for `ArrowSeries[T1]`, you can't go to `ArrowSeries[T2]` without a `@classmethod` that removes `T1` from scope.

---

My brain has fully melted working on this, hope the above made sense 🫠 
If not (https://typing.readthedocs.io/en/latest/spec/generics.html#scoping-rules-for-type-variables)

MarcoGorelli:
thanks for working on this

is it necessary to make `ArrowSeries` generic in the PyArrow type? would it work to just keep that out for now?

narwhals/_arrow/group_by.py
```diff
@@ -139,15 +140,26 @@ def agg(self: Self, *exprs: ArrowExpr) -> ArrowDataFrame:
...
+            key_values = pc.binary_join_element_wise(
+                *it, null_handling="replace", null_replacement=null_token
+            )
+        else:
+            key_values = pc.binary_join_element_wise(
+                *it,
+                "",
+                null_handling="replace",
+                null_replacement=null_token,
+            )
```
dangotbanned, line 162: Unless the stubs are wrong here, the original code seems to add the `""` on the end?

I've done this to preserve the current behavior at runtime - but haven't tested if the `TYPE_CHECKING` would produce the same output

narwhals/_arrow/namespace.py
```diff
@@ -351,16 +361,21 @@ def concat_str(
...
+            it = (s._native_series for s in compliant_series_list)
+            if TYPE_CHECKING:
+                # NOTE: stubs indicate `separator` would get appended to the end, instead of between elements
+                result_series = pc.binary_join_element_wise(
+                    *it, null_handling=null_handling
+                )
+            else:
+                result_series = pc.binary_join_element_wise(
+                    *it, separator, null_handling=null_handling
+                )
```
dangotbanned, line 378: Similar to https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1954877868

dangotbanned:
> thanks for working on this
> 
> is it necessary to make `ArrowSeries` generic in the PyArrow type? would it work to just keep that out for now?

100% needed to resolve the issues I'm afraid 😔

Without the `TypeVar`, most of the `@overload`s end matching `Expression`.
This was the main source of errors, since we don't appear to use `Expression` anywhere, it just introduces a huge amount of noise 

--- 

Having read through a lot of the code (but not using `pyarrow` much personally) I am curious as to why we've not used `Expression` much/at all?

It seems to be available in our min version

https://arrow.apache.org/docs/11.0/python/generated/pyarrow.dataset.Expression.html

dangotbanned:
Down to **32 errors** 😁!!!!!!!!!!!

https://github.com/narwhals-dev/narwhals/actions/runs/13313871223/job/37182957093?pr=2007

narwhals/_arrow/series.py
```diff
@@ -247,65 +318,67 @@ def max(self: Self, *, _return_py_scalar: bool = True) -> Any:
...
-            result = pa.concat_arrays([pa.nulls(n, ca.type), *ca[:-n].chunks])
+            arrays = [pa.repeat(None, n).cast(self._type), *ca[:-n].chunks]
         elif n < 0:
-            result = pa.concat_arrays([*ca[-n:].chunks, pa.nulls(-n, ca.type)])
+            arrays = [*ca[-n:].chunks, pa.repeat(None, -n).cast(self._type)]
         else:
-            result = ca
-        return self._from_native_series(result)
+            return self._from_native_series(ca)
+        return self._from_native_series(pa.concat_arrays(arrays))
```
dangotbanned, line 345: ### TODO
- [ ] Investigate [why this was an issue](https://github.com/narwhals-dev/narwhals/actions/runs/13317234868/job/37194138357?pr=2007) in `3.9` only
- [ ] Probably replace this *solution* with a a re-wrapped/typed `pa.nulls()`

narwhals/_arrow/typing.py
```diff
@@ -11,7 +12,40 @@
...
+Marker for working code that fails on the stubs.
+
+Common issues:
+- Annotated for `Array`, but not `ChunkedArray`
+- Relies on typing information that the stubs don't provide statically
+- Missing attributes
+- Incorrect return types
+- Inconsistent use of generic/concrete types
+- `_clone_signature` used on signatures that are not identical
+"""
```
dangotbanned, line 48: > [!NOTE]
>  See for more context https://github.com/python/typeshed/blob/a410f251ee535950eb4440ad41ae4c00c87e6a67/stdlib/_typeshed/__init__.pyi#L45-L50

I've been sprinkling these in with a comment when all else fails, e.g:

https://github.com/narwhals-dev/narwhals/blob/1b537e75027ff69eaf5bb8f2bd5d88e9a6897a83/narwhals/_arrow/series.py#L486-L490

If the stub issues get resolved in the future, this will be a lot easier to fix than just using `Any` directly

narwhals/_interchange/dataframe.py
```diff
@@ -127,6 +127,8 @@ def to_pandas(self: Self) -> pd.DataFrame:
     def to_arrow(self: Self) -> pa.Table:
         from pyarrow.interchange import from_dataframe  # ignore-banned-import()
 
+        # NOTE: mypy: error: Module not callable  [operator]
+        # NOTE: pyright: Module is not callable [reportCallIssue]
         return from_dataframe(self._interchange_frame)
```
dangotbanned, line 132: ### TODO
- Figure out whether the stubs are accurate
- Maybe `from_dataframe` is **both** and module and a function?

dangotbanned: https://github.com/apache/arrow/blob/4f71de3cbf13adc72fd56c64a170f8bb3fe8c2d9/python/pyarrow/interchange/__init__.py#L20

```py
from .from_dataframe import from_dataframe
```

😩

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/4d4a7220006cf3517a4946eaa117e7eee0954617)

narwhals/_arrow/series.py
dangotbanned: > [!IMPORTANT]
> I only ever got that failure on a single CI run.
> Still planning to redo this to avoid the *potential* of introducing an error

narwhals/_arrow/series.py
```diff
@@ -109,133 +166,151 @@ def __narwhals_namespace__(self: Self) -> ArrowNamespace:
...
+        _return_py_scalar: bool = True,
+    ) -> float:
+        return maybe_extract_py_scalar(pc.mean(self._native_series), _return_py_scalar)
 
-    def median(self: Self, *, _return_py_scalar: bool = True) -> int:
+    def median(
+        self: ArrowSeries[pa.FloatScalar] | ArrowSeries[pa.DoubleScalar],
+        *,
+        _return_py_scalar: bool = True,
+    ) -> float:
```
dangotbanned, line 306: ### TODO
- [ ] Can these use numeric types besides `(Float|Double)Scalar`?

narwhals/_arrow/series.py
```diff
@@ -247,65 +322,67 @@ def max(self: Self, *, _return_py_scalar: bool = True) -> Any:
...
-        return maybe_extract_py_scalar(  # type: ignore[no-any-return]
+        return maybe_extract_py_scalar(
             pc.variance(self._native_series, ddof=ddof), _return_py_scalar
         )
 
     def skew(self: Self, *, _return_py_scalar: bool = True) -> float | None:
         ser = self._native_series
-        ser_not_null = pc.drop_null(ser)
+        # NOTE: stub issue with `pc.subtract`, `pc.mean` and `pa.ChunkedArray`
+        ser_not_null: Incomplete = ser.drop_null()
```
dangotbanned, line 364: > [!NOTE]
> stub issue with `pc.subtract`, `pc.mean` and `pa.ChunkedArray`

narwhals/_arrow/series.py
```diff
@@ -377,44 +460,60 @@ def alias(self: Self, name: str) -> Self:
...
         return self._from_native_series(pc.abs(self._native_series))
 
-    def cum_sum(self: Self, *, reverse: bool) -> Self:
+    def cum_sum(
+        self: ArrowSeries[_NumericScalarT], *, reverse: bool
+    ) -> ArrowSeries[_NumericScalarT]:
         native_series = self._native_series
+        # NOTE: stub only permits `NumericArray`
+        # https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L140
+        cum_sum: Incomplete = pc.cumulative_sum
```
dangotbanned, line 472: > [!NOTE]
> stub only permits `NumericArray`
https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L140

narwhals/_arrow/series.py
```diff
@@ -377,44 +460,60 @@ def alias(self: Self, name: str) -> Self:
...
         )
         return self._from_native_series(result)
 
-    def round(self: Self, decimals: int) -> Self:
+    def round(
+        self: ArrowSeries[_NumericScalarT], decimals: int
+    ) -> ArrowSeries[_NumericScalarT]:
+        # NOTE: stub only permits `NumericArray`
+        # https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L140
+        pc_round: Incomplete = pc.round
```
dangotbanned, line 485: > [!NOTE]
> stub only permits `NumericArray`
https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L140

narwhals/_arrow/series.py
```diff
@@ -377,44 +460,60 @@ def alias(self: Self, name: str) -> Self:
...
+        pc_any: Incomplete = pc.any
+        return maybe_extract_py_scalar(
+            pc_any(self._native_series, min_count=0), _return_py_scalar
         )
 
     def all(self: Self, *, _return_py_scalar: bool = True) -> bool:
-        return maybe_extract_py_scalar(  # type: ignore[no-any-return]
-            pc.all(self._native_series, min_count=0), _return_py_scalar
+        # NOTE: stub restricts to `BooleanArray`, should be based on truthiness
+        pc_all: Incomplete = pc.all
```
dangotbanned, line 506: > [!NOTE]
> stub restricts to `BooleanArray`, should be based on truthiness
> `pc.any` copies `pc.all` 
(https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L215)

narwhals/_arrow/series.py
```diff
@@ -728,7 +820,8 @@ def to_dummies(self: Self, *, separator: str, drop_first: bool) -> ArrowDataFram
 
         series = self._native_series
         name = self._name
-        da = series.dictionary_encode(null_encoding="encode").combine_chunks()
+        # NOTE: stub is missing attributes (https://arrow.apache.org/docs/python/generated/pyarrow.DictionaryArray.html)
+        da: Incomplete = series.combine_chunks().dictionary_encode(null_encoding="encode")
```
dangotbanned, line 824: > [!NOTE]
> stub is missing attributes `dictionary`, `indices`
(https://arrow.apache.org/docs/python/generated/pyarrow.DictionaryArray.html)

narwhals/_arrow/series.py
```diff
@@ -762,33 +855,36 @@ def quantile(
...
             self, lower_bound, self._backend_version
         )
         _, upper_bound = broadcast_and_extract_native(
             self, upper_bound, self._backend_version
         )
-        arr = pc.max_element_wise(arr, lower_bound)
+        # NOTE: stubs are missing `ChunkedArray` support
+        # https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L948-L954
+        max_element_wise: Incomplete = pc.max_element_wise
+        arr = max_element_wise(arr, lower_bound)
```
dangotbanned, line 879: > [!NOTE]
> stubs are missing `ChunkedArray` support
(https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L948-L954)

narwhals/_arrow/series.py
```diff
@@ -955,6 +1051,9 @@ def rolling_var(
         count_in_window = valid_count - valid_count.shift(window_size).fill_null(
             value=0, strategy=None, limit=None
         )
+        # NOTE: stubs are missing `ChunkedArray` support
+        # https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L948-L954
+        max_element_wise: Incomplete = pc.max_element_wise
```
dangotbanned, line 1056: Same as https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1956150928

narwhals/_arrow/series.py
```diff
@@ -1026,33 +1127,47 @@ def hist(  # noqa: PLR0915
...
+                    cast("pc.NumericOrTemporalArray", self._native_series), lower
+                ),
+                width,
+            )
+            bin_indices: pa.ChunkedArray[Any] = cast(
+                "pa.ChunkedArray[Any]", pc.floor(bin_proportions)
+            )
+
+            # NOTE: stubs leave unannotated
+            if_else: Incomplete = pc.if_else
```
dangotbanned, line 1157: > [!NOTE]
> stubs introduce an `Unknown` here
(https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L1829)

narwhals/_arrow/series.py
```diff
@@ -1061,38 +1176,31 @@ def _hist_from_bin_count(
...
+                # align bin ids to all possible bin ids (populate in missing bins)
+                .join(possible, keys="values", join_type="right outer")
                 .sort_by("values")
             )
-            counts = counts.set_column(  # empty bin intervals should have a 0 count
-                0, "counts", pc.coalesce(counts.column("counts"), 0)
+            # empty bin intervals should have a 0 count
+            counts_coalesce = cast(
+                "pa.Array[Any]",
+                pc.coalesce(cast("pa.Array[Any]", counts.column("counts")), lit(0)),
```
dangotbanned, line 1186: > [!NOTE]
> stubs missing support for `pa.ChunkedArray`
https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L200
https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/compute.pyi#L1826-L1828

narwhals/_arrow/series.py
```diff
@@ -1026,33 +1127,47 @@ def hist(  # noqa: PLR0915
 
         from narwhals._arrow.dataframe import ArrowDataFrame
 
-        def _hist_from_bin_count(
-            bin_count: int,
-        ) -> tuple[Sequence[int], Sequence[int | float], Sequence[int | float]]:
+        def _hist_from_bin_count(bin_count: int):  # type: ignore[no-untyped-def] # noqa: ANN202
```
dangotbanned, line 1130: The whole of `ArrowSeries.hist` is too complex to bother with typing nested inner-method return types.

Falling back to inference is fine.

Kinda concerned though that `bin_left` is returned and never used?

https://github.com/narwhals-dev/narwhals/blob/3bdf0e8b024b4700f8b6adf443a033274369a1a5/narwhals/_arrow/series.py#L1214-L1224

narwhals/_arrow/series.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/27da267556c7229b61e320a793729c090753e834)

narwhals/_arrow/group_by.py
dangotbanned: https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join_element_wise.html#pyarrow-compute-binary-join-element-wise

> Join string arguments together, with the last argument as separator.
> 
> Concatenate the strings except for the last one. The last argument in strings is inserted between each given string

Well that's unintuitive :weary:

So using this function *as-intended*, results in a warning from a type checker

narwhals/_arrow/series.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/c43d09263fa546cdc061d41584e8511cabc6c3e4)

dangotbanned:
> is it necessary to make `ArrowSeries` generic in the PyArrow type? would it work to just keep that out for now?

@MarcoGorelli had a bit of an interesting development with this.
So, regardless of how accurate `pyarrow-stubs` is - by annotating the `ArrowSeries` methods we can at least chain those together with accurate static feedback.

I haven't gone too deep into that, but have made a start with:
- [`series_str.py`](https://github.com/narwhals-dev/narwhals/pull/2007/files#diff-d6322061ea7eab03471b03326394eab426be6f7dabb2d49f34df157f751b4936)
- [`series_dt.py`](https://github.com/narwhals-dev/narwhals/pull/2007/files#diff-2f9581b78e3c21fdbcd18dd70e42bd184ff4c62737b2764de9955dd958c837f4)
- some of the easier ones in [`series.py`](https://github.com/narwhals-dev/narwhals/pull/2007/files#diff-f595265984b673e1c612607ebc3426434ee7162684840b99cd73fcce59fd6a7f)

narwhals/_arrow/dataframe.py
```diff
@@ -532,9 +534,9 @@ def with_row_index(self: Self, name: str) -> Self:
             df.append_column(name, row_indices).select([name, *cols])
         )
 
-    def filter(self: Self, predicate: ArrowExpr | list[bool]) -> Self:
```
MarcoGorelli, line 535: why `None`? i don't think we should allow that?

narwhals/_arrow/group_by.py
MarcoGorelli, line 162: this looks a bit odd to me - can we just `type: ignore` if the stubs are wrong and report upstream?

narwhals/_arrow/utils.py
```diff
@@ -11,18 +13,89 @@
...
+def nulls_like(
+    n: int, series: ArrowSeries[pa.Scalar[DataTypeT_co]]
+) -> pa.Array[pa.Scalar[DataTypeT_co]]:
+    """Create a strongly-typed Array instance with all elements null.
+
+    Uses the type of `series`, without upseting `mypy`.
+    """
+    if TYPE_CHECKING:
+        return pa.repeat(None, n).cast(series._type)
+    return pa.nulls(n, series._type)
```
MarcoGorelli, line 97: same comment, can we `type: ignore` and report upstream if there's a stubs error?

narwhals/_arrow/group_by.py
dangotbanned: @MarcoGorelli I discovered partly what was going on in https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1956407428

Realising I hadn't mentioned this anywhere 🤦‍♂️ - but I was hoping that all of these notes on the PR would something we can report upstream

narwhals/_arrow/dataframe.py
dangotbanned: image

This is very similar to #2012 

https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/_stubs_typing.pyi#L26

MarcoGorelli: Ah thanks!

narwhals/_arrow/group_by.py
MarcoGorelli: Sure, thanks - happy to get this in if we resolve the conditionals (and optionally report upstream)

narwhals/_arrow/utils.py
dangotbanned: @MarcoGorelli would you be okay with (https://github.com/narwhals-dev/narwhals/pull/2007/commits/adef6a00eb1b10cf4abcd9aba2a6de31eb1b45ce)


I prefer that, since we can easily find all refs like:
image

I would've added as a suggestion, but the import was outside the range

dangotbanned: Some more context in (https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1955246312)

narwhals/_arrow/series.py
```diff
@@ -723,11 +723,7 @@ def to_frame(self: Self) -> ArrowDataFrame:
         )
 
     def to_pandas(self: Self) -> pd.Series:
-        if TYPE_CHECKING:
-            return self._native_series.to_pandas()
-        import pandas as pd  # ignore-banned-import()
-
-        return pd.Series(self._native_series, name=self.name)
+        return self._native_series.to_pandas().rename(self._name)
```
dangotbanned, line 727: Huh, I really thought this was gonna fail - nice surprise

narwhals/_arrow/series.py
```diff
@@ -109,133 +164,148 @@ def __narwhals_namespace__(self: Self) -> ArrowNamespace:
...
+    def __invert__(self: ArrowSeries[pa.BooleanScalar]) -> ArrowSeries[pa.BooleanScalar]:
+        return self._from_native_series(
+            pc.invert(cast("pa.BooleanArray", self._native_series))
+        )
+
+    @property
+    def _type(self: ArrowSeries[pa.Scalar[DataTypeT_co]]) -> DataTypeT_co:
+        if TYPE_CHECKING:
+            return self._native_series[0].type
+        return self._native_series.type
```
dangotbanned, line 279: I don't think there's a way to get this working *without* the `TYPE_CHECKING` block.

I'm using this in a few places to resolve `ChunkedArray` erasing its `type` property:

https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/__lib_pxi/table.pyi#L59-L63

```py
def type(self) -> DataType: ...
```

The type is preserved for `Array`, so here I'm stealing the generic from that `type` property:

https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/__lib_pxi/array.pyi#L1058-L1070

```py
def type(self: Array[Scalar[_DataType_CoT]]) -> _DataType_CoT: ...
```

narwhals/_arrow/series.py
```diff
@@ -33,33 +40,77 @@
...
+@overload
+def maybe_extract_py_scalar(
+    value: pa.Scalar[Any] | Any,
+    return_py_scalar: bool,  # noqa: FBT001
+) -> Any: ...
+
 
 def maybe_extract_py_scalar(value: Any, return_py_scalar: bool) -> Any:  # noqa: FBT001
+    if TYPE_CHECKING:
+        return value.as_py()
```
dangotbanned, line 97: There's some similarities between this one and (https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1957124735)

Part of this is recreating a subset of the `@overload`(s) in 
https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/__lib_pxi/scalar.pyi#L62-L105

But I'm also needing to lie, since `.as_py()` isn't available in all versions.

For a lot of the cases where `maybe_extract_py_scalar` is used - this avoids needing to do a `[no-any-return]` - since we have `pa.Scalar[_BasicDataType[_AsPyType]]` provided by https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1957124735

MarcoGorelli (commented):
first, thanks a tonne for your efforts, really appreciate it

second, I think this might be doing too much -  mainly i'm concerned about there being both logic changes (e.g. `to_pandas`) and typing changes. can we keep them to separate PRs? i'm concerned about missing things with too large PRs

narwhals/_arrow/series.py
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2007#pullrequestreview-2619421873

@MarcoGorelli no worries, feel free to revert this part or I'll give it a go in a little bit

dangotbanned: Reverted in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/f8617ff1a0d3cfd0232512605dd1f653939129aa)

dangotbanned: > the ~~ignore(s)~~ will show up for `mypy` after https://github.com/narwhals-dev/narwhals/pull/2008 (I assume)

I meant the error(s)

narwhals/_arrow/group_by.py
```diff
@@ -68,7 +70,7 @@ def agg(self: Self, *exprs: ArrowExpr) -> ArrowDataFrame:
             )
             raise ValueError(msg)
 
-        aggs: list[tuple[str, str, pc.FunctionOptions | None]] = []
```
MarcoGorelli, line 71: why set this to `Any`? isn't it useful (at least for readability) to know what's inside the list?

dangotbanned: So there were a couple of issues here

`list` being invariant:

```
Argument of type "list[tuple[str, str, FunctionOptions | None]]" cannot be assigned to parameter "aggregations" of type "list[tuple[str, str]] | list[tuple[str, str, FunctionOptions]]" in function "aggregate"
  Type "list[tuple[str, str, FunctionOptions | None]]" is not assignable to type "list[tuple[str, str]] | list[tuple[str, str, FunctionOptions]]"
    "list[tuple[str, str, FunctionOptions | None]]" is not assignable to "list[tuple[str, str]]"
      Type parameter "_T@list" is invariant, but "tuple[str, str, FunctionOptions | None]" is not the same as "tuple[str, str]"
      Consider switching from "list" to "Sequence" which is covariant
    "list[tuple[str, str, FunctionOptions | None]]" is not assignable to "list[tuple[str, str, FunctionOptions]]"
      Type parameter "_T@list" is invariant, but "tuple[str, str, FunctionOptions | None]" is not the same as "tuple[str, str, FunctionOptions]"
      Consider switching from "list" to "Sequence" which is covariant
```

And the stubs annotation **not** including `None`:

```py
aggregations: list[tuple[str, str]] | list[tuple[str, str, FunctionOptions]]
```

https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/__lib_pxi/table.pyi#L642-L644

dangotbanned: `list[Any]` seemed like the most reasonable option to me.

Another one could be changing the runtime code to pass a 2-tuple instead of 3-tuple (w/ maybe `None`).
But that would still cause the invariant `list` issue

MarcoGorelli: `aggs: list[tuple[str, str, Any]]` type checks fine when I run this locally

dangotbanned: > `aggs: list[tuple[str, str, Any]]` type checks fine when I run this locally

Sounds good to me @MarcoGorelli , maybe add an alias for the `tuple[...]` part for readability?

dangotbanned: Then you'd have:

```py
aggs: list[Whatever]
```

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/ba2b3a5216aa7d8f2f5723cfb47193d736cbce2a)

tests/series_only/is_ordered_categorical_test.py
```diff
@@ -51,7 +53,8 @@ def test_is_definitely_not_ordered_categorical(
 
 @pytest.mark.xfail(reason="https://github.com/apache/arrow/issues/41017")
 def test_is_ordered_categorical_pyarrow() -> None:
-    s = pa.chunked_array(
-        [pa.array(["a", "b"], type=pa.dictionary(pa.int32(), pa.string(), ordered=True))]
-    )
-    assert nw.is_ordered_categorical(nw.from_native(s, series_only=True))
+    tp = pa.dictionary(pa.int32(), pa.string(), ordered=True)
+    s = pa.chunked_array([pa.array(["a", "b"], type=tp)], type=tp)
```
MarcoGorelli, line 57: this line is raising
```
pyarrow.lib.ArrowTypeError: Array chunks must all be same type
```

narwhals/typing.py
```diff
@@ -272,6 +273,8 @@ def __native_namespace__(self) -> ModuleType: ...
 _2DArray: TypeAlias = "_NDArray[tuple[int, int]]"  # noqa: PYI042, PYI047
 _AnyDArray: TypeAlias = "_NDArray[tuple[int, ...]]"  # noqa: PYI047
 
+_ScalarT_co = TypeVar("_ScalarT_co", bound="pa.Scalar", covariant=True)  # noqa: PYI018
```
MarcoGorelli, line 276: this is only used for pyarrow? i think this can go in `_arrow/typing.py`, no?

narwhals/_arrow/utils.py
```diff
@@ -11,18 +13,89 @@
...
+from narwhals.utils import is_compliant_expr
+from narwhals.utils import is_compliant_series
 from narwhals.utils import isinstance_or_issubclass
 
 if TYPE_CHECKING:
     from typing import TypeVar
 
+    from pyarrow.__lib_pxi import (  # pyright: ignore[reportMissingModuleSource]
+        types as pa_types,  # pyright: ignore[reportMissingModuleSource]
+    )
```
MarcoGorelli, line 25: why are we importing from `pyarrow.__lib_pxi`? this looks really private

narwhals/typing.py
dangotbanned: I think I put it there because it is used outside of `_arrow`?

MarcoGorelli: sure but having it in `narwhals/typing.py` named like this makes it looks like it's any Scalar (which we should probably introduce at some point rather than using `Any`), rather than a pyarrow scalar

narwhals/_arrow/utils.py
dangotbanned: There's a few places, but one is to redefine the type guards to work statically:

https://github.com/narwhals-dev/narwhals/blob/5dcd2f312621647ab7e915425ddb4b0e4fbb5a3d/narwhals/_arrow/utils.py#L47-L70

These are only useful at runtime:

https://github.com/zen-xu/pyarrow-stubs/blob/d97063876720e6a5edda7eb15f4efe07c31b8296/pyarrow-stubs/types.pyi

MarcoGorelli:
> [MarcoGorelli](https://github.com/MarcoGorelli) 100% needed to resolve the issues I'm afraid 😔
> Without the TypeVar, most of the @overloads end up matching Expression.

Would it work to introduce an alias for `ChunkedArray[Any]` and use that? Otherwise if we have to import from `pyarrow.__lib_pxi` then this risks breaking, and also becomes more complicated than what most contributors can probably handle

narwhals/typing.py
dangotbanned: > sure but having it in `narwhals/typing.py` named like this makes it looks like it's any Scalar (which we should probably introduce at some point rather than using `Any`), rather than a pyarrow scalar

That's fair @MarcoGorelli

dangotbanned: Made a typo, but resolved in (https://github.com/narwhals-dev/narwhals/pull/2007/commits/9c267c873ff14f9c17b32da60e3efb90885beb60)

narwhals/_arrow/utils.py
```diff
@@ -204,24 +272,18 @@ def broadcast_and_extract_native(
...
     return lhs._native_series, rhs
 
 
 def broadcast_and_extract_dataframe_comparand(
-    length: int,
-    other: Any,
-    backend_version: tuple[int, ...],
-) -> Any:
+    length: int, other: ArrowSeries, backend_version: tuple[int, ...]
+) -> pa.BooleanArray:
```
MarcoGorelli, line 286: 🤔 this doesn't have to be boolean, right? plenty of binary operations don't return booleans?

dangotbanned:
@MarcoGorelli I'm gonna leave you to it for now to avoid conflicts, but could you give me a shout when everything is ready please?

I'm getting warnings from `pyright` for some of these changes

image

MarcoGorelli:
sure, I think those pyright issues should be addressed now (and it would be really good to get pyright running in CI too)

downstream test failures should all be unrelated to this PR

MarcoGorelli (approved):
happy to get this in anyway, thanks @dangotbanned , really appreciate it!

dangotbanned:
> happy to get this in anyway, thanks @dangotbanned , really appreciate it!
> 
>  [ 68747470733a2f2f692e666c756666792e63632f32564e3573524e4c48306864733477676c4248666e73744a6c4a5843346d674a2e676966 ](https://camo.githubusercontent.com/80ca1eaaad9c14ad2aab79c88213205a9fe3fe9ad5056270a16c65eb5264e10d/68747470733a2f2f692e666c756666792e63632f32564e3573524e4c48306864733477676c4248666e73744a6c4a5843346d674a2e676966) [ ](https://camo.githubusercontent.com/80ca1eaaad9c14ad2aab79c88213205a9fe3fe9ad5056270a16c65eb5264e10d/68747470733a2f2f692e666c756666792e63632f32564e3573524e4c48306864733477676c4248666e73744a6c4a5843346d674a2e676966)

Thanks @MarcoGorelli, both `mypy` and `pyright` seem to be good with `_arrow` now 🎉

dangotbanned:
> sure, I think those pyright issues should be addressed now (and it would be really good to get pyright running in CI too)
> 
> downstream test failures should all be unrelated to this PR

@MarcoGorelli IIRC `marimo` has `pyright` in their CI - which could be worth a look

narwhals/_arrow/series_dt.py
```diff
@@ -172,46 +155,48 @@ def nanosecond(self: Self) -> ArrowSeries[pa.Int64Scalar]:
...
-    def total_milliseconds(self: Self) -> ArrowSeries[pa.Int64Scalar]:
-        ser: ArrowSeries[pa.TimestampScalar] = self._compliant_series
+    def total_milliseconds(self: Self) -> ArrowSeries:
+        ser: ArrowSeries = self._compliant_series
         arr = ser._native_series
-        unit = ser._type.unit
+        unit = ser._type.unit  # type: ignore[attr-defined]
         unit_to_milli_factor = {
             "s": 1e3,  # seconds
             "ms": 1,  # milli
```
dangotbanned: A bit dissapointed this change reintroduced the need for the `#type: ignore`s that were fixed

narwhals/dependencies.py
```diff
@@ -210,7 +211,9 @@ def is_polars_series(ser: Any) -> TypeIs[pl.Series]:
     return (pl := get_polars()) is not None and isinstance(ser, pl.Series)
 
 
-def is_pyarrow_chunked_array(ser: Any) -> TypeIs[pa.ChunkedArray]:
+def is_pyarrow_chunked_array(
+    ser: Any | ArrowChunkedArray,
+) -> TypeIs[ArrowChunkedArray]:
```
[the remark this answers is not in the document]
MarcoGorelli: ok, sure

MarcoGorelli:
Going to go ahead and merge this as I think it's a clear improvement, and we can address some smaller items like https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1958143258 and https://github.com/narwhals-dev/narwhals/pull/2007#discussion_r1958138696 as follow-ups

Thanks again Dan!

PR #2011, opened 2025-02-14
fix: sqlframe false positives in compliance checks

camriddell:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [ ] Documented the changes

In the `narwhals.from_native` function, `sqlframe` had to be explicitly checked before the `is_compliant*` checks due to false positives from `hasattr(sqlframe_df, "__narwhals_dataframe__")`.

1. Instead we can use `inspect.getattr_static` (thanks for the suggestion @dangotbanned) to avoid this addition and we can move the `sqlframe` specific code to be in line with the other backends.

2. I also added a test to ensure `sqlframe` cases are handled properly.
3. I removed a test that will break in the future, this test was passing due to a side effect caused by `hasattr` on a `pandas` DataFrame enabling the use of nonhashable values as column names (raised as an issue in pandas https://github.com/pandas-dev/pandas/issues/60925)

### DIFF
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -741,9 +741,7 @@ def _from_native_impl(  # noqa: PLR0915
             level="lazy",
         )
 
-    # SQLFrame
-    # This one needs checking before extensions as `hasattr` always returns `True`.
-    if is_sqlframe_dataframe(native_object):  # pragma: no cover
+    elif is_sqlframe_dataframe(native_object):  # pragma: no cover
         from narwhals._spark_like.dataframe import SparkLikeLazyFrame
 
         if series_only:
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -386,32 +386,8 @@ def _from_native_impl(  # noqa: PLR0915
         msg = "Invalid parameter combination: `eager_only=True` and `eager_or_interchange_only=True`"
         raise ValueError(msg)
 
-    # SQLFrame
-    # This one needs checking before extensions as `hasattr` always returns `True`.
-    if is_sqlframe_dataframe(native_object):  # pragma: no cover
-        from narwhals._spark_like.dataframe import SparkLikeLazyFrame
-
-        if series_only:
-            msg = "Cannot only use `series_only` with SQLFrame DataFrame"
-            raise TypeError(msg)
-        if eager_only or eager_or_interchange_only:
-            msg = "Cannot only use `eager_only` or `eager_or_interchange_only` with SQLFrame DataFrame"
-            raise TypeError(msg)
-        import sqlframe._version
-
-        backend_version = parse_version(sqlframe._version)
-        return LazyFrame(
-            SparkLikeLazyFrame(
-                native_object,
-                backend_version=backend_version,
-                version=version,
-                implementation=Implementation.SQLFRAME,
-            ),
-            level="lazy",
-        )
-
     # Extensions
-    elif is_compliant_dataframe(native_object):
+    if is_compliant_dataframe(native_object):
         if series_only:
             if not pass_through:
                 msg = "Cannot only use `series_only` with dataframe"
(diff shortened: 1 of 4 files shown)


camriddell:
@MarcoGorelli for the coverage is there a preference among
1. Add `sqlframe` to the pytest GitHub CI
2. Remove this test
3. `pragma: nocover` the tests

tests/translate/from_native_test.py
camriddell, line 1: @MarcoGorelli for the coverage is there a preference among
1. Add `sqlframe` to the pytest GitHub CI
2. Remove this test
3. `pragma: nocover` the tests

FBruzzesi: Just FYI, I am taking a look for how to fully support sqlframe - it's still missing a few features (not many) to pass all our tests. I would say that for now it's ok to `pragma: no cover` it

narwhals/translate.py
```diff
@@ -765,6 +741,30 @@ def _from_native_impl(  # noqa: PLR0915
             level="lazy",
         )
 
+    # SQLFrame
+    # This one needs checking before extensions as `hasattr` always returns `True`.
```
FBruzzesi, line 745: not applicable anymore 😉

tests/translate/from_native_test.py
MarcoGorelli: definitely interested in testing sqlframe in CI, last  time i checked it was too far to being passing - but they've resolved a lot of issues recently, so if we can run them in ci now then that would be great

else for now we can `no cover` the test

narwhals/utils.py
```diff
@@ -1275,16 +1276,24 @@ def dtype_matches_time_unit_and_time_zone(
     )
 
 
+def _hasattr_static(obj: Any, attr: str) -> bool:
+    try:
+        getattr_static(obj, attr)
+    except AttributeError:
+        return False
+    return True
```
FBruzzesi, line 1284: Not sure if you like it better but if we use a default value then:

```suggestion
    return getattr_static(obj, attr, None) is not None
```

or 

```suggestion
    return bool(getattr_static(obj, attr, None))
```

FBruzzesi (commented):
Thanks @camriddell 🚀 Left a couple of suggestion and (maybe) addressed your question

narwhals/utils.py
camriddell: What if the attribute exists and returns None? Not that I would expect our uses of `'__narwhals_*__'` to ever return `None`, but it would end up creating an edgecase. If you want to avoid the try/except we can create a short-lived sentinel of our own and check for that.

```python
from inspect import getattr_static

def hasattr_static(obj, attr):
    sentinel = object()
    return getattr_static(obj, attr, sentinel) is not sentinel

class T:
    x = None

print(
    bool(getattr_static(T, "x", None)), # False; though x exists as an attribute!
    hasattr_static(T, "x"),             # True; which is correct!
    sep='\n'
)
```

MarcoGorelli (approved):
you're a magician 🪄 

thanks! this may fix a `cudf.pandas` issue too!

MarcoGorelli:
pending on https://github.com/narwhals-dev/narwhals/pull/2011#discussion_r1956517442 😉

camriddell:
failing downstream tests seem to be unrelated. I know that `panel` just released a new version which may have broken `marimo` if they don't have an upperbound version constraint.

FBruzzesi:
Cool thanks! pointblank also just merged a new feature, we can update its dependency on another PR

narwhals/utils.py
dangotbanned: Sorry I'm late to this, but I'd vote for the `sentinel = object()` version

This is a pretty hot code path and for `<3.11` you'll be adding overhead for the exception handling https://bugs.python.org/issue40222

camriddell: I'll open up a PR with this change and we can have further discussion there!

PR #2012, opened 2025-02-14
chore(typing): resolve time unit/zone `set` invariance

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related PR (#1822)

## Checklist

- [ ] Tests added
- [ ] Documented the changes

image

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -40,7 +40,7 @@ from narwhals.exceptions import InvalidOperationError
 
 if TYPE_CHECKING:
     from types import ModuleType
-    from typing import AbstractSet
+    from typing import AbstractSet as Set
     from typing import Protocol
 
     import pandas as pd
@@ -1243,15 +1243,15 @@ def check_column_names_are_unique(columns: list[str]) -> None:
 def _parse_time_unit_and_time_zone(
     time_unit: TimeUnit | Iterable[TimeUnit] | None,
     time_zone: str | timezone | Iterable[str | timezone | None] | None,
-) -> tuple[set[TimeUnit], AbstractSet[str | None]]:
-    time_units: set[TimeUnit] = (
+) -> tuple[Set[TimeUnit], Set[str | None]]:
+    time_units: Set[TimeUnit] = (
         {"ms", "us", "ns", "s"}
         if time_unit is None
         else {time_unit}
         if isinstance(time_unit, str)
         else set(time_unit)
     )
-    time_zones: AbstractSet[str | None] = (
+    time_zones: Set[str | None] = (
         {None}
         if time_zone is None
         else {str(time_zone)}
@@ -1262,10 +1262,7 @@ def _parse_time_unit_and_time_zone(
 
 
 def dtype_matches_time_unit_and_time_zone(
-    dtype: DType,
-    dtypes: DTypes,
-    time_units: set[TimeUnit],
-    time_zones: AbstractSet[str | None],
+    dtype: DType, dtypes: DTypes, time_units: Set[TimeUnit], time_zones: Set[str | None]
 ) -> bool:
     return (
         (dtype == dtypes.Datetime)
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -39,6 +39,7 @@ from narwhals.exceptions import InvalidOperationError
 
 if TYPE_CHECKING:
     from types import ModuleType
+    from typing import AbstractSet
     from typing import Protocol
 
     import pandas as pd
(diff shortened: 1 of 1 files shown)


narwhals/utils.py
```diff
@@ -1241,15 +1242,15 @@ def check_column_names_are_unique(columns: list[str]) -> None:
 def _parse_time_unit_and_time_zone(
     time_unit: TimeUnit | Iterable[TimeUnit] | None,
     time_zone: str | timezone | Iterable[str | timezone | None] | None,
-) -> tuple[set[str], set[str | None]]:
-    time_units = (
+) -> tuple[set[TimeUnit], AbstractSet[str | None]]:
```
FBruzzesi, line 1245: Why wouldn't 

```suggestion
) -> tuple[set[TimeUnit], set[str | None]]:
```
not work?

I though [`AbstractSet` was deprecated](https://docs.python.org/3/library/typing.html#typing.AbstractSet)

FBruzzesi (commented):
Thanks @dangotbanned , I have a question for 50% of the changes 🙈

narwhals/utils.py
dangotbanned: Good question @FBruzzesi 

It very much is deprecated, but until we up our minimum python to `3.9` we can't use this as a generic https://docs.python.org/3/library/collections.abc.html#collections.abc.Set

`AbstractSet` and `Set` both have a covariant `TypeVar` in the stubs https://github.com/python/typeshed/blob/cc8ca939c0477a49fcce0554fa1743bd5c656a11/stdlib/typing.pyi#L627

FBruzzesi: I feel like I am missing something, how come that for one variable we can use it and not in the other?

Even with a py3.8 venv I am getting no issues with pyright after annotating the return type and the respective variables to `set[TimeUnit]` and `set[str | None]` 🤔

dangotbanned: > Even with a py3.8 venv I am getting no issues with pyright after annotating the return type and the respective variables to set[TimeUnit] and set[str | None] 🤔

Hmm, I'll need to take a look at this tomorrow.
IIRC, the difference was `set[TimeUnit]` can be used now because because it isn't a union.
But `set[str|None]` got expanded to `set[str] | set[None] | set[str | None]` - which was the issue

dangotbanned: `AbstractSet[str | None]` can match any of or all of those 3 (because of the variance)

FBruzzesi: Thanks for explaining! I am approving this in the meanwhile, but let me know if you manage to double check whenever you have the time 🙏🏼

FBruzzesi (approved):
Thanks @dangotbanned 🚀

NarwhalsKittyGIF

narwhals/utils.py
dangotbanned: Thanks @FBruzzesi will do!

Absolutely need some time away from typing after #2007 :dizzy_face:

dangotbanned: Okay so I know what the issue was now @FBruzzesi

By using `AbstractSet`, we can **safely** drop the inline `time_zones` annotation.

> [!NOTE]
> I forgot to do this though 🤦‍♂️  

<details><summary>Image 1</summary>
<p>

image

</p>
</details> 


We can *also* switch every annotation to `set[str | None]`, as you discovered in https://github.com/narwhals-dev/narwhals/pull/2012#discussion_r1956712085:

<details><summary>Image 2</summary>
<p>

image

</p>
</details> 



But this requires annotating it like that one every usage.
Dropping the inline annotation shows the warning:

<details><summary>Image 3</summary>
<p>

image

</p>
</details> 



Originally, I added a new alias for this - but later scrapped it since it only appears in these two functions:

```py
_TimeZones: TypeAlias = AbstractSet[str | None]]
```

---


However, leaving the variables to inference *only* works for `pyright` sadly:

```py
# mypy
narwhals/utils.py:1257: error: Argument 1 to <set> has incompatible type "str"; expected "None"  [arg-type]
narwhals/utils.py:1259: error: Set comprehension has incompatible type set[str | None]; expected set[str]  [misc]
narwhals/utils.py:1259: error: Set comprehension has incompatible type set[str | None]; expected set[None]  [misc]
```

I'm just pushing a little tweak that makes more sense to me

dangotbanned: See (https://github.com/narwhals-dev/narwhals/pull/2012/commits/b38403c0ecad54e39d6a471861412724a7313b26)

MarcoGorelli:
thanks all!

PR #2014, opened 2025-02-14
perf: avoid try/except in _hasattr_static

camriddell:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

In #2011 @dangotbanned pointed out that `try/except` is expensive in Python<3.11 and opted to use a sentinel to avoid those statements.

This codepath is triggered 3 times for every call to `from_native` so it will be regularly hit multiple times by end users so this could be worth doing.

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1277,11 +1277,8 @@ def dtype_matches_time_unit_and_time_zone(
 
 
 def _hasattr_static(obj: Any, attr: str) -> bool:
-    try:
-        getattr_static(obj, attr)
-    except AttributeError:
-        return False
-    return True
+    sentinel = object()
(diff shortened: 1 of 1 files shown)


FBruzzesi (approved):
Thanks @camriddell and thanks @dangotbanned for monitoring we don't introduce overheads 🚀

NarwhalDolphinGIF

dangotbanned:
Thanks @camriddell @FBruzzesi

Issue #2017, opened 2025-02-15
Allow to run `pytest narwhals --doctest-modules` without extra dependencies


FBruzzesi:
# Description

Reported in [discord thread](https://discord.com/channels/1235257048170762310/1237430779584446546/1340245707822465085) by @DeaMariaLeon 

`pytest narwhals --doctest-modules` ends up requiring extra dependencies such as ibis, although:

- `narwhals/_ibis/*.py` has no doctest to run (same as any other `narwhals/_<compliant>` module)
- We don't have any doctest with ibis

Can we use pytest [`norecursedir`](https://docs.pytest.org/en/stable/reference/reference.html#confval-norecursedirs) config to skip those files/directories in the first place without drawbacks?

dangotbanned:
I noticed that as well, managed to get it *"working"* with:
```cmd
pytest narwhals --doctest-modules --doctest-ignore-import-errors
```

One day, I'll install the timezone database for `pyarrow` 😉.
For now, I'm finding it helpful for spotting this issue (https://github.com/vega/altair/pull/3674):
```py
pyarrow.lib.ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "C:\Users\danie\Downloads\tzdata"
C:\Users\danie\Documents\GitHub\narwhals\narwhals\expr_dt.py:1282: UnexpectedException
====================================================== short test summary info ============
FAILED narwhals/dtypes.py::narwhals.dtypes.Datetime
FAILED narwhals/expr_dt.py::narwhals.expr_dt.ExprDateTimeNamespace.to_string
FAILED narwhals/series_dt.py::narwhals.series_dt.SeriesDateTimeNamespace.replace_time_zone
FAILED narwhals/series_dt.py::narwhals.series_dt.SeriesDateTimeNamespace.to_string
FAILED narwhals/dataframe.py::narwhals.dataframe.DataFrame.write_csv
FAILED narwhals/expr_dt.py::narwhals.expr_dt.ExprDateTimeNamespace.replace_time_zone
=============================================== 6 failed, 357 passed, 2 skipped in 8.24s ==
```

PR #2019, opened 2025-02-15
tests: Allow to run pytest narwhals --doctest-modules without ibis

DeaMariaLeon:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2017
- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

@FBruzzesi : I'm not sure if you meant (in the issue) that we should detect automatically the dependencies that we want to skip...? In case we only need to skip ibis, this does it. 

Per the link you shared about `norecursedirs` : "Default patterns are '*.egg', '.*', '_darcs', 'build', 'CVS', 'dist', 'node_modules', 'venv', '{arch}'. Setting a `norecursedirs` replaces the default."

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -165,7 +165,7 @@ required-imports = ["from __future__ import annotations"]
 docstring-code-format = true
 
 [tool.pytest.ini_options]
-norecursedirs =  ['*.egg', '.*', '_darcs', 'build', 'CVS', 'dist', 'node_modules', 'venv', '{arch}', "_ibis"]
+norecursedirs =  ['*.egg', '.*', '_darcs', 'build', 'CVS', 'dist', 'node_modules', 'venv', '{arch}', 'narwhals/_*']
 testpaths = ["tests"]
 filterwarnings = [
   "error",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -165,6 +165,7 @@ required-imports = ["from __future__ import annotations"]
 docstring-code-format = true
 
 [tool.pytest.ini_options]
+norecursedirs =  ['*.egg', '.*', '_darcs', 'build', 'CVS', 'dist', 'node_modules', 'venv', '{arch}', "_ibis"]
```
MarcoGorelli, line 168: thanks - should we add `_spark_like` / etc? would `"narwhals/_*"` work?

FBruzzesi: Something like `narwhals/_*` is what I had in mind but didn't check if that would work 🙃

DeaMariaLeon: 🙂.. it seems to work.

FBruzzesi (approved):
Thanks @DeaMariaLeon 🚀

NarwhalChubbiwhalGIF

MarcoGorelli (approved):
thanks!

DeaMariaLeon:
Two GIF's?! Thaaanks! I feel like a queen 👑

PR #1991, opened 2025-02-11
feat: support casting to and from spark-like structs

FBruzzesi:
## Reason

There are multiple reason for this PR to happen 😁

- Eventually I would like to support `Schema.to_pyspark`
- For some integration it might be useful to have: 
  - `nw.struct` emulating [`pl.struct`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.struct.html)
  - `.struct.unnest()` and/or `Frame.unnest`

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1743

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below

I am having a hard time testing this 🤔

PR #1807, opened 2025-01-13
feat: has_operation

camriddell:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1610 
- Closes #\<issue number\>

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below

Review Items
- [ ] Does this conflict with Narwhals internal stable/version API?
- [ ] Fix the [typing](https://github.com/narwhals-dev/narwhals/pull/1807/files#diff-9ce6eeb9a2683854ccb2d70d1590a00e0dddefa8da9f431394b7530d8c56fbfdR1212) here (specified return type is a shallow proxy to the real behavior) 

Wanted to get feedback for an idea I had to implement `has_operation`. The current API implements the following changes

- custom `metaproperty` decorator that behaves similarly to `property` (only the getter syntax other functionality would need to be implemented or would need to inherit from property itself).
  - enables chained syntax to move from an Expression/Series to its constituent namespaces without requiring an instance
  - `nw.Expr.dt.date` is valid syntax.
- add `has_operation` to check if a given function passed using the above syntax is available for any given backend.

What would be nice for this feature would be if these components were more readily accessible from the narwhals namespace, dynamic imports based on internalized mappings can become tricky to maintain (see `narwhals/utils:has_operation locals.backend_mapping`) Instead these expressions could be added to the Implementation enum providing more informative metadata (such as import location lookups as well as namespace import names).

---

Feedback on the concept as a whole or the specific implementation is welcome! If @MarcoGorelli likes this approach, I do believe some additional work is required here to get this working correctly with the internal versioning interface so guidance in this spot would be appreciated.

Here are a few examples to show what this PR contains

```python
from narwhals import has_operation
import narwhals as nw
import polars as pl
import pandas as pd
import pyarrow as pa
import duckdb
import dask.dataframe as dd


print(
    f"{has_operation(pl, nw.Expr.mean)                   =  }",  #  True
    f"{has_operation(pl, nw.Expr.dt.date)                =  }",  #  True
    f"{has_operation(pl, nw.Series.mean)                 =  }",  #  True
    f"{has_operation(pl, nw.Series.str.to_uppercase)     =  }",  #  True
[...]

PR #2025, opened 2025-02-16
fix(RFC): Use metaclass for safe `DType` attr access

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Mentioned in
- https://github.com/narwhals-dev/narwhals/pull/1991#discussion_r1957328569
- https://github.com/narwhals-dev/narwhals/pull/1807#discussion_r1956331625

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- I started with an implementation for `Datetime`, when used in `_polars` only
- Was expecting this to be a more complex task, but realised I could use the same idea as https://github.com/vega/altair/blob/f61a22e4fb0fb0822e7e3fb440b716d93a261474/altair/expr/__init__.py#L23

This gives you a read-only default for:

```py
Datetime.time_unit
```

Which is unaffected by the instance-level version:

```py
dtype = Datetime("ms")
dtype.time_unit
```

I imagined this would also be useful `Duration`:

https://github.com/narwhals-dev/narwhals/blob/43d072b41b8d4098c89b6798da46199e31368513/narwhals/dtypes.py#L527

Originally thought this would be beneficial for the `NestedType` classes, but I'm seeing that they *don't* have defaults

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -613,27 +613,26 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
         # convert to it?
         return "category"
     if isinstance_or_issubclass(dtype, dtypes.Datetime):
-        dt_time_unit = getattr(dtype, "time_unit", "us")
-        dt_time_zone = getattr(dtype, "time_zone", None)
-
         # Pandas does not support "ms" or "us" time units before version 2.0
-        # Let's overwrite with "ns"
         if implementation is Implementation.PANDAS and backend_version < (
             2,
         ):  # pragma: no cover
             dt_time_unit = "ns"
+        else:
+            dt_time_unit = dtype.time_unit
 
         if dtype_backend == "pyarrow":
-            tz_part = f", tz={dt_time_zone}" if dt_time_zone else ""
+            tz_part = f", tz={tz}" if (tz := dtype.time_zone) else ""
             return f"timestamp[{dt_time_unit}{tz_part}][pyarrow]"
         else:
-            tz_part = f", {dt_time_zone}" if dt_time_zone else ""
+            tz_part = f", {tz}" if (tz := dtype.time_zone) else ""
             return f"datetime64[{dt_time_unit}{tz_part}]"
     if isinstance_or_issubclass(dtype, dtypes.Duration):
-        du_time_unit = getattr(dtype, "time_unit", "us")
+        du_time_unit = dtype.time_unit
         if implementation is Implementation.PANDAS and backend_version < (
             2,
         ):  # pragma: no cover
+            # NOTE: I think this assignment was a typo, it never gets used (should be du_)
             dt_time_unit = "ns"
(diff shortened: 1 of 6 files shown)


narwhals/_polars/utils.py
```diff
@@ -190,13 +191,11 @@ def narwhals_to_native_dtype(
...
-    if dtype == dtypes.Datetime or isinstance(dtype, dtypes.Datetime):
-        dt_time_unit: TimeUnit = getattr(dtype, "time_unit", "us")
-        dt_time_zone = getattr(dtype, "time_zone", None)
-        return pl.Datetime(dt_time_unit, dt_time_zone)  # type: ignore[arg-type]
-    if dtype == dtypes.Duration or isinstance(dtype, dtypes.Duration):
-        du_time_unit: TimeUnit = getattr(dtype, "time_unit", "us")
-        return pl.Duration(time_unit=du_time_unit)  # type: ignore[arg-type]
+    if isinstance_or_issubclass(dtype, dtypes.Datetime):
+        # [s] is not allowed for polars
+        return pl.Datetime(dtype.time_unit, dtype.time_zone)  # type: ignore[arg-type]
```
dangotbanned, line 196: `time_unit="s"` isn't allowed for `polars`, but we pass it through anyway?

narwhals/_pandas_like/utils.py
```diff
@@ -613,27 +613,26 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
+            tz_part = f", {tz}" if (tz := dtype.time_zone) else ""
             return f"datetime64[{dt_time_unit}{tz_part}]"
     if isinstance_or_issubclass(dtype, dtypes.Duration):
-        du_time_unit = getattr(dtype, "time_unit", "us")
+        du_time_unit = dtype.time_unit
         if implementation is Implementation.PANDAS and backend_version < (
             2,
         ):  # pragma: no cover
+            # NOTE: I think this assignment was a typo, it never gets used (should be du_)
             dt_time_unit = "ns"
```
MarcoGorelli, line 636: good catch, thanks, should be `du`

MarcoGorelli (commented):
I didn't think i liked metaclasses, but this is pretty neat - thanks @dangotbanned !

narwhals/_pandas_like/utils.py
dangotbanned: Hopefully fixed in (https://github.com/narwhals-dev/narwhals/pull/2025/commits/50a43f3261d6c01923abd6e27037229a837eb44a)

I wasn't sure on changing it - in case it changed some behavior - but I suppose it would be fixing a bug

MarcoGorelli (approved):
thanks @dangotbanned !

PR #2028, opened 2025-02-16
perf: Avoid evaluating expr twice in mean_horizontal

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -237,14 +237,9 @@ class ArrowNamespace(CompliantNamespace[ArrowSeries]):
         dtypes = import_dtypes_module(self._version)
 
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
-            series = (
-                s.fill_null(0, strategy=None, limit=None)
-                for _expr in exprs
-                for s in _expr(df)
-            )
-            non_na = (
-                1 - s.is_null().cast(dtypes.Int64()) for _expr in exprs for s in _expr(df)
-            )
+            expr_results = [s for _expr in exprs for s in _expr(df)]
+            series = (s.fill_null(0, strategy=None, limit=None) for s in expr_results)
+            non_na = (1 - s.is_null().cast(dtypes.Int64()) for s in expr_results)
             return [reduce(operator.add, series) / reduce(operator.add, non_na)]
 
         return self._create_expr_from_callable(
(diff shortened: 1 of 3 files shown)


narwhals/_arrow/namespace.py
```diff
@@ -237,14 +237,9 @@ def mean_horizontal(self: Self, *exprs: ArrowExpr) -> IntoArrowExpr:
...
-                for _expr in exprs
-                for s in _expr(df)
-            )
-            non_na = (
-                1 - s.is_null().cast(dtypes.Int64()) for _expr in exprs for s in _expr(df)
-            )
+            expr_results = [s for _expr in exprs for s in _expr(df)]
+            series = (s.fill_null(0, strategy=None, limit=None) for s in expr_results)
+            non_na = (1 - s.is_null().cast(dtypes.Int64()) for s in expr_results)
             return [reduce(operator.add, series) / reduce(operator.add, non_na)]
```
dangotbanned, line 243: This doesn't keep the type checker happy, but does pass `pytest`?

```suggestion
            return [sum(series) / sum(non_na)]
```

dangotbanned: From https://docs.python.org/3/howto/functional.html#the-functools-module

MarcoGorelli: thanks, i wasn't aware that `sum` could be used like that

not sure if we should though - as you said, it fails type checking, but also, `sum` it usually used with numbers

dangotbanned: @MarcoGorelli I was also surprised! 

I remembered reading that page again recently and thought I'd give a try - no worries about keeping as is though

dangotbanned: @MarcoGorelli there's an upcoming `ruff` rule to be on the lookout for:

- https://github.com/astral-sh/ruff/pull/16131

Beyond this specific case, I think the cases like this would benefit:
`[a for b in c for a in b]`

Would be part of the ([`FURB`](https://docs.astral.sh/ruff/rules/#refurb-furb)) **preview** group, which I've quite liked using in https://github.com/vega/altair/blob/f61a22e4fb0fb0822e7e3fb440b716d93a261474/pyproject.toml#L156-L163

Issue #1818, opened 2025-01-16
refactor: use sqlframe instead of pyspark for testing?


MarcoGorelli:
PySpark is a massive dependency and hard to install

SQLFrame follows the same API and it much more lightweight. I earlier said it wasn't ready, but I reported some bugs to them and they've addressed all of them. It may be worth giving them another go?

FBruzzesi:
Not sure if using it as a complete alternative will suffice. In any case, SQLFrame is now running in CI 🚀

MarcoGorelli:
nice! maybe for now we can keep both?

FBruzzesi:
Yes I would say to keep them both - maybe we can have pyspark in a single run to make it more lightweight?

dangotbanned:
Both will be type checked in CI with https://github.com/narwhals-dev/narwhals/blob/11bde2e95c93d10f4cc04a389a259da2a05d1f6b/.github/workflows/typing.yml

MarcoGorelli:
we're using both now, so we can probably close this

PR #2038, opened 2025-02-18
chore: Finalize support for SQLFrame

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1818 

## Checklist

- [x] Tests added
- [ ] Documented the changes


narwhals/_spark_like/dataframe.py
```diff
@@ -42,7 +45,10 @@ def __init__(
         backend_version: tuple[int, ...],
         version: Version,
         implementation: Implementation,
+        validate_column_names: bool,
     ) -> None:
+        if validate_column_names:
+            check_column_names_are_unique(native_dataframe.columns)
```
FBruzzesi, line 51: While pyspark does it out of the box, sqlframe with duckdb backend doesn't.
I think it is good to have it regardless in order to standardize errors and error messages

narwhals/_spark_like/dataframe.py
```diff
@@ -413,16 +453,51 @@ def explode(self: Self, columns: list[str]) -> Self:
...
+                        self._F.col(col_name).alias(col_name)
+                        if col_name != columns[0]
+                        else self._F.explode_outer(col_name).alias(col_name)
+                        for col_name in column_names
+                    ]
+                ),
+                validate_column_names=False,
             )
-        )
+        elif self._implementation.is_sqlframe():
```
FBruzzesi, line 468: This is the same workaround as per duckdb native. As order is not guaranteed, I think that's ok

narwhals/_spark_like/expr.py
```diff
@@ -436,13 +447,9 @@ def _is_finite(_input: Column) -> Column:
 
     def is_in(self: Self, values: Sequence[Any]) -> Self:
         def _is_in(_input: Column) -> Column:
-            return _input.isin(values)
+            return _input.isin(values) if values else self._F.lit(False)  # noqa: FBT003
```
FBruzzesi, line 450: Related to #2031

tests/expr_and_series/cast_test.py
```diff
@@ -236,7 +236,9 @@ def test_cast_datetime_tz_aware(
 
 
 def test_cast_struct(request: pytest.FixtureRequest, constructor: Constructor) -> None:
-    if any(backend in str(constructor) for backend in ("dask", "modin", "cudf")):
+    if any(
+        backend in str(constructor) for backend in ("dask", "modin", "cudf", "sqlframe")
+    ):
```
FBruzzesi, line 241: Regardless having it more generic in line 256, I could not manage to make the test pass with sqlframe.
Might require investigation.

```terminal
self = <sqlframe.duckdb.session.DuckDBSession object at 0x7fcf53fa0380>
sql = 'WITH "t26313807" AS (SELECT CAST("a" AS MAP(TEXT, TEXT)) AS "a" FROM (VALUES (MAP([\'movie \', \'rating\'], [\'Cars\'...AS DOUBLE)} AS "a" FROM "t26313807") SELECT CAST("a" AS STRUCT("movie" TEXT, "rating" DOUBLE)) AS "a" FROM "t10314570"'

    def _execute(self, sql: str) -> None:
>       self._last_result = self._cur.execute(sql)  # type: ignore
E       duckdb.duckdb.BinderException: Binder Error: STRUCT to STRUCT cast must have at least one matching member

.venv/lib/python3.12/site-packages/sqlframe/duckdb/session.py:75: BinderException
```

FBruzzesi:
CI failure is.... interesting?! 🤔

narwhals/_spark_like/expr.py
MarcoGorelli: does it work to use `contains`, as per the latest change?

FBruzzesi: I might have missed a lot this week. Not sure which change you refer to 🤔

FBruzzesi: Ok I might be missing something - pyspark contains has different behaviour, and for sqlframe I rather not specialize for a specific dialect

MarcoGorelli: ah ok if pyspark's contains is different let's keep it like this 👍

FBruzzesi: Just to expand a bit on this: [pyspark.sql.functions.contains](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.contains.html#pyspark-sql-functions-contains) would be the equivalent of `.str.contains`

eakmanrq:
Thanks for the tag and sorry for the delay I didn't see this notification at first.

This does look strange. If I'm reading the error correctly, it seems like the first time SQLFrame is used, regardless of test, this warning from numpy is thrown. Purest is seeing this warning as unhandled and therefore considering the test a failure. So I believe the test is actually passing in terms of the logic but pytest isn't happy about the raised numpy warning. 

I'm trying to figure out why though. One thing I noticed is that you are using DuckDB 1.2.0 here and SQLFrame technically doesn't support that yet. I will be adding it soon but was waiting on a SQLGlot release with a fix. I can't explain though why that matters. 

Looking through SQLFrame I don't use numpy directly and I'm not using private methods. So I can't really explain this yet. Also it looks like the 3.11 test would pass if it wasn't cancelled so this could be specific to 3.13. 

Let me add DuckDB 1.2.0 support to SQLFrame and see if that luckily solves this. Otherwise we might need to see if we can repro locally and figure out where exactly this numpy warnings coming from.

FBruzzesi:
Thanks @eakmanrq for your time and help!

Let me try to help with some more context:

> * See if it can reproduce for you locally when running 3.13?

Locally I can replicate the following:

- Failure happens **only** for python 3.13, both for duckdb 1.2 and 1.1.3
- No issues for python<3.13 with both duckdb versions

> This does look strange. If I'm reading the error correctly, it seems like the first time SQLFrame is used, regardless of test, this warning from numpy is thrown. Purest is seeing this warning as unhandled and therefore considering the test a failure. So I believe the test is actually passing in terms of the logic but pytest isn't happy about the raised numpy warning.

Yes that seems to be exactly the case - There are cases in which we filter warning in tests, yet this might be a tricky one since:

- It depends on order of test collection/execution
- We are not too happy to filter it globally

> Looking through SQLFrame I don't use numpy directly and I'm not using private methods. So I can't really explain this yet. Also it looks like the 3.11 test would pass if it wasn't cancelled so this could be specific to 3.13.

Yes and yes!

> Let me add DuckDB 1.2.0 support to SQLFrame and see if that luckily solves this. Otherwise we might need to see if we can repro locally and figure out where exactly this numpy warnings coming from.

As a fairly minimal repro, this issue is independent of what operation is run and seems associated with pytest. I can run the following to reproduce in python 3.13:

**repro_test.py**

```py


def test_repro():
    session = DuckDBSession()

    data = {"foo": [1, 3, 2]}
    session.createDataFrame([*zip(*data.values())], schema=[*data.keys()])
```

```terminal
pytest repro_test.py
```

```terminal
uv pip freeze | grep -E "duck|sqlframe"                                  

Using Python 3.13.0 environment at: .venv13
duckdb==1.1.3
sqlframe==3.22.0
```

eakmanrq:
Thanks @FBruzzesi that repro is very helpful. I indeed could reproduce locally and this is actually a duckdb only repro (doesn't involve SQLFrame): 

```
import duckdb
from duckdb.typing import VARCHAR


def test_repro():
    conn = duckdb.connect()
    conn.create_function("blah", lambda x: x, return_type=VARCHAR)
```

I actually think the issue is this line in DuckDB which triggers this warning in later versions of numpy: https://github.com/duckdb/duckdb/blob/cd0d0da9b1f475632a21e11a7c00cc726bedaacc/tools/pythonpkg/src/python_udf.cpp#L485

This is the line in SQLFrame that triggers it: https://github.com/eakmanrq/sqlframe/blob/main/sqlframe/duckdb/session.py#L49

Currently SQLFrame just creates a Python UDF for `soundex` support which is a fairly niche function so I could potentially remove it. On the other hand though I may add more Python UDFs in the future if needed in order to get full PySpark compatibility support. So even if I remove this now I could end up adding a UDF in the future. I think the core issue here is that DuckDB should update their code to be better compatible with later versions of `numpy`. 

Let me know what you think.

MarcoGorelli:
thanks all for investigations!

shall we report this to duckdb, and here just silence the warning in `pyproject.toml`? i don't think there's anything we can do about it on the narwhals side?

eakmanrq:
Yeah I think silencing on Narwhal end is the right choice. Up to you if you want to open an issue on DuckDB Github.

FBruzzesi:
@eakmanrq I can't thank you enough for your support! and thanks @MarcoGorelli for reviewing!

I committed the change to ignore the warning for now and opened an issue in duckdb ([duckdb#16370](https://github.com/duckdb/duckdb/issues/16370))

tests/conftest.py
```diff
@@ -210,7 +210,7 @@ def sqlframe_pyspark_lazy_constructor(
     "pyspark": pyspark_lazy_constructor,  # type: ignore[dict-item]
     # We've reported several bugs to sqlframe - once they address
     # them, we can start testing them as part of our CI.
```
MarcoGorelli, line 212: remove this?

MarcoGorelli (approved):
thanks @FBruzzesi - just a minor comment then feel free to merge

regarding duplicate column names, we probably want to avoid validating them at every step (see https://github.com/narwhals-dev/narwhals/issues/2091) but that can be a follow-up

tests/conftest.py
FBruzzesi:
```suggestion
```

PR #2049, opened 2025-02-19
refactor(typing): Avoid more `ignore[attr-defined]` on `Datetime`

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

Made possible by
- #2025
- #1997

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2050, opened 2025-02-19
feat(DRAFT): `DType.__eq__` as a `TypeIs[...]`

dangotbanned:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/2049#discussion_r1961925131
- https://github.com/narwhals-dev/narwhals/pull/2049#discussion_r1961957453

## Checklist

- [ ] Tests added
- [ ] Documented the changes

https://typing.readthedocs.io/en/latest/spec/narrowing.html
> Type checkers should assume that type narrowing should be applied to the expression that is passed as the first positional argument to a user-defined type guard.
> If the type guard function accepts more than one argument, no type narrowing is applied to those additional argument expressions.
> 
> If a type guard function is implemented as an instance method or class method, the **first positional argument maps to the second parameter (after “self” or “cls”)**.

So `__eq__` currently describes the **opposite** of the spec:

```py
import narwhals as nw

dtype = nw.Boolean()
second_parameter = nw.Int64

dtype == second_parameter
dtype.__eq__(second_parameter)

# Equivalent to
isinstance_or_issubclass(second_parameter, type(dtype))
isinstance_or_issubclass(nw.Int64, nw.Boolean)
issubclass(nw.Int64, nw.Boolean)
```


narwhals/dtypes.py
```diff
@@ -458,7 +472,7 @@ def time_zone(cls) -> str | None:
         return None
 
 
-class Datetime(TemporalType, metaclass=_DatetimeMeta):
+class Datetime(TemporalType, Generic["UnitT", "ZoneT"], metaclass=_DatetimeMeta):
```
dangotbanned, line 475: ### TODO
- Change this to a `typing.TypeVar`, without a default
- Move that `TypeVar` out of the type checking block

dangotbanned: Nice that `mypy` is the only job that passed for once https://github.com/narwhals-dev/narwhals/actions/runs/13419148834/job/37487353343?pr=2050#step:8:1

utils/check_api_reference.py
```diff
@@ -215,7 +215,7 @@
 
 # DTypes
 dtypes = [
-    i for i in nw.dtypes.__dir__() if i[0].isupper() and not i.isupper() and i[0] != "_"
+    i for i in nw.dtypes.__all__ if i[0].isupper() and not i.isupper() and i[0] != "_"
```
dangotbanned, line 218: ## FIXME

Need some help getting this work again 
https://results.pre-commit.ci/run/github/760058710/1739990897.pkejAgUjQ3CdKYXvgIDqPw

I don't understand what the two checks are doing https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes

Or why this seems to include `Literal`?

https://github.com/narwhals-dev/narwhals/blob/228e478dab65685adb2c421963805af4ee8b247c/utils/check_api_reference.py#L42-L49

dangotbanned:
@MarcoGorelli I've added a little write-up for what I believe is the blocking issue.

Maybe this is a naive question, why was `__eq__` chosen for this instead of `__(instance|subclass)check__`?
https://docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks

dangotbanned:
Closing as not planned

MarcoGorelli:
> @MarcoGorelli I've added a little write-up for what I believe is the blocking issue.
> 
> Maybe this is a naive question, why was `__eq__` chosen for this instead of `__(instance|subclass)check__`? https://docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks

thanks @dangotbanned for looking into this, sorry i missed it

we just mirrored what Polars does with regards to `__eq__`

dangotbanned:
> > @MarcoGorelli I've added a little write-up for what I believe is the blocking issue.
> > Maybe this is a naive question, why was `__eq__` chosen for this instead of `__(instance|subclass)check__`? [docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks](https://docs.python.org/3/reference/datamodel.html#customizing-instance-and-subclass-checks)
> 
> thanks @dangotbanned for looking into this, sorry i missed it
> 
> we just mirrored what Polars does with regards to `__eq__`

@MarcoGorelli
Oh wow, can't believe I didn't realise that 🤦‍♂️

All of this is near-identical to `polars`, including a comment 😉 

https://github.com/pola-rs/polars/blob/0aa889bef845ed902c634f900d4ed703c2eb55aa/py-polars/polars/datatypes/classes.py#L528-L562

https://github.com/narwhals-dev/narwhals/blob/2bcc6bbfe188889c15ae7eab544f5ba5e91ba695/narwhals/dtypes.py#L335-L367

PR #2051, opened 2025-02-20
chore(typing): enable typing checks for `pyspark`

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes


tests/frame/unpivot_test.py
```diff
@@ -106,3 +106,14 @@ def test_unpivot_mixed_types(
...
     assert result.collect_schema().dtypes() == expected_dtypes
+
+
+def test_unpivot_index_none(constructor: Constructor) -> None:
+    df = nw.from_native(constructor(data))
+    expected_data = {
+        "variable": ["b", "b", "b", "c", "c", "c"],
+        "value": [1, 3, 5, 2, 4, 6],
+    }
+    result = df.unpivot(on=["b", "c"], index=None)
```
EdAbati, line 117: looks like pyspark doesn't want a `None` index

dangotbanned:
@EdAbati do we want this merged before or after #2038?

narwhals/_spark_like/expr.py
```diff
@@ -473,7 +473,7 @@ def is_null(self: Self) -> Self:
 
     def is_nan(self: Self) -> Self:
         def _is_nan(_input: Column) -> Column:
-            return self._F.when(self._F.isnull(_input), None).otherwise(
+            return self._F.when(self._F.isnull(_input), None).otherwise(  # type: ignore[no-any-return]
```
dangotbanned, line 476: We should add `*._pyspark.*` to the rule I changed here [`8c02a95` (#2077)](https://github.com/narwhals-dev/narwhals/pull/2077/commits/8c02a9534ffef6771dd3885da68f8edff427206a)

https://github.com/narwhals-dev/narwhals/blob/8609786a1cdc29055cc422255736c8688cec0a5a/pyproject.toml#L247-L255

narwhals/_spark_like/expr.py
```diff
@@ -27,7 +27,7 @@
     from narwhals.utils import Version
 
 
-class SparkLikeExpr(CompliantExpr["SparkLikeLazyFrame", "Column"]):
+class SparkLikeExpr(CompliantExpr["SparkLikeLazyFrame", "Column"]):  # type: ignore[type-var]
```
dangotbanned, line 30: Do you think it would be helpful to link the PR about why we need to do this?

I did that in (https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1967281895) for `DaskExpr` btw

narwhals/_spark_like/dataframe.py
```diff
@@ -152,12 +152,12 @@ def _collect_to_arrow(self) -> pa.Table:
                 else:  # pragma: no cover
                     raise
         else:
-            native_pyarrow_frame = self._native_frame.toArrow()
+            native_pyarrow_frame = self._native_frame.toArrow()  # type: ignore[operator]
```
dangotbanned, line 155: This one is interesting :thinking:

I'm guessing the `[operator]` refers to `__call__`?

I think it is pointing to `self._native_frame: pyspark.sql.DataFrame` not being understood.

This is on `main`:
image

narwhals/_spark_like/namespace.py
```diff
@@ -28,7 +28,7 @@
     from narwhals.utils import Version
 
 
-class SparkLikeNamespace(CompliantNamespace["SparkLikeLazyFrame", "Column"]):
+class SparkLikeNamespace(CompliantNamespace["SparkLikeLazyFrame", "Column"]):  # type: ignore[type-var]
```
dangotbanned, line 31: Same as **<REPLACE ME AFTER REVIEW PLS>**

dangotbanned:
I'd really like this PR to be resolving warnings vs just adding ignore comments.
I do understand we don't have stubs to work with, so @EdAbati let me know if this is an unreasonable ask

EdAbati:
Thanks for the comments, I'll move  the `unpivot` fix out so we can concentrate on discussing typing here. I'll reply later

EdAbati:
Addressed some of the comments thanks :)

I will check again later today/tomorrow if there is anything more that can be fixed instead of being ignored.

narwhals/_spark_like/dataframe.py
dangotbanned: > I think this is because `toArrow()` is a [new method in PySpark 4.0](https://spark.apache.org/docs/4.0.0-preview2/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.toArrow.html) (that is not yet released or tested in our CI)

Oh I see

Yeah I've installed `pyspark` now and can see `pyspark.sql.DataFrame` **is understood** but prior to `4.0` we're falling back to `__getattr__`

https://github.com/apache/spark/blob/7feb911c814b3b0e285bd2b146085668e964f014/python/pyspark/sql/dataframe.py#L3418-L3419

So bringing it all back to (https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1967298532)

> I'm guessing the `[operator]` refers to `__call__`?

We're being warned that `Column` is not callable, but that's not the type being called

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2051/commits/1167ba3d4c82e5556b1e68490a0625e9c89bb3d0)

@EdAbati I did a similar thing in (#2007)
I find it makes reviewing again in the future easier - since we have a complete trail for the issue

image

EdAbati: Very neat!

narwhals/_spark_like/utils.py
```diff
@@ -55,14 +57,14 @@ def native_to_narwhals_dtype(
     if isinstance(dtype, spark_types.DecimalType):
         return dtypes.Decimal()
     if isinstance(dtype, spark_types.ArrayType):
+        dtype = cast(pyspark_types.ArrayType, dtype)
```
dangotbanned, line 60: I might be able to get this typing correctly by letting `pyright` do some inference on `pyspark_types`

The issue at the moment is that `ModuleType` is opaque wrt the members of the module

dangotbanned: There's this functionality that isn't captured in the type system yet - but you can see on hover sometimes.
An import gets the *"type"* `Module("dotted.name")` and it can resolve what is available in the public api

FBruzzesi:
> Regarding "do we want this merged before or after #2038?"
> 
> I guess it is up to @FBruzzesi. Maybe we want to enable type checking first in case we'd catch a sneaky bug? I think should be fine either way

I might have missed this thread and merged already 🥲

narwhals/_spark_like/utils.py
dangotbanned: I'd been thinking about trying to utilize it here

image

But this case is simpler in comparison - since I think there's just 2 modules right?

dangotbanned: Here's a similar case in `altair` where leaving the return type **un-annotated** gives you access to this special `Overload[...]` *"type"*

https://github.com/vega/altair/blob/a196172329fb6d03529bf21a7f58c8bb0f698dd4/altair/utils/core.py#L727-L766

dangotbanned: We could simplify the `TYPE_CHECKING` by just always returning `pyspark.sql.types`

Which would be aligned with

https://github.com/narwhals-dev/narwhals/blob/a43aca045834ff402994e7d9633bd5630bdeb129/narwhals/_spark_like/utils.py#L14

https://github.com/narwhals-dev/narwhals/blob/a43aca045834ff402994e7d9633bd5630bdeb129/narwhals/_spark_like/utils.py#L25

narwhals/_spark_like/utils.py
```diff
@@ -22,45 +22,45 @@
 
 @lru_cache(maxsize=16)
 def native_to_narwhals_dtype(
-    dtype: pyspark_types.DataType,
-    version: Version,
-    spark_types: ModuleType,
+    dtype: pyspark_types.DataType, version: Version, spark_types: ModuleType
 ) -> DType:  # pragma: no cover
```
dangotbanned, line 26: I feel like this is a valid warning from `pyright`
https://github.com/narwhals-dev/narwhals/actions/runs/13525351228/job/37794375285?pr=2051

```py
narwhals/_spark_like/dataframe.py:293:29 - error: Argument of type "Module("pyspark.sql.types")" cannot be assigned to parameter "spark_types" of type "Hashable" in function "__call__"
    "__hash__" is not present (reportArgumentType)
1 error, 0 warnings, 0 informations 
```

Modules aren't hashable - am I missing something?

narwhals/_spark_like/utils.py
dangotbanned: @FBruzzesi 
I imagine you spent a while with this sub-package in (#2038)

Would be interested to hear what you think of the DX (with `pyright` or `pylance`) from checking out this PR 😅

You might notice there's a lot more IDE feedback:

image

image

image

dangotbanned: > @MarcoGorelli I forgot what was the reason for not returning SQLFrame modules from base - wouldn't SQLFrame itself take care of dispatching to the engine? That would avoid having to add a dedicated block for type checking

@FBruzzesi that's a really good point - looking at `sqlframe` it does seem to handle this for us 

https://github.com/eakmanrq/sqlframe/blob/b29873ecb517bf1eb6212778988dadff64317e4f/sqlframe/__init__.py

FBruzzesi: > Would be interested to hear what you think of the DX (with pyright or pylance) from checking out this PR 😅

Yeah that looks much better to what I dealt with 🤩

> @FBruzzesi that's a really good point - looking at sqlframe it does seem to handle this for us

There could be no reason, we might just have missed it 🙈

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/13527240491/job/37800780977?pr=2051

This one is gonna take a little extra work to fix

image

dangotbanned (approved):
Thanks for working on this PR @EdAbati!

Had a lot of fun spiraling through the issues that came up here.
I'll leave this open to let someone else merge - since I've made a lot of changes that I'd also want reviewed :smile:

narwhals/_spark_like/dataframe.py
```diff
@@ -26,19 +27,29 @@
...
     from typing_extensions import Self
+    from typing_extensions import TypeAlias
 
     from narwhals._spark_like.expr import SparkLikeExpr
     from narwhals._spark_like.group_by import SparkLikeLazyGroupBy
     from narwhals._spark_like.namespace import SparkLikeNamespace
     from narwhals.dtypes import DType
     from narwhals.utils import Version
 
+    SQLFrameDataFrame: TypeAlias = _SQLFrameDataFrame[Any, Any, Any, Any, Any]
```
FBruzzesi, line 42: Damn that's a hell of a generic 😂

narwhals/_spark_like/dataframe.py
```diff
@@ -67,7 +82,12 @@ def _F(self: Self) -> Any:  # noqa: N802
         return functions
 
     @property
-    def _native_dtypes(self: Self) -> Any:
+    def _native_dtypes(self: Self):  # type: ignore[no-untyped-def] # noqa: ANN202
+        if TYPE_CHECKING:
+            from pyspark.sql import types
+
+            return types
```
FBruzzesi, line 89: Is there any drawback to this? I have barely seen these conditional in non-import blocks

FBruzzesi (commented):
I must admit I still get easily lost in the typing shenanigans - can't spot anything that does not sound reasonable, but it's not a review that you should trust either 🙈

narwhals/_spark_like/dataframe.py
dangotbanned: > Is there any drawback to this? I have barely seen these conditional in non-import blocks

@FBruzzesi 

Oh let me introduce you to how `pytest` works then 😄 

https://github.com/pytest-dev/pytest/blob/9cf2cae94355cd83ad7e8d88f976c5a524c98cfb/src/_pytest/mark/structures.py#L432-L485
https://github.com/pytest-dev/pytest/blob/9cf2cae94355cd83ad7e8d88f976c5a524c98cfb/src/_pytest/mark/structures.py#L505-L512
https://github.com/pytest-dev/pytest/blob/9cf2cae94355cd83ad7e8d88f976c5a524c98cfb/src/_pytest/_py/path.py#L203-L217

dangotbanned: `polars` does some pretty complex stuff here - which includes tricking the type checker to recognise these symbols & getting lazy imports from them

https://github.com/pola-rs/polars/blob/45ec22a18c0cbdc67f56b3230f8b76149679ef02/py-polars/polars/dependencies.py

https://github.com/pola-rs/polars/blob/45ec22a18c0cbdc67f56b3230f8b76149679ef02/py-polars/polars/dataframe/frame.py#L88-L90

FBruzzesi: 🤯🤯

narwhals/_spark_like/dataframe.py
dangotbanned: > Damn that's a hell of a generic 😂

Ikr!
All the `Unknown`(s) in this gave me quite the scare 
https://github.com/narwhals-dev/narwhals/actions/runs/13527240491/job/37800780977?pr=2051

narwhals/_spark_like/dataframe.py
dangotbanned: > Is there any drawback to this? I have barely seen these conditional in non-import blocks

I'm not sure I'd say *drawback* - but I would recommend only trying this as a last resort.
Here - I'm using it to describe something the type system doesn't support yet.

Another use was https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1965921386 - for a backwards compat bug fix.

I'd much rather do things in a simple way whenever possible

dangotbanned: I did a similar thing here to convert these functions into typeguards

https://github.com/zen-xu/pyarrow-stubs/blob/c482d8f146a8cf2cff55cc4060148fe4883f9309/pyarrow-stubs/types.pyi

https://github.com/narwhals-dev/narwhals/blob/81353a350d8849aa56aa3cf84fb8a5fbd28737e3/narwhals/_arrow/utils.py#L18-L62

dangotbanned:
> from a quick-ish scan this looks good, but i didn't check the details, though if you're all happy with it feel free to ship it 🚢
> 
>  [ 68747470733a2f2f692e666c756666792e63632f34357a51333477304b775758736768304b4d346735715a34766a526a307877322e676966 ](https://camo.githubusercontent.com/af26b5507c769c18e41ff0cb5c1affbdf7c3d119d71e93e6fd8e893c57960bf9/68747470733a2f2f692e666c756666792e63632f34357a51333477304b775758736768304b4d346735715a34766a526a307877322e676966) [ ](https://camo.githubusercontent.com/af26b5507c769c18e41ff0cb5c1affbdf7c3d119d71e93e6fd8e893c57960bf9/68747470733a2f2f692e666c756666792e63632f34357a51333477304b775758736768304b4d346735715a34766a526a307877322e676966)

Ah perfect thanks @MarcoGorelli yeah I'm happy with it if you are ♥️

PR #2059, opened 2025-02-20
chore: remove more `kwargs` from dask, dont treat `ddof` as expressifiable arg

MarcoGorelli:
I don't like how currently we're muddying the waters between "expressifiable arguments" and "kwargs", which leads us to having to pass `ddof` as an expressifiable argument in `std` and `var`

I'd like to start by separating the concepts, and not passing `ddof` as an expressifiable argument

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -48,6 +48,8 @@ class DaskExpr(CompliantExpr["dx.Series"]):
         alias_output_names: Callable[[Sequence[str]], Sequence[str]] | None,
         backend_version: tuple[int, ...],
         version: Version,
+        # Kwargs with metadata which we may need in group-by agg
+        # (e.g. `ddof` for `std` and `var`).
         kwargs: dict[str, Any] | None = None,
     ) -> None:
         self._call = call
@@ -139,8 +141,6 @@ class DaskExpr(CompliantExpr["dx.Series"]):
         # First argument to `call` should be `dx.Series`
         call: Callable[..., dx.Series],
         expr_name: str,
-        # Kwargs with metadata which we may need in group-by agg
-        # (e.g. `ddof` for `std` and `var`).
         kwargs: dict[str, Any] | None = None,
         **expressifiable_args: Self | Any,
     ) -> Self:
@@ -152,9 +152,7 @@ class DaskExpr(CompliantExpr["dx.Series"]):
                 for key, value in expressifiable_args.items()
             }
             for native_series in native_series_list:
-                result_native = call(
-                    native_series, **(kwargs or {}), **other_native_series
-                )
+                result_native = call(native_series, **other_native_series)
                 native_results.append(result_native)
             return native_results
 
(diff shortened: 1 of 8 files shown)


narwhals/_dask/expr.py
```diff
@@ -133,14 +134,14 @@ def func(df: DaskLazyFrame) -> list[dx.Series]:
...
             version=version,
-            kwargs={},
         )
 
     def _from_call(
         self: Self,
         # First argument to `call` should be `dx.Series`
         call: Callable[..., dx.Series],
         expr_name: str,
+        kwargs: dict[str, Any] | None = None,
```
EdAbati, line 144: What if we call this `call_kwargs` i.e. `kwargs` for the `call` function ?

dangotbanned: @EdAbati Couldn't you just pass `call` as a `partial` in the case where this is needed?

https://docs.python.org/3/library/functools.html#partial-objects

- Avoids needing this parameter entirely
- "binding" of arguments is done once, instead of on each use
- Also supports positional arguments

So instead of `call`, `call_kwargs` - you'd have `call.func`, `call.keywords`.
But the keywords themselves aren't exposed here

narwhals/_dask/expr.py
```diff
@@ -310,12 +311,16 @@ def max(self: Self) -> Self:
 
     def std(self: Self, ddof: int) -> Self:
         return self._from_call(
-            lambda _input, ddof: _input.std(ddof=ddof).to_series(), "std", ddof=ddof
+            lambda _input: _input.std(ddof=ddof).to_series(),
+            "std",
+            kwargs={"ddof": ddof},
         )
```
dangotbanned, line 317: Related to https://github.com/narwhals-dev/narwhals/pull/2059#discussion_r1964966715

I'm not following why `kwargs` is used at all here:

https://github.com/narwhals-dev/narwhals/blob/4dda548fc36ebcef3bb70e8d4581de1774879060/narwhals/_dask/group_by.py#L54-L55

https://github.com/narwhals-dev/narwhals/blob/4dda548fc36ebcef3bb70e8d4581de1774879060/narwhals/_dask/group_by.py#L148-L150

Am I reading this wrong, or are the same `kwargs` applied twice?

narwhals/_dask/expr.py
dangotbanned: > if you don't know the class that would be passed.

Is that why you're storing `function_name`, in addition to the `lambda`?

MarcoGorelli:
thanks both for comments!

i'd like to refactor this mechanism more broadly, so i'll hold off `functools` gymnastics for the time being (even though it does look really nice!), but definitely interested in revisiting in the future

narwhals/_dask/expr.py
MarcoGorelli: yup, that's right

dangotbanned (approved):
Sounds good @MarcoGorelli

MarcoGorelli:
thanks all!

PR #2071, opened 2025-02-23
chore: use more expression metadata functions

MarcoGorelli:
temporary-ish refactor while I build up to something else, but this should simplify the ExprMetadata class changes anyway? these could just become methods

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2067, opened 2025-02-22
chore: use `change_metadata_kind`

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2074, opened 2025-02-23
refactor: Use real class for `ExprMetadata` 

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2067
- #2071

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -388,9 +388,9 @@ class ExprMetadata:
         """Change kind and set `order_dependent` to True."""
         return ExprMetadata(kind, order_dependent=True)
 
-
-def default_selector_metadata() -> ExprMetadata:
-    return ExprMetadata(ExprKind.TRANSFORM, order_dependent=False)
+    @staticmethod
+    def selector() -> ExprMetadata:
+        return ExprMetadata(ExprKind.TRANSFORM, order_dependent=False)
 
 
 def combine_metadata(*args: IntoExpr | object | None, str_as_lit: bool) -> ExprMetadata:
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -325,10 +325,6 @@ def evaluate_output_names_and_aliases(
     return output_names, aliases
 
 
-def operation_is_order_dependent(*args: IntoExpr | Any) -> bool:
-    return any(is_expr(x) and x._metadata.is_order_dependent() for x in args)
-
-
 class ExprKind(Enum):
     """Describe which kind of expression we are dealing with.
 
@@ -360,7 +356,7 @@ class ExprMetadata:
         self._kind: ExprKind = kind
         self._order_dependent: bool = order_dependent
 
-    def __init_subclass__(cls, /, *args: Any, **kwds: Any) -> Never:
+    def __init_subclass__(cls, /, *args: Any, **kwds: Any) -> Never:  # pragma: no cover
         msg = f"Cannot subclass {cls.__name__!r}"
         raise TypeError(msg)
 
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -10,7 +10,6 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 from typing import Sequence
-from typing import TypedDict
 from typing import TypeVar
 from typing import overload
 
@@ -327,7 +326,7 @@ def evaluate_output_names_and_aliases(
 
 
 def operation_is_order_dependent(*args: IntoExpr | Any) -> bool:
-    return any(is_expr(x) and x._metadata["is_order_dependent"] for x in args)
+    return any(is_expr(x) and x._metadata.is_order_dependent() for x in args)
 
 
 class ExprKind(Enum):
@@ -354,13 +353,7 @@ class ExprKind(Enum):
     """e.g. `nw.col('a').drop_nulls()`"""
 
 
-class ExprMetadata(TypedDict):
-    kind: ExprKind
-    is_order_dependent: bool
-
-
-# TODO @dangotbanned: replace `ExprMetadata` and rename
-class ExprMeta:
+class ExprMetadata:
     __slots__ = ("_kind", "_order_dependent")
 
     def __init__(self, kind: ExprKind, /, *, order_dependent: bool) -> None:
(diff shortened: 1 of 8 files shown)


dangotbanned:
@MarcoGorelli didn't realise #2072 was coming oops

MarcoGorelli:
ah sorry - I can fixup the merge conflicts from here if you like, thanks for doing this!

dangotbanned:
> ah sorry - I can fixup the merge conflicts from here if you like, thanks for doing this!

Perfect thanks @MarcoGorelli happy for you to takeover 🙌

narwhals/_expression_parsing.py
```diff
@@ -353,27 +353,40 @@ class ExprKind(Enum):
...
+        return self.kind in {ExprKind.AGGREGATION, ExprKind.LITERAL}
+
+    def is_changes_length(self) -> bool:
+        return self.kind is ExprKind.CHANGES_LENGTH
+
+    def with_kind(self, kind: ExprKind, /) -> ExprMetadata:
+        """Change metadata kind, leaving all other attributes the same."""
+        return ExprMetadata(kind, order_dependent=self.is_order_dependent())
+
+    def with_dependence(self, order_dependent: bool = True, /) -> ExprMetadata:  # noqa: FBT001, FBT002
```
dangotbanned, line 387: Disgreeing with `ruff` on this one - making it keyword-only doesn't make sense here

narwhals/stable/v1/__init__.py
```diff
@@ -1050,7 +1048,7 @@ def arg_true(self: Self) -> Self:
         """
         return self.__class__(
             lambda plx: self._to_compliant_expr(plx).arg_true(),
-            change_kind_and_make_order_dependent(self._metadata, ExprKind.CHANGES_LENGTH),
+            ExprMetadata(ExprKind.CHANGES_LENGTH, order_dependent=True),
```
dangotbanned, line 1051: > Another one could be having methods on `ExprKind` to allow `ExprKind -> ExprMetadata` - while keeping things [Fluent](https://en.wikipedia.org/wiki/Fluent_interface)

https://github.com/narwhals-dev/narwhals/pull/2074#issuecomment-2676923483

So this would be something like:

```py
# ExprMetadata(ExprKind.CHANGES_LENGTH, order_dependent=True)
ExprKind.CHANGES_LENGTH.with_dependence()`
```

dangotbanned:
Also for a similar interface in the stdlib:
https://github.com/python/cpython/blob/ac455678e7c9f12e9cdbdf5d31667cd12974ca92/Lib/pathlib.py#L292-L793

narwhals/_expression_parsing.py
MarcoGorelli: there's no operation that can set `is_order_dependent` to false anyway, so i think we can remove the argument

dangotbanned: But there are operations that pass it the result of some other function - which *could* be, right?

dangotbanned: IIRC I used that when it was evaluated before a `lambda` and then passed in

MarcoGorelli (approved):
thanks @dangotbanned ! looks good to me on green

narwhals/_expression_parsing.py
```diff
@@ -353,31 +353,48 @@ class ExprKind(Enum):
...
+        """Set `order_dependent` to True, leaving all other attributes the same."""
+        return ExprMetadata(self.kind, order_dependent=True)
+
+    def with_kind_and_order_dependence(self, kind: ExprKind, /) -> ExprMetadata:
+        """Change kind and set `order_dependent` to True."""
+        return ExprMetadata(kind, order_dependent=True)
+
+
+def default_selector_metadata() -> ExprMetadata:
+    return ExprMetadata(ExprKind.TRANSFORM, order_dependent=False)
```
dangotbanned:
```suggestion
    @staticmethod
    def selector():
        return ExprMetadata(ExprKind.TRANSFORM, order_dependent=False)
```

dangotbanned (commented):
One little tweak, since the function already has `metadata` in the name

narwhals/_expression_parsing.py
dangotbanned: Thanks @MarcoGorelli

narwhals/_expression_parsing.py
```diff
@@ -353,31 +349,48 @@ class ExprKind(Enum):
...
+class ExprMetadata:
+    __slots__ = ("_kind", "_order_dependent")
+
+    def __init__(self, kind: ExprKind, /, *, order_dependent: bool) -> None:
+        self._kind: ExprKind = kind
+        self._order_dependent: bool = order_dependent
+
+    def __init_subclass__(cls, /, *args: Any, **kwds: Any) -> Never:  # pragma: no cover
+        msg = f"Cannot subclass {cls.__name__!r}"
+        raise TypeError(msg)
```
FBruzzesi, line 361: TIL 👀

narwhals/_expression_parsing.py
```diff
@@ -353,31 +349,48 @@ class ExprKind(Enum):
...
+        return self.kind is ExprKind.TRANSFORM
 
+    def is_aggregation_or_literal(self) -> bool:
+        return self.kind in {ExprKind.AGGREGATION, ExprKind.LITERAL}
 
-def change_kind(md: ExprMetadata, kind: ExprKind) -> ExprMetadata:
-    # Change metadata kind, leaving all other attributes the same.
-    return ExprMetadata(kind=kind, is_order_dependent=md["is_order_dependent"])
+    def is_changes_length(self) -> bool:
+        return self.kind is ExprKind.CHANGES_LENGTH
```
FBruzzesi, line 377: Considering all other kinds to be names and not verbs, should we change this in `LENGTH_CHANGING`?
I was itching at reading `is_changes_length` 😂

narwhals/_expression_parsing.py
```diff
@@ -353,31 +349,48 @@ class ExprKind(Enum):
...
 
+    def with_kind(self, kind: ExprKind, /) -> ExprMetadata:
+        """Change metadata kind, leaving all other attributes the same."""
+        return ExprMetadata(kind, order_dependent=self.is_order_dependent())
```
FBruzzesi, line 381: Nitpick here and for the following two methods: I am not a fan of `with_*`, I would like to put up for discussion to use `set_*` 😇

narwhals/_expression_parsing.py
```diff
@@ -353,31 +349,48 @@ class ExprKind(Enum):
...
+    def is_order_dependent(self) -> bool:
+        return self._order_dependent
 
-def default_metadata() -> ExprMetadata:
-    return ExprMetadata(kind=ExprKind.TRANSFORM, is_order_dependent=False)
+    def is_transform(self) -> bool:
+        return self.kind is ExprKind.TRANSFORM
 
+    def is_aggregation_or_literal(self) -> bool:
+        return self.kind in {ExprKind.AGGREGATION, ExprKind.LITERAL}
```
FBruzzesi, line 374: Any advantage in these being properties? 🤔

FBruzzesi (commented):
I am trying to catch up with a lot of stuff, and I stopped for some nitpicks! Feel free to ignore

narwhals/_expression_parsing.py
dangotbanned: > Nitpick here and for the following two methods: I am not a fan of `with_*`, I would like to put up for discussion to use `set_*` 😇

https://github.com/python/cpython/blob/ac455678e7c9f12e9cdbdf5d31667cd12974ca92/Lib/pathlib.py#L292-L793

@FBruzzesi `set_` implies mutability, `with_` implies new object *with* changes

narwhals/_expression_parsing.py
dangotbanned: @FBruzzesi could do, but `is_*` is also a pretty common function/method name pattern

`pathlib` and `polars` do this for example

narwhals/_expression_parsing.py
dangotbanned: Also `set_` really reminds me of java 🤮

narwhals/_expression_parsing.py
MarcoGorelli: been thinking about this - how about `filtration`?

narwhals/_expression_parsing.py
FBruzzesi: > @FBruzzesi `set_` implies mutability, `with_` implies new object _with_ changes

Oh wow, I have probably seen `set_*` misused in a bunch of (very much used) projects then 🙈

narwhals/_expression_parsing.py
dangotbanned: > TIL 👀

@FBruzzesi this is another use case for it btw https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-12054784

narwhals/_expression_parsing.py
FBruzzesi: True, thanks for coming up with examples (that I use daily)! It makes it so much easier to make up my mind :)

narwhals/_expression_parsing.py
FBruzzesi: I tried to think of an alternative, but I don't have it, so it would probably do the job 😁

dangotbanned: @MarcoGorelli @FBruzzesi 

maybe have a flip through this to see if they have a term in their IR https://github.com/ibis-project/ibis/tree/913fa51d16e6165411f5a7f272ad5d9a42b51962/ibis/expr/operations

PR #2078, opened 2025-02-23
chore: add `has_open_windows` to `ExprMetadata` instead of `is_order_dependent`

MarcoGorelli:
One step closer to supporting `lf.select(nw.col('a').cum_sum('b', order_by='c'))`. Getting there!

This PR does not (yet) have any user-facing impact. But...we're not that far off from being able to make changes which are!

The gist of the change is to distinguish between:
- transform: preserves length
- window: preserves length and the last node is a window function (e.g. `cum_sum`, `rolling_mean`, `shift`, ...)

Then:
- and open window is a window function not immediately followed by `over` with `order_by` specified
- calling `over` immediately after a window function reduces `n_open_windows` by `1`
- by the time we evaluate an expression, then:
  - in the lazy API, there must be 0 open windows
  - in the eager API, there are no restrictions, as physical row order is always defined

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


MarcoGorelli:
ooh nice, looks like the tubular test suite helped find a bug 🙌

narwhals/stable/v1/__init__.py
```diff
@@ -1047,7 +1047,7 @@ def arg_true(self: Self) -> Self:
         """
         return self.__class__(
             lambda plx: self._to_compliant_expr(plx).arg_true(),
-            self._metadata.with_kind_and_order_dependence(ExprKind.CHANGES_LENGTH),
+            self._metadata.with_extra_open_window(),
```
MarcoGorelli, line 1050: this should probably still be changes_length

narwhals/_expression_parsing.py
```diff
@@ -489,7 +505,7 @@ def apply_n_ary_operation(
     )
     kinds = [infer_kind(comparand, str_as_lit=str_as_lit) for comparand in comparands]
 
-    broadcast = any(kind is ExprKind.TRANSFORM for kind in kinds)
+    broadcast = any(kind in {ExprKind.TRANSFORM, ExprKind.WINDOW} for kind in kinds)
```
MarcoGorelli, line 508: need to make a `.preserves_length` function, this is way too easy to get wrong

narwhals/_expression_parsing.py
```diff
@@ -343,18 +343,30 @@ class ExprKind(Enum):
...
+    - `nw.col('a').cum_sum()`
+    - `(nw.col('a')+1).cum_sum()`
+
+    non-examples:
+    - `nw.col('a').cum_sum()+1`
+    - `nw.col('a').cum_sum().mean()`
+    """
 
     CHANGES_LENGTH = auto()
     """e.g. `nw.col('a').drop_nulls()`"""
```
dangotbanned, line 361: Following up (https://github.com/narwhals-dev/narwhals/pull/2074#discussion_r1966889317)


What about `ExprKind.REDUCTION`?

It seems like you're using `CHANGES_LENGTH` for the same kinds of ops as https://github.com/ibis-project/ibis/blob/913fa51d16e6165411f5a7f272ad5d9a42b51962/ibis/expr/operations/reductions.py

The only outlier I see is `Expr.map_batches` - everything else is a reduction in length (changes in one direction):

https://github.com/narwhals-dev/narwhals/blob/8609786a1cdc29055cc422255736c8688cec0a5a/narwhals/expr.py#L647-L653

narwhals/expr.py
```diff
@@ -1535,12 +1538,17 @@ def over(self: Self, *keys: str | Iterable[str]) -> Self:
         if self._metadata.is_changes_length():
             msg = "`.over()` can not be used for expressions which change length."
             raise LengthChangingExprError(msg)
+        kind = ExprKind.TRANSFORM
+        n_open_windows = self._metadata.n_open_windows
+        if _order_by is not None and self._metadata.is_window():
+            n_open_windows -= 1
+        metadata = ExprMetadata(kind, n_open_windows=n_open_windows)
```
dangotbanned, line 1545: Is this the only context that can *decrement* `n_open_windows`?
Every other usage I've seen seems to *increment*.

What if you had:

```py
class ExprMetadata:
    n_windows: int

    def open_window(self):
        return ExprMetadata(self.kind, n_windows=self._n_windows + 1)
    def close_window(self):
        return ExprMetadata(self.kind, n_windows=self._n_windows - 1)
```

MarcoGorelli: yes this is the only one

narwhals/_expression_parsing.py
dangotbanned: > i think reduction risks sounding like an aggregation [mongodb.com/docs/manual/reference/operator/aggregation/reduce](https://www.mongodb.com/docs/manual/reference/operator/aggregation/reduce/)

@MarcoGorelli it is though isn't it?
Following the logic of `MongoDB` - a reduction is a kind of aggregation


I hadn't thought to look here before - but there also seems to be an overlap with `polars-rs`
https://github.com/pola-rs/polars/tree/28cc37ad607eff376ef462cc9ce98f26c216e84f/crates/polars-expr/src/reduce

And IIUTC - they're what is shown in https://docs.pola.rs/api/python/stable/reference/expressions/aggregation.html

MarcoGorelli: yes but `changes_length` isn't an aggregation, it's things like `Expr.drop_nulls` which change the length but don't aggregate

dangotbanned: > yes but `changes_length` isn't an aggregation, it's things like `Expr.drop_nulls` which change the length but don't aggregate

Oops my bad - I misread all of these 🤦‍♂️ 
- `arg_true` -> `arg_max`
- `n_unique` -> `unique`
- `head` -> `first`
- `tail` -> `last`

Absolutely wild 😄

MarcoGorelli: yeah probably

MarcoGorelli:
thanks for reviews! shipping this then so we can work on further developments

PR #2064, opened 2025-02-21
refactor: Generic `CompliantSelector`

dangotbanned:
# Tracking
- [ ] #2084 
- [x] #2051
  - After that merges, move these into (#2115)
  	- https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1970545635
  	- https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1970546493
  	- https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1970546759
  	- https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1970547095
  - That will support moving these fixes out as well
  	- https://github.com/narwhals-dev/narwhals/pull/2064#issuecomment-2676872619
- [x] #2101

## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2060
- #2058
- #2057
- #2044
- #1942

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Expecting quite a nice reduction in copy/pasted code

- [x] `_arrow`
- [x] `_dask`
- [x] `_duckdb`
- [x] `_pandas_like`
- [x] `_spark_like`
- [x] ~~Investigate~~ fix `3.8` [protocol bug](https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1966282177)

PR #2087, opened 2025-02-24
chore(typing): Resolve `_dask` errors

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/2064#issuecomment-2676872619

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Just following up some things I temporarily added to #2064.

Sadly this is about all that I can do, given the typing coverage of (https://github.com/dask/dask)

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -35,7 +35,7 @@ if TYPE_CHECKING:
     from narwhals.utils import Version
 
 
-class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):
+class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):  # pyright: ignore[reportInvalidTypeArguments] (#2044)
     _implementation: Implementation = Implementation.DASK
 
     def __init__(
@@ -454,7 +454,7 @@ class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):
                 _input.dtype, self._version, self._implementation
             )
             if dtype.is_numeric():
-                return _input != _input  # noqa: PLR0124
+                return _input != _input  # pyright: ignore[reportReturnType] # noqa: PLR0124
             msg = f"`.is_nan` only supported for numeric dtypes and not {dtype}, did you mean `.is_null`?"
             raise InvalidOperationError(msg)
 
@@ -492,7 +492,7 @@ class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):
                 col_token,
                 backend_version=self._backend_version,
                 implementation=self._implementation,
-            )
+            )  # pyright: ignore[reportAssignmentType]
             first_distinct_index = _input.groupby(_name).agg({col_token: "min"})[
                 col_token
             ]
@@ -509,7 +509,7 @@ class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):
                 col_token,
                 backend_version=self._backend_version,
                 implementation=self._implementation,
-            )
+            )  # pyright: ignore[reportAssignmentType]
             last_distinct_index = _input.groupby(_name).agg({col_token: "max"})[col_token]
             return _input[col_token].isin(last_distinct_index)
 
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -487,16 +487,11 @@ class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):  # pyright: ignore[
         def func(_input: dx.Series) -> dx.Series:
             _name = _input.name
             col_token = generate_temporary_column_name(n_bytes=8, columns=[_name])
-            _input = add_row_index(
-                _input.to_frame(),
-                col_token,
-                backend_version=self._backend_version,
-                implementation=self._implementation,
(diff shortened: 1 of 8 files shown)


narwhals/_dask/namespace.py
```diff
@@ -347,32 +345,31 @@ def __init__(
...
             return [then_series.where(condition)]
 
-        if is_compliant_expr(self._otherwise_value):
-            otherwise_value: dx.Series | object = self._otherwise_value(df)[0]
+        if isinstance(self._otherwise_value, DaskExpr):
+            otherwise_value = self._otherwise_value(df)[0]
         else:
             otherwise_value = self._otherwise_value
-        (otherwise_series,) = align_series_full_broadcast(df, otherwise_value)
+        otherwise_series = align_series_full_broadcast(df, otherwise_value)[0]
```
dangotbanned, line 370: FYI, I'll follow this up giving the same treatment for `_pandas` & `_arrow`.

Found they all had the same problems in (https://github.com/narwhals-dev/narwhals/pull/2064/commits/ef85eda6599d0a0fce261c35416983756fa456e5)

MarcoGorelli: this was intentional, the way i've written it would error if there was more than one value, is there a way to keep it?

dangotbanned: > this was intentional, the way i've written it would error if there was more than one value, is there a way to keep it?

@MarcoGorelli are you refering to the unpacking part? 

I'll admit that wasn't needed to resolve any issues. I've just never seen that syntax used before and this seemed a bit clearer to me?

Would you be open to reworking `align_series_full_broadcast` into something that (IMO) better documents what is going on here?

I can take a look again tomorrow, but IIRC the function was fairly simple internally - so adding a param to ensure `n_len` and then raising could be something?

dangotbanned: Happy for you to revert that bit if it's a deal-breaker

narwhals/_dask/expr.py
```diff
@@ -487,31 +487,23 @@ def is_first_distinct(self: Self) -> Self:
...
-                implementation=self._implementation,
+            frame = add_row_index(
+                _input.to_frame(), col_token, self._backend_version, self._implementation
             )
-            first_distinct_index = _input.groupby(_name).agg({col_token: "min"})[
-                col_token
-            ]
-            return _input[col_token].isin(first_distinct_index)
+            first_distinct_index = frame.groupby(_name).agg({col_token: "min"})[col_token]
+            return frame[col_token].isin(first_distinct_index)
```
FBruzzesi, line 494: That _one_ character change! Amazing 👌

FBruzzesi (commented):
Thanks @dangotbanned , just a tiny comment regarding other binary operations :)

narwhals/_dask/expr.py
```diff
@@ -255,7 +255,7 @@ def __ne__(self: Self, other: DaskExpr) -> Self:  # type: ignore[override]
             lambda _input, other: _input.__ne__(other), "__ne__", other=other
         )
 
-    def __ge__(self: Self, other: DaskExpr) -> Self:
+    def __ge__(self: Self, other: DaskExpr | Any) -> Self:
```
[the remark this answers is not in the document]
dangotbanned: @FBruzzesi I can definitely see the argument for that

You might notice that in (https://github.com/narwhals-dev/narwhals/pull/2087/commits/cdf5e423d6fa34c75fce048511b0d57c59bb39e3) I only updated signatures that were already used - but triggered a warning.

My thought process was:

- I've not used (https://github.com/dask/dask) before
- (https://github.com/dask/dask) doesn't have any annotations
- `Any` isn't a very useful hint
  - Using anything *narrower* will probably conflict with the lack of upstream typing

So all of that is to say - I'm not against it - but I'm not confident I could make the right decisions on more broadly updating them.

Does that all make sense?
I know it can seem strange typing things partially like this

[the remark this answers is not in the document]
dangotbanned: When a package does stuff like this https://github.com/dask/dask/blob/cf48d546955faddc3eb4d492e70a05b900068402/dask/dataframe/dask_expr/_collection.py#L2635-L2675
It truly makes everyone's life harder 😞

narwhals/_dask/namespace.py
MarcoGorelli: yeah I'd prefer to keep the unpacking please

it's not about the length of the series but about the number of series returned from the method

dangotbanned: Got it, should be good to go in (https://github.com/narwhals-dev/narwhals/pull/2087/commits/f799337b6952fec3d2a0c8086271804cf7cde1a9) @MarcoGorelli

narwhals/_dask/expr.py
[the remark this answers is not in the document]
FBruzzesi: Oh wow! Yes I fully agree, especially with:

> Any isn't a very useful hint

I pointed that out because only a couple of binary ops were addressed here.

MarcoGorelli (approved):
amazing, thanks both!

Issue #2091, opened 2025-02-25
perf: only check for duplicate column names in DuckDB on collection


MarcoGorelli:
We currently check for duplicate column names in some places for DuckDB for operations which may introduce duplicates

However, we probably don't need to do that, right? We can just do the check in `collect`. If someone introduces duplicates but then doesn't select them, we don't actually need to error

FBruzzesi:
Commenting here for visibility. Anyone interested in tackling this, the same approach can be used in the spark-like case, since I introduced duplicate columns validation (mostly for sqlframe case)

MarcoGorelli:
just tried this out quickly and it seems fine, will submit something shortly

PR #2093, opened 2025-02-25
perf: use cached property for schema/columns lazy frames

MarcoGorelli:
thanks @dangotbanned for highlighting this area of improvement!

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -47,6 +47,7 @@ class DaskLazyFrame(CompliantLazyFrame):
         self._backend_version = backend_version
         self._implementation = Implementation.DASK
         self._version = version
+        self._cached_schema: dict[str, DType] | None = None
         validate_backend_version(self._implementation, self._backend_version)
         if validate_column_names:
             check_column_names_are_unique(native_dataframe.columns)
@@ -138,7 +139,7 @@ class DaskLazyFrame(CompliantLazyFrame):
 
     @property
     def columns(self: Self) -> list[str]:
-        return self._native_frame.columns.tolist()
+        return list(self.schema)
 
     def filter(self: Self, predicate: DaskExpr) -> Self:
         # `[0]` is safe as the predicate's expression only returns a single column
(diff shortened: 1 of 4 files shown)


narwhals/_dask/dataframe.py
```diff
@@ -194,13 +195,15 @@ def drop_nulls(self: Self, subset: list[str] | None) -> Self:
...
-        }
+        if self._cached_schema is None:
+            native_dtypes = self._native_frame.dtypes
+            self._cached_schema = {
+                col: native_to_narwhals_dtype(
+                    native_dtypes[col], self._version, self._implementation
+                )
+                for col in self._native_frame.columns
+            }
+        return self._cached_schema
```
dangotbanned, line 206: Was the issue with `@cached_property` that the returned value was mutable?

If so, something like this might help?:

```py
from types import MappingProxyType

schema: dict[str, DType] = {}
immutable_schema = MappingProxyType(schema)
```

dangotbanned: `3.13` being a problem sounds like GIL/threading-related?

MarcoGorelli: i don't know what the issue was, but here's the traceback https://github.com/narwhals-dev/narwhals/actions/runs/13523765828/job/37789003669

MarcoGorelli: 3.10 on windows too

dangotbanned: Is it related to #2091?

MarcoGorelli: i don't think so

MarcoGorelli: putting a breakpoint in, it seems that `.columns` is sometimes just reporting the wrong value

MarcoGorelli: the current solution is simple enough and understandable, we probably don't need another library's method for it?

dangotbanned: Shouldn't we be caching `_native_frame.schema` and `_native_frame.columns`?

That was the part I was imagining was expensive.

The `narwhals` parts are already cached in `native_to_narwhals_dtype`

MarcoGorelli: isn't this doing the same thing? if `self._collect_schema` is not `None`, it means it was already computed, so we don't trigger another call to `_native_frame.schema`

dangotbanned: > interesting, thanks! maybe we should remove that `lru_cache` for pyspark 🤔

Thought it was worth a shot - I'm gonna be pretty surprised if the timing of the two PRs resolved this 😄

dangotbanned: If not I can mark (https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1970074662) as a false positive

MarcoGorelli: nvm messed up my git history, one sec

MarcoGorelli:
ah same error...ok, gonna remove the `lru_cache` for now, and do the manual `cached_property`

MarcoGorelli:
pyarrow is completely eager so computing schema there should really be practically free

MarcoGorelli:
gonna ship this then, thanks a tonne for your help here! 🙌

PR #2096, opened 2025-02-25
feat: align `over` signature with Polars

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -1494,14 +1494,16 @@ class Expr:
         )
 
     def over(
-        self: Self, *keys: str | Iterable[str], _order_by: str | None = None
+        self: Self,
+        *partition_by: str | Iterable[str],
+        _order_by: str | None = None,
     ) -> Self:
         """Compute expressions over the given groups.
 
         Arguments:
-            keys: Names of columns to compute window expression over.
-                  Must be names of columns, as opposed to expressions -
-                  so, this is a bit less flexible than Polars' `Expr.over`.
+            partition_by: Names of columns to compute window expression over.
+                Must be names of columns, as opposed to expressions -
+                so, this is a bit less flexible than Polars' `Expr.over`.
             _order_by: Unused, but this is building up to something.
 
         Returns:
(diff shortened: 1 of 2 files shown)


narwhals/expr.py
```diff
@@ -1494,14 +1494,16 @@ def sample(
...
         """Compute expressions over the given groups.
 
         Arguments:
-            keys: Names of columns to compute window expression over.
-                  Must be names of columns, as opposed to expressions -
-                  so, this is a bit less flexible than Polars' `Expr.over`.
+            partition_by: Names of columns to compute window expression over.
+                Must be names of columns, as opposed to expressions -
+                so, this is a bit less flexible than Polars' `Expr.over`.
             _order_by: Unused, but this is building up to something.
```
FBruzzesi, line 1507: I just noticed this 😂

narwhals/expr.py
```diff
@@ -1494,14 +1494,16 @@ def sample(
         )
 
     def over(
-        self: Self, *keys: str | Iterable[str], _order_by: str | None = None
+        self: Self,
+        *partition_by: str | Iterable[str],
```
FBruzzesi, line 1498: Any reason to not align with:

```py
partition_by: str | Iterable[str],
*more_exprs: str,
```

at the Narwhals level only?
I think the main difference in being able to pass `partition_by` as keyword argument.

FBruzzesi (approved):
Tiny question but clearly looks good 
HiThereGIF (2)

narwhals/expr.py
MarcoGorelli: like this it's a bit more natural not pass any `partition_by`, whereas the Polars way it's required

FBruzzesi: So are we planning to default it to None eventually to allow `.max().over()`?

MarcoGorelli: the signature from this PR allows that, if you don't pass any positional arguments then it's an empty over

FBruzzesi: 🤦🏼‍♂️

PR #2098, opened 2025-02-25
perf: avoid full broadcast for otherwise_value in when/then/otherwise

MarcoGorelli:
A full broadcast is inevitable for the `then` value, but for the `otherwise` value we can avoid it

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/utils.py
+++ b/narwhals/_spark_like/utils.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from narwhals.utils import Version
 
 
+# note: don't lru_cache this as `ModuleType` isn't hashable
 def native_to_narwhals_dtype(
     dtype: pyspark_types.DataType,
     version: Version,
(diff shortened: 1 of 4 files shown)


narwhals/_spark_like/utils.py
```diff
@@ -19,6 +19,7 @@
     from narwhals.utils import Version
 
 
+# note: don't lru_cache this as `ModuleType` isn't hashable
```
dangotbanned, line 22: Not sure why, but VSCode gives some nice highlights with all caps

```suggestion
# NOTE: don't lru_cache this as `ModuleType` isn't hashable
```

PR #2058, opened 2025-02-20
chore: remove `kwargs` from `selector` function

MarcoGorelli:
follow-up to #2057 - there's a lot of `kwargs` we're passing around that we don't need

i'd generally like to refactor that whole mechanism (we'll need it for when we want to support `rolling_mean(n).over(group, order_by=id)`), but for a start we can remove them when they're not needed

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2102, opened 2025-02-26
perf: Avoid redefining `lambda` in `*Namespace.all`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Mentioned in (https://github.com/narwhals-dev/narwhals/pull/2058#discussion_r1964137667)
- Originally (https://github.com/narwhals-dev/narwhals/pull/2064/commits/028098e4b5091f73fca2f812830d54f109875777) and (https://github.com/narwhals-dev/narwhals/pull/2064/commits/c597ba7c2b06b1d1db29e9b7f254aa3be47b5d86)

## Checklist

- [ ] Tests added
- [x] Documented the changes

Previously, a new function was created in each of these locations - but it didn't bind any arguments - so there was no benefit to using a `lambda`

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -26,6 +26,7 @@ from narwhals._expression_parsing import combine_alias_output_names
 from narwhals._expression_parsing import combine_evaluate_output_names
 from narwhals.typing import CompliantNamespace
 from narwhals.utils import Implementation
+from narwhals.utils import get_column_names
 from narwhals.utils import import_dtypes_module
 
 if TYPE_CHECKING:
@@ -161,7 +162,7 @@ class ArrowNamespace(CompliantNamespace[ArrowDataFrame, ArrowSeries]):
             ],
             depth=0,
             function_name="all",
-            evaluate_output_names=lambda df: df.columns,
+            evaluate_output_names=get_column_names,
             alias_output_names=None,
             backend_version=self._backend_version,
             version=self._version,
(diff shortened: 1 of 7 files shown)


MarcoGorelli:
thanks - could you clarify how this helps please? i don't understand the difference, and there aren't any arguments to bind in this function?

dangotbanned:
> thanks - could you clarify how this helps please? i don't understand the difference,

Sure thing @MarcoGorelli

Right now, every time `all()` is called that line is equivalent to:

```py
def _(df):
    return df.columns

...

columns = _(df)
```

By equivalent - I mean if you call `all()` 3 times - that would expand to this:

```py
def _1(df):
    return df.columns
...
columns = _1(df)
def _2(df):
    return df.columns
...
columns = _2(df)
def _3(df):
    return df.columns
...
columns = _3(df)
```

This change skips defining function each time, so those 3 calls would all be:

```py
columns = get_column_names(df)
columns = get_column_names(df)
columns = get_column_names(df)
```

The difference is we didn't need to create a new inline function each time.

>  and there aren't any arguments to bind in this function?

What I mean is the scope created by `lambda df: df.columns` doesn't *bind* an argument in the definition.
By comparison, this `lambda` binds `self._implementation`, `self._backend_version`, `self._version` from the outer scope.

That would be a use-case I'd understand for a `lambda` - since it utilizes `self`

https://github.com/narwhals-dev/narwhals/blob/8888e2c503e40121bcff607a4e3486675d46e181/narwhals/_pandas_like/namespace.py#L125-L132

Also some links:
- https://docs.python.org/3/reference/expressions.html#lambda
- https://docs.python.org/3/howto/functional.html#small-functions-and-the-lambda-expression
- https://discuss.python.org/t/what-is-the-purpose-of-lambda-expressions/12415/2

MarcoGorelli (approved):
sure, thanks @dangotbanned

PR #2103, opened 2025-02-26
chore: deprecate `LazyFrame.gather_every` in main namespace (but maintain it in stable.v1)

### LINKED ISSUE #2047: deprecate LazyFrame.gather_every (but keep around in stable.v1)
towards https://github.com/narwhals-dev/narwhals/discussions/1657

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2047

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -3158,7 +3158,7 @@ class LazyFrame(BaseFrame[FrameT]):
             "Note: this will remain available in `narwhals.stable.v1`.\n"
             "See https://narwhals-dev.github.io/narwhals/backcompat/ for more information.\n"
         )
-        issue_deprecation_warning(msg, _version="1.28.0")
+        issue_deprecation_warning(msg, _version="1.29.0")
 
         return super().gather_every(n=n, offset=offset)
 
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -34,6 +34,7 @@ from narwhals.utils import find_stacklevel
 from narwhals.utils import flatten
 from narwhals.utils import generate_repr
 from narwhals.utils import is_sequence_but_not_str
+from narwhals.utils import issue_deprecation_warning
 from narwhals.utils import parse_version
 
 if TYPE_CHECKING:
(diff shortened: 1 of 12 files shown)


FBruzzesi:
CI has issue with building duckdb==1.3.0.dev756  🤔

I can see other PRs run in the same blocker

narwhals/dataframe.py
```diff
@@ -3109,13 +3110,25 @@ def lazy(self: Self) -> Self:
...
+
         Arguments:
             n: Gather every *n*-th row.
             offset: Starting index.
 
         Returns:
             The LazyFrame containing only the selected rows.
         """
+        msg = (
+            "`LazyFrane.gather_every` is deprecated and will be removed in a future version.\n\n"
```
MarcoGorelli, line 3126: typo

narwhals/dataframe.py
```diff
@@ -3109,13 +3110,25 @@ def lazy(self: Self) -> Self:
...
+            "Note: this will remain available in `narwhals.stable.v1`.\n"
+            "See https://narwhals-dev.github.io/narwhals/backcompat/ for more information.\n"
+        )
+        issue_deprecation_warning(msg, _version="1.28.0")
```
MarcoGorelli, line 3130: 1.29?

MarcoGorelli (approved):
thanks! couple of typos then feel free to ship it

narwhals/dataframe.py
FBruzzesi: I checked how it was used in the other cases e.g. in `Expr.tail` https://github.com/narwhals-dev/narwhals/blob/77a0150ffb5121086d74b0f9a10ec26781162529/narwhals/expr.py#L1784

and I noticed that it corresponds to the current version (at the time) and in the docs we mention the release right after. I am not sure if this was intentional or not - the description for `_version` made it ambiguous to me:

```py
_version: Narwhals version when the warning was introduced. Just used for internal
            bookkeeping.
```

narwhals/dataframe.py
FBruzzesi:
```suggestion
            "`LazyFrame.gather_every` is deprecated and will be removed in a future version.\n\n"
```

narwhals/dataframe.py
MarcoGorelli: ah sorry 😳 this is just internal anyway so as long as it's approximately right it's probably fine 😄

MarcoGorelli: if we want to be consistent going forwards, i think the intention was probably for it to be the next minor release after the current one

if it ends up in a patch release in between then no big deal

FBruzzesi: Should I change the other instances in which it is used to match what's in the doc?

PR #2106, opened 2025-02-27
fill_null with expression

### LINKED ISSUE #2041: [Enh]: Expr.fill_null(value=an_expression)

Hi, would it be possible (and wanted) to extend `Expr.fill_null` so that it accepts not just a scalar but also an Expression? 

Personally, I am looking to solve a situation where null values of one column should be replaced by values of another (hopefully non-null) column but in general, the purpose would be to get closer to feature parity with polars (& pandas).

I would happily try to implement it (for polars and pandas at the least).

_No response_

_No response_

janpipek:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2041 

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -670,7 +670,8 @@ class ArrowSeries(CompliantSeries):
         dtype = ser.type
 
         if value is not None:
-            res_ser = self._from_native_series(pc.fill_null(ser, lit(value, dtype)))  # type: ignore[attr-defined]
+            _, value = extract_native(self, value)
+            res_ser = self._from_native_series(pc.fill_null(ser, value))  # type: ignore[attr-defined]
         elif limit is None:
             fill_func = (
                 pc.fill_null_forward if strategy == "forward" else pc.fill_null_backward
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -667,7 +667,6 @@ class ArrowSeries(CompliantSeries):
             )
 
         ser = self._native_series
-        dtype = ser.type
 
         if value is not None:
             _, value = extract_native(self, value)
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -11,7 +11,9 @@ from typing import cast
 from duckdb import CaseExpression
 from duckdb import CoalesceOperator
 from duckdb import ColumnExpression
+from duckdb import Expression
 from duckdb import FunctionExpression
+from duckdb import InvalidInputException
 from duckdb.typing import DuckDBPyType
 
 from narwhals._duckdb.expr_dt import DuckDBExprDateTimeNamespace
@@ -450,9 +452,13 @@ class DuckDBExpr(CompliantExpr["DuckDBLazyFrame", "duckdb.Expression"]):  # type
             msg = "todo"
             raise NotImplementedError(msg)
 
-        return self._from_call(
-            lambda _input: CoalesceOperator(_input, lit(value)), "fill_null"
-        )
+        def func(_input: Expression, value: Any) -> Expression:
+            try:
+                return CoalesceOperator(_input, value)
+            except InvalidInputException:
+                return CoalesceOperator(_input, lit(value))
+
+        return self._from_call(func, "fill_null", value=value)
 
     def cast(self: Self, dtype: DType | type[DType]) -> Self:
         def func(_input: duckdb.Expression) -> duckdb.Expression:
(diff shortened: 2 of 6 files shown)


narwhals/_duckdb/expr.py
```diff
@@ -450,9 +452,13 @@ def fill_null(self: Self, value: Any, strategy: Any, limit: int | None) -> Self:
...
             raise NotImplementedError(msg)
 
-        return self._from_call(
-            lambda _input: CoalesceOperator(_input, lit(value)), "fill_null"
-        )
+        def func(_input: Expression, value: Any) -> Expression:
+            try:
+                return CoalesceOperator(_input, value)
+            except InvalidInputException:
+                return CoalesceOperator(_input, lit(value))
```
janpipek, line 459: Is there a more clever way to check we are looking at a literal?

narwhals/expr.py
```diff
@@ -1352,27 +1353,29 @@ def fill_null(
             ...     {
             ...         "a": [2, None, None, 3],
             ...         "b": [2.0, float("nan"), float("nan"), 3.0],
+            ...         "c": [1, 2, 3, 4],
             ...     }
             ... )
             >>> df = nw.from_native(df_native)
             >>> df.with_columns(
-            ...     nw.col("a", "b").fill_null(0).name.suffix("_nulls_filled")
```
janpipek, line 1359: Sorry if stripping the "_nulls" can be misleading. I tried to keep the table at a reasonable width.

narwhals/_duckdb/expr.py
MarcoGorelli: thanks for working on this! can you check what we do for other functions where an input might be an expression? i think we should have some logic for this

janpipek: Ha, actually the mere changing the way how value is passed to _from_call did the job. The try-block is unnecessary :-)

MarcoGorelli:
nice one, thanks! any reason pyspark is xfailed?

MarcoGorelli:
😆 maybe it's just something that still needs doing

will check again later but this looks like it might be the solution - nice one!

janpipek:
Sorry, I still need to fix the doctest (thanks, CI)...

MarcoGorelli:
downstream test failures are unrelated btw, you can ignore them

FBruzzesi:
> nice one, thanks! any reason pyspark is xfailed?

@MarcoGorelli we.... simply don't have `fill_null` implemented for spark-like 🙈

MarcoGorelli (approved):
thanks @janpipek !

if you fancied trying pyspark as a follow-up, that would be welcome! you can use `sqlframe` to test it locally if you wish, that's much lighter and easier to setup, you can run its tests with `pytest --constructors=sqlframe`

narwhals/_arrow/series.py
```diff
@@ -667,10 +667,10 @@ def fill_aux(
             )
 
         ser = self._native_series
-        dtype = ser.type
 
         if value is not None:
-            res_ser = self._from_native_series(pc.fill_null(ser, lit(value, dtype)))  # type: ignore[attr-defined]
+            _, value = extract_native(self, value)
+            res_ser = self._from_native_series(pc.fill_null(ser, value))  # type: ignore[attr-defined]
```
dangotbanned, line 673: @janpipek could you explain this change for me please?

Not sure what the commit message is about (https://github.com/narwhals-dev/narwhals/pull/2106/commits/101bfb8bd0f5b99eec777e3ecb3de15be30c9cde)

The reason I ask is because this is causing a merge conflict for me - and I don't want to ignore your change if it solved something

image

janpipek: I am not sure what you're trying to achieve (whatever `pc` is) so I will just explain my change: In the previous version, `value` was always taken as literal. The change allows it to be any expression and hence the `_, value = extract_native(self, value)` line. I guess you should be able to use this converted `value` instead of `lit(value)` in your line.

Hope this helps.

janpipek: Feel free to change anything. I have very little arrow knowledge and hence, I was implementing only against the unit test.

dangotbanned: Thanks for explaining @janpipek!

I was expecting the same thing there as the annotation didn't change - only the description in https://github.com/narwhals-dev/narwhals/pull/2106/files#diff-8dbec323eb770e099734f31eb6dd160668841972f5549ea2d00779ea65853ec5

No worries - I've opened #2129 to address it 🙂

Issue #2068, opened 2025-02-22
docs: make docstrings concise in expr_str, expr_dt, series_str, ... and more


MarcoGorelli:
See https://github.com/narwhals-dev/narwhals/pull/1974 for an example of how to make the docstring examples more concise

We've still got the overly long examples in several docstrings

skritsotalakis:
I can start with expr_str and series_str

dangotbanned:
@MarcoGorelli do you know of any that are outstanding?

MarcoGorelli:
this is probably done - thanks all for your help here, really appreciate it 🙏

dangotbanned:
> this is probably done - thanks all for your help here, really appreciate it 🙏

Happy to help out where I can :blush:

PR #2117, opened 2025-02-28
docs: more concise docstrings(#2068)

skritsotalakis:
* expr_dt
  * expr_cat
  * dataframe
  * dtypes
  * exp_str - fix slice-
  * expr_list
  * expr_name

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

### DIFF
--- a/narwhals/expr_dt.py
+++ b/narwhals/expr_dt.py
@@ -28,34 +28,13 @@ class ExprDateTimeNamespace(Generic[ExprT]):
 
         Examples:
             >>> from datetime import datetime
-            >>> import pandas as pd
             >>> import polars as pl
-            >>> import pyarrow as pa
             >>> import narwhals as nw
-            >>> from narwhals.typing import IntoFrameT
-            >>>
-            >>> data = {
-            ...     "a": [datetime(2012, 1, 7, 10, 20), datetime(2023, 3, 10, 11, 32)]
-            ... }
-            >>> df_pd = pd.DataFrame(data).convert_dtypes(dtype_backend="pyarrow")
-            >>> df_pl = pl.DataFrame(data)
-            >>> df_pa = pa.table(data)
-
-            We define a library agnostic function:
-
-            >>> def agnostic_dt_date(df_native: IntoFrameT) -> IntoFrameT:
-            ...     df = nw.from_native(df_native)
-            ...     return df.select(nw.col("a").dt.date()).to_native()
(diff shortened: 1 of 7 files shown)


narwhals/expr_dt.py
```diff
@@ -1012,52 +554,23 @@ def total_microseconds(self: Self) -> ExprT:
             however, pandas may change the dtype to float when there are missing values,
             consider using `fill_null()` and `cast` in this case.
 
+        Examples:
         Examples:
```
MarcoGorelli, line 558: no need to have this twice

MarcoGorelli (commented):
Thanks!

could you please double-check for accidental duplication like the one I highlighted?

also, the data in many of these `expr_dt` examples can probably be shortened so it doesn't span as many lines, but that's ok as a follow-up

narwhals/expr_dt.py
```diff
@@ -426,53 +239,21 @@ def second(self: Self) -> ExprT:
...
             >>> import narwhals as nw
-            >>> from narwhals.typing import IntoFrameT
-            >>>
-            >>> data = {
-            ...     "datetime": [
-            ...         datetime(1978, 1, 1, 1, 1, 1),
-            ...         datetime(2024, 10, 13, 5, 30, 14),
-            ...         datetime(2065, 1, 1, 10, 20, 30),
-            ...     ]
-            ... }
```
skritsotalakis, line 441: @MarcoGorelli 

Changing the column name to "a" and inserting 2 datatimes instead of 3 doesn't improve much in long datetime entries. Introducing ```from datetime import datatime as dtime``` reduces the output to 3 lines (bellow example) but i m not sure this is desirable.
            >>> from datetime import datetime as dtime
            >>> df_native = pa.table(
            ...     {"a": [dtime(1978, 1, 1, 1, 1, 1), dtime(2065, 1, 1, 10, 20, 30)]}
            ... )

MarcoGorelli: sure thanks, don't worry about it then

if you just check of repetitions like `Examples` twice or accidental whitespace then we can get this in 👍

MarcoGorelli (approved):
thanks @skritsotalakis

PR #2118, opened 2025-02-28
ci: Add darts to downstream tests

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

[darts#2661](https://github.com/unit8co/darts/pull/2661) just got merged

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Darts entire test suite takes a bit too much for our use case - Narwhals is used in `TimeSeries` class, that's what I am adding as test here

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -523,10 +523,13 @@ jobs:
       - name: install-narwhals-dev
         run: |
           uv pip uninstall narwhals --system
-          uv pip install -e . --system
+          # TODO(FBruzzesi): Install polars from darts requirements when they introcude it
+          uv pip install -e ".[polars]" --system
       - name: show-deps
         run: uv pip freeze
       - name: Run tests
         run: |
           cd darts
           pytest darts/tests/test_timeseries.py
+          pytest darts/tests/test_timeseries_multivariate.py
+          pytest darts/tests/test_static_covariates.py
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -491,3 +491,40 @@ jobs:
           touch tests/utils/__init__.py
           pytest tests
         timeout-minutes: 15
+  
+  darts:
+    strategy:
+      matrix:
+        python-version: ["3.11"]
+        os: ["ubuntu-latest"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "pyproject.toml"
+      - name: clone-darts
+        run: |
+          git clone https://github.com/unit8co/darts.git --depth=1
+          cd darts
+          git log
+      - name: install-deps
+        run: |
+          cd darts
+          uv pip install -r requirements/core.txt -r requirements/dev.txt --system
+      - name: install-narwhals-dev
+        run: |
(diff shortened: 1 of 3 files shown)


FBruzzesi:
Hey @dennisbader and @authierj, we try to run test suites for downstream libraries that use narwhals in our CI, to make sure that we don't break anything by accident.

As mentioned in the description, darts test suite takes ~30mins, but narwhals is used just in `TimeSeries` test. Do you think that's enough for now to test?

dennisbader:
Hi @FBruzzesi, that's very nice 🚀 

I'd also add the following tests (they're very lightweight):
- `test_timeseries_multivariate.py`, contains narwhals related checks for multivariate DataFrames.
- `test_static_covariates.py`, contains narwhals related checks for DataFrames and static covariates.
- we'll be adding checks using polars as backend soon. Polars will only be an optional dependency and not be part of our `core.txt` requirements file (nor `dev.txt`). But it looks like you already install it for the tests, so all good 👍

FBruzzesi:
> I'd also add the following tests (they're very lightweight):
> 
> * `test_timeseries_multivariate.py`, contains narwhals related checks for multivariate DataFrames.
> * `test_static_covariates.py`, contains narwhals related checks for DataFrames and static covariates.

Thanks @dennisbader , I just added these two tests as well

> * we'll be adding checks using polars as backend soon. Polars will only be an optional dependency and not be part of our `core.txt` requirements file (nor `dev.txt`). But it looks like you already install it for the tests, so all good 👍

Actually we don't install polars if not needed - I added it now, with a TODO note to use darts specification when you are going to introduce it in CI/some requirements

.github/workflows/downstream_tests.yml
```diff
@@ -493,3 +493,43 @@ jobs:
...
+          git log
+      - name: install-deps
+        run: |
+          cd darts
+          uv pip install -r requirements/core.txt -r requirements/dev.txt --system
+      - name: install-narwhals-dev
+        run: |
+          uv pip uninstall narwhals --system
+          # TODO(FBruzzesi): Install polars from darts requirements when they introcude it
+          uv pip install -e ".[polars]" --system
```
dennisbader, line 527: We thought about not adding an extra "polars" to the darts package setup and let the user manually install it. Otherwise we'd also have to add pyarrow, ... extras.

dennisbader (approved):
Looks great @FBruzzesi 🚀 

Just added a minor comment

.github/workflows/downstream_tests.yml
FBruzzesi: Thanks for the info @dennisbader
I will remove the TODO then, but I am just wondering, are you going to manually install in CI for testing?

dennisbader: That was our plan, yes. We will add a `requirements/optional.txt` requirements file that we only install in one of our CI checks that is supposed to install "all" + optional dependencies. See here:

https://github.com/unit8co/darts/blob/393355275fd9d7a0a59678687f92b13d591aba47/.github/workflows/develop.yml#L70

Would you advice against that?

FBruzzesi: That's perfectly fine, we can eventually use such file then - similar for how we handle it for [plotly in CI](https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml#L309)

FBruzzesi:
I am going to merge as is for now - I believe the TODO is still relevant for the future 😉

dennisbader:
Hi @FBruzzesi, just wanted to let you know that we've activated polars in our tests now (in this [PR](https://github.com/unit8co/darts/pull/2701)). We added polars to `requirements/optional.txt`.

Issue #2083, opened 2025-02-24
nw.exclude


MarcoGorelli:
I think this one could be quite nice to add, alongside our existing `nw.nth`, `nw.all` and `nw.col`. As with `nw.col`, I'd like to keep this one limited to column names - for selection with dtypes / regexes, we have selectors

PR #2079, opened 2025-02-23
perf: avoid repeating constant iteration

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Either evaluate prior to defining the function, or skip converting a tuple to a list

PR #2122, opened 2025-03-01
feat: Adds `nw.exclude`

thomasjpfan:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes https://github.com/narwhals-dev/narwhals/issues/2083

## Checklist

- [x] Tests added
- [x] Documented the changes

This PR adds `nw.exclude` and unit tests.

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -5,6 +5,7 @@ from functools import reduce
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
+from typing import Container
 from typing import Iterable
 from typing import Literal
 from typing import Sequence
@@ -120,15 +121,14 @@ class ArrowNamespace(CompliantNamespace[ArrowDataFrame, ArrowSeries]):
             *column_names, backend_version=self._backend_version, version=self._version
         )
 
-    def exclude(self: Self, *column_names: str) -> ArrowExpr:
+    def exclude(self: Self, column_names: Container[str]) -> ArrowExpr:
         from narwhals._arrow.series import ArrowSeries
 
         def evaluate_output_names(df: ArrowDataFrame) -> Sequence[str]:
-            exclude_columns = set(column_names)
             return [
                 column_name
                 for column_name in df.columns
-                if column_name not in exclude_columns
+                if column_name not in column_names
             ]
 
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -121,14 +121,14 @@ class ArrowNamespace(CompliantNamespace[ArrowDataFrame, ArrowSeries]):
             *column_names, backend_version=self._backend_version, version=self._version
         )
 
-    def exclude(self: Self, column_names: Container[str]) -> ArrowExpr:
+    def exclude(self: Self, excluded_names: Container[str]) -> ArrowExpr:
         from narwhals._arrow.series import ArrowSeries
 
         def evaluate_output_names(df: ArrowDataFrame) -> Sequence[str]:
             return [
                 column_name
                 for column_name in df.columns
-                if column_name not in column_names
+                if column_name not in excluded_names
             ]
 
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -120,6 +120,39 @@ class ArrowNamespace(CompliantNamespace[ArrowDataFrame, ArrowSeries]):
             *column_names, backend_version=self._backend_version, version=self._version
         )
 
+    def exclude(self: Self, *column_names: str) -> ArrowExpr:
+        from narwhals._arrow.expr import ArrowExpr
+        from narwhals._arrow.series import ArrowSeries
+
+        def evaluate_output_names(df: ArrowDataFrame) -> Sequence[str]:
+            exclude_columns = set(column_names)
+            return [
+                column_name
+                for column_name in df.columns
+                if column_name not in exclude_columns
+            ]
+
+        def func(df: ArrowDataFrame) -> list[ArrowSeries]:
+            return [
+                ArrowSeries(
+                    df._native_frame[column_name],
+                    name=column_name,
+                    backend_version=df._backend_version,
+                    version=df._version,
+                )
+                for column_name in evaluate_output_names(df)
+            ]
+
+        return ArrowExpr(
+            func,
+            depth=0,
+            function_name="exclude",
+            evaluate_output_names=evaluate_output_names,
+            alias_output_names=None,
+            backend_version=self._backend_version,
+            version=self._version,
+        )
+
     def nth(self: Self, *column_indices: int) -> ArrowExpr:
         from narwhals._arrow.expr import ArrowExpr
 
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -121,7 +121,6 @@ class ArrowNamespace(CompliantNamespace[ArrowDataFrame, ArrowSeries]):
         )
 
     def exclude(self: Self, *column_names: str) -> ArrowExpr:
-        from narwhals._arrow.expr import ArrowExpr
         from narwhals._arrow.series import ArrowSeries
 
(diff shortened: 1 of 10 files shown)


narwhals/_arrow/namespace.py
```diff
@@ -120,6 +120,39 @@ def col(self: Self, *column_names: str) -> ArrowExpr:
...
+
+        return ArrowExpr(
+            func,
+            depth=0,
+            function_name="exclude",
+            evaluate_output_names=evaluate_output_names,
+            alias_output_names=None,
+            backend_version=self._backend_version,
+            version=self._version,
+        )
```
FBruzzesi, line 154: You can safe a couple of lines by doing:

```suggestion
        return self._create_expr_from_callable(
            func=func,
            depth=0,
            function_name="exclude",
            evaluate_output_names=evaluate_output_names,
            alias_output_names=None,
        )
```

narwhals/_arrow/namespace.py
```diff
@@ -120,6 +120,39 @@ def col(self: Self, *column_names: str) -> ArrowExpr:
...
+        def func(df: ArrowDataFrame) -> list[ArrowSeries]:
+            return [
+                ArrowSeries(
+                    df._native_frame[column_name],
+                    name=column_name,
+                    backend_version=df._backend_version,
+                    version=df._version,
+                )
+                for column_name in evaluate_output_names(df)
+            ]
```
FBruzzesi, line 144: I wonder if we could tweak `ArrowExpr.from_column_names` and use it here 🤔 It might be a refactor follow up

FBruzzesi (commented):
Thanks for the contribution @thomasjpfan 🚀

I left a few comments - those in the `_arrow` module apply to all other modules. In general we might try lower some repetition: `evaluate_output_names` is always the same function for all backends and the selection is quite close to `ComlpliantExpr.from_column_names`. Maybe we could generalize that a bit to be re-usable

narwhals/functions.py
```diff
@@ -1040,6 +1040,42 @@ def func(plx: Any) -> Any:
...
+        |  shape: (2, 1)   |
+        |  ┌─────┐         |
+        |  │ b   │         |
+        |  │ --- │         |
+        |  │ i64 │         |
+        |  ╞═════╡         |
+        |  │ 3   │         |
+        |  │ 4   │         |
+        |  └─────┘         |
+        └──────────────────┘
```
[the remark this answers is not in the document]
thomasjpfan: I updated to use `nw.exclude("c", "a")` to show case excluding two items.

tests/expr_and_series/exclude_test.py
```diff
@@ -0,0 +1,30 @@
...
+if TYPE_CHECKING:
+    from tests.utils import Constructor
+
+
+@pytest.mark.parametrize(
+    ("exclude_selector", "expected_cols"),
+    [
+        (nw.exclude("a"), ["b", "z"]),
+        (nw.exclude("b", "z"), ["a"]),
+    ],
```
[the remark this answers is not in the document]
thomasjpfan: Updated the PR with these. I do not think we support `nw.exclude(["b"], "z")` (mixing list and strings)

narwhals/_arrow/namespace.py
thomasjpfan: The refactor would be to accept able callable that returns the names.


```python
@classmethod
def from_column_names(
    cls,
    get_column_names,  # callable
    function_name,
    backend_version,
    version,
):
    def func(df):
        return [... for column_name in get_column_names(df)]

    return cls(
        func,
        function_name=function_name,
        evaluate_output_names=get_column_names
    )
```

Then `exclude` and `col` gets refactored to:

```python
def col(self: Self, *column_names: str):
    return ArrowExpr.from_column_names(
        lambda _: column_names, function_name="col", ...
    )

def exclude(self: Self, *column_names: str):
    def get_column_names(df) -> Sequence[str]:
        exclude_names = set(column_names)
        return [
            column_name
            for column_name in df.columns
            if column_name not in exclude_names
        ]

     return ArrowExpr.from_column_names(
         get_column_names,
         function_name="exclude", ...
     )
```

---

I did not want to increase the scope of this PR by changing the signature of `from_column_names` and changing `col`. I'm okay with doing a quick follow up.

narwhals/functions.py
```diff
@@ -1040,6 +1040,42 @@ def func(plx: Any) -> Any:
...
+    """
+
+    def func(plx: Any) -> Any:
+        return plx.exclude(*flatten(names))
```
dangotbanned, line 1074: I think you'd be able to simplify each backend by changing this to:

```suggestion
        return plx.exclude(set(flatten(names))
```

Then changing what they accept to be a single (non-variadic) argument:
```py
column_names: set[str]
```

narwhals/functions.py
```diff
@@ -1040,6 +1040,42 @@ def func(plx: Any) -> Any:
...
+        |  │ b   │         |
+        |  │ --- │         |
+        |  │ i64 │         |
+        |  ╞═════╡         |
+        |  │ 3   │         |
+        |  │ 4   │         |
+        |  └─────┘         |
+        └──────────────────┘
+    """
+
```
dangotbanned, line 1072: https://github.com/narwhals-dev/narwhals/pull/2122/files/d733bb24b79c1f79f3033b8433e82f76cf056c1d#r1976440039

Defining it here would save the cost of re-evaluating on each call
```suggestion
    exclude_names = frozenset(flatten(column_names))

```

Then passing that to `plx.exclude(...)`

narwhals/functions.py
dangotbanned: Could alternatively annotate it as:
```py
column_names: Container[str]
```

Since only `__contains__` is used

narwhals/functions.py
```diff
@@ -1040,6 +1040,42 @@ def func(plx: Any) -> Any:
...
+    def func(plx: Any) -> Any:
+        return plx.exclude(frozenset(flatten(names)))
```
MarcoGorelli, line 1074: thanks for updating!

I think Dan's suggestion was to do something like
```python
flattened = frozenset(flatten(names))
    def func(plx: Any) -> Any:
        return plx.exclude(flattened)
return Expr(func, ExprMetadata.selector())
```
so that `flatten` doesn't get called repeatedly (admittedly, we should do the for `nth`, which you probably used a template? sorry for that if so)

dangotbanned: Yeah that was my thinking @MarcoGorelli, like
- #2079

Should be good in (https://github.com/narwhals-dev/narwhals/pull/2122/commits/8fa7acd128aace64afd4d20dcfaf63eb43823860)

narwhals/functions.py
dangotbanned: Resolved via (https://github.com/narwhals-dev/narwhals/pull/2122#discussion_r1976480393)

narwhals/functions.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2122/commits/07deb3a365d11a4ea1ee09313538a6af9aa3842f)

narwhals/_arrow/namespace.py
FBruzzesi: Thanks @thomasjpfan

>The refactor would be to accept able callable that returns the names.

Yes I think we should aim for that

> I did not want to increase the scope of this PR by changing the signature of from_column_names and changing col. I'm okay with doing a quick follow up.

Happy to keep it as follow up

FBruzzesi (approved):
Thanks @thomasjpfan 🚀 

SwiftmoTwitchGIF

narwhals/stable/v1/__init__.py
```diff
@@ -1812,6 +1812,18 @@ def col(*names: str | Iterable[str]) -> Expr:
     return _stableify(nw.col(*names))
 
 
+def exclude(*names: str | Iterable[str]) -> Expr:
```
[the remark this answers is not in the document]
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2122/commits/d733bb24b79c1f79f3033b8433e82f76cf056c1d)

dangotbanned (approved):
Thanks everyone (especially @thomasjpfan) :tada:

PR #2126, opened 2025-03-01
feat: Add support for full join

### LINKED ISSUE #2120: [Enh]: Add full join to Narwhals

I think the join method should have a full option, to include rows from both left and right. Happy to implement on my own :)

This option is available in most dataframe libraries in some form or fashion, it would be a nice to have. 

Some native apis support this out of the box, for those who don't it may require a manual implementation.

_No response_

_No response_

tylerriccio33:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2120 

## Checklist

- [x] Tests added
- [x] Documented the changes

I did not add explicit documentation other than the proper type hints in the respective libraries. If you think the option needs an example then I will insert!

I did run into an interesting issue relating to polars... Arrow and pandas automatically coalesce the join key but polars does not. In my pr I left it up to the native library to implement, but I could see reasons why this isn't optimal. I'd like the perspective of the maintainers on this aspect, and am happy to implement whatever solution is in line w/the libraries' philosophy.


FBruzzesi:
Hey @tylerriccio33 , thanks a ton for your contribution 🙏🏼

> I did not add explicit documentation other than the proper type hints in the respective libraries. If you think the option needs an example then I will insert!

It's ok to not have an explicit example, maybe simply add a line to the list to describe what's a full join in the docstrings for [DataFrame](https://github.com/narwhals-dev/narwhals/blob/9387ecd70bc2bbeddc7fe2c4c7d29818b496f150/narwhals/dataframe.py#L1706-L1710) and [LazyFrame](https://github.com/narwhals-dev/narwhals/blob/9387ecd70bc2bbeddc7fe2c4c7d29818b496f150/narwhals/dataframe.py#L2997-L3001).

> I did run into an interesting issue relating to polars... Arrow and pandas automatically coalesce the join key but polars does not. In my pr I left it up to the native library to implement, but I could see reasons why this isn't optimal.

We should aim for the same behaviour across all libraries. If that's not the native one, we should try to remap it. Ideally to the polars one (as narwhals is a subset of polars). In this specific case:

- It might not be trivial doing so
- @MarcoGorelli might have more insights if polars behavior is going to change in the future

---

One last consideration regarding the failing tests: for the backends that you didn't address, you might need to explicitly xfail the unit test. You can find many examples in which we do:

```py
def some_test(...):
    if <unsupported_backend> in str(constructor):
    
    ...
```

MarcoGorelli:
thanks for your pr! yeah i think we need to match the default coalescing behaviour from polars if we want to support this

tylerriccio33:
In the updated pr I do the below:
- Update the docstings in LazyFrame and DataFrame
- Assert the non-coalesce behavior in the test
- Remap this coalesce logic for pyarrow

I only implemented this functionality for polars and pyarrow in this pr, for the simple fact that it was easy. Each library requires requires a bespoke solution if it does not expose a coalesce-keys-like argument. Pandas looks like it'll be particularly hard.

Is the bare minimum implementation for polars and arrow acceptable at this time? I can open separate prs for each library? I can assure you I'll need to fix duckdb at some point for my job; pandas I can't guarantee. Is this in line w/your philosophy or does the implementation need coverage of the core libraries?

MarcoGorelli:
thanks for your pr! i think this would need implementing for all backends

FBruzzesi:
> In the updated pr I do the below:
> 
> * Update the docstings in LazyFrame and DataFrame
> * Assert the non-coalesce behavior in the test
> * Remap this coalesce logic for pyarrow

Thanks @tylerriccio33 🙌🏼

> Pandas looks like it'll be particularly hard.

If you first add the suffix to the right frame keys and then merge, you should be able to match the behaviour

```py
df1 = pd.DataFrame({"id": [1, 2, 3], "value1": ["A", "B", "C"]})
df2 = pd.DataFrame({"id_right": [2, 3, 4], "value2": ["X", "Y", "Z"]})

df1.merge(df2, how="outer", left_on="id", right_on="id_right")
    id value1  id_right value2
0  1.0      A       NaN    NaN
1  2.0      B       2.0      X
2  3.0      C       3.0      Y
3  NaN    NaN       4.0      Z
```

Notice that if `<key>_right` already exists, you should match the polars error:

```terminal
DuplicateError: column with name 'id_right' already exists

You may want to try:
- renaming the column prior to joining
- using the `suffix` parameter to specify a suffix different to the default one ('_right')
```

> Is the bare minimum implementation for polars and arrow acceptable at this time? I can open separate prs for each library? I can assure you I'll need to fix duckdb at some point for my job; pandas I can't guarantee. Is this in line w/your philosophy or does the implementation need coverage of the core libraries?

While we don't have _core_ libraries, we should aim to at least a few backends

tylerriccio33:
Apologies for the big pr, if it must be implemented in many backends I don't see another way. I ran into logic issues making the join consistently return the same results.

tests/frame/join_test.py
```diff
@@ -10,13 +10,119 @@
...
+    df2 = {"foo": [1, 2, 3], "foo_right": [1, 2, 3]}
+    df_left = nw_main.from_native(constructor(df1))
+    df_right = nw_main.from_native(constructor(df2))
+
+    # polars throws `NarwhalsError`, everything else should raise `DuplicateError`
+    with pytest.raises((DuplicateError, nw.exceptions.NarwhalsError)):
+        df_left.join(df_right, on="foo", how="full")
+
+
+pytest.main([__file__, "--all-cpu-constructors"])
```
tylerriccio33, line 123: I accidentally left this artifact in here while testing. If you would kindly remove it :)

FBruzzesi: I pushed one commit and removed it

FBruzzesi:
Hey @tylerriccio33 thanks for the contribution, I think we are very close.

> Apologies for the big pr, if it must be implemented in many backends I don't see another way.

Don't worry, we have seen _way_ worse 😂

---

I added one commit to solve a conflict and to address some nitpicks.

What's left is:

- polars pre v1.0.0 - it does not support `how="full"`, so either we xfail that or implement it from our side
- pyspark is also failing. I am ok to xfail that for now and follow up to address a fix

MarcoGorelli:
Thanks @tylerriccio33 and @FBruzzesi !

I didn't check all the details but @FBruzzesi feel free to ship it if you're happy with it!

FBruzzesi (approved):
Thanks @tylerriccio33 - let's ship this, so that it makes it in today's release 🔥

NarwhalChubbiwhalGIF

PR #2119, opened 2025-02-28
feat(RFC): Spec'd-out `CompliantExpr`

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/issues/2044#issuecomment-2674262833
- #1807

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below
This all came together really quickly after wondering *if* it would be possible to improve the typing in `expr.py`.

For example, we currently get no feedback from a type checker here:
https://github.com/narwhals-dev/narwhals/blob/77a0150ffb5121086d74b0f9a10ec26781162529/narwhals/expr.py#L1070-L1074

That seemed pretty simple to solve - all we need is to annotate `_to_compliant_expr` (https://github.com/narwhals-dev/narwhals/commit/0a95012b52e39a0afe5b323cfca1c74659118c08).

However, that change led me down *this* rabbit hole of fully spec-ing out `CompliantExpr` - to be able to type everything correctly.

During that process - I spotted a link between (#2044) and (#1807).
I've borrowed from @camriddell's [`not_implemented`](https://github.com/narwhals-dev/narwhals/pull/1807#issuecomment-2612810281) idea (thanks btw :wink:) and *implemented* a very low-tech version of it.
With some extra magic - things are looking quite interesting now

{screenshot}

PR #2130, opened 2025-03-02
refactor: adds `_Stores(Native|Compliant)`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Spin-off from (#1301)
  - [comment 1](https://github.com/narwhals-dev/narwhals/pull/1301#issuecomment-2676238941)
  - [comment 2](https://github.com/narwhals-dev/narwhals/pull/1301#issuecomment-2689337509)
- https://github.com/narwhals-dev/narwhals/pull/2119#discussion_r1976024714

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Originally (https://github.com/narwhals-dev/narwhals/pull/1301#issuecomment-2686042530)
> @FBruzzesi I've tested this out for some parts of `_arrow` over on [main...stores-native-compliant](https://github.com/narwhals-dev/narwhals/compare/main...stores-native-compliant)
> 
> Really think it's a nice improvement to readabilty and [DRY](https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself)s things up well 🙂
> 
> With a lot of the repetition out of the way, I was able to spot a few more things that could be written more concisely.
> 
> Not sure if this is worth opening an issue for - but if we went ahead with it - then it'd probably make sense to try and do it in one or more big chunks to avoid conflicts

> [!NOTE]
> The parts that I've touched in `_arrow/*.py` are ready for review.
> Other modules have only had the *bare-minimum* changes to meet the new protocol(s)
> The goal would be to repeat this kind of pattern throughout `narwhals` (*pending feedback*)


narwhals/utils.py
```diff
@@ -112,6 +111,59 @@ class _StoresColumns(Protocol):
...
+
+
+class _ExprNamespace(  # type: ignore[misc] # noqa: PYI046
+    _StoresCompliant[CompliantExprT_co], Protocol[CompliantExprT_co]
+):
+    _compliant_expr: CompliantExprT_co
+
+    @property
+    def compliant(self) -> CompliantExprT_co:
+        return self._compliant_expr
```
dangotbanned, line 164: The core idea is providing (read-only) accessors and fast-paths to commonly used methods.

These are what I've felt were needed so far, but they can get more specific like for temporal properties:

https://github.com/narwhals-dev/narwhals/blob/5c642eb46c690a4b5829eebeac28f3b3be454dc3/narwhals/_arrow/series_dt.py#L24-L31

Ideally, I think these **should not be used** for methods accepting more than a single argument.
That way it keeps things pretty clean and doesn't introduce backend-specific details at this high of a level

narwhals/_arrow/series_dt.py
```diff
@@ -96,148 +96,99 @@ def timestamp(self: Self, time_unit: TimeUnit) -> ArrowSeries:
...
-        return self._compliant_series._from_native_series(
-            pc.second(self._compliant_series._native_series)
-        )
+        return self.from_native(pc.second(self.native))
 
     def millisecond(self: Self) -> ArrowSeries:
-        return self._compliant_series._from_native_series(
-            pc.millisecond(self._compliant_series._native_series)
-        )
+        return self.from_native(pc.millisecond(self.native))
```
dangotbanned, line 123: This is the most typical change - where it becomes significantly eaiser to see the **unique** thing the method is doing

### Before
```py
    def millisecond(self: Self) -> ArrowSeries:
        return self._compliant_series._from_native_series(
            pc.millisecond(self._compliant_series._native_series)
        )
```

### After
```py
    def millisecond(self: Self) -> ArrowSeries:
        return self.from_native(pc.millisecond(self.native))
```

narwhals/utils.py
```diff
@@ -118,6 +117,59 @@ class _StoresColumns(Protocol):
...
+    @property
+    def native(self) -> NativeT_co:
+        """Return the native object."""
+        ...
+
+
+class _StoresCompliant(Protocol[CompliantT_co]):
+    @property
+    def compliant(self) -> CompliantT_co:
+        """Return the narwhals object."""
```
MarcoGorelli, line 140: can we call it "compliant object"?

we probably need to document this better, but in my head the hierarchy is:
- narwhals object: nw.Series, nw.DataFrame, nw.LazyFrame, nw.Expr
- compliant object: PandasLikeSeries, PolarsDataFrame, ...
- native object: pd.Series, pa.Table, ...

MarcoGorelli (approved):
thanks @dangotbanned ! love these simplifications

just one really minor comment

narwhals/utils.py
dangotbanned: Yeah sure, "compliant object" makes more sense!

I think I mixed up the two (or three) when I wrote that doc

dangotbanned: Will take a look at merge conflicts tomorrow

narwhals/utils.py
```diff
@@ -118,6 +117,75 @@ class _StoresColumns(Protocol):
...
+
+
+class _StoresNative(Protocol[NativeT_co]):
+    """Provides access to a native object.
+
+    Native objects are types like:
+
+    >>> from pandas import Series
+    >>> from pyarrow import Table
+    """
```
dangotbanned, line 137: 100% expecting this doc to be questioned.

Was the only style that VSCode seemed to parse correctly:

image

<details><summary>Alternatives</summary>
<p>


image

image

image


</p>
</details>

narwhals/utils.py
dangotbanned: Resolved in:
- (https://github.com/narwhals-dev/narwhals/pull/2130/commits/7c4dea31ea498c068176b8cfdfd254ae38fbf8f1)
- (https://github.com/narwhals-dev/narwhals/pull/2130/commits/0d41a7a3b6799fb57286e9a06e2a3958d3c463db)

PR #2134, opened 2025-03-02
refactor: allow callable in `CompliantExpr.from_column_names` to reuse for `CompliantNamespace.<all|col|exclude>`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Follow-up on [comment](https://github.com/narwhals-dev/narwhals/pull/2122#discussion_r1976370057)

### DIFF
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -87,12 +87,11 @@ class ArrowExpr(CompliantExpr["ArrowDataFrame", ArrowSeries]):
     @classmethod
     def from_column_names(
         cls: type[Self],
-        *column_names: str,
+        get_column_names: Callable[[ArrowDataFrame], Sequence[str]],
+        function_name: str,
         backend_version: tuple[int, ...],
         version: Version,
     ) -> Self:
-        from narwhals._arrow.series import ArrowSeries
-
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
             try:
                 return [
@@ -102,10 +101,10 @@ class ArrowExpr(CompliantExpr["ArrowDataFrame", ArrowSeries]):
                         backend_version=df._backend_version,
                         version=df._version,
                     )
-                    for column_name in column_names
+                    for column_name in get_column_names(df)
                 ]
             except KeyError as e:
-                missing_columns = [x for x in column_names if x not in df.columns]
+                missing_columns = [x for x in get_column_names(df) if x not in df.columns]
                 raise ColumnNotFoundError.from_missing_and_available_column_names(
                     missing_columns=missing_columns, available_columns=df.columns
                 ) from e
@@ -113,8 +112,8 @@ class ArrowExpr(CompliantExpr["ArrowDataFrame", ArrowSeries]):
         return cls(
             func,
             depth=0,
-            function_name="col",
-            evaluate_output_names=lambda _df: column_names,
+            function_name=function_name,
+            evaluate_output_names=get_column_names,
             alias_output_names=None,
             backend_version=backend_version,
             version=version,
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -87,7 +87,7 @@ class ArrowExpr(CompliantExpr["ArrowDataFrame", ArrowSeries]):
     @classmethod
     def from_column_names(
         cls: type[Self],
-        get_column_names: Callable[[ArrowDataFrame], Sequence[str]],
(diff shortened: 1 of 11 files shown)


narwhals/_arrow/namespace.py
```diff
@@ -117,37 +117,29 @@ def __init__(
...
 
+        def get_column_names(_: ArrowDataFrame) -> Sequence[str]:
+            return column_names
+
         return ArrowExpr.from_column_names(
-            *column_names, backend_version=self._backend_version, version=self._version
+            get_column_names=get_column_names,
+            function_name="col",
+            backend_version=self._backend_version,
+            version=self._version,
```
dangotbanned, line 127: I think `get_column_names` is a bad name for this.

I chose that name to mean **exactly** this:

https://github.com/narwhals-dev/narwhals/blob/9d9fde2b43af33746efce9d730c6b97c2681bc51/narwhals/utils.py#L1355-L1356

Where `get_columns` could easily be mistaken for being related to these:

https://github.com/narwhals-dev/narwhals/blob/9d9fde2b43af33746efce9d730c6b97c2681bc51/narwhals/typing.py#L88-L99

---

Maybe its a personal taste thing - but I think something like `evaluate_*` makes a lot more sense because you're performing a transformation

narwhals/_arrow/expr.py
```diff
@@ -87,12 +87,11 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
     @classmethod
     def from_column_names(
         cls: type[Self],
-        *column_names: str,
+        get_column_names: Callable[[ArrowDataFrame], Sequence[str]],
```
dangotbanned, line 90: Putting (https://github.com/narwhals-dev/narwhals/pull/2134#discussion_r1976732319) aside - how about:

```py
@classmethod
    def from_column_names(
        cls: type[Self],
        get_column_names: Sequence[str],
        ...
```

Then you could avoid needing to create a new function for each backend and just reuse `utils.get_column_names` in `evaluate_output_names=get_column_names,`

- https://github.com/narwhals-dev/narwhals/pull/2102#issuecomment-2685117308

dangotbanned: Ah I saw this covers `col` and `exclude`.

In that case maybe:
```py
get_column_names: Sequence[str] | Callable[[ArrowDataFrame], Sequence[str]]
```

And then add a check for `callable(...)` somewhere?

narwhals/_dask/namespace.py
```diff
@@ -67,29 +67,27 @@ def func(df: DaskLazyFrame) -> list[dx.Series]:
...
         )
 
     def exclude(self: Self, excluded_names: Container[str]) -> DaskExpr:
-        def evaluate_output_names(df: DaskLazyFrame) -> Sequence[str]:
+        def get_column_names(df: DaskLazyFrame) -> Sequence[str]:
             return [
                 column_name
                 for column_name in df.columns
                 if column_name not in excluded_names
             ]
```
dangotbanned: ooh now that I'm thinking about it - isn't this the part we'd want to factor-out?

```py
# nw.utils.py (or a method on something)

def exclude_columns(frame: _StoresColumns, names: Container[str]):
    return [name for name in frame.columns if name not in names]
```

That should work for every backend IIRC?

dangotbanned: You could even do this if you needed to match the single-arg signature?

```py
from functools import partial

fn = partial(exclude_columns, names=excluded_names)
```

dangotbanned:
Apologies for the *stream-of-consciousness* @FBruzzesi 😉

FBruzzesi:
> Apologies for the _stream-of-consciousness_ @FBruzzesi 😉

No worries! There are a lot of good points and I realized `all` could be refactored as well! I should have opened this as draft to begin with. 

I am quite happy with it now - I should have addressed all your points except [comment](https://github.com/narwhals-dev/narwhals/pull/2134#discussion_r1976735018) for which I have mixed feelings.

narwhals/_dask/namespace.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2134/commits/efd5a227a082adebffd11a36fe3149445edacf3a)

FBruzzesi:
Thanks @dangotbanned 🙌🏼

PR #2138, opened 2025-03-03
feat: support anonymous expressions in `over` context for pandas-like, support `over` for multi-partitioned dask

MarcoGorelli:
This simplifies a lot on the pandas side, and unlocks `.over` for dask

On the downside, for some very old versions of pandas (pre 1.1), some tests would fail...but tbh those versions are so old i'm not overly bothered. pandas 1.0 is >5 years old...

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -534,9 +534,9 @@ class DaskExpr(CompliantExpr["DaskLazyFrame", "dx.Series"]):  # pyright: ignore[
         # pandas is a required dependency of dask so it's safe to import this
         from narwhals._pandas_like.group_by import AGGREGATIONS_TO_PANDAS_EQUIVALENT
 
-        if not is_elementary_expression(self):
+        if not is_elementary_expression(self):  # pragma: no cover
             msg = (
-                "Only elementary expressions are supported for `.over` in pandas-like backends.\n\n"
+                "Only elementary expressions are supported for `.over` in dask.\n\n"
                 "Please see: "
                 "https://narwhals-dev.github.io/narwhals/pandas_like_concepts/improve_group_by_operation/"
             )
--- a/narwhals/_dask/expr.py
+++ b/narwhals/_dask/expr.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import re
+import warnings
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
@@ -14,6 +16,7 @@ from narwhals._dask.utils import maybe_evaluate_expr
 from narwhals._dask.utils import narwhals_to_native_dtype
 from narwhals._expression_parsing import ExprKind
 from narwhals._expression_parsing import evaluate_output_names_and_aliases
+from narwhals._expression_parsing import is_elementary_expression
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import InvalidOperationError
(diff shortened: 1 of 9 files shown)


narwhals/_expression_parsing.py
```diff
@@ -250,7 +250,8 @@ def is_simple_aggregation(expr: CompliantExpr[Any, Any]) -> bool:
 
         - nw.col('a').filter(nw.col('b')>nw.col('c')).max()
 
-    because then, we can use a fastpath in pandas.
+    Elementary ones are the only ones supported properly in
```
EdAbati, line 253:
```suggestion
    Elementary expressions are the only ones supported properly in
```
is this what you meant? :)

narwhals/_dask/expr.py
```diff
@@ -528,29 +531,40 @@ def null_count(self: Self) -> Self:
...
+
+        if not is_elementary_expression(self):  # pragma: no cover
             msg = (
-                "`Expr.over` is not supported for Dask backend with multiple partitions."
+                "Only elementary expressions are supported for `.over` in dask.\n\n"
+                "Please see: "
+                "https://narwhals-dev.github.io/narwhals/pandas_like_concepts/improve_group_by_operation/"
             )
             raise NotImplementedError(msg)
+        function_name = re.sub(r"(\w+->)", "", self._function_name)
```
EdAbati, line 544: should we add (in a followup) a new property that does this?

narwhals/_pandas_like/expr.py
```diff
@@ -48,6 +48,32 @@
...
+    function_name: str, kwargs: dict[str, object]
+) -> dict[str, object]:
+    unsupported_reverse_msg = (
+        "Cumulative operation with `reverse=True` is not supported in "
+        "over context for pandas-like backend."
+    )
+    if function_name == "shift":
+        pandas_kwargs: dict[str, object] = {"periods": kwargs["n"]}
+    elif function_name == "rank":
+        _method = kwargs["method"]
```
EdAbati, line 61: just out of curiosity, would it be possible that the "method" key (or "descending") is not present?

narwhals/_pandas_like/expr.py
```diff
@@ -48,6 +48,32 @@
 }
 
 
+def window_kwargs_to_pandas_equivalent(
+    function_name: str, kwargs: dict[str, object]
+) -> dict[str, object]:
```
EdAbati, line 53: Again out of curiosity, is `dict[str, Any]` different from `dict[str, object]` ?

EdAbati (approved):
This is a great simplification!

I left mostly questions for follow ups :) but IMO we can merge

narwhals/_pandas_like/expr.py
dangotbanned: It is different, but maybe not in a meaningful way here.

Personally, I'd use `Any` unless there's a good reason for `object`

https://typing.readthedocs.io/en/latest/spec/special-types.html#any

narwhals/_dask/expr.py
MarcoGorelli: yup, sure. this whole mechanism should probably be refactored anyway

narwhals/_pandas_like/expr.py
MarcoGorelli: no, we always pass it

narwhals/_pandas_like/expr.py
MarcoGorelli: Isn't `object` the safest choice? Else we risk using some attribute which doesn't exist https://adamj.eu/tech/2021/05/07/python-type-hints-use-object-instead-of-any/

dangotbanned: @MarcoGorelli I mean there doesn't seem to be any attribute access here?

Stricly speaking - this would probably benefit from either a `dict[Literal[...], Any]`, a `TypedDict` or a non-`dict` data structure.

`pandas_kwargs` seems to have well-defined keys - so a narrower type could be:

```py

_PandasKey: TypeAlias = Literal[
    "periods", "method", "ascending", "na_option", "pct", "skip_na"
]
```

For `kwargs`, you seem to be inferring the following:

```py
_KwargsKey: TypeAlias = Literal["n", "method", "descending", "reverse"]
```

Putting that all together - I think this is much more useful than `dict[str, object]`:

```py
_WindowFnName: TypeAlias = Literal[
    "cum_sum", "cum_min", "cum_max", "cum_prod", "cum_count", "shift", "rank", "diff"
]
_PandasKey: TypeAlias = Literal[
    "periods", "method", "ascending", "na_option", "pct", "skipna"
]
_KwargsKey: TypeAlias = Literal["n", "method", "descending", "reverse"]


def window_kwargs_to_pandas_equivalent(
    function_name: _WindowFnName, kwargs: dict[_KwargsKey, Any]
) -> dict[_PandasKey, Any]:
    unsupported_reverse_msg = (
        "Cumulative operation with `reverse=True` is not supported in "
        "over context for pandas-like backend."
    )
    if function_name == "shift":
        pandas_kwargs: dict[_PandasKey, Any] = {"periods": kwargs["n"]}
    elif function_name == "rank":
        _method = kwargs["method"]
        pandas_kwargs = {
            "method": "first" if _method == "ordinal" else _method,
            "ascending": not kwargs["descending"],
            "na_option": "keep",
            "pct": False,
        }
    elif function_name.startswith("cum_"):  # Cumulative operation
        if kwargs["reverse"]:
            raise NotImplementedError(unsupported_reverse_msg)
        pandas_kwargs = {"skipna": True}
    else:
        pandas_kwargs = {}
    return pandas_kwargs
```

And now you get warnings leading to the actual source of `object`.
You would be able to type this all correctly by switching to [look before you leap](https://realpython.com/python-lbyl-vs-eafp/#the-look-before-you-leap-lbyl-style)

image

PR #2140, opened 2025-03-03
refactor: Use `chain.from_iterable` more

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- https://github.com/astral-sh/ruff/pull/16131
- https://github.com/narwhals-dev/narwhals/pull/2028#discussion_r1957854827

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable
- Even if there were no performance benefit, I find this much easier to read than nested comprehensions
- There are likely a lot more places we could do this
  - Kinda difficult to search for

### DIFF
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import operator
 from functools import partial
 from functools import reduce
+from itertools import chain
 from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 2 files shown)


MarcoGorelli (approved):
thanks @dangotbanned !

narwhals/_arrow/namespace.py
```diff
@@ -238,7 +236,7 @@ def mean_horizontal(self: Self, *exprs: ArrowExpr) -> IntoArrowExpr:
         dtypes = import_dtypes_module(self._version)
 
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
-            expr_results = [s for _expr in exprs for s in _expr(df)]
+            expr_results = list(chain.from_iterable(expr(df) for expr in exprs))
```
FBruzzesi, line 239: Nitpick: can we materialize the generator into a tuple instead of a list (here and elsewhere)? 😇

dangotbanned: @FBruzzesi sure, shouldn't see why not

I only picked a `list` since that was what already there.

I think there's probably some performance left on the table, by picking the right data structures for each task.

[`deque`](https://docs.python.org/3/library/collections.html#collections.deque) is usually what I prefer when you need mutability but not random access

PR #1876, opened 2025-01-27
chore: refactor `name` namespaces to lower code duplication

FBruzzesi:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2146, opened 2025-03-04
feat: Added `struct` namespace with `field` method.

osoucy:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

Introduction of the strut namespace for applicable expressions and series with an initial `field` function that returns a field of a struct.


narwhals/_polars/expr.py
```diff
@@ -391,3 +395,14 @@ def func(*args: Any, **kwargs: Any) -> PolarsExpr:
...
+
+class PolarsExprStructNamespace:
+    def __init__(self: Self, expr: PolarsExpr) -> None:
+        self._expr = expr
+
+    def field(self: Self, name: str) -> PolarsExpr:
+        native_expr = self._expr._native_expr
+        native_result = native_expr.struct.field(name)
+
+        return self._expr._from_native_expr(native_result)
```
EdAbati, line 408: I cannot try because I'm on mobile, but could we use __getattr__ here like we did above? 👀

docs/api-completeness/expr_struct.md
```diff
@@ -0,0 +1,5 @@
+# Expr.str
```
EdAbati, line 1:
```suggestion
# Expr.struct
```

docs/api-completeness/series_struct.md
```diff
@@ -0,0 +1,5 @@
+# Series.str
```
EdAbati, line 1:
```suggestion
# Series.struct
```

narwhals/_pandas_like/series_struct.py
```diff
@@ -0,0 +1,18 @@
...
+    from narwhals._pandas_like.series import PandasLikeSeries
+
+
+class PandasLikeSeriesStructNamespace:
+    def __init__(self: Self, series: PandasLikeSeries) -> None:
+        self._compliant_series = series
+
+    def field(self: Self, name: str) -> PandasLikeSeries:
+        return self._compliant_series._from_native_series(
+            self._compliant_series._native_series.apply(lambda x: x[name]).rename(name),
```
MarcoGorelli, line 17: Instead of using apply can we use https://pandas.pydata.org/pandas-docs/dev/reference/api/pandas.Series.struct.field.html and only support this for pyarrow-backed dtypes?

EdAbati (commented):
This looks great! Thank you for doing this 🙌🏼

I've added a couple of comments 'on the go' (I'll check the rest later)

narwhals/expr_struct.py
```diff
@@ -0,0 +1,45 @@
...
+            ...     {
+            ...         "user": [
+            ...             {"id": "0", "name": "john"},
+            ...             {"id": "1", "name": "jane"},
+            ...         ]
+            ...     }
+            ... )
+            >>> df = nw.from_native(df_native)
+            >>> df.with_columns(name=nw.col("user").struct.field("name"))
+        """
```
FBruzzesi, line 41: Output is missing and doctest check would fail without it

narwhals/series_struct.py
```diff
@@ -0,0 +1,43 @@
...
+            >>> import narwhals as nw
+            >>> s_native = pl.Series(
+            ...     [
+            ...         {"id": "0", "name": "john"},
+            ...         {"id": "1", "name": "jane"},
+            ...     ]
+            ... )
+            >>> s = nw.from_native(s_native, series_only=True)
+            >>> s.struct.field("name").to_native()
+        """
```
FBruzzesi, line 40: Same as for the Expr case, output is missing

FBruzzesi (commented):
Thanks @osoucy - it generally looks very promising. Left an additional comment to those of Edo and Marco

narwhals/_pandas_like/series_struct.py
osoucy: I feel the majority of Pandas users don't leverage pyarrow-backed dtypes. I added a switch to check and use the `struct` namespace if available and fallbacks on the `apply` function if it's not the case.

narwhals/expr_struct.py
osoucy: The output has to be manually generated? There is no command that I can use to generate it?

narwhals/_polars/expr.py
osoucy: We can! I'm not sure what is the purpose though?

narwhals/_pandas_like/series_struct.py
```diff
@@ -0,0 +1,23 @@
...
+    def __init__(self: Self, series: PandasLikeSeries) -> None:
+        self._compliant_series = series
+
+    def field(self: Self, name: str) -> PandasLikeSeries:
+        series = self._compliant_series._native_series
+
+        if hasattr(series, "struct"):
+            series = series.struct.field(name)
+        else:
+            series = series.apply(lambda x: x[name])
```
MarcoGorelli, line 21: can we just error here? I think it's ok to require that people use pyarrow-dtypes for `struct` operations in pandas

osoucy: Well, it's your call, but I think many users would benefit from having this feature available even if they are not using pyarrow data types. And I suspect most people are not using it.

dangotbanned: > can we just error here? I think it's ok to require that people use pyarrow-dtypes for `struct` operations in pandas

If that's the intent, maybe enforcing that in either of these places would make more sense?

https://github.com/okube-ai/narwhals/blob/37dc7fc34b45dbc42ce3e590a43a8674e24ca80f/narwhals/_pandas_like/series.py#L1111-L1113

https://github.com/okube-ai/narwhals/blob/37dc7fc34b45dbc42ce3e590a43a8674e24ca80f/narwhals/_pandas_like/series_struct.py#L11-L13

That way - `field` doesn't need error handling and any future methods won't either.
All `struct` ops will require `pyarrow`

MarcoGorelli: yeah init should be fine - we could do the same for the list namespace. or just use `.struct` and let pandas raise its own error

osoucy: Done!

narwhals/_arrow/expr_struct.py
```diff
@@ -0,0 +1,21 @@
...
+
+class ArrowExprStructNamespace:
+    def __init__(self: Self, expr: ArrowExpr) -> None:
+        self._compliant_expr = expr
+
+    def field(self: Self, name: str) -> ArrowExpr:
+        self._compliant_expr._evaluate_output_names = lambda _col: [name]
+        return reuse_series_namespace_implementation(
+            self._compliant_expr, "struct", "field", name=name
+        )
```
dangotbanned, line 21: I feel like mutating the current expression might cause issues elsewhere.

@osoucy was there a particular reason you chose this?

I can't seem to find any other cases like this (outside of the new `expr_struct` code)

osoucy: I had to use it becase "facilitator" methods such as `reuse_series_namespace_implementation` or `_compliant_expr._from_call` does not allow to control the output expression / series name. I could not find other examples where we explicitly had to rename the output. Not having this resulted in raising errors because of a mismatch between the expected expresssion name and the received expression name (or something similar).

The other option would be to not use facilitator methods and simply intantiate an expression / series class from scratch, but it would mean some code duplication.

dangotbanned: Thanks for the detail @osoucy that was helpful

I think that the pattern used here might be something to work backwards from
https://github.com/narwhals-dev/narwhals/blob/b986bfd5768759c2125c12ffdcd34911caf9e34c/narwhals/_pandas_like/expr_name.py

It seems quite a bit more complex than what you might need, but hopefully you can see how it avoids mutating the current `Expr`

https://github.com/narwhals-dev/narwhals/blob/b986bfd5768759c2125c12ffdcd34911caf9e34c/narwhals/_pandas_like/expr_name.py#L63-L84

@FBruzzesi per #1876

narwhals/_polars/expr.py
EdAbati: Thanks! the idea should be that given we are using the Polars interface we don't need to reimplement most of the methods for this backend :)

docs/api-completeness/series_struct.md
FBruzzesi: This is still relevant

narwhals/_polars/expr.py
osoucy: Ah... got it! So it would replace the `def field` implementation. Make sense! I made the change :)

osoucy:
Are we good to merge? Anything else to add?

docs/api-completeness/expr_struct.md
```diff
@@ -0,0 +1,5 @@
+# Expr.struct
+
+| Method       | arrow              | duckdb             | pandas-like        | polars             | spark-like         |
+|--------------|--------------------|--------------------|--------------------|--------------------|--------------------|
+| field        | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
```
MarcoGorelli, line 5: these files get autogenerated, can you show how they work for `expr_dt` / `expr_str` please?

MarcoGorelli (commented):
thanks @osoucy !

i'll check the `evaluate_output_names` / `alias_output_names` more closely, I'm not 100% sure about those

docs/api-completeness/expr_struct.md
osoucy: Oh sorry! I did not realized they were automatically generated. I will remove them. However, I'm not sure what you mean by "can you show how they work for expr_dt / expr_str please?".

MarcoGorelli: sorry I meant "check", not "show"

osoucy: Got it! Fixed!

narwhals/_arrow/expr_struct.py
```diff
@@ -0,0 +1,24 @@
...
+    def __init__(self: Self, expr: ArrowExpr) -> None:
+        self._compliant_expr = expr
+
+    def field(self: Self, name: str) -> ArrowExpr:
+        return reuse_series_namespace_implementation(
+            self._compliant_expr,
+            "struct",
+            "field",
+            name=name,
+            evaluate_output_names=lambda _col: [name],
```
MarcoGorelli, line 23: For all implementations, instead of overwriting `evaluate output names`,  can we just use `.alias`?

A good test would be `nw.col('a').struct.field('b').name.keep()`. I think for polars the resulting column name would be 'a' we should check that we do the same

narwhals/_arrow/series_struct.py
```diff
@@ -17,5 +17,5 @@ def __init__(self: Self, series: ArrowSeries) -> None:
     def field(self: Self, name: str) -> ArrowSeries:
         self._compliant_series._name = name
```
MarcoGorelli, line 18: you can't mutate `self._compliant_series`, you'll need `alias` here too

osoucy: Done.

dangotbanned: For the `Arrow*` stuff I'd suggest using `.compliant` and `.native`

- #2130

They weren't available when you started the PR @osoucy

dangotbanned: Some more context https://github.com/narwhals-dev/narwhals/pull/2130#discussion_r1976720783

MarcoGorelli: this is fine as a follow-up

osoucy: At least the current implementation

```
    def field(self: Self, name: str) -> ArrowSeries:
        return self._compliant_series._from_native_series(
            pc.struct_field(self._compliant_series.alias(name)._native_series, name),
        )
```
avoids mutating the compliant series. Maybe I can look into @dangotbanned in a future PR?

dangotbanned: All good @osoucy 😊

osoucy: So we are good to merge? :)

MarcoGorelli:
there's some unnecessary `alias` calls - I tried making a commit and pushing to your branch, but was denied access

could you check that the "allow edits by maintainers" checkbox is checked? you should see something like this

image

osoucy:
> there's some unnecessary `alias` calls - I tried making a commit and pushing to your branch, but was denied access
> 
> could you check that the "allow edits by maintainers" checkbox is checked? you should see something like this
> 
> image

This option is not available because the fork is created inside of my organization instead of inside my personnal account. I added you as a contributor to the organization. You should be able to push to the branch now.

narwhals/_pandas_like/series_list.py
```diff
@@ -16,6 +16,9 @@
 
 class PandasLikeSeriesListNamespace:
     def __init__(self: Self, series: PandasLikeSeries) -> None:
+        if not hasattr(series._native_series, "list"):
+            msg = "Series must be of PyArrow List type to support struct namespace."
```
MarcoGorelli, line 20: typo

narwhals/_expression_parsing.py
```diff
@@ -217,9 +217,12 @@ def reuse_series_namespace_implementation(
         expr: expression object.
         series_namespace: The Series namespace (e.g. `dt`, `cat`, `str`, `list`, `name`)
         attr: name of method.
+        evaluate_output_names: Output names function.
+        alias_output_names: Alias output names function.
```
MarcoGorelli, line 221: outdated

MarcoGorelli (approved):
thanks @osoucy ! great feature 🙌

osoucy:
Awesome! Thanks for the guidance and review!

PR #2132, opened 2025-03-02
feat: support `cum_sum` for lazy backends

MarcoGorelli:
Demo of this work:

For eager backends, just like now, you can keep using `cum_sum` liberally, there are no new restrictions introduced by this PR:
```python
In [16]: df
Out[16]: 
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|     a  b  c  i   |
|  0  a  1  5  0   |
|  1  a  2  4  1   |
|  2  b  3  3  2   |
|  3  b  5  2  4   |
|  4  b  3  1  3   |
└──────────────────┘

In [17]: df.with_columns(b_cum_sum = (nw.col('b')**2).cum_sum())
Out[17]: 
┌────────────────────────┐
|   Narwhals DataFrame   |
|------------------------|
|   a  b  c  i  b_cum_sum|
|0  a  1  5  0          1|
|1  a  2  4  1          5|
|2  b  3  3  2         14|
|3  b  5  2  4         39|
|4  b  3  1  3         48|
└────────────────────────┘
```

You can also optionally specify `_order_by` (for now private, whilst we build out the functionality):
```python
In [18]: df.with_columns(b_cum_sum = (nw.col('b')**2).cum_sum().over(_order_by='i'))
Out[18]: 
┌────────────────────────┐
|   Narwhals DataFrame   |
|------------------------|
|   a  b  c  i  b_cum_sum|
|0  a  1  5  0          1|
|1  a  2  4  1          5|
|2  b  3  3  2         14|
|3  b  5  2  4         48|
|4  b  3  1  3         23|
└────────────────────────┘
```
You can also partition by a column, but for pandas there's the usual limitation that only elementary expressions are supported:
```python
In [19]: df.with_columns(b_cum_sum = nw.col('b').cum_sum().over('a', _order_by='i'))
Out[19]: 
┌────────────────────────┐
|   Narwhals DataFrame   |
|------------------------|
|   a  b  c  i  b_cum_sum|
|0  a  1  5  0          1|
|1  a  2  4  1          3|
|2  b  3  3  2          3|
|3  b  5  2  4         11|
|4  b  3  1  3          6|
└────────────────────────┘
```
And now, for the new functionality which this unlocks. Here is where we start enabling new things. For lazy backends, the above is supported, but specifying `_order_by` is required. Example using SQLFrame:
```python
In [29]: lf
Out[29]: 
┌────────────────────────────────────────────────────────────────────┐
|                         Narwhals LazyFrame                         |
[...]

PR #1826, opened 2025-01-19
feat: add pyspark str namespace `to_datetime`

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [ ] Documented the changes

PR #2152, opened 2025-03-05
chore(typing): `SparkLikeExpr` properties

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Porting over (#2051), didn't realize this was declared twice until (#2132)


dangotbanned:
@MarcoGorelli if you merge this ahead of #2132 - you'll be able to get more useful IDE feedback from `SparkLikeExpr.(_F|_Window)` 😊

narwhals/_spark_like/expr_str.py
```diff
@@ -107,47 +108,28 @@ def to_lowercase(self: Self) -> SparkLikeExpr:
...
 
 
-@overload
-def strptime_to_pyspark_format(format: None) -> None: ...
-
-
-@overload
-def strptime_to_pyspark_format(format: str) -> str: ...
+def is_naive(obj: Any) -> bool:
+    return obj is not None and {"%s", "%z", "Z"}.isdisjoint(obj)
```
EdAbati, line 127:
```suggestion
def is_naive(format: str) -> bool:
    return {"%s", "%z", "Z"}.isdisjoint(format)
```

just to signal that this checks the date time format (should we maybe call it `is_naive_format`?). 
Also based on how it is used, `format` will always be a non-null str right?

dangotbanned: Thanks @EdAbati, yeah `is_naive_format` sounds good to me

I renamed `format` to `obj` to avoid a ruff warning for shadowing `builtins.format`

Also narrowing from `Any` to `str` sounds good as well.
I originally wrote it using a `TypeIs` but removed that since it gave a false-positive on unreachability

dangotbanned: Happy for you to make those changes and merge 🙌

dangotbanned: Actually I think the current `set` needs to be changed to:

```py
{"s", "z", "Z"}
```

I'm relying on `format` splitting into a sequence of characters - but two of them a 2-char sequences.

I wonder why that didn't fail a test?

narwhals/_spark_like/expr.py
```diff
@@ -78,7 +79,11 @@ def func(df: SparkLikeLazyFrame) -> Sequence[Column]:
         )
 
     @property
-    def _F(self: Self) -> Any:  # noqa: N802
+    def _F(self: Self):  # type: ignore[no-untyped-def] # noqa: ANN202, N802
+        if TYPE_CHECKING:
+            from pyspark.sql import functions
+
+            return functions
```
EdAbati, line 86: 🤯

dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1970400205

FBruzzesi: This is a deja vu 😂

narwhals/_spark_like/expr_str.py
EdAbati: > I wonder why that didn't fail a test?

I actually haven't seen a test covering this 👀 double checking with the datetime wizard @MarcoGorelli , is it something that we are not testing on purpose?

dangotbanned: > > I wonder why that didn't fail a test?
> 
> I actually haven't seen a test covering this 👀 double checking with the datetime wizard @MarcoGorelli , is it something that we are not testing on purpose?

@EdAbati 😮 

IMO there's more than enough `narwhals`-own logic here to warrant testing

dangotbanned: Probably related to
- #1818

dangotbanned: Maybe I'm reading this wrong, but I dont think we have coverage for **any backend** with a tz-aware format for `.str.to_datetime(...)`

https://github.com/narwhals-dev/narwhals/blob/3103b20afc1a523137f74003478c77a9debc2781/tests/expr_and_series/str/to_datetime_test.py

dangotbanned: @MarcoGorelli really need some help on writing a test for this 🙏 

Right now I've got something *kinda* working for a lot of backends.
But it fails to parse for `sqlframe`, `pyarrow`.

Also [`polars.Expr.str.to_datetime`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.to_datetime.html) seems to have more options - that might be able to preserve the timezone offset?

### `test_to_datetime_tz_aware`

```py
@pytest.mark.parametrize(
    ("fmt", "data", "expected_pandas", "expected_polars_duckdb"),
    [
        (
            "%Y-%m-%d %H:%M:%S%z",
            {"a": ["2020-01-01 12:34:56+0200"]},
            "2020-01-01 12:34:56+02:00",
            "2020-01-01 10:34:56+00:00",
        )
    ],
)
def test_to_datetime_tz_aware(
    constructor: Constructor,
    fmt: str,
    data: dict[str, Sequence[str]],
    expected_pandas: str,
    expected_polars_duckdb: str,
) -> None:
    constructor_str = str(constructor)
    if "pandas" in constructor_str or "dask" in constructor_str:
        expected = expected_pandas
    elif "polars" in constructor_str or "duckdb" in constructor_str:
        expected = expected_polars_duckdb
    else:
        expected = "fails"
    result = (
        nw.from_native(constructor(data))
        .lazy()
        .select(b=nw.col("a").str.to_datetime(fmt))
        .collect()
        .item(row=0, column="b")
    )

    assert str(result) == expected
```

I can fully understand nope-ing out of handling timezones for all these different packages 😒

dangotbanned: Alright so (https://github.com/narwhals-dev/narwhals/pull/2152/commits/e19dcdf5f747a350c69b2a06236bbdcd886d6b27) -> (https://github.com/narwhals-dev/narwhals/pull/2152/commits/194ef83ca93d9c91c9efd3626251643e42f458bd) now has a single test

`sqlframe` works, `pyspark` apprently doesn't https://github.com/narwhals-dev/narwhals/actions/runs/13700361438/job/38312364899?pr=2152

Absolutely don't want to do timezones again 😩

dangotbanned: @FBruzzesi I see you added `strp_to_pyspark_format` in #1826

AFAICT there were never any tests for tz-aware timestamps and right now I can't get one to work 😞 

Tried my best, but as of (https://github.com/narwhals-dev/narwhals/pull/2152/commits/ce3145db05be055d3387260351d12538b5b3dc49) I'm giving up

MarcoGorelli (commented):
thanks for looking into this!

I think this PR is doing too much, please either address typing or fix `str.to_datetime`

the former should be fine, some comments for the latter:
- can we keep `format` even if it overrides the builtin?
- we should probably make sure that we consistently parse into UTC, so there should only be a single `expected` for `test_to_datetime_tz_aware`

I can take care of the timezone handling separately, but it does need splitting out

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2152#pullrequestreview-2667604929

Thanks for the review @MarcoGorelli
I think you're right, I bit off a lil more than I could chew here.

Caught me at a good time with the feedback - hopefully I've addressed it all in:
- (https://github.com/narwhals-dev/narwhals/pull/2152/commits/4470a99d18d244962559c210fbcf8f26ca34504d)
- (https://github.com/narwhals-dev/narwhals/pull/2152/commits/77a264ce2dd27fcd037f2ba56d279b27b30bff09)

narwhals/_spark_like/expr.py
```diff
@@ -91,7 +96,11 @@ def _F(self: Self) -> Any:  # noqa: N802
         return functions
 
     @property
-    def _native_dtypes(self: Self) -> Any:
+    def _native_dtypes(self: Self):  # type: ignore[no-untyped-def] # noqa: ANN202
```
MarcoGorelli, line 99: why remove `-> Any`?

MarcoGorelli: trying this locally, ` -> type[functions]` seems to work fine? it still requires a `type: ignore` internally, but so does the approach in this PR anyway

dangotbanned: This isn't a valid type?

```py
# where `functions` is a module
type[functions]
```

Using `# type: ignore[no-untyped-def]` is explained pretty visually here https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1969753588

Shortest answer - `Any` prevents the inference from working.

If you're getting `type[functions]` to work locally - I can only assume its because it gets ignored as an invalid type maybe?

MarcoGorelli: thanks, i'll check

narwhals/_spark_like/expr_str.py
```diff
@@ -107,47 +107,29 @@ def to_lowercase(self: Self) -> SparkLikeExpr:
...
 
 
-@overload
-def strptime_to_pyspark_format(format: None) -> None: ...
-
-
-@overload
-def strptime_to_pyspark_format(format: str) -> str: ...
+def is_naive_format(format: str) -> bool:  # noqa: A002
+    return {"s", "z", "Z"}.isdisjoint(format)
```
MarcoGorelli, line 127: i don't really like how we're changing the logic here, `'z'` in a format string doesn't work by itself, it's `'%z'` we need to check for

the rest looks good though, well done!

PR to address time zone aware parsing: https://github.com/narwhals-dev/narwhals/pull/2166

dangotbanned: Thanks @MarcoGorelli 

I'm happy for that bit to be reverted and just keep the splitting out into the new function?

dangotbanned: This kind of `any(x in (...))` check would probably be good

https://github.com/MarcoGorelli/narwhals/blob/fdb2f4812945e9fab459b6bc473e095b18fe61a7/narwhals/_pandas_like/series_str.py#L88

MarcoGorelli: sure! if you're happy with https://github.com/narwhals-dev/narwhals/pull/2166 to add a test and adjust some logic, happy to then take this?

narwhals/_spark_like/expr.py
MarcoGorelli: sure let's keep it like this, ignore me here, sorry

dangotbanned: No worries

narwhals/_spark_like/expr_str.py
dangotbanned: Hopefully got it done (https://github.com/narwhals-dev/narwhals/pull/2152/commits/311c7ea4d556e7530db45c30e6d25a2d3dfdf5d5)

MarcoGorelli (approved):
thanks @dangotbanned much appreciated!

PR #2149, opened 2025-03-05
refactor: adds `_compliant` sub-package

dangotbanned:
Closes #2044

## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes: https://github.com/narwhals-dev/narwhals/issues/2044#issuecomment-2674262833
- https://github.com/narwhals-dev/narwhals/pull/2119#issuecomment-2700860171
- Supersedes (#2055)
- #2130
- #2064

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Direct follow-up to (#2119), aiming to **make space** for similar changes for other `Compliant*` protocols.

This PR spiralled a bit, but tackles a few quite-related issues.

### Lazy-only typing
#2044 has been resolved through the use of the new protocol `NativeExpr`.

https://github.com/narwhals-dev/narwhals/blob/b3bdb3b3ba52dada9cadde3c50d68520f1270a09/narwhals/_compliant/expr.py#L65-L73

### `(Arrow|PandasLike)*` deduplication
All of the new classes prefixed with `Eager` now provide the common functionality between the two  **Eager-only** backends.
The big one is `EagerExpr`

https://github.com/narwhals-dev/narwhals/blob/b3bdb3b3ba52dada9cadde3c50d68520f1270a09/narwhals/_compliant/expr.py#L257-L260

A **very unexpected** part of that was that they could also share `*ExprNamespace` implementations

- (https://github.com/narwhals-dev/narwhals/pull/2149/commits/5a13b63a12f211bbc58977b017bbf3783d61cf54)
- (https://github.com/narwhals-dev/narwhals/pull/2149/commits/277782219ea18a22fd5387ac03f9c6f658f1de3d)

Probably the most visual part of the PR as it removes the need for **12 modules**

### `*Namespace` protocols

Working on the `Eager` namespaces led to the new generic protocols for both `(expr|series)_*` namespaces https://github.com/narwhals-dev/narwhals/blob/b3bdb3b3ba52dada9cadde3c50d68520f1270a09/narwhals/_compliant/any_namespace.py

These can be re-used for all backends 😄 

### Misc
[...]

Issue #1485, opened 2024-12-02
clean: remove unused code in _pandas_like/series.py


MarcoGorelli:
This code is now unused:
```python
        # In pandas, copy-on-write becomes the default in version 3.
        # So, before that, we need to explicitly avoid unnecessary
        # copies by using `copy=False` sometimes.
        if self._implementation is Implementation.PANDAS and self._backend_version < (
            3,
            0,
            0,
        ):
            self._use_copy_false = True
        else:
            self._use_copy_false = False
```
anyone fancy cleaning it up?

MarcoGorelli:
closed by https://github.com/narwhals-dev/narwhals/pull/1489

PR #2169, opened 2025-03-08
refactor: try `pd.Series` without `copy=False`

dangotbanned:
> [!WARNING]
> Just testing CI 

- Experimenting with an idea from (#2149)
- Trying to understand the different behavior from `create_compliant_series`


dangotbanned:
@MarcoGorelli I've traced this back all the way to https://github.com/narwhals-dev/narwhals/commit/b2b03c8c932166d5ca2b5293003997bab605cddb#diff-a35b1d0015dbbb57ecbf08e3a5b51cca3960c157491746f8f38b1867e191dbf5

Would you be able to help me understand this?

AFAICT, the default is and has always been `False` 🤔

And there are various conditions in other functions that determine when/when not to do this?
https://github.com/narwhals-dev/narwhals/blob/944f6da692160b93ae4c1196855e51ef0243a3e8/narwhals/_pandas_like/utils.py

## `pandas==main`
https://github.com/pandas-dev/pandas/blob/f1b00b8e373aad54c83585b94076f53ae8e89da9/pandas/core/series.py#L266

## `pandas==3.0.0.dev0`
https://github.com/pandas-dev/pandas/blob/dc37a6d035474c25932d805cfe50a47b04916a83/pandas/core/series.py#L286

## `pandas==2.0.0`
https://github.com/pandas-dev/pandas/blob/bdb509f95a8c0ff16530cedb01c2efc822c0d314/pandas/core/series.py#L275

## `pandas==1.0.0`
https://github.com/pandas-dev/pandas/blob/b687cd4d9e520666a956a60849568a98dd00c672/pandas/core/series.py#L157

## `pandas==0.25.0`
https://github.com/pandas-dev/pandas/blob/13653812b43aab0d4be29177f9c849c07414a7bc/pandas/core/series.py#L174

## `pandas==0.19.0`
https://github.com/pandas-dev/pandas/blob/12b3264d17780b6b38f747353bc80cfd910e6dc8/pandas/core/series.py#L128

dangotbanned:
My hope is to do something like (https://github.com/narwhals-dev/narwhals/pull/2149/commits/612df692b356f602c44738362174c2d3df1163c5) - but for `PandasLike`

dangotbanned:
The only thing I've come across was the dev release notes

https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#copy-keyword

They mention a lot of places `copy` will be removed - but none of them are in `pd.Series(...)`

dangotbanned:
# Summary

@MarcoGorelli apologies for the earlier ping before I did my homework!

AFAICT, `copy=False` has been used in `narwhals` for performance benefits:

- #1547
- #1485
- #1252
- #1247
- #937
- #275

## Copy-on-Write

I'm surprised that enabling [Copy-on-Write](https://pandas.pydata.org/docs/user_guide/copy_on_write.html) either globally (risky?) or via a [context manager](https://pandas.pydata.org/docs/reference/api/pandas.option_context.html#pandas.option_context) (preferred) hasn't been discussed before.

Looking from the outside (non-`pandas` user) - that seems like the safer route to get to the same goal of reducing overhead IMO.


## Original goal

Wasn't expecting that pulling at this thread would lead here 😅 

All I wanted to know was **why do these two functions co-exist and behave differently**:

https://github.com/narwhals-dev/narwhals/blob/1e45855366fd2a0bb36f17c08e12be89ebc2a125/narwhals/_pandas_like/utils.py#L150-L163

https://github.com/narwhals-dev/narwhals/blob/1e45855366fd2a0bb36f17c08e12be89ebc2a125/narwhals/_pandas_like/utils.py#L264-L277

MarcoGorelli:
copy on write as only been optionally available since pandas 2 (i think) and even then it took some versions to iron out, i don't think we can enable it on behalf of the user

it may well be that `_create_compliant_series` overlaps with `_from_iterable`, it looks like I introduced this when Narwhals was about 1 week old and still just a little experiment

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2169#issuecomment-2714639155

Thanks for getting back to me @MarcoGorelli 

> copy on write as only been optionally available since pandas 2 (i think) and even then it took some versions to iron out,

Yeah I was hoping someone else might be able to interpret how this might line up with the current `copy=False` usage:
https://pandas.pydata.org/docs/user_guide/copy_on_write.html#copy-on-write-cow
> Copy-on-Write was first introduced in version **1.5.0**.
> Starting from version **2.0** most of the optimizations that become possible through CoW are implemented and supported.
> All possible optimizations are supported starting from pandas **2.1**.
> 
> CoW will be enabled by default in version **3.0**.

Would've been nice if they had a table showing which version enabled CoW for specfic features

> i don't think we can enable it on behalf of the user

Totally understandable! 
I'm out-of-my-depth with `pandas` so I trust you on this 🙏 


> it may well be that `_create_compliant_series` overlaps with `_from_iterable`

I think the branch `copy` is [first used in](https://github.com/pandas-dev/pandas/blob/f1b00b8e373aad54c83585b94076f53ae8e89da9/pandas/core/series.py#L396-L401) seems to handle the `np.ndarray` vs other `Iterable`(s) case for us right?


> it looks like I introduced this when Narwhals was about 1 week old and still just a little experiment

👶 `narwhals`

dangotbanned:
Closing as not planned

PR #2178, opened 2025-03-09
feat: support passing multiple arguments positionally to get_native_namespace

### LINKED ISSUE #2160: [Enh]: `get_native_namespace` could check several objects
When figuring out how to do support for narwhals in fairlearn, I was concerned with how to make sure that a user passes objects from the same dataframe library, which would then determine the namespace of returned objects.

I have seen there is `nw.get_native_namespace()` and I would suggest an enhancement for it similar to something we have in scikit-learn: There is a [get_namespace()](https://github.com/scikit-learn/scikit-learn/blob/5f50a8706f46e971aa053791a1441ac8a4ebd623/sklearn/utils/_array_api.py#L497) function for Array API support and you can pass multiple objects to check. 

`get_native_namespace` could also error or return False if the objects don't share a common namespace. 

I would find it practical for the current use case in fairlearn and I would imagine other projects would also benefit from it. Would it make sense to add a functionality like this one?

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2160

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -40,11 +40,6 @@ from narwhals.dependencies import is_sqlframe_dataframe
 from narwhals.utils import Version
 
 if TYPE_CHECKING:
-    import pandas as pd
-    import polars as pl
-    import pyarrow as pa
-
-    from narwhals._arrow.typing import ArrowChunkedArray
     from narwhals.dataframe import DataFrame
     from narwhals.dataframe import LazyFrame
     from narwhals.series import Series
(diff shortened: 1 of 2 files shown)


EdAbati:
Nice!👌 thanks

EdAbati (approved):

EdAbati:
(FYI I edited the description so it closes the issue automatically)

MarcoGorelli:
thanks for your review!

Issue #1541, opened 2024-12-08
enh: let `Enum` take arguments, allow it in construction


MarcoGorelli:
We should be able to do
```
s = nw.new_series('foo', ['a', 'b', 'c'], dtype=nw.Enum(['a', 'b', 'c', ''d]), native_namespace=pd)
```
and have it create a Series with dtype Categorical and categories 'a', 'b', 'c', 'd'

camriddell:
@MarcoGorelli I can take this one on

MarcoGorelli:
awesome! please do let us know if anything trips you up

camriddell:
@MarcoGorelli let me know if the following question is a bit out of scope and I should focus only on adding some extra code to patch `narwhals/functions.py:_new_series_impl` given the native dtype, instead of the broader handling of Categoricals/Enums
internally.

---

In the following code snippet what should the expected dtype of column `"a"` be?

```python
 dfpd = pd.DataFrame({"a": ["a", "b"]}, dtype="category")
 df = nw.from_native(dfpd).select(nw.col("a").cast(nw.Enum(["a", "b"])))
```

The ambiguous portion is that internally `pd.Category` maps to `nw.Category`, then (to my understanding) the conversion is applied internally such that when we specify casting from `nw.Category` → `nw.Enum(...)`, internally pandas would move from `pd.CategoricalDtype` to another `pd.CategoricalDtype`, but then narwhals reinterprets the new `pd.CategoricalDtype` to become a `nw.Category`.

Given the above consideration, would it be best to:

a. Converting to an `nw.Enum` from pandas native space is not allowed and will raise an error (current behavior). However we can accommodate a construction with `nw.new_series` via a conditional branch.
b. Converting to an `nw.Enum` from pandas native space creates `pd.CategoricalDtype(..., ordered=True)`, then narwhals uses that `ordered=True` flag to signify whether a `pd.CategoricalDtype` should result in `nw.Categroical` or `nw.Enum`. The ordered flag here actually works quite nicely as `polars.Enum` respects the ordering of the values inputted.
c. The code snippet above returns a schema of `{'a': nw.Categorical}` which may surprise end users since there was an explicit request to convert to `nw.Enum(...)`.

I have a working codebase for option B and can open a PR if you want to take a look for some more fine-grained discussion.

MarcoGorelli:
thanks for looking at this so closely - i'm gonna have to think about this one, not 100% sure atm 😄

[...]

PR #2192, opened 2025-03-11
feat: let `nw.Enum` accept categories, map pandas ordered categorical to Enum (only in main namespace, not stable.v1)

camriddell:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1541

## Checklist

- [x] Tests added
- [ ] Documented the changes

Adds support for the `nw.Enum` datatype for pandas (backed by `pandas.CategoricalDtype(…, ordered=True)`

The current implementation diverges from pandas/Polars in two broad ways
1. We do not check for None, NaN, or Null (both pandas and Polars raise when they construct a CategoricalDtype/Enum with these in the categories list.
2. pandas allows arbitrary (hashable) objects to be stored as the categories, whereas Polars only allows integers. The current implementation is type-hinted to follow suit with pandas, but we do not perform this check instead letting the backend library raise as needed.

```python
>>> import narwhals as nw
>>> import pandas as pd
>>> s = nw.new_series('foo', ['a', 'b', 'c'], dtype=nw.Enum(['a', 'b', 'c', 'd']), native_namespace=pd)
>>> s
┌───────────────────────────────────────────────┐
|                Narwhals Series                |
|-----------------------------------------------|
|0    a                                         |
|1    b                                         |
|2    c                                         |
|Name: foo, dtype: category                     |
|Categories (4, object): ['a' < 'b' < 'c' < 'd']|
└───────────────────────────────────────────────┘
```


MarcoGorelli:
thanks! it's encouraging that this doesn't break downstream tests

sorry i didn't get round to it for tomorrow's release, will try to get it in for next week's one 👍

narwhals/_pandas_like/utils.py
```diff
@@ -569,8 +575,17 @@ def narwhals_to_native_dtype(  # noqa: PLR0915
...
     if isinstance_or_issubclass(dtype, dtypes.Enum):
-        msg = "Converting to Enum is not (yet) supported"
-        raise NotImplementedError(msg)
+        if isinstance(dtype, dtypes.Enum):
+            try:
+                import pandas as pd  # ignore-banned-import
+            except ImportError as exc:  # pragma: no cover
+                msg = f"Unable to convert to {dtype} to to the following exception: {exc.msg}"
+                raise ImportError(msg) from exc
+            return pd.CategoricalDtype(categories=dtype.categories, ordered=True)
```
MarcoGorelli, line 584: not sure if we can do something pandas-specific here, as this is used by cudf and modin too - could we generalise?

narwhals/_pandas_like/utils.py
```diff
@@ -352,7 +354,11 @@ def non_object_native_to_narwhals_dtype(dtype: str, version: Version) -> DType:
         return dtypes.String()
     if dtype in {"bool", "boolean", "boolean[pyarrow]", "bool[pyarrow]"}:
         return dtypes.Boolean()
-    if dtype == "category" or dtype.startswith("dictionary<"):
+    if dtype.startswith("dictionary<"):
+        return dtypes.Categorical()
+    if dtype == "category":
+        if native_dtype.ordered:
+            return dtypes.Enum(categories=native_dtype.categories)
         return dtypes.Categorical()
```
MarcoGorelli, line 362: this would be a breaking change, so i'm not totally sure about it - could we preserve the current behaviour in `v1` and only make this change in the main namespace? the `version` variable is available in this function, you can use that

narwhals/_pandas_like/utils.py
dangotbanned: I'm a little confused by this. 
`pandas` is already a a module-level import?

https://github.com/narwhals-dev/narwhals/blob/5550ad897992e3a639cea3fb0cca5bede98da6d7/narwhals/_pandas_like/utils.py#L13

camriddell: @dangotbanned you're right- I pulled this code from a pretty old branch I had so that must have just been leftover. I'll delete it.

@MarcoGorelli I'll look into generalizing `cudf` and `modin`

MarcoGorelli:
thanks Cam - looks like there's a xpass
```
FAILED tests/series_only/cast_test.py::test_cast_to_enum_v1[modin[pyarrow]]
```

narwhals/stable/v1/_dtypes.py
```diff
@@ -72,6 +72,36 @@ def __hash__(self: Self) -> int:
         return hash(self.__class__)
 
 
+class Enum(NwEnum):
+    """A fixed categorical encoding of a unique set of strings.
```
camriddell, line 76: @dangotbanned the typing here gets a bit wonky as we currently need `v1._dtypes...` implementations to inherit from what is defined in `nw.dtypes`. However `nw.dtypes.Enum` had its call signature changed which should not be propagated down to `v1._dtypes.Enum` so I implemented this functionality to skip a level of inheritance on its defined methods.

If feels like I may have something backwards here though? Would love to hear you thoughts.

dangotbanned: Thanks for the ping @camriddell, will take a look in the morning

dangotbanned: Definitely a tricky one, but I have a few ideas I'm gonna try out today.

I did a [search of existing usage](https://github.com/search?q=nw.Enum+language%3APython&type=code&l=Python&p=1) and looked at [what we allow in tests](https://github.com/narwhals-dev/narwhals/blob/b2046697f01ebe1cdf88e63ca194f6da5752f740/tests/series_only/cast_test.py#L116-L146).

I think our main concern should be **preserving the behavior of** `isinstance(..., nw.Enum)`

dangotbanned: I'm wanting to avoid subclassing, since this is a pretty clear [Liskov substitution principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle) violation (not your fault, just how `v1` inheriting from `main` works)

MarcoGorelli: i think it's ok to allow `Enum` to accept `categories` in v1 as well, so long as `== nw.Enum` keeps working - can you check what we do for `Datetime` and `Duration`? I think something similar might work?

dangotbanned: > i think it's ok to allow `Enum` to accept `categories` in v1 as well, so long as `== nw.Enum` keeps working - can you check what we do for `Datetime` and `Duration`? I think something similar might work?

Lol @MarcoGorelli the timing on this 😅 (https://github.com/narwhals-dev/narwhals/pull/2192/commits/105e39453b7a0c25bfb4c36f72a453d7399a4ad2)

tests/dtypes_test.py
```diff
@@ -393,3 +397,25 @@ def test_cast_decimal_to_native() -> None:
...
+
+
+def test_enum_v1_is_enum_unstable() -> None:
+    enum_v1 = nw.Enum()
+    enum_unstable = unstable_nw.Enum(("a", "b", "c"))
+    assert isinstance(enum_v1, unstable_nw.Enum)
+    assert issubclass(nw.Enum, unstable_nw.Enum)
+
+    assert enum_v1 == unstable_nw.Enum
+    assert enum_v1 != enum_unstable
```
dangotbanned, line 421: I've got another case locally that I think is a bug, since it fails

```py
assert enum_unstable != nw.Enum
```

```py
E AssertionError: assert Enum(categories=['a', 'b', 'c']) != <class 'narwhals.stable.v1._dtypes.Enum'>
E +  where <class 'narwhals.stable.v1._dtypes.Enum'> = nw.Enum

tests\dtypes_test.py:422: AssertionError
```

narwhals/dtypes.py
```diff
@@ -443,9 +446,30 @@ class Enum(DType):
...
+        #       should the intersection of the above be caught at the narwhals layer?
+        if isinstance(categories, type) and issubclass(categories, enum.Enum):
+            categories = (getattr(v, "value", v) for v in categories.__members__.values())
+        self.categories = [*categories]
+
+    def __eq__(self: Self, other: object) -> bool:
+        # allow comparing object instances to class
+        if type(other) is type:
+            return other is Enum
+        return isinstance(other, type(self)) and self.categories == other.categories
```
dangotbanned, line 464: @camriddell you're using a `list` to describe:

> A **fixed** categorical encoding of a **unique** set of strings.

`pl.Enum`, for reference, uses a `pl.Series` and makes quite a lot of checks to ensure they can keep that promise

https://github.com/pola-rs/polars/blob/py-1.26.0/py-polars/polars/datatypes/classes.py#L645-L727

This all type checks fine, and passes through at runtime - despite being the wrong types and non-unique:

```py
import narwhals as nw

very_invalid = nw.Enum(
    ["beluga", "narwhal", "orca", (), nw.Enum([]), "narwhal", ((),), "narwhal", "orca"]
)
>>> very_invalid
Enum(categories=['beluga', 'narwhal', 'orca', (), Enum(categories=[]), 'narwhal', ((),), 'narwhal', 'orca'])
```

Since a `list` was used, I can break the *"fixed"* promise with:

```py
very_invalid.categories.insert(0, [9, 2, 3, 5])
>>> very_invalid
Enum(categories=[[9, 2, 3, 5], 'beluga', 'narwhal', 'orca', (), Enum(categories=[]), 'narwhal', ((),), 'narwhal', 'orca'])
```

Or go further with:

```py
very_invalid.categories.sort(key=lambda x: not isinstance(x,str))
>>> very_invalid
Enum(categories=['beluga', 'narwhal', 'orca', 'narwhal', 'narwhal', 'orca', [9, 2, 3, 5], (), Enum(categories=[]), ((),)])
```

dangotbanned: I think we might need to restrict inputs to a `nw.Series` - since we'd need access to a backend to make the same checks.

Possibly raise on `Enum.__setattr__` as well 🤔

camriddell: 1. Agreed on performing uniqueness checks during instantiation like Polars.
2. I'm agnostic on enforcing that inputs should be only strings, apologies for  (while this is core to Polars, pandas differs here by allowing any hashable object).
    - Originally, my thought was to let the backends figure out what they can/can't handle when we convert to them during runtime. But this would mean that code that works for one backend simply wouldn't work for another, so perhaps taking the strictest set of requirements is favorable?

I'm still a fan of using basic data structures here for agnostic data modeling as it is less complex than having a dtype that needs a backend for creation. For immutability, we can coerce to a tuple instead of a list at instantiation so that the above manipulations become disallowed.

As an aside, Polars doesn't make guarantees about immutability. I don't know if we should go out of our way to force users to not manipulate data in this manner either.

```python
>>> e = pl.Enum(['a', 'b', 'c'])
>>> e.categories
shape: (3,)
Series: 'category' [str]
[
        "a"
        "b"
        "c"
]
>>> e.categories[0] = 'd'
>>> e
Enum(categories=['d', 'b', 'c'])
```

narwhals/dtypes.py
```diff
@@ -443,9 +446,30 @@ class Enum(DType):
        >>> data = ["beluga", "narwhal", "orca"]
        >>> s_native = pl.Series(data, dtype=pl.Enum(data))
        >>> nw.from_native(s_native, series_only=True).dtype
-       Enum
+       Enum(categories=['beluga', 'narwhal', 'orca'])
     """
 
+    def __init__(self, categories: Iterable[Hashable] | type[enum.Enum]) -> None:
+        # TODO(Unassigned): pandas errors on NaN, NA, NaT OR duplicated value category
+        #       Polars errors on Null, NaN OR duplicated OR any non-string category
```
camriddell, line 454: @MarcoGorelli @dangotbanned any thoughts on whether we should perform a NaN value check at instantiation? Or should we let each backend handle their cases as to these kinds of checks.

Note that @dangotbanned and I are contemplating the duplicated/non-string categories in another thread.

narwhals/dtypes.py
dangotbanned: Thanks for following up on the mutability bit!

If we have a choice of only `list` or `tuple`, then I think choosing the later *at least* hints the intent 🙂

camriddell: I'll push an update factoring this in and we can see if it is satisfactory.

dangotbanned: I'd prefer going for the stricter requirements of `polars`.

It would avoid the concerns raised in (https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035810308).

You'd also be able to use `narwhals` to roundtrip `Enum` conversion between backends.
A different set of rules for `pandas` means that could only work in one direction

MarcoGorelli: > I'm agnostic on enforcing that inputs should be only strings

I don't think we should interfere with pandas here - if someone passes a pandas series which has non-string categories, then that's fine

narwhals/dtypes.py
MarcoGorelli: i'd suggest leaving all of these to the backends

narwhals/_dask/utils.py
```diff
@@ -125,6 +125,16 @@ def narwhals_to_native_dtype(dtype: DType | type[DType], version: Version) -> An
         return "object"  # pragma: no cover
     if isinstance_or_issubclass(dtype, dtypes.Boolean):
         return "bool"
+    if isinstance_or_issubclass(dtype, dtypes.Enum):
+        if version is Version.V1:
+            msg = "Converting to Enum is not supported in V1"
```
MarcoGorelli, line 130: "in `narwhals.stable.v1`"

narwhals/_polars/utils.py
```diff
@@ -111,7 +111,9 @@ def native_to_narwhals_dtype(
     if dtype == pl.Categorical:
         return dtypes.Categorical()
     if dtype == pl.Enum:
-        return dtypes.Enum()
+        if version is Version.V1:
+            return dtypes.Enum()  # type: ignore[call-arg]
+        return dtypes.Enum(dtype.categories)  # type: ignore[attr-defined]
```
MarcoGorelli, line 116: does it not work to copy what we do for `Datetime`? i think it should be fine to pass `dtype.categories` in all cases

tests/frame/schema_test.py
```diff
@@ -89,7 +90,8 @@ class Foo: ...
 
 
 @pytest.mark.skipif(PANDAS_VERSION < (2, 0, 0), reason="too old")
-def test_dtypes() -> None:
+@pytest.mark.parametrize("nw", [nw_main, nw])
+def test_dtypes(nw: ModuleType) -> None:
```
MarcoGorelli, line 94: i don't think we need to parametrise these, can we can do the v1-specific test in the new `v1_test.py`

MarcoGorelli (commented):
awesome, this looks very close

narwhals/dtypes.py
dangotbanned: > > I'm agnostic on enforcing that inputs should be only strings
> 
> I don't think we should interfere with pandas here - if someone passes a pandas series which has non-string categories, then that's fine

@MarcoGorelli We'd probably want to test for more than just this then right?
https://github.com/camriddell/narwhals/blob/02b60e04c3d0617ff850f4236d52aaae6fe8ac0d/tests/dtypes_test.py#L402-L404

```py
@pytest.mark.parametrize(
    "categories", [["a", "b"], ["a", None], [1, 2], enum.Enum("test", "a b")]
)
```

I imagine using `nw.Series` would work fine, but always good to cover that if its anticipated 🙂

narwhals/_polars/utils.py
dangotbanned: @MarcoGorelli that would be breaking though wouldn't it? 

https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2019784440

> this would be a breaking change, so i'm not totally sure about it - could we preserve the current behaviour in v1 and only make this change in the main namespace? the version variable is available in this function, you can use that

MarcoGorelli: not sure i see why it would be breaking, is there any code which someone could have written with stable.v1 which would stop working or work differently because of this change?

dangotbanned: If we make `v1.Enum` have a required argument, this would turn into a `TypeError`:

```py
from narwhals.stable import v1 as nw_v1

>>> nw_v1.Schema({"column 1": nw_v1.String(), "column_2": nw_v1.Enum()})
Schema([('column 1', String), ('column_2', Enum)])
```

MarcoGorelli: ah right, thanks - all good then 👍

dangotbanned: I will note though that I didn't find any usage like that from a search.

But I can see it existing since we advise it in the docs for `Schema`:

>  *instantiated*

https://github.com/narwhals-dev/narwhals/blob/52355ea5224bb30521f5935a54d80c43608ecefe/narwhals/stable/v1/__init__.py#L608-L615

narwhals/dtypes.py
MarcoGorelli: not sure we can really use `nw.Series` as lazy backends don't support it

not sure i see the issue with `tuple`?

narwhals/stable/v1/_dtypes.py
```diff
@@ -72,6 +72,38 @@ def __hash__(self: Self) -> int:
...
+       >>> import polars as pl
+       >>> import narwhals.stable.v1 as nw
+       >>> data = ["beluga", "narwhal", "orca"]
+       >>> s_native = pl.Series(data, dtype=pl.Enum(data))
+       >>> nw.from_native(s_native, series_only=True).dtype
+       Enum
+    """
+
+    def __new__(cls: type[Self]) -> Self:
+        return super(NwEnum, cls).__new__(cls)
```
MarcoGorelli, line 90: skill issue on my part, but why do we need `__new__` here? the tests seem to pass without it

narwhals/dtypes.py
dangotbanned: > not sure we can really use `nw.Series` as lazy backends don't support it


@MarcoGorelli For the test I'm talking about testing this, which all works and passes type checking:

```py
import polars as pl

import narwhals as nw

categories = pl.Series(["a", "b", "c"])
enum = pl.Enum(categories)
>>> enum
Enum(categories=['a', 'b', 'c'])

categories_nw = nw.from_native(categories, series_only=True)
enum_nw = nw.Enum(categories_nw)
>>> enum_nw
Enum(categories=('a', 'b', 'c'))

enum_nw_direct = nw.Enum(pl.Series(["a", "b", "c"]))
>>> enum_nw_direct
Enum(categories=('a', 'b', 'c'))
```

We could align the `__repr__` with `polars` as well - but not too important

dangotbanned: > lazy backends don't support it

That part isn't an issue here, since we're just relying on an object with `__iter__` defined when we do 

https://github.com/camriddell/narwhals/blob/3357c551f23d261d5b6f04ea06db72540bef7af2/narwhals/dtypes.py#L460

```py
            self.categories = tuple(categories)
```

MarcoGorelli: ah sorry I thought you were suggesting to store `categories` in a `nw.Series`, instead perhaps you're suggesting that `nw.Series` should work as an input to `categories=`? if so, i agree, and I think that that should indeed work

MarcoGorelli (approved):
thanks @camriddell , and @dangotbanned for reviews!

i think this is mostly ready?

narwhals/dtypes.py
dangotbanned: Yeah apologies for the confusion, I did suggest that before! 😅

Just wanting to be *extra* cautious, since it currently works but isn't documented or included in tests - so could easily be lost in a refactor later

dangotbanned:
> thanks @camriddell , and @dangotbanned for reviews!
> 
> i think this is mostly ready?

Thanks @MarcoGorelli 

Yeah I think after (https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2041081799) - my only other unresolved note was adding some guards like in https://github.com/pola-rs/polars/blob/py-1.26.0/py-polars/polars/datatypes/classes.py#L645-L727

@camriddell seemed on board with the suggestion - we've just been a bit more active 😉
- https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035728791
- https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035794136

narwhals/stable/v1/_dtypes.py
dangotbanned: Ah I see you'd already resolved this @MarcoGorelli (https://github.com/narwhals-dev/narwhals/pull/2192/commits/000013cd3756e2507c7dbf0ee86b5d43ddd0646d#diff-0d9c3865bf2fab7dc5c9b174968b375828bc0578de5e5b5e7ac9ce93a4711f0c)

camriddell:
> > thanks @camriddell , and @dangotbanned for reviews!
> > i think this is mostly ready?
> 
> Thanks @MarcoGorelli
> 
> Yeah I think after ([#2192 (comment)](https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2041081799)) - my only other unresolved note was adding some guards like in https://github.com/pola-rs/polars/blob/py-1.26.0/py-polars/polars/datatypes/classes.py#L645-L727
> 
> @camriddell seemed on board with the suggestion - we've just been a bit more active 😉
> 
> * [Feat: let `nw.Enum` accept categories, map pandas ordered categorical to Enum (only in main namespace, not stable.v1) #2192 (comment)](https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035728791)
> * [Feat: let `nw.Enum` accept categories, map pandas ordered categorical to Enum (only in main namespace, not stable.v1) #2192 (comment)](https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035794136)
> 
> My concern there is if we leave it to raise when it reaches each backend - we won't have consistent errors and the traceback might not be super clear where the origin was

I think I'm in favor of taking the strictest route here, and that is to mimic Polars again:
1. Only allow containers of strings as valid categories
2. Explicitly disallow `None` and `float('nan')`
3. Explicitly disallow

 
Adding onto this discussion:

If we only let backends raise, we will hit an issue where some code only work with specific backends which reduces the purpose of Narwhals. With the current Enum targeting pandas_like and Polars, I see this primarily happening in the space that Polars ONLY allows strings as valid category members, yet pandas allows any hashable Py object. I'm leaning towards taking the stricter set of requirements (only strings) and sticking with Polars behavior.

For the checks themselves:
- detecting duplicates is very doable.
- detecting if inputs are `NaN` can likely be relied on by verifying `value != value` as this follows IEEE-754 impl. so it should be fairly uniform across backends (I believe we also use this semantic in other places internally).
- detecting if inputs are Null may be problematic since we would need to verify is one passed along None, `pandas.NA`, and any other "Null" representing value out there.

Is it too aggressive to propose that categories should only be a container of Python built-in types and error on duplicate values, `float('nan')` and `None`? I'll need to do a usage check to see how frequently `Enum`s are derived in patterns opposite to this, I would guess that they are most often created from explicit user-defined sequences using Python built-ins. 

That said, maybe supporting both arbitrary Py Sequences and narwhals Series can be done for the most flexible approach? That way we can support either pattern for users.

docs/backcompat.md
```diff
@@ -113,6 +114,11 @@ before making any change.
 
 The following are differences between the main Narwhals namespace and `narwhals.stable.v1`:
 
+- Since Narwhals 1.35:
+
+  - pandas' ordered categoricals get mapped to `nw.Enum` instead of `nw.Categorical`.
+  - `nw.Enum` must accept categories at instantiation.
```
dangotbanned, line 120:
```suggestion
  - `nw.Enum` must be provided `categories` at instantiation.
```

dangotbanned: @camriddell will try to do a proper review tomorrow

narwhals/_dask/utils.py
```diff
@@ -125,6 +125,20 @@ def narwhals_to_native_dtype(dtype: DType | type[DType], version: Version) -> An
...
+        if version is Version.V1:
+            msg = "Converting to Enum is not supported in narwhals.stable.v1"
+            raise NotImplementedError(msg)
+        if isinstance(dtype, dtypes.Enum):
+            import pandas as pd
+
+            # NOTE: `pandas-stubs.core.dtypes.dtypes.CategoricalDtype.categories` is too narrow
+            # Should be one of the `ListLike*` types
+            # https://github.com/pandas-dev/pandas-stubs/blob/8434bde95460b996323cc8c0fea7b0a8bb00ea26/pandas-stubs/_typing.pyi#L497-L505
+            return pd.CategoricalDtype(dtype.categories, ordered=True)  # pyright: ignore[reportArgumentType]
```
dangotbanned, line 138: @MarcoGorelli does this seem like it could be widened upstream (https://github.com/pandas-dev/pandas-stubs)?

I've traced back the runtime check to `is_list_like`:
- https://github.com/pandas-dev/pandas/blob/5fef9793dd23867e7b227a1df7aa60a283f6204e/pandas/_libs/lib.pyx#L1236-L1282

The current annotation only permits `list[Any]` as a non-`pandas` `Sequence`
- https://github.com/pandas-dev/pandas-stubs/blob/8434bde95460b996323cc8c0fea7b0a8bb00ea26/pandas-stubs/core/dtypes/dtypes.pyi#L36

dangotbanned (commented):
Thanks for the PR @camriddell!

I've left some *non-blocking* comments/questions.
Looking pretty ready to me 🎉

narwhals/_dask/utils.py
dangotbanned: @camriddell ignore this, I only meant to add as a comment - not the review 🫣

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2192#pullrequestreview-2775252785

> This is going to cause issues for people even just inspecting the schema of a dataframe

That's a good point @MarcoGorelli 

If someone is currently doing that operation, on `v1` would look like this:

```py
import pandas as pd

from narwhals.stable import v1 as nw_v1

s = pd.Series([1, 2, 3], dtype=pd.CategoricalDtype(ordered=True))
>>> nw_v1.from_native(s, series_only=True).dtype
Categorical
```

So far we've had two options:
1. Being lax with categories [1](https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2035728791), [2](https://github.com/narwhals-dev/narwhals/pull/2192#discussion_r2040722920)
2. Using the stricter rules from `polars` [3](https://github.com/narwhals-dev/narwhals/pull/2192#issuecomment-2803171089), [4](https://github.com/narwhals-dev/narwhals/pull/2192#issuecomment-2799896654)

I see two other tweaks we could do to option **2** - when we can't meet the constraints of [`pl.Enum`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Enum.html)
- Just continue mapping `pd.CategoricalDtype` -> `nw.Categorical`
  - No change in behavior, breaks no-one
- Use an alternative constructor for `pd.CategoricalDtype` -> `nw.Enum`
  - So we'd still reject `nw.Enum([1, 2, 3])`
  - But we'd allow existing ordered categoricals to be represented by an ordered type

I think either of those would solve the problem, but I think the simplest is to just keep using `nw.Categorical`

dangotbanned:
## `altair`-related
@MarcoGorelli 
> In particular, it would be a breaking change for Altair users, who'd no longer be able to plot pandas dataframes where columns are of categorical dtype and have non-string categories. 
> It's probably not showing up at the moment in the downstream tests because we were careful to use `narwhals.stable.v1`

I could be wrong, but I don't think we have any paths that would hit this - even if we *weren't* using `v1`?
AFAICT, `pandas` is handled natively - since the type conversion logic predates `narwhals` and (I assume) we didn't wanna make a breaking change.

### Impl
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/altair/utils/data.py#L316-L333
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/altair/utils/core.py#L674-L680
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/altair/utils/core.py#L704-L714
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/altair/utils/core.py#L388-L394
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/altair/utils/core.py#L249-L252

### Tests
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/tests/utils/test_utils.py#L46-L47
- https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/tests/vegalite/v5/test_api.py#L238

### Docs
- https://altair-viz.github.io/user_guide/encodings/index.html#encoding-data-types

MarcoGorelli:
This is the part that would break in Altair:

https://github.com/vega/altair/blob/f1e0049e6f6669ec46ec462cec81ce62aae8cbf2/altair/utils/core.py#L670-L671

It would also affect Plotly and other libraries

I think it's fine to be laxer here - Polars only allows string column names, but we allow pandas dataframe with non-string column names. Similarly, we can allow pandas dataframes with non-string categories

I think it's legit to do something like
```
s: nw.Series
categories = list(s.dtype.categories)
categories.append(new_value)
nw.new_series('a', values, dtype=nw.Enum(categories))
```
where the categories are taken from user inputs. If the user is starting with something which a backend permits, they can continue with that, no issues

MarcoGorelli:
sure, thanks for explaining

true, there is a risk that someone writes something which doesn't end up working for polars, but i'd rather accept that risk that disallow people from passing valid pandas dataframes to narwhals

i think one reason for narwhals' relatively rapid growth has been that, relative to similar/competing projects, we've put a lot of emphasis on there not being any cost to existing pandas users

narwhals/_dask/utils.py
dangotbanned: @MarcoGorelli gentle nudge on this, in case it was missed

MarcoGorelli: hey - yeah, probably, the pandas stubs definitely don't get all the attention they probably deserve

narwhals/_pandas_like/utils.py
```diff
@@ -211,7 +211,9 @@ def rename(
 
 
 @functools.lru_cache(maxsize=16)
-def non_object_native_to_narwhals_dtype(dtype: str, version: Version) -> DType:
+def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> DType:
+    dtype = str(native_dtype)
+
```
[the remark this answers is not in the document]
camriddell: > Which would mean we do the str(...) call twice now. Just an observation, not sure if there is a cost to that

I think the cost of a repeated call to `str` should be fairly negligible, we can always come back later to refactor if a profiler disagrees with this statement and this leads to a larger overhead.

> Are all non-object pandas data types guaranteed to be immutable?
> I think str was used because it is hashable, so is safe to use in [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)

Since the tests pass, I am at least confident that all of the datatypes are **hashable**, however whether that hash is something meaningful or just the default `id(self) / 16` then caching may not be reliable. That said, perhaps we can also leave as is for now, then if we catch wind of a slow down in the future we can revisit it? Trying to avoid the pre-mature optimization scenarios here.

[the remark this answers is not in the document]
dangotbanned: @camriddell agreed on the `str` part.

My concern on the hashability though is related to https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1970074662

Right now we won't get a warning like that because we have:

```py
native_dtype: Any
```

However - good news!
I changed it to this locally:

```py
def non_object_native_to_narwhals_dtype(
    native_dtype: pd.api.extensions.ExtensionDtype, version: Version
) -> DType:
```

And [followed though](https://github.com/pandas-dev/pandas/blob/5f354ca51f24cfb3e6a11c0c3367d0abf8676637/pandas/core/dtypes/base.py#L78) to the docs to find:

>     ExtensionDtypes are required to be hashable. The base class provides

Looks like we're all good 🙂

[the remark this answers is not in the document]
camriddell: Great find on that one, thanks so much for diving in there!

camriddell:
@MarcoGorelli I believe the current version meets the changes you requested? When you have a chance can you take another look?

MarcoGorelli (approved):
thanks, looks good to me!

@dangotbanned any objections?

dangotbanned:
> thanks, looks good to me!
> 
> @dangotbanned any objections?
> 

@MarcoGorelli just wanna double check this was what you asked for?

[remove enum duplication/null checks](https://github.com/narwhals-dev/narwhals/pull/2192/commits/3902a29f6aea952125484bfdd45018e29c87c4dc)

I thought in (https://github.com/narwhals-dev/narwhals/pull/2192#issuecomment-2812967946) you just wanted to allow non-strings - not allow duplicates and `None`

MarcoGorelli:
thanks!

the backends themselves already disallow duplicates and nulls, so tbh i'm not super-bothered, especially given that people will usually be inspecting schemas of dataframes containing enums rather than making new ones

dangotbanned (approved):

PR #2206, opened 2025-03-14
feat!: remove require at least one expression be passed to lazyframe select and lazyframe.with_columns, remove lazyframe.clone

### LINKED ISSUE #2205: api: disallow `LazyFrame.select()`
`lf.select()` is a bit problematic and varies a bit across backends, as for duckdb/sqlframe it seems we don't have a way to return a table with 0 rows and 0 columns

I propose to just require that at least one expression be passed to `LazyFrame.select`, problem solved

MarcoGorelli:
closes #2205

This doesn't break any downstream tests, and `narwhals.LazyFrame` isn't widely used (yet 🤞 ) so I think it's OK to be slightly more aggressive with changes there

We can always yank / revert if anyone complains and go through the usual process

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/extremes.yml
+++ b/.github/workflows/extremes.yml
@@ -142,8 +142,7 @@ jobs:
       - name: install-reqs
         run: uv pip install -e ".[tests]" --system
       - name: install-kaggle
-        # https://github.com/Kaggle/kaggle-api/issues/718
-        run: uv pip install "kaggle<1.7.4" --system
+        run: uv pip install kaggle --system
       - name: Download Kaggle notebook artifact
         env:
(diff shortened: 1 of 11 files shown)


MarcoGorelli:
discussed yesterday in the call, no objections

Issue #1888, opened 2025-01-28
Deprecate `native_namespace` in favour of `backend` in IO functions


MarcoGorelli:
`native_namespace` is annoyingly long, and retrieving native namespaces can be tedious

I think `from_dict` (and other IO functions) could have a `backend: ModuleType | Implementation | EagerBackend` argument, so all of the following could be supported:
```python
data = {'a': [1,2,3], 'b': [4,5,6]}

nw.from_dict(data, backend=pyarrow)
nw.from_dict(data, backend='pyarrow')
nw.from_dict(data, backend=nw.Implementation.PYARROW)
```

`EagerBackend` could be `Literal['pandas', 'polars', 'pyarrow']`

# Tasks
- [x] #1931
- [x] #2063
- [x] #2212
- [x] #2216
- [x] #2218 
- [x] #2217
- [x] #2215
- [x] #2219

MarcoGorelli:
I'm going to reopen this as it's not yet completed for all IO functions

`read_csv`, `read_parquet`, `new_series`, are all missing

dangotbanned:
Also `from_numpy`, `from_arrow`, `new_series`

PR #2212, opened 2025-03-14
chore: Deprecate `native_namespace` in favour of `backend` in `new_series`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #1888

## Checklist

- [ ] Tests added
- [x] Documented the changes

- `validate_native_namespace_and_backend` had a static version `1.25.1`
- Didn't cover double `None` thats needed here and in #2063
- Added a decorator to make this easier for the rest
  - `from_numpy`
  - `from_arrow`

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -238,47 +238,46 @@ jobs:
           cd tea-tasting
           pdm run test
 
-  tubular:
-    strategy:
-      matrix:
-        python-version: ["3.12"]
-        os: [ubuntu-latest]
+  # temporarily commented out until they address the `new_series` deprecation
+  # tubular:
+  #   strategy:
+  #     matrix:
+  #       python-version: ["3.12"]
+  #       os: [ubuntu-latest]
 
-    runs-on: ${{ matrix.os }}
-    steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-      - name: Install uv
-        uses: astral-sh/setup-uv@v5
-        with:
-          enable-cache: "true"
-          cache-suffix: ${{ matrix.python-version }}
-          cache-dependency-glob: "pyproject.toml"
-      - name: clone-tubular
-        run: |
-          git clone https://github.com/lvgig/tubular --depth=1
-          cd tubular
-          git log
-      - name: install-basics
-        run: uv pip install --upgrade tox virtualenv setuptools pytest-env --system
-      - name: install-tubular-dev
-        run: |
-          cd tubular
-          uv pip install -e ".[dev]" --system
-      - name: install-narwhals-dev
-        run: |
-          uv pip uninstall narwhals --system
-          uv pip install -e . --system
-      - name: show-deps
-        run: uv pip freeze
-      - name: Run pytest
-        run: |
-          cd tubular
-          # temporarily ignoring warnings until they address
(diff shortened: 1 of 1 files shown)


dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/13864172160/job/38799425325?pr=2212

@MarcoGorelli do you know if `tubular` is using `v1`?

dangotbanned:
If we ever want to get more fancy with the deprecations, `polars` has some nice generalized decorators
https://github.com/pola-rs/polars/blob/b9d9ab1da52b343de87cf8fea95d99a37a9baa26/py-polars/polars/_utils/deprecation.py

narwhals/functions.py
```diff
@@ -220,24 +236,20 @@ def new_series(
...
     """
-    return _new_series_impl(
-        name,
-        values,
-        dtype,
-        native_namespace=native_namespace,
-        version=Version.MAIN,
-    )
+    backend = cast("ModuleType | Implementation | str", backend)
+    return _new_series_impl(name, values, dtype, backend=backend, version=Version.MAIN)
```
dangotbanned, line 240: FWIW, I'm not that happy about needing:
- `typing.cast` on `backend`
- `# noqa: ARG001` on `native_namespace`

It seemed like the *lowest friction* solution to me.

## Rough alternative ideas
- Including the `_*_impl` function within the decorator
  - Not sure what the body of this function would look like after though
- Spending more time looking into how the [typing spec](https://typing.python.org/en/latest/spec/) wants *signature-changing* decorators to work
  - I'll be honest I don't have much hope for this
  - IIRC, pre-pending positional-only args is the main supported case
- Rethinking how this mechanism works for `stable` vs `v1`
  - E.g. maybe we'd want to do things differently with `_stableify` as well?
  - Stacking multiple decorators may be helpful in the future

MarcoGorelli:
thanks - looks like there's a few places where they're not using `v1`, so I think it's ok if they get a deprecationwarning

dangotbanned:
> thanks @dangotbanned !
> 
> we could temporarily use `-W ignore` for the tubular tests, until they address this?

Thanks @MarcoGorelli 

Yeah that sounds good - happy for you to add that 🙌

MarcoGorelli:
looks like they're testing the messages a bit too tightly
```
 E           AssertionError: assert '`native_name...nformation.\n' == 'ModeImputer:...rst candidate'
E             
E             - ModeImputer: The Mode of column float_col is tied, will sort in descending order and return first candidate
E             + `native_namespace` is deprecated, please use `backend` instead.
E             + 
E             + Note: `native_namespace` will remain available in `narwhals.stable.v1`.
E             + See https://narwhals-dev.github.io/narwhals/backcompat/ for more information.
```
temporarily turning it off then

MarcoGorelli:
Hi @limlam96 - just looping you in here, as your test suite will probably start failing because of this

I think if you either:
- use `narwhals.stable.v1` consistently
- bump the minimum narwhals version and use `new_series(..., backend=)` instead of `new_series(..., native_namespace=)`

then it should become green again

dangotbanned:
@MarcoGorelli I'll leave this in your hands to merge

We can probably get the remaining parts of #1888 done pretty quickly afterwards

MarcoGorelli:
sure thanks!

PR #2220, opened 2025-03-15
Implement `fill_null` for spark-like

### LINKED ISSUE #2210: enh: implement `fill_null` for spark-like

lucas-nelson-uiuc:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2210 

## Checklist

- [x] Tests added
- [ ] Documented the changes

Supports filling null values with:
- scalar values (e.g. `fill_null(value=99)`)
- expressions (e.g. `fill_null(value=nw.col("c"))`
- strategies ("forward" and "backward")
- limits

Also removed the checks for the PySpark constructor in the respective tests.


lucas-nelson-uiuc:
Also thinking of rewriting this to be more like `clip`, where the different strategies are their own functions that we feed into `self._from_call`.

For example, we can define `_fill_null_forward` and `_fill_null_backward` similar to `_clip_upper` and `_clip_lower`

https://github.com/narwhals-dev/narwhals/blob/1b8d5487efa61affdb9c0bf776392eb880df9c78/narwhals/_spark_like/expr.py#L419-L445

dangotbanned:
> Not sure why some tests are failing - some point back to `duckdb`, specifically `sqlframe.duckdb.functions` not having `monotonically_increasing_id()`

Seems like only `"spark"` supports that function

https://github.com/eakmanrq/sqlframe/blob/d694e3acc827b94a9736ee38e4c2a8e0c924024b/sqlframe/base/functions.py#L685-L687

```py
@meta(unsupported_engines=["duckdb", "postgres", "bigquery", "snowflake", "redshift"])
def monotonically_increasing_id() -> Column:
    return Column.invoke_anonymous_function(None, "MONOTONICALLY_INCREASING_ID")
```

narwhals/_spark_like/expr.py
```diff
@@ -557,6 +557,47 @@ def func(
...
+                )
+                window_spec = (
+                    self._Window()
+                    .orderBy(self._F.monotonically_increasing_id())
+                    .rowsBetween(0, upper_limit)
+                )
+                value = self._F.first(_input, ignorenulls=True).over(window_spec)
+            if not isinstance(value, SparkLikeExpr):
+                value = self._F.lit(value)
+            return self._F.ifnull(_input, value)
```
dangotbanned, line 597: Could you explain what this check is for?

It introduces this error https://github.com/narwhals-dev/narwhals/actions/runs/13875048994/job/38826287314?pr=2220

Because now the line below *could* be `SparkLikeExpr`:

```py
return self._F.ifnull(_input, value)
```

But every call would only accept the **native** (`Column`) and not the **compliant** (`SparkLikeExpr`) type

narwhals/_spark_like/expr.py
```diff
@@ -557,6 +557,47 @@ def func(
...
+            if strategy == "forward":
+                lower_limit = (
+                    self._Window().unboundedPreceding if limit is None else -limit
+                )
+                window_spec = (
+                    self._Window()
+                    .orderBy(self._F.monotonically_increasing_id())
+                    .rowsBetween(lower_limit, 0)
+                )
+                value = self._F.last(_input, ignorenulls=True).over(window_spec)
```
MarcoGorelli, line 584: can we start with just implementing `value` please?

order-dependent operations need to be implemented differently (see `cum_sum` and `rolling_sum` for examples)

MarcoGorelli (changes requested):
thanks @lucas-nelson-uiuc for your PR!

let's start with just supporting `value`

narwhals/_spark_like/expr.py
lucas-nelson-uiuc: Without that check I kept getting an error if I tried passing an expression as the fill value - something like `nw.col("a").fill_null(nw.col("c"))`. I can revisit it in a moment after removing the strategy/limit parameters

lucas-nelson-uiuc: New version no longer includes this check - thanks for pointing this out

narwhals/_spark_like/expr.py
lucas-nelson-uiuc: yea ofc ^ just updated to only include `value`

MarcoGorelli (commented):
thanks @lucas-nelson-uiuc !

this looks correct, there's just some failing tests, if you can get CI green 🥦 then I think we can ship this

narwhals/_spark_like/expr.py
```diff
@@ -557,6 +557,17 @@ def func(
 
         return self._with_window_function(func)
 
+    def fill_null(
+        self,
+        value: Any | None = None,
+        strategy: Literal["forward", "backward"] | None = None,
+        limit: int | None = None,
```
MarcoGorelli, line 564:
```suggestion
        value: Any | None,
        strategy: Literal["forward", "backward"] | None,
        limit: int | None,
```
we generally avoid specifying default at the compliant level

MarcoGorelli (approved):
thanks @lucas-nelson-uiuc , much appreciated!

narwhals/_spark_like/expr.py
```diff
@@ -557,6 +557,17 @@ def func(
 
         return self._with_window_function(func)
 
+    def fill_null(
+        self,
+        value: Any | None,
+        strategy: Literal["forward", "backward"] | None,
+        limit: int | None,
```
MarcoGorelli, line 564: sorry hadn't spotted that we're now silently ignoring these

if we want this to be part of tomorrow's release, can we raise if `strategy` is not `None`?

longer-term solution would be to change the metadata kind in `Expr.fill_null` to `WINDOW` if `strategy` is passed, and preserve it otherwise

lucas-nelson-uiuc: yea no problem - would you want the same for `limit`?

```python
if strategy is not None: # or (limit is not None)
    msg = "Support for strategies is not yet implemented."
```

Issue #2124, opened 2025-03-01
Tracking: `pyarrow-stubs` upstream fixes


dangotbanned:
Following #2007, there were a lot of things that still needed addressing upstream.

The end-goal will be removing/greatly reducing the need for the [`Incomplete`](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+path%3A%2F%5Enarwhals%5C%2F_arrow%5C%2F%2F+%22Incomplete%22&type=code&rgh-link-date=2025-02-20T12%3A00%3A41Z) marker and excessive use of `typing.cast(...)` - to get things working

## PRs
I expect this list may grow longer - depends on how many things are solved by (https://github.com/zen-xu/pyarrow-stubs/pull/181)

- [x] https://github.com/zen-xu/pyarrow-stubs/pull/177
- [ ] https://github.com/zen-xu/pyarrow-stubs/pull/181
- [ ] https://github.com/zen-xu/pyarrow-stubs/pull/182
- [ ] https://github.com/zen-xu/pyarrow-stubs/pull/183

dangotbanned:
We're currently blocked on upgrading from ([`pyarrow-stubs==19.2`](https://github.com/zen-xu/pyarrow-stubs/releases/tag/19.2))

I created a repro a while back but forgot to post it anywhere 😅

Before bumping the stubs, **we realistically need all of these to pass again** - otherwise we'll start getting tons of false negatives *again*

*Specifically*, fixes in (https://github.com/zen-xu/pyarrow-stubs/pull/183) have **regressed** following (https://github.com/zen-xu/pyarrow-stubs/pull/233)

That problematic PR was the reason given for closing (https://github.com/zen-xu/pyarrow-stubs/pull/208) and (https://github.com/zen-xu/pyarrow-stubs/pull/209) which would have fixed more for us 😞 

## Repro

```py
# t.py
# 19.2
# pyright: 0 errors, 0 warnings, 0 informations
# mypy: Success: no issues found in 1 source file
# ---------------
# 20.0.0.20250716
# pyright: 22 errors, 0 warnings, 0 informations
# mypy: Found 13 errors in 1 file (checked 1 source file)
from __future__ import annotations

from typing import TYPE_CHECKING, Any

import pyarrow as pa
import pyarrow.compute as pc

if TYPE_CHECKING:
    from collections.abc import Iterable

tp_1 = pa.dictionary(pa.int32(), pa.string())
[...]

Issue #2230, opened 2025-03-17
Formalise `Compliant*` protocols


dangotbanned:
Probably should've opened this sooner to track existing efforts, but better late than never.

# Tasks
- [x] #2130
  - [x] #2249 
  - [x] #2290
  - [x] #2293
  - [ ] #2294
- [x] #2149
- [x] #2064
- [x] #2119
- [x] #2202
- [x] #2223
  - [x] #2251
- [x] #2232
- [x] #2236
- [x] #2252
  - See (https://github.com/narwhals-dev/narwhals/issues/2184#issuecomment-2710918460)
- [x] #2261

# Related
- #2116
- #1942
- #1886
- #2222

PR #2232, opened 2025-03-17
chore: Spec `CompliantLazyFrame`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2230
- #1644
- #2239 

## Checklist

- [ ] Tests added
- [x] Documented the changes

- [x] Reduce `Incomplete` markers (https://github.com/narwhals-dev/narwhals/pull/2232/commits/df00903afbd9e00e4a942ddbf3aa303a8cb7f906)
- [x] Get feedback on what to do with deprecations


dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/13902763286/job/38898383145?pr=2232

lol, need to pause this PR to resolve #2124

narwhals/_dask/dataframe.py
```diff
@@ -425,3 +427,11 @@ def unpivot(
...
             )
         )
+
+    explode = not_implemented()
+    to_arrow = not_implemented.deprecated(
+        "only if version is v1, keep around for backcompat"
+    )
+    to_pandas = not_implemented.deprecated(
+        "only if version is v1, keep around for backcompat"
+    )
```
dangotbanned, line 437: Preview of what (https://github.com/narwhals-dev/narwhals/pull/2232/commits/b15976e91984b6375b0d0759e49c3734727514d1) enabled

image

dangotbanned:
@MarcoGorelli @FBruzzesi 

I really feel like I need some guidance on what the *intended* API is for `CompliantLazyFrame`.
Working through the PR has left me fairly unsure - particularly the deprecated bits 🤔

FBruzzesi:
> I really feel like I need some guidance on what the _intended_ API is for `CompliantLazyFrame`. Working through the PR has left me fairly unsure - particularly the deprecated bits 🤔

@dangotbanned I fell so behind with the progress you made in the last 2-3 weeks, so I am not sure I can help 🙈

Regarding deprecation, that would not affect v1 - would it work to have `CompliantLazyFrame` be the spec for main, and (something like) `V1CompliantLazyFrame` inherit from it and add those deprecated methods?

dangotbanned:
> @dangotbanned I fell so behind with the progress you made in the last 2-3 weeks, so I am not sure I can help 🙈

aha no worries @FBruzzesi - it might look like a lot, but really its just formally defining what was already there 🙂 

> Regarding deprecation, that would not affect v1 - would it work to have `CompliantLazyFrame` be the spec for main, and (something like) `V1CompliantLazyFrame` inherit from it and add those deprecated methods?

That could work.
One thing to keep in mind is we'd need a `V1` and `MAIN` (non-prefixed) version of **all the classes** (I'm talking `_arrow`, `_dask`, `_...`).

Initially that might seem insane - but a big positive is we wouldn't need to pass around the `Version` for every new object.
An idea I had a while back (https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-12054784) may come in handy for that - if we went down that route.

Definitely is something we need to have a clear plan for *before* moving to `V2`

MarcoGorelli:
Do we need to include deprecated or v1-only methods at all? Can the compliant protocols only cover the main namespace?

dangotbanned:
> Do we need to include deprecated or v1-only methods at all? Can the compliant protocols only cover the main namespace?

Good questions @MarcoGorelli!

We don't *need* to include anything 😉 
More seriously though, I'm hoping I can tease out what role(s) we *want* the protocols to serve?

## Example
> only cover the main namespace

If we take that approach and then move to `v2`, all `Compliant*` are required to implement `v2` but not `v1`.
That seems like it *could* be a footgun for 2 reasons:

1. We might *accidentally* remove things from a class and break compatibility with `v1`.
  i. We lose the type safety that would've caught that statically
2. `v1` is not frozen when we move to `v2`
  i. All changes in `v2` have the *potential* to impact a `v1` implementation - unless the method depends on nothing outside of it

## What does this PR describe?

### Normal
We've got the main namespace covered with regular definitions like this.
The assumption would be every implementation **must** have that member or a `not_implemented`:
https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_compliant/dataframe.py#L189

### Deprecation (1)

Methods that were deprecated in `v1` look like:
https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_compliant/dataframe.py#L234-L235

New information I've captured is **deprecated functionality, that was not implemented**.
My assumption is that this will never be implemented:

https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_duckdb/dataframe.py#L478-L483

I think that's a useful distinction from **functionality that is not implemented, but may be in the future**:

https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_duckdb/dataframe.py#L484

### Deprecation (2)

There's a second kind of deprecation I'm not too sure on, like this:

https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_compliant/dataframe.py#L236-L239

<details><summary>Where <code>duckdb</code> is the <i>ugly</i> duckling</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_duckdb/dataframe.py#L200-L212

</p>
</details> 

<details><summary><i>ugly</i> because it introduces</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/df00903afbd9e00e4a942ddbf3aa303a8cb7f906/narwhals/_spark_like/dataframe.py#L483-L488

</p>
</details> 


## What else can we describe?
Outside of this PR, I've been using `@unstable` - which *currently* has no runtime or static behavior.

<details><summary><code>@unstable</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/275c5b639c0171a4cc23451f01795742826537fa/narwhals/utils.py#L1507-L1529

</p>
</details> 

<details><summary>Usage</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/a38c9db04b1c51dce5681338d7c98ebe95792e5a/narwhals/_compliant/series.py#L203-L210

</p>
</details> 

# Summary
Apologies this got quite long!

Bringing it all back to:

> what role(s) we *want* the protocols to serve?

- We can encode as *much* or as *little* information as you like into the protocols.
- The [Perfect backwards compatibility policy](https://narwhals-dev.github.io/narwhals/backcompat/) seems an important concept to the project.
- IMO, we have a better chance at avoiding future regressions by encoding deprecations into `Compliant*` protocols

> [!NOTE]
> I'm very open to ideas on how to express these concepts differently (*or not at all*)

MarcoGorelli:
> Where duckdb is the ugly duckling

same story with Ibis

Shall we start with just typing according to the main namespace?

dangotbanned:
> > Where duckdb is the ugly duckling
> 
> same story with Ibis
> 
> Shall we start with just typing according to the main namespace?

So that would mean drop these two:

- `CompliantLazyFrame.to_arrow`
- `CompliantLazyFrame.to_pandas`

And then skip *"introducing"*:

- `(Dask|Polars|SparkLike)LazyFrame.to_arrow`
- `(Dask|Polars|SparkLike)LazyFrame.to_pandas`

Yeah that makes sense to me @MarcoGorelli

narwhals/_compliant/dataframe.py
```diff
@@ -155,26 +160,110 @@ def write_csv(self, file: str | Path | BytesIO | None) -> str | None: ...
     def write_parquet(self, file: str | Path | BytesIO) -> None: ...
 
 
-class CompliantLazyFrame(Protocol):
+class CompliantLazyFrame(
+    _StoresNative[NativeFrameT_co], Protocol[CompliantExprT_contra, NativeFrameT_co]
+):
```
dangotbanned, line 165: I'm trying to establish a pattern of always using the **native** type last if present

narwhals/_compliant/dataframe.py
```diff
@@ -155,26 +160,112 @@ def write_csv(self, file: str | Path | BytesIO | None) -> str | None: ...
...
+        ...
+
+    def aggregate(self, *exprs: CompliantExprT_contra) -> Self:
+        """`select` where all args are aggregations or literals.
+
+        (so, no broadcasting is necessary).
+        """
+        ...
+
+    def _change_version(self, version: Version) -> Self: ...
```
dangotbanned, line 185: Just to provide some more context on this message (https://github.com/narwhals-dev/narwhals/pull/2232/commits/ac25fde74461dff381907d2198d04cbb05efea95)

I *thought* I'd got pretty far along in being able to resolve (https://github.com/narwhals-dev/narwhals/issues/2239#issuecomment-2733748367)

I had `pyright` happy, but then gave `mypy` a try and :cold_sweat:...

```cmd
>>> mypy
Found 795 errors in 140 files (checked 355 source files)
```

dangotbanned: Adding `_change_version` was the only salvageable part of that ordeal

narwhals/_compliant/dataframe.py
```diff
@@ -136,7 +141,7 @@ def unique(
         subset: Sequence[str] | None,
         *,
         keep: Literal["any", "first", "last", "none"],
-        maintain_order: bool | None,
+        maintain_order: bool | None = None,
```
MarcoGorelli, line 144: why add a default?

narwhals/_compliant/dataframe.py
```diff
@@ -155,26 +160,112 @@ def write_csv(self, file: str | Path | BytesIO | None) -> str | None: ...
...
+        right_on: str | None,
+        by_left: Sequence[str] | None,
+        by_right: Sequence[str] | None,
+        strategy: Literal["backward", "forward", "nearest"],
+        suffix: str,
+    ) -> Self: ...
+    def lazy(self, *, backend: Implementation | None) -> Self | Incomplete:
+        # The `backend`` argument has no effect but we keep it here for
+        # backwards compatibility because in `narwhals.stable.v1`
+        # function `.from_native()` will return a DataFrame for DuckDB.
```
MarcoGorelli, line 233: can we keep this out of the compliant protocol and just adjust in duckdb/ibis?

MarcoGorelli (commented):
thanks, looks good, just a couple of comments

narwhals/_compliant/dataframe.py
dangotbanned: So I needed to supress the messages in the commit descriptions of:

(https://github.com/narwhals-dev/narwhals/pull/2232/commits/e70cc4719ce2c81ccab3805463b74cc2edc8933d)
(https://github.com/narwhals-dev/narwhals/pull/2232/commits/17ca5e8e48d027402163b044d0a0ec5dd7c1904f)

But they do need the default since `LazyFrame` calls it without `maintain_order`

https://github.com/narwhals-dev/narwhals/blob/4100cb7c72d9680ae1ef2f62d7fcc73a8c3d2494/narwhals/dataframe.py#L2721-L2731

dangotbanned: The culprit is this - which was needed as soon as you mixed `CompliantDataFrame` and `CompliantLazyFrame`

https://github.com/narwhals-dev/narwhals/blob/4100cb7c72d9680ae1ef2f62d7fcc73a8c3d2494/narwhals/_pandas_like/dataframe.py#L763-L777

narwhals/_compliant/dataframe.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2232/commits/9a523fffa2696976f6781ff68fb6e3671272d952)

narwhals/_compliant/dataframe.py
MarcoGorelli: this is fine - I guess we could call it `_with_version` for consistency with how we're now naming similar methods?

narwhals/_compliant/dataframe.py
```diff
@@ -155,26 +160,102 @@ def write_csv(self, file: str | Path | BytesIO | None) -> str | None: ...
...
+        self, *by: str, descending: bool | Sequence[bool], nulls_last: bool
+    ) -> Self: ...
+    @deprecated("`LazyFrame.tail` is deprecated and will be removed in a future version.")
+    def tail(self, n: int) -> Self: ...
+    def unique(
+        self,
+        subset: Sequence[str] | None,
+        *,
+        keep: Literal["any", "first", "last", "none"],
+        maintain_order: bool | None = None,
```
MarcoGorelli, line 242: still slightly bothered by this appearing with the compliantlazyframe spec...will take a look, i'd like to think we can avoid it

narwhals/_compliant/dataframe.py
dangotbanned: > this is fine - I guess we could call it `_with_version` for consistency with how we're now naming similar methods?

@MarcoGorelli Yeah I like that
I think we should open an issue to identify all the places we could change to that style.

I had a similar thought here - but reverted to keep the PR from spiraling:
- [`9e0611d` (#2236)](https://github.com/narwhals-dev/narwhals/pull/2236/commits/9e0611d4840a906321153bddbcddc1fc1644729d)

narwhals/_compliant/dataframe.py
dangotbanned: Definitely isn't ideal - since it leaks how `PandasLike` & `Arrow` work into the lazy side.

It might be avoided by not *explicitly* subclassing `CompliantLazyFrame` in `EagerDataFrame`:

https://github.com/narwhals-dev/narwhals/blob/3b6506bc859eb776266fd1f696ee88ade018f6f4/narwhals/_compliant/dataframe.py#L255-L259

Then just attempting to duck type the bits we need by redefining them in `EagerDataFrame`?

MarcoGorelli: I just tried
```diff
diff --git a/narwhals/_compliant/dataframe.py b/narwhals/_compliant/dataframe.py
index 8c51b8b8..d3108b7e 100644
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -239,7 +239,6 @@ class CompliantLazyFrame(
         subset: Sequence[str] | None,
         *,
         keep: Literal["any", "first", "last", "none"],
-        maintain_order: bool | None = None,
```
and pytest and pyright worked fine, am I missing something (I probably am 😄 )?

dangotbanned: @MarcoGorelli turns out I only needed to add the default to `CompliantDataFrame`

Resolved in (https://github.com/narwhals-dev/narwhals/pull/2232/commits/f906cce3808f3bbbfeccd4cd7db1ad48d3f25702)

MarcoGorelli: hmm
```
        if maintain_order:
            msg = "`maintain_order=True` is not supported for LazyFrame.unique."
            raise ValueError(msg)
        if maintain_order is not None:
            msg = (
                "`maintain_order` has no effect and is only kept around for backwards-compatibility. "
                "You can safely remove this argument."
            )
            warn(message=msg, category=UserWarning, stacklevel=find_stacklevel())
```
tbh I think we should remove this from lazyframe, it doesn't belong there

gonna try this, if none of the downstream tests fail i suggest we just go ahead with it

dangotbanned: That is allowed because `CompliantDataFrame` can add default params - but not required ones when doing:

```py
class Mixed(CompliantDataFrame, CompliantLazyFrame): ...
```

dangotbanned: > I just tried
> 
> ```diff
> diff --git a/narwhals/_compliant/dataframe.py b/narwhals/_compliant/dataframe.py
> index 8c51b8b8..d3108b7e 100644
> --- a/narwhals/_compliant/dataframe.py
> +++ b/narwhals/_compliant/dataframe.py
> @@ -239,7 +239,6 @@ class CompliantLazyFrame(
>          subset: Sequence[str] | None,
>          *,
>          keep: Literal["any", "first", "last", "none"],
> -        maintain_order: bool | None = None,
> ```
> 
> and pytest and pyright worked fine, am I missing something (I probably am 😄 )?

@MarcoGorelli I think we're out of sync on the messages 😄 
I arrived there as well

narwhals/_compliant/dataframe.py
```diff
@@ -238,7 +238,7 @@ def unique(
         self,
         subset: Sequence[str] | None,
         *,
-        keep: Literal["any", "first", "last", "none"],
+        keep: Literal["any", "none"],
```
dangotbanned, line 241: This was overkill

The real problem to solve was `duckdb` using `keep: str`

https://github.com/narwhals-dev/narwhals/blob/a6361cda0a6374d3b176bba7025e9821b87c8171/narwhals/_duckdb/dataframe.py#L349

narwhals/_compliant/dataframe.py
dangotbanned: Moved to

- #2248

Issue #2123, opened 2025-03-01
enh: Passing arguments to `pa.Table.to_pandas`


MarcoGorelli:
### Discussed in https://github.com/narwhals-dev/narwhals/discussions/2121

<div type='discussions-op-text'>

<sup>Originally posted by **dhirschfeld** February 28, 2025</sup>
In my current code I have a very large `pa.Table` that I want to convert to `pd.DataFrame`. In the [`to_pandas`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas) call I specify `split_blocks` and `self_destruct` as I don't want 2 copies of the data in memory.

I'm thinking of porting this code to `narwhals` so was wondering if there's a way to do the same - convert to a `pa.Table` to different format without doubling the memory requirements?

I'm interest in converting to either `pandas` or `polars` hence why I thought `narwhals` might be a good fit.</div>

dangotbanned:
I think improved performance characteristics are a good reason to *not* unify things like (https://github.com/narwhals-dev/narwhals/issues/2056#issuecomment-2674314926).

If anyone wants to pursue this - I've added a branch with the typing tools to support it https://github.com/narwhals-dev/narwhals/compare/main...pa-table-to-pandas

https://github.com/narwhals-dev/narwhals/blob/112cfd93f965eb787ea479f83e2d58e8e8f822c6/narwhals/_arrow/typing.py#L61-L79

You'd still need to add the runtime handling of providing the wrong keywords.
*Statically* - they'd only appear for `pa.Table`:

https://github.com/narwhals-dev/narwhals/blob/112cfd93f965eb787ea479f83e2d58e8e8f822c6/narwhals/dataframe.py#L658-L665

dangotbanned:
Coming back to this as we missed something that I noticed in #2879


The request said:

> In my current code I have a very large `pa.Table` that I want to convert to `pd.DataFrame`. In the [to_pandas](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas) call I specify `split_blocks` and `self_destruct` as I don't want 2 copies of the data in memory.


But I wasn't aware at the time that this was something you can also do in `polars`:

- https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_pandas.html
- https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.to_pandas.html

[...]

PR #2235, opened 2025-03-17
feat: Enable spark-like backends in `scan_parquet`

### LINKED ISSUE #2229: [Enh]: Allow SparkLike implementation in `scan_parquet`
Spotted in #2228 - we can't use `scan_parquet` with any SparkLike implementation.

I didn't think this through too much, but we might _require_ one kwarg to be a spark session or a [`pyspark.sql.DataFrameReader`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrameReader.parquet.html#pyspark-sql-dataframereader-parquet) (respective classes in SQLFrame), and use that to read the parquet file 

This comes with the drawback of deciding how to ~~call~~ name the kwarg to use

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2229 

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -1131,12 +1131,16 @@ def _scan_parquet_impl(
         import pyarrow.parquet as pq  # ignore-banned-import
 
         native_frame = pq.read_table(source, **kwargs)
-    elif implementation is Implementation.SQLFRAME:
-        session = kwargs.pop("session")
-        native_frame = session.read.format("parquet").load(source)
-    elif implementation is Implementation.PYSPARK:
-        session = kwargs.pop("session")
-        native_frame = session.read.format("parquet").options(**kwargs).load(source)
+    elif implementation.is_spark_like():
+        if (session := kwargs.pop("session", None)) is None:
+            msg = "Spark like backends require a session object to be passed in `kwargs`."
+            raise ValueError(msg)
+
+        native_frame = (
+            session.read.format("parquet").load(source)
+            if implementation is Implementation.SQLFRAME
+            else session.read.format("parquet").options(**kwargs).load(source)
+        )
 
     else:  # pragma: no cover
         try:
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -1044,15 +1044,29 @@ def scan_parquet(
     For the libraries that do not support lazy dataframes, the function reads
     a parquet file eagerly and then converts the resulting dataframe to a lazyframe.
 
+    !!! note
+        Spark like backends require a session object to be passed in `kwargs`.
+
+        For instance:
+
+        ```py
+        import narwhals as nw
+        from sqlframe.duckdb import DuckDBSession
+
+        nw.scan_parquet(source, backend="sqlframe", session=DuckDBSession())
+        ```
+
     Arguments:
         source: Path to a file.
         backend: The eager backend for DataFrame creation.
             `backend` can be specified in various ways:
 
             - As `Implementation.<BACKEND>` with `BACKEND` being `PANDAS`, `PYARROW`,
-                `POLARS`, `MODIN` or `CUDF`.
-            - As a string: `"pandas"`, `"pyarrow"`, `"polars"`, `"modin"` or `"cudf"`.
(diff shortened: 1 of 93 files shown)


FBruzzesi:
@MarcoGorelli @dangotbanned this is potentially ready for review - I opened it as draft for two reasons:

- is `session` keyword ok in your opinion?
- SQLFrame does not allow for `.options(..)` to be set 

Let me know what you think :)

narwhals/functions.py
```diff
@@ -1078,6 +1093,19 @@ def scan_parquet(
         |     0  1   4     |
         |     1  2   5     |
         └──────────────────┘
+        >>> nw.scan_parquet(
+        ...     "file.parquet", backend="sqlframe", session=DuckDBSession()
+        ... ).collect()  # doctest:+SKIP
```
dangotbanned, line 1098: Just basing this assumption off of the example.

If we have the `backend`, could we do something like this?

```py

backend = "sqlframe"
implementation = Implementation.from_backend(backend)
if implementation.is_sqlframe():
    from sqlframe.base.session import _BaseSession

    session = _BaseSession.builder.getOrCreate()
```

I'm basing this off type annotations alone btw 😅

narwhals/functions.py
```diff
@@ -1103,6 +1131,13 @@ def _scan_parquet_impl(
         import pyarrow.parquet as pq  # ignore-banned-import
 
         native_frame = pq.read_table(source, **kwargs)
+    elif implementation is Implementation.SQLFRAME:
+        session = kwargs.pop("session")
```
EdAbati, line 1135: What about having a friendlier error message (that repeats the advice in the docstring) like:

```suggestion
        session = kwargs.pop("session", None)
        if not session:
            raise ValueError("Spark like backends require a session object to be passed in `kwargs`.)
```

EdAbati (commented):
You have my +1 of this. 👌 This will allow us to run the TCPH tests on pyspark too 👀 (even if they'll make it a bit slower so probaly not in CI)

narwhals/functions.py
dangotbanned: You could use something like (https://github.com/narwhals-dev/narwhals/issues/2123#issuecomment-2692160476) to add an `@overload` for it.

I suppose the match might be:

```py

from typing import Any
from typing import Literal
from typing import TypedDict
from typing import overload

if TYPE_CHECKING:
    from types import ModuleType

    from typing_extensions import Never

class _SparkLikeScanParquetKwds(TypedDict):
    ... # declare whatever is required/optional

_SparkLikeBackend: TypeAlias = Literal[Implementation.PYSPARK, Implementation.SQLFRAME, "pyspark", "sqlframe"]


@overload
def scan_parquet(
    source: str,
    *,
    backend: None = ...,
    native_namespace: None = ...,
    **kwargs: Any,
) -> Never: ...
@overload
def scan_parquet(
    source: str,
    *,
    backend: _SparkLikeBackend,
    native_namespace: None = ...,  # noqa: ARG001
    **kwargs: Unpack[_SparkLikeScanParquetKwds],
) -> LazyFrame[Any]: ...
@overload
def scan_parquet(
    source: str,
    *,
    backend: ModuleType | Implementation | str | None = None, # <------- Probably needs to be narrowed
    native_namespace: ModuleType | None = None,  # noqa: ARG001
    **kwargs: Any,
) -> LazyFrame[Any]: ...
def scan_parquet(
    source: str,
    *,
    backend: ModuleType | Implementation | str | None = None,
    native_namespace: ModuleType | None = None,  # noqa: ARG001
    **kwargs: Any,
) -> LazyFrame[Any]:
```

dangotbanned: Important thing here is `ModuleType` can't appear in the more specific `@overload` definitions (I think) - to avoid an `[overlapping-overload]`

PR #2238, opened 2025-03-18
chore: move dev dependencies to `dependency-groups`

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

I propose moving all the dev dependencies from `optional-dependencies` to [`dependency-groups`](https://packaging.python.org/en/latest/specifications/dependency-groups/). I think this is a bit cleaner what do you think? :)

uv related resources:
- https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-groups
- https://docs.astral.sh/uv/reference/cli/#uv-pip-install--group

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -236,7 +236,7 @@ jobs:
       - name: run-pytest
         run: |
           cd tea-tasting
-          export PYTEST_ADDOPTS="--rootdir=tea-tasting"
+          export PYTEST_ADDOPTS="--rootdir=$(pwd)"
           pdm run test
 
   # temporarily commented out until they address the `new_series` deprecation
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -236,6 +236,7 @@ jobs:
       - name: run-pytest
         run: |
           cd tea-tasting
+          # empty pytest.ini to avoid pytest using narwhals configs
           touch pytest.ini
           pdm run test
 
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -236,6 +236,7 @@ jobs:
       - name: run-pytest
         run: |
           cd tea-tasting
+          export PYTEST_ADDOPTS="--rootdir=tea-tasting"
           pdm run test
 
   # temporarily commented out until they address the `new_series` deprecation
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -228,7 +228,7 @@ jobs:
         run: |
           cd tea-tasting
           pdm remove narwhals
-          pdm add ./..[tests]
+          pdm add ./..
       - name: show-deps
         run: |
           cd tea-tasting
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -236,7 +236,7 @@ jobs:
       - name: run-pytest
         run: |
           cd tea-tasting
-          export PYTEST_ADDOPTS="--rootdir=$(pwd)"
+          touch pytest.ini
           pdm run test
 
   # temporarily commented out until they address the `new_series` deprecation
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -38,6 +44,24 @@ dask = ["dask[dataframe]>=2024.8"]
 duckdb = ["duckdb>=1.0"]
 ibis = ["ibis-framework>=6.0.0", "rich", "packaging", "pyarrow_hotfix"]
 sqlframe = ["sqlframe>=3.22.0"]
+core = [
+  "duckdb",
+  "pandas",
+  "polars",
+  "pyarrow",
+  "sqlframe"
+]
```
MarcoGorelli, line 53: should we move `core` over to `dependency-groups` too?

EdAbati:
> I had no idea you could use `dependency-groups` with `uv pip` 🤯

It was just [released in `uv`](https://github.com/astral-sh/uv/releases/tag/0.6.7) `0.6.7`. I was looking forward to this 😅

.github/workflows/pytest.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: install-reqs
         # Python3.8 is technically at end-of-life, so we don't test everything
-        run: uv pip install -e ".[tests, core]" backports.zoneinfo --system
+        run: uv pip install -e ".[pandas,polars,pyarrow]" backports.zoneinfo --group tests --system
```
EdAbati, line 31: Only installing what is tested (not all `core`) (`constructors=pandas,pyarrow,polars[eager],polars[lazy]`)

dangotbanned:
> > I had no idea you could use `dependency-groups` with `uv pip` 🤯
> 
> It was just [released in `uv`](https://github.com/astral-sh/uv/releases/tag/0.6.7) `0.6.7`. I was looking forward to this 😅

Ah that's great!

This should make a future transition to the [projects](https://docs.astral.sh/uv/concepts/projects/) interface much easier.
I've been interesting in moving to `[dependency-groups]` for `altair` - this PR could serve as a nice example of that 🙂 
- https://github.com/vega/altair/issues/3725

.github/workflows/downstream_tests.yml
```diff
@@ -228,14 +228,16 @@ jobs:
...
+          pdm add ./..
       - name: show-deps
         run: |
           cd tea-tasting
           pdm list
       - name: run-pytest
         run: |
           cd tea-tasting
+          # empty pytest.ini to avoid pytest using narwhals configs
+          touch pytest.ini
```
EdAbati, line 240: lol this was fun to debug. 😅
`tea-tasting` doesn't have pytest configs in the `pyproject.toml` so `pytest` was using the `narwhals` config (and therefore requiring narwhals testing dependencies)

Can anyone think of a cleaner way to do this?

pyproject.toml
```diff
@@ -38,6 +44,29 @@ dask = ["dask[dataframe]>=2024.8"]
...
+core-tests = [
+  {include-group = "core"},
+  {include-group = "tests"},
+]
+local-dev = [
+  {include-group = "dev"},
+  {include-group = "core"},
+  {include-group = "docs"}
+]
```
dangotbanned, line 66: **This is a question, not a suggestion**

Is declaring dependent groups *before* their dependencies a common pattern?

I looked at this for a while before realizing `tests` and `typing` were declared as a later group

MarcoGorelli:
@EdAbati it's occurred to me that this is going to break cuDF's tests against Narwhals

Their CI has broken a few times because of Narwhals, it's probably good if we can keep it more stable

They currently do `install -U -e ".[dev]"` and then run tests

If there's anything missing, could we make a PR to update their CI please? Here's their CI file https://github.com/rapidsai/cudf/blob/branch-25.04/ci/test_narwhals.sh

PR #2241, opened 2025-03-18
[WIP/rolling] test: Scope dataframe library imports to fixtures in `conftest.py`

mgorny:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1726
- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Restarting #1896 from scratch. I'm thinking of working on this as a "rolling" pull request. I'll be doing atomic changes whenever possible, and push them early. Feel free to review and comment anytime, or even possibly merge smaller batches as you see fit. While this won't make it possible to run the whole test suite without a particular dependency until it's all done, it shouldn't cause any regressions for the existing use cases — and have some improvements for setups without dependencies (e.g. the first changes from immediate failure without `polars` to being able to run 924 tests).

My aim is to make at least `duckdb`, `polars` and `pyarrow` optional. However, I also went for making `pandas` optional whenever that was a low-hanging fruit as well. If you disagree with that, please let me know and I'll move `pandas` back to global scope.


MarcoGorelli:
definitely in favour, please ping when ready for review and i can ship it

mgorny:
@MarcoGorelli, thanks! I'm going to take a break now, so please review if you have a minute :-).

dangotbanned:
I've only had a glance, but it looked like there's lots of `try-import-except-pass`

@mgorny you should be able to simplify that pattern by using https://docs.pytest.org/en/stable/reference/reference.html#pytest-importorskip

If you need to be able to continue parts of a test, then an alternative ishttps://docs.python.org/3.12/library/importlib.html#importlib.util.find_spec

The pattern there would be:
```py

if find_spec("polars"):
    import polars
    # do polars stuff
else:
    ... 

# continue the rest of the test

**Review by MarcoGorelli — COMMENTED** [review]
thanks @mgorny ! Generally looks good 👍 

My only comment would be - let's split out the tests rather than doing all these `try/except/pass`

For `test_is_into_dataframe`, for example:
```python
def test_is_into_dataframe_pl() -> None:
    # test Polars using importorskip
    ...

def test_is_into_dataframe_pa() -> None:
    # test PyArrow using importorskip
    ...
```

This is already a fairly large PR (32 files), feel free to keep them smaller for fast merges 🚀

mgorny:
It is green!

mgorny:
Thanks!

Issue #2253, opened 2025-03-19
feat: add support for `replace_strict` for SparkLike


lucas-nelson-uiuc:
Hey everyone,

Noticed that `replace_strict` is not listed in the not-implemented section of the expr script:

https://github.com/narwhals-dev/narwhals/blob/a2796344571f8efe68306ef4cb263ad11980e0ff/narwhals/_spark_like/expr.py#L621-L631

I know we talked about holding off in #1724 - just wanted to check on how we're feeling about that now?

Happy to put together a basic version that doesn't implement any window/collect expressions. Lmk!

FBruzzesi:
Hey @lucas-nelson-uiuc, thanks for the issue! Yes this is definitly something we want to support 👌

FBruzzesi:
Closed by #3282 (with the requirement of passing a default value). I don't expect that we can do better than that

PR #2254, opened 2025-03-20
Spark expr replace strict

lucas-nelson-uiuc:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2253 

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Currently supports replacing one value with another value and specifying `return_dtype`. See comments for issue I'm running into.

### DIFF
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -496,21 +496,17 @@ class SparkLikeExpr(LazyExpr["SparkLikeLazyFrame", "Column"]):
 
         def _replace_strict(
             _input: Column,
-            old: Column,  # should be Sequence[Any]
-            new: Column,  # should be Sequence[Any]
+            old: Column,  # or a Sequence[Column]
+            new: Column,  # or a Sequence[Column]
         ) -> Column:
-            mapper = self._F.create_map(old, new)
-            return mapper[_input]
+            mapping_expr = self._F.create_map(old, new)
+            return mapping_expr[_input]
 
-        result = self._from_call(
-            _replace_strict,
-            "replace_strict",
-            old=old,
-            new=new,
-        )
+        result = self._from_call(_replace_strict, "replace_strict", old=old, new=new)
 
         if return_dtype is not None:
             result = result.cast(return_dtype)
+
         return result
 
     def round(self: Self, decimals: int) -> Self:
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -486,7 +486,7 @@ class SparkLikeExpr(LazyExpr["SparkLikeLazyFrame", "Column"]):
         old: ExprKind | Sequence[Any] | Mapping[Any, Any],
         new: ExprKind | Sequence[Any] | None = None,
         default: ExprKind | None = None,
-        return_dtype: DType | None = None,
+        return_dtype: DType | type[DType] | None = None,
     ) -> Self:
         if new is None:
             if not isinstance(old, Mapping):
@@ -501,9 +501,9 @@ class SparkLikeExpr(LazyExpr["SparkLikeLazyFrame", "Column"]):
             old: Sequence[Any],
             new: Sequence[Any],
             default: ExprKind,
-            return_dtype: DType,
+            return_dtype: DType | type[DType],
         ) -> Column:
-            mapper = self._F.create_map(chain(*zip(old, new)))
+            mapper = self._F.create_map(*chain(*zip(old, new)))
             mapping_expr = mapper[_input]
 
             if default is not None:
--- a/narwhals/_spark_like/expr.py
+++ b/narwhals/_spark_like/expr.py
@@ -484,6 +484,7 @@ class SparkLikeExpr(LazyExpr["SparkLikeLazyFrame", "Column"]):
         self: Self,
         old: ExprKind | Sequence[Any] | Mapping[Any, Any],
         new: ExprKind | Sequence[Any] | None = None,
+        return_dtype: DType | type[DType] | None = None,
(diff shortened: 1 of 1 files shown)


narwhals/_spark_like/expr.py
```diff
@@ -479,6 +480,35 @@ def _len(_input: Column) -> Column:
...
+
+        def _replace_strict(
+            _input: Column,
+            old: Column,  # or a Sequence[Column]
+            new: Column,  # or a Sequence[Column]
+        ) -> Column:
+            mapping_expr = self._F.create_map(old, new)
+            return mapping_expr[_input]
+
+        result = self._from_call(_replace_strict, "replace_strict", old=old, new=new)
```
lucas-nelson-uiuc, line 505: Most likely confusing myself with this `self._from_call`...

The approach I'm taking is to use `pyspark.sql.functions.create_map` to create a mapping expression for us. Currently, it only accepts one constant for `old` and another constant for `new`.

Running into two issues when trying to expand the functionality:
- if the user passes a sequence of values to replace, `self._from_call` will convert `old` and `new` into array columns; instead, I want `old` and `new` to be a sequence of columns by the time we access it in `_replace_strict` so that we can use it with `create_map`
- this approach doesn't allow us to pass expressions to `old` or `new`

dangotbanned:
@lucas-nelson-uiuc I was a bit confused when I saw (#2253)

So right now `replace_strict` is declared as `not_implemented` for **all lazy backends** here:

https://github.com/narwhals-dev/narwhals/blob/a2796344571f8efe68306ef4cb263ad11980e0ff/narwhals/_compliant/expr.py#L827-L847

That is why you're seeing `mypy` yelling in (https://github.com/narwhals-dev/narwhals/actions/runs/13960816458/job/39081638109?pr=2254)

However, if `replace_strict` is **possible** for `SparkLikeExpr` - then you'll need to move this:

https://github.com/narwhals-dev/narwhals/blob/a2796344571f8efe68306ef4cb263ad11980e0ff/narwhals/_compliant/expr.py#L846

Into each of these sections:

https://github.com/narwhals-dev/narwhals/blob/a2796344571f8efe68306ef4cb263ad11980e0ff/narwhals/_dask/expr.py#L632-L634

https://github.com/narwhals-dev/narwhals/blob/a2796344571f8efe68306ef4cb263ad11980e0ff/narwhals/_duckdb/expr.py#L488-L502

I hope that all makes sense, but let me know if you have any issues 🙂

narwhals/_spark_like/expr.py
MarcoGorelli: thanks for your pr!

i don't think we currently accept expressions here for other backends anyway, so I think we don't need to support that just yet

narwhals/_spark_like/expr.py
```diff
@@ -560,6 +562,65 @@ def _len(_input: Column) -> Column:
...
+
+            mapped_col = (
+                mapping_expr[_input]
+                if self._implementation.is_pyspark()
+                else mapping_expr.getItem(_input)
+            )
+
+            try:
+                result = self._F.when(validation_expr, mapped_col).otherwise(
+                    self._F.assert_true(self._F.lit(False))  # noqa: FBT003
```
FBruzzesi, line 607: SQLFrame does not have `assert_true` for duckdb backend, that's why I am still xfailing it in tests

narwhals/_spark_like/expr.py
```diff
@@ -560,6 +562,65 @@ def _len(_input: Column) -> Column:
...
+                        self._F.array([self._F.lit(k) for k in non_null_keys]), _input
+                    )
+                )
+            )
+
+            mapped_col = (
+                mapping_expr[_input]
+                if self._implementation.is_pyspark()
+                else mapping_expr.getItem(_input)
+            )
```
FBruzzesi, line 603: PySpark raises a warning for `mapping_expr.getItem(_input)`, while SQLFrame raises an exception for `mapping_expr[_input]` (Column is not callable)

narwhals/_spark_like/expr.py
```diff
@@ -560,6 +562,65 @@ def _len(_input: Column) -> Column:
...
+                    _input.isNull() & self._F.lit(has_null_key),
+                    self._F.lit(True),  # noqa: FBT003
+                )
+                .when(_input.isNull() & ~self._F.lit(has_null_key), self._F.lit(False))  # noqa: FBT003
+                .otherwise(
+                    self._F.array_contains(
+                        self._F.array([self._F.lit(k) for k in non_null_keys]), _input
+                    )
+                )
+            )
```
FBruzzesi, line 597: I might need an extra look at this

narwhals/_spark_like/expr.py
```diff
@@ -560,6 +562,65 @@ def _len(_input: Column) -> Column:
...
+                else mapping_expr.getItem(_input)
+            )
+
+            try:
+                result = self._F.when(validation_expr, mapped_col).otherwise(
+                    self._F.assert_true(self._F.lit(False))  # noqa: FBT003
+                )
+            except Exception as exc:
+                msg = "replace_strict did not replace all non-null values."
+                raise ValueError(msg) from exc
```
FBruzzesi, line 611: Also, I am not really able to capture any spark exception at this level

FBruzzesi (commented):
I pushed a commit with (maybe) some improvement and a step for validation.
I think that `replace` (non-strict) might be much easier 😂 not really sure why we started with the strict version

PR #2262, opened 2025-03-21
chore: Add `CompliantExpr._with_metadata`

MarcoGorelli:
This avoid the need to have `kind` as an argument in `over`, and also gives us a way to use `ExprMetadata` at the compliant level in other places (e.g. `group_by.agg`), so we can get rid of tracking `function_name` at the compliant level

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


MarcoGorelli:
@dangotbanned I suspect (and have good reason to believe) that you can probably think of a better way to do this, in which case, feel free to add a commit to do so 🙏

dangotbanned:
> @dangotbanned I suspect (and have good reason to believe) that you can probably think of a better way to do this, in which case, feel free to add a commit to do so 🙏

Thanks for the complimentary ping @MarcoGorelli 😍

I'll take a look later today

dangotbanned:
I'm just looking from my phone, apologies for the stray comment.

Would it make sense to add a `with_metadata` implementation for `EagerExpr` and/or `LazyExpr`?

You can always override it if there's something unique for a given backend (e.g. `window_function`).
I'd *think* there'd be some potential for reuse on that front

narwhals/expr.py
```diff
@@ -1583,9 +1583,9 @@ def over(
...
 
         return self.__class__(
-            lambda plx: self._to_compliant_expr(plx).over(
-                flat_partition_by, order_by=order_by, kind=self._metadata.kind
-            ),
+            lambda plx: self._to_compliant_expr(plx)
+            .with_metadata(self._metadata)
+            .over(flat_partition_by, order_by=order_by),
             metadata,
         )
```
dangotbanned, line 1590: @MarcoGorelli 

Is it important that `with_metadata` uses `self._metadata` but the new `Expr` uses `metadata`?

If so, I'd probably alias both of them before this to something like:

```py
current_meta = self._metadata
next_meta = ExprMetadata(
    kind,
    n_open_windows=n_open_windows,
    is_multi_output=current_meta.is_multi_output,
)
```

Then the two are more clearly distinct in the return:

```py
return self.__class__(
    lambda plx: self._to_compliant_expr(plx)
    .with_metadata(current_meta)
    .over(flat_partition_by, order_by=order_by),
    next_meta,
)
```

MarcoGorelli: looks good!

MarcoGorelli: > Is it important that with_metadata uses self._metadata but the new Expr uses metadata?

yup, very important, so your distinction between "current" and "next" probably helps with readability 👍

narwhals/_arrow/expr.py
```diff
@@ -52,6 +52,21 @@ def __init__(
...
+            function_name=self._function_name,
+            evaluate_output_names=self._evaluate_output_names,
+            alias_output_names=self._alias_output_names,
+            backend_version=self._backend_version,
+            version=self._version,
+            depth=self._depth,
+            call_kwargs=self._call_kwargs,
+        )
+        expr._metadata = metadata
+        return expr
```
dangotbanned, line 69: This kinda applies to all of them, just picking the first impl.

Was it intentional that there isn't a way to preserve `_metadata`?

I know that there's more going on at the `nw.Expr` level -- but if I call any method method *after* `.with_metadata` it gets lost right?

dangotbanned: Since `CompliantExpr.__init__` always resets to `None`

MarcoGorelli: yup, that's right - I think that's correct, because further operations might change the metadata. e.g. if you have `nw.col('a').abs()` and then do `.drop_nulls()`, then the `ExprKind` changes

dangotbanned:
Apologies for all the questions before doing any commits 😅 

Wanna make sure any changes I make are preserving your intent

narwhals/_arrow/expr.py
MarcoGorelli: I think most of the time, accessing the metadata from the compliant level isn't necessary, so my thinking was that it should be ok to just call `with_metadata` from the narwhals level when needed

not against always passing around the metadata at the compliant level if that's possible though

dangotbanned: Got it, thanks @MarcoGorelli

narwhals/expr.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2262/commits/eddcaa34b318a40c20ddd4d85c293d5b1fe56573)

dangotbanned (approved):
LGTM, thanks @MarcoGorelli 

All 3 of the `LazyExpr` constructors differ - so I only did the `EagerExpr` part of (https://github.com/narwhals-dev/narwhals/pull/2262#issuecomment-2743779041)

Will leave it to you for merging/making further tweaks 🙌

MarcoGorelli:
awesome, thanks @dangotbanned !

Issue #2174, opened 2025-03-09
implement remaining order-dependent lazy operations


MarcoGorelli:
https://github.com/narwhals-dev/narwhals/pull/2168 implements `rolling_sum` for lazy backends

We should do the same for other order-dependent operations:
- [x] all other rolling_* operations
- [x] all other cum_* operations
- [x] diff
- [x] shift
- [ ] rank
- [x] is_first_distinct
- [x] is_last_distinct

Contributions towards the above would be extremely welcome. Feel free to use #2168 as a template

You'll probably need to add some xfails / skips in some cases. See https://github.com/narwhals-dev/narwhals/issues/1893 for some heuristics for which to use and when, and feel free to ask here or on discord about anything

DeaMariaLeon:
I would like to work on `rolling_mean`.  (Or did you mean **all** rolling at once??)

MarcoGorelli:
😄 one at a time is fine!

marenwestermann:
Then I'll start with `rolling_var`. 🙂

marvinl803:
I would like to work on rolling_std.

DeaMariaLeon:
Believe it or not I was working on `rolling_mean` @raisadz. 

Pinging @marenwestermann, it seems that the one function you took has also been done. Just FYI so you don't waste time like I did this morning.

MarcoGorelli:
ah sorry it had been 3 weeks so we went ahead with the pr

raisadz:
@DeaMariaLeon I would appreciate it if you don't speak with me (or other contributors) like this in the future. It's been three weeks since you started and there were no drafts so I wanted to complete this issue as it was marked as high priority.

DeaMariaLeon:
That's fine @MarcoGorelli. I just wanted to let Maren know.

DeaMariaLeon:
I am sorry for deleting your last comment @raisadz. I was just very offended. 

Obviously my comment before that sounded offensive to you, so I'm also sorry for that. 

Finally,  I can't restore the comment, github doesn't let me.

MarcoGorelli:
No worries, thanks for apologising!

It's very hard to tell people's tone from online comments, and I don't think either of you wrote or intended anything to be offensive - my general advice is, if in doubt, put one extra emoji rather than one fewer 😄 💡 ❤

I think this is the first mishap we've had in the wonderful Narwhals community, let's make sure it's the last 🚀 

[...]

PR #2263, opened 2025-03-21
feat: support window operations for DuckDB

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -85,7 +85,7 @@ jobs:
       - name: install-reqs
         run: uv pip install -e ".[modin, dask]" --group core-tests --group extra --system
       - name: install duckdb nightly
-        run: uv pip install -U --pre duckdb
+        run: uv pip install -U --pre duckdb --system
       - name: install pyspark
         run: uv pip install -e ".[pyspark]" --system
         # PySpark is not yet available on Python3.12+
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -84,6 +84,8 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: install-reqs
         run: uv pip install -e ".[modin, dask]" --group core-tests --group extra --system
+      - name: install duckdb nightly
+        run: uv pip install -U --pre duckdb
       - name: install pyspark
         run: uv pip install -e ".[pyspark]" --system
         # PySpark is not yet available on Python3.12+
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -21,6 +21,7 @@ from narwhals._duckdb.expr_list import DuckDBExprListNamespace
 from narwhals._duckdb.expr_name import DuckDBExprNameNamespace
 from narwhals._duckdb.expr_str import DuckDBExprStringNamespace
 from narwhals._duckdb.expr_struct import DuckDBExprStructNamespace
+from narwhals._duckdb.utils import generate_partition_by_sql
 from narwhals._duckdb.utils import lit
 from narwhals._duckdb.utils import maybe_evaluate_expr
 from narwhals._duckdb.utils import narwhals_to_native_dtype
@@ -487,12 +488,11 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
                     for expr in self._call(df)
                 ]
         else:
+            partition_by_sql = generate_partition_by_sql(*partition_by)
 
             def func(df: DuckDBLazyFrame) -> list[duckdb.Expression]:
                 return [
-                    SQLExpression(
-                        f"{expr} over (partition by {', '.join([f'"{x}"' for x in partition_by])})"
-                    )
+                    SQLExpression(f"{expr} over {partition_by_sql}")
                     for expr in self._call(df)
                 ]
 
@@ -542,12 +542,7 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
                 order_by_sql = "order by " + ", ".join(
                     f'"{x}" asc nulls first' for x in order_by
                 )
-            if partition_by:
-                partition_by_sql = "partition by " + ",".join(
-                    f'"{x}"' for x in partition_by
-                )
-            else:
-                partition_by_sql = ""
+            partition_by_sql = generate_partition_by_sql(*partition_by)
             sql = f"sum ({_input}) over ({partition_by_sql} {order_by_sql} rows between unbounded preceding and current row)"
             return SQLExpression(sql)
 
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -492,7 +492,7 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
 
             def func(df: DuckDBLazyFrame) -> list[duckdb.Expression]:
                 return [
-                    SQLExpression(f"{expr} over {partition_by_sql}")
+                    SQLExpression(f"{expr} over ({partition_by_sql})")
                     for expr in self._call(df)
                 ]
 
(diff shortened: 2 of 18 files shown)


narwhals/_duckdb/expr.py
```diff
@@ -8,6 +8,7 @@
 from typing import Sequence
 from typing import cast
 
+import duckdb
```
FBruzzesi, line 11: Should we follow the same pattern as below and directly import `from duckdb import SQLExpression`?

narwhals/_duckdb/expr.py
```diff
@@ -447,6 +512,61 @@ def round(self: Self, decimals: int) -> Self:
...
+                    f'"{x}" asc nulls first' for x in order_by
+                )
+            if partition_by:
+                partition_by_sql = "partition by " + ",".join(
+                    f'"{x}"' for x in partition_by
+                )
+            else:
+                partition_by_sql = ""
+            sql = f"sum ({_input}) over ({partition_by_sql} {order_by_sql} rows between unbounded preceding and current row)"
+            return duckdb.SQLExpression(sql)
```
FBruzzesi, line 536: I can see a lot of re-usability if everything here for #2174 - only "word" to input is "sum" and other operations 🙌🏼

.github/workflows/pytest.yml
```diff
@@ -84,6 +84,8 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: install-reqs
         run: uv pip install -e ".[modin, dask]" --group core-tests --group extra --system
+      - name: install duckdb nightly
+        run: uv pip install -U --pre duckdb --system
```
FBruzzesi, line 88: Should we wait for the duckdb release before merging this PR? You are ahead of them 😂

FBruzzesi (commented):
@MarcoGorelli this was an amazing streaming! I left two nitpick comments and a suggestion for whoever wants to have some fun working on #2174 🥦

.github/workflows/pytest.yml
MarcoGorelli: nah let's be one step ahead so when they release we're ready 🔥

narwhals/_duckdb/expr.py
```diff
@@ -461,6 +528,56 @@ def round(self: Self, decimals: int) -> Self:
...
 
+    def cum_sum(self, *, reverse: bool) -> Self:
+        def func(
+            _input: duckdb.Expression,
+            partition_by: Sequence[str],
+            order_by: Sequence[str],
+        ) -> duckdb.Expression:
+            if reverse:
+                order_by_sql = "order by " + ", ".join(
+                    f'"{x}" desc nulls last' for x in order_by
```
dangotbanned, line 539: @MarcoGorelli I commented this on the stream, not sure it was ever tried out, anyways now in context 🙂 
```suggestion
                    f'{x!r} desc nulls last' for x in order_by
```

- https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals

narwhals/_duckdb/expr.py
```diff
@@ -438,6 +471,40 @@ def null_count(self: Self) -> Self:
...
 
+    def over(
+        self: Self,
+        partition_by: Sequence[str],
+        order_by: Sequence[str] | None,
+    ) -> Self:
+        if self._backend_version < (1, 3):
+            msg = "At least version 1.3 of DuckDB is required for `over` operation."
+            raise NotImplementedError(msg)
+        if (window_function := self._window_function) is not None:
```
dangotbanned, line 482: AFAIK, you only need the `is not None` when the type overrides `__bool__`. 

Going to guess this is from experience battling `pandas`, `polars`, and `numpy` who all tell you off for using `__bool__`

```suggestion
        if (window_function := self._window_function):
```

narwhals/_duckdb/typing.py
```diff
@@ -0,0 +1,16 @@
...
+if TYPE_CHECKING:
+    import duckdb
+
+    class WindowFunction(Protocol):
+        def __call__(
+            self,
+            _input: duckdb.Expression,
+            partition_by: Sequence[str],
+            order_by: Sequence[str],
+        ) -> duckdb.Expression: ...
```
dangotbanned, line 16: I've been meaning to ask about this for a while now.

Is that naming convention *intended* to signal *positional-only* for `_input`?

It seems similar to a [convention](https://peps.python.org/pep-0570/#using-as-a-per-argument-marker) that was common *before* [PEP 570 – Python Positional-Only Parameters](https://peps.python.org/pep-0570/).

### Current
```py
    class WindowFunction(Protocol):
        def __call__(
            self,
            _input: duckdb.Expression,
            partition_by: Sequence[str],
            order_by: Sequence[str],
        ) -> duckdb.Expression: ...
```

### Before PEP 570
```py
    class WindowFunction(Protocol):
        def __call__(
            self,
            __input: duckdb.Expression,
            partition_by: Sequence[str],
            order_by: Sequence[str],
        ) -> duckdb.Expression: ...
```

### After PEP 570
```py
    class WindowFunction(Protocol):
        def __call__(
            self,
            input: duckdb.Expression,
            /,
            partition_by: Sequence[str],
            order_by: Sequence[str],
        ) -> duckdb.Expression: ...
```

narwhals/_duckdb/expr.py
MarcoGorelli: i don't think that's the same, we need double quotes around the column names as that's what sql expects (single quotes are string literals)

narwhals/_duckdb/typing.py
MarcoGorelli: yeah `_input` wasn't a very good name, i don't think i'd put much thought into it at the time and now it's all over

we could use `native_expr` in its place?

narwhals/_duckdb/expr.py
dangotbanned: > i don't think that's the same, we need double quotes around the column names as that's what sql expects (single quotes are string literals)

Does it not respect the outer quotes?

I would've thought these two used different internal ones:

```py
f'{x!r} desc nulls last' for x in order_by
f"{x!r} desc nulls last" for x in order_by
```

narwhals/_duckdb/typing.py
dangotbanned: > we could use `native_expr` in its place?

Whatever you feel works 👍 

### Side note
One of the cool things about *positional-only* args is that you can use different names - and still work the same at runtime and to a type checker.
Was a bit of a 🤯 when I learned that

MarcoGorelli:
thanks all for your reviews!

tests/expr_and_series/str/to_datetime_test.py
```diff
@@ -216,15 +216,12 @@ def test_to_datetime_tz_aware(
         pytest.skip()
     if is_pyarrow_windows_no_tzdata(constructor):
         pytest.skip()
-    if "sqlframe" in str(constructor):
-        # https://github.com/eakmanrq/sqlframe/issues/325
-        request.applymarker(pytest.mark.xfail)
```
dangotbanned, line 221: @MarcoGorelli  I'm also needing this since the PR merged 🤔 

The comment is the error that started showing up for me, but not in CI

```diff
diff --git a/tests/expr_and_series/str/to_datetime_test.py b/tests/expr_and_series/str/to_datetime_test.py
index 412485d0..4ed208f6 100644
--- a/tests/expr_and_series/str/to_datetime_test.py
+++ b/tests/expr_and_series/str/to_datetime_test.py
@@ -14,6 +14,7 @@ from tests.utils import PANDAS_VERSION
 from tests.utils import PYARROW_VERSION
 from tests.utils import assert_equal_data
 from tests.utils import is_pyarrow_windows_no_tzdata
+from tests.utils import is_windows
 
 if TYPE_CHECKING:
     from tests.utils import Constructor
@@ -219,6 +220,15 @@ def test_to_datetime_tz_aware(
     if "cudf" in str(constructor):
         # cuDF does not yet support timezone-aware datetimes
         request.applymarker(pytest.mark.xfail)
+    if "sqlframe" in str(constructor) and format is not None and is_windows():
+        #
+        # E       duckdb.duckdb.InvalidInputException: Invalid Input Error:
+        #         Could not parse string "2020-01-01 01:02:03+0100" according to format specifier "%Y-%-m-%-d %-H:%-M:%-SZ"
+        # E       2020-01-01 01:02:03+0100
+        # E                           ^
+        # E       Error: Literal does not match, expected Z
+        #
+        request.applymarker(pytest.mark.xfail)
     context = (
         pytest.raises(NotImplementedError)
         if any(x in str(constructor) for x in ("duckdb",)) and format is None


```

PR #2280, opened 2025-03-24
fix: Expr.mode broadcasting

### LINKED ISSUE #2273: [Bug]: `Expr.mode` inconsistent output when applied to multiple columns
### Describe the bug

When applying expr.mode() on multiple columns (e.g. `df.select(nw.all().mode())`) and one returns one and another multiple values the output is incosistent. This is due to to an issue on the `polars` side of the operation. 

In `polars` this is dealt by using `df.select(nw.all().mode().first())` (https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.mode.html).

```python

import polars as pl
import pandas as pd
import narwhals as nw
from narwhals.typing import IntoFrame

data = {
    "a": [1, 1, 2, 2, 3],
    "b": [1, 2, 3, 3, 4],
}

df_pl = pl.DataFrame(data)
df_pd = pd.DataFrame(data)

def agnostic_mode(df_native: IntoFrame) -> IntoFrame:
    df = nw.from_native(df_native)
    return df.select(nw.all().mode())

agnostic_mode(df_pd)

agnostic_mode(df_pl)
```

### Expected results

```python
agnostic_mode(df_pd)
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|        a  b      |
|     0  1  3      |
|     1  2  3      |
└──────────────────┘

agnostic_mode(df_pl)
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|        a  b      |
|     0  1  3      |
|     1  2  3      |
└──────────────────┘
```

### Actual results

```python
agnostic_mode(df_pd)
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|        a  b      |
|     0  1  3      |
|     1  2  3      |
└──────────────────┘

agnostic_mode(df_pl)
  File "/home/kstelios/Projects/narwhals-dev/narwhals/dataframe.py", line 1321, in select
    return super().select(*exprs, **named_exprs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kstelios/Projects/narwhals-dev/narwhals/dataframe.py", line 180, in select
    self._compliant_frame.select(*compliant_exprs),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kstelios/Projects/narwhals-dev/narwhals/_polars/dataframe.py", line 171, in func
    raise catch_polars_exception(e, self._backend_version) from None
narwhals.exceptions.ShapeError: Series b, length 1 doesn't match the DataFrame height of 2

If you want expression: col("b").mode() to be broadcasted, ensure it is a scalar (for instance by adding '.first()').
```

```shell
System:
    python: 3.12.7 (main, Oct 16 2024, 04:37:19) [Clang 18.1.8 ]
executable: /home/kstelios/Projects/narwhals-dev/.venv/bin/python3
   machine: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 1.31.0
       pandas: 2.2.3
       polars: 1.24.0
         cudf:
        modin: 0.32.0
      pyarrow: 19.0.1
        numpy: 2.2.3
```

### Relevant log output

```shell

```

MarcoGorelli:
closes #2273

This is definitely not THE solution, we need to find a better way of preserving the `_broadcast` attribute in some cases

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -132,13 +132,23 @@ class ArrowSeries(EagerSeries["ArrowChunkedArray"]):
     def _from_native_series(
         self: Self,
         series: ArrowArray | ArrowChunkedArray,
+        *,
+        preserve_broadcast: bool = False,
     ) -> Self:
-        return self.__class__(
+        # In cases when operations are known to not affect whether a result should
+        # be broadcast, we can pass `preverse_broadcast=True`.
+        # Set this with care - it should only be set for unary expressions which don't
+        # change length or order, such as `.alias` or `.fill_null`. If in doubt, don't
+        # set it, you probably don't need it.
+        result = self.__class__(
             chunked_array(series),
             name=self._name,
             backend_version=self._backend_version,
             version=self._version,
         )
+        if preserve_broadcast:
+            result._broadcast = self._broadcast
+        return result
 
     @classmethod
     def from_iterable(
@@ -435,12 +445,14 @@ class ArrowSeries(EagerSeries["ArrowChunkedArray"]):
         return self.native.to_numpy()
 
     def alias(self: Self, name: str) -> Self:
-        return self.__class__(
+        result = self.__class__(
             self.native,
             name=name,
             backend_version=self._backend_version,
             version=self._version,
         )
+        result._broadcast = self._broadcast
+        return result
 
     @property
     def dtype(self: Self) -> DType:
@@ -505,14 +517,16 @@ class ArrowSeries(EagerSeries["ArrowChunkedArray"]):
         return self._from_native_series(res)
 
     def is_null(self: Self) -> Self:
-        return self._from_native_series(self.native.is_null())
+        return self._from_native_series(self.native.is_null(), preserve_broadcast=True)
 
     def is_nan(self: Self) -> Self:
-        return self._from_native_series(pc.is_nan(self.native))
+        return self._from_native_series(pc.is_nan(self.native), preserve_broadcast=True)
 
     def cast(self: Self, dtype: DType | type[DType]) -> Self:
         data_type = narwhals_to_native_dtype(dtype, self._version)
-        return self._from_native_series(pc.cast(self.native, data_type))
+        return self._from_native_series(
+            pc.cast(self.native, data_type), preserve_broadcast=True
+        )
 
     def null_count(self: Self, *, _return_py_scalar: bool = True) -> int:
         return maybe_extract_py_scalar(self.native.null_count, _return_py_scalar)
@@ -655,10 +669,7 @@ class ArrowSeries(EagerSeries["ArrowChunkedArray"]):
             series = fill_func(self.native)
         else:
             series = fill_aux(self.native, limit, strategy)
-        result = self._from_native_series(series)
-        if self._broadcast:
-            result._broadcast = True
-        return result
+        return self._from_native_series(series, preserve_broadcast=True)
 
     def to_frame(self: Self) -> ArrowDataFrame:
         from narwhals._arrow.dataframe import ArrowDataFrame
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -28,6 +28,7 @@ from narwhals._arrow.utils import native_to_narwhals_dtype
 from narwhals._arrow.utils import nulls_like
 from narwhals._arrow.utils import pad_series
 from narwhals._compliant import EagerSeries
+from narwhals._expression_parsing import ExprKind
 from narwhals.dependencies import is_numpy_array_1d
 from narwhals.exceptions import InvalidOperationError
 from narwhals.utils import Implementation
@@ -654,7 +655,10 @@ class ArrowSeries(EagerSeries["ArrowChunkedArray"]):
             series = fill_func(self.native)
         else:
             series = fill_aux(self.native, limit, strategy)
-        return self._from_native_series(series)
+        result = self._from_native_series(series)
+        if self._broadcast:
+            result._broadcast = True
+        return result
 
     def to_frame(self: Self) -> ArrowDataFrame:
         from narwhals._arrow.dataframe import ArrowDataFrame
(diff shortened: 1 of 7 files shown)


narwhals/_arrow/series.py
```diff
@@ -654,7 +655,10 @@ def fill_aux(
             series = fill_func(self.native)
         else:
             series = fill_aux(self.native, limit, strategy)
-        return self._from_native_series(series)
+        result = self._from_native_series(series)
+        if self._broadcast:
+            result._broadcast = True
+        return result
```
MarcoGorelli, line 661: definitely not the right solution

perhaps `_from_native_series` needs a "preserve_broadcast" boolean argument, which defaults to `False` but which we can selectively set to `True` for functions which we know don't change the length?

MarcoGorelli: @dangotbanned curious to hear your opinion on this if you have interest/time

dangotbanned: Thanks for the ping @MarcoGorelli!

I haven't taken a look at this yet, but did find the current `._broadcast` stuff a bit odd while working on `when-then-otherwise`.

`Dask` seems to have something that *looks* similar - but I didn't dig too deep into it

dangotbanned: Do you have info in `Expr._metadata` that would be useful to have passed down during `Expr.mode`?

MarcoGorelli: if we could make sure that compliant exprs are always have full `ExprMetadata` available, then maybe we wouldn't even need `_broadcast`

🤔 that should be feasible, and might be simpler

dangotbanned: Nice one @MarcoGorelli - always a bonus if that simplifies things

If you can figure out what's up with `pyspark` - I can take a look at #2282

tests/expr_and_series/mode_test.py
```diff
@@ -12,25 +14,23 @@
...
     expected = {"a": [1, 2]}
     assert_equal_data(result, expected)
 
 
-def test_mode_series(
-    constructor_eager: ConstructorEager, request: pytest.FixtureRequest
-) -> None:
-    if "pyarrow_table" in str(constructor_eager):
-        # TODO(unassigned): reimplement mode for pyarrow
-        request.applymarker(pytest.mark.xfail)
```
MarcoGorelli, line 32: unexpected side-effect of this PR - fixing these two issues with pyarrow's mode 🥳

narwhals/_arrow/series.py
```diff
@@ -131,13 +132,23 @@ def _change_version(self: Self, version: Version) -> Self:
...
+        # set it, you probably don't need it.
+        result = self.__class__(
             chunked_array(series),
             name=self._name,
             backend_version=self._backend_version,
             version=self._version,
         )
+        if preserve_broadcast:
+            result._broadcast = self._broadcast
+        return result
```
dangotbanned, line 151: > [!IMPORTANT]
> Haven't had a chance to review today, just an isolated comment

Would it not make more sense to document this in `EagerSeries`?

https://github.com/narwhals-dev/narwhals/blob/9da549e86f83c96df19ab71916e7d4335c89c1e3/narwhals/_compliant/series.py#L273-L281

The signature from `CompliantSeries._from_native_series` is being overridden in the same way here and in `PandasLikeSeries`

https://github.com/narwhals-dev/narwhals/blob/9da549e86f83c96df19ab71916e7d4335c89c1e3/narwhals/_compliant/series.py#L67

If you wanted to get real funky, you could do something like `EagerSeries._from_scalar`

https://github.com/narwhals-dev/narwhals/blob/9da549e86f83c96df19ab71916e7d4335c89c1e3/narwhals/_compliant/series.py#L280-L281

https://github.com/narwhals-dev/narwhals/blob/9da549e86f83c96df19ab71916e7d4335c89c1e3/narwhals/_arrow/series.py#L153-L156

narwhals/_arrow/series.py
dangotbanned: > if we could make sure that compliant exprs are always have full `ExprMetadata` available, then maybe we wouldn't even need `_broadcast`
> 
> 🤔 that should be feasible, and might be simpler

@MarcoGorelli did you run into an issue with that idea?

I wasn't expecting to see `._broadcast` appearing in more places

narwhals/_arrow/series.py
dangotbanned: Moved the doc in (https://github.com/narwhals-dev/narwhals/pull/2280/commits/e8f0e0369987569c18b5ffa1031fa24047447cb8)

narwhals/_arrow/series.py
```diff
@@ -504,14 +512,16 @@ def is_between(
         return self._from_native_series(res)
 
     def is_null(self: Self) -> Self:
-        return self._from_native_series(self.native.is_null())
+        return self._from_native_series(self.native.is_null(), preserve_broadcast=True)
 
     def is_nan(self: Self) -> Self:
-        return self._from_native_series(pc.is_nan(self.native))
+        return self._from_native_series(pc.is_nan(self.native), preserve_broadcast=True)
```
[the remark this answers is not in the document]
dangotbanned: The same methods, but for `nw.Expr`, do not *add* `ExprMetadata` - which makes me think the concern is *unique* to `Series`

MarcoGorelli:
thanks @dangotbanned for your review!

yes, this can probably be moved further up or reworked further - but, I think it's at least a step forwards, and avoids broadcasting based on the length of the results (which is what we did before 😳 )

PR #2282, opened 2025-03-24
chore: always set `CompliantExpr._metadata`

MarcoGorelli:
This could be a step towards simplifying https://github.com/narwhals-dev/narwhals/pull/2280

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/2280#discussion_r2010361846)

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/group_by.py
```diff
@@ -174,9 +174,9 @@ def agg(self: Self, *aggs: Expr | Iterable[Expr], **named_aggs: Expr) -> LazyFra
...
         plx = self._df.__narwhals_namespace__()
         compliant_aggs = (
-            *(x._to_compliant_expr(plx)._with_metadata(x._metadata) for x in flat_aggs),
+            *(x._to_compliant_expr(plx) for x in flat_aggs),
             *(
-                value._to_compliant_expr(plx).alias(key)._with_metadata(value._metadata)
+                value.alias(key)._to_compliant_expr(plx)
                 for key, value in named_aggs.items()
             ),
         )
```
dangotbanned, line 182: Woah - much better

MarcoGorelli:
Cool, CI seems to be resolved 🥳 Some downstream libraries are failing due to the new setuptools release, but I think we can just wait for setuptools to hopefully revert their change and for this to resolve itself

narwhals/group_by.py
MarcoGorelli: 😄 can i interpret this as approval?

narwhals/expr.py
```diff
@@ -50,11 +50,16 @@
...
         metadata: ExprMetadata,
     ) -> None:
         # callable from CompliantNamespace to CompliantExpr
-        self._to_compliant_expr: _ToCompliant = to_compliant_expr
+        def func(plx: CompliantNamespace[Any, Any]) -> CompliantExpr[Any, Any]:
+            result = to_compliant_expr(plx)
+            result._metadata = self._metadata
+            return result
+
+        self._to_compliant_expr: _ToCompliant = func
```
dangotbanned, line 62: I'm a *little bit* confused about this part.

Is there any chance you could walk me through the steps leading up to `to_compliant_expr`?

narwhals/group_by.py
dangotbanned: > 😄 can i interpret this as approval?

*Visually* yeah 🎉 - *conceptually* I need some help with (https://github.com/narwhals-dev/narwhals/pull/2282#discussion_r2010769604)

narwhals/expr.py
dangotbanned: I know everything is somewhat circular - so maybe the cycle between 1 `Expr` to another?

MarcoGorelli: yup, sure! let's see what happens when we make `nw.col('a')`

First, we create `nw.col('a')`, which creates:
```python
Narwhals Expr
metadata: ExprMetadata(kind: ExprKind.TRANSFORM, n_open_windows: 0, expansion_kind: ExpansionKind.SINGLE)
```

To get a pandas-like compliant Expr (`PandasLikeExpr`), you can evaluate `_to_compliant_expr` with a `PandasLikeNamespace`:
```python
In [10]: import pandas as pd
    ...: import narwhals as nw
    ...: from narwhals._pandas_like.namespace import PandasLikeNamespace
    ...: from narwhals._pandas_like.utils import Implementation
    ...: from narwhals.utils import parse_version, Version
    ...: 
    ...: pn = PandasLikeNamespace(
    ...:     implementation=Implementation.PANDAS,
    ...:     backend_version=parse_version(pd.__version__),
    ...:     version=Version.MAIN,
    ...: )

In [11]: nw.col('a')._to_compliant_expr(pn)
Out[11]: PandasLikeExpr(depth=0, function_name=col)

In [12]: nw.col('a')._to_compliant_expr(pn)._metadata
Out[12]: ExprMetadata(kind: ExprKind.TRANSFORM, n_open_windows: 0, expansion_kind: ExpansionKind.SINGLE)
```
Note how the `metadata` from the Narwhals-level was propagated through to `PandasLikeExpr._metadata` - that's what this PR enables.

Similarly:
```python
In [13]: nw.col('a').diff()
Out[13]: 
Narwhals Expr
metadata: ExprMetadata(kind: ExprKind.WINDOW, n_open_windows: 1, expansion_kind: ExpansionKind.SINGLE)

In [14]: nw.col('a').diff()._to_compliant_expr(pn)
Out[14]: PandasLikeExpr(depth=1, function_name=col->diff)

In [15]: nw.col('a').diff()._to_compliant_expr(pn)._metadata
Out[15]: ExprMetadata(kind: ExprKind.WINDOW, n_open_windows: 1, expansion_kind: ExpansionKind.SINGLE)
```

dangotbanned: Thanks @MarcoGorelli that was super helpful!

Give me a few mins so I can reply from *not* my phone 🙂

dangotbanned: Okay AFAICT, following this PR - every `CompliantExpr` should have `CompliantExpr._metadata` set.

However, due to this occurring *after* `CompliantExpr.__init__` it can still be `None`

https://github.com/narwhals-dev/narwhals/blob/e67ae25f46558987234df1c08a1b9d5f5eb0a00d/narwhals/_compliant/expr.py#L81-L87

*Morally*, we know that `._metadata` **should** never be `None`.
But we'll need to prove that to `mypy`/`pyright` on every attribute access like in:

https://github.com/narwhals-dev/narwhals/blob/e67ae25f46558987234df1c08a1b9d5f5eb0a00d/narwhals/_compliant/expr.py#L285

What I'm trying to get at is - would there be a way to get to the same goal **but** avoid having a `None`?

MarcoGorelli: It can still be `None`, because it gets reset when doing operations at the compliant level (because they might change the metadata, so it can only be set "from above")

```python
In [3]: nw.col('a').diff()._to_compliant_expr(pn)._metadata
Out[3]: ExprMetadata(kind: ExprKind.WINDOW, n_open_windows: 1, expansion_kind: ExpansionKind.SINGLE)

In [4]: nw.col('a')._to_compliant_expr(pn).diff()._metadata
```

dangotbanned: > It can still be `None`, because it gets reset when doing operations at the compliant level (because they might change the metadata, so it can only be set "from above")

Is that a desirable property - or the result of how the implementation works *currently*?

dangotbanned (approved):
Thanks @MarcoGorelli 🎉 

I'm approving as-is, but I definitely wanna explore that thread more in the future (https://github.com/narwhals-dev/narwhals/pull/2282#discussion_r2010819556)

narwhals/expr.py
MarcoGorelli: I think it's desirable if metadata resolution only happens at one level, and I think the top level is probably the best place for that to happen, as then it can be done independently of any backend. For example, you can just do `nw.all().is_first_distinct()` and inspect the metadata - among other things, this allows uniform error messages to be raised for certain unsupported combinations of operations

I think maybe we shouldn't have `_broadcast` at all, and that some `ExprMetadata` method can determine whether broadcasting should happen. If the metadata is usually known at the compliant level, then that should be possible. I think there's relatively few cases where we compose compliant expressions (though there are some, like in `sum_horizontal` we call `is_null`), but for those few cases we should be able to ensure that some `preserve_metadata` flag is called

Definitely happy to work towards a better solution though

MarcoGorelli:
Thanks Dan!

I'm also 100% interested in exploring it, I'm sure we can do better! thanks for your collaboration

PR #2285, opened 2025-03-25
build: Include tests in source distributions

### LINKED ISSUE #2284: [Enh]: Include tests in source distribution

I am adding Narwhals package to [Gentoo Linux](https://gentoo.org). It works great, and working with is a pleasure!

For Gentoo packages, we really want to run tests as part of the package building/install process. While using GitHub archives is acceptable, we strongly prefer using source distributions from PyPI, and this is especially important now that PyPI supports signed attestations.

Include all the files needed to run tests in the source distribution.

The alternative is to download autogenerated GitHub archives. However, unlike source distributions, they are not signed and they technically don't have 100% guarantee that the file checksum won't change in the future.

_No response_

mgorny:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2284

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Include test files in the source distribution, to make it possible to run them while repackaging from sources.  This permits Linux distributions to use signed PyPI release artficats rather than having to use the git repository directly.


MarcoGorelli (changes requested):
thanks for your pr but i'd rather not increase the package size

mgorny:
> thanks for your pr but i'd rather not increase the package size

Does it matter? Regular users install from wheels anyway. Almost all users of source distributions are downstream repackagers who actually want tests.

> > rather than having to use the git repository directly
> 
> just for my understanding, what's the downside of using the git repository?

`git archive` has no stability guarantees, and the stability of produced artifacts relies entirely on keeping the same compressor version. GitHub already broke that once in the past, and the immediate result is that all pipelines relying on GitHub archives broke over hash changes.

On top of that, PyPI supports signed attestations which provide an additional layer of authentication.

MarcoGorelli:
So this wouldn't affect the wheel, but only the `.tar.gz` file that gets uploaded?

mgorny:
Yes. I suppose the diff can be confusing, but I've verified that Hatchling includes only the `narwhals` package itself in the wheel.

MarcoGorelli:
Thanks - would it work if we just included a subset of tests? I'm just wondering if people will see a large tar gz file on github and draw the wrong conclusion

mgorny:
That wouldn't work for us, since we want to run as many tests as possible (and they are fast enough not to justify skipping some). Not sure how much that can actually help you either, given that without tests, the tarball is already 253 KiB (with tests, it's 432 KiB — I'm not really convinced people can really infer anything from these two numbers).

Another option would be to generate a separate tarball and upload it somewhere (e.g. attach to GitHub release) — but that's more work. An extra advantage of that approach is that you would be able to use a stronger compression.

MarcoGorelli:
thanks for measuring + explaining!

i don't know much about packaging (and virtually nothing about linux distributions) but i'll look into this shortly, we can probably get this into the next release

thanks for your interest here!

EdAbati (commented):
Thank you for this PR!
I'm also definitely not an expert in packaging/linux distributions but I took this opportunity to learn more about it 😄

It looks like there is a bit of debate whether is a requirement to include `tests` in `sdist` or not and it seems that the answer is "it depends but there is no harm in doing so": see ["Should sdists include docs and tests?"](https://discuss.python.org/t/should-sdists-include-docs-and-tests/14578)"

The [Python Packaging documentation](https://packaging.python.org/en/latest/discussions/package-formats/#what-is-a-wheel) says at some point:
> Wheels are meant to contain exactly what is to be installed, and nothing more. In particular, wheels should never include **tests** and **documentation**, while **sdists commonly do.**

I checked what some of the most common library do. `pandas` and `scikit-learn` have tests in both `wheel` and `sdist` because their tests are close to the source code not in a separate folder. Our friend `polars` has `docs` and `tests` (and other stuff) in `sdists` but not in `wheel`. See `tar.gz`:

Screenshot 2025-04-01 at 08 56 17


In our case we could use the [`tool.hatch.build.target.<target>` functionality](https://hatch.pypa.io/dev/config/build/#patterns) to include what we want. I tried locally and this seems to do the trick:

```
[tool.hatch.build.targets.sdist]
include = [
  "narwhals/*",
  "tests/*",
  # or docs too?
  # "docs/*",
]

[tool.hatch.build.targets.wheel]
include = [
  "narwhals/*",
]
```

I think `include` is still better than `exclude` so that we don't have to update each time we add a file in root (e.g. just noticed that we ship the `Makefile` in the `.tar.gz`)

MarcoGorelli:
thanks @EdAbati !

@mgorny does the 'include' suggestion work for you?

.github/workflows/publish_to_pypi.yml
```diff
@@ -26,13 +26,11 @@ jobs:
...
-      run: >-
-        python3 -m
-        pip install
-        build
-        --user
+      run: python3 -m pip install build --user
     - name: Build a binary wheel and a source tarball
       run: python3 -m build
+    - name: Check wheel content
+      run: python3 utils/check_wheel_content.py
```
FBruzzesi, line 33: @MarcoGorelli there could be a simpler way of doing this than the 20 lines script below, but I thought it might be a good and quick sanity check

utils/check_dist_content.py
```diff
@@ -0,0 +1,41 @@
...
+    # Allow only 'narwhals' ans 'tests' folders
+    sdist_dirs = {m.name.split("/")[1] for m in sdist_file.getmembers()}
+    allowed_sdist_dirs = {
+        "narwhals",
+        "tests",
+        "pyproject.toml",
+        "PKG-INFO",
+        "LICENSE.md",
+        "README.md",
+        ".gitignore",
```
FBruzzesi, line 34: Are these expected? Apparently 

```toml
[tool.hatch.build.targets.sdist]
include = [
  "narwhals/*",
  "tests/*",
]
```
will include them

EdAbati: Mmm I think so (based on also what polars does)

EdAbati (approved):
It looks great to me!

Thanks @FBruzzesi for updating :) @mgorny will this fix the issue?

utils/check_dist_content.py
mgorny: Yes. Hatchling always includes them (they are in current sdists too), and well, at least first four are expected. `.gitignore` is kinda debatable, but apparently Hatchling uses it sometimes to determine what to install ([not necessarily in a beneficial way](https://github.com/pypa/hatch/issues/1964)).

FBruzzesi:
@MarcoGorelli I will wait for your feedback here before merging 😇

MarcoGorelli (approved):
sure let's do this, thanks all!

mgorny:
Thanks!

PR #2293, opened 2025-03-25
chore: Update `._native_series` refs

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2230
- #2290

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- I've left out `PandasLikeSeries*Namespace`
- Planning to do a bigger change to align with `Arrow*`
  - See #2130

Issue #2084, opened 2025-02-24
Tracking: Drop support for python `3.8`


dangotbanned:
Had quite the ordeal trying to avoid the bugs of `Protocol` in (https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1965921386).
If that compatibility code gets merged, I'd really like to remove it ASAP.
It adds a lot of complexity for a version almost **5 months past [EOL](https://devguide.python.org/versions/#python-release-cycle)**

> [!NOTE]
> Collected more info on `Protocol` issues in (https://github.com/narwhals-dev/narwhals/pull/2294#discussion_r2014534830)

I collected a lot of info a while back in a similar issue for `altair`
- https://github.com/vega/altair/issues/3497

### Downstream `"<=3.8"`
All projects listed in [README.md#used-by](https://github.com/narwhals-dev/narwhals/blob/8609786a1cdc29055cc422255736c8688cec0a5a/README.md#used-by) have dropped `3.8` support, except:

- [ ] [`plotly`](https://github.com/plotly/plotly.py/blob/c54a2bdf1655caaaa3b7b71fbfc38a5584767bd5/pyproject.toml#L38)
  - [ ] https://github.com/plotly/plotly.py/issues/5054
- [x] [`scikit-lego`](https://github.com/koaning/scikit-lego/blob/7f2cc5cd7fd38100761dc88f23178fa37cb34c38/pyproject.toml#L12)
  - [x] https://github.com/koaning/scikit-lego/issues/734

### Upstream
- [x] https://github.com/python/typing_extensions/issues/568
  - **Dev dependency only**

FBruzzesi:
As mentioned in Discord, `timebasedcv` will keep working with the latest Narwhals version that supports python 3.8 (for python 3.8), thus let's not consider that a blocker for dropping support for 3.8.
The library is quite feature complete for now and I am not planning many developments that depends on new narwhals features

dangotbanned:
Got it, thanks @FBruzzesi

dangotbanned:
Might be useful to decide on something like this moving forward

- https://github.com/python/typing_extensions/issues/571

dangotbanned:
## Minimum version bumps
Dropping support for `3.8` will trigger these bumps


| Package  | Minimum | Source                                                                                   |
| -------- | ------- | ---------------------------------------------------------------------------------------- |
[...]

PR #2294, opened 2025-03-25
chore: Add `EagerSeries*Namespace` protocols

### LINKED ISSUE #2230: Formalise `Compliant*` protocols

dangotbanned:
Closes #2230 

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2130
- #2293
- #2230
- #2119
- #2084

## Checklist

- [ ] Tests added
- [x] Documented the changes

- [X] Figure out `<3.11` compatible approach
  - Venture into this one at your own peril (https://github.com/narwhals-dev/narwhals/pull/2294#discussion_r2014534830)
- [x] `PandasLike*`
  - [x] `CatNamespace`
  - [x] `DateTimeNamespace`
  - [x] `ListNamespace`
  - [x] `StringNamespace`
  - [x] `StructNamespace`
- [x] Un-break `generate_backend_completeness`
  - (https://github.com/narwhals-dev/narwhals/pull/2294#issuecomment-2755683513)
  - (https://github.com/narwhals-dev/narwhals/pull/2294/commits/3ff77a057091aa97c4734c5a3274be4db698b671)


narwhals/_compliant/series.py
```diff
@@ -284,3 +294,78 @@ def __narwhals_namespace__(self) -> EagerNamespace[Any, Self, Any]: ...
...
+
+    def __init__(self, series: EagerSeriesT_co, /) -> None:
+        self._compliant_series = series
+
+
+class EagerSeriesCatNamespace(  # type: ignore[misc]
+    _SeriesNamespace[EagerSeriesT_co, NativeSeriesT_co],
+    CatNamespace[EagerSeriesT_co],
+    Protocol[EagerSeriesT_co, NativeSeriesT_co],
+): ...
```
dangotbanned, line 343: Now that these are in the right place, I can collect my thoughts on the **3** solutions we can use for `Protocol` definintions with an `__init__`.

# Problem
Defining a constructor in a `Protocol` can produce two different bugs.
I've documented that in (https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1965921386) - but the short of it is `3.8` and `3.(9|10)` have issues.

# Solutions
## 1. `Protocol38`
This was the first one I discovered (for #2064) and I really dislike it due to the complexity.

*"Fixing"* `3.8` requires tricking the type checker into thinking `typing.Generic` is `typing.Protocol`.

<details><summary>Imports</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/expr.py#L40-L48

</p>
</details> 


The side-effect is **every** sub-protocol is then **required** to use that `Protocol38` definition or face runtime errors - that can no-longer be caught statically

```py
E   TypeError: Protocols can only inherit from other protocols, got <class 'narwhals._compliant.series.EagerSeriesNamespace'>

```

*"Fixing"* `3.(9|10)` requires redefining `__init__` anyway in the impl.
Which mostly negates the benefit of defining in the `Protocol` 🤦‍♂️

<details><summary>Repeat <code>__init__</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/expr.py#L332-L345

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_arrow/expr.py#L33-L55

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_pandas_like/expr.py#L69-L91

</p>
</details> 

## 2. Define the constructor another way
(#2261) introduced a different solution - but still needed to use `Protocol38` to avoid the metaclass issues of the first solution.

I like this **a lot more** as it feels more *intuitive* from the outside - given the clear chain of `@classmethod` and `@property` definitions.

<details><summary><code>when-then</code></summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_arrow/namespace.py#L241-L242

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_arrow/namespace.py#L282-L285

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/when_then.py#L72-L82

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/when_then.py#L63-L64

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/when_then.py#L67-L70

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/narwhals/_compliant/when_then.py#L94-L115

</p>
</details> 

The only downside I could see is [`__new__`](https://docs.python.org/3/reference/datamodel.html#object.__new__) being a less-understood method than [`__init__`](https://docs.python.org/3/reference/datamodel.html#object.__init__).
Personally it took me a while to understand when it makes sense to use. 
Even now that I do - I didn't think of it as a solution until a month after (#2064) merged 😅 

## 3. Avoid constructors *entirely* in `Protocol`(s)
That is the solution this comment is attached to.

It might be easy to miss, but `_SeriesNamespace` is a `Protocol` and `EagerSeriesNamespace` is a `Generic` class.

The interesting aspect of that is the `Protocol`(s) represent the the interface - but the concrete impl is separate.
This approach makes sense when there interface and implementation differ greatly.
Here it is `ArrowSeries` and `PandasLikeSeries` - but we go *fully* [structural](https://typing.python.org/en/latest/spec/glossary.html#term-structural) for all of `Polars`

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/14091955981/job/39470550165?pr=2294

Haven't figured out *why* yet, but the `.pivot(...)` here is throwing for `Series.dt`

https://github.com/narwhals-dev/narwhals/blob/697bb2c0f40fcd91a674391feb511b2f83f3e1fe/utils/generate_backend_completeness.py#L103-L111

First guess would be the extra properties for `PandasLike` in https://github.com/narwhals-dev/narwhals/blob/7daa8f6200a4599661faa9d0ecca08ec87e31a2f/narwhals/_pandas_like/utils.py#L813-L824

MarcoGorelli (approved):
thanks @dangotbanned !

PR #2300, opened 2025-03-26
feat: Allow spark-like backends in `scan_csv` function

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -908,13 +908,15 @@ def _scan_csv_impl(
         if (session := kwargs.pop("session", None)) is None:
             msg = "Spark like backends require a session object to be passed in `kwargs`."
             raise ValueError(msg)
+
+        csv_reader = session.read.format("csv")
         native_frame = (
-            session.read.format("csv").load(source)
+            csv_reader.load(source)
             if (
                 implementation is Implementation.SQLFRAME
-                and (parse_version(version("sqlframe"))) < (3, 27, 0)
+                and parse_version(version("sqlframe")) < (3, 27, 0)
             )
-            else session.read.format("csv").options(**kwargs).load(source)
+            else csv_reader.options(**kwargs).load(source)
         )
     else:  # pragma: no cover
         try:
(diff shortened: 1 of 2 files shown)


narwhals/functions.py
```diff
@@ -904,6 +904,18 @@ def _scan_csv_impl(
...
+            msg = "Spark like backends require a session object to be passed in `kwargs`."
+            raise ValueError(msg)
+        native_frame = (
+            session.read.format("csv").load(source)
+            if (
+                implementation is Implementation.SQLFRAME
+                and (parse_version(version("sqlframe"))) < (3, 27, 0)
+            )
+            else session.read.format("csv").options(**kwargs).load(source)
+        )
```
EdAbati, line 918:
```suggestion
        reader = session.read.format("csv")
        if not (
            implementation is Implementation.SQLFRAME
            and (parse_version(version("sqlframe"))) < (3, 27, 0)
        ):
            reader = reader.options(**kwargs)
        native_frame = reader.load(source)
```

I remember you commenting about readibility in the previous `scan_parquet` PR. I just thought about this alternative (but if we change here we should probably make `scan_parquet` consistent)

EdAbati (approved):
It looks 👌 the tests look much better now. Just added a minor comment about readibility but it is not a blocker

narwhals/functions.py
dangotbanned: You could also use the method `is_sqlframe`

```py
implementation is Implementation.SQLFRAME
implementation.is_sqlframe()
```

FBruzzesi:
Thanks @EdAbati and @dangotbanned for the reviews! I ended up pushing something in the middle between how it was and your suggestion!

@dangotbanned I know about `is_sqlframe`, but I was also considering to avoid using `is_spark_like()` a few lines above, since we don't do that for other backends - it might totally be because when we created this function we didn't have the methods, but I consider `is_<backend>(_like)` to be more user facing (for some weird reason in my head)

PR #2301, opened 2025-03-26
feat: Add `DuckDBExpr.is_unique()` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

Requires `SQLExpression`

### DIFF
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -613,6 +613,13 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
 
         return self._from_call(func)
 
+    def is_unique(self: Self) -> Self:
+        def func(_input: duckdb.Expression) -> duckdb.Expression:
+            sql = f"count(*) over (partition by {_input})"
+            return SQLExpression(sql) == lit(1)
+
+        return self._from_call(func)
+
     @property
     def str(self: Self) -> DuckDBExprStringNamespace:
         return DuckDBExprStringNamespace(self)
@@ -635,4 +642,3 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
 
     drop_nulls = not_implemented()
     unique = not_implemented()
-    is_unique = not_implemented()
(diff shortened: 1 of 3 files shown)


narwhals/_duckdb/expr.py
```diff
@@ -613,6 +613,13 @@ def func(_input: duckdb.Expression) -> duckdb.Expression:
 
         return self._from_call(func)
 
+    def is_unique(self: Self) -> Self:
+        def func(_input: duckdb.Expression) -> duckdb.Expression:
+            sql = f"count(*) over (partition by {_input})"
+            return SQLExpression(sql) == lit(1)  # type: ignore[no-any-return, unused-ignore]
```
dangotbanned, line 619: @FBruzzesi I trust that this is correct, but my brain can't connect the dots :pensive:

If we pretend `pl.Expr.is_unique` didn't exist - what would this expression translate to in `polars` syntax?

FBruzzesi: Sure!

Not sure what the clear way to show this, I can think of a few options:

The 1-1 translation should be:

```py
pl.col("x").len().over("x") ==1
```

The extended (or the way to do it in pandas):

```py
df.groupby("x").transform("size") == 1
```

dangotbanned (approved):
Thanks @FBruzzesi, (https://github.com/narwhals-dev/narwhals/pull/2301#discussion_r2016536613) isn't blocking by any means - just wanted to understand it with an API I'm more familar with 🙂

narwhals/_duckdb/expr.py
FBruzzesi: It should be `over` the same column:

```py
result = (
    df.select(
        pl.col("a").len().over("a")==1,
        pl.col("b").len().over("b")==1,
        "index"
    ).sort("index")
```

dangotbanned: Aha yeah that did the trick - nice one @FBruzzesi - feel free to merge 🙌

Issue #2116, opened 2025-02-28
API: io functions for v2


MarcoGorelli:
Currently we're a bit inconsistent with IO functions:
- from_native
- from_dict
- from_arrow
- new_series
- from_numpy

I think we might be better off movings towards:
- `from_native`: may return Series, DataFrame, or LazyFrame, depending on the input
- `DataFrame.from_native`: like `from_native(..., eager_only=True, allow_series=False)`
- `LazyFrame.from_native`: like `from_native(...).lazy()`
- `Series.from_native`: like `from_native(..., series_only=True)`
- `DataFrame.from_dict`, `DataFrame.from_arrow`, `DataFrame.from_numpy`. Remove the top-level ones
- `Series.from_iterable`, `Series.from_numpy`. Remove `nw.new_series`, that's an inconsistent name

All of this would only affect the main namespace and stable.v2 onwards, stable.v1 should remain unaffected

dangotbanned:
@MarcoGorelli I had a similar thought yesterday - but slightly smaller in scope.

`from_native` could be split into:
- `from_frame` (optionally as a data/lazy pair)
- `from_series`

Those would actually be shorter than the current version:

```py
from typing import TYPE_CHECKING

import polars as pl

import narwhals as nw

if TYPE_CHECKING:
    from narwhals.typing import IntoFrame
    from narwhals.typing import IntoSeries

data = {"a": [1, 2, 3]}

frame: IntoFrame = pl.DataFrame(data)
lazy: IntoFrame = pl.DataFrame(data).lazy()
series: IntoSeries = pl.Series(data["a"])

# v1
nw.from_native(frame)
nw.from_native(lazy, eager_only=False)
nw.from_native(series, allow_series=True)
nw.from_native(series, series_only=True)

# v2
nw.from_frame(frame)
nw.from_frame(frame, eager_only=False)
nw.from_series(series)
```

I hadn't thought about adding `@classmethod`(s) though - I do like that as well!

MarcoGorelli:
Thanks!

Yeah I think i'd prefer classmethods

For `into_`, I think that's what I'd expect a method to be called - e.g. `df.into_frame()`, rather than `nw.into_frame(df)`. But we can't control what methods upstream libraries have, so we may be stuck with `nw.from_` / `nw.DataFrame.from_` etc...

dangotbanned:
https://github.com/narwhals-dev/narwhals/issues/2116#issuecomment-2691018631

[...]

PR #2306, opened 2025-03-28
refactor: Add `CompliantDataFrame.from_arrow`

dangotbanned:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues
- Part of #2116

## Checklist

- [x] Tests added
- [x] Documented the changes

- [x] Refine the error handling for each backend
  - Most of it can live in `ArrowDataFrame.from_arrow`
- [x] Widen to accept `pa.Table`
  - since we support `pyarrow>=11`, I think this function can still provide some value for those earlier versions
  - The `Arrow*` impl has similar logic to `pa.table` in `pyarrow==13.0.0`


dangotbanned:
@MarcoGorelli I'm close to finishing, just wanted to check in and see if you had any thoughts on supporting `pa.Table` here?

I left some notes in the description on why it makes sense to me 🙂

Would be very easy to support across all versions

MarcoGorelli:
thanks @dangotbanned ! sure, should be good to accept pa.table pre-pycapsule

dangotbanned:
> sure, should be good to accept pa.table pre-pycapsule

Perfect, thanks @MarcoGorelli 

Still need to test how `pa.table` responds to `pa.Table` for post-pycapsule.
I *think* it`d use `__arrow_c_stream__`, but there's only one way to find out

tests/from_pycapsule_test.py
```diff
@@ -25,15 +37,17 @@ def test_from_arrow_to_arrow() -> None:
...
-@pytest.mark.xfail(PYARROW_VERSION < (14,), reason="too old")
 def test_from_arrow_to_polars(monkeypatch: pytest.MonkeyPatch) -> None:
     pytest.importorskip("polars")
     import polars as pl
 
     tbl = pa.table({"ab": [1, 2, 3], "ba": [4, 5, 6]})
     monkeypatch.delitem(sys.modules, "pandas")
-    df = nw.from_native(tbl, eager_only=True)
-    result = nw.from_arrow(df, backend=pl)
+    if PYARROW_VERSION < (14,):
```
dangotbanned, line 46: https://github.com/narwhals-dev/narwhals/actions/runs/14134927029/job/39604239875?pr=2306

Feel like this would be good to have coverage for, but not sure how feasible that is with the current CI.
Adding a `# pragma: no cover` seems wrong

dangotbanned: Also, it means the previous `xfail`(s) were never triggered?

narwhals/_arrow/dataframe.py
```diff
@@ -93,6 +96,26 @@ def __init__(
         self._version = version
         validate_backend_version(self._implementation, self._backend_version)
 
+    @classmethod
+    def from_arrow(cls, data: IntoArrowTable, /, *, context: _FullContext) -> Self:
+        backend_version = context._backend_version
+        if isinstance(data, pa.Table):
+            native = data
+        elif backend_version >= (14,) or isinstance(data, Collection):
```
dangotbanned, line 104: Supporting `Collection` here might seem strange.

My thinking is that *were* a `list | tuple | dict` of `pyarrow` objects passed - `pa.table` would still be capable of turning that into a `pa.Table`:
https://github.com/apache/arrow/blob/7c1b96e26910dd553ee0ccf0792e1854af6b0021/python/pyarrow/table.pxi#L4993-L5000

narwhals/_pandas_like/dataframe.py
```diff
@@ -114,6 +116,29 @@ def __init__(
...
+            native = tbl.to_pandas()
+        elif implementation.is_modin():  # pragma: no cover
+            from modin.pandas.utils import from_arrow as mpd_from_arrow
+
+            native = mpd_from_arrow(tbl)
+        elif implementation.is_cudf():  # pragma: no cover
+            native = implementation.to_native_namespace().DataFrame.from_arrow(tbl)
+        else:  # pragma: no cover
+            msg = "congratulations, you entered unreachable code - please report a bug"
+            raise AssertionError(msg)
```
MarcoGorelli, line 133: lol, can't believe i wrote this, we may want to make a utility with a consistent message for these at some point

MarcoGorelli (approved):
nice, thanks @dangotbanned !

tests/from_pycapsule_test.py
MarcoGorelli: no strong preference

narwhals/_pandas_like/dataframe.py
dangotbanned: Certainly made me smile when I copied it over 😄

PR #2310, opened 2025-03-28
feat: add `rank` for Lazy backends

### LINKED ISSUE #2174: implement remaining order-dependent lazy operations

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2174

## Checklist

- [x] Tests added
- [ ] Documented the changes

Only `min` and `dense` methods are implemented for DuckDB and PySpark without `.over()`.


FBruzzesi:
Hey @raisadz thanks for the PR. I didn't have the time to look at this, I just wanted to drop a comment and ping @marvinl803 who also took a look at `rank` and maybe can share some insights for the remaining `method`s 🙌🏼

MarcoGorelli (commented):
thanks for working on this 🙌 !

it's occurred to me that we class `rank` as `ExprKind.WINDOW` and that, after `narwhals.stable.v1`, it won't be supported to use `rank` if not followed by `over(order_by=...)`

Here it works because the import is `import narwhals.stable.v1 as nw`, but would raise for `import narwhals as nw`

🤔 gonna have to think about a solution here, not super-sure

_maybe_, just maybe, we may want to introduce `nw.rank` instead of `Expr.rank`, because then that would compose more naturally with `over`. And for eager backends, `nw.col('a').rank()` does the same thing as `nw.rank().over('a')`

🤔 i'm not totally sure, will do a bit more thinking on this one, but i think we can come up with something

MarcoGorelli:
Having slept on this, I'd suggest:
```
nw.rank().over(order_by='a')  # equivalent to pl.col('a').rank()
```

An alternative could be `nw.rank('a', partition_by='b')`, but then what happens if a user writes `nw.rank('a', partition_by='b').over('b')`? I think in general, any expression that uses `over` internally is problematic

The only solution I can currently think of is: `nw.rank().over(order_by='a')`, and deprecate `Expr.rank` (but keep it around in `stable.v1` - this is especially important here because, IIRC, we already have downstream libraries using `nw.Expr.rank`)

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2310#issuecomment-2763250852

@MarcoGorelli I haven't looked into this more than your two comments, so apologies in advance if this seems too simplified of a view.

IMO, the *current* `narwhals` internal implementation of the `polars` API **shouldn't** lead the decision on the `narwhals` public API.
*If* that is the case here, the general concept (if you're unfamiliar) is called a [leaky abstraction](https://en.m.wikipedia.org/wiki/Leaky_abstraction).

---

I'm not sure how likely this would be, but another concern might be - what if `polars` adds a `pl.rank` in the future that has different behaviour?
The list of [sugar functions](https://docs.pola.rs/api/python/stable/reference/expressions/functions.html) is still growing

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2310#issuecomment-2763275571

@MarcoGorelli that seems consistent with the [`pl.Expr.rank` docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.rank.html) 👍 

I don't know if this already plays a factor in the `narwhals` API.
Do you think about *aiming* to cover sequences of expressions that are **demonstrated** in the `polars` docs?
That seems like a good compass to me - even if we can't always get there

MarcoGorelli:
Yeah i think it's good to aim for as much expressivity as Polars offers, with the restriction that for LazyFrame we only allow relational expressions - so, `lf.select(nw.col('a').drop_nulls(), nw.col('b').drop_nulls())` will never be allowed. The rest should in theory just be limited by how clear our mental model and implemenation is

narwhals/_duckdb/expr.py
```diff
@@ -692,6 +692,34 @@ def func(_input: duckdb.Expression) -> duckdb.Expression:
...
+        def _rank(_input: duckdb.Expression) -> duckdb.Expression:
+            if descending:
+                by_sql = f"{_input} desc nulls last"
+            else:
+                by_sql = f"{_input} asc nulls last"
+            order_by_sql = f"order by {by_sql}"
+            sql = (
+                f"CASE WHEN {_input} IS NULL THEN NULL "
+                f"ELSE {func_name}() OVER ({order_by_sql}) END"
+            )
```
MarcoGorelli, line 718: can we flip this round (case when not-null), like you did for the spark-like case?

MarcoGorelli (approved):
thanks @raisadz - I think we can ship this, just left a minor comment

MarcoGorelli:
i'll address https://github.com/narwhals-dev/narwhals/pull/2310#discussion_r2019986932 now, gonna start by getting this in though

PR #2314, opened 2025-03-29
feat: disallow `.over(order_by=...)` for non-order-dependent operations (as it signals a user error)

MarcoGorelli:
We currently allow `nw.col('a').mean().over(order_by='i')`, but probably shouldn't because the `order_by` has no effect there and is a sign of a user-error

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -187,8 +187,8 @@ jobs:
         run: |
           cd py-shiny
           # temporary pin to get ci green
-          uv pip install "pyflakes<3.3.0" "flake8<7.2.0" --system
           make narwhals-install-shiny
+          uv pip install "pyflakes<3.3.0" "flake8<7.2.0" --system
       - name: install-narwhals-dev
         run: |
           uv pip uninstall narwhals --system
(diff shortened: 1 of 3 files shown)


dangotbanned:
> We currently allow `nw.col('a').mean().over(order_by='i')`, but probably shouldn't because the `order_by` has no effect there and is a sign of a user-error

That does sound reasonable, but what does `polars` do in that scenario?

AFAIK, `polars` *usually* allows no-ops. I'm guessing because they'd be optimized out anyway

MarcoGorelli:
thanks!

yeah but tbh i think it's potentially misleading that Polars no-ops, else a user might expect `pl.col('a').sum().over('b', order_by='c')` to behave like sql `sum(a) over (partition by b order by c)`, whereas it's actually like ``sum(a) over (partition by b)``

PR #2315, opened 2025-03-29
chore: Add `Compliant*.from_native`

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Part of #2116

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Following roughly the same pattern as `from_numpy`

- [x] `FromNative` protocol
- [x] `CompliantSeries`
- [x] `CompliantDataFrame`
- [x] `CompliantNamespace` (instance method)
  - [x] `EagerNamespace` 
  - [x] `PolarsNamespace`
  - [x] `LazyNamespace`

PR #2304, opened 2025-03-27
refactor: Add `CompliantDataFrame.from_dict`

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2283
- #2302
- Part of #2116

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2283, opened 2025-03-24
refactor: `*(Namespace|DataFrame).from_numpy`

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/2196#issuecomment-2726773113
- #2116

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- This PR is the first big step towards (#2116), following all of the work in (#2230)
- I chose `from_numpy` as all the supported backends already had a `to_numpy`, to complete the `NumpyConvertible` pair
- Experimenting with the possibility of typing support from `functions.py`
  - Everything checks out with `pyright`, but `mypy` had issues with defining `overload`(s) on `Implementation`
  - Another option *could* be having a `nw.Namespace` class to provide a scope for the single `Implementation` -> `CompliantNamespace` context

PR #2324, opened 2025-04-01
feat: Adds private `Namespace` class

dangotbanned:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Mentioned in (#2283)
- Related
  - #2315
  - #2304
  - #2116

## Checklist

- [X] Tests added
- [X] Documented the changes

Had this idea nagging away at me while working on (#2116).

### Super high-level
This PR adds a `Namespace` class, that can be created via either:

```py
Namespace.from_backend(...)
Namespace.from_native_object(...)
```

Supporting these new `@classmethod`s is a new `@property` for `Implementation`:

```py
Implementation._backend_version
```

### Notes
I've tried to keep things isolated from the rest of `narwhals` - for now - but **the goal would be replacing similar logic** from all of:
- `functions.py`
- `translate.py`
- `utils.py`

> [!IMPORTANT]
> Not planning to make this a public class

## Example
This *further* simplifies (#2283) to a one-liner, with the added bonus of typing from a `backend` string

image


narwhals/utils.py
```diff
@@ -531,6 +533,26 @@ def is_sqlframe(self: Self) -> bool:
...
+            into_version = native
+        elif self is Implementation.PYSPARK:
+            into_version = get_pyspark()
+        elif self is Implementation.DASK:
+            into_version = get_dask()
+        else:
+            import sqlframe._version
+
+            into_version = sqlframe._version
+        return parse_version(into_version)
```
dangotbanned, line 554: This could be used in place of *most* calls we currently do to `parse_version`.

Here we'd only need `Implementation` and use the same property for all of them.

- `is_sqlframe_dataframe`
- `SparkLikeLazyFrame`
- `sqlframe._version`

https://github.com/narwhals-dev/narwhals/blob/4d2b9d57f70e52b7c78ca3f41e228be2ceb96cfa/narwhals/translate.py#L755-L775

If you then combine this PR with (#2315), we can do:

```py

from typing import cast

from narwhals._namespace import Namespace
import narwhals as nw

if TYPE_CHECKING:
    from narwhals._spark_like.dataframe import SQLFrameDataFrame
    
native_object = cast("SQLFrameDataFrame", "pretend im a sqlframe")
compliant = Namespace.from_native_object(native_object).from_native(native_object)
nw.LazyFrame(compliant, level="lazy")
```

narwhals/_namespace.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @overload
+    @classmethod
+    def from_backend(cls, backend: ModuleType, /) -> Namespace[CompliantNamespaceAny]: ...
+
+    @classmethod
+    def from_backend(
+        cls: type[Namespace[Any]], backend: IntoBackend, /
+    ) -> Namespace[Any]:
+        """Instantiate from native namespace module, string, or Implementation.
```
dangotbanned, line 192: Intended to replace a util I added recently https://github.com/narwhals-dev/narwhals/blob/4d2b9d57f70e52b7c78ca3f41e228be2ceb96cfa/narwhals/utils.py#L569-L571

narwhals/_namespace.py
```diff
@@ -0,0 +1,354 @@
...
+        _partition_type: type[pd.DataFrame]
+
+    class _NativeCuDF(Protocol):
+        def to_pylibcudf(self, *args: Any, **kwds: Any) -> Any: ...
+
+    class _ModinDataFrame(Protocol):
+        _pandas_class: type[pd.DataFrame]
+
+    class _ModinSeries(Protocol):
+        _pandas_class: type[pd.Series[Any]]
```
dangotbanned, line 117: I tried using the actual types first (https://github.com/narwhals-dev/narwhals/pull/2324/commits/acb5787fb3237c6ada2782eae2ee40d03317aa0f), but they broke the `@overload`(s).

These seem specific enough to match *only* the intended target(s)

narwhals/_namespace.py
```diff
@@ -0,0 +1,354 @@
...
+def is_native_polars(obj: Any) -> TypeIs[_NativePolars]:
+    return (pl := get_polars()) is not None and isinstance(
+        obj, (pl.DataFrame, pl.Series, pl.LazyFrame)
+    )
+
+
+def is_native_arrow(obj: Any) -> TypeIs[_NativeArrow]:
+    return (pa := get_pyarrow()) is not None and isinstance(
+        obj, (pa.Table, pa.ChunkedArray, pa.Array)
+    )
```
dangotbanned, line 319: See (https://github.com/narwhals-dev/narwhals/pull/2315#issuecomment-2764252252) for how this links back to `from_native`

narwhals/_namespace.py
```diff
@@ -0,0 +1,363 @@
...
+    SparkLike: TypeAlias = Literal[
+        _SparkLike, Implementation.PYSPARK, Implementation.SQLFRAME
+    ]
+    EagerOnly: TypeAlias = "PandasLike | Arrow"
+    EagerAllowed: TypeAlias = "EagerOnly | Polars"
+    LazyOnly: TypeAlias = "SparkLike | Dask | DuckDB"
+    LazyAllowed: TypeAlias = "LazyOnly | Polars"
+
+    BackendName: TypeAlias = "_EagerAllowed | _LazyAllowed"
+    IntoBackend: TypeAlias = "BackendName | Implementation | ModuleType"
```
dangotbanned, line 105: I'd wanna make these aliases public, somewhat superseding (#1917)

- [ ] #1917

Aiming for the same benefits mentioned in (https://github.com/narwhals-dev/narwhals/pull/1914#issuecomment-2629460801) - but with `@overload` matching on `Literal`s

narwhals/_namespace.py
```diff
@@ -0,0 +1,363 @@
...
+        self._compliant_namespace = namespace
+
+    def __init_subclass__(cls, *args: Any, version: Version, **kwds: Any) -> None:
+        super().__init_subclass__(*args, **kwds)
+
+        if isinstance(version, Version):
+            cls._version = version
+        else:
+            msg = f"Expected {Version} but got {type(version).__name__!r}"
+            raise TypeError(msg)
```
dangotbanned, line 152: Finally managed to follow up on the last example in (https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-12054784)

<details><summary>All the bad variants</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/29ac3341aacc6ed1ed5162ccfa306ddf8e01fe4a/tests/namespace_test.py#L147-L168

</p>
</details> 


The correct version is a single line 🙂 

https://github.com/narwhals-dev/narwhals/blob/29ac3341aacc6ed1ed5162ccfa306ddf8e01fe4a/narwhals/stable/v1/_namespace.py#L10

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2356#discussion_r2032967492)

@FBruzzesi this PR has some interesting typing stuff as well 😅

narwhals/utils.py
```diff
@@ -563,6 +566,26 @@ def _alias(self) -> LiteralString:
         }
         return mapping[self]
 
+    @property
+    def _backend_version(self) -> tuple[int, ...]:
```
MarcoGorelli, line 570: should this be a function (not property) given it does actual work?

dangotbanned: @MarcoGorelli good question!

So I chose a property because
- it doesn't take any arguments
- `_backend_version` is an attribute everywhere else

> given it does actual work?

Although I couldn't find anything in the `@property` docs about it, `@cached_property` mentions that they can be computed https://docs.python.org/3/library/functools.html#functools.cached_property

dangotbanned: `packaging` implements parsing (mostly) via properties btw

https://github.com/pypa/packaging/blob/3910129009b25dd1aa1fe32e644bc891188c56fe/src/packaging/specifiers.py

MarcoGorelli: Sure, thanks - I'd prefer it to be a function so it's clearer that it's not a free operation

dangotbanned: > Sure, thanks - I'd prefer it to be a function so it's clearer that it's not a free operation

No problem 🙂

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2324/commits/c9865ab092a89a94d1701f087d065b475456e3b3)

MarcoGorelli (approved):
thanks @dangotbanned

dangotbanned:
> thanks @dangotbanned
> 

thanks @MarcoGorelli!

Just going to make the replacements now in `utils` and `functions` - where this'll be used first

- https://github.com/narwhals-dev/narwhals/pull/2324#discussion_r2023222391
- https://github.com/narwhals-dev/narwhals/pull/2283

dangotbanned:
@MarcoGorelli after (https://github.com/narwhals-dev/narwhals/pull/2324#commits-pushed-03a08ef) I'm gonna call it here, too easy to get carried away 😅 

Hopefully you get an idea for how much of `from_native` this can simplify

PR #2331, opened 2025-04-03
chore: use less sql for duckdb, do extra validation

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -356,8 +356,7 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
     def unique(
         self: Self, subset: Sequence[str] | None, *, keep: Literal["any", "none"]
     ) -> Self:
-        subset_ = subset if keep == "any" else (subset or self.columns)
-        if subset_:
+        if subset_ := subset if keep == "any" else (subset or self.columns):
             # Sanitise input
             if any(x not in self.columns for x in subset_):
                 msg = f"Columns {set(subset_).difference(self.columns)} not found in {self.columns}."
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -277,6 +277,7 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
             condition: duckdb.Expression = reduce(and_, it)
             rel = self.native.set_alias("lhs").join(
                 other.native.set_alias("rhs"),
+                # NOTE: Fixed in `--pre` https://github.com/duckdb/duckdb/pull/16933
                 condition=condition,  # type: ignore[arg-type, unused-ignore]
                 how=native_how,
             )
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -251,18 +251,20 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
             if self._backend_version < (1, 1, 4):
                 msg = f"DuckDB>=1.1.4 is required for cross-join, found version: {self._backend_version}"
                 raise NotImplementedError(msg)
-            rel = self.native.set_alias("lhs").cross(
-                other.native.set_alias("rhs")
-            )  # pragma: no cover
+            rel = self.native.set_alias("lhs").cross(other.native.set_alias("rhs"))
         else:
             # help mypy
             assert left_on is not None  # noqa: S101
             assert right_on is not None  # noqa: S101
-            condition = " and ".join(
-                f'lhs."{left}" = rhs."{right}"' for left, right in zip(left_on, right_on)
+            it = (
+                col(f'lhs."{left}"') == col(f'rhs."{right}"')
+                for left, right in zip(left_on, right_on)
             )
+            condition: duckdb.Expression = reduce(and_, it)
             rel = self.native.set_alias("lhs").join(
-                other.native.set_alias("rhs"), condition=condition, how=native_how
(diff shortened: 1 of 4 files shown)


narwhals/_duckdb/dataframe.py
```diff
@@ -251,18 +251,20 @@ def join(
...
-            )  # pragma: no cover
+            rel = self.native.set_alias("lhs").cross(other.native.set_alias("rhs"))
         else:
             # help mypy
             assert left_on is not None  # noqa: S101
             assert right_on is not None  # noqa: S101
-            condition = " and ".join(
-                f'lhs."{left}" = rhs."{right}"' for left, right in zip(left_on, right_on)
+            it = (
+                col(f'lhs."{left}"') == col(f'rhs."{right}"')
```
dangotbanned, line 260: Love to see the new `col` alias getting some use 😊

narwhals/_duckdb/dataframe.py
```diff
@@ -319,6 +322,8 @@ def join_asof(
                 select.append(f'rhs."{name}" as "{name}{suffix}"')
             elif right_on is None or name not in {right_on, *by_right}:
                 select.append(f'"{name}"')
+        # Replace with Python API call once
```
dangotbanned, line 325:
```suggestion
        # TODO @MarcoGorelli: Replace with Python API call once
```

narwhals/_duckdb/dataframe.py
```diff
@@ -449,6 +454,8 @@ def unpivot(
 
         unpivot_on = ", ".join(f'"{name}"' for name in on_)
         rel = self.native  # noqa: F841
+        # Replace with Python API once
```
dangotbanned, line 457:
```suggestion
        # TODO @MarcoGorelli: Replace with Python API once
```

narwhals/_duckdb/utils.py
```diff
@@ -235,3 +235,11 @@ def generate_order_by_sql(*order_by: str, ascending: bool) -> str:
...
         by_sql = ", ".join([f'"{x}" desc nulls last' for x in order_by])
     return f"order by {by_sql}"
+
+
+def assert_type(obj: Any, valid_types: tuple[type, ...]) -> None:
+    # Use this for extra (possibly redundant) validation in places where we
+    # use SQLExpression, as an extra guard against unsafe inputs.
+    if not isinstance(obj, valid_types):  # pragma: no cover
+        msg = f"Expected int, got: {type(obj)}"
+        raise TypeError(msg)
```
dangotbanned, line 245: I like the motivation for this, but have a few notes:

- I don't like that this shadows [`typing.assert_type`](https://docs.python.org/3/library/typing.html#typing.assert_type) and does *somewhat* of the inverse of it
  - This function **only** does runtime checking
  - `typing.assert_type` **only** does static checking
- The name includes `assert`, but no `assert` keyword is used
  - An alternative could be `ensure_type`?

IMO, you could generalize this to work in more places by doing something like this:


```py

from typing import Any
from typing import Callable
from typing import Sequence

if TYPE_CHECKING:
    from typing_extensions import TypeIs

def is_int(obj: Any) -> TypeIs[int]:
    return isinstance(obj, int)


def is_none(obj: Any) -> TypeIs[None]:
    return obj is None


def is_sequence(obj: Any) -> TypeIs[Sequence[Any]]:
    return isinstance(obj, Sequence)

# NOTE: It *might* be possble to type this using a `TypeVarTuple`
def ensure_type(obj: Any, *guards: Callable[[Any], TypeIs[Any]]) -> None:
    if not any(guard(obj) for guard in guards):
        msg = "..."
        raise TypeError(msg)


def example(window_size: int | None):
    ensure_type(window_size, is_int, is_none)
```

By using functions, you're not limited by using `type` objects.
You'd evenutally run into needing to check the inner type of containers

narwhals/_duckdb/expr.py
```diff
@@ -561,8 +565,8 @@ def func(window_inputs: WindowInputs) -> duckdb.Expression:
                 )
             else:
                 partition_by_sql = f"partition by {window_inputs.expr}"
-            sql = f"row_number() over({partition_by_sql} {order_by_sql}) == 1"
-            return SQLExpression(sql)  # type: ignore[no-any-return, unused-ignore]
+            sql = f"row_number() over({partition_by_sql} {order_by_sql})"
```
dangotbanned, line 568: Could we use `FunctionExpression` here?

This *feels* like it *could* work
```py
import duckdb

row_number = duckdb.FunctionExpression("row_number")
```

narwhals/_spark_like/utils.py
```diff
@@ -43,7 +43,7 @@ def __init__(
 # NOTE: don't lru_cache this as `ModuleType` isn't hashable
 def native_to_narwhals_dtype(
     dtype: _NativeDType, version: Version, spark_types: ModuleType
-) -> DType:  # pragma: no cover
+) -> DType:
```
dangotbanned, line 46: Nice nice nice

dangotbanned (commented):
Thanks for this @MarcoGorelli

Just left a few comments - looking great for the most part :tada:

narwhals/_duckdb/utils.py
dangotbanned: Reminded me of https://discuss.python.org/t/singledispatch-support-for-pep-585-generic-types/86772

narwhals/_duckdb/expr.py
MarcoGorelli: true, but there's no user input here and it would just get translated to `row_number()` anyway

dangotbanned (approved):
Forgot to approve before, thanks @MarcoGorelli 🎉

PR #2335, opened 2025-04-04
fix: unpivot was unnecessary failing for PySpark/SQLFrame when neither `on` nor `index` were passed

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -495,7 +495,7 @@ class SparkLikeLazyFrame(CompliantLazyFrame["SparkLikeExpr", "SQLFrameDataFrame"
         else:  # pragma: no cover
             pass
 
-        ids = () if index is None else tuple(index)
+        ids = tuple(index) if index else ()
         values = (
             tuple(set(self.columns).difference(set(ids))) if on is None else tuple(on)
         )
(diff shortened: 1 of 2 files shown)


tests/frame/unpivot_test.py
```diff
@@ -64,11 +64,12 @@ def test_unpivot(
     on: str | list[str] | None,
     index: list[str] | None,
     expected: dict[str, list[float]],
-    request: pytest.FixtureRequest,
 ) -> None:
-    if on is None and index is None and "polars" not in str(constructor):
-        # TODO(2082): add support in other backends
-        request.applymarker(pytest.mark.xfail)
```
MarcoGorelli, line 71: the reason this test was failing was because of the mixed types (which I don't think we'd want to support anyway), if we make the types consistent then, with a fix in spark-like code, it passes

narwhals/_spark_like/dataframe.py
```diff
@@ -495,7 +495,7 @@ def unpivot(
         else:  # pragma: no cover
             pass
 
-        ids = tuple(self.columns) if index is None else tuple(index)
+        ids = () if index is None else tuple(index)
```
dangotbanned, line 498:
```suggestion
        ids = tuple(index) if index else ()
```

dangotbanned (approved):
Nice one thanks @MarcoGorelli

tests/frame/unpivot_test.py
FBruzzesi: 🎻🚀

Issue #2337, opened 2025-04-04
enh: Add `polars._typing` aliases


dangotbanned:
## Description
`polars` [defines a pretty extensive set](https://github.com/pola-rs/polars/blob/bab7cefdb987972571542757d1f55d04db9cd087/py-polars/polars/_typing.py) of `TypeAlias`(s) and I think there's plenty we could *steal* to improve our UX.

I'd want to take it a step *further* and add trailing docstrings to them - that way we get some docs *"for free"* when used in the internals 🙂 

## Examples
I'm just picking out one example for each *kind* of alias - but there are **many others**

### `Literal`(s)
Here it would contribute to (#1942) as we have **14x** of this spread across the backends:

```py
how: Literal["horizontal", "vertical", "diagonal"]
```

#### `polars.concat`
```py
def concat(
    items: Iterable[PolarsType],
    *,
    how: ConcatMethod = "vertical",
    rechunk: bool = False,
    parallel: bool = True,
) -> PolarsType:
```

#### `narwhals.concat`
```py
@overload
def concat(
    items: Iterable[DataFrame[IntoDataFrameT]],
    *,
    how: Literal["horizontal", "vertical", "diagonal"] = "vertical",
) -> DataFrame[IntoDataFrameT]: ...


@overload
def concat(
    items: Iterable[LazyFrame[IntoFrameT]],
    *,
    how: Literal["horizontal", "vertical", "diagonal"] = "vertical",
) -> LazyFrame[IntoFrameT]: ...


@overload
def concat(
    items: Iterable[DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT]],
    *,
    how: Literal["horizontal", "vertical", "diagonal"] = "vertical",
) -> DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT]: ...


def concat(
    items: Iterable[DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT]],
    *,
    how: Literal["horizontal", "vertical", "diagonal"] = "vertical",
) -> DataFrame[IntoDataFrameT] | LazyFrame[IntoFrameT]:
```
### Python types
We have cases like this, that would benefit from aliases like `PythonLiteral`, `NumericLiteral`, etc.

https://github.com/narwhals-dev/narwhals/blob/d2a9f423688c14e524d28e7d282bfa5de3c888be/narwhals/expr.py#L1951-L1957

## Related
- https://github.com/pola-rs/polars/pull/17282
[...]

PR #2339, opened 2025-04-04
refactor(typing): Simplify `concat` signature

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Noticed during #2337
- Did something similar in https://github.com/narwhals-dev/narwhals/pull/2283#discussion_r2014022110

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- Using a constrained `TypeVar`, like in `polars`
- Marking as *"fix"*, since (335ed0531faee5abcdb3735b4d6004ed761a7d4b) now warns on a failure case

### DIFF
--- a/narwhals/stable/v1/__init__.py
+++ b/narwhals/stable/v1/__init__.py
@@ -99,6 +99,7 @@ if TYPE_CHECKING:
     from narwhals.typing import _1DArray
     from narwhals.typing import _2DArray
 
+    FrameT = TypeVar("FrameT", "DataFrame[Any]", "LazyFrame[Any]")
     IntoSeriesT = TypeVar("IntoSeriesT", bound="IntoSeries", default=Any)
     T = TypeVar("T", default=Any)
 else:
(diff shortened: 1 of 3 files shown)


narwhals/stable/v1/__init__.py
```diff
@@ -2096,7 +2073,7 @@ def concat(
     Raises:
         TypeError: The items to concatenate should either all be eager, or all lazy
     """
-    return _stableify(nw.concat(items, how=how))
+    return cast("FrameT", _stableify(nw.concat(items, how=how)))
```
dangotbanned, line 2076: Not *super* happy about using `cast` - but it'll do *for now*

dangotbanned:
> nice!
> 
> i was taking a look at concat too, and noticed that we should be disallowing horizontal concatenation for lazyframes
> 
> #2341
> 
> in light of that, is the simplification here still valid?

Thanks @MarcoGorelli 

Hmm, I'll need to have a think

Could you explain a bit more about why this isn't allowed?

(#2340)

> we shouldn't allow this operation at all given that row order is undefined

I'm not making the connection to row order?

It seems supported in `polars` unconditionally

https://github.com/pola-rs/polars/blob/py-1.26.0/py-polars/polars/functions/eager.py#L26-L304

tests/frame/invalid_test.py
```diff
@@ -83,7 +83,7 @@ def test_validate_laziness() -> None:
         TypeError,
         match=("The items to concatenate should either all be eager, or all lazy"),
     ):
-        nw.concat([nw.from_native(df, eager_only=True), nw.from_native(df).lazy()])
+        nw.concat([nw.from_native(df, eager_only=True), nw.from_native(df).lazy()])  # type: ignore[type-var]
```
dangotbanned, line 86: https://github.com/narwhals-dev/narwhals/pull/2339#issuecomment-2779294590

@MarcoGorelli I have no idea if that made sense 😅 

I'm trying to say - if we merge (#2341) then I still want to base the new signatures on the new `TypeVar`.

We currently raise here, but the `@overload`s pass as valid

MarcoGorelli (approved):
thanks @dangotbanned

PR #2347, opened 2025-04-04
fix(typing): Narrow `TypeVar` used in `Series`

### LINKED ISSUE #2343: Narrow `TypeVar` used in `Series`
Should be the easiest to fix in (#2239)

dangotbanned:
Will close #2343

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2343

## Checklist

- [x] Tests added
- [ ] Documented the changes

- Changing `IntoSeries` gets us *most* of the way there
- Settled on a bit of a creative solution

### DIFF
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -311,7 +311,7 @@ def from_native(  # noqa: D417
     eager_only: bool = False,
     series_only: bool = False,
     allow_series: bool | None = None,
-    **kwds: Any,  # noqa: ARG001
+    **kwds: Any,
 ) -> LazyFrame[IntoLazyFrameT] | DataFrame[IntoFrameT] | Series[IntoSeriesT] | T:
     """Convert `native_object` to Narwhals Dataframe, Lazyframe, or Series.
 
@@ -357,6 +357,9 @@ def from_native(  # noqa: D417
     pass_through = validate_strict_and_pass_though(
         strict, pass_through, pass_through_default=False, emit_deprecation_warning=True
     )
+    if kwds:
+        msg = f"from_native() got an unexpected keyword argument {next(iter(kwds))!r}"
+        raise TypeError(msg)
 
     return _from_native_impl(  # type: ignore[no-any-return]
         native_object,
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -50,6 +50,7 @@ if TYPE_CHECKING:
     from narwhals.typing import IntoLazyFrameT
     from narwhals.typing import IntoSeries
     from narwhals.typing import IntoSeriesT
+    from narwhals.typing import SeriesT
 
 T = TypeVar("T")
 
@@ -126,6 +127,10 @@ def to_native(
         msg = f"Expected Narwhals object, got {type(narwhals_object)}."
         raise TypeError(msg)
     return narwhals_object
+
+
+@overload
+def from_native(native_object: SeriesT, **kwds: Any) -> SeriesT: ...
 
 
 @overload
@@ -298,7 +303,7 @@ def from_native(
 ) -> Any: ...
 
 
-def from_native(
+def from_native(  # noqa: D417
     native_object: IntoLazyFrameT | IntoFrameT | IntoSeriesT | IntoFrame | IntoSeries | T,
     *,
     strict: bool | None = None,
@@ -306,6 +311,7 @@ def from_native(
     eager_only: bool = False,
     series_only: bool = False,
     allow_series: bool | None = None,
+    **kwds: Any,  # noqa: ARG001
 ) -> LazyFrame[IntoLazyFrameT] | DataFrame[IntoFrameT] | Series[IntoSeriesT] | T:
     """Convert `native_object` to Narwhals Dataframe, Lazyframe, or Series.
 
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -37,6 +37,7 @@ if TYPE_CHECKING:
     class NativeLazyFrame(NativeFrame, Protocol):
         def explain(self, *args: Any, **kwargs: Any) -> Any: ...
 
+    # TODO @dangotbanned: `nw.Series` **cannot** be allowed to match this!!!
     class NativeSeries(Sized, Iterable[Any], Protocol):
         def filter(self, *args: Any, **kwargs: Any) -> Any: ...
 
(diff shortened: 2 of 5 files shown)


tests/translate/from_native_test.py
```diff
@@ -367,3 +367,33 @@ def test_from_native_lazyframe() -> None:
...
+        nw_series_depth_2 = unstable_nw.Series(nw_series, level="full")
+        # NOTE: Checking that the type is `Series[Unknown]`
+        assert_type(nw_series_depth_2, unstable_nw.Series)
+
+        # TODO @dangotbanned: Fix this one
+        # Current:
+        #   `unstable_nw.Series[unstable_nw.Series[pl.Series]]``
+        # Goal:
+        #   `unstable_nw.Series[pl.Series]`
+        assert_type(nw_series_passthrough, unstable_nw.Series[pl.Series])
```
dangotbanned, line 399: ## FIXME
What we need is for `from_native` to return the same type on these branches

https://github.com/narwhals-dev/narwhals/blob/d2a9f423688c14e524d28e7d282bfa5de3c888be/narwhals/translate.py#L387-L391

*Instead*, this introduces a level of nesting

dangotbanned: Ignore `mypy` (https://github.com/narwhals-dev/narwhals/actions/runs/14271920405/job/40006873171?pr=2347)

I'm working on resolving this one

image

narwhals/typing.py
```diff
@@ -37,6 +37,7 @@ def join(self, *args: Any, **kwargs: Any) -> Any: ...
     class NativeLazyFrame(NativeFrame, Protocol):
         def explain(self, *args: Any, **kwargs: Any) -> Any: ...
 
+    # TODO @dangotbanned: `nw.Series` **cannot** be allowed to match this!!!
```
dangotbanned, line 40: `NativeSeries` is now what is used for `IntoSeries`.

An issue - that's new to me - is that `Series` **also matches** `NativeSeries` 🤦‍♂️ 

So for the original alias:

```py
IntoSeries: TypeAlias = Union["Series[Any]", "NativeSeries"]
```

`nw.Series` can be a match in *3 positions*
```py
IntoSeries: TypeAlias = Union["Series[Any]", "NativeSeries"]
#                             ^--1--^ ^2^    ^-----3-----^
```

tests/translate/from_native_test.py
dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/2347/commits/6bd47b49ed8259126d286964b6a0110c96e340e8) is a very heavy-handed fix.

My thinking is
- Passing a `narwhals` object to `from_native` really should never have been supported
  - None of these are native types: `nw.Series | nw.DataFrame | nw.LazyFrame`
- Removing that support would likely break someones workflow
- **Ignoring all flags** requires only adding **1** new `@overload` per-class
  - None of those need to deal with recursion

dangotbanned: Was expecting the `v1` backport (https://github.com/narwhals-dev/narwhals/pull/2347/commits/253da29bc8e02d71c2af411661639bbe67634b93) would surface some downstream issues - but all seems fine 🤔

narwhals/translate.py
```diff
@@ -128,6 +129,10 @@ def to_native(
     return narwhals_object
 
 
+@overload
+def from_native(native_object: SeriesT, **kwds: Any) -> SeriesT: ...
```
dangotbanned, line 133: ## But why does this work?
By having an **unconditional** `@overload` when a `nw.Series` is passed - it ensures no other `@overload`(s) are checked.

If we were to spell out all the valid cases, they would overlap with the `NativeSeries` cases (see https://github.com/narwhals-dev/narwhals/pull/2347#discussion_r2029314787)

```py
narwhals.from_native(native_series, allow_series=True)
narwhals.from_native(native_series, series_only=True)
```

Needing to have `**kwds` is *unfortunate* - [but](https://peps.python.org/pep-0020/)

> practicality beats purity

I'm open to any other solutions that can pass the new tests 🙂

MarcoGorelli: shall we raise at runtime `if kwds`?

MarcoGorelli (commented):
thanks @dangotbanned ! just one comment

narwhals/translate.py
dangotbanned: > shall we raise at runtime `if kwds`?

Yeah I was thinking about that as well!

I wasn't sure
- what the message should be?
- could that logic be shared for both `v1` and `main`?

narwhals/translate.py
```diff
@@ -351,6 +357,9 @@ def from_native(
     pass_through = validate_strict_and_pass_though(
         strict, pass_through, pass_through_default=False, emit_deprecation_warning=True
     )
+    if kwds:
+        msg = f"from_native() got an unexpected keyword argument {next(iter(kwds))!r}"
+        raise TypeError(msg)
```
dangotbanned, line 362: > [!IMPORTANT]
> This matches the runtime behavior of `python`.

 ```py
import polars as pl
import narwhals as nw

df = pl.DataFrame({"a": ["A", "B", "A"]})
nw_df = nw.from_native(df)
nw.to_native(nw_df, keyword_1="a", keyword_2="b")
```

Only the first bad argument is ever included:
```py
Cell In[5], line 39
     37 df = pl.DataFrame({"a": ["A", "B", "A"]})
     38 nw_df = nw.from_native(df)
---> 39 nw.to_native(nw_df, keyword_1="a", keyword_2="b")

TypeError: to_native() got an unexpected keyword argument 'keyword_1'
```

narwhals/translate.py
dangotbanned: @MarcoGorelli resolved in:

- (https://github.com/narwhals-dev/narwhals/pull/2347/commits/124239a54ec2460ef23ecbc817d6fcaa5e85e843)
- (https://github.com/narwhals-dev/narwhals/pull/2347/commits/8f6d90aebf7bee5d6ba0dfa55c525a2cef54bb97)
- (https://github.com/narwhals-dev/narwhals/pull/2347#discussion_r2029839147)

MarcoGorelli (approved):
thanks @dangotbanned !

PR #2358, opened 2025-04-08
fix: noxfile.py & better align it against GitHub Actions CI

camriddell:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

1. fixes `noxfile.py` as it had improper quoting of arguments that are passed to a subprocess leading to errors.
2. aligns the contents `noxfile.py` against the current CI in GitHub Actions `.github/workflows/*`
    - pytest for coverage tests
    - minimum/pretty_old/not_so_old
    - nightly (isn't a perfect replication due to our current Polars nightly build process)
    - random (slightly refactored `utils/generate_random_versions.py`)
        - Probably not incredibly useful for local work
        - side note—this should probably be seed dependent for reproducibility or print out its installed requirements within nox.

@MarcoGorelli I wasn't able to replicate your nightly release for Polars (which is fine) just wanted to ensure that was explicitly pointed out during review.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from tempfile import NamedTemporaryFile
 from typing import TYPE_CHECKING
 
 import nox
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -8,80 +9,218 @@
     from nox.sessions import Session
 
 nox.options.default_venv_backend = "uv"
-nox.options.reuse_venv = "yes"
+nox.options.reuse_venv = "never"
 
-PYTHON_VERSIONS = ["3.8", "3.9", "3.10", "3.11", "3.12"]
+PYTHON_VERSIONS = {
+    "pytest": ["3.8", "3.10", "3.11", "3.12", "3.13"],
```
EdAbati, line 15: I noticed this too. In CI we test on python `3.9` only in the `random` test.

Is it something deliberate?

noxfile.py
```diff
@@ -8,80 +9,218 @@
...
+
+    elif session.python in {"3.11", "3.13"}:
+        session.install(
+            "-e", ".[modin, dask]", "--group", "core-tests", "--group", "extra"
+        )
+        session.install("-U", "--pre", "duckdb")
+        if session.python != "3.13":
+            with NamedTemporaryFile() as f:
+                f.write(b"setuptools<78\n")
+                session.install("-b", f.name, "-e", ".[pyspark]")
```
EdAbati, line 64: Out of curiosity, why do we need to create a file for setuptools?

Maybe we could add the reason in a comment too?

utils/generate_random_versions.py
```diff
@@ -57,16 +58,13 @@
...
-content = f"pandas=={pandas_version}\nnumpy=={numpy_version}\npolars=={polars_version}\npyarrow=={pyarrow_version}\n"
-with open("random-requirements.txt", "w") as fd:
-    fd.write(content)
-
-with open("pyproject.toml") as fd:
-    content = fd.read()
-content = content.replace(
-    'filterwarnings = [\n  "error",\n]',
-    "filterwarnings = [\n  \"error\",\n  'ignore:distutils Version classes are deprecated:DeprecationWarning',\n]",
-)
```
EdAbati, line 70: is this not needed anymore?

EdAbati (commented):
Thank you for doing this!

It looks good to me, I just let a couple of questions below.

Unfortunately on my Mac when I run `nox` `tests/series_only/hist_test.py` fails:

```
FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pandas[pyarrow]] - ExceptionGroup: Hypothesis found 2 distinct failures. (2 sub-exceptions)
FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pandas[nullable]] - ExceptionGroup: Hypothesis found 2 distinct failures. (2 sub-exceptions)
FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pyarrow] - AssertionError: Mismatch at index 0: None != 1.0
```

It doesn't happen if I simply run `pytest` in the virtual env. (all library versions look the same)
Did you see something similar?

MarcoGorelli:
thanks for your PR!

tbh i think local development comes with so many personal preferences that i'm half-inclined to not have a nox file at all and leave people free to set up their own local testing config how they wish. then we don't need to maintain it

noxfile.py
camriddell: This was (naively) copied from the current CI:
https://github.com/narwhals-dev/narwhals/blob/4144497ebd9f4d20b4ef1c99b784d59ed5ab6e1d/.github/workflows/pytest.yml#L89-L92

Perhaps @MarcoGorelli remembers something here?

noxfile.py
camriddell: Tracked back to this commit: https://github.com/narwhals-dev/narwhals/commit/e718d1f100bc38f8fa84112ef0e589d37fdb4717

My guess is that it mainly reduces the time it takes for tests to run without sacrificing too much validity.
Though I'll kick this one to @MarcoGorelli as well if he remembers.

utils/generate_random_versions.py
camriddell: Ah thank you! I was going to leave a comment for @MarcoGorelli about this.

The matching string doesn't exist in the `pyproject.toml` file any more (note the position of the closing square bracket `]`) so this doesn't replace anything.

https://github.com/narwhals-dev/narwhals/blob/4144497ebd9f4d20b4ef1c99b784d59ed5ab6e1d/pyproject.toml#L204-L227

camriddell:
> [#2358 (comment)](https://github.com/narwhals-dev/narwhals/pull/2358#issuecomment-2785653929)
> 
> Feeling the same mostly 🤔
> 
> > then we don't need to maintain it
> 
> I think this PR could be _more compelling_ if we were to define [conflicting dependency groups](https://docs.astral.sh/uv/concepts/projects/config/#conflicting-dependencies) in our `pyproject.toml`.
> 
> For example, if `minimum`/`minimum_versions` were defined in one place - then having a `noxfile.py` would have a lower maintenance burden.
> 
> ```python
> @nox.session(python=PYTHON_VERSIONS["minimum"])
> def minimum(session: Session) -> None:
>     session.install(
>         "pandas==0.25.3",
>         "polars==0.20.3",
>         "numpy==1.17.5",
>         "pyarrow==11.0.0",
>         "pyarrow-stubs<17",
>         "scipy==1.5.0",
>         "scikit-learn==1.1.0",
>         "duckdb==1.0",
>         "tzdata",
>         "backports.zoneinfo",
>     )
>     session.install("-e", ".", "--group", "tests")
> ```
> 
> Even this would be simpler, since it could just target some combination of groups
> 
> https://github.com/narwhals-dev/narwhals/blob/01bf4ac6fca244ea697c34c2993bc5899e7e0bd9/.github/workflows/extremes.yml#L27-L43

Wholeheartedly agree here. I wanted to get something up quickly since `noxfile.py` was fully broken in its current state.

I'd like to take that next step of merging the metadata options (dependency groups, coverage values, etc.) into the pyproject.toml and having both nox and GH CI use that as a single-source of truth. 

That said, if @MarcoGorelli is inclined to remove `noxfile.py` the perhaps we should start there and I'll hold onto this for personal use.

camriddell:
> Thank you for doing this!
> 
> It looks good to me, I just let a couple of questions below.
> 
> Unfortunately on my Mac when I run `nox` `tests/series_only/hist_test.py` fails:
> 
> ```
> FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pandas[pyarrow]] - ExceptionGroup: Hypothesis found 2 distinct failures. (2 sub-exceptions)
> FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pandas[nullable]] - ExceptionGroup: Hypothesis found 2 distinct failures. (2 sub-exceptions)
> FAILED tests/series_only/hist_test.py::test_hist_count_hypothesis[pyarrow] - AssertionError: Mismatch at index 0: None != 1.0
> ```
> 
> It doesn't happen if I simply run `pytest` in the virtual env. (all library versions look the same) Did you see something similar?

I had `nox` tests pass locally on my end—so maybe there is some flakiness in the hypothesis/hist tests that hasn't been picked up by existing CI? I'm going to dive back into the `hist` implementations anyways to factor in the new hist behavior in Polars so I'll take a look at this failure then as well.

Was there a specific test group causes the failure here for you (ie. one of the pytest_coverage, minimum/old/not_so_old/nightly)? Or was this present in each in all of them?

dangotbanned:
Just wanted to note that I didn't touch `noxfile.py` when dropping `3.8` as this PR was open and fixing it

- #2679

PR #2361, opened 2025-04-08
Fix issue with missing TypeVar when using older versions of typing_extensions

mnxoid:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

It's a simple fix so I didn't go through the hassle of setting up the environment, please feel free to reformat if you already have it installed.
The problem I'm addressing is an import error when the version of `typing_extensions` present is below 4.4.0 where `TypeVar` was added. I do so by using the else-clause fallback which worked in my case (I monkey-patched the `typing_extensions` package to make my HoloViews visualizations work again)

### DIFF
--- a/narwhals/_translate.py
+++ b/narwhals/_translate.py
@@ -16,11 +16,12 @@ if TYPE_CHECKING:
 
 else:  # pragma: no cover
     import sys
+    from importlib.metadata import version
     from importlib.util import find_spec
 
     if sys.version_info >= (3, 13):
         from typing import TypeVar
-    elif find_spec("typing_extensions"):
+    elif find_spec("typing_extensions") and version("typing_extensions") >= "4.4.0":
         from typing_extensions import TypeVar
     else:
         from typing import TypeVar as _TypeVar


dangotbanned:
Thanks for the PR @mnxoid!

Should be able to resolve any *possible* issues and get this merged today 🎉

dangotbanned (approved):
Thanks again @mnxoid

MarcoGorelli:
thanks @mnxoid !

tbh i don't think we should be importing from typing_extensions at runtime at all, will take a look later in the week

dangotbanned:
> tbh i don't think we should be importing from typing_extensions at runtime at all, will take a look later in the week

@MarcoGorelli just for some more context on why I opted for it there.

In this PR, you can see how `TypeVar` changed as a result of [PEP 696](https://peps.python.org/pep-0696/)

- https://github.com/python/typeshed/pull/11925/files

`TypeVar`(s) in that module **must** be available at runtime - as they are used as a type param when [subclassing](https://peps.python.org/pep-0696/#subclassing).
Therefore, I wanted to make a *best-effort* to preserve the **intended** runtime type when it was safe to do so.

I just missed the case @mnxoid ran into with an old version of `typing_extensions` 🤦‍♂️

MarcoGorelli:
All good, thanks for explaining!

Issue #2348, opened 2025-04-05
`hist` hypothesis failure in CI


MarcoGorelli:
```
FAILED tests/series_only/hist_test.py::test_hist_bin_hypotheis[pandas[nullable]] - AssertionError: Mismatch at index 0: 0 != 1
Expected: {'breakpoint': [2.0], 'count': [1]}
Got: {'breakpoint': [2.0], 'count': [0]}
Falsifying example: test_hist_bin_hypotheis(
    constructor_eager=pandas_nullable_constructor,
    data=[1.0],
    bin_deltas=[1.0, 1.0],
)
Explanation:
    These lines were always and only run by failing examples:
        /home/runner/work/narwhals/narwhals/tests/utils.py:122

You can reproduce this example by temporarily adding @reproduce_failure('6.130.8', b'AXicY9Sw/8AAAYxwJoLFAABLvgQJ') as a decorator on your test case
FAILED tests/series_only/hist_test.py::test_hist_bin[polars[eager]-False-params1] - AssertionError: Mismatch at index 0: 2 != 1
Expected: {'count': [1, 3, 1]}
Got: {'count': [2, 3, 1]}
FAILED tests/series_only/hist_test.py::test_hist_bin_hypotheis[pyarrow] - AssertionError: Mismatch at index 0: 0 != 1
Expected: {'breakpoint': [2.0], 'count': [1]}
Got: {'breakpoint': [2.0], 'count': [0]}
Falsifying example: test_hist_bin_hypotheis(
    constructor_eager=pyarrow_table_constructor,
    data=[1.0],
    bin_deltas=[1.0, 1.0],
)
Explanation:
    These lines were always and only run by failing examples:
        /home/runner/work/narwhals/narwhals/tests/utils.py:122

You can reproduce this example by temporarily adding @reproduce_failure('6.130.8', b'AXicY9Sw/8AAAYxwJoLFAABLvgQJ') as a decorator on your test case
FAILED tests/series_only/hist_test.py::test_hist_bin_hypotheis[pandas] - AssertionError: Mismatch at index 0: 0 != 1
Expected: {'breakpoint': [2.0], 'count': [1]}
Got: {'breakpoint': [2.0], 'count': [0]}
Falsifying example: test_hist_bin_hypotheis(
    constructor_eager=pandas_constructor,
    data=[1.0],
    bin_deltas=[1.0, 1.0],
)
Explanation:
    These lines were always and only run by failing examples:
[...]

PR #2362, opened 2025-04-08
feat(typing): Add `polars` python type aliases

### LINKED ISSUE #2360: [enh]: Add `polars._typing` python type aliases
https://github.com/pola-rs/polars/blob/bab7cefdb987972571542757d1f55d04db9cd087/py-polars/polars/_typing.py#L85-L89

```py
NumericLiteral: TypeAlias = Union[int, float, "Decimal"]
TemporalLiteral: TypeAlias = Union["date", "time", "datetime", "timedelta"]
NonNestedLiteral: TypeAlias = Union[NumericLiteral, TemporalLiteral, str, bool, bytes]
# Python literal types (can convert into a `lit` expression)
PythonLiteral: TypeAlias = Union[NonNestedLiteral, list[Any]]
```

dangotbanned:
Closes #2360

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2360
- Parent issue: #2337

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- [x] Main changes
- [x] Review `ndarray` support (#588)

After adding the aliases themselves, I went through and filled some gaps we have in our typing with them:
- `CompliantExpr.clip` was entirely missing
- Public API (and their internal counterparts)
  - `nw.(Expr|Series).clip`
  - `nw.(Expr|Series).fill_null`
  - `nw.lit`
  - `nw.when().then(...).otherwise(...)`
- `_expression_parsing.py`
  - A lot of these were called by the other functions/methods
  - Which meant their parameters types could be more narrowly defined
- `_arrow.utils.py`
  - And usage of those functions

I'm pretty happy with the end result, we've got fewer places typed as `Any` or `object` - which makes the code easier to reason with if you're unfamiliar


tests/expr_and_series/when_test.py
```diff
@@ -96,9 +99,9 @@ def test_value_expression(constructor: Constructor) -> None:
 def test_otherwise_numpy_array(constructor_eager: ConstructorEager) -> None:
     df = nw.from_native(constructor_eager(data))
 
-    result = df.select(
-        nw.when(nw.col("a") == 1).then(-1).otherwise(np.array([0, 9, 10])).alias("a_when")
-    )
+    arr: _1DArray = np.zeros([3], np.dtype(np.int64))
+    arr[:3] = 0, 9, 10
+    result = df.select(nw.when(nw.col("a") == 1).then(-1).otherwise(arr).alias("a_when"))
```
dangotbanned, line 104: @MarcoGorelli I noticed you've done some work on the `numpy` stubs recently.

Do you know of a more reliable (and readable 🙃) way to create an array with this type?



```py

import numpy as np

if TYPE_CHECKING:

PlzLetMeBeAnOverload: TypeAlias = np.ndarray[tuple[int], np.dtype[np.int64]]
```

I can only seem to get there via `np.arange`, but the test has elements `(0, 9, 10)`.

`np.zeros` would work for `float64`, but can't seem to match an `int`-related overload 😞

dangotbanned:
I'm not sure if these aliases need documenting *yet*.

Quite happy with what their introduction has led to *outside* of the public changes.
I've found it much easier to reason with things, knowing that `object` or `Any` really just meant one of these:

https://github.com/narwhals-dev/narwhals/blob/418162630b9892907b99da4490af174f024f43bd/narwhals/typing.py#L307-L312

dangotbanned:
Found some more places to use the new aliases https://github.com/narwhals-dev/narwhals/pull/2362#event-17193702490

dangotbanned:
CI failures aren't related to these changes, see
- https://github.com/narwhals-dev/narwhals/issues/2348#issuecomment-2790558234

dangotbanned:
@MarcoGorelli I've updated the description to try and make the review more approachable 🙂

tests/expr_and_series/when_test.py
MarcoGorelli: no sure i follow what you're asking, sorry 😳

MarcoGorelli (approved):
thanks @dangotbanned !

tests/expr_and_series/when_test.py
dangotbanned: Ah my bad, didn't really explain it well.

I wanted to call a `numpy` function and have `mypy` understand that it is a `_1DArray`.

Really struggled to get anything working, which is why the section this comment is under looks very odd

PR #2363, opened 2025-04-08
fix: Preserve pandas column name attribute

### LINKED ISSUE #1483: bug: `select` and `with_columns` don't always preserve column index name for pandas-like
```python
import pandas as pd
# import narwhals.stable.v1 as nw
import narwhals as nw

df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
df.columns.name = 'foo'

print(nw.from_native(df).select(c=nw.col('a')+1).to_native())
print(nw.from_native(df).with_columns(c=nw.col('a')+1).to_native())
print(nw.from_native(df).to_native())
print(nw.from_native(df).to_native().assign(c=lambda df: df['a']+1))
```

outputs
```python
   c
0  2
1  3
2  4
   a  b  c
0  1  4  2
1  2  5  3
2  3  6  4
foo  a  b
0    1  4
1    2  5
2    3  6
foo  a  b  c
0    1  4  2
1    2  5  3
2    3  6  4
```

expected:
```python
foo  c
0    2
1    3
2    4
foo  a  b  c
0    1  4  2
1    2  5  3
2    3  6  4
foo  a  b
0    1  4
1    2  5
2    3  6
foo  a  b  c
0    1  4  2
1    2  5  3
2    3  6  4
```

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1483 

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/_dask/dataframe.py
```diff
@@ -110,7 +110,7 @@ def collect(
             from narwhals._pandas_like.dataframe import PandasLikeDataFrame
 
             return PandasLikeDataFrame(
-                result,
+                result.rename_axis(columns=self.native.columns.name),
```
FBruzzesi, line 113: This might be one step ahead of [dask#11874](https://github.com/dask/dask/issues/11874)

narwhals/_pandas_like/dataframe.py
```diff
@@ -248,7 +249,7 @@ def _with_version(self: Self, version: Version) -> Self:
 
     def _with_native(self: Self, df: Any, *, validate_column_names: bool = True) -> Self:
         return self.__class__(
-            df,
+            df.rename_axis(columns=self._native_columns_name),
```
FBruzzesi, line 252: Should we use `rename_axis(..., copy=False)`?

narwhals/_pandas_like/namespace.py
FBruzzesi, line 1: Can rollback this, or compress it even more

narwhals/translate.py
```diff
@@ -488,8 +488,8 @@ def _from_native_impl(  # noqa: PLR0915
         return DataFrame(
             PandasLikeDataFrame(
                 native_object,
-                backend_version=parse_version(pd),
                 implementation=Implementation.PANDAS,
+                backend_version=parse_version(pd),
```
FBruzzesi, line 492: Mental sanity for symmetry with other pandas-like and order of the spec 😅

tests/preserve_pandas_like_columns_name_attr_test.py
```diff
@@ -0,0 +1,28 @@
...
+            reason="Dataframe columns is a list and do not have a `name` like a pandas Index does"
+        )
+
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8.0, 9.0]}
+    df_native = constructor(data)
+    df_native.columns.name = "foo"  # type: ignore[union-attr]
+
+    df = nw.from_native(df_native)
+
+    result = df.with_columns(b=nw.col("a") + 1, c=nw.col("a") * 2).select("c", "b")
```
FBruzzesi, line 25: I wanted to concatenate two methods here, mostly for the sake of it.
Might be worth reparametrizing it

narwhals/_dask/dataframe.py
FBruzzesi, line 1: Dask `select` and `with_columns` have no issues since we use `.assign` method for both!

MarcoGorelli:
thanks! i'll do a little test to check there's no accidental / surprise overhead

narwhals/_dask/dataframe.py
```diff
@@ -112,7 +113,12 @@ def collect(
...
             return PandasLikeDataFrame(
-                result,
+                rename_axis(
+                    result,
+                    implementation=Implementation.PANDAS,
+                    backend_version=parse_version(pd),
+                    columns=self.native.columns.name,
+                ),
                 implementation=Implementation.PANDAS,
                 backend_version=parse_version(pd),
```
dangotbanned, line 123:
```suggestion
            implementation = Implementation.PANDAS
            backend_version = parse_version(pd)
            return PandasLikeDataFrame(
                rename_axis(
                    result,
                    implementation=implementation,
                    backend_version=backend_version,
                    columns=self.native.columns.name,
                ),
                implementation=implementation,
                backend_version=backend_version,
```

narwhals/_pandas_like/utils.py
```diff
@@ -130,35 +129,6 @@ def align_and_extract_native(
...
-            else {}
-        )
-        return implementation.to_native_namespace().concat(dfs, axis=1, **extra_kwargs)
-
-    else:  # pragma: no cover
-        msg = f"Expected pandas-like implementation ({PANDAS_LIKE_IMPLEMENTATION}), found {implementation}"
-        raise TypeError(msg)
-
-
 def vertical_concat(
```
dangotbanned, line 132: Happy to split this (+ the other `*_concat`) into another PR 🙂 

Just thought it would have a bigger impact than (https://github.com/narwhals-dev/narwhals/pull/2363/files/f28599f7189b5a2128a5cce17675808f63c9a22b#r2034071157) alone

FBruzzesi: Yes I think it would be cleaner! I shouldn't have edited `concat` - I didn't realize immediately that `pd.concat` does not maintain the dataframe columns name

dangotbanned: No worries @FBruzzesi, moving it now 👍

FBruzzesi:
Messed up the merge 🥲

MarcoGorelli:
thanks!

> Let me know what your thoughts are and if we need to check something more extensively

i'm still slightly concerned about there being extra copies in old versions, i'll take another look

FBruzzesi:
> thanks!
> 
> > Let me know what your thoughts are and if we need to check something more extensively
> 
> i'm still slightly concerned about there being extra copies in old versions, i'll take another look

No rush! I keep merging main to avoid having to deal with a large conflicting diff if that happens to be the case

narwhals/_pandas_like/dataframe.py
```diff
@@ -251,7 +253,12 @@ def _with_version(self, version: Version) -> Self:
 
     def _with_native(self, df: Any, *, validate_column_names: bool = True) -> Self:
         return self.__class__(
-            df,
+            rename_axis(
+                df,
+                implementation=self._implementation,
+                backend_version=self._backend_version,
+                columns=self._native_columns_name,
+            ),
```
MarcoGorelli, line 261: 🤔 i'm wondering if we could/should just set `df.columns.name = self._native_columns_name`

else we need to trust pandas' copy-on-write / `copy=False` working properly, which i'm not sure i do

FBruzzesi: I trust your judgement here - but should I would be concerned as a pandas user?

MarcoGorelli (approved):
thanks @FBruzzesi !

I don't trust pandas' `rename_axis`, nor do I trust that copy-on-write will take effect properly, we would need more comprehensive benchmarks to be sure

but, in `select` and `with_columns`, we create new objects, so it should be safe to just use `.columns.name = ...` there

PR #2366, opened 2025-04-09
ci: Run PySpark selectively

### LINKED ISSUE #2364: ci: PySpark is too slow, only run it selectively?
Our tests in CI currently take 7 minutes (!), largely because of PySpark

I think this is way too slow, especially given that we're mostly testing the same functionality with sqlframe

Our CI jobs often end up "queued", probably because we're using quite a lot of minutes

Perhaps we could have that pyspark tests only run:
- if we apply some "pyspark" label to PRs
- before releases

so we can keep the contributor experience smoother

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2364 

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/pytest-pyspark.yml
+++ b/.github/workflows/pytest-pyspark.yml
@@ -1,0 +1,37 @@
+name: PyTest PySpark constructor
+
+on:
+  pull_request:
+    types: [opened, labeled, synchronize]  # Triggers on PR creation, updates, and label changes
+
+env:
+  PY_COLORS: 1
+
+jobs:
+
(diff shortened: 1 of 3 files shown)


FBruzzesi:
Opened PR with `ci` label only:

image

then adding `pyspark` label:

image

PR #2368, opened 2025-04-10
refactor: Simplify `PandasLikeNamespace.concat`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Originally https://github.com/narwhals-dev/narwhals/pull/2363#discussion_r2037731196

## Checklist

- [ ] Tests added
- [x] Documented the changes

- [x] `horizontal_concat`
- [x] `vertical_concat`
- [x] `diagonal_concat`


narwhals/_pandas_like/namespace.py
```diff
@@ -223,48 +223,62 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
+    def _concat_diagonal(self, dfs: Sequence[NDFrameT], /) -> NDFrameT:
+        """Concatenate (native) DataFrames diagonally."""
+        concat = self._implementation.to_native_namespace().concat
+        if self._implementation.is_pandas() and self._backend_version < (3,):
+            if self._backend_version < (1,):
+                return concat(dfs, axis=0, copy=False, sort=False)
+            return concat(dfs, axis=0, copy=False)
+        return concat(dfs, axis=0)
+
+    def _concat_horizontal(self, dfs: Sequence[NDFrameT], /) -> NDFrameT:
```
FBruzzesi, line 235: Shouldn't the output type always be `pd.DataFrame` regardless of  input in `dfs`?

```suggestion
    def _concat_horizontal(self, dfs: Sequence[NDFrameT], /) -> pd.DataFrame:
```

Here a snippet:

```py
import pandas as pd

s1 = pd.Series([1, 2, 3])
s2 = pd.Series(["a", "b", "c"])

type(pd.concat([s1], axis=1))
# pandas.core.frame.DataFrame

type(pd.concat([s1, s2], axis=1))
# pandas.core.frame.DataFrame
```

FBruzzesi (commented):
Thanks @dangotbanned 🚀 I like this version a lot!
I might have spotted something in the type annotation?!

narwhals/_pandas_like/namespace.py
dangotbanned: Well spotted! 

So I added that type for `horizontal`, which gets used with a `Series` as well (IIRC in `group_by` maybe?).

`vertical` used some property that is only available on `DataFrame`, so I had to make that one narrower.

`diagonal` (I think) should work with either type, but is currently only used for `DataFrame`.

Replying from my phone, hope that all makes sense 😅

dangotbanned: Oh I see what you mean now!

dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2368#discussion_r2038333140
@FBruzzesi I think you'd also need to change the type of `dfs` to be:

```py
Sequence[pd.DataFrame] | Sequence[pd.Series[Any]]
```

If you kept the `TypeVar` in `pyright` will tell you off 😉

dangotbanned: @FBruzzesi quite happy with this now (https://github.com/narwhals-dev/narwhals/pull/2368/commits/b6d07114073faf31fe5e3f03ad998acd0fab0583)

Thanks for the `pandas` 🧠 

image

image


image

MarcoGorelli (approved):
thanks both

narwhals/_pandas_like/namespace.py
```diff
@@ -223,48 +227,66 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
+    def _concat(self):  # type: ignore[no-untyped-def] # noqa: ANN202
+        """Return the **native** equivalent of `pd.concat`."""
+        # NOTE: Leave un-annotated to allow `@overload` matching via inference.
+        if TYPE_CHECKING:
+            import pandas as pd
+
+            return pd.concat
+        return self._implementation.to_native_namespace().concat
+
+    def _concat_diagonal(self, dfs: Sequence[NDFrameT], /) -> pd.DataFrame:
```
MarcoGorelli, line 240: actually sorry, is the input to `concat_diagonal` ever not `pd.DataFrame`? can't this be `Sequence[pd.DataFrame]`?

dangotbanned: @MarcoGorelli (https://github.com/narwhals-dev/narwhals/pull/2368#discussion_r2038402018)

> So I added that type for `horizontal`, which gets used with a `Series` as well (IIRC in `group_by` maybe?).
> 
> `vertical` used some property that is only available on `DataFrame`, so I had to make that one narrower.
> 
> `diagonal` (I think) should work with either type, but is currently only used for `DataFrame`.

So the thing I was going for is annotating as wide as possible, since this *could* accept either.

I will admit it is strange to think *diagonal* for `Series` though 🤔

narwhals/_pandas_like/namespace.py
```diff
@@ -223,48 +227,66 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
+        if self._implementation.is_pandas() and self._backend_version < (3,):
+            if self._backend_version < (1,):
+                return self._concat(dfs, axis=VERTICAL, copy=False, sort=False)
+            return self._concat(dfs, axis=VERTICAL, copy=False)
+        return self._concat(dfs, axis=VERTICAL)
```
dangotbanned, line 245: @MarcoGorelli should `_concat_vertical` be matching this?

I just copied it over, but they are different:

https://github.com/narwhals-dev/narwhals/blob/657aef9d27e98b82c6f93ec1ad4751c598f7c285/narwhals/_pandas_like/namespace.py#L273-L275

narwhals/_pandas_like/namespace.py
MarcoGorelli: i'd say let's annotate according to current usage

dangotbanned: That also works 🙂

dangotbanned: Happy for you to make the change or I can get another commit up later?

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2368/commits/0aee5bf647cabd17121648b15936a879430fff5b)

PR #2371, opened 2025-04-11
chore: `@requires` for backend method constraints

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

Adds a method decorator for consistent error messages on unsupported versions

Main use case is for enforcing a minimum version, e.g.:

```py
class PolarsExpr:
    @requires.backend_version((1,))
    def rolling_var(
        self: Self, window_size: int, *, min_samples: int, center: bool, ddof: int
    ) -> Self: ...
```

Still have some other cases I wanna explore:
- [x] Providing a hint (`PolarsNamespace.nth`)

### `PandasLike*`
`.pivot` is a good example of multiple cases dependent on more than just `backend_version`:

https://github.com/narwhals-dev/narwhals/blob/d63ec65ef400f370dc1d56764cacdaaf84248fb9/narwhals/_pandas_like/dataframe.py#L1032-L1049

### Parameter combinations
I'd only want to include the *version-based* parameter restrictions on `@requires`:
- `DuckDBLazyFrame.join(how="cross")`
- `DuckDBExpr.broadcast(kind=ExprKind.AGGREGATION)`
- `PolarsDataFrame.__array__(copy=...)`
- `PolarsExpr.over(order_by=...)`

It seems there are more *blanket* parameter restrictions, but they don't make sense as a `@requires`:
- `DaskExpr.cum_*(reverse=True)`
- `DaskExpr.rolling_*(ddof=2)` (or not `1`)
- `(Dask|DuckDB)Expr.quantile(interpolation="(nearest|higher|lower|midpoint)")`
- `DuckDBLazyFrame.join_asof(strategy="nearest")`
- `(DuckDB|SparkLike)Expr.fill_null(strategy=...)`
- `(DuckDB|SparkLike)Expr.rank(method="(average|max|ordinal)")`
- `DuckDBExprStringNamespace.to_datetime(format=None)`

### How to document this?
Like with `not_implemented`, we could mark the methods with runtime metadata. Then consume that in a script for something like https://narwhals-dev.github.io/narwhals/api-completeness/


dangotbanned:
Been think about an alternative spelling which could be more easily adapted for other requirements.

<details><summary>Imports</summary>
<p>

```py

if TYPE_CHECKING:

    from narwhals._polars.expr import PolarsExpr
```

</p>
</details> 

## Current
### Decorator constructor
```py
class requires:  # noqa: N801
    def __init__(self, *, min_version: tuple[int, ...], hint: str = "") -> None:
        self._min_version: tuple[int, ...] = min_version
        self._hint: str = hint
        self._wrapped_name: str
```
### Usage
```py
class DuckDBExpr:
    @requires(min_version=(1, 3))
    def is_first_distinct(self) -> Self: ...

class PolarsNamespace:
    @requires(
        min_version=(1, 0, 0), hint="Please use `col` for columns selection instead."
    )
    def nth(self, *indices: int) -> PolarsExpr: ...
```

## Alternative
Would be pretty easy to add support for multi-`implementation` cases this way - without it affecting the more common usage.

I also like that we can stay aligned with the `backend_version` name, while still indicating it is a minimum

### Decorator constructor
```py
class requires:  # noqa: N801
    @classmethod
    def backend_version(cls, min_version: tuple[int, ...], /, hint: str = "") -> Self:
        obj = cls.__new__(cls)
        obj._min_version = min_version
        obj._hint = hint
        return obj
```

### Usage
```py
class DuckDBExpr:
    def is_first_distinct(self) -> Self: ...

class PolarsNamespace:
    @requires.backend_version(
        (1, 0, 0), "Please use `col` for columns selection instead."
    )
    def nth(self, *indices: int) -> PolarsExpr: ...
```

narwhals/utils.py
```diff
@@ -1779,3 +1803,67 @@ def _not_implemented_error(what: str, who: str, /) -> NotImplementedError:
...
+            minimum: Minimum backend version.
+            hint: Optional suggested alternative.
+
+        Returns:
+            An exception-raising decorator.
+        """
+        obj = cls.__new__(cls)
+        obj._min_version = minimum
+        obj._hint = hint
+        return obj
```
dangotbanned, line 1842: I want to switch every thing over to this style, see for comparison (https://github.com/narwhals-dev/narwhals/pull/2371#issuecomment-2799886235)

narwhals/utils.py
```diff
@@ -539,6 +542,27 @@ def is_sqlframe(self: Self) -> bool:
         """
         return self is Implementation.SQLFRAME  # pragma: no cover
 
+    @property
+    def _alias(self) -> LiteralString:
+        """Friendly name for errors.
```
FBruzzesi, line 547: Nice touch! This might be used in many other places in which we raise error specifically for a given implementation 💡

narwhals/utils.py
FBruzzesi: It looks pretty nice and more explicit 👌🏼

FBruzzesi (commented):
Thanks @dangotbanned , it looks very good!

I personally like the `requires.backend_version(...)` style, as it leaves the possibility to expand to further if needed. Aside such personal preference, I don't have anything to object with the PR itself :)

dangotbanned:
> Thanks @dangotbanned , it looks very good!
> 
> I personally like the `requires.backend_version(...)` style, as it leaves more room to expand if needed. Aside such personal preference, I don't have anything to object with the PR itself :)

Thanks @FBruzzesi 😍

narwhals/utils.py
dangotbanned: Thanks @FBruzzesi!

Yeah I noticed we seem to use those names in the errors I've replaced here and in the docstrings like [`Implementation.is_cudf`](https://narwhals-dev.github.io/narwhals/api-reference/implementation/#narwhals.Implementation.is_cudf)
But was suprised they hadn't been defined in one place yet

narwhals/utils.py
dangotbanned: Great I'll go with this one then for all of them 😄

narwhals/utils.py
FBruzzesi: I assume that in practice it rarely happened that a minimum version exception was raised as downstream libraries require higher versions than our minimum ones - at least that's my impression when working on some of them.

narwhals/utils.py
```diff
@@ -1808,22 +1808,28 @@ def _not_implemented_error(what: str, who: str, /) -> NotImplementedError:
...
-        An exception-raising decorator.
+    Examples:
+        >>> from narwhals.utils import requires, Implementation
+        >>> class SomeBackend:
+        ...     _implementation = Implementation.PYARROW
+        ...     _backend_version = 20, 0, 0
+        ...
+        ...     @requires.backend_version((9000, 0, 0))
+        ...     def really_complex_feature(self) -> str:
+        ...         return "hello"
```
FBruzzesi, line 1823: 😂 I had a god laugh

FBruzzesi (approved):
It's a yes from me!

DogHighFiveGIF

MarcoGorelli:
nice, thanks both!

PR #2384, opened 2025-04-13
perf: Only use first element to sniff types in sequences in `__getitem__` and `filter`

MarcoGorelli:
urrgh, I noticed there's some parts where we're iterating over objects which are used to filter/select rows, which we should never be doing

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1271,9 +1271,9 @@ def is_sequence_but_not_str(sequence: Any | Sequence[_T]) -> TypeIs[Sequence[_T]
     return isinstance(sequence, Sequence) and not isinstance(sequence, str)
 
 
-def is_boolean_list(obj: Any) -> TypeIs[list[bool]]:
-    # Check if an object is a list of booleans, only sniffing the first element.
-    return isinstance(obj, list) and obj and isinstance(obj[0], bool)  # type: ignore[return-value]
+def is_list_of(obj: Any, tp: type[_T]) -> TypeIs[list[type[_T]]]:
+    # Check if an object is a list of `tp`, only sniffing the first element.
+    return bool(isinstance(obj, list) and obj and isinstance(obj[0], tp))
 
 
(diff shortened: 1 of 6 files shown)


narwhals/utils.py
```diff
@@ -1271,6 +1271,11 @@ def is_sequence_but_not_str(sequence: Any | Sequence[_T]) -> TypeIs[Sequence[_T]
     return isinstance(sequence, Sequence) and not isinstance(sequence, str)
 
 
+def is_boolean_list(obj: Any) -> TypeIs[list[bool]]:
+    # Check if an object is a list of booleans, only sniffing the first element.
+    return isinstance(obj, list) and obj and isinstance(obj[0], bool)  # type: ignore[return-value]
```
dangotbanned, line 1276: I noticed there's another path where you check a similar thing for `str`.

Making the guard generic *could* let you reuse it for both 
```suggestion
def is_list_of(obj: Any, tp: type[_T]) -> TypeIs[list[type[_T]]]:
    # Check if an object is a list of `tp`, only sniffing the first element.
    return bool(isinstance(obj, list) and obj and isinstance(obj[0], tp))
```

```py
is_boolean_list(obj)
is_list_of(obj, bool)
```

dangotbanned: Similar function in https://github.com/pola-rs/polars/blob/cd3e4ee7995f95bfb186d6f9b323631db246df60/py-polars/polars/_utils/various.py#L87-L89

MarcoGorelli:
sure - thanks @dangotbanned !

PR #2391, opened 2025-04-16
perf: Avoid module-level `importlib.util.find_spec`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_translate.py
+++ b/narwhals/_translate.py
@@ -6,21 +6,6 @@ from typing import Iterable
 from typing import Mapping
 from typing import Protocol
 
-_MIN_TYPING_EXTENSIONS = 4, 4, 0
-_TYPING_EXTENSIONS = "typing_extensions"
-
-
-def _typing_extensions_has_pep_696() -> bool:  # pragma: no cover
-    from importlib.metadata import version
-    from importlib.util import find_spec
-
-    from narwhals.utils import parse_version
-
-    if find_spec(_TYPING_EXTENSIONS):
-        return parse_version(version(_TYPING_EXTENSIONS)) >= _MIN_TYPING_EXTENSIONS
-    return False
-
-
 if TYPE_CHECKING:
     import pyarrow as pa
     from typing_extensions import Self
@@ -34,8 +19,6 @@ else:  # pragma: no cover
 
     if sys.version_info >= (3, 13):
         from typing import TypeVar
-    elif _typing_extensions_has_pep_696():
-        from typing_extensions import TypeVar
     else:
         from typing import TypeVar as _TypeVar
 
--- a/tests/no_imports_test.py
+++ b/tests/no_imports_test.py
@@ -15,6 +15,7 @@ def test_polars(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.delitem(sys.modules, "pandas")
     monkeypatch.delitem(sys.modules, "numpy")
     monkeypatch.delitem(sys.modules, "pyarrow", raising=False)
+    monkeypatch.delitem(sys.modules, "typing_extensions", raising=False)
     monkeypatch.delitem(sys.modules, "duckdb", raising=False)
     monkeypatch.delitem(sys.modules, "dask", raising=False)
     monkeypatch.delitem(sys.modules, "ibis", raising=False)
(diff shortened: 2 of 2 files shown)


dangotbanned:
@MarcoGorelli could we slow down a bit on this one?

My read on this, was that the import wasn't the issue?

> > Sure, thanks - would it help if we just removed the runtime import from Narwhals anyway?
> 
> Sure, if you don't need it. (But dynamic import with `find_spec` is not really the problematic part - it is the metadata query. Because while our analysis will miss the reference to `typing_extensions` made via `find_spec`, there are plenty of direct `import typing_extensions` later on, which allow us to pick up the module - mostly because we do not handle `TYPE_CHECKING` as no-op codepath...).
> 
> You could also catch the `importlib.metadata.PackageNotFoundError` when calling `importlib.metadata.version(_TYPING_EXTENSIONS)` in [narwhals-dev/narwhals@`7059f5b`/narwhals/_translate.py#L20 (blame)](https://github.com/narwhals-dev/narwhals/blame/7059f5b77a0e57b5ec95c688f2072d4134bd4858/narwhals/_translate.py#L20).


I just wanna be sure I understand the issue fully first

dangotbanned:
@MarcoGorelli okay it seems we're all good!


https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/902#issuecomment-2810594839
> As far as I'm concerned, there's no need to change anything on your side.

I followed up and checked, the fix they've added is pretty standard and they've used it in lots of popular packages
- https://github.com/search?q=repo%3Apyinstaller/pyinstaller-hooks-contrib%20copy_metadata&type=code

MarcoGorelli:
sure, thanks, should we just address

> You could also catch the importlib.metadata.PackageNotFoundError when calling importlib.metadata.version(_TYPING_EXTENSIONS) in

in case that affects anyone else?

dangotbanned:
> sure, thanks, should we just address
> 
> > You could also catch the importlib.metadata.PackageNotFoundError when calling importlib.metadata.version(_TYPING_EXTENSIONS) in
> 
> in case that affects anyone else?

@MarcoGorelli how would you feel about making that a *sub-task* of adding a `_typing_compat` module?
- https://docs.astral.sh/ruff/settings/#lint_typing-modules
- https://mypy.readthedocs.io/en/stable/config_file.html#confval-custom_typing_module
- `pyright` seems to not need config

We could have that extra safety there, **plus**:
- A [single](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/stable/v1/__init__.py#L113-L117) place to handle `TypeVar` defaults
- Cut down on the [`Protocol38`](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/_compliant/expr.py#L41-L49) [boilerplate](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/_compliant/group_by.py#L30-L38) [that](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/_compliant/when_then.py#L33-L41) [I've littered](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/_compliant/selectors.py#L23-L34)
- Fun stuff like [`@deprecated`](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/utils.py#L1679-L1694)

MarcoGorelli:
I don't really mind, but based  on https://github.com/narwhals-dev/narwhals/issues/2479 I still think we may want to consider removing this - the `else` branch you've added seems to cover pre-3.13 behaviour perfectly

dangotbanned (approved):
> I don't really mind, but based on #2479 I still think we may want to consider removing this
Thanks @MarcoGorelli!

(#2479) points towards a cost that we can **only** avoid by not doing the `find_spec` call.
Since a `TypeVar` **must** be defined at the module-level (*tried to find a ref but had no luck* 😔) the only option we have is to use the `else` branch.

Appreciate you bearing with me on this 😅 


I'd still like to clean up the bits mentioned in (https://github.com/narwhals-dev/narwhals/pull/2391#issuecomment-2832442331) soon - but we might as well get the perf boost in now 🎁

MarcoGorelli:
sure! thanks a tonne for your attention to detail, here and everywhere

Issue #2389, opened 2025-04-15
[Enh]: allow slicing LazyFrame


gdementen:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

I am currently creating adapters so that my Qt-based UI (https://github.com/larray-project/larray-editor/) becomes a lot more generic instead of being only useful for our own project. The goal is to make it able to display any grid-like structure (arrays, dataframes, or even some files). 

### Please describe the purpose of the new feature or describe the problem to solve.

For obvious performance reasons, it is much faster to only compute what is actually visible on the screen, so my adapters try to do exactly that. I implemented that easily using the direct Polars API for both DataFrame and LazyFrame (and Ibis Table) with great result. However, when using the Narwhals Lazy API (on top of a Polars LazyFrame), I have not found any direct way to select specific rows other than top rows. I had to resort to an ugly workaround which works but makes things uncomfortably slow :

```python
wh_index = data.with_row_index('_index')
filter_ = (nw.col('_index') >= v_start) & (nw.col('_index') < v_stop)
# .select also implicitly drops _index
lazy_sub_df = wh_index.filter(filter_).select(columns[h_start:h_stop])
```

I know of https://narwhals-dev.github.io/narwhals/basics/order_dependence/ and the fact that lazyFrame.tail() is deprecated but I would advocate for implementing slicing anyway. After all, even if one cannot guarantee the order (unless an explicit sort was used) and the order is arbitrary, it makes sense to paginate the result. Another argument is that all SQL engines I know of support some kind of offset command for pagination (whether via limit/offset or offset fetch which is standard SQL AFAIK). Some SQL engine require sorting, for offset to work, but in my (limited) experience this is rather the exception than the rule.

### Suggest a solution if possible.

One option might be to allow slicing only if the query includes a sort operation. It would not help my specific case though.

MarcoGorelli:
thanks @gdementen for the request

could you clarify how you do this for polars please? using `.head` / `tail`?

gdementen:
Polars supports slicing LazyFrames directly, so this is enough:
```python
data[v_start:v_stop].select(columns[h_start:h_stop])
```

MarcoGorelli:
thanks, i hadn't realised that `slice` was supported in `LazyFrame.__getitem__`

[...]

PR #2325, opened 2025-04-01
feat: Allow expressions as `group_by` keys

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #1385
- https://github.com/narwhals-dev/narwhals/issues/2225#issuecomment-2746592849
- https://github.com/narwhals-dev/narwhals/issues/2225#issuecomment-2747281275

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below

@MarcoGorelli there is [_one_ test failing](https://github.com/narwhals-dev/narwhals/blob/b940b2cbadd3d35d0f8813cbb5b1830f43480f2f/tests/frame/select_test.py#L57-L58) - which is grouping by a pandas column with a boolean name (and similarly with integer name in scikit-lego)

@dangotbanned I am crying for help with the typing 🙈

Other comment will address in the code itself

PR #2020, opened 2025-02-15
refactor(typing): Add `__getitem__` selector aliases

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

Discussed in: 
- https://github.com/narwhals-dev/narwhals/pull/1963#issuecomment-2645456279
- https://github.com/narwhals-dev/narwhals/pull/1963#issuecomment-2645527746

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- Just demonstrating what it looks like for a single backend (`pandas`)
  - Can do the others if anyone else is on board (*besides myself and @EdAbati*) 
- The goal is to have something that's *easier* to understand, than the current `@overload`(s)
- Stealing pretty directly from (https://github.com/pola-rs/polars/blob/4eb4fb7527d26d3ecf9b80cc4ad92809064f8c35/py-polars/polars/_typing.py#L281)

PR #2393, opened 2025-04-17
feat: Improve ``DataFrame.__getitem__`` consistency

### LINKED ISSUE #2405: enh: allow slicing `DataFrame` with native objects
Currently, the following works:
```python
import pandas as pd
import narwhals as nw
import numpy as np

df_nw = nw.from_native(pd.DataFrame({'a':[1,2,3], 'b': [4,5,6], 'c': [0,2,1]}), eager_only=True)
print(df_nw['a'][df_nw['c'].to_native()])
```
```
┌─────────────────────┐
|   Narwhals Series   |
|---------------------|
|0    1               |
|2    3               |
|1    2               |
|Name: a, dtype: int64|
└─────────────────────┘
````

However, this doesn't:
```python
df_nw = nw.from_native(pd.DataFrame({'a':[1,2,3], 'b': [4,5,6], 'c': [0,2,1]}), eager_only=True)
print(df_nw[df_nw['c'].to_native()])
```
```
  File "/home/marcogorelli/polars-api-compat-dev/t.py", line 9, in <module>
    print(df_nw[df_nw['c'].to_native()])
          ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/polars-api-compat-dev/narwhals/dataframe.py", line 915, in __getitem__
    raise TypeError(msg)
TypeError: Expected str or slice, got: <class 'pandas.core.series.Series'>
```

In https://github.com/narwhals-dev/narwhals/pull/2393, I'm preserving the current behaviour. I just wanted to open this as a potential follow-up to make sure it doesn't get lost - for consistency, we should probably allow native objects for slicing in `DataFrame.__getitem__`, insofar as we allow `Sequence[int]`

Keeping this out of #2393 though as that's already quite large

MarcoGorelli:
closes #2405

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


dangotbanned:
Yes please

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +373,41 @@ def _numpy_column_names(
         data: _2DArray, columns: Sequence[str] | None, /
     ) -> list[str]:
         return list(columns or (f"column_{x}" for x in range(data.shape[1])))
+
+    def gather(self, indices: Any) -> Self: ...
```
dangotbanned, line 377: Very early nit (sorry)

Would `gather` make more sense as `_gather`?

I'm just thinking about confusing it with ...
https://github.com/narwhals-dev/narwhals/blob/a56c02d8644be488d7b379dfa077869c5abc4e75/narwhals/_compliant/dataframe.py#L137

dangotbanned: There's also [`pl.DataFrame.slice`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.slice.html) - not sure if that is close enough 🤔

narwhals/dataframe.py
```diff
@@ -797,6 +798,9 @@ def estimated_size(self: Self, unit: SizeUnit = "b") -> int | float:
         """
         return self._compliant_frame.estimated_size(unit=unit)
 
+    @overload
+    def __getitem__(self: Self, item: tuple[int, int | str]) -> Any: ...
```
dangotbanned, line 802: Maybe now would be a time to revisit this PR I abandoned?
- #2020

I think we could get more use out of the individual aliases in this model you've used - where they'd have their own method

https://github.com/narwhals-dev/narwhals/blob/2858cc9813479bba97696e82a270a842251e65a3/narwhals/_pandas_like/dataframe.py#L189-L201

---

After seeing (#2389), I did some more digging and these were an interesting read:

- https://github.com/pola-rs/polars/blob/ece6e633e4ea2be42fde5346181c32b1bcffa7bc/py-polars/polars/_utils/getitem.py
- https://github.com/pola-rs/polars/blob/ece6e633e4ea2be42fde5346181c32b1bcffa7bc/py-polars/polars/_utils/slice.py

MarcoGorelli: sure, thanks - happy for you to take over the PR if it interests you

dangotbanned: > sure, thanks - happy for you to take over the PR if it interests you

All yours @MarcoGorelli, I'm in no rush to understand these slicing complexities 😂

narwhals/_polars/namespace.py
```diff
@@ -255,7 +255,7 @@ def concat_str(
     @property
     def selectors(self) -> CompliantSelectorNamespace[PolarsDataFrame, PolarsSeries]:
         return cast(
-            "CompliantSelectorNamespace[PolarsDataFrame, PolarsSeries]",
+            "CompliantSelectorNamespace[PolarsDataFrame, PolarsSeries]",  # pyright: ignore[reportInvalidTypeArguments]
```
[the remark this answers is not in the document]
dangotbanned: That usually pops up if you've added something a protocol, but not implemented it for polars.

The errors aren't that helpful when the protocol doesn't use inheritance sadly 😔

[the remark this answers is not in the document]
dangotbanned: The rule to follow is - if you don't need something for `polars`, move it into the `Eager*` sub-protocol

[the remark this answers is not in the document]
dangotbanned: If there isn't anything new, try checking the annotations are still valid

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +380,45 @@ def _numpy_column_names(
...
+
+        if not is_null_slice(rows):
+            is_native_series = self.__narwhals_namespace__()._series._is_native
+            if isinstance(rows, int):
+                compliant = compliant._gather([rows])
+            elif isinstance(rows, (slice, range)):
+                compliant = compliant._gather_slice(rows)
+
+            elif is_sequence_like_ints(rows) or is_native_series(rows):
+                compliant = compliant._gather(rows)
```
dangotbanned, line 419: Feel like I didn't do a good enough job documenting this

- https://github.com/narwhals-dev/narwhals/pull/2315/files#diff-76bea174dd4497eac2915b830209c9b96f3b9004d6be41b9d2405fcf8e2e45a2

tests/frame/getitem_test.py
```diff
@@ -242,3 +246,65 @@ def test_get_item_works_with_tuple_and_list_indexing_and_str(
 ) -> None:
     nw_df = nw.from_native(constructor_eager(data), eager_only=True)
     nw_df[row_idx, col]
+
+
+def test_getitem_ndarray_columns(constructor_eager: ConstructorEager) -> None:
+    data = {"col1": ["a", "b", "c", "d"], "col2": np.arange(4), "col3": [4, 3, 2, 1]}
+    nw_df = nw.from_native(constructor_eager(data), eager_only=True)
+    arr: np.ndarray[tuple[int], np.dtype[np.int64]] = np.array([0, 1])  # pyright: ignore[reportAssignmentType]
```
dangotbanned, line 254:
```suggestion
    arr = np.arange(2)
```

narwhals/utils.py
```diff
@@ -1295,6 +1297,54 @@ def is_sequence_but_not_str(sequence: Any | Sequence[_T]) -> TypeIs[Sequence[_T]
...
+        or isinstance(sequence, Series)
+    )
+
+
+def is_slice_ints(obj: object) -> bool:
+    return isinstance(obj, slice) and (
+        isinstance(obj.start, int)  # e.g. [1:]
+        or isinstance(obj.stop, int)  # e.g. [:3]
+        or (obj.start is None and obj.stop is None)  # e.g. [::2]
+    )
```
dangotbanned, line 1330: @MarcoGorelli I'm trying to work out the typing for this.

The examples you give would be:

```py
def is_slice_ints(
    obj: object,
) -> TypeIs[slice[int, None, None] | slice[None, int, None] | slice[None, None, int]]: ...
```

But I can't tell if you want this instead:

```py
def is_slice_ints(
    obj: object,
) -> TypeIs[slice[int, Any, Any] | slice[Any, int, Any] | slice[Any, Any, int]]: ...
```

Since the first two cases imply anything is allowed for `step`

narwhals/dataframe.py
dangotbanned: > sure, thanks - happy for you to take over the PR if it interests you

I've been reeled back in with the typing 😅

narwhals/utils.py
```diff
@@ -1295,6 +1299,54 @@ def is_sequence_but_not_str(sequence: Any | Sequence[_T]) -> TypeIs[Sequence[_T]
...
+def is_sequence_like_ints(sequence: Any | Sequence[_T]) -> bool:
+    np = get_numpy()
+    return (
+        (
+            isinstance(sequence, Sequence)
+            and (
+                (len(sequence) > 0 and isinstance(sequence[0], int))
+                or (len(sequence) == 0)
+            )
+        )
```
dangotbanned, line 1311: This has a false positive for:

```py
>>> is_sequence_like_ints("")
True
```

narwhals/utils.py
MarcoGorelli: i think the latter

narwhals/dataframe.py
```diff
@@ -826,12 +833,15 @@ def __getitem__(
...
+            | tuple[int, int | str]
+            | tuple[int | slice | Sequence[int] | _1DArray | Series[Any], int | str]
             | tuple[
-                slice | Sequence[int] | _1DArray, slice | Sequence[int] | Sequence[str]
+                int | slice | Sequence[int] | _1DArray | Series[Any],
+                slice | Sequence[int] | Sequence[str] | _1DArray | Series[Any],
             ]
         ),
-    ) -> Series[Any] | Self:
+    ) -> Series[Any] | Self | Any:
```
dangotbanned, line 844: 🤯 @MarcoGorelli 

Is it alright if we get these worked out first? https://github.com/narwhals-dev/narwhals/pull/2393#discussion_r2050476211

I'm having a hard time following these, especially since everything else seems to have dropped typing 🤔?

https://github.com/MarcoGorelli/narwhals/blob/b503c2f2bb41b4490aaa310cdb764698789350d5/narwhals/_compliant/dataframe.py#L384-L394

https://github.com/MarcoGorelli/narwhals/blob/b503c2f2bb41b4490aaa310cdb764698789350d5/narwhals/dataframe.py#L901-L904

narwhals/_polars/namespace.py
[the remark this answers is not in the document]
MarcoGorelli: nice, thanks for resolving! i should probably have spent a little longer on it 😄

narwhals/dataframe.py
MarcoGorelli: yup, definitely!

dangotbanned: > yup, definitely

Great - adding to my list 😎

narwhals/typing.py
```diff
@@ -309,6 +310,22 @@ def __native_namespace__(self) -> ModuleType: ...
...
+
+# Annotations for `__getitem__` methods
+_Slice: TypeAlias = "slice[Any, Any, Any]"
+SingleIndexSelector: TypeAlias = int
+SingleNameSelector: TypeAlias = str
+MultiIndexSelector: TypeAlias = "_Slice | Sequence[int] | Series[Any] | _1DArray"
+MultiNameSelector: TypeAlias = "_Slice | Sequence[str] | Series[Any] | _1DArray"
+BooleanMask: TypeAlias = "Sequence[bool] | Series[Any] | _1DArray"
+SingleColSelector: TypeAlias = "SingleIndexSelector | SingleNameSelector"
+MultiColSelector: TypeAlias = "MultiIndexSelector | MultiNameSelector | BooleanMask"
```
dangotbanned, line 327: Copied `BooleanMask` from `polars`, but I don't think we need it.

1. We didn't explictly state it was supported before
2. `bool` being a subclass of `int` makes it redundant to add here https://github.com/python/typeshed/blob/1b267b25f2c726291e4e6627b7567f2c8dc04b60/stdlib/builtins.pyi#L906-L907

dangotbanned: I removed the explicit reference to `range` for the same reason (`Sequence[int]`)
https://github.com/python/typeshed/blob/1b267b25f2c726291e4e6627b7567f2c8dc04b60/stdlib/builtins.pyi#L1241-L1241

narwhals/dataframe.py
```diff
@@ -881,40 +884,52 @@ def __getitem__(
...
-
-        else:
-            msg = f"Expected str or slice, got: {type(item)}"
-            raise TypeError(msg)
+        if rows is None and columns is None:
+            return self
+
+        compliant = self._compliant_frame
+        rows = to_native(rows, pass_through=True)
+        columns = to_native(columns, pass_through=True)
```
dangotbanned, line 920: In relation to (https://github.com/narwhals-dev/narwhals/pull/2393/commits/f05c3a2440f8bdec74c139e34b88f5474b100f66), this is the last major questionable part

It converts all of the narrowing into an `Any`, but I think thats hiding too much of what's going on here.

What are the *non*-noop cases for `rows` and `columns`?

narwhals/typing.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2393/commits/f4aed5fa34f29fe17cc3d1c8c6dd3d3473eda121)

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2393#event-17328071292)

@MarcoGorelli you thought you could sneak this support in but I see you 😉

narwhals/typing.py
```diff
@@ -309,6 +310,21 @@ def __native_namespace__(self) -> ModuleType: ...
...
+
+
+# Annotations for `__getitem__` methods
+_Slice: TypeAlias = "slice[Any, Any, Any]"
+SingleIndexSelector: TypeAlias = int
+SingleNameSelector: TypeAlias = str
+MultiIndexSelector: TypeAlias = "_Slice | Sequence[int] | Series[Any] | _1DArray"
+MultiNameSelector: TypeAlias = "_Slice | Sequence[str] | Series[Any] | _1DArray"
+SingleColSelector: TypeAlias = "SingleIndexSelector | SingleNameSelector"
+MultiColSelector: TypeAlias = "MultiIndexSelector | MultiNameSelector"
```
dangotbanned, line 326: I'm thinking the descriptions in `__getitem__` would be a better fit as tiny docs for each of these aliases.
They're a bit of a difficult read in the [API Reference](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.__getitem__) and in VSCode.

Would also give us something helpful on hover if we forgot what all of this means 😅 

## Docstring (source code)

https://github.com/narwhals-dev/narwhals/blob/0eff60b9d516f36d8a5f3a863115ecbfdb65eb3d/narwhals/dataframe.py#L835-L860



## API Reference

image

## Docstring (VSCode)

image

narwhals/utils.py
dangotbanned: Related tweaks in (https://github.com/narwhals-dev/narwhals/pull/2393/commits/46be441789c6131f33625e463db610f950e7e413)

- Checking `step` is an `int`
- Renaming to align with `polars` aliases
  - `is_slice_ints` -> `is_slice_index`
  - `is_null_slice` -> `is_slice_none`
    - Really like how `slice(None)` becomes the alias `_SliceNone` 😄

narwhals/typing.py
```diff
@@ -309,6 +310,27 @@ def __native_namespace__(self) -> ModuleType: ...
...
+# Overloaded sequence of strings
+_StrIndexer: TypeAlias = Any  # noqa: PYI047
+
+
+# Annotations for `__getitem__` methods
+_Slice: TypeAlias = "slice[Any, Any, Any]"
+_SliceNone: TypeAlias = "slice[None, None, None]"
+_SliceIndex: TypeAlias = (
+    "slice[int, Any, Any] | slice[Any, int, Any] | slice[None, None, int] | _SliceNone"
+)
```
dangotbanned, line 324: For now at least, `_SliceNone` needs to be included in the `_SliceIndex` alias.

Quite a lot of tests flared up with issues when I went as strict as requiring at least one `int`.
I think all the cases with `slice(None)` return a `DataFrame`, so I didn't need to add any overloads this way

narwhals/dataframe.py
MarcoGorelli: > What are the non-noop cases for rows and columns?

non sure what you're asking sorry, could you clarify please?

dangotbanned: > > What are the non-noop cases for rows and columns?
> 
> non sure what you're asking sorry, could you clarify please?

Sure @MarcoGorelli 

Okay so you're using this:

```py
to_native(rows, pass_through=True)
```

Because `pass_through` is there, `rows` may either:
- Return unchanged (noop case)
- Be converted into a native object (non-noop case)

But regardless of what happens, `rows` is reassigned as `Any` and erases all the [narrowing](https://typing.python.org/en/latest/spec/narrowing.html) we'd done so far.

That makes everything afterwards more difficult to follow

image

## Question
Are we doing anything more than this?

```py
if isinstance(rows, Series):
    rows_native = rows.to_native()
if isinstance(columns, Series):
    columns_native = columns.to_native()
```

MarcoGorelli: sure,
```python
if isinstance(rows, Series):
    rows_native = rows.to_native()
if isinstance(columns, Series):
    columns_native = columns.to_native()
```
should be perfectly fine 👍

dangotbanned: Actually now I've typed that all out, wouldn't it make more sense to pass `CompliantSeries` down, instead of `NativeSeries`?

I haven't moved onto the compliant impls yet - but I imagine they need to do another conversion?

MarcoGorelli: sure, that might be simpler / more consistent

dangotbanned:
@MarcoGorelli just fixing conflicts from #2399 now 👍

narwhals/dataframe.py
MarcoGorelli: done in [d24f1b7](https://github.com/narwhals-dev/narwhals/pull/2393/commits/d24f1b71fbbd6a968685f24186e94e309da075e1), it is indeed better, thanks 👍

dangotbanned: Nice one, thanks @MarcoGorelli!

As you might have been able to tell, I've been a bit distracted by (#2325) - so thanks again picking this back up 🚀

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +381,49 @@ def _numpy_column_names(
...
         return list(columns or (f"column_{x}" for x in range(data.shape[1])))
+
+    def _gather(self, indices: _IntIndexer) -> Self: ...
+    def _gather_slice(self, indices: slice | range) -> Self: ...
+    def _select_indices(self, indices: _IntIndexer) -> Self: ...
+    def _select_labels(self, indices: _StrIndexer) -> Self: ...
+    def _select_slice_of_indices(self, indices: slice | range) -> Self: ...
+    def _select_slice_of_labels(self, indices: slice | range) -> Self: ...
+
+    def __getitem__(self, item: tuple[Any, Any]) -> Self:
```
dangotbanned, line 392: At some point we should probably work out which aliases are going to be used for each of these.

*Maybe* also align the method names to the aliases?
https://github.com/MarcoGorelli/narwhals/blob/d24f1b71fbbd6a968685f24186e94e309da075e1/narwhals/typing.py#L327-L332

But I'm happy with just the annotations, if not 🙂

MarcoGorelli:
ooh, interesting CI failure from scikit-lego:
```python
__________________________ test_timegapsplit_summary ___________________________
[gw2] linux -- Python 3.11.12 /opt/hostedtoolcache/Python/3.11.12/x64/bin/python
    def test_timegapsplit_summary():
        cv = TimeGapSplit(
            date_serie=df["date"],
            train_duration=timedelta(days=5),
            valid_duration=timedelta(days=3),
            gap_duration=timedelta(days=0),
        )
    
>       summary = cv.summary(X_train)
tests/test_model_selection/test_timegapsplit.py:237: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
sklego/model_selection.py:285: in summary
    update_split_info(train_info, j, "train", summary)
sklego/model_selection.py:268: in update_split_info
    dates = X_index_df["__date__"][indices]
../narwhals/series.py:178: in __getitem__
    return self._with_compliant(self._compliant_series[idx])
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <narwhals._pandas_like.series.PandasLikeSeries object at 0x7f5bf039c5d0>
item = 0    0
1    1
2    2
3    3
4    4
Name: tmp, dtype: int64
    def __getitem__(self, item: Any) -> Self:
        if is_slice_none(item):
            return self
        if isinstance(item, int):
            return self._gather([item])
        elif isinstance(item, (slice, range)):
            return self._gather_slice(item)
        elif is_compliant_series(item):
            return self._gather(item.native)
        elif is_sequence_like_ints(item):
            return self._gather(item)
        else:
            msg = f"Unreachable code, got unexpected type: {type(item)}"
>           raise AssertionError(msg)
E           AssertionError: Unreachable code, got unexpected type: <class 'pandas.core.series.Series'>
../narwhals/_compliant/series.py:339: AssertionError
```

Looks like indexing there was done using a (native) pandas Series, as opposed to it being wrapped in a Narwhals Series

MarcoGorelli:
We may need to allow array-like inputs (which implement `__array__`) and call `np.asarray` on them if necessary

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2393#issuecomment-2817131493

Apologies if I'm reading this wrong, @MarcoGorelli just to clarify, are you suggesting that we need to support undocumented behavior if we discover it downstream?

Maybe I'm being naive, but I would've viewed that test failure as being the discovery of a bug in the current implementation - since we hadn't intended to support that use case 🤔

dangotbanned:
@MarcoGorelli thanks for the ping!

FYI, I'm expecting this to be a big one, so **before** the actual review I'm just adding some comments 👍

narwhals/typing.py
```diff
@@ -311,19 +311,19 @@ def __native_namespace__(self) -> ModuleType: ...
...
+_T = TypeVar("_T")
+_Slice: TypeAlias = "slice[_T, Any, Any] | slice[Any, _T, Any] | slice[None, None, _T]"
 _SliceNone: TypeAlias = "slice[None, None, None]"
-_SliceIndex: TypeAlias = (
-    "slice[int, Any, Any] | slice[Any, int, Any] | slice[None, None, int] | _SliceNone"
-)
+_SliceIndex: TypeAlias = "_Slice[int] | _SliceNone"
 """E.g. `[1:]` or `[:3]` or `[::2]`."""
 
+_SliceName: TypeAlias = "_Slice[str] | _SliceNone"
```
dangotbanned, line 320: I'm keeping the `_SliceNone` outside of the alias *for now* - hopefully should mean we can avoid overlaps in the compliant methods *without* it

narwhals/utils.py
```diff
@@ -1818,3 +1874,20 @@ def wrapper(instance: _ContextT, *args: P.args, **kwds: P.kwargs) -> R:
...
+
+
+def convert_str_slice_to_int_slice(
+    str_slice: slice | range, columns: list[str]
+) -> tuple[int | None, int | None, int | None]:
+    start = (
+        columns.index(cast("str", str_slice.start))
+        if str_slice.start is not None
+        else None
+    )
```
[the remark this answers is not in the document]
dangotbanned:
```py
from narwhals.utils import convert_str_slice_to_int_slice

columns = ["0", "1", "2", "3", "4", "5"]

>>> convert_str_slice_to_int_slice(slice("0", "2"), columns)
(0, 2, None)

>>> convert_str_slice_to_int_slice(range("0", "2"), columns)
TypeError: 'str' object cannot be interpreted as an integer
```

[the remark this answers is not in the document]
dangotbanned: Removing the `cast` reveals this 

image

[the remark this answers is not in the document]
MarcoGorelli: i don't think `range('0', '2')` is valid python 
```python
>>> range("0", "2")
TypeError: 'str' object cannot be interpreted as an integer
```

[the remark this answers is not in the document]
MarcoGorelli: ah i see, we can just remove the range here? sure

dangotbanned:
In addition to the [8 errors from `mypy`](https://github.com/narwhals-dev/narwhals/actions/runs/14561453323/job/40845271948?pr=2393) in (https://github.com/narwhals-dev/narwhals/pull/2393/commits/de2d9ba9bf21a3ca73d99e93d7db0e37e67aa335) `pyright` has 23 for me locally - after removing all the `cast`s 🤔 

I'm just having a lil break now before diving into what all the issues are.

I think (https://github.com/narwhals-dev/narwhals/pull/2393#discussion_r2051747123) probably accounts for a few, but another one is this kind of thing


https://github.com/MarcoGorelli/narwhals/blob/de2d9ba9bf21a3ca73d99e93d7db0e37e67aa335/narwhals/_compliant/dataframe.py#L410-L421

Here no narrowing took place, where I would assumed we'd reached the end of the `int` cases

image

And then at the end, we would want to see `Never` - but instead we've got most of the types that we started with

image

narwhals/utils.py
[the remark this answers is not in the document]
dangotbanned: > ah i see, we can just remove the range here? sure

Yes lol

narwhals/dataframe.py
```diff
@@ -879,40 +882,54 @@ def __getitem__(
...
-            msg = f"Expected str or slice, got: {type(item)}"
-            raise TypeError(msg)
+        compliant = self._compliant_frame
+
+        if isinstance(rows, int) and isinstance(columns, (int, str)):
+            return self.item(rows, columns)
+        if isinstance(columns, (int, str)):
+            col_name = columns if isinstance(columns, str) else self.columns[columns]
+            series = self.get_column(col_name)
+            return series[rows] if rows is not None else series
```
dangotbanned, line 922: Here, `rows` is apprently:

```py
MultiIndexSelector | SingleColSelector | slice[str, Any, Any] | slice[Any, str, Any] | None
```

But that doesn't match a `Series` overload 😔

narwhals/series.py
```diff
@@ -169,10 +172,25 @@ def __getitem__(self, idx: int | slice | Sequence[int] | Self) -> Any | Self:
...
-            self._compliant_series[to_native(idx, pass_through=True)]
-        )
+            idx = int(idx) if not isinstance(idx, int) else idx
+            return self._compliant_series.item(idx)
+
+        if isinstance(idx, self.to_native().__class__):
+            idx = self._with_compliant(self._compliant_series._with_native(idx))
+
+        # For Series.__getitem__, we only
+        if not is_index_selector(idx):
```
dangotbanned, line 182: @MarcoGorelli I don't suppose you remember the other half of this sentence? 😄

narwhals/dataframe.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2393/commits/7edc0ca5b50e28c8e02841a8c9ccb5523db0fb7c)

This is what we like to see 😎 

image

narwhals/series.py
MarcoGorelli: I think it was probably something like "we only allow for integers rather than labels", but we can probably delete the comment

MarcoGorelli: thanks a tonne for everything you've done here, seriously impressive + useful + amazing! 🙏 🙌

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +395,54 @@ def _numpy_column_names(
...
+    def _select_slice_of_indices(self, indices: _SliceIndex | range) -> Self: ...
+    def _select_slice_of_labels(self, item: _SliceName) -> Self: ...
+
+    def __getitem__(
+        self,
+        item: tuple[
+            SingleIndexSelector | MultiIndexSelector | CompliantSeriesAny,
+            MultiIndexSelector | MultiColSelector | CompliantSeriesAny,
+        ],
+    ) -> Self:
```
dangotbanned, line 412: ## TODO
We need to do something different for the `Compliant`-level annotations.

They contain `nw.Series[Any]` in each of:
- `SizedMultiIndexSelector`
- `MultiIndexSelector`
- `SizedMultiNameSelector`
- `MultiNameSelector`
- `MultiColSelector`

One option could be making some of these aliases generic in the *Series* position, like:

```py
SizedMultiNameSelector: TypeAlias = Sequence[str] | T | _1DArray
MultiNameSelector: TypeAlias = _SliceName | SizedMultiNameSelector[T]
```

So then we could reuse the same alias at different levels:

```py

# Currently used everywhere, we only need it at narwhals-level
MultiNameSelector[Series[Any]]

# We'd use that here, instead of adding `CompliantSeriesAny` in retroactively
MultiNameSelector[CompliantSeriesAny]
```

narwhals/series.py
dangotbanned: > thanks a tonne for everything you've done here, seriously impressive + useful + amazing! 🙏 🙌

Ah no problem, happy to help 😍

narwhals/_arrow/dataframe.py
```diff
@@ -248,118 +248,46 @@ def get_column(self, name: str) -> ArrowSeries:
...
-                return self._with_native(self.native.select(cast("Indices", item)))
-            if isinstance(item, Sequence) and len(item) == 0:
-                return self._with_native(self.native.slice(0, 0))
-            return self._with_native(self.native.take(cast("Indices", item)))
+    def _gather(self, item: SizedMultiIndexSelector) -> Self:
+        if len(item) == 0:
+            return self._with_native(self.native.slice(0, 0))
+        if self._backend_version < (18,) and isinstance(item, tuple):
+            item = list(item)
+        return self._with_native(self.native.take(item))
```
dangotbanned, line 256: @MarcoGorelli if you can put together a minimal repro for each of the `pyarrow` errors, I can fix upstream if they're false negatives? 

5/19 we have left from `pyright` are in `_arrow/(dataframe|series)`

narwhals/_compliant/dataframe.py
dangotbanned: These are also the same issue, where we could specialize the alias(s) with `pa.ChunkedArray` to fix it

image

narwhals/_polars/dataframe.py
```diff
@@ -257,54 +265,70 @@ def collect_schema(self) -> dict[str, DType]:
...
 
-    def __getitem__(self, item: Any) -> Any:
+    def __getitem__(
+        self,
+        item: tuple[
+            SingleIndexSelector | MultiIndexSelector | PolarsSeries,
+            MultiIndexSelector | MultiColSelector | PolarsSeries,
+        ],
+    ) -> Any:
+        rows, columns = item
```
dangotbanned, line 275: So this is the main source of grief left.

I understand it was typed as `Any` before, but I think having this typed **accurately** will really help us stay in sync with `polars` in the long-term.

If there's a change upstream, it'll be soooooo much easier to spot if we get a heads-up from `pyright` or `mypy`

MarcoGorelli: yup, agree - ok to leave this as a follow-up?

dangotbanned: > yup, agree - ok to leave this as a follow-up?

Thanks @MarcoGorelli yeah just open an issue for it - so we don't forget  🙏

dangotbanned:
> @MarcoGorelli thanks for the ping!
> 
> FYI, I'm expecting this to be a big one, so **before** the actual review I'm just adding some comments 👍

Still need to do the review pass 😅

Compared to (https://github.com/narwhals-dev/narwhals/pull/2393#issuecomment-2817221905) we're much closer now!

narwhals/_compliant/dataframe.py
dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/2393/commits/693c4eb009ae8f2de12db10e2e5623359227dfc9) is going to introduce loads of new errors 🙈 

Sorry @MarcoGorelli! I didn't think you were active (https://github.com/narwhals-dev/narwhals/pull/2393/commits/dc71fd90e5a040970e24f68a90b790cad0cc05db)

dangotbanned: It's a partial impl of this task, just need to do the backend parts (https://github.com/narwhals-dev/narwhals/actions/runs/14571479901/job/40869426241?pr=2393)

That'll eventually fix a load of other stuff - but the rest step is breaking things 😄

tests/hypothesis/getitem_test.py
```diff
@@ -155,79 +155,23 @@ def test_getitem(
...
-    # df[..., ::step] is not fine:
-    # TypeError: Expected slice of integers or strings, got: <class 'slice'>
-    assume(
-        not (
-            isinstance(selector, tuple)
-            and isinstance(selector[1], slice)
-            and selector[1].start is None
-            and selector[1].stop is None
-        )
-    )
```
MarcoGorelli, line 224: @sjdenny check this out, we can finally remove these 🥳 

Thank you _so_ much for having set up this hypothesis test, it was _so_ useful to finding edge cases 🙌 legend

narwhals/_compliant/dataframe.py
dangotbanned: Getting closer with (https://github.com/narwhals-dev/narwhals/pull/2393/commits/2a5e8668660c50129f1c3f94f12cc73ea4eed122)

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +395,54 @@ def _numpy_column_names(
         data: _2DArray, columns: Sequence[str] | None, /
     ) -> list[str]:
         return list(columns or (f"column_{x}" for x in range(data.shape[1])))
+
+    def _gather(self, indices: SizedMultiIndexSelector[NativeSeries]) -> Self: ...
+    def _gather_slice(self, indices: _SliceIndex | range) -> Self: ...
+    def _select_indices(self, indices: SizedMultiIndexSelector[NativeSeries]) -> Self: ...
+    def _select_labels(self, indices: SizedMultiNameSelector[NativeSeries]) -> Self: ...
```
dangotbanned, line 402: I'm gonna need to add `NativeSeriesT` to `EagerDataFrame` to resolve these Liskov violations https://github.com/narwhals-dev/narwhals/actions/runs/14572060963/job/40871004012?pr=2393

Buuuuut, it'll *also* plug the final hole we have in [`EagerDataFrame.__narwhals_namespace__`](https://github.com/MarcoGorelli/narwhals/blob/af8f152e246aab4081c32e3a79fe99030c7adbd5/narwhals/_compliant/dataframe.py#L358-L360)

```py
EagerNamespace[Self, EagerSeriesT, EagerExprT, NativeFrameT, Any]
```

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2393/commits/308e197c8ea2ba8afc1ba2e4097dfe1c74b44bb5)

narwhals/_polars/dataframe.py
```diff
@@ -257,54 +267,70 @@ def collect_schema(self) -> dict[str, DType]:
...
+                    native = native[rows, :]  # pyright: ignore[reportArgumentType,reportCallIssue]
+                elif is_compliant_series(rows):
+                    native = native[rows.native, :]  # pyright: ignore[reportArgumentType,reportCallIssue]
+                elif is_sequence_like(rows):
+                    native = native[rows, :]  # pyright: ignore[reportArgumentType,reportCallIssue]
+                else:
+                    msg = f"Unreachable code, got unexpected type: {type(rows)}"
+                    raise AssertionError(msg)
+
+            return self._with_native(native)  # pyright: ignore[reportArgumentType]
```
dangotbanned, line 333: The main issue is all the assignments to `native`.

Gradually, it widens from `pl.DataFrame` to `pl.DataFrame | pl.Series | Any`

narwhals/_pandas_like/dataframe.py
```diff
@@ -281,135 +283,40 @@ def get_column(self, name: str) -> PandasLikeSeries:
...
-            elif isinstance(item[1], int):
-                native_series = self.native.iloc[item]
-            else:  # pragma: no cover
-                msg = f"Expected str or int, got: {type(item[1])}"
-                raise TypeError(msg)
+    def _gather_slice(self, item: _SliceIndex | range) -> Self:
+        return self._with_native(
+            self.native.iloc[slice(item.start, item.stop, item.step), :],
+            validate_column_names=False,
+        )
```
dangotbanned, line 294: The difference between this and `_select_slice_index` is incredibly subtle.

For `ArrowDataFrame` there *seems* to be a bigger difference, but I'm struggling to understand both 😵

Where is the line drawn between `_gather*` and `_select*`?

narwhals/_compliant/dataframe.py
```diff
@@ -396,15 +396,14 @@ def _numpy_column_names(
...
+    def _select_multi_index(
+        self, item: SizedMultiIndexSelector[NativeSeriesT]
     ) -> Self: ...
-    def _select_labels(self, indices: SizedMultiNameSelector[NativeSeriesT]) -> Self: ...
-    def _select_slice_of_indices(self, indices: _SliceIndex | range) -> Self: ...
-    def _select_slice_of_labels(self, item: _SliceName) -> Self: ...
-
+    def _select_multi_name(self, item: SizedMultiNameSelector[NativeSeriesT]) -> Self: ...
+    def _select_slice_index(self, item: _SliceIndex | range) -> Self: ...
+    def _select_slice_name(self, item: _SliceName) -> Self: ...
```
dangotbanned, line 406: Btw I'm not *necessarily* saying these are the names we should use.

Just that whatever names we **do** settle on have a clear trace back to the aliases (https://github.com/narwhals-dev/narwhals/pull/2393#discussion_r2051562650)

narwhals/_pandas_like/dataframe.py
MarcoGorelli: gather is for rows, select is for columns

dangotbanned:
> > @MarcoGorelli thanks for the ping!
> > FYI, I'm expecting this to be a big one, so **before** the actual review I'm just adding some comments 👍
> 
> Still need to do the review pass 😅
> 
> Compared to ([#2393 (comment)](https://github.com/narwhals-dev/narwhals/pull/2393#issuecomment-2817221905)) we're much closer now!

## Update
Reviewing now 🎉

narwhals/_compliant/dataframe.py
MarcoGorelli: sure, no objections

narwhals/_compliant/series.py
```diff
@@ -316,6 +321,23 @@ def __narwhals_namespace__(
     def _to_expr(self) -> EagerExpr[Any, Any]:
         return self.__narwhals_namespace__()._expr._from_series(self)  # type: ignore[no-any-return]
 
+    def _gather(self, item: SizedMultiIndexSelector[NativeSeriesT]) -> Self: ...
+    def _gather_slice(self, item: _SliceIndex | range) -> Self: ...
+    def __getitem__(self, item: Any) -> Self:
```
dangotbanned, line 326: Really think we need an annotation here.

3/5 branches below call `_gather` - but I don't know what we started with 😞

Ideally, we could capture those in one branch and then refine from there - to prepare the argument we pass.

Then `slice | range` is dealt with separately

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +395,55 @@ def _numpy_column_names(
...
+                elif is_compliant_series(columns):
+                    compliant = self._select_multi_index(columns.native)
+                else:
+                    compliant = compliant._select_multi_index(columns)
+            elif isinstance(columns, slice):
+                compliant = compliant._select_slice_name(columns)
+            elif is_compliant_series(columns):
+                compliant = self._select_multi_name(columns.native)
+            elif is_sequence_like(columns):
+                compliant = self._select_multi_name(columns)
```
dangotbanned, line 431: Similar thought to (https://github.com/narwhals-dev/narwhals/pull/2393#discussion_r2052548358)

- `_select_multi_index` has 2 calls (one is a bit messy due to the `compliant = self`)
- `_select_multi_name` has the same thing

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +395,55 @@ def _numpy_column_names(
...
+        if not is_slice_none(rows):
+            if isinstance(rows, int):
+                compliant = compliant._gather([rows])
+            elif isinstance(rows, (slice, range)):
+                compliant = compliant._gather_slice(rows)
+            elif is_compliant_series(rows):
+                compliant = compliant._gather(rows.native)
+            elif is_sized_multi_index_selector(rows):
+                compliant = compliant._gather(rows)
+            else:
```
dangotbanned, line 445: Is this the same as?

https://github.com/MarcoGorelli/narwhals/blob/2c0100778978fbc2f562ba25810c5184809e9920/narwhals/_compliant/series.py#L329-L336

narwhals/utils.py
```diff
@@ -1256,10 +1266,64 @@ def parse_columns_to_drop(
...
+        (
+            is_sequence_but_not_str(obj)
+            and ((len(obj) > 0 and isinstance(obj[0], int)) or (len(obj) == 0))
+        )
+        or (is_numpy_array_1d(obj) and np.issubdtype(obj.dtype, np.integer))
+        or (
+            (is_narwhals_series(obj) or is_compliant_series(obj))
+            and obj.dtype.is_integer()
+        )
+    )
```
dangotbanned, line 1289: I think this should be split out into multiple guards

E.g.

- `is_numpy_array_1d_int`
- `is_narwhals_series_int`
- `is_compliant_series_int`

That way, we can be more specfic about what is allowed at each level **and** not need the `Any` in `SizedMultiIndexSelector`

dangotbanned (commented):
Thanks for the PR @MarcoGorelli!

Some really big steps towards *demystifying* for me 😄 and looks very close to merging (*despite the review length*)

tests/frame/getitem_test.py
```diff
@@ -242,3 +255,120 @@ def test_get_item_works_with_tuple_and_list_indexing_and_str(
...
+    arr = np.arange(2)
+    result = nw_df[:, arr]
+    expected = {"col1": ["a", "b", "c", "d"], "col2": [0, 1, 2, 3]}
+    assert_equal_data(result, expected)
+
+
+def test_getitem_ndarray_columns_labels(constructor_eager: ConstructorEager) -> None:
+    data = {"col1": ["a", "b", "c", "d"], "col2": np.arange(4), "col3": [4, 3, 2, 1]}
+    nw_df = nw.from_native(constructor_eager(data), eager_only=True)
+    arr: np.ndarray[tuple[int], np.dtype[Any]] = np.array(["col1", "col2"])  # pyright: ignore[reportAssignmentType]
```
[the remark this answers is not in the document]
MarcoGorelli: tbh i think it's clearer as-is

tests/frame/getitem_test.py
```diff
@@ -242,3 +255,120 @@ def test_get_item_works_with_tuple_and_list_indexing_and_str(
...
+    data = {"a": [1, 2, 3]}
+    with pytest.raises(TypeError, match="Tuples cannot"):
+        nw.from_native(constructor_eager(data), eager_only=True)[(1, 2, 3)]
+
+
+def test_slice_with_series(
+    constructor_eager: ConstructorEager, request: pytest.FixtureRequest
+) -> None:
+    if "pandas_pyarrow" in str(constructor_eager):
+        request.applymarker(pytest.mark.xfail)
```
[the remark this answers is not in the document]
MarcoGorelli: bug on their side, i've opend an issue

narwhals/_arrow/dataframe.py
```diff
@@ -80,8 +81,9 @@
     PromoteOptions: TypeAlias = Literal["none", "default", "permissive"]
 
 
-class ArrowDataFrame(EagerDataFrame["ArrowSeries", "ArrowExpr", "pa.Table"]):
-    # --- not in the spec ---
+class ArrowDataFrame(
+    EagerDataFrame["ArrowSeries", "ArrowExpr", "pa.Table", "pa.ChunkedArray[Any]"]
```
lorentzenchr, line 85: What about pyarrow's `RecordBatch`?

narwhals/_compliant/dataframe.py
```diff
@@ -99,7 +115,13 @@ def from_numpy(
         schema: Mapping[str, DType] | Schema | Sequence[str] | None,
     ) -> Self: ...
     def __array__(self, dtype: Any, *, copy: bool | None) -> _2DArray: ...
-    def __getitem__(self, item: Any) -> CompliantSeriesT | Self: ...
+    def __getitem__(
+        self,
+        item: tuple[
+            SingleIndexSelector | MultiIndexSelector[CompliantSeriesT],
+            MultiIndexSelector[CompliantSeriesT] | MultiColSelector[CompliantSeriesT],
```
lorentzenchr, line 122:
```suggestion
            SingleColSelector | MultiColSelector[CompliantSeriesT],
```
Not sure. But it sounds more logical to me that way (note, I'm a 100% narwhals newbie).

narwhals/_compliant/dataframe.py
```diff
@@ -369,3 +395,55 @@ def _numpy_column_names(
...
+        self, item: SizedMultiIndexSelector[NativeSeriesT]
+    ) -> Self: ...
+    def _select_multi_name(self, item: SizedMultiNameSelector[NativeSeriesT]) -> Self: ...
+    def _select_slice_index(self, item: _SliceIndex | range) -> Self: ...
+    def _select_slice_name(self, item: _SliceName) -> Self: ...
+    def __getitem__(
+        self,
+        item: tuple[
+            SingleIndexSelector | MultiIndexSelector[EagerSeriesT],
+            MultiIndexSelector[EagerSeriesT] | MultiColSelector[EagerSeriesT],
```
lorentzenchr, line 411:
```suggestion
            SingleColSelector | MultiColSelector[EagerSeriesT],
```
Same here?

lorentzenchr (commented):
@MarcoGorelli I think this PR will be super useful for scikit-learn.
Do you also envision to allow for boolean masks, especially for row indexing? (Not a must, but - again - useful.)

tests/frame/getitem_test.py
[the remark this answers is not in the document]
dangotbanned: Thanks @MarcoGorelli 

- [ ] https://github.com/pandas-dev/pandas/issues/61311

narwhals/_arrow/dataframe.py
dangotbanned: > What about pyarrow's `RecordBatch`?

@lorentzenchr what is the question here?

lorentzenchr: Should it read
```suggestion
    EagerDataFrame["ArrowSeries", "ArrowExpr", "pa.Table", "pa.ChunkedArray[Any]", "pa.RecordBatch"]
```

dangotbanned: > Should it read

@lorentzenchr that introduces a runtime `TypeError`

```py
TypeError: Too many arguments for <class 'narwhals._compliant.dataframe.EagerDataFrame'>; actual 5, expected 4
```

What meaning were you trying to convey by adding `pa.RecordBatch` there?

MarcoGorelli:
Thanks for your feedback! 

> Do you also envision to allow for boolean masks, especially for row indexing?

Maybe, in the meantime you can use `filter` for that

narwhals/_compliant/dataframe.py
MarcoGorelli: Thanks for your suggestion!

I don't think it's right though, the singlecolselector case returns a Series and that's handled separately

tests/hypothesis/getitem_test.py
sjdenny: Amazing!  Glad it was useful :)

narwhals/dataframe.py
```diff
@@ -879,41 +886,58 @@ def __getitem__(
...
+        if isinstance(rows, Series):
+            rows = rows._compliant_series
+        if isinstance(columns, Series):
+            columns = columns._compliant_series
+        if rows is None:
+            return self._with_compliant(compliant[:, columns])
+        if columns is None:
+            return self._with_compliant(compliant[rows, :])
+        return self._with_compliant(compliant[rows, columns])
```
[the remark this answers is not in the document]
MarcoGorelli: if possible, sure

narwhals/_polars/namespace.py
```diff
@@ -251,7 +251,7 @@ def concat_str(
...
```
MarcoGorelli, line 254: trying to resolve this again

narwhals/_arrow/dataframe.py
```diff
@@ -248,118 +250,61 @@ def get_column(self, name: str) -> ArrowSeries:
...
-        elif isinstance(item, Sequence) or is_numpy_array_1d(item):
-            if isinstance(item, Sequence) and len(item) > 0 and isinstance(item[0], str):
-                return self._with_native(self.native.select(cast("Indices", item)))
-            if isinstance(item, Sequence) and len(item) == 0:
-                return self._with_native(self.native.slice(0, 0))
-            return self._with_native(self.native.take(cast("Indices", item)))
+    def _select_multi_index(
+        self, columns: SizedMultiIndexSelector[ArrowChunkedArray]
+    ) -> Self:
+        selector: Sequence[int] | Sequence[str]
```
MarcoGorelli, line 284: @dangotbanned this isn't right, it can't be `Sequence[str]` if it gets here

narwhals/_polars/namespace.py
dangotbanned: @MarcoGorelli 

Compare what changed in `CompliantDataFrame` to what changed in `PolarsDataFrame`.

Then do the same for `CompliantSeries` and `PolarsSeries`.

Is anything missing?
Are the signatures the same?
Are the annotations equivalent?

narwhals/_compliant/dataframe.py
MarcoGorelli: same response 😉  the `SingleColSelector` gets handled before we get here

narwhals/_polars/namespace.py
MarcoGorelli: 😄 nice, thanks for your patience here 🙏

narwhals/_arrow/dataframe.py
dangotbanned: I'm happy the annotation could be removed (for CI) - but I'm still getting this locally 🤔 

image

dangotbanned: @MarcoGorelli yeah all good - there was a fix 3 days ago 😄 

- https://github.com/numpy/numpy/pull/28650

MarcoGorelli: it's not for CI, it's for readibility / documentation too - `_select_multi_index` is for selecting columns by index. not by name. so it can't be `Sequence[str]`

narwhals/_arrow/dataframe.py
```diff
@@ -248,118 +250,61 @@ def get_column(self, name: str) -> ArrowSeries:
...
+        if isinstance(columns, pa.ChunkedArray):
+            # TODO @dangotbanned: Fix upstream with `pa.ChunkedArray.to_pylist(self) -> list[Any]:`
+            selector = cast("Sequence[int]", columns.to_pylist())
+        # TODO @dangotbanned: Fix upstream, it is actually much narrower
+        # **Doesn't accept `ndarray`**
+        elif is_numpy_array(columns):
+            selector = columns.tolist()
+        else:
+            selector = columns
+        return self._with_native(self.native.select(selector))
```
dangotbanned, line 294: @MarcoGorelli I'm absolutely baffled by `pyarrow` on this one 🤣 

- https://github.com/zen-xu/pyarrow-stubs/pull/221

`np.ndarray` is allowed **only** when it contains `np.str_` - because that passes a `isinstance(obj, str)` check.

The [current stubs](https://github.com/zen-xu/pyarrow-stubs/blob/f899bb35e10b36f7906a728e9f8acf3e0a1f9f64/pyarrow-stubs/_stubs_typing.pyi#L32) suggest **only** `np.integer` is allowed.

Also - no native `pyarrow` objects will work 🤦‍♂️

narwhals/utils.py
MarcoGorelli: sure, ok if we do it as follow-up? we're passed 100 messages here, it's getting hard to follow

narwhals/_compliant/dataframe.py
MarcoGorelli: same as https://github.com/narwhals-dev/narwhals/pull/2393#discussion_r2054846542

MarcoGorelli:
thanks all for comments 🙏 really appreciate it!

especially:
- @lorentzenchr for highlighting this area of improvement - we'd added `__getitem__` a while ago but then hadn't touched it in some time, really glad to have revisited it
- @sjdenny for the hypothesis tests, which helped avoid regressing on some edge cases
- @dangotbanned for your patient reviews and help with annotations, which make everything SO much better

PR #2395, opened 2025-04-18
fix: Use `np.generic` in `is_numpy_scalar`

### LINKED ISSUE #2394: [Bug]: `is_numpy_scalar` reeturns `True` for non-numpy scalars
### Describe the bug

The function signature of `narwhals.dependencies.is_numpy_scalar` says that it narrows the type to `np.generic`m but it is not necessarily the case, as `numpy.is_scalar` also returns `True` for Python scalars. A potential fix would be to just do an explicit `isinstance(..., np.generic)` check instead of using `is_scalar`.

Also, a related question: is `is_numpy_scalar` considered to be a part of the public API? I don't think it's in the docs, but I think it could be useful.

```python
import numpy as np
from narwhals.dependencies import is_numpy_scalar

isinstance(1, np.generic)  # False
is_numpy_scalar(1)  # True, but should be False
```

### Expected results

`is_numpy_scalar(x) == True` should imply `isinstance(x, np.generic) == True`.

### Actual results

It does not for Python scalars (see example).

```shell
System:
    python: 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:30) [Clang 18.1.8 ]
executable: /Users/stanmart/Library/Caches/rattler/cache/cached-envs-v0/ipython-c454637721dc3b0a/bin/python3.13
   machine: macOS-15.4-arm64-arm-64bit-Mach-O

Python dependencies:
     narwhals: 1.35.0
       pandas:
       polars:
         cudf:
        modin:
      pyarrow:
        numpy: 2.2.4
```

### Relevant log output

```shell

```

dangotbanned:
Closes #2394

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2394

## Checklist

- [x] Tests added
- [ ] Documented the changes

- Still keeping this utility internal
- We didn't see an issue, as it was always used on the same branch as an `int` guard

### DIFF
--- a/narwhals/dependencies.py
+++ b/narwhals/dependencies.py
@@ -13,7 +13,6 @@ if TYPE_CHECKING:
     import duckdb
     import ibis
     import modin.pandas as mpd
-    import numpy as np
     import pandas as pd
     import polars as pl
     import pyarrow as pa
@@ -32,6 +31,7 @@ if TYPE_CHECKING:
     from narwhals.typing import _1DArray
     from narwhals.typing import _2DArray
     from narwhals.typing import _NDArray
+    from narwhals.typing import _NumpyScalar
     from narwhals.typing import _ShapeT
 
 # We silently allow these but - given that they claim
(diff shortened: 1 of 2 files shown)


FBruzzesi (approved):
Thanks @dangotbanned 😎

DolphinGriddyGIF (3)

PR #2401, opened 2025-04-19
feat: Add support for DuckDB Enum

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -130,9 +130,8 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
         if not enum_content:  # pragma: no cover
             msg = "Can not cast / initialize Enum without categories present"
             raise ValueError(msg)
-        values_pattern = r"'(.*?)'"
-        values = re.findall(values_pattern, enum_content.group(1))
-        return dtypes.Enum(values)
+        categories = re.findall(r"'(.*?)'", enum_content.group(1))
+        return dtypes.Enum(categories)
 
     if duckdb_dtype.startswith("STRUCT"):
         matchstruc_ = re.findall(r"(\w+)\s+(\w+)", duckdb_dtype)
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -122,7 +122,7 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
     if duckdb_dtype == "INTERVAL":
         return dtypes.Duration()
     if duckdb_dtype.startswith("ENUM"):
-        if version is Version.V1:
+        if version is Version.V1:  # pragma: no cover
             msg = "Converting to Enum is not supported in narwhals.stable.v1"
             raise NotImplementedError(msg)
 
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -8,6 +8,7 @@ from typing import Sequence
 
 import duckdb
 
+from narwhals.utils import Version
 from narwhals.utils import import_dtypes_module
 from narwhals.utils import isinstance_or_issubclass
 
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
ooh nice, thanks!

@camriddell fancy taking a look at this one?

narwhals/_duckdb/utils.py
```diff
@@ -120,6 +121,19 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
         return dtypes.Boolean()
     if duckdb_dtype == "INTERVAL":
         return dtypes.Duration()
+    if duckdb_dtype.startswith("ENUM"):
+        if version is Version.V1:  # pragma: no cover
+            msg = "Converting to Enum is not supported in narwhals.stable.v1"
+            raise NotImplementedError(msg)
```
FBruzzesi, line 127: I mean in principle we have Enum in v1 but we cannot store the categories. Not sure what we should do in this case

narwhals/_duckdb/utils.py
```diff
@@ -120,6 +121,19 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
...
+
+        enum_content = re.search(r"ENUM\((.*?)\)", duckdb_dtype)
+        if not enum_content:  # pragma: no cover
+            msg = "Can not cast / initialize Enum without categories present"
+            raise ValueError(msg)
```
FBruzzesi, line 132: This should never be the case

narwhals/_duckdb/utils.py
```diff
@@ -120,6 +121,19 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
...
+        values_pattern = r"'(.*?)'"
+        values = re.findall(values_pattern, enum_content.group(1))
+        return dtypes.Enum(values)
```
FBruzzesi, line 135:
```suggestion
        categories = re.findall(r"'(.*?)'", enum_content.group(1))
        return dtypes.Enum(categories)
```

narwhals/_duckdb/utils.py
MarcoGorelli: i think what you'd done is fine

narwhals/_duckdb/utils.py
```diff
@@ -120,6 +121,18 @@ def native_to_narwhals_dtype(duckdb_dtype: str, version: Version) -> DType:
...
+    if duckdb_dtype.startswith("ENUM"):
+        if version is Version.V1:  # pragma: no cover
+            msg = "Converting to Enum is not supported in narwhals.stable.v1"
+            raise NotImplementedError(msg)
+
+        enum_content = re.search(r"ENUM\((.*?)\)", duckdb_dtype)
+        if not enum_content:  # pragma: no cover
+            msg = "Can not cast / initialize Enum without categories present"
+            raise ValueError(msg)
+        categories = re.findall(r"'(.*?)'", enum_content.group(1))
```
MarcoGorelli, line 133: TIL that duckdb types have a `children` attribute, which we could use instead of this to make it more robust
```python
In [30]: duckdb.sql("""SELECT 'happy'::ENUM ('sad', 'ok', 'happy')""").types[0]
Out[30]: ENUM('sad', 'ok', 'happy')

In [31]: duckdb.sql("""SELECT 'happy'::ENUM ('sad', 'ok', 'happy')""").types[0].children[0][1]
Out[31]: ['sad', 'ok', 'happy']
```
could we use that?

MarcoGorelli (approved):
thanks @FBruzzesi 

we should probably investigate if there's a way to build duckdb types from python

FBruzzesi:
> thanks @FBruzzesi
> 
> we should probably investigate if there's a way to build duckdb types from python

So the fun fact is that there is a way, yet this PR/feature makes it un-usable:

```py
enum_type(
    name="enum_type",
    type=duckdb_types.VARCHAR,
    values=list(dtype.categories)
)
```
> duckdb.duckdb.NotImplementedException: Not implemented Error: enum_type creation method is not implemented yet

🥲

PR #2406, opened 2025-04-20
chore: DuckDB dtype parsing using `.id` and `.children` properties

### LINKED ISSUE #2402: chore: use `.id` and `.children` for parsing duckdb types
We're currently doing string-matching, but our code would be much more robust if we used the `.children` attribute

For example, if we have
```python
In [47]: rel
Out[47]:
┌───────────┐
│     a     │
│  int64[]  │
├───────────┤
│ [1, 2, 3] │
│ [4, 5]    │
└───────────┘
```
then we can get the type robustly with:
```python
In [48]: rel.types[0].id
Out[48]: 'list'
```
and the inner type with
```python
In [49]: rel.types[0].children
Out[49]: [('child', BIGINT)]

In [50]: rel.types[0].children[0][1].id
Out[50]: 'bigint'
```

## Related
- https://github.com/narwhals-dev/narwhals/pull/2401#discussion_r2051683215

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2402 

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import re
 from functools import lru_cache
 from typing import TYPE_CHECKING
 from typing import Any
@@ -12,6 +11,8 @@ from narwhals.utils import import_dtypes_module
 from narwhals.utils import isinstance_or_issubclass
 
 if TYPE_CHECKING:
+    from duckdb.typing import DuckDBPyType
+
     from narwhals._duckdb.dataframe import DuckDBLazyFrame
     from narwhals._duckdb.expr import DuckDBExpr
     from narwhals.dtypes import DType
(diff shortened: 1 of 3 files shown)


narwhals/_duckdb/utils.py
```diff
@@ -79,75 +80,68 @@ def evaluate_exprs(
...
+            shape.insert(0, size[1])
+
+        inner = native_to_narwhals_dtype(child[1], version=version)
+        return dtypes.Array(inner=inner, shape=tuple(shape))
+
+    return _non_nested_native_to_narwhals_dtype(duckdb_dtype_id, version)
+
+
+@lru_cache(maxsize=16)
+def _non_nested_native_to_narwhals_dtype(duckdb_dtype_id: str, version: Version) -> DType:
```
FBruzzesi, line 118: As suggested in [#2402 comment](https://github.com/narwhals-dev/narwhals/issues/2402#issuecomment-2817191433)

narwhals/_duckdb/utils.py
```diff
@@ -79,75 +80,68 @@ def evaluate_exprs(
...
+    if duckdb_dtype_id == "array":
+        child, size = duckdb_dtype.children
+        shape: list[int] = [size[1]]
+
+        while child[1].id == "array":
+            child, size = child[1].children
+            shape.insert(0, size[1])
+
+        inner = native_to_narwhals_dtype(child[1], version=version)
+        return dtypes.Array(inner=inner, shape=tuple(shape))
```
FBruzzesi, line 112: Arguably, this is the only tricky dtype. The reason being that multi dimensional arrays are parsed from the outer to the inner shape. As an example from our test suite:

```py
duckdb_dtype
# BIGINT[3][2]

child, size = duckdb_dtype.children

child
# ('child', BIGINT[3])

child[1].id
# 'array'

size
# ('size', 2)
```
the expected narwhals type is `Array(Int64(), (3, 2))`, yet if we were to run it recursively, we would end up with `Array(Int64(), (2, 3))` instead

MarcoGorelli (approved):
nice, really satisfying!

narwhals/_duckdb/utils.py
```diff
@@ -79,75 +80,68 @@ def evaluate_exprs(
...
+        "timestamp": dtypes.Datetime(),
+        # TODO(marco): is UTC correct, or should we be getting the connection timezone?
+        # https://github.com/narwhals-dev/narwhals/issues/2165
+        "timestamp with time zone": dtypes.Datetime(time_zone="UTC"),
+        "boolean": dtypes.Boolean(),
+        "interval": dtypes.Duration(),
+        "decimal": dtypes.Decimal(),
+        "time": dtypes.Time(),
+        "blob": dtypes.Binary(),
+    }.get(duckdb_dtype_id, dtypes.Unknown())
```
FBruzzesi, line 144: @MarcoGorelli sorry for the direct ping - I was wondering if should we create this dict for V1 and MAIN on top level on the module instead of for each call, or do it in another way? I didn't put too much thought into it 🤔

MarcoGorelli: I'd imagine this is fine

FBruzzesi: Merging then 😉

PR #2407, opened 2025-04-20
perf: cache `columns` property for lazyframe's

### LINKED ISSUE #2403: perf: cache `.columns` like we cache `.schema`
For lazyframes we generally cache `.schema`:

https://github.com/narwhals-dev/narwhals/blob/db1c5a34784aca2cd1951da8c664f17a8f47cad3/narwhals/_duckdb/dataframe.py#L204-L215

For `.columns`, we should probably use the same logic, and have `_cached_columns`, but if `_cached_schema` was already calculated, then we use that

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related [issue #\<issue number\>](https://github.com/narwhals-dev/narwhals/pull/2325#discussion_r2051684847)
- Closes #2403 

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -192,7 +192,7 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
             new_columns_map.pop(name).alias(name)
             if name in new_columns_map
             else col(name)
-            for name in self.native.columns
+            for name in self.columns
         ]
         result.extend(value.alias(name) for name, value in new_columns_map.items())
         return self._with_native(self.native.select(*result))
@@ -223,7 +223,7 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
                 if self._cached_schema is not None
                 else self.native.columns
             )
-        return self._cached_columns.copy()
+        return self._cached_columns
 
     def to_pandas(self) -> pd.DataFrame:
         # only if version is v1, keep around for backcompat
@@ -295,9 +295,9 @@ class DuckDBLazyFrame(CompliantLazyFrame["DuckDBExpr", "duckdb.DuckDBPyRelation"
             )
 
         if native_how in {"inner", "left", "cross", "outer"}:
-            select = [col(f'lhs."{x}"') for x in self.native.columns]
-            for name in other.native.columns:
-                col_in_lhs: bool = name in self.native.columns
+            select = [col(f'lhs."{x}"') for x in self.columns]
+            for name in other.columns:
+                col_in_lhs: bool = name in self.columns
                 if native_how == "outer" and not col_in_lhs:
                     select.append(col(f'rhs."{name}"'))
                 elif (native_how == "outer") or (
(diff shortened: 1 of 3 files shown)


narwhals/_spark_like/dataframe.py
```diff
@@ -187,7 +188,13 @@ def _iter_columns(self) -> Iterator[Column]:
...
     @property
     def columns(self) -> list[str]:
-        return list(self.schema)
+        if self._cached_columns is None:
+            self._cached_columns = (
+                list(self.schema)
+                if self._cached_schema is not None
+                else self.native.columns
+            )
+        return self._cached_columns.copy()
```
FBruzzesi, line 197: spark-like is the main reason to return a copy. For `.join` we have tests for joining a dataframe with itself, which might be  not so rare; now within the compliant `join` method, we instantiate `self.columns` and `other.columns` - if we don't get copies, then we end up with the most classical python list error.

Question is: should we move the copy there or be safe-er have have it in `.columns`? For now I added a `.copy` call in each method, but I am curious to hear your thoughts

MarcoGorelli: not sure i see the issue, sorry - when we join it with itself, we create a new `narwhals.LazyFrame`, no?

MarcoGorelli: maybe i need to try running it to see 😄

MarcoGorelli: ok i see, it's the line
```
col_order = left_columns
```
and then we append to `col_order`

shouldnt' we just do
```python
col_order = left_columns.copy()
```
?

narwhals/_duckdb/dataframe.py
```diff
@@ -216,7 +217,13 @@ def schema(self) -> dict[str, DType]:
...
     @property
     def columns(self) -> list[str]:
-        return list(self.schema)
+        if self._cached_columns is None:
+            self._cached_columns = (
+                list(self.schema)
+                if self._cached_schema is not None
+                else self.native.columns
+            )
+        return self._cached_columns.copy()
```
MarcoGorelli, line 226: tests pass without the copy, we always create a new list in `join` for duckdb

narwhals/_spark_like/dataframe.py
MarcoGorelli: or return a tuple to make sure nobody can mutate it 😉

dangotbanned: > or return a tuple to make sure nobody can mutate it 😉

Please 🙏🙏🙏

We should really only be using a list when:
- Passing to a backend that demands it
- We are actively mutating the sequence **and** need random access

Otherwise, use of these depending on the actual task
- `tuple`
- `set`
- `deque`
- `Iterator`

FBruzzesi: (Sorry I missed this last comment notification)

@dangotbanned I would agree, yet here that's just what is natively returned (which coincide with what polars does) 

Not sure if you are suggesting to store `._cached_columns` as a tuple, and `.columns` to return `list(self._cached_columns)`

MarcoGorelli (approved):
thanks @FBruzzesi !

narwhals/_spark_like/dataframe.py
MarcoGorelli: splitting this off into https://github.com/narwhals-dev/narwhals/issues/2408

Issue #805, opened 2024-08-17
[Enh]: Better benchmarking routine


FBruzzesi:
### Please describe the purpose of the new feature or describe the problem to solve.

There are some features that can require extra attention or are worth benchmarking to understand if worth implementing. For example I am thinking of #500 and #743.

### Suggest a solution if possible.

I checked how other libraries do that, specifically pydantic. They use [codspeed](https://codspeed.io/) which seems to have a free tier for public repos.

Question is: what to benchmark?! Would TPCH queries in main vs branch be a reasonable test?

### If you have tried alternatives, please describe them below.

Currently very manual effort on kaggle

DeaMariaLeon:
Hey, I installed Codspeed for benchmarking on pydata/sparse. I'm transforming benchmarks from asv to codspeed "as we speak".  

>Would TPCH queries in main vs branch be a reasonable test?

Codpeed tests the opened PR against main, if that was the question. 🤔 
It can be setup to block the merge if there is a regression. It sends the report as a comment on the PR. 
If you need details (or help) let me know. 
Those are my 15 cents. 😇

FBruzzesi:
Hey Dea, thanks for the input. That's what happens when I open issues in a rush. Let me try to clarify some points and ideas.

My understanding is that one can mark some test for benchmarking, and I am wondering what could these test be.

> TPCH tests with a lot of data, doesn't it?

One option is to run TPCH queries with the subset of the data we have in `tests/data/` folder. It should not take as long as the actual TPCH benchmarking.

> Codpeed tests the opened PR against main, if that was the question. 🤔

Yes that is exactly my point: PR (branch) vs main, so I am getting the process right  👌
I wonder though if it could be trigger only, as it is definitly an overkill for most PRs.

> If you need details (or help) let me know.

I have never used it so far, I am happy to give it a spin, but expect to be pinged for help 🙈

DeaMariaLeon:
>I wonder though if it could be trigger only, as it is definitly an overkill for most PRs.

It doesn't say in the documentation. I guess one could do it "somehow" with the CI, but I don't think that it's an out-of-the-box option. You can select if the report is sent all the time to the PR, or only if there is a failure/improvement... but that's all they mention.

>but expect to be pinged for help

I truly doubt that you'll ever need help from me 😁 .. but sure!

FBruzzesi:
[...]

Issue #566, opened 2024-07-21
feat: add "lazy-only" level of support


MarcoGorelli:
If we look at:
- dask.dataframe
- ibis (personally I'd prefer to wait for them to complete the removal of pandas+pyarrow as required dependencies for all backends)
- duckdb (either via Ibis, if https://github.com/ibis-project/ibis/issues/9629 is addressed, or directly via their relational API)

then it doesn't seem too hard to support the `narwhals.LazyFrame` API without running into footguns. We may need to disallow operations which operate on multiple columns' row orders independently (the alternative is doing a join on row order behind the scenes, which I'm not too keen, as we'd be making an expensive operation look cheap)

Each one can then have an associated eager class, which is what things get transformed to if you call `.collect`:
- dask goes to pandas
- ibis goes to pyarrow
- duckdb goes to pyarrow

We can start with this, and possibly consider expanding support later. As they say - "in open source, 'no' is temporary, but 'yes' is forever" - doubly so in a library like Narwhals with a stable API policy 😄 

Furthermore, we'd like to make it easier to [extend Narwhals](https://narwhals-dev.github.io/narwhals/extending/#extending-narwhals), so that if library authors disagree with us and want to add complexity which we wouldn't be happy with maintaining, they can be free to extend Narwhals themselves - all they'd need to do is to implement a few dunder-methods

binste:
I'm a fan of Ibis and use it regularly, mainly together with https://github.com/binste/dbt-ibis and I find it exciting to see that you're looking at it as a potential layer around duckdb.

However, given that one of the benefits of narwhals is a low overhead, I just wanted to note that all the type validation and abstractions in Ibis do add overhead which can add up. It might not be noticeable when working interactively with larger amounts of data but I had use cases in web applications where I had to rewrite code back from Ibis to pure Duckdb or Pandas due to this.

Here's a quick example, loosely based on [this DuckDB-Ibis example](https://duckdb.org/docs/guides/python/ibis). Ibis 9.1, duckdb 1.0.

```python
import ibis

# Create some example database
con = ibis.connect("duckdb://penguins.ddb")
con.create_table(
    "penguins", ibis.examples.penguins.fetch().to_pyarrow(), overwrite = True
)
```

```python
%%timeit
# reconnect to the persisted database (dropping temp tables)
con = ibis.connect("duckdb://penguins.ddb")
penguins = con.table("penguins")
penguins = (
[...]

PR #2414, opened 2025-04-21
perf: Delay converting categories to tuple for Enum

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -14,6 +14,7 @@ from typing import cast
 import pandas as pd
 
 from narwhals._compliant.series import EagerSeriesNamespace
+from narwhals.dtypes import _DelayedCategories
 from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import DuplicateError
 from narwhals.exceptions import ShapeError
@@ -311,7 +312,7 @@ def native_categorical_to_narwhals_dtype(
     if version is Version.V1:
         return dtypes.Categorical()
     if native_dtype.ordered:
-        return dtypes.Enum(get_categories)
+        return dtypes.Enum(_DelayedCategories(get_categories))
     return dtypes.Categorical()
 
 
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -5,6 +5,7 @@ import re
 from contextlib import suppress
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import Callable
 from typing import Sequence
 from typing import Sized
 from typing import TypeVar
@@ -255,7 +256,7 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> 
         return dtypes.Categorical()
     if dtype == "category":
         return native_categorical_to_narwhals_dtype(
-            native_dtype, version, native_dtype.categories
+            native_dtype, version, lambda: tuple(native_dtype.categories)
         )
     if (match_ := PATTERN_PD_DATETIME.match(dtype)) or (
         match_ := PATTERN_PA_DATETIME.match(dtype)
@@ -302,13 +303,15 @@ def object_native_to_narwhals_dtype(
 
 
 def native_categorical_to_narwhals_dtype(
-    native_dtype: pd.CategoricalDtype, version: Version, categories: pd.Index[Any]
+    native_dtype: pd.CategoricalDtype,
+    version: Version,
+    get_categories: Callable[[], tuple[str, ...]],
 ) -> DType:
     dtypes = import_dtypes_module(version)
     if version is Version.V1:
         return dtypes.Categorical()
     if native_dtype.ordered:
-        return dtypes.Enum(categories)
+        return dtypes.Enum(get_categories)
     return dtypes.Categorical()
 
 
@@ -331,7 +334,7 @@ def native_to_narwhals_dtype(
     if str_dtype == "category" and implementation.is_cudf():
         # https://github.com/rapidsai/cudf/issues/18536
         return native_categorical_to_narwhals_dtype(
-            native_dtype, version, native_dtype.categories.to_pandas()
+            native_dtype, version, lambda: tuple(native_dtype.categories.to_pandas())
         )
     if str_dtype != "object":
         return non_object_native_to_narwhals_dtype(native_dtype, version)
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -502,6 +502,8 @@ class Enum(DType):
         return hash((self.__class__, tuple(self.categories)))
 
     def __repr__(self) -> str:  # pragma: no cover
+        if self._categories is None:
+            return f"{type(self).__name__}(categories=[...])"
         return f"{type(self).__name__}(categories={list(self.categories)!r})"
 
 
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -476,13 +476,16 @@ class Enum(DType):
         self, categories: Iterable[str] | type[enum.Enum] | Callable[[], tuple[str, ...]]
     ) -> None:
         self._get_categories: Callable[[], tuple[str, ...]]
+        self._categories = tuple[str, ...] | None
         if isinstance(categories, type) and issubclass(categories, enum.Enum):
             self._get_categories = lambda: tuple(member.value for member in categories)
+            self._categories = self._get_categories()
         elif callable(categories):
             self._get_categories = categories
+            self._categories = None
         else:
             self._get_categories = lambda: tuple(categories)
-        self._categories: tuple[str, ...] | None = None
+            self._categories = self._get_categories()
 
     @property
     def categories(self) -> tuple[str, ...]:
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -491,10 +491,10 @@ class Enum(DType):
         self._delayed_categories: _DelayedCategories | None = None
         self._cached_categories: tuple[str, ...] | None = None
 
-        if isinstance(categories, type) and issubclass(categories, enum.Enum):
-            self._cached_categories = tuple(member.value for member in categories)
-        elif isinstance(categories, _DelayedCategories):
+        if isinstance(categories, _DelayedCategories):
             self._delayed_categories = categories
+        elif isinstance(categories, type) and issubclass(categories, enum.Enum):
+            self._cached_categories = tuple(member.value for member in categories)
         else:
             self._cached_categories = tuple(categories)
 
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -476,7 +476,7 @@ class Enum(DType):
         self, categories: Iterable[str] | type[enum.Enum] | Callable[[], tuple[str, ...]]
     ) -> None:
         self._get_categories: Callable[[], tuple[str, ...]]
-        self._categories = tuple[str, ...] | None
+        self._categories: tuple[str, ...] | None
         if isinstance(categories, type) and issubclass(categories, enum.Enum):
             self._get_categories = lambda: tuple(member.value for member in categories)
             self._categories = self._get_categories()
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -461,6 +461,18 @@ class Categorical(DType):
     """
 
 
+class _DelayedCategories:
+    """Store callable which produces tupleified version of Enum's categories.
+
+    !!! warning
+        This class is not meant to be instantiated directly and exists
+        for internal usage only.
+    """
+
+    def __init__(self, get_categories: Callable[[], tuple[str, ...]]) -> None:
+        self.get_categories = get_categories
+
+
 class Enum(DType):
     """A fixed categorical encoding of a unique set of strings.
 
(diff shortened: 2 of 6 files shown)


narwhals/_pandas_like/utils.py
```diff
@@ -255,7 +256,7 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) ->
         return dtypes.Categorical()
     if dtype == "category":
         return native_categorical_to_narwhals_dtype(
-            native_dtype, version, native_dtype.categories
+            native_dtype, version, lambda: tuple(native_dtype.categories)
```
MarcoGorelli, line 259: :thinking: wait does it actually delay it 😄 gonna do an experiment

MarcoGorelli: based on
```python
In [1]: def func(x):
   ...:     print('here')
   ...:     return tuple(x)
   ...: 

In [2]: values = [1,2,3]

In [3]: fun = lambda: func(values)

In [4]: fun()
here
Out[4]: (1, 2, 3)
```
it looks like this does indeed delay the `tuple` call

MarcoGorelli: Calling tuple on a Series amounts to iterating over rows in python. For cudf it would mean transferring data from gpu to cpu

For series with large numbers of categories this can be an issue

dangotbanned: Could we wait for a user to report this as an issue?

I'm understanding of wanting to be lightweight, but my first thought is this complicates the public API - for a *theoretical* concern of a single backend 🤔

MarcoGorelli: Most users don't report issues unfortunately 

This affect pandas and polars too

dangotbanned: Could you do a benchmark showing the cost per backend?

FBruzzesi: Chiming in to share some mixed feelings as well. Disclaimer: I didn't follow the Enum PR closely so maybe there is the answer to this question already

@MarcoGorelli what if we do as in polars and store a narwhals Series as categories instead of a tuple? Doing so, we would not need to call tuple on any object

camriddell: > Calling tuple on a Series amounts to iterating over rows in python.

At the same time, are large Series expected to be passed in to Enums? Presumptuously, Enum typed Series are likely to contain a small number of categories (<100 or <1000?) due to the inverse relationship between cardinality and 
 performance.

camriddell:
Doesn't delaying this operation just kick the can down the road to _when_ we materialize the categories?

For example in pandas, we shift the (possibly slow?) `tuple(categories)` call from `Enum.__init__` code to `Enum(…).categories` attribute access,  which happens when we convert the `narwhals.Enum` to a `pandas.CategoricalDtype(…, ordered=True)`.

https://github.com/narwhals-dev/narwhals/blob/70ca2b99512cbd00017d68e7afafcd7be26b79c5/narwhals/_pandas_like/utils.py#L504-L506

In the above pattern we pay the iteration price of converting to a tuple either way, we're just shifting _when_ that price is paid.

---

Furthermore the thunk pattern here is subject to external manipulation as well, which is something that converting immediately to a new container avoided:

```python
>>> categories = [*'abcd']
>>> def get_categories(cats):
...     return tuple(cats)
... 
>>> getter = lambda: get_categories(categories)
>>> getter()
('a', 'b', 'c', 'd')
>>> categories.append('e')
>>> getter()
('a', 'b', 'c', 'd', 'e')
```

---

Are there, perhaps some scenario(s) where this pattern allows us to fully skip the materialization of categories as a tuple and passthrough straight to the backend?

MarcoGorelli:
> Are there, perhaps some scenario(s) where this pattern allows us to fully skip the materialization of categories as a tuple and passthrough straight to the backend?

I'm keen to avoid it when users call `.schema`. If you just call schema to,  say,  select all numeric columns, then you just care that something is Enum, you don't care about the inner categories

MarcoGorelli:
First, thanks all for comments, really appreciate it 🙏 

Regarding external manipulation, if we use a `_DelayedCategories` class for this, would that address the concern? Then, it's not something users could use (unless they go looking for trouble, in which case they're left to their own devices). It also simplifies the typing, which is always nice

To elaborate on why I think this is important:
- I don't we can say how many unique categories users' dtypes should have. It could be 10, and it could be 1 million or 100 times that
- There's still memory savings to be had even if there are millions of unique categories with fewer repetitions
- `df.schema` should be a quick operation, regardless of what's inside

In most Narwhals use-cases, people inspect schemas and then make decisions based on the column types. For those decisions, they typically don't need to know the exact categories of an `Enum` column, they just need to know that it's an `Enum`. If they need the exact categories, then OK, they can get those, and I'm happy to kick the can down the road and do the more expensive part when they ask for it (e.g. `s.dtype.categories`)

To give a concrete example:
```python
In [11]: s = pl.Series([f'a_{i}' for i in range(100_000)])

In [12]: s = s.append(s)

In [13]: s = s.append(s)

In [14]: s.to_frame().estimated_size()
Out[14]: 2755560

In [15]: s.cast(pl.Enum(s.unique())).to_frame().estimated_size()
Out[15]: 2288890

In [16]: s_cast = s.cast(pl.Enum(s.unique()))

In [17]: %time _ = tuple(s_cast.dtype.categories)
CPU times: user 15.3 ms, sys: 0 ns, total: 15.3 ms
Wall time: 15.1 ms
```

`s` ends up having only `400_000` rows here, and we'd end up having 15ms of extra overhead just to inspect the dataframe schema even if we don't care about the concrete categories

tbh I think that's too much, and it looks like [so do some of our users](https://github.com/narwhals-dev/narwhals/issues/566#issuecomment-2243562384)

I think this achieves a nice balance:
- if users just want to know if a column is `Enum`, they can do so cheaply
- if they want to know the exact categories, then it's a little more expensive than always returning the native type, but at least they get a consistent type (`tuple`) for all backends

Happy to acknowledge that I might be wrong here 💯 , just wanted to lay out my reasoning

MarcoGorelli:
🤔 though admittedly, the fact we're ending up with a test which fails in CI but passes locally doesn't give me a tonne of confidence in this

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2819423821

Thanks for explaining @MarcoGorelli 🙂 

I agree with your motivation for this PR - we should avoid introducing overhead where possible.

I want to be sure we've identified and measured the right issue.
I did ask some questions in (https://github.com/narwhals-dev/narwhals/pull/2414#discussion_r2052666219), (https://github.com/narwhals-dev/narwhals/pull/2414#discussion_r2052730737) - but these one's are more narrowly focused on your benchmark 🤓 

## `__iter__` vs `to_list`

> To give a concrete example:
> ...
> `s` ends up having only `400_000` rows here, and we'd end up having 15ms of extra overhead just to inspect the dataframe schema even if we don't care about the concrete categories

I want to zoom in a bit more on the part where I see overhead:

```py
import pandas as pd
import polars as pl

import narwhals as nw

categories = [f"a_{i}" for i in range(100_000)]
series_pl = pl.Series(categories)
enum_pl = pl.Enum(series_pl)

enum_slow_nw = nw.Enum(enum_pl.categories)
```

Couldn't we just use `Series.to_list()` here instead?
Based on an old comment of yours 😉 (https://github.com/narwhals-dev/narwhals/issues/1443#issuecomment-2508957161) - I'd expect that to be faster

```py
enum_maybe_faster_nw = nw.Enum(enum_pl.categories.to_list())
```

It seems like the `pandas` side would also support this:

```py
enum_pd = pd.CategoricalDtype(categories, ordered=True)
nw.Enum(enum_pd.categories.to_list())
```

### Question
Does this have *any* effect on overhead for you?

## Number of categories
Jumping straight to `100_000` without measuring anything in between doesn't give us the full picture.
Personally, I've never seen `pl.Enum` used in that way and *would assume* the more common case is much lower.

### Questions
What is the overhead of the following steps?:
- `5`
- `10`
- `50`
- `100`
- `1_000`
- `5_000`
- `10_000`
- `50_000`
- `100_000`

Is this equal across backends, or higher in specific cases? (https://github.com/narwhals-dev/narwhals/pull/2414#discussion_r2052688917)

## Relative overhead
> we'd end up having **15ms** of extra overhead

I think this figure needs the context of other operations for comparison - particularly (https://github.com/narwhals-dev/narwhals/issues/805#issuecomment-2799042816).

I'm not sure what a dataset with `100_000` unique categories *typically* looks like.
I'd guess it would have *at least* `100_000` rows, but probably an order of magnitude more - or it could be empty! 😅 

### Question
Is **15ms** of overhead noticeably higher than the overhead `narwhals` adds when operating on larger datasets?

narwhals/dtypes.py
```diff
@@ -460,35 +461,62 @@ class Categorical(DType):
...
+    """Store callable which produces tupleified version of Enum's categories.
+
+    !!! warning
+        This class is not meant to be instantiated directly and exists
+        for internal usage only.
+    """
+
+    def __init__(self, get_categories: Callable[[], tuple[str, ...]]) -> None:
+        self.get_categories = get_categories
+
```
dangotbanned, line 474: This way we can avoid adding `categories: _DelayedCategories` 

```suggestion

    def __iter__(self) -> Iterator[str]:
        yield from self.get_categories()

```

narwhals/dtypes.py
```diff
@@ -460,35 +461,62 @@ class Categorical(DType):
...
     Examples:
        >>> import narwhals as nw
        >>> nw.Enum(["beluga", "narwhal", "orca"])
        Enum(categories=['beluga', 'narwhal', 'orca'])
     """
 
-    categories: Sequence[str]
+    def __init__(
+        self, categories: Iterable[str] | type[enum.Enum] | _DelayedCategories
+    ) -> None:
```
dangotbanned, line 490: Dependent on (https://github.com/narwhals-dev/narwhals/pull/2414#discussion_r2054049105)

But checks out after making that change 🙂
```suggestion
    def __init__(self, categories: Iterable[str] | type[enum.Enum]) -> None:
```

narwhals/dtypes.py
```diff
@@ -460,35 +461,63 @@ class Categorical(DType):
...
+
+        if isinstance(categories, _DelayedCategories):
+            self._delayed_categories = categories
+        elif isinstance(categories, type) and issubclass(categories, enum.Enum):
+            self._cached_categories = tuple(member.value for member in categories)
         else:
-            self.categories = tuple(categories)
+            self._cached_categories = tuple(categories)
+
+    @property
```
camriddell, line 502: would [functools.cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) be appropriate here?

dangotbanned: @camriddell there seems to be an issue in CI with that and python `3.13`.

I'm not sure if anyone found the source of the issue, but we did try than for `.schema` but ran into a bug 😔

MarcoGorelli:
> Is 15ms of overhead noticeably higher than the overhead narwhals adds when operating on larger datasets?

The Narwhals overhead currently depends on the complexity of the operation, not on the size of the dataset

> What is the overhead of the following steps?:

sorry i'm travelling now and with limited time, but i don't think it's necessary to time all those sizes

>  Personally, I've never seen pl.Enum used in that way and would assume the more common case is much lower.

Sure, but I don't think we should be making assumptions about users' use-cases

MarcoGorelli:
thanks all for comments, really appreciate it 🙏 

i understand this may be viewed as unnecessary, but i really think it's really important that `nw.from_native(df).schema` be as snappy as possible

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2823238213

@MarcoGorelli now that the rush of getting this in for (https://github.com/narwhals-dev/narwhals/releases/tag/v1.36.0) is out of the way - would you be able to re-read the *`__iter__` vs `to_list`* section of (https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2821189253)?

Based on **your earlier benchmark** (https://github.com/narwhals-dev/narwhals/issues/1443#issuecomment-2508957161) - I expected this to cut the materialization cost (possibly in half?)
I believe this would've addressed @camriddell's point (https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2819158013) of not just shifting the cost around - but actually reducing it as well 🙂

MarcoGorelli:
sure, we can definitely use `to_list` as well 👍 so long as that's in addition to delaying materilisation, rather than instead of it

dangotbanned:
> sure, we can definitely use `to_list` as well 👍 so long as that's in addition to delaying materilisation, rather than instead of it

Yeah @MarcoGorelli that's fine with me 😄 

https://github.com/narwhals-dev/narwhals/pull/2414#discussion_r2052666219
> Why is this concern narrowly-scoped to `Enum` and not in every instance we use a `tuple`?

I did have an idea for generalizing `_DelayedCategories` into `_DelayedIterable` (or `_DeferredIterable`).
Which if we switched to `to_list` could making this change:

```py
# _polars.utils.py

dtypes.Enum(_DelayedCategories(lambda: tuple(dtype.categories)))
dtypes.Enum(_DelayedIterable(dtype.categories.to_list))
```

Which would also avoid creating a new function (`lambda`) and be reusable if we come across other cases you wish to optimize 🙂

PR #2417, opened 2025-04-22
feat: Add support for spark connect

### LINKED ISSUE #2189: [Enh]: Support for Spark Connect

_No response_

Hey everyone,

Going through a migration at from Spark RDD (`pyspark.sql`) to Spark Connect (`pyspark.sql.connect`). Couple differences between the two frameworks that I'm still figuring out ([link](https://docs.databricks.com/aws/en/compute/serverless/limitations#limitations-overview)). However, as of writing, running `nw.from_native` will fail since there is no support for Spark Connect DataFrames.

Would it be possible to support Spark Connect DataFrames? To my understanding:

- the Spark Connect API looks and feels almost the same as the existing Spark RDD API (which narwhals supports)
- pretty sure the only updates narwhals would need to make are in the implementation details (checking `sys.modules` for `pyspark.sql.connect`, importing from `pyspark.sql.connect` when necessary, etc.)

Happy to talk through this some more - also willing to test this out in the meantime to make sure it's possible. Just wanted to get the group's take on it first.

Thanks!

Also looking into SQLFrame - haven't seen any discourse on Spark Connect yet in their repo though.

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related [discord thread](https://discord.com/channels/1235257048170762310/1272835922924273694/1363922011612778576)
- Closes #2189

## Checklist

- [x] Tests added
- [ ] Documented the changes

Opening as draft as:

- datetime tests are failing (infer format and timezone aware)
- cast to struct type is failing (even though the conversion internally seems fine - needs more investigation)
- some naming/choices might need discussion (see code comments)
- very unsure if the CI will work as expected

### DIFF
--- a/.github/workflows/pytest-pyspark.yml
+++ b/.github/workflows/pytest-pyspark.yml
@@ -61,11 +61,11 @@ jobs:
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
 
-      - name: Set up JDK
+      - name: Install Java 17
         uses: actions/setup-java@v4
         with:
-          java-version: '11'
-          distribution: 'temurin'
+          distribution: zulu
+          java-version: 17
 
       - name: install-reqs
         run: uv pip install -e . --group core-tests --group extra --system


narwhals/_namespace.py
```diff
@@ -351,4 +362,8 @@ def is_native_pandas_like(obj: Any) -> TypeIs[_NativePandasLike]:
 
 
 def is_native_spark_like(obj: Any) -> TypeIs[_NativeSparkLike]:
-    return is_native_pyspark(obj) or is_native_sqlframe(obj)
+    return (
+        is_native_sqlframe(obj)
+        or is_native_pyspark(obj)
+        or is_native_pyspark_connect(obj)
```
FBruzzesi, line 368: `is_native_sqlframe` must come first, otherwise `is_native_pyspark_connect` might:

- `(pyspark_sql := get_pyspark_connect()) is not None` -> True
- `pyspark_sql.connect.dataframe.DataFrame` -> raise no attribute `connect`

This is due to the fact that `get_pyspark_connect` is "fake" and actually returns spark.sql. I don't really like this and might come back to it to improve it

narwhals/_spark_like/dataframe.py
```diff
@@ -179,6 +179,13 @@ def _collect_to_arrow(self) -> pa.Table:
                     return pa.Table.from_pydict(data, schema=pa.schema(schema))
                 else:  # pragma: no cover
                     raise
+        elif (
+            self._implementation is Implementation.PYSPARK_CONNECT
+            and self._backend_version < (4,)
+        ):
+            import pyarrow as pa  # ignore-banned-import
+
+            return pa.Table.from_pandas(self.native.toPandas())
```
FBruzzesi, line 188: This...got me a laugh and a tear!

narwhals/_spark_like/dataframe.py
```diff
@@ -294,7 +301,7 @@ def drop(self, columns: Sequence[str], *, strict: bool) -> Self:
         return self._with_native(self.native.drop(*columns_to_drop))
 
     def head(self, n: int) -> Self:
-        return self._with_native(self.native.limit(num=n))
+        return self._with_native(self.native.limit(n))
```
FBruzzesi, line 304: No argument named `num` 🤷🏼‍♀️

narwhals/dependencies.py
```diff
@@ -112,6 +113,11 @@ def get_pyspark_sql() -> Any:
     return sys.modules.get("pyspark.sql", None)
 
 
+def get_pyspark_connect() -> Any:
+    """Get pyspark.sql.connect module (if already imported - else return None)."""
+    return sys.modules.get("pyspark.sql", None)
```
FBruzzesi, line 118:
```suggestion
def get_pyspark_connect() -> Any:
    """Get pyspark.sql module (if already imported - else return None)."""
    return sys.modules.get("pyspark.sql", None)
```

narwhals/utils.py
```diff
@@ -199,6 +200,8 @@ class Implementation(Enum):
     """Ibis implementation."""
     SQLFRAME = auto()
     """SQLFrame implementation."""
+    PYSPARK_CONNECT = auto()
```
FBruzzesi, line 203: This or 
```suggestion
    PYSPARKCONNECT = auto()
```
?

pyproject.toml
```diff
@@ -217,6 +218,9 @@ filterwarnings = [
   'ignore:.*The distutils package is deprecated and slated for removal in Python 3.12:DeprecationWarning:pyspark',
   'ignore:.*distutils Version classes are deprecated. Use packaging.version instead.*:DeprecationWarning:pyspark',
   'ignore:.*is_datetime64tz_dtype is deprecated and will be removed in a future version.*:DeprecationWarning:pyspark',
+  'ignore:.*No Partition Defined for Window operation! Moving all data to a single partition.*:UserWarning:pyspark',
```
FBruzzesi, line 221: This looks quite bad honestly! I will try to do some digging or if anyone knows :)

pyproject.toml
```diff
@@ -217,6 +218,9 @@ filterwarnings = [
...
+  # TODO(FBruzzesi): This is raised in `hist_test.py` from pandas??? How does spark-connect affect/changes pandas behavior?
+  'ignore:.*invalid value encountered in cast.*:RuntimeWarning',
```
FBruzzesi, line 223: This other warning is just WTF 🤔

narwhals/utils.py
dangotbanned: I prefer `PYSPARK_CONNECT`

FBruzzesi:
> Just adding the original tag as well to be careful 🙂 ([#2417 (comment)](https://github.com/narwhals-dev/narwhals/pull/2417#event-17360125974))

Thanks @dangotbanned , as I might need a few iterations before getting the CI right (just to make it start) I didn't want to trigger everything all the time 🙈
But yes, before making a final call to this PR, we should definitly run _all_ the jobs

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2417#issuecomment-2822469146

No worries @FBruzzesi - trust your jhudgement here 😄

.github/workflows/pytest-pyspark.yml
```diff
@@ -35,3 +35,49 @@ jobs:
...
+        run: echo "setuptools<78" | uv pip install -b - -e . "pyspark[connect]==3.5.5" --system
+      - name: show-deps
+        run: uv pip freeze
+
+      - name: Download Spark
+        run: |
+          curl -sL https://downloads.apache.org/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3.tgz
+          tar xzf spark-${SPARK_VERSION}-bin-hadoop3.tgz
+          echo "SPARK_HOME=$PWD/spark-${SPARK_VERSION}-bin-hadoop3" >> $GITHUB_ENV
+          echo "$PWD/spark-${SPARK_VERSION}-bin-hadoop3/bin" >> $GITHUB_PATH
```
FBruzzesi, line 73: So apparently this is not a good idea.

Searching for `start-connect-server.sh` in across all public github actions (yaml files in `.github` folder) here is what we get: [search](https://github.com/search?q=%22start-connect-server.sh%22+language%3AYAML+path%3A%2F%5E%5C.github%5C%2Fworkflows%5C%2F%2F&type=code&p=1).

Which to me does not really look great since all the repos are `apache/spark` forks 🥲 with [one exception](https://github.com/golfalot/sedona/blob/974fc0748c747a6692751de754bf3cd90af30744/.github/workflows/python.yml#L139), which however uses that string to skip the spark connect test.

I will sleep on this and keep looking for alternatives 🤔

narwhals/_namespace.py
FBruzzesi: Fixed!

pyproject.toml
FBruzzesi: I can reproduce this with _local_ session with pyspark and old pandas/python versions

.github/workflows/pytest-pyspark.yml
FBruzzesi: Exciting news! It seems to work with the latest commit (https://github.com/narwhals-dev/narwhals/pull/2417/commits/05e0bcc4c1545e2a8fef787a6900e145e2340a74). It just takes a while to set up

narwhals/_spark_like/dataframe.py
```diff
@@ -142,43 +142,53 @@ def _with_native(self, df: SQLFrameDataFrame) -> Self:
             implementation=self._implementation,
         )
 
+    def _to_arrow_schema(self) -> pa.Schema:  # pragma: no cover
```
FBruzzesi, line 145: Reason for being a method rather than a util function is that we need to access a bunch of instance states

narwhals/_spark_like/expr.py
```diff
@@ -772,7 +774,7 @@ def _rank(_input: Column) -> Column:
             else:
                 order_by_cols = [self._F.asc_nulls_last(_input)]
 
-            window = self._Window().orderBy(order_by_cols)
+            window = self._Window().partitionBy(self._F.lit(1)).orderBy(order_by_cols)
```
FBruzzesi, line 777: Left over from #2429

tests/conftest.py
```diff
@@ -191,13 +196,15 @@ def _constructor(obj: Data) -> PySparkDataFrame:
...
 
-            return (
+            frame = (
                 session.createDataFrame([*zip(*_obj.values())], schema=[*_obj.keys()])
                 .repartition(2)
                 .orderBy(index_col_name)
                 .drop(index_col_name)
             )
 
+            return cast("PySparkDataFrame", frame)
```
FBruzzesi, line 206: @dangotbanned perdoname por mi vida loca 😂

MarcoGorelli (approved):
thanks @FBruzzesi !

narwhals/_spark_like/dataframe.py
```diff
@@ -142,43 +142,53 @@ def _with_native(self, df: SQLFrameDataFrame) -> Self:
...
+        from narwhals._arrow.utils import narwhals_to_native_dtype
+
+        schema: list[tuple[str, pa.DataType]] = []
+        nw_schema = self.collect_schema()
+        native_schema = self.native.schema
+        for key, value in nw_schema.items():
+            try:
+                native_dtype = narwhals_to_native_dtype(value, self._version)
+            except Exception as exc:  # noqa: BLE001,PERF203
+                native_spark_dtype = native_schema[key].dataType  # type: ignore[index]
```
dangotbanned, line 157: @FBruzzesi Could you address this performance issue ([`PERF203`](https://docs.astral.sh/ruff/rules/try-except-in-loop/)) or leave a comment on why it is unavoidable please?

I'd personally `try` to avoid ([`BLE001`](https://docs.astral.sh/ruff/rules/blind-except/)) and using exception handling entirely - but understand they were already here 🙂

dangotbanned (changes requested):
Thanks @FBruzzesi!
I was summoned (https://github.com/narwhals-dev/narwhals/pull/2417#discussion_r2059683788) for typing - but spotted a possible performance regression that would impact other spark-like backends

narwhals/dependencies.py
```diff
@@ -230,6 +236,14 @@ def is_pyspark_dataframe(df: Any) -> TypeIs[pyspark_sql.DataFrame]:
     )
 
 
+def is_pyspark_connect_dataframe(df: Any) -> TypeIs[PySparkConnectDataFrame]:
+    """Check whether `df` is a PySpark Connect DataFrame without importing PySpark."""
+    return bool(
+        (pyspark_sql := get_pyspark_connect()) is not None
+        and isinstance(df, pyspark_sql.dataframe.DataFrame)
```
EdAbati, line 243: nitpick (sorry - just because by reading the variable name I initially thought we were importing `pyspark.sql`)

```suggestion
        (pyspark_connect := get_pyspark_connect()) is not None
        and isinstance(df, pyspark_connect.dataframe.DataFrame)
```

narwhals/_spark_like/expr.py
```diff
@@ -75,7 +75,8 @@ def broadcast(self, kind: Literal[ExprKind.AGGREGATION, ExprKind.LITERAL]) -> Se
 
         def func(df: SparkLikeLazyFrame) -> Sequence[Column]:
             return [
-                result.over(df._Window().partitionBy(df._F.lit(1))) for result in self(df)
+                result.over(df._Window().partitionBy(self._F.lit(1)))
```
EdAbati, line 78: should we also change `df._Window` to `self._Window`?

EdAbati (commented):
Thank you for doing this! Looks great to me, just left minor comments :)

narwhals/_spark_like/dataframe.py
FBruzzesi: In principle we could first try to check if `self.collect_schema()` has any unknown type. If it doesn't then it should be possible to call `.to_arrow()` - if it does then this workaround is needed

.github/workflows/pytest-pyspark.yml
```diff
@@ -35,3 +35,75 @@ jobs:
...
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "pyproject.toml"
+
+      - name: Set up JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: '11'
+          distribution: 'temurin'
```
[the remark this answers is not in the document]
FBruzzesi: > I have ~0 knownledge about Java.

That makes 2 of us

> When do this come from should we add the source in a comment?

I can try to bump the version or follow what they do

tests/conftest.py
dangotbanned: Hopefully fixed in (https://github.com/narwhals-dev/narwhals/pull/2417/commits/f297ac4adab53fc6e41e3c43f5c39c793f630707)

FBruzzesi: Nice thanks! That's one more trick that I need to add to my belt!

dangotbanned: Same trick as the rest `spark_like` typing 😉

You need to pick **one** of the imports for `TYPE_CHECKING`, which goes first. Then the real code goes afterwards. 
A type checker will view the *non*-`TYPE_CHECKING` as unreachable - [since they *"think"* it is always `True`](https://typing.python.org/en/latest/guides/libraries.html#type-checking) - whereas [at runtime it is always `False`](https://github.com/python/cpython/blob/28a2fd031e3e9c28bf32670c86b60ea74d0c470d/Lib/typing.py#L3417-L3418)

https://github.com/narwhals-dev/narwhals/blob/9122aef1072843d92bac28caeee2c9f3950118a1/narwhals/_spark_like/dataframe.py#L73-L98

https://github.com/narwhals-dev/narwhals/blob/9122aef1072843d92bac28caeee2c9f3950118a1/narwhals/_spark_like/utils.py#L243-L272

EdAbati (approved):
Nice! Thank you :)

narwhals/_spark_like/dataframe.py
FBruzzesi: @dangotbanned sorry for the direct ping - let's figure out what to write - to me the explanation be in ln158 is quite good.

I will need your approval to merge 😎

MarcoGorelli:
thanks @FBruzzesi ! going to go ahead an ship this

happy to discuss whether the `PERF203` ruff code can be avoided separately, but i don't think it's a blocker, and it was already there to begin with (probably because i added it 😳 ) it's not introduced by this PR

narwhals/_spark_like/dataframe.py
dangotbanned: Hey sorry I lost this @FBruzzesi

I started trying to address it, but couldn't get the tests working locally 😔

PR #2421, opened 2025-04-23
perf: Optimize, generalize `_DelayedCategories` -> `_DeferredIterable`

### LINKED ISSUE #2412: perf: avoid doing work when mapping ordered categorical to Enum
Currently, when mapping an ordered categorical to Enum, we call `tuple` on the categories, or even `to_pandas()` on the categories for cudf

This should be unnecessary, I think we should only compute the categories as a tuple when it becomes necessary:
- when inspecting `dtype.categories`
- when printing the dtype

dangotbanned:
Closes #2412

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- #2414
- https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2823734330

## Checklist

- [x] Tests added
- [x] Documented the changes

Related to the name change from `_Delayed*` -> `_Deferred`.
I think that's a more common term for a lazy operation. When I see *delay*, it makes me think of time delay instead

- https://peps.python.org/pep-0749/
- https://github.com/ibis-project/ibis/blob/bd926f0fbda023124453b824aa24dc4a50e27d72/ibis/common/deferred.py
- https://discuss.python.org/t/backquotes-for-deferred-expression/70577

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -6,6 +6,8 @@ from contextlib import suppress
 from itertools import chain
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import Callable
+from typing import Literal
 from typing import Sequence
 from typing import Sized
 from typing import TypeVar
@@ -300,20 +302,33 @@ def object_native_to_narwhals_dtype(
 
 
 def native_categorical_to_narwhals_dtype(
-    native_dtype: pd.CategoricalDtype, version: Version
+    native_dtype: pd.CategoricalDtype,
+    version: Version,
+    implementation: Literal[Implementation.CUDF] | None = None,
 ) -> DType:
     dtypes = version.dtypes
     if version is Version.V1:
         return dtypes.Categorical()
     if native_dtype.ordered:
-        return dtypes.Enum(_DeferredIterable(native_dtype.categories.to_list))
+        into_iter = (
+            _cudf_categorical_to_list(native_dtype)
+            if implementation is Implementation.CUDF
+            else native_dtype.categories.to_list
+        )
+        return dtypes.Enum(_DeferredIterable(into_iter))
     return dtypes.Categorical()
 
 
+def _cudf_categorical_to_list(native_dtype: Any) -> Callable[[], list[Any]]:
+    # NOTE: https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.dtypes.categoricaldtype/#cudf.core.dtypes.CategoricalDtype
+    def fn() -> list[Any]:
+        return native_dtype.categories.to_arrow().to_pylist()
+
+    return fn
+
+
 def native_to_narwhals_dtype(
-    native_dtype: Any,
-    version: Version,
-    implementation: Implementation,
+    native_dtype: Any, version: Version, implementation: Implementation
 ) -> DType:
     str_dtype = str(native_dtype)
 
@@ -329,8 +344,9 @@ def native_to_narwhals_dtype(
     if str_dtype == "category" and implementation.is_cudf():
         # https://github.com/rapidsai/cudf/issues/18536
         # https://github.com/rapidsai/cudf/issues/14027
-        # https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.dtypes.categoricaldtype/#cudf.core.dtypes.CategoricalDtype
-        native_dtype = native_dtype.to_pandas()
+        return native_categorical_to_narwhals_dtype(
+            native_dtype, version, Implementation.CUDF
+        )
     if str_dtype != "object":
         return non_object_native_to_narwhals_dtype(native_dtype, version)
     elif implementation is Implementation.DASK:
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -5,7 +5,6 @@ import re
 from contextlib import suppress
 from typing import TYPE_CHECKING
 from typing import Any
-from typing import Callable
 from typing import Sequence
 from typing import Sized
 from typing import TypeVar
@@ -13,12 +12,12 @@ from typing import TypeVar
 import pandas as pd
 
 from narwhals._compliant.series import EagerSeriesNamespace
-from narwhals.dtypes import _DelayedCategories
 from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import DuplicateError
 from narwhals.exceptions import ShapeError
 from narwhals.utils import Implementation
 from narwhals.utils import Version
+from narwhals.utils import _DeferredIterable
 from narwhals.utils import import_dtypes_module
 from narwhals.utils import isinstance_or_issubclass
 
@@ -255,9 +254,7 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> 
     if dtype.startswith("dictionary<"):
         return dtypes.Categorical()
     if dtype == "category":
-        return native_categorical_to_narwhals_dtype(
-            native_dtype, version, lambda: tuple(native_dtype.categories)
-        )
+        return native_categorical_to_narwhals_dtype(native_dtype, version)
     if (match_ := PATTERN_PD_DATETIME.match(dtype)) or (
         match_ := PATTERN_PA_DATETIME.match(dtype)
     ):
@@ -303,15 +300,13 @@ def object_native_to_narwhals_dtype(
 
 
 def native_categorical_to_narwhals_dtype(
-    native_dtype: pd.CategoricalDtype,
-    version: Version,
-    get_categories: Callable[[], tuple[str, ...]],
+    native_dtype: pd.CategoricalDtype, version: Version
 ) -> DType:
     dtypes = import_dtypes_module(version)
     if version is Version.V1:
         return dtypes.Categorical()
     if native_dtype.ordered:
-        return dtypes.Enum(_DelayedCategories(get_categories))
+        return dtypes.Enum(_DeferredIterable(native_dtype.categories.to_list))
     return dtypes.Categorical()
 
 
@@ -333,9 +328,7 @@ def native_to_narwhals_dtype(
         return arrow_native_to_narwhals_dtype(native_dtype.pyarrow_dtype, version)
     if str_dtype == "category" and implementation.is_cudf():
         # https://github.com/rapidsai/cudf/issues/18536
-        return native_categorical_to_narwhals_dtype(
-            native_dtype, version, lambda: tuple(native_dtype.categories.to_pandas())
-        )
+        return native_categorical_to_narwhals_dtype(native_dtype, version)
     if str_dtype != "object":
         return non_object_native_to_narwhals_dtype(native_dtype, version)
     elif implementation is Implementation.DASK:
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -319,7 +319,9 @@ def native_categorical_to_narwhals_dtype(
     return dtypes.Categorical()
 
 
-def _cudf_categorical_to_list(native_dtype: Any) -> Callable[[], list[Any]]:
+def _cudf_categorical_to_list(
+    native_dtype: Any,
+) -> Callable[[], list[Any]]:  # pragma: no cover
     # NOTE: https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.dtypes.categoricaldtype/#cudf.core.dtypes.CategoricalDtype
     def fn() -> list[Any]:
         return native_dtype.categories.to_arrow().to_pylist()
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -328,7 +328,9 @@ def native_to_narwhals_dtype(
         return arrow_native_to_narwhals_dtype(native_dtype.pyarrow_dtype, version)
     if str_dtype == "category" and implementation.is_cudf():
         # https://github.com/rapidsai/cudf/issues/18536
-        return native_categorical_to_narwhals_dtype(native_dtype, version)
+        # https://github.com/rapidsai/cudf/issues/14027
+        # https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.dtypes.categoricaldtype/#cudf.core.dtypes.CategoricalDtype
+        native_dtype = native_dtype.to_pandas()
     if str_dtype != "object":
         return non_object_native_to_narwhals_dtype(native_dtype, version)
     elif implementation is Implementation.DASK:
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1961,6 +1961,6 @@ class _DeferredIterable(Generic[_T]):
         yield from self._into_iter()
 
     def to_tuple(self) -> tuple[_T, ...]:
-        """Collect and return as a `tuple`."""
+        # Collect and return as a `tuple`.
         it = self._into_iter()
         return it if isinstance(it, tuple) else tuple(it)
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1959,3 +1959,8 @@ class _DeferredIterable(Generic[_T]):
 
     def __iter__(self) -> Iterator[_T]:
         yield from self._into_iter()
+
+    def to_tuple(self) -> tuple[_T, ...]:
+        """Collect and return as a `tuple`."""
+        it = self._into_iter()
+        return it if isinstance(it, tuple) else tuple(it)
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -13,7 +13,9 @@ from typing import TYPE_CHECKING
 from typing import Any
 from typing import Callable
 from typing import Container
+from typing import Generic
 from typing import Iterable
+from typing import Iterator
 from typing import Literal
 from typing import Protocol
 from typing import Sequence
@@ -99,7 +101,7 @@ if TYPE_CHECKING:
     FrameOrSeriesT = TypeVar(
         "FrameOrSeriesT", bound=Union[LazyFrame[Any], DataFrame[Any], Series[Any]]
     )
-    _T = TypeVar("_T")
+
     _T1 = TypeVar("_T1")
     _T2 = TypeVar("_T2")
     _T3 = TypeVar("_T3")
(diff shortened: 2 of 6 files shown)


dangotbanned:
Interesting!

https://github.com/narwhals-dev/narwhals/actions/runs/14616480347/job/41005794828?pr=2421

So `pl.Enum.categories` used to be a `list` but is now  `pl.Series`.

Will need to pinpoint the version that changed

tests/dtypes_test.py
```diff
@@ -435,6 +435,8 @@ def test_enum_categories_immutable() -> None:
     dtype = nw.Enum(["a", "b"])
     with pytest.raises(TypeError, match="does not support item assignment"):
         dtype.categories[0] = "c"  # type: ignore[index]
+    with pytest.raises(AttributeError):
```
dangotbanned, line 438: Using a `match` here is quite a chore across python versions

https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/tests/expr/test_expr.py#L127-L136

narwhals/utils.py
```diff
@@ -1873,3 +1876,13 @@ def convert_str_slice_to_int_slice(
...
+
+
+class _DeferredIterable(Generic[_T]):
+    """Store a callable producing an iterable to defer collection until we need it."""
+
+    def __init__(self, into_iter: Callable[[], Iterable[_T]], /) -> None:
+        self._into_iter: Callable[[], Iterable[_T]] = into_iter
+
+    def __iter__(self) -> Iterator[_T]:
+        yield from self._into_iter()
```
MarcoGorelli, line 1888: bit skeptical about this, is its perf any different from directly calling `tuple` on `into_iter()`?

MarcoGorelli (commented):
thanks @dangotbanned - does this still work for cudf?

regarding the name, i had https://docs.rs/chrono/latest/chrono/format/struct.DelayedFormat.html in mind, but no objections to "deferred"

narwhals/utils.py
dangotbanned: > bit skeptical about this, is its perf any different from directly calling `tuple` on `into_iter()`?
@MarcoGorelli 

I think that *specifically* might be tricky to benchmark.
What we had before is not *quite* equivalent to this - in isolation - but what we get in `Enum` is the same.

There's also two aspects of the refactor that I wanna highlight

## Performance
Using `polars` as an example, we've got two differences at the conversion site:

- By not using a `lambda`, we save the cost of creating a new function **per-`nw.Enum`**
  - I think this could be a factor for packages that create a large number of `Enum`s - but each one doesn't necessarily have a high number of categories
- By switching from (implicit) `.__iter__` to `.to_list` - we skip the intermediate python-side buffering
  - https://github.com/pola-rs/polars/blob/8d30e79a2e63e14d54c47b6e12b901d539ae5b42/py-polars/polars/series/series.py#L1273-L1284
  - https://github.com/pola-rs/polars/blob/8d30e79a2e63e14d54c47b6e12b901d539ae5b42/py-polars/polars/series/series.py#L4142-L4156

```py
import polars as pl

elements = "a", "b", "c"
some_enum = pl.Enum(elements)

into_before = lambda: tuple(some_enum.categories)  # noqa: E731
into_after = some_enum.categories.to_list
```

## Flexibility
The use of `_DeferredIterable` is currently always with `tuple`

https://github.com/narwhals-dev/narwhals/blob/01edb180582446bff31d739daeda0463986f8bd1/narwhals/dtypes.py#L491-L492


```py
import polars as pl

from narwhals.utils import _DeferredIterable

elements = "a", "b", "c"
some_enum = pl.Enum(elements)
into_after = some_enum.categories.to_list
deferred = _DeferredIterable(into_after)

>>> tuple(deferred)
('a', 'b', 'c')
```

What if we want to use it in a place where we need to return a `list`?

```py
>>> list(deferred)
['a', 'b', 'c']
```

A `set`?

```py
>>> set(deferred)
{'a', 'b', 'c'}
```

How about multiple `_DeferredIterable`s?

```py
from itertools import chain

deferred_upper = _DeferredIterable(some_enum.categories.str.to_uppercase().to_list)

>>> dict.fromkeys(chain(deferred, deferred_upper))
{'a': None, 'b': None, 'c': None, 'A': None, 'B': None, 'C': None}
```

What if we wanted to add some validation into the mix?

```py
def validate_first(iterable: Iterable[Any]) -> tuple[str, ...]:
    it = iter(iterable)
    first = next(it)
    if isinstance(first, str):
        return (first, *it)
    msg = f"Expected `str`, but got {type(first).__name__!r}\n{first!r}"
    raise TypeError(msg)

>>> validate_first(deferred)
('a', 'b', 'c')
```

> [!IMPORTANT]
> All of that is already possible, required no changes to the backend/dtype code, and stays efficient because we preserved the laziness 😎 


## Misc links
I'm not referencing any of this stuff, just plopping it here if you want to dive deeper 😉 
- https://peps.python.org/pep-0380/
- https://docs.python.org/3/howto/functional.html#generators
- https://docs.python.org/3/reference/simple_stmts.html#yield
- https://docs.python.org/3/reference/expressions.html#yield-expressions
- https://docs.python.org/3/glossary.html#term-generator

dangotbanned:
> thanks @dangotbanned - does this still work for cudf?

Well spotted - I think I unintentionally lost that part - will add back in!

narwhals/_pandas_like/utils.py
```diff
@@ -333,9 +328,9 @@ def native_to_narwhals_dtype(
         return arrow_native_to_narwhals_dtype(native_dtype.pyarrow_dtype, version)
     if str_dtype == "category" and implementation.is_cudf():
         # https://github.com/rapidsai/cudf/issues/18536
-        return native_categorical_to_narwhals_dtype(
-            native_dtype, version, lambda: tuple(native_dtype.categories.to_pandas())
-        )
+        # https://github.com/rapidsai/cudf/issues/14027
+        # https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.core.dtypes.categoricaldtype/#cudf.core.dtypes.CategoricalDtype
+        native_dtype = native_dtype.to_pandas()
```
dangotbanned, line 333: I'm just going by the docs - haven't got `cudf` installed.

The change is converting the dtype, instead of converting the inner `Index`/`Series`.

@MarcoGorelli which of the two would you want to prioritize?

1. Deferring each time, but *maybe* paying the cost more than once later? (previous)
2. Caching each *possibly* expensive evaluation? (this)

MarcoGorelli:
still not sure i get it, sorry
```python

In [1]: big_tuple = tuple(range(100_000))

In [2]: class Foo:
   ...:     def __init__(self):
   ...:         self.big_tuple = big_tuple
   ...:     def __iter__(self):
   ...:         yield from self.big_tuple
   ...:

In [3]: foo = Foo()

In [4]: %timeit tuple(foo.big_tuple)
39.9 ns ± 1.77 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [5]: %timeit tuple(foo)
4.57 ms ± 54.2 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

isn't it more inefficient to go via `__iter__`, rather than just calling `tuple` on the iterable?

narwhals/_pandas_like/utils.py
MarcoGorelli: there's a kaggle notebook linked in the contributing guide, if you fork that one and run it there on kaggle it should work

i think the `to_pandas` part needs deferring (or `to_arrow`, as we do in `to_list` - that may be better, come to think of it)

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2421#issuecomment-2832620340

@MarcoGorelli does this hold up when `big_tuple` is an iterable *besides* a `tuple`?

The cases we're dealing with are all `list`, which might not be as fast as something that's immutable 🤔

Also you'd still need to call `tuple` on the `list`, which would result in `__iter__` being called anyway right?

dangotbanned:
@MarcoGorelli 
Here's a repeat of (https://github.com/narwhals-dev/narwhals/pull/2414#issuecomment-2819423821) on my PC

<details><summary>Diff (adding previous version back)</summary>
<p>

```diff
diff --git a/narwhals/utils.py b/narwhals/utils.py
index 595269a74..c5ed57216 100644
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1877,3 +1877,8 @@ class _DeferredIterable(Generic[_T]):
 
     def __iter__(self) -> Iterator[_T]:
         yield from self._into_iter()
+
+
+class _DelayedCategories:
+    def __init__(self, get_categories: Callable[[], tuple[str, ...]]) -> None:
+        self.get_categories = get_categories

```

</p>
</details> 

```py
import polars as pl

from narwhals.utils import _DeferredIterable
from narwhals.utils import _DelayedCategories

categories = [f"a_{i}" for i in range(100_000)]
series_pl = pl.Series(categories)
enum_pl = pl.Enum(series_pl)
```

```py
%%timeit -n 1000
before = _DelayedCategories(lambda: tuple(enum_pl.categories))
before.get_categories()
```
> 6 ms ± 7.59 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

```py
%%timeit -n 1000
after = _DeferredIterable(enum_pl.categories.to_list)
tuple(after)
```
> 5.83 ms ± 56.3 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

Given that both of these results are much lower than your time of `15.1 ms` - you might see more of a difference on your end

---

I'd still want to test something that looks at a large number of smaller (but unique) `Enum`s, (as mentioned in https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2061269042).
I would think that is closer to real-world usage and *expect* there might be more of a benefit from fewer `lambda`s created

dangotbanned:
# Benchmark (Part 2)

> **Before**
> 
> > 5.78 ms ± 24.8 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> 
> **After**
> 
> > 5.42 ms ± 11 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
> 
> ## Follow-up
> 
> I'd still want to test something that looks at a large number of smaller (but unique) `Enum`s, (as mentioned in [#2421 (comment)](https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2061269042)). I would think that is closer to real-world usage and _expect_ there might be more of a benefit from fewer `lambda`s created

@MarcoGorelli Thought I'd share this as I was surprised in **how much** of an impact it had

```py
from itertools import batched

import polars as pl

from narwhals.utils import _DeferredIterable
from narwhals.utils import _DelayedCategories

N_CATEGORIES = 100_000
N_BATCH_SIZE = 20

categories = [f"a_{i}" for i in range(N_CATEGORIES)]
enums = [pl.Enum(batch) for batch in batched(categories, N_BATCH_SIZE)]
```

**Before**
```py
%%timeit -n 1000
for enum in enums:
    before = _DelayedCategories(lambda enum_=enum: tuple(enum_.categories))
    result = before.get_categories()
```
> 20.5 ms ± 257 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

**After**
```py
%%timeit -n 1000
for enum in enums:
    after = _DeferredIterable(enum.categories.to_list)
    result = tuple(after)
```
> 9.03 ms ± 20.9 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

`N_BATCH_SIZE = 20` was just the first thing I tested.
I was thinking something more like 5-10 might be more common?

## 10
**Before**
> 34.8 ms ± 208 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

**After**
> 12.1 ms ± 61.7 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)

narwhals/_pandas_like/utils.py
dangotbanned: @MarcoGorelli I went with the `to_arrow().to_pylist()` path in (https://github.com/narwhals-dev/narwhals/pull/2421/commits/c51597a7268aa568a195cbe5444cdc66c348072c)

narwhals/utils.py
dangotbanned: Thanks @MarcoGorelli 

I'll try benchmarking that to compare against the other two, across (https://github.com/narwhals-dev/narwhals/pull/2421#issuecomment-2833369121), (https://github.com/narwhals-dev/narwhals/pull/2421#issuecomment-2835786977) later today

My understanding was that they'd have the same/very similar costs, since `yield from` is just delegating to the same (`list.__iter__`) method called in both situations.
I'm happy to learn I'm wrong though 😄

dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2065983839)
Back once again @MarcoGorelli 

<details><summary>Diff</summary>
<p>

```diff
diff --git a/narwhals/utils.py b/narwhals/utils.py
index d737d045b..e0d9a1553 100644
--- a/narwhals/utils.py
+++ b/narwhals/utils.py
@@ -1959,3 +1959,11 @@ class _DeferredIterable(Generic[_T]):
 
     def __iter__(self) -> Iterator[_T]:
         yield from self._into_iter()
+
+    def to_tuple(self) -> tuple[_T, ...]:
+        return tuple(self._into_iter())
+
+
+class _DelayedCategories:
+    def __init__(self, get_categories: Callable[[], tuple[str, ...]]) -> None:
+        self.get_categories = get_categories

```

</p>
</details>

<details><summary>Benchmark repro</summary>
<p>

## Setup

```py
from itertools import batched

import polars as pl

import narwhals as nw
from narwhals.utils import _DeferredIterable
from narwhals.utils import _DelayedCategories

N_CATEGORIES = 100_000

categories = [f"a_{i}" for i in range(N_CATEGORIES)]
enum_single = pl.Enum(categories)
enums_len_20 = [pl.Enum(batch) for batch in batched(categories, 20)]
enums_len_10 = [pl.Enum(batch) for batch in batched(categories, 10)]
enums_len_5 = [pl.Enum(batch) for batch in batched(categories, 5)]
```
## `enum_single`
```py
%%timeit -n 100
before = _DelayedCategories(lambda: tuple(enum_single.categories))
result = before.get_categories()
```
```py
%%timeit -n 100
after_iter = _DeferredIterable(enum_single.categories.to_list)
result = tuple(after_iter)
```
```py
%%timeit -n 100
after_to_tuple = _DeferredIterable(enum_single.categories.to_list)
result = after_to_tuple.to_tuple()
```
## `enums_len_20`
```py
%%timeit -n 100
for enum in enums_len_20:
    before = _DelayedCategories(lambda enum_=enum: tuple(enum_.categories))
    result = before.get_categories()
```
```py
%%timeit -n 100
for enum in enums_len_20:
    after_iter = _DeferredIterable(enum.categories.to_list)
    result = tuple(after_iter)
```
```py
%%timeit -n 100
for enum in enums_len_20:
    after_to_tuple = _DeferredIterable(enum.categories.to_list)
    result = after_to_tuple.to_tuple()
```
## `enums_len_10`
```py
%%timeit -n 100
for enum in enums_len_10:
    before = _DelayedCategories(lambda enum_=enum: tuple(enum_.categories))
    result = before.get_categories()
```
```py
%%timeit -n 100
for enum in enums_len_10:
    after_iter = _DeferredIterable(enum.categories.to_list)
    result = tuple(after_iter)
```
```py
%%timeit -n 100
for enum in enums_len_10:
    after_to_tuple = _DeferredIterable(enum.categories.to_list)
    result = after_to_tuple.to_tuple()
```
## `enums_len_5`
```py
%%timeit -n 100
for enum in enums_len_5:
    before = _DelayedCategories(lambda enum_=enum: tuple(enum_.categories))
    result = before.get_categories()
```
```py
%%timeit -n 100
for enum in enums_len_5:
    after_iter = _DeferredIterable(enum.categories.to_list)
    result = tuple(after_iter)
```
```py
%%timeit -n 100
for enum in enums_len_5:
    after_to_tuple = _DeferredIterable(enum.categories.to_list)
    result = after_to_tuple.to_tuple()
```

</p>
</details> 

## Results
> I'm happy to learn I'm wrong though 😄
@dangotbanned 

And that I was!

Good call @MarcoGorelli 🧠 - that gives us a nice boost across all the tests I've used so far
<table>
  <thead align="center">
    <tr>
      <th>name</th>
      <th>enum_single</th>
      <th>enums_len_20</th>
      <th>enums_len_10</th>
      <th>enums_len_5</th>
    </tr>
  </thead>
  <tbody align="right">
    <tr>
      <td align="center">before</td>
      <td>5ms</td>
      <td>20ms</td>
      <td>35ms</td>
      <td>64ms</td>
    </tr>
    <tr>
      <td align="center">after_iter</td>
      <td>5ms</td>
      <td>9ms</td>
      <td>12ms</td>
      <td>20ms</td>
    </tr>
    <tr>
      <td align="center">after_to_tuple</td>
      <td>3ms</td>
      <td>5ms</td>
      <td>7ms</td>
      <td>12ms</td>
    </tr>
  </tbody>
</table>

I *think* when we **already** have a `list` - the `tuple` call can use a more optimized route in (https://github.com/python/cpython/blob/bba14c3e0129dc6c36c304018ab60a5580043f55/Objects/tupleobject.c#L371-L389)

Whereas I *thought* all calls to `tuple(iterable)` resulted in `iterable.__iter__()` 😅

dangotbanned: I think a path forward for us that benefits from both
- the performance of (https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2065726257)
- the flexibility shown in (https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2061269042)

Would be to add `_DeferredIterable.to_<collection>` methods (*starting with* `to_tuple`) - so that the caller can choose exactly what they need.
`__iter__` still makes sense to me as the best for the *general case* - but a `tuple` is **measurably better** for this task 🙂

@MarcoGorelli does that work for you as a compromise?

With the alternative being just removing `__iter__` and only having the `to_tuple` method

dangotbanned: Resolved-ish in (https://github.com/narwhals-dev/narwhals/pull/2421/commits/f153da3a30f533dd1f4d123d7bc3a1abeb0dc700)

MarcoGorelli (approved):
nice, thanks @dangotbanned , and thanks for bearing with me on this one 😅

dangotbanned:
> nice, thanks @dangotbanned , and thanks for bearing with me on this one 😅

Thank you @MarcoGorelli 😍 

Appreciate your patience with me as well 😉 

I'm pretty happy with what we arrived at (https://github.com/narwhals-dev/narwhals/pull/2421#discussion_r2066300353) and have some lessons learned for #805

Issue #1942, opened 2025-02-05
Commitment to keep package size down


MarcoGorelli:
Narwhals started off with the objective of being a lightweight compatibility layer. As we've been adding feature and supported backends, the package size has been growing

There's a lot of essential dataframe functionality, and a lot of libraries that people want to support, so some increase in size since the earliest days is expected. But we do need to monitor it, it does need to stay under control, and Narwhals does need to stay lightweight.

**Commitment**: I'd like to suggest a hard commitment that:
- The Narwhals wheel size will never go above 500 kB. It's currently 305 kB
- Narwhals' size on disk will never go above 5000 kB (i.e.: when you make a virtual environment, the difference in size before vs after doing `pip install narwhals`. This includes some cached files which Python generates, but still, I think it's good to monitor the overall size). It's currently 3789 kB.
- Never introduce any required dependencies nor compiled code

https://github.com/narwhals-dev/narwhals/issues/1886 will probably increase our size a bit more. I think that's OK, as Ibis is a library that a few maintainers have said that they want to support. But it does bring us closer to the limits.

Some strategies to reduce size are:
- Reduce overly-long docstrings. Some examples of how to do this are in https://github.com/narwhals-dev/narwhals/pull/1939 and https://github.com/narwhals-dev/narwhals/pull/1915
- More code-sharing. https://github.com/narwhals-dev/narwhals/pull/1876 is a nice example, and I think there's more opportunities to do this
- Directly implement some methods at the Narwhals level, instead of at the compliant level. For example, is `is_duplicated` just the negation of `is_unique`?
- Freeze new features which don't have a use case. `Series.hist` is fine because it's been requested by Marimo, so it has a clear use case. Anything without a clear use-case, I think we may need to put the brakes on, at least until https://github.com/narwhals-dev/narwhals/issues/1886 is resolved
- See if there's any linter configurations that would reduce the size. I really don't want to minify Narwhals - legibility is important - but maybe there's some simple settings we can tweak, like line length / grouping imports / commas, that can reduce size a bit "for free"

Any help towards this goal would be appreciated - thank you, and thank you to everyone who has contributed in any way to Narwhals 🙏

dangotbanned:
This might be only a **neglible** win, but `# type: ignore[no-any-return]` shows up **159** times in `narwhals`.

I think you could remove these by adding [`warn_return_any`](https://mypy.readthedocs.io/en/stable/config_file.html#confval-warn_return_any)` = false` here:

[...]

PR #2427, opened 2025-04-23
docs: Remove inheritable docstrings

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- #1942
- #2068
- #1939

## Checklist

- [ ] Tests added
- [ ] Documented the changes

https://docs.python.org/3/library/inspect.html#inspect.getdoc

AFAIK, adding docstrings in the `v1` classes/methods is probably doing more harm than good:
- Best-case, we're duplicating information that we need to keep in sync
- Worst-case, we override a docstring that *had* examples with a version **without examples**

I think we should only be adding docstrings if they are different to what is on main.
Things like deprecations, breaking changes, etc

### Before
https://github.com/narwhals-dev/narwhals/blob/f47ad1f0f390b747186a425d1ce483dc59322298/narwhals/stable/v1/__init__.py#L233-L244
image

### After
https://github.com/narwhals-dev/narwhals/blob/a571160020f2eaece7c084b0b050eb4efcd735d4/narwhals/stable/v1/__init__.py#L183-L185
image

## Tasks
- [x] Figure out why only [`3.13`](https://github.com/narwhals-dev/narwhals/actions/runs/14627489787/job/41042874806?pr=2427) is working, but all [earlier versions break formatting](https://github.com/narwhals-dev/narwhals/actions/runs/14627489787/job/41043140947?pr=2427)
  - Resolved in (https://github.com/narwhals-dev/narwhals/pull/2427/commits/04eaf9b46d68213bed912cec9bfc5b9d86fd1827)
- [x] Rewrite `stable_api_test` with less copy/pasting
- [x] Remove repeat docs from `v1`
  - [x] `DataFrame`
  - [x] `LazyFrame`
  - [x] `Series`
  - [x] `Expr`
  - [x] `Schema`
  - [x] `Datetime`
  - [x] `Duration`
- [x] Handle the class-level docs
  - See https://github.com/narwhals-dev/narwhals/pull/2427#issuecomment-2827648318

### DIFF
--- a/tests/stable_api_test.py
+++ b/tests/stable_api_test.py
@@ -118,7 +118,8 @@ def test_dataframe_docstrings() -> None:
             assert remove_docstring_examples(
                 doc.replace("import narwhals.stable.v1 as nw", "import narwhals as nw")
             ) == remove_docstring_examples(getattr(df, item).__doc__)
-        else:
+        else:  # pragma: no cover
+            # NOTE: Would only be possible if `main` didn't define a doc
             assert getdoc(getattr(df, item)) is None
 
 
--- a/tests/stable_api_test.py
+++ b/tests/stable_api_test.py
@@ -4,6 +4,7 @@ from datetime import datetime
 from datetime import timedelta
 from inspect import getdoc
 from typing import Any
+from typing import Iterator
 
 import pytest
 
@@ -105,24 +106,29 @@ def test_stable_api_docstrings() -> None:
         assert remove_docstring_examples(v1_doc) == remove_docstring_examples(doc), item
 
 
+def _iter_api_method_docs(obj: Any, *exclude: str) -> Iterator[tuple[str, str]]:
+    for name in dir(obj):
+        if (
+            not name.startswith("_")
+            and name not in exclude
+            and (doc := getdoc(getattr(obj, name)))
+        ):
+            yield name, doc
+
+
 def test_dataframe_docstrings() -> None:
     pytest.importorskip("polars")
     import polars as pl
 
     df_v1 = nw_v1.from_native(pl.DataFrame())
     df = nw.from_native(pl.DataFrame())
-    api = [i for i in df.__dir__() if not i.startswith("_")]
-    for item in api:
-        method = getattr(df_v1, item)
-        if doc_v1 := getdoc(method):
-            doc = getdoc(getattr(df, item))
-            assert doc
-            assert remove_docstring_examples(
-                doc_v1.replace("import narwhals.stable.v1 as nw", "import narwhals as nw")
-            ) == remove_docstring_examples(doc)
-        else:  # pragma: no cover
-            # NOTE: Would only be possible if `main` didn't define a doc
-            assert getdoc(getattr(df, item)) is None
+    for method_name, doc in _iter_api_method_docs(df):
+        doc_v1 = getdoc(getattr(df_v1, method_name))
+        assert doc_v1
+        doc_v1 = doc_v1.replace(
+            "import narwhals.stable.v1 as nw", "import narwhals as nw"
+        )
+        assert remove_docstring_examples(doc_v1) == remove_docstring_examples(doc)
 
 
 def test_lazyframe_docstrings() -> None:
--- a/tests/stable_api_test.py
+++ b/tests/stable_api_test.py
@@ -90,7 +90,7 @@ def test_stable_api_completeness() -> None:
 def test_stable_api_docstrings() -> None:
     main_namespace_api = nw.__all__
     for item in main_namespace_api:
-        if getattr(nw, item).__doc__ is None:
+        if (doc := getdoc(getattr(nw, item))) is None:
             continue
         if item in {"from_native", "narwhalify"}:
             # `eager_or_interchange` param was removed from main namespace,
@@ -100,9 +100,9 @@ def test_stable_api_docstrings() -> None:
             # In v1 this was Polars-only, after that pandas ordered categoricals
             # started to be mapped to it too, so the docstring changed.
             continue
-        v1_doc = remove_docstring_examples(getattr(nw_v1, item).__doc__)
-        nw_doc = remove_docstring_examples(getattr(nw, item).__doc__)
-        assert v1_doc == nw_doc, item
+        v1_doc = getdoc(getattr(nw_v1, item))
+        assert v1_doc is not None
+        assert remove_docstring_examples(v1_doc) == remove_docstring_examples(doc), item
 
 
 def test_dataframe_docstrings() -> None:
--- a/tests/stable_api_test.py
+++ b/tests/stable_api_test.py
@@ -135,38 +135,32 @@ def test_lazyframe_docstrings() -> None:
     pytest.importorskip("polars")
     import polars as pl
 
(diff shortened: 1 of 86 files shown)


dangotbanned:
### TODO
- Figure out why only [`3.13`](https://github.com/narwhals-dev/narwhals/actions/runs/14627489787/job/41042874806?pr=2427) is working, but all earlier versions break formatting https://github.com/narwhals-dev/narwhals/actions/runs/14627489787/job/41043140947?pr=2427

dangotbanned:
Ahhh we had a nice streak of no unrelated downstream CI issues going there for a bit 😢 

https://github.com/narwhals-dev/narwhals/actions/runs/14629063988/job/41047719471?pr=2427

dangotbanned:
# Feedback please!

We have a few options for dealing with the **missing class-level** docstrings

## 1. Revert that part
Simplest option, lowers my code-golf score but no issues beyond that 🙂 

## 2. Getting funky
I can get them to appear **statically**, but **only** when the constructor is called:

image

Locally, I've [adapted the typing I used in `altair`](https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/altair/utils/core.py#L727).
The usage is simply decorating `<subclass>.__init__` with the parent's context

image

### Bad news
But that doesn't extend to hovering over the class itself 😞 

image

## 3. Getting funkier
If we did something more dynamic we could get the class-level doc.
However, that would only be displayed in a notebook/kernel/live environment.

It *might* be possible to combine **2** & **3**, but I don't know if they'd interact well across all environments.

### Personal note
99% of the time I have kernel completions disabled - since I find them too noisy with `pylance` at the same time.

So while I'm presenting this as an option - I'd prefer an alternative for my own sake 😄 

```json
"jupyter.enableKernelCompletions": false
```

dangotbanned:
> thanks @dangotbanned ! tbh I don't think we need class-level docstrings too much, no objections to removing them

Thanks for taking a look @MarcoGorelli

I can't tell if you were saying *either*:
- You're in favor of a particular option in (https://github.com/narwhals-dev/narwhals/pull/2427#issuecomment-2827648318)
- *or* a **fourth** path where we remove the class-level docstrings 😄 

For the example I gave in (https://github.com/narwhals-dev/narwhals/pull/2427#issuecomment-2827648318) I can understand why they might not seem like a big loss - since we're just missing an [admonition](https://squidfunk.github.io/mkdocs-material/setup/extensions/python-markdown/#admonition).

However, there's also `Datetime` and `Duration` which have a bit more utility

### Main
image

### V1 (currently in this PR)
image

### V1 (using **2. Getting funky**)

image

image

Also includes an example we [didn't have before](https://github.com/narwhals-dev/narwhals/blob/41475ade607f773b7145eb9b85d5f85221b9be6a/narwhals/stable/v1/_dtypes.py#L40-L51)

image

<details><summary>Possible with minimal changes to <code>Datetime</code></summary>
<p>

image


</p>
</details>

tests/stable_api_test.py
```diff
@@ -99,62 +101,66 @@ def test_stable_api_docstrings() -> None:
...
-                "import narwhals.stable.v1 as nw", "import narwhals as nw"
-            )
-        ) == remove_docstring_examples(getattr(df, item).__doc__), item
+    ser_v1 = nw_v1.from_native(pl.Series(), series_only=True)
+    ser = nw.from_native(pl.Series(), series_only=True)
+    for method_name, doc in _iter_api_method_docs(ser):
+        doc_v1 = getdoc(getattr(ser_v1, method_name))
+        assert doc_v1
+        doc_v1 = doc_v1.replace(
+            "import narwhals.stable.v1 as nw", "import narwhals as nw"
```
MarcoGorelli, line 161: i don't think we use this one in the docstrings anymore, we can probably remove it?

MarcoGorelli (commented):
thanks @dangotbanned , changes to docstrings look good

could you summarise the gist of the change in tests/stable_api_test.py please?

tests/stable_api_test.py
dangotbanned: Yeah should be safe to remove, I couldn't find any docstrings that had that import.
I tried to find the PR that removed them, but didnt have any luck 😔

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2427/commits/433cc0b5929828ca0d226c1c11f4d208ffd77d88)

tests/stable_api_test.py
```diff
@@ -99,62 +101,57 @@ def test_stable_api_docstrings() -> None:
...
+
+
+def _iter_api_method_docs(obj: Any, *exclude: str) -> Iterator[tuple[str, str]]:
+    for name in dir(obj):
+        if (
+            not name.startswith("_")
+            and name not in exclude
+            and (doc := getdoc(getattr(obj, name)))
+        ):
+            yield name, doc
```
dangotbanned, line 116: > could you summarise the gist of the change in tests/stable_api_test.py please?

## Getting *gist*-y with it

@MarcoGorelli the changes are centred on this new helper generator function, which **mostly** is just replacing the repetition in each of:
- `test_dataframe_docstrings`
- `test_lazyframe_docstrings`
- `test_series_docstrings`

<details><summary>Previous</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/89448231342acda3358efea20419203e6ea35c56/tests/stable_api_test.py#L107-L157

</p>
</details> 

## Notes
- Used [`dir(...)`](https://docs.python.org/3/library/functions.html#dir) instead of [`__dir__()`](https://docs.python.org/3/reference/datamodel.html#object.__dir__)
- Collapsed the various `if <predicate>: continue` statements into one that can be parametrized
  - The groups in `test_lazyframe_docstrings` with comments became the name of variables passed to `*exclude`
    - This is *intended* to make the code *self-documenting*, but it seems to have not worked 😢
- Used [`inspect.getdoc(...)`](https://docs.python.org/3/library/inspect.html#inspect.getdoc) instead of [`__doc__`](https://docs.python.org/3/reference/datamodel.html#method.__doc__)
  - The latter broke following (https://github.com/narwhals-dev/narwhals/pull/2427/commits/0a5c5e7b8a4fa244b387551a26850829d5129201), whereas the next commit (https://github.com/narwhals-dev/narwhals/pull/2427/commits/64f6ef992c03e1c1115639e78ff23e5936903965) worked as documented

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/14693968316/job/41233109700?pr=2427

@MarcoGorelli this seems to have popped up after merging (https://github.com/narwhals-dev/narwhals/pull/2325) 🤔 

```
[XPASS(strict)] 
 =========================== short test summary info ============================
FAILED tests/group_by_test.py::test_group_by_expr[polars[lazy]-0]
= 1 failed, 9881 passed, 147 skipped, 519 xfailed, 1 xpassed in 222.44s (0:03:42) =
Error: Process completed with exit code 1.
```

MarcoGorelli (approved):
there's just been a new polars release, could well be that

thanks @dangotbanned !

PR #2432, opened 2025-04-25
refactor: remove `Implementation._alias` in favor of non-auto values

FBruzzesi:
## Related
- #2324 
- #2371

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

-  [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

This is a breaking change on paper, yet ~one~ two less method to keep in sync

### DIFF
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -680,7 +680,7 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
             if self._backend_version < (1, 3):  # pragma: no cover
                 msg = (
                     f"`fill_null` with `strategy={strategy}` is only available in "
-                    "'DuckDB>=1.3.0'."
+                    "'duckdb>=1.3.0'."
                 )
                 raise NotImplementedError(msg)
 
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -678,7 +678,10 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
     ) -> Self:
         if strategy is not None:
             if self._backend_version < (1, 3):  # pragma: no cover
-                msg = f"`fill_null` with `strategy={strategy}` is only available in DuckDB>=1.3.0."
+                msg = (
+                    f"`fill_null` with `strategy={strategy}` is only available in "
+                    "'DuckDB>=1.3.0'."
+                )
                 raise NotImplementedError(msg)
 
             def _fill_with_strategy(window_inputs: WindowInputs) -> duckdb.Expression:
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -678,10 +678,7 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
     ) -> Self:
         if strategy is not None:
             if self._backend_version < (1, 3):  # pragma: no cover
-                msg = (
-                    f"`fill_null` with `strategy={strategy}` is only available in "
-                    "'duckdb>=1.3.0'."
-                )
+                msg = f"`fill_null` with `strategy={strategy}` is only available in 'duckdb>=1.3.0'."
                 raise NotImplementedError(msg)
 
             def _fill_with_strategy(window_inputs: WindowInputs) -> duckdb.Expression:
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -44,6 +44,7 @@ if TYPE_CHECKING:
     from narwhals._duckdb.typing import WindowFunction
     from narwhals._expression_parsing import ExprMetadata
     from narwhals.dtypes import DType
+    from narwhals.typing import FillNullStrategy
     from narwhals.typing import NonNestedLiteral
     from narwhals.typing import NumericLiteral
     from narwhals.typing import RankMethod
@@ -670,16 +671,41 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
         )
 
     def fill_null(
-        self, value: Self | NonNestedLiteral, strategy: Any, limit: int | None
+        self,
+        value: Self | NonNestedLiteral,
+        strategy: FillNullStrategy | None,
+        limit: int | None,
     ) -> Self:
         if strategy is not None:
-            msg = "todo"
-            raise NotImplementedError(msg)
+            if self._backend_version < (1, 3):  # pragma: no cover
+                msg = f"`fill_null` with `strategy={strategy}` is only available in DuckDB>=1.3.0."
+                raise NotImplementedError(msg)
 
-        def func(_input: duckdb.Expression, value: Any) -> duckdb.Expression:
+            def _fill_with_strategy(window_inputs: WindowInputs) -> duckdb.Expression:
+                order_by_sql = generate_order_by_sql(
+                    *window_inputs.order_by, ascending=True
+                )
+                partition_by_sql = generate_partition_by_sql(*window_inputs.partition_by)
+
+                fill_func = "last_value" if strategy == "forward" else "first_value"
+                _limit = "unbounded" if limit is None else limit
+                rows_between = (
+                    f"{_limit} preceding and current row"
+                    if strategy == "forward"
+                    else f"current row and {_limit} following"
+                )
+                sql = (
+                    f"{fill_func}({window_inputs.expr} ignore nulls) over "
+                    f"({partition_by_sql} {order_by_sql} rows between {rows_between})"
+                )
+                return SQLExpression(sql)  # type: ignore[no-any-return, unused-ignore]
+
+            return self._with_window_function(_fill_with_strategy)
+
+        def _fill_constant(_input: duckdb.Expression, value: Any) -> duckdb.Expression:
             return CoalesceOperator(_input, value)
 
-        return self._with_callable(func, value=value)
+        return self._with_callable(_fill_constant, value=value)
 
     def cast(self, dtype: DType | type[DType]) -> Self:
         def func(_input: duckdb.Expression) -> duckdb.Expression:
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -345,13 +345,7 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "duckdb.Expression"]):
                 raise ValueError(msg)
             return [name]
 
-        return self.__class__(
-            self._call,
-            evaluate_output_names=self._evaluate_output_names,
-            alias_output_names=alias_output_names,
-            backend_version=self._backend_version,
-            version=self._version,
-        )
+        return self._with_alias_output_names(alias_output_names)
 
     def abs(self) -> Self:
         return self._with_callable(lambda _input: FunctionExpression("abs", _input))
--- a/narwhals/_enum.py
+++ b/narwhals/_enum.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+# ruff: noqa: ARG004
 from enum import Enum
 from typing import Any
 
@@ -38,10 +39,7 @@ class NoAutoEnum(Enum):
 
     @staticmethod
     def _generate_next_value_(
-        name: str,  # noqa: ARG004
-        start: int,  # noqa: ARG004
-        count: int,  # noqa: ARG004
-        last_values: list[Any],  # noqa: ARG004
+        name: str, start: int, count: int, last_values: list[Any]
     ) -> Any:
         msg = "Creating values with `auto()` is not allowed. Please provide a value manually instead."
         raise ValueError(msg)
--- a/narwhals/_enum.py
+++ b/narwhals/_enum.py
@@ -6,10 +6,9 @@ from typing import Any
 
 
 class NoAutoEnum(Enum):
-    """Enum base class that prohibits the use of auto() for value assignment.
+    """Enum base class that prohibits the use of enum.auto() for value assignment.
 
-    This class inherits from enum.Enum but overrides the value generation mechanism
-    to prevent automatic value assignment via enum.auto().
+    This behavior is achieved by overriding the value generation mechanism.
 
     Examples:
         >>> from enum import auto
@@ -29,12 +28,7 @@ class NoAutoEnum(Enum):
         ValueError: Creating values with `auto()` is not allowed. Please provide a value manually instead.
 
     Raises:
-        TypeError: If `auto()` is attempted to be used for any enum member value.
-
-    Note:
-        All enum values must be explicitly assigned. This is useful when you want
-        to maintain full control over enum values and prevent automatic numbering.
-
+        ValueError: If `auto()` is attempted to be used for any enum member value.
     """
 
     @staticmethod
--- a/narwhals/_enum.py
+++ b/narwhals/_enum.py
@@ -11,7 +11,7 @@ import sys
 
 if sys.version_info >= (3, 11):
     from enum import StrEnum
-else:
+else:  # pragma: no cover
     from enum import Enum
     from typing import TYPE_CHECKING
 
--- a/narwhals/_enum.py
+++ b/narwhals/_enum.py
@@ -1,0 +1,44 @@
+"""Compatibility module for newer [`enum`] features, like [`StrEnum`].
+
+[`enum`]: https://docs.python.org/3/library/enum.html
+[`StrEnum`]: https://docs.python.org/3/library/enum.html#enum.StrEnum
+"""
+
+from __future__ import annotations
+
+# ruff: noqa: ARG004
+import sys
+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum
+else:
+    from enum import Enum
+    from typing import TYPE_CHECKING
+
+    if TYPE_CHECKING:
+        from typing import Any
+
+        from typing_extensions import Self
+
+    class ReprEnum(Enum): ...
+
+    class StrEnum(str, ReprEnum):
+        def __new__(cls, value: str) -> Self:
+            if not isinstance(value, str):
+                msg = f"{value!r} is not a string"
+                raise TypeError(msg)
+            member = str.__new__(cls, value)
+            member._value_ = value
+            return member
+
+        def __str__(self) -> str:
+            return str(self.value)
+
+        @staticmethod
+        def _generate_next_value_(
+            name: str, start: int, count: int, last_values: list[Any]
+        ) -> Any:
+            return name.lower()
+
+
+__all__ = ["StrEnum"]
--- a/narwhals/_enum.py
+++ b/narwhals/_enum.py
@@ -1,44 +1,50 @@
-"""Compatibility module for newer [`enum`] features, like [`StrEnum`].
-
-[`enum`]: https://docs.python.org/3/library/enum.html
-[`StrEnum`]: https://docs.python.org/3/library/enum.html#enum.StrEnum
-"""
-
 from __future__ import annotations
 
-# ruff: noqa: ARG004
-import sys
-
-if sys.version_info >= (3, 11):
-    from enum import StrEnum
-else:  # pragma: no cover
-    from enum import Enum
-    from typing import TYPE_CHECKING
-
-    if TYPE_CHECKING:
-        from typing import Any
-
-        from typing_extensions import Self
-
-    class ReprEnum(Enum): ...
-
-    class StrEnum(str, ReprEnum):
-        def __new__(cls, value: str) -> Self:
-            if not isinstance(value, str):
-                msg = f"{value!r} is not a string"
-                raise TypeError(msg)
-            member = str.__new__(cls, value)
-            member._value_ = value
-            return member
-
-        def __str__(self) -> str:
-            return str(self.value)
-
-        @staticmethod
-        def _generate_next_value_(
-            name: str, start: int, count: int, last_values: list[Any]
-        ) -> Any:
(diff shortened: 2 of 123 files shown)


narwhals/utils.py
```diff
@@ -199,30 +199,33 @@ def namespace(self) -> type[Namespace[Any]]:
 class Implementation(Enum):
     """Implementation of native object (pandas, Polars, PyArrow, ...)."""
 
-    PANDAS = auto()
-    """Pandas implementation."""
-    MODIN = auto()
+    PANDAS = "pandas"
+    """pandas implementation."""
```
FBruzzesi, line 203: pandas should be lower case - see [citing, Brand and logo](https://pandas.pydata.org/about/citing.html)

tests/dtypes_test.py
```diff
@@ -466,7 +466,7 @@ def test_enum_repr() -> None:
     result = nw.Enum(["a", "b"])
     assert "Enum(categories=['a', 'b'])" in repr(result)
     result = nw.Enum(nw.Implementation)
-    assert "Enum(categories=[1, 2, 3" in repr(result)
+    assert "Enum(categories=['Pandas', 'Modin', 'cuDF'" in repr(result)
```
FBruzzesi, line 469:
```suggestion
    assert "Enum(categories=['pandas', 'Modin', 'cuDF'" in repr(result)
```

narwhals/utils.py
```diff
@@ -1846,7 +1828,7 @@ def _ensure_version(self, instance: _FullContext, /) -> None:
         if instance._backend_version >= self._min_version:
             return
         method = self._wrapped_name
-        backend = instance._implementation._alias
+        backend = instance._implementation
```
FBruzzesi, line 1831: Should we do 
```suggestion
        backend = str(instance._implementation)
```
to be more explicit?

dangotbanned: So in (#2371) I only split it up that way for readability.

If you can get the same output with `str(...)`, then you could use `!s`

```py
msg = f"`{method}` is only available in {backend!s}>={minimum!r}, found version {found!r}."
```

https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals

narwhals/utils.py
```diff
@@ -199,30 +199,33 @@ def namespace(self) -> type[Namespace[Any]]:
...
     """Dask implementation."""
-    DUCKDB = auto()
+    DUCKDB = "DuckDB"
     """DuckDB implementation."""
-    IBIS = auto()
+    IBIS = "Ibis"
     """Ibis implementation."""
-    SQLFRAME = auto()
+    SQLFRAME = "SQLFrame"
     """SQLFrame implementation."""
```
dangotbanned, line 221: You might wanna do a quick search for everywhere these names are currently used (that aren't referencing `Implementation`)

IIRC `"PyArrow"` was used quite a bit - but like (https://github.com/narwhals-dev/narwhals/pull/2432/files/ceab751df9facbbf7bd460357578bda57b47e38b#r2060319755) I don't think that's canonical

dangotbanned:
@FBruzzesi this isn't an objection, just wanted to say I don't think we need to class this as breaking 🙂 

For all we know [`auto()`](https://docs.python.org/3/library/enum.html#enum.auto) could be returning anything.
Our docs don't even display the value https://narwhals-dev.github.io/narwhals/api-reference/implementation/

EdAbati (commented):
I love this semplification!

I am wondering if we could even go further and make the strings consistent with the ones in the dictionary inside the `from_string` method.

If we do (and I suggest to go all lowercase), we can even remove that method and just do `Implementation(str_implementation)` -> `Implementation("pandas") == Implementation.PANDAS` 

Thoughts? :)

(Shame we cannot use yet [`StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum) that is only for Python 3.11 I think)

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2432#pullrequestreview-2794743074

@EdAbati have you checked out (#2324)?

I needed to add all the lowercase strings there, which would be simplified if the `Enum` used the same representation 🙂

FBruzzesi:
> @FBruzzesi this isn't an objection, just wanted to say I don't think we need to class this as breaking 🙂

That's good to hear!

> For all we know [`auto()`](https://docs.python.org/3/library/enum.html#enum.auto) could be returning anything. Our docs don't even display the value https://narwhals-dev.github.io/narwhals/api-reference/implementation/

Yeah I also think that no one is using/should be rely on these values as (by faulty design), UNKNOWN keeps changing each time we add a new backend.

> I am wondering if we could even go further and make the strings consistent with the ones in the dictionary inside the from_string method.

@EdAbati I thought about it (which is what `StrEnum` would do out of the box) - I don't have a strong opinion on it, and it might be worth it for the sake of simplicity (of a class which is mostly of internal use)

narwhals/utils.py
```diff
@@ -206,11 +206,29 @@ def dtypes(self) -> DTypes:
...
+        return str(self.value)
+
+    def _generate_next_value_(
+        name: str,  # noqa: N805
+        start: int,
+        count: int,
+        last_values: list[Any],
+    ) -> Any:
+        # See: https://docs.python.org/3.10/library/enum.html#using-automatic-values
+        return name.lower()
```
dangotbanned, line 228: (https://github.com/narwhals-dev/narwhals/actions/runs/14681912112/job/41205629117?pr=2432)

Seems like this might need more compat code

- `3.11`: https://github.com/python/cpython/blob/ab4746261870bcdbb6e7c5a5b307465fbe517790/Lib/enum.py#L1301-L1305
- `3.12`: https://github.com/python/cpython/blob/aeb3a6f61af53ed3fbf31f0b3704f49b71ac553c/Lib/enum.py#L1335-L1340

dangotbanned: Also fun fact I just learned...

The `enum` module doubled in size between:
- `3.10`: https://github.com/python/cpython/blob/6322edd260e8cad4b09636e05ddfb794a96a0451/Lib/enum.py
- `3.11`: https://github.com/python/cpython/blob/ab4746261870bcdbb6e7c5a5b307465fbe517790/Lib/enum.py

dangotbanned: Hopefully (https://github.com/narwhals-dev/narwhals/pull/2432/commits/261a99419158a285f7fb554d6e8c52bdb5432eec) should work better @FBruzzesi 

We'd also be able to reuse `StrEnum` in other places - rather than the port being tied to `Implementation`-only

FBruzzesi:
@dangotbanned I got a [fix for the CI](https://github.com/narwhals-dev/narwhals/pull/2440)

narwhals/utils.py
```diff
@@ -275,19 +276,10 @@ def from_string(
...
-            "dask": Implementation.DASK,
-            "duckdb": Implementation.DUCKDB,
-            "ibis": Implementation.IBIS,
-            "sqlframe": Implementation.SQLFRAME,
-        }
-        return mapping.get(backend_name, Implementation.UNKNOWN)
+        try:
+            return cls(backend_name)
+        except ValueError:
+            return Implementation.UNKNOWN
```
EdAbati, line 282: Another way to implement this would be to override the `_missing_` `classmethod`, I think: https://docs.python.org/3.14/library/enum.html#enum.Enum._missing_

If we have this fallback on `Implementation.UNKNOWN`, then everywhere in the code when we use `Implementation.from_string(str_)` we can just use `Implementation(str_)`

dangotbanned: @EdAbati 

That does sound nice to have, but from what I can see we're only using `Implementation.from_string` in two places.
Everywhere else is just using `Implementation.from_backend`

https://github.com/narwhals-dev/narwhals/blob/a260885919debcde1d270cb2904b0e2b74d8a3ef/narwhals/utils.py#L298-L315

This test would actually be simpler if we just switched to `Implementation.from_backend` 🤔 

https://github.com/narwhals-dev/narwhals/blob/a260885919debcde1d270cb2904b0e2b74d8a3ef/tests/frame/lazy_test.py#L76-L79

narwhals/utils.py
dangotbanned: The other part I've thought of is `Implementation.PYSPARK_CONNECT`.
AFAICT, that would currently use `"pyspark_connect"`.
Might we want to use one of these instead?

https://github.com/narwhals-dev/narwhals/blob/3605e76e3d73c32273e2045502b11d9d592c00ab/pyproject.toml#L42

I originally added `_alias` for error messages (https://github.com/narwhals-dev/narwhals/pull/2371#discussion_r2049395015) but `pyspark_connect` looks more like a package (but isn't).

I'm not sure if that made any sense 😄

narwhals/utils.py
```diff
@@ -1861,7 +1830,7 @@ class requires:  # noqa: N801
         >>> backend.really_complex_feature()
         Traceback (most recent call last):
             ...
-        NotImplementedError: `really_complex_feature` is only available in PyArrow>='9000.0.0', found version '20.0.0'.
+        NotImplementedError: `really_complex_feature` is only available in pyarrow>='9000.0.0', found version '20.0.0'.
```
dangotbanned, line 1833: If we're going with something that looks more like a package/spec (https://github.com/narwhals-dev/narwhals/pull/2432#discussion_r2066023233) - I'd revise the message to look more like:

```py
"""
NotImplementedError: `really_complex_feature` is only available in "pyarrow>=9000.0.0", found version '20.0.0'.
"""
```

narwhals/utils.py
FBruzzesi: Sorry @dangotbanned I had a long day!
I must have messed up with a merge - Taking a look at this now. 

We could use `SPARK_CONNECT = "pyspark[connect]"`, but let's make sure everyone is agreeing on this

narwhals/_duckdb/expr.py
```diff
@@ -678,7 +678,10 @@ def fill_null(
     ) -> Self:
         if strategy is not None:
             if self._backend_version < (1, 3):  # pragma: no cover
-                msg = f"`fill_null` with `strategy={strategy}` is only available in DuckDB>=1.3.0."
+                msg = (
+                    f"`fill_null` with `strategy={strategy}` is only available in "
+                    "'DuckDB>=1.3.0'."
```
dangotbanned, line 683:
```suggestion
                    "'duckdb>=1.3.0'."
```

narwhals/utils.py
dangotbanned: @FBruzzesi typing is looking good now!

image

FBruzzesi:
> But they _don't_ show up in `dir()`? 🤔

Mmmh this is odd! I can definitly see all the string methods in `dir(...)` and can call them 

<details><summary> screenshot </summary>

image

</details>

> And aren't there at runtime?
> 
> ```python
> >>> nw.Implementation.POLARS.casefold()
> AttributeError: 'Implementation' object has no attribute 'casefold'
> ```

and can call them

image

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2432#issuecomment-2840194090

I'm not at my pc now, but I was running `3.13`.
Are you using something below `3.11` by chance?

FBruzzesi:
> Okay so I know I added `StrEnum` in ([261a994](https://github.com/narwhals-dev/narwhals/commit/261a99419158a285f7fb554d6e8c52bdb5432eec)). But how about we just replace that with our own (e.g. `StrLikeEnum` or a better name) that has the new `Enum` behavior but without the `str` subclass?
> 
> The `auto()` part had it's downsides ([#2432 (comment)](https://github.com/narwhals-dev/narwhals/pull/2432#discussion_r2066019463)) but can be preserved if we want. The stuff in `__new__` I don't think _requires_ mixing in `str`.
> 
> I did a full read-through of (https://docs.python.org/3/howto/enum.html) again last night. I think the main reason we'd use a `str` (data) mix-in - is _if_ **we wanted to inherit methods**. My view is **we don't** - since `Implementation` has a lot of methods that we'd want to prioritize.

Thanks @dangotbanned , yes I think that it's ok to avoid using `StrEnum` - the main advantage was allowing `auto()` out of the box, however as we want to have more control over the values it's ok to not use it.

If we ever need some string method we can always add that later on

narwhals/_enum.py
```diff
@@ -1,44 +1,50 @@
...
 
-[`enum`]: https://docs.python.org/3/library/enum.html
-[`StrEnum`]: https://docs.python.org/3/library/enum.html#enum.StrEnum
-"""
+from enum import Enum
+from typing import Any
 
-from __future__ import annotations
 
-# ruff: noqa: ARG004
```
dangotbanned: 🙄 

https://github.com/narwhals-dev/narwhals/blob/571e2fafe2fc935512af430e5d5c8fb86e780750/narwhals/_enum.py#L39-L45

FBruzzesi: whooops! ctrl+A -> delete the file.
Sorry will change it back

dangotbanned: All good 😄

MarcoGorelli:
I haven't managed to follow everything (quite short on time lately 😩 ) but some thoughts I'd like to share are:

**backwards compatibility**

What we absolutely need to preserve is:
```python
df: nw.DataFrame
s = nw.new_series('a', [1,2,3], implementation=df.implementation)
```
or whatever the exact syntax is

It's true that `UNKNOWN` has been changing between versions, but I think that's the only one, right? Could we just assign say `999` to `UNKNOWN` and keep the others stable?

On board with removing `auto` anyway, it came from a community contribution and i should probably have pushed back on it

dangotbanned:
> **backwards compatibility**
> 
> What we absolutely need to preserve is:
> 
> ```python
> df: nw.DataFrame
> s = nw.new_series('a', [1,2,3], implementation=df.implementation)
> ```
> 
> or whatever the exact syntax is

That part is all good 👍 

> It's true that `UNKNOWN` has been changing between versions, but I think that's the only one, right? Could we just assign say `999` to `UNKNOWN` and keep the others stable?

@MarcoGorelli while I was testing out how the typing works for the new (`str`) `_value_`(s), the only thing I discovered that *could* break is:

```py
import narwhals as nw

>>> nw.Implementation.PANDAS
<Implementation.PANDAS: 1>

>>> nw.Implementation(1)
<Implementation.PANDAS: 1>
```

We haven't documented this and our tests haven't caught anyone doing it, so I don't think there's a huge concern.

You already can't do much with the `int` value, e.g.:

```py
import narwhals as nw

>>> nw.Implementation.PANDAS == 1
False

>>> nw.Implementation.PANDAS is 1
<>:3: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
<>:3: SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
[C:\Users\danie\AppData\Local\Temp\ipykernel_22668\3208834624.py:3](file:///C:/Users/danie/AppData/Local/Temp/ipykernel_22668/3208834624.py:3): SyntaxWarning: "is" with 'int' literal. Did you mean "=="?
  nw.Implementation.PANDAS is 1
False
```

Places we ask for an `Implementation` don't accept an `int`:

```py
import narwhals as nw

>>> nw.new_series("a", [1, 2, 3], backend=1)
File /GitHub/narwhals/narwhals/utils.py:372, in Implementation.to_native_namespace(self)
    369     return pyspark.sql
    371 msg = "Not supported Implementation"  # pragma: no cover
--> 372 raise AssertionError(msg)

AssertionError: Not supported Implementation
```

narwhals/utils.py
```diff
@@ -215,35 +216,37 @@ def dtypes(self) -> DTypes:
...
     """Ibis implementation."""
-    SQLFRAME = auto()
+    SQLFRAME = "sqlframe"
     """SQLFrame implementation."""
-    PYSPARK_CONNECT = auto()
+    PYSPARK_CONNECT = "pyspark[connect]"
     """PySpark Connect implementation."""
-
-    UNKNOWN = auto()
+    UNKNOWN = "unknown"
```
dangotbanned, line 244: Would there be any benefit in differentiating the value here for `Implementation.UNKNOWN`?

Some other options:

```py
UNKNOWN = "UNKNOWN"
UNKNOWN = "<unknown>"
UNKNOWN = "..."
UNKNOWN = "?"
UNKNOWN = "extension"
UNKNOWN = "USER DEFINED"
```

I don't have a strong opinion on either or the current one.
Just thought it *could* be clearer that this one is not like the others 🤔

narwhals/_duckdb/expr.py
```diff
@@ -678,7 +678,10 @@ def fill_null(
     ) -> Self:
         if strategy is not None:
             if self._backend_version < (1, 3):  # pragma: no cover
-                msg = f"`fill_null` with `strategy={strategy}` is only available in DuckDB>=1.3.0."
+                msg = (
+                    f"`fill_null` with `strategy={strategy}` is only available in "
+                    "'duckdb>=1.3.0'."
+                )
```
dangotbanned, line 684:
```suggestion
                msg = f"`fill_null` with `strategy={strategy}` is only available in 'duckdb>=1.3.0'."
```

narwhals/translate.py
```diff
@@ -665,10 +665,15 @@ def _from_native_impl(  # noqa: PLR0915
     elif is_native_spark_like(native_object):  # pragma: no cover
         ns_spark = version.namespace.from_native_object(native_object)
         if series_only:
-            msg = f"Cannot only use `series_only` with {ns_spark.implementation._alias} DataFrame"
+            msg = (
+                f"Cannot only use `series_only` with {ns_spark.implementation} DataFrame"
+            )
```
dangotbanned, line 670:
```suggestion
            msg = f"Cannot only use `series_only` with {ns_spark.implementation} DataFrame"
```

narwhals/translate.py
```diff
@@ -665,10 +665,15 @@ def _from_native_impl(  # noqa: PLR0915
...
             raise TypeError(msg)
         if eager_only or eager_or_interchange_only:
-            msg = f"Cannot only use `eager_only` or `eager_or_interchange_only` with {ns_spark.implementation._alias} DataFrame"
+            msg = (
+                "Cannot only use `eager_only` or `eager_or_interchange_only` with "
+                f"{ns_spark.implementation} DataFrame"
+            )
```
dangotbanned, line 676:
```suggestion
            msg = f"Cannot only use `eager_only` or `eager_or_interchange_only` with {ns_spark.implementation} DataFrame"
```

narwhals/_enum.py
```diff
@@ -0,0 +1,48 @@
...
+        ValueError: Creating values with `auto()` is not allowed. Please provide a value manually instead.
+
+    Raises:
+        TypeError: If `auto()` is attempted to be used for any enum member value.
+
+    Note:
+        All enum values must be explicitly assigned. This is useful when you want
+        to maintain full control over enum values and prevent automatic numbering.
+
+    """
```
dangotbanned, line 38: It seems like what this `Enum` does is repeated a few too many times to me:

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L7-L8

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L9

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L11-L12

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L29

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L32

https://github.com/narwhals-dev/narwhals/blob/59614e352c03f439f7bdee83d3327d85f9fe973f/narwhals/_enum.py#L35-L36

Do we need *all* of these?
I like the examples and think they explain everything clearly 🎉

dangotbanned (commented):
I think we're pretty much ready to 🚢  @FBruzzesi

narwhals/_enum.py
FBruzzesi: I guess I was a bit _too_ verbose (and also wrong in the error mentioned) - I am from mobile and will address later tonight

narwhals/utils.py
FBruzzesi: I also don't have a strong opinion - maybe `UNKNOWN_EXTENSION = "unknown extension" is the most verbose but also more descriptive?!

narwhals/_enum.py
FBruzzesi: Addressed in https://github.com/narwhals-dev/narwhals/pull/2432/commits/51c0e3ae253d10aea3815ffc4e7bf0ed0248cb50

dangotbanned (approved):
Thanks @FBruzzesi, LGTM ❤️ 

(https://github.com/narwhals-dev/narwhals/pull/2432#discussion_r2068863654) isn't blocking for me - but maybe worth seeing if @EdAbati or @MarcoGorelli have a preference?

FBruzzesi:
@dangotbanned I know [this point is still open](https://github.com/narwhals-dev/narwhals/pull/2432?notification_referrer_id=NT_kwDOAo1WGLQxNjA3MDMyMzMxNTo0MjgxNzA0OA#discussion_r2068863654), but shall we merge for now and eventually come back to it if needed?

dangotbanned:
> @dangotbanned I know [this point is still open](https://github.com/narwhals-dev/narwhals/pull/2432?notification_referrer_id=NT_kwDOAo1WGLQxNjA3MDMyMzMxNTo0MjgxNzA0OA#discussion_r2068863654), but shall we merge for now and eventually come back to it if needed?

Yeah go for it @FBruzzesi - thanks for working on this!

PR #2440, opened 2025-04-27
tests: un-xfail polars group by expr streaming

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

Polars made a release 🔥

### DIFF
--- a/tests/group_by_test.py
+++ b/tests/group_by_test.py
@@ -501,26 +501,14 @@ def test_fancy_functions(constructor: Constructor) -> None:
             ["y"],
         ),
     ],
-    ids=range(6),
 )
 def test_group_by_expr(
-    request: pytest.FixtureRequest,
     constructor: Constructor,
     keys: list[nw.Expr],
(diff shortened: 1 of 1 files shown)


dangotbanned (approved):
Nice one @FBruzzesi 🎉 

Just wanna double check - do we definitely not need this gated for earlier versions?

FBruzzesi:
> Nice one @FBruzzesi 🎉 
> 
> Just wanna double check - do we definitely not need this gated for earlier versions?

In CI we don't really pin versions (unless for specific jobs), so the latest should always be used. This is even more true for the streaming engine case

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2440#issuecomment-2833614167

Sounds good to me, thanks @FBruzzesi

Issue #2248, opened 2025-03-19
Consistently prefix `(_?)with_` for "this, but with `x` changed" 


dangotbanned:
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/2232#discussion_r2003027883_

> > this is fine - I guess we could call it `_with_version` for consistency with how we're now naming similar methods?
> 
> @MarcoGorelli Yeah I like that
> I think we should open an issue to identify all the places we could change to that style.
>
> I had a similar thought here - but reverted to keep the PR from spiraling:
> - [`9e0611d` (#2236)](https://github.com/narwhals-dev/narwhals/pull/2236/commits/9e0611d4840a906321153bddbcddc1fc1644729d)

## Candidates
> [!IMPORTANT]
> - Just for internal API
> - `from_` becomes reserved for `@classmethod`(s)

| Current                                                                 | Proposed                                           |
| ----------------------------------------------------------------------- | -------------------------------------------------- |
| `Compliant*._change_version()`                                          | `Compliant*._with_version()`                       |
| `Compliant(Data\|Lazy)Frame._from_native_frame()`                       | `Compliant*._with_native()`                        |
| `CompliantSeries._from_native_series()`                                 | Same as above                                      |
| `(Dask\|DuckDB\|SparkLike)Expr._from_call()`                            | `LazyExpr._with_callable()`                        |
| `Expr._from_callable()`                                                 | `Expr._with_callable()`                            |
| `(Dask\|DuckDB\|SparkLike)ExprNameNamespace._from_alias_output_names()` | `LazyExprNameNamespace._with_alias_output_names()` |

## Related
- [ ] #2262

dangotbanned:
@MarcoGorelli I think I've found all the *maybe withs* 😉

PR #2442, opened 2025-04-28
chore: Parsing functions -> enum constructors

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Similar #2248

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Same thinking as (#2248), but for some functions that *might not* immediately appear to be alternative constructors.

Now on usage, it is very clear exactly the type we get when called.
Should make discovery a little easier as well

### DIFF
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -403,7 +403,6 @@ def combine_metadata(  # noqa: PLR0915
                         result_expansion_kind = expansion_kind
                     else:
                         result_expansion_kind = result_expansion_kind & expansion_kind
-
             kind = metadata.kind
             if kind is ExprKind.AGGREGATION:
                 has_aggregations = True


narwhals/_expression_parsing.py
```diff
@@ -198,6 +198,9 @@ class ExpansionKind(Enum):
     def is_multi_unnamed(self) -> bool:
         return self is ExpansionKind.MULTI_UNNAMED
 
+    def is_multi_output(self) -> bool:
+        return self in {ExpansionKind.MULTI_NAMED, ExpansionKind.MULTI_UNNAMED}
+
```
dangotbanned: This one wasn't a constructor, just noticed while making the other changes.

Having it as a `TypeIs` would make sense if we were using it for narrowing.
However, we weren't, so it was just more verbose and less ergonomic

Issue #2450, opened 2025-04-29
[Bug]: `AttributeError` when using narwhalified Polars DataFrame with Joblib Parallel Processing


dennisbader:
### Description

I encountered an error when using the narwhals package with Polars and Joblib for parallel processing. The error occurs when attempting to convert a Polars DataFrame using nw.from_native(df) within a parallelized function.

#### Background info: 

We encountered this issue when trying to implement narwhals in another Darts method in [this PR](https://github.com/unit8co/darts/pull/2766).

### Steps or code to reproduce the bug

Here is the code to reproduce the error:

```py
from joblib import Parallel, delayed

import pandas as pd
import polars as pl
import narwhals as nw


def do_something(df):
    # parallel does not work with narwhalified polars
    return df

dfs_pandas = [nw.from_native(pd.DataFrame({"col1": [0, 2], "col2": [3, 7]}))]
dfs_polars = [nw.from_native(pl.DataFrame({"col1": [0, 2], "col2": [3, 7]}))]

# with pandas it works
Parallel(n_jobs=-1)(delayed(do_something)(df_) for df_ in dfs_pandas)

# with polars it does not work
Parallel(n_jobs=-1)(delayed(do_something)(df_) for df_ in dfs_polars)
```

### Expected results

Parallel processing should work as in the case with narwhalified pandas DataFrame.

### Actual results

Parallel processing does not work with narwhalified polars DataFrame.

### Please run narwhals.show_version() and enter the output below.

```shell
System:
    python: 3.11.11 (main, Dec 11 2024, 10:25:04) [Clang 14.0.6 ]
executable: /Users/username/miniconda3/envs/darts/bin/python
   machine: macOS-15.3.2-arm64-arm-64bit
Python dependencies:
     narwhals: 1.37.0
       pandas: 2.2.3
       polars: 1.27.1
         cudf: 
        modin: 
      pyarrow: 19.0.1
        numpy: 2.2.5
```

### Relevant log output

```shell
/Users/username/miniconda3/envs/darts/bin/python /Users/username/projects/unit8/darts_test_code/bug_fixes/bug_polars_parallel.py 
joblib.externals.loky.process_executor._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/Users/username/miniconda3/envs/darts/lib/python3.11/site-packages/narwhals/_polars/dataframe.py", line 227, in func
    return self._from_native_object(getattr(self.native, attr)(*pos, **kwds))
[...]

PR #2451, opened 2025-04-29
fix: Ensure Polars wrapped in Narwhals can be passed to Joblib

MarcoGorelli:
closes https://github.com/narwhals-dev/narwhals/issues/2450

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_polars/dataframe.py
```diff
@@ -70,6 +70,37 @@
...
+        "tail",
+        "to_arrow",
+        "to_pandas",
+        "unique",
+        "with_columns",
+        "write_csv",
+        "write_parquet",
+    ]
+)
+
```
dangotbanned, line 103: Thanks @MarcoGorelli, I've got two suggestions.

1. Wouldn't we also need this for all the other `Polars*` classes that use `__getattr__` as a proxy?
I'm thinking a search for `: Method[` would highlight which ones - but also the `(Series|Expr)*Namespace` classes
2. If we're defining all the cases up front, have you thought about using a descriptor instead?
`__getattr__` makes sense at the moment because we're *technically* allowing things *outside* of the `narwhals` API to pass through. In practice, I understand that might not be possible from the user-side - but it could mean we can just rely on `python` to raise the `AttributeError` instead of checking and raising ourselves 🤔

MarcoGorelli: Thanks @dangotbanned !

1. I don't think is a concern, as users work with `DataFrame` / `LazyFrame` / `Series` structures. Nobody would make a function which accepts `series.dt` as an input :) 
2. I'm not familiar with descriptors, but open to suggestions

dangotbanned: > Thanks @dangotbanned !
> 
> 1. I don't think is a concern, as users work with `DataFrame` / `LazyFrame` / `Series` structures. Nobody would make a function which accepts `series.dt` as an input :)

Sounds good 👍 

> 2. I'm not familiar with descriptors, but open to suggestions

If you're looking to include this PR in a bugfix - we can punt this to another issue?

Some useful homework though 😉 

- [datamodel>Class instances](https://docs.python.org/3/reference/datamodel.html#id4)
- [`object.__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__)
- [datamodel>descriptors](https://docs.python.org/3/reference/datamodel.html#descriptors)
- [`object.__dict__`](https://docs.python.org/3/reference/datamodel.html#object.__dict__)

In short, I think we can reduce the performance overhead of **always** looking up (and failing) in order to reach `__getattr__` - since you've defined the full set of methods that are permitted.
So the idea would be to move a cost that is currently paid *per-instance* and *per-method* into one that is *per-type* for the fixed set of methods 😏 

Pinging @camriddell to check if what I'm saying seems legit 😄

MarcoGorelli: sure, thanks!

camriddell: After looking at this might I propose the use of a decorator to dynamically add these methods to the class and then do away with the dynamic look up/attribute fetching via `__getattr__` altogether?

The pattern here typically looks like

```python
def define_attributes(attributes):
    def decorator(cls):
        for attr in attributes:
            setattr(cls, attr, ...) # insert dynamically defined function here
        return cls
    return decorator
         
@define_attributes(['min', 'max'])
class T:
    pass

t = T()
print(f'{t.min = }') # t.min = Ellipsis
```

Whereas the descriptor pattern may look like:

```python
class PassThrough:
    def __set_name__(self, owner, name):
        self._name = name
        return name

    def __get__(self, instance, owner):
        # func = getattr(..., self._name) # get the underlying function

        if instance is None:
            return ... # special return for class look ups
        return ...     # return for instance look ups


class T:
    min = PassThrough()

t = T()
print(f'{t.min = }')
```

The main tradeoffs between these two
- The decorator pattern is fairly declarative (passing attribute names as a `list[str]`) and the logic may be easier to follow
- The descriptor pattern is very declarative; but repetitive and a big "magic"

The nice thing about either approach is that is enhances discoverability (finding available attribute via `dir` [without overriding `__dir__`]).

---

@dangotbanned, as you pointed out `__getattr__` is invoked as a final effort when `__getattribute__` fails noting that `__getattribute__` will check in the instance `__dict__`, then the class `__dict__`, then the rest of the MRO class `__dict__`s (this was off the top of my head, so may be slightly inaccurate).
Though I am not concerned about speed here, the difference between calling through `__getattr__` vs not will likely be a couple hundred nanoseconds? This is negligible compared to the computation time.

If performance is a consideration, we could probably use the defined method strings as a whitelist in `__getattribute__` (as opposed to a blacklist in `__getattr__`).

```python
class T:
    def __getattribute__(self, attribute):
        if attribute in INHERITED_METHODS:
            return ... # some dynamic func
        return object.__getattribute__(self, attribute) # pass the rest along
```

dangotbanned: Thanks for accepting the summons and giving such a detailed overview @camriddell 😄 

For a language with a [guiding principle](https://peps.python.org/pep-0020/) of

> There should be one-- and preferably only one --obvious way to do it.

We certainly have a few options!

## Performance

I'm planning to spin this off into an issue, as I'd like to figure out a good set of benchmarks *first* and then try out the various solutions.

> Though I am not concerned about speed here, the difference between calling through __getattr__ vs not will likely be a couple hundred nanoseconds? This is negligible compared to the computation time.

I agree that the overhead for a **single call** is *likely* negligible. I'm more concerned with it being overhead that is added in **so much** of the core API.
If we take `PolarsDataFrame` as an example. It has 23 methods that currently hit the `__getattr__` path.

I'd argue that *at least* **1** of these **5** (*particularly the first **3***) methods are *likely* to appear in most queries.
Whatever cost there is (if any) can very quickly add up when you add method chaining into the mix

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L110

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L120

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L125

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L121

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L116

> [!NOTE]
> To be clear, I'm saying the overhead *seems* to be an unknown that I'd like to quantify.
> Just that it seems reasonable *to me* that there *could* be overhead

## Decorator vs Descriptor
> The main tradeoffs between these two
> 
> - The decorator pattern is fairly declarative (passing attribute names as a list[str]) and the logic may be easier to follow
> - The descriptor pattern is very declarative; but repetitive and a big "magic"
> 
> The nice thing about either approach is that is enhances discoverability (finding available attribute via dir [without overriding __dir__]).

**Fun fact**: in `altair` we use a decorator to add multiple descriptors!
Although I wouldn't recommend it here 😅 

<details><summary><code>altair</code> property setter methods</summary>
<p>

This all pre-dates when I started working on the project, but I did manage to simplify it somewhat recently in (https://github.com/vega/altair/pull/3659)

https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/altair/vegalite/v5/schema/channels.py#L269-L271

https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/altair/vegalite/v5/schema/channels.py#L493-L749

https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/altair/utils/schemapi.py#L1677-L1682

https://github.com/vega/altair/blob/ea3a6e206b31407714e3619458b3cb9d9c4d7e92/altair/utils/schemapi.py#L1628-L1674

</p>
</details> 


The reason I was so quick to jump to using a descriptor for this case though was that *for typing*, we already need to add a lot of repetitive boilerplate like this:

<details><summary>Typing boilerplate</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L67-L71

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L104-L129

</p>
</details> 


We'd still need that if we used a decorator, plus the new set defining the names

<details><summary><code>INHERITED_METHODS</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/b1c7e8e18a18cc542f2aac73ce264f184c9cba3f/narwhals/_polars/dataframe.py#L73-L101

</p>
</details> 

But if we were to use a ([generic](https://docs.python.org/3/reference/datamodel.html#emulating-generic-types)) descriptor, I think we'd be able to write something like this which does the job of both 🤯:

```py
class PolarsDataFrame:
    clone = Proxy["Self"]()
    collect = Proxy["CompliantDataFrame[Any, Any, Any]"]()
    drop_nulls = Proxy["Self"]()
    estimated_size = Proxy["int | float"]()
    explode = Proxy["Self"]()
    filter = Proxy["Self"]()
```

> [!NOTE]
> Using [string forward references](https://typing.python.org/en/latest/spec/annotations.html#string-annotations) here, since they were previously not in a runtime scope

Issue #2264, opened 2025-03-21
[Enh]: Add functionality to convert from python to narwhals dtype


FBruzzesi:
Polars has [parse_py_type_into_dtype](https://github.com/pola-rs/polars/blob/d95e343e04d47bcbeafc815a2d9855aa5099f1bb/py-polars/polars/datatypes/_parse.py#L72). It could be handy to have the same

dangotbanned:
I'd thought about this before, but figured it might've been out of scope.

It could be nice to have these:

```py
# NOTE: Implemented on each subclass
nw.DType.from_python
nw.DType.to_python

nw.Schema.from_python
nw.Schema.to_python
```

- https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.DataType.html
- https://github.com/pola-rs/polars/blob/py-1.25.2/py-polars/polars/schema.py#L202-L218

dangotbanned:
@FBruzzesi I forgot to mention, I've got part of the `from_python` side in (#2572) for inferring the `DType` in `lit(...)` like `polars` does:

## Impl

https://github.com/narwhals-dev/narwhals/blob/01f4d26346f3046db5d34c330b372356540ab9da/narwhals/_plan/common.py#L323-L338

## Usage

https://github.com/narwhals-dev/narwhals/blob/01f4d26346f3046db5d34c330b372356540ab9da/narwhals/_plan/common.py#L281-L294

https://github.com/narwhals-dev/narwhals/blob/01f4d26346f3046db5d34c330b372356540ab9da/narwhals/_plan/demo.py#L56-L66

FBruzzesi:
Thanks @dangotbanned - would it be worth it to start to prioritize this issue and work on it separately of #2572?

dangotbanned:
> Thanks [@dangotbanned](https://github.com/dangotbanned) - would it be worth it to start to prioritize this issue and work on it separately of [#2572](https://github.com/narwhals-dev/narwhals/pull/2572)?

Yeah @FBruzzesi definitely!

I would recommend landing (#2626) **first** though, since we'd be able to expand the definition of `IntoDType` to include *some* python types as well.
[...]

PR #2460, opened 2025-04-30
chore: Adopting more `ruff` complexity rules

### LINKED ISSUE #2376: [internal]: Adopting/following more `ruff` complexity rules
## Related
- https://github.com/narwhals-dev/narwhals/issues/1942#issuecomment-2763647131
- https://discord.com/channels/1235257048170762310/1272835922924273694/1355268679083753594
- https://github.com/narwhals-dev/narwhals/commit/4f02f06c3c17b10a79c539113c00b1a8c35c5a95

## Description
All of the rules I'm linking have clear examples demonstrating why you'd use them.
The general theme they share is producing *smaller* and more *generic* functions/methods/classes.

The **bad news** is these aren't auto-fixes, but intended as a warning when you've written something that *may* be difficult for others to follow.

The **good news** is we can gradually introduce them via [`per-file-ignores`](https://docs.astral.sh/ruff/settings/#lint_per-file-ignores).
See [altair/pyproject.toml](https://github.com/vega/altair/blob/a6d0b7d5720cb414f3a3b61e58cdbe2bcbadcce4/pyproject.toml#L249-L250) for an example of enabling a rule group for a single module.

Adopt the following **ignored** rules:
- https://docs.astral.sh/ruff/rules/complex-structure/
- https://docs.astral.sh/ruff/rules/too-many-return-statements/
- https://docs.astral.sh/ruff/rules/too-many-branches/
- ~~https://docs.astral.sh/ruff/rules/superfluous-else-return/~~
  - Doesn't play nicely with coverage

*Possibly* adopt these **preview** rules:
- https://docs.astral.sh/ruff/rules/too-many-locals/
- https://docs.astral.sh/ruff/rules/too-many-boolean-expressions/
- https://docs.astral.sh/ruff/rules/too-many-nested-blocks/

More strictly **follow** rules we *frequently* `# noqa: ...`
- https://docs.astral.sh/ruff/rules/too-many-statements/

dangotbanned:
Will close #2376

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [X] 🐳 Other

## Related issues

- Closes #2376

## Checklist

- [ ] Tests added
- [x] Documented the changes

Would need some buy-in from others that they'd like to see this stuff simplified/broken up 🙂


FBruzzesi:
Thanks @dangotbanned - I am a bit perplexed about this one as of now, mostly due to

> Currently just highlighting areas that need some love

This does not mean I am against it, but as of now I don't see the benefits nor what it would imply to adopt them

> Would need some buy-in from others

Could you maybe show how _one_ of those `noqa` would change to satisfy these rules?

According to such example it might be easier for everyone to evaluate how the future might look like 😁

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2460#issuecomment-2845685986

Sure thing @FBruzzesi I'll take a stab at one tomorrow!
The general theme would be writing shorter, more reusable and easier to understand functions.

The aim of the limits set by the rules are to *nudge* you when something you're writing is *starting* to get difficult for others to understand.
You don't *necessarily* need to follow them all the time, but I've found it keeps me on my toes - always considering *"will this make sense to someone else, without writing a comment?"*

narwhals/_arrow/utils.py
```diff
@@ -154,7 +154,7 @@ def native_to_narwhals_dtype(dtype: pa.DataType, version: Version) -> DType:
     return dtypes.Unknown()  # pragma: no cover
 
 
-def narwhals_to_native_dtype(dtype: DType | type[DType], version: Version) -> pa.DataType:
+def narwhals_to_native_dtype(dtype: DType | type[DType], version: Version) -> pa.DataType:  # noqa: C901, PLR0912
```
dangotbanned, line 157: For all of the `narwhals_to_native_dtype` functions, I'd recommend the same solution.
It is a bit much (to implement) for this PR, but hopefully it gives an example of where these kinds of changes can lead 🙂 

## General
We could *instead* have methods on each `DType` class like:

```py
to_arrow
to_dask
to_duckdb
to_polars
# NOTE: Could be more granular for these
to_pandas_like
to_spark_like
```

- Each would be much shorter (the goal of the rules) and very easy to digest.
- They'd be **more performant** as we'd skip all of the `isinstance` checks 
  - By just calling the same method for every single case
- We could also make these public very easily
  - I imagine they'd be helpful in the same way as (#1912) 😉 
- Unsupported conversions are handled by the abscence of methods
  - We wouldn't need to write this as python raises for us
  - https://github.com/narwhals-dev/narwhals/blob/162dd5fbf9878155ed64d16043040f6b075f6101/narwhals/_polars/utils.py#L198-L200
  - `AttributeError: 'Decimal' object has no attribute 'to_arrow'`

## Impl
We would want to borrow/adapt the [`classinstmethod`](https://github.com/pola-rs/polars/blob/479b9d24da685fecad3d5682b7bc82ffe0f52a45/py-polars/polars/datatypes/classes.py#L32-L37) descriptor from `polars`.
They use that for [`to_python`](https://github.com/pola-rs/polars/blob/479b9d24da685fecad3d5682b7bc82ffe0f52a45/py-polars/polars/datatypes/classes.py#L230-L246) which is the same kind of operation - and is something we might want in the future anyway (#2264) 😏 

## `V1`
Another big benefit of this is that we'd move the `MAIN` vs `V1` logic into the `V1` data types.
I think this would make it really clear for us what the behavior of `Enum` should be:

https://github.com/narwhals-dev/narwhals/blob/162dd5fbf9878155ed64d16043040f6b075f6101/narwhals/_polars/utils.py#L184-L191

narwhals/_pandas_like/utils.py
```diff
@@ -213,7 +213,7 @@ def rename(
 
 
 @functools.lru_cache(maxsize=16)
-def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> DType:
+def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> DType:  # noqa: C901, PLR0912
```
dangotbanned, line 216: This could benefit from something like the example in ([`PLR0911`](https://docs.astral.sh/ruff/rules/too-many-return-statements/#too-many-return-statements-plr0911))

<details><summary><code>PANDAS_LIKE_TO_NW_PRIMITIVE</code></summary>
<p>


```py
PANDAS_LIKE_TO_NW_PRIMITIVE = {
    "Int64": "Int64",
    "Int64[pyarrow]": "Int64",
    "int64": "Int64",
    "int64[pyarrow]": "Int64",
    "Int32": "Int32",
    "Int32[pyarrow]": "Int32",
    "int32": "Int32",
    "int32[pyarrow]": "Int32",
    "Int16": "Int16",
    "Int16[pyarrow]": "Int16",
    "int16": "Int16",
    "int16[pyarrow]": "Int16",
    "Int8": "Int8",
    "Int8[pyarrow]": "Int8",
    "int8": "Int8",
    "int8[pyarrow]": "Int8",
    "UInt64": "UInt64",
    "UInt64[pyarrow]": "UInt64",
    "uint64": "UInt64",
    "uint64[pyarrow]": "UInt64",
    "UInt32": "UInt32",
    "UInt32[pyarrow]": "UInt32",
    "uint32": "UInt32",
    "uint32[pyarrow]": "UInt32",
    "UInt16": "UInt16",
    "UInt16[pyarrow]": "UInt16",
    "uint16": "UInt16",
    "uint16[pyarrow]": "UInt16",
    "UInt8": "UInt8",
    "UInt8[pyarrow]": "UInt8",
    "uint8": "UInt8",
    "uint8[pyarrow]": "UInt8",
    "Float64": "Float64",
    "Float64[pyarrow]": "Float64",
    "double[pyarrow]": "Float64",
    "float64": "Float64",
    "float64[pyarrow]": "Float64",
    "Float32": "Float32",
    "Float32[pyarrow]": "Float32",
    "float32": "Float32",
    "float32[pyarrow]": "Float32",
    "float[pyarrow]": "Float32",
    "large_string[pyarrow]": "String",
    "string": "String",
    "string[pyarrow]": "String",
    "string[python]": "String",
    "bool": "Boolean",
    "bool[pyarrow]": "Boolean",
    "boolean": "Boolean",
    "boolean[pyarrow]": "Boolean",
}
```

</p>
</details> 

We can reduce the first 12 branches into a single case

```py
def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> DType:  # noqa: C901, PLR0912
    dtype = str(native_dtype)

    dtypes = version.dtypes
    if normalized := PANDAS_LIKE_TO_NW_PRIMITIVE.get(dtype):
        return getattr(dtypes, normalized)()
    elif ...
```

narwhals/_arrow/series_dt.py
```diff
@@ -44,7 +44,7 @@ def convert_time_zone(self, time_zone: str) -> ArrowSeries:
         ser = self.replace_time_zone("UTC") if self.time_zone is None else self.compliant
         return self.with_native(ser.native.cast(pa.timestamp(self.unit, time_zone)))
 
-    def timestamp(self, time_unit: TimeUnit) -> ArrowSeries:
+    def timestamp(self, time_unit: TimeUnit) -> ArrowSeries:  # noqa: C901, PLR0912
```
dangotbanned, line 47: > [!NOTE]
> TODO 

Okay I'll do this one in this PR - have been wanting to refactor it for a while anyway

dangotbanned: @FBruzzesi first worked example in (https://github.com/narwhals-dev/narwhals/pull/2460/commits/6666fe23490ada4f745f0d57296698d658100115)

I *could have* reduced the LOC as well, but I think the typing helps reduce the amount of info you need to keep in your head 🙂

narwhals/_arrow/utils.py
FBruzzesi: Those are some great observations 😎

> It is a bit much (to implement) for this PR

Yes I agree 👌🏼

> We could instead have methods on each DType class like:

All the points and benefits here are very nice indeed - my only disagreement would be on:

> Unsupported conversions are handled by the abscence of methods

which I would rather raise a NotImplemented or NotSupported error on a case by case basis. Would that be even simpler to type check 😉?

narwhals/_arrow/series_dt.py
FBruzzesi, line 1: I am already sold from this diff 👌🏼😉

narwhals/_arrow/utils.py
dangotbanned: > which I would rather raise a NotImplemented or NotSupported error on a case by case basis. Would that be even simpler to type check 😉?

Yeah @FBruzzesi that sounds like a better option actually nice 🙂

narwhals/_arrow/series_dt.py
FBruzzesi: Yep that's great (as commented [here](https://github.com/narwhals-dev/narwhals/pull/2460#discussion_r2072454774)).

I would propose to:

- Wrap up this PR as is (or any minor adjustment if needed)
- Keep #2376 as tracker for the functionalities that had to add `#noqa: <ANY NEW RULE>` or open a new issue for it so that we can progressively tackle them

dangotbanned: Thanks @FBruzzesi - I've finished moving all of them into child issues of 

- #2485

Issue #2438, opened 2025-04-27
enh: simple `dt.truncate`


MarcoGorelli:
I think it would be amazing to support some version of truncate

In Polars it's fairly complex because it allows for mixing and matching its bespoke string language

In practice, i'd conjecture that 99% of `dt.truncate` cases involve a single duration (e.g. `truncate('1d')`) and that `truncate('1d2ms3h')` is exceedingly rare

I think we should allow for simple truncation, where we accept a string `every` which must be of the form: `{number}{interval}`

Initially at least it could be: only strings, no timedelta, no expressions. Keep it simple initially, we can add complexity later

Pretty keen on having this, if anyone wants to submit a PR I'd like to merge it quite quickly

So, examples of what we accept:
- `dt.truncate('1d')`
- `dt.truncate('2d')`
- `dt.truncate('1mo')`

The `number` part must be strictly positive

The interval part can be one of:

        - 1ns   (1 nanosecond)
        - 1us   (1 microsecond)
        - 1ms   (1 millisecond)
        - 1s    (1 second)
        - 1m    (1 minute)
        - 1h    (1 hour)
        - 1d    (1 calendar day)
        - 1w    (1 calendar week) ***
        - 1mo   (1 calendar month)
        - 1q    (1 calendar quarter)
        - 1y    (1 calendar year)

***: 'week' is arguably fairly non-trivial. No objections to keeping this one out of an initial implementation and doing it as a follow-up

PR #2465, opened 2025-05-01
feat: add `dt.truncate`

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2438
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

## Comments

- The intervals larger than "d" are not implemented in pandas/dask. It seems that `dt.to_period()` can be used for them but it seems to be buggy so I didn't use it. Maybe there is a better way.
- Unit of 'w' is not implemented as there seems to be differences in how different backends calculate them.
- sqlframe seems to have some issue with adding a timezone, reported this in https://github.com/eakmanrq/sqlframe/issues/383 so hopefully after it is fixed, we can just remove the xfail.
- There seems to be different origin dates used by different backends so there is a condition on multiples used for 'mo', 'q', and 'y' (investigated by @MarcoGorelli ).

### DIFF
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -10,6 +10,8 @@ import pyarrow.compute as pc
 from narwhals._arrow.utils import ArrowSeriesNamespace
 from narwhals._arrow.utils import floordiv_compat
 from narwhals._arrow.utils import lit
+from narwhals.utils import parse_interval_string
+from narwhals.utils import unit_to_str
 
 if TYPE_CHECKING:
     from narwhals._arrow.series import ArrowSeries
@@ -189,3 +191,10 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
         }
         factor = lit(unit_to_nano_factor[self.unit], type=pa.int64())
         return self.with_native(pc.multiply(self.native, factor).cast(pa.int64()))
+
+    def truncate(self, every: str) -> ArrowSeries:
+        number, unit = parse_interval_string(every)
+        unit_dict = unit_to_str()
+        return self.with_native(
+            pc.floor_temporal(self.native, multiple=int(number), unit=unit_dict[unit])  # type: ignore[call-overload]
+        )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -7,11 +7,11 @@ from typing import cast
 import pyarrow as pa
 import pyarrow.compute as pc
 
+from narwhals._arrow.utils import UNITS_DICT
 from narwhals._arrow.utils import ArrowSeriesNamespace
 from narwhals._arrow.utils import floordiv_compat
 from narwhals._arrow.utils import lit
 from narwhals.utils import parse_interval_string
-from narwhals.utils import unit_to_str
 
 if TYPE_CHECKING:
     from narwhals._arrow.series import ArrowSeries
@@ -194,10 +194,9 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
 
     def truncate(self, every: str) -> ArrowSeries:
         number, unit = parse_interval_string(every)
-        unit_dict = unit_to_str()
         if unit == "q" and number != "1":
             msg = f"Only 1{unit} is currently supported for PyArrow."
             raise NotImplementedError(msg)
         return self.with_native(
-            pc.floor_temporal(self.native, multiple=int(number), unit=unit_dict[unit])  # type: ignore[call-overload]
+            pc.floor_temporal(self.native, multiple=int(number), unit=UNITS_DICT[unit])  # type: ignore[call-overload]
         )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -195,6 +195,9 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
     def truncate(self, every: str) -> ArrowSeries:
         number, unit = parse_interval_string(every)
         unit_dict = unit_to_str()
+        if unit == "q" and number != "1":
+            msg = f"Only 1{unit} is currently supported for PyArrow."
+            raise NotImplementedError(msg)
         return self.with_native(
             pc.floor_temporal(self.native, multiple=int(number), unit=unit_dict[unit])  # type: ignore[call-overload]
         )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -201,5 +201,5 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
             )
             raise NotImplementedError(msg)
         return self.with_native(
-            pc.floor_temporal(self.native, multiple=multiple, unit=UNITS_DICT[unit])  # type: ignore[call-overload]
+            pc.floor_temporal(self.native, multiple=multiple, unit=UNITS_DICT[unit])
         )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -194,12 +194,6 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
 
     def truncate(self, every: str) -> ArrowSeries:
         multiple, unit = parse_interval_string(every)
-        if unit == "q" and multiple != 1:
-            msg = (
-                "For 'q' unit only multiple 1 is currently supported for PyArrow.\n"
-                f"Got multiple {multiple!s}."
-            )
-            raise NotImplementedError(msg)
         return self.with_native(
             pc.floor_temporal(self.native, multiple=multiple, unit=UNITS_DICT[unit])
         )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -193,10 +193,13 @@ class ArrowSeriesDateTimeNamespace(ArrowSeriesNamespace):
         return self.with_native(pc.multiply(self.native, factor).cast(pa.int64()))
 
     def truncate(self, every: str) -> ArrowSeries:
-        number, unit = parse_interval_string(every)
-        if unit == "q" and number != "1":
-            msg = f"Only 1{unit} is currently supported for PyArrow."
+        multiple, unit = parse_interval_string(every)
+        if unit == "q" and multiple != 1:
+            msg = (
+                "For 'q' unit only multiple 1 is currently supported for PyArrow.\n"
+                f"Got multiple {multiple!s}."
+            )
             raise NotImplementedError(msg)
         return self.with_native(
-            pc.floor_temporal(self.native, multiple=int(number), unit=UNITS_DICT[unit])  # type: ignore[call-overload]
+            pc.floor_temporal(self.native, multiple=multiple, unit=UNITS_DICT[unit])  # type: ignore[call-overload]
         )
--- a/narwhals/_arrow/series_dt.py
+++ b/narwhals/_arrow/series_dt.py
@@ -11,7 +11,7 @@ from narwhals._arrow.utils import UNITS_DICT
 from narwhals._arrow.utils import ArrowSeriesNamespace
 from narwhals._arrow.utils import floordiv_compat
 from narwhals._arrow.utils import lit
-from narwhals.utils import parse_interval_string
+from narwhals._duration import parse_interval_string
 
 if TYPE_CHECKING:
     from narwhals._arrow.series import ArrowSeries
--- a/narwhals/_dask/expr_dt.py
+++ b/narwhals/_dask/expr_dt.py
@@ -2,11 +2,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from narwhals._duration import parse_interval_string
 from narwhals._pandas_like.utils import calculate_timestamp_date
 from narwhals._pandas_like.utils import calculate_timestamp_datetime
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals.utils import Implementation
-from narwhals.utils import parse_interval_string
 
 if TYPE_CHECKING:
     import dask.dataframe.dask_expr as dx
--- a/narwhals/_dask/expr_dt.py
+++ b/narwhals/_dask/expr_dt.py
@@ -6,6 +6,7 @@ from narwhals._pandas_like.utils import calculate_timestamp_date
 from narwhals._pandas_like.utils import calculate_timestamp_datetime
 from narwhals._pandas_like.utils import native_to_narwhals_dtype
 from narwhals.utils import Implementation
+from narwhals.utils import parse_interval_string
 
 if TYPE_CHECKING:
     import dask.dataframe.dask_expr as dx
@@ -156,4 +157,13 @@ class DaskExprDateTimeNamespace:
         return self._compliant_expr._with_callable(
             lambda _input: _input.dt.total_seconds() * 1_000_000_000 // 1,
             "total_nanoseconds",
+        )
+
+    def truncate(self, every: str) -> DaskExpr:
+        _, unit = parse_interval_string(every)
+        if unit == "m":  # replace "m" unit with "min"
+            every += "in"
+        return self._compliant_expr._with_callable(
+            lambda _input: _input.dt.floor(every),
+            "truncate",
         )
--- a/narwhals/_dask/expr_dt.py
+++ b/narwhals/_dask/expr_dt.py
@@ -164,8 +164,8 @@ class DaskExprDateTimeNamespace:
         if unit in {"mo", "q", "y"}:
             msg = f"Truncating to {unit} is not supported yet for dask."
             raise NotImplementedError(msg)
-        if unit == "m":  # replace "m" unit with "min"
-            every += "in"
+        if unit == "m":
+            every = every.replace("m", "min", count=1)
         return self._compliant_expr._with_callable(
             lambda _input: _input.dt.floor(every),
             "truncate",
--- a/narwhals/_dask/expr_dt.py
+++ b/narwhals/_dask/expr_dt.py
@@ -161,6 +161,9 @@ class DaskExprDateTimeNamespace:
 
     def truncate(self, every: str) -> DaskExpr:
         _, unit = parse_interval_string(every)
+        if unit in {"mo", "q", "y"}:
+            msg = f"Truncating to {unit} is not supported yet for dask."
+            raise NotImplementedError(msg)
         if unit == "m":  # replace "m" unit with "min"
             every += "in"
         return self._compliant_expr._with_callable(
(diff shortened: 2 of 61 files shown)


narwhals/utils.py
```diff
@@ -1964,3 +1964,36 @@ def to_tuple(self) -> tuple[_T, ...]:
...
+        A tuple of number and unit parsed from the interval string.
+    """
+    format = r"(\d+)([a-zA-Z]+)"
+    interval = re.match(format, every)
+    if interval is None:
+        msg = f"Invalid frequency string: {every}."
+        raise ValueError(msg)
+    number = interval.group(1)
+    unit = interval.group(2)
+    return number, unit
```
dangotbanned, line 1982: @raisadz I think this could remove the need for validating the unit elsewhere and might be a bit faster with [`re.compile`](https://docs.python.org/3/library/re.html#re.compile)

```suggestion

# You could remove the ones that aren't supported yet/not planned
_TIME_UNIT = "ns", "us", "ms", "mo", "m", "s", "h", "d", "w", "q", "y"
_PATTERN_INTERVAL = re.compile(rf"^(?P<number>\d+)(?P<unit>{'|'.join(_TIME_UNIT)})$")


def parse_interval_string(every: str) -> tuple[str, str]:
    """Parse a string like "1d", "2h", "3m" into a tuple of (number, unit).

    Returns:
        A tuple of number and unit parsed from the interval string.
    """
    if match := _PATTERN_INTERVAL.match(every):
        return match["number"], match["unit"]
    msg = f"Invalid frequency string: {every}."
    raise ValueError(msg)
```

dangotbanned:
This *might* be handy if you're interested in how the sausage was made 😉 

- https://github.com/pola-rs/polars/blob/ad1c7194d3f6b8be792924d76b3616e418aa4159/crates/polars-time/src/windows/duration.rs

narwhals/utils.py
raisadz: Thanks for the suggestion @dangotbanned , looks great!

narwhals/utils.py
```diff
@@ -2000,3 +2003,35 @@ def to_tuple(self) -> tuple[_T, ...]:
...
+        "q": "quarter",
+        "mo": "month",
+        "d": "day",
+        "h": "hour",
+        "m": "minute",
+        "s": "second",
+        "ms": "millisecond",
+        "us": "microsecond",
+        "ns": "nanosecond",
+    }
```
MarcoGorelli, line 2037: could this dictionary just be a constant in `_duckdb/utils.py` and `_arrow/utils.py`?

narwhals/_arrow/series_dt.py
```diff
@@ -189,3 +191,13 @@ def total_nanoseconds(self) -> ArrowSeries:
         }
         factor = lit(unit_to_nano_factor[self.unit], type=pa.int64())
         return self.with_native(pc.multiply(self.native, factor).cast(pa.int64()))
+
+    def truncate(self, every: str) -> ArrowSeries:
+        number, unit = parse_interval_string(every)
+        unit_dict = unit_to_str()
+        if unit == "q" and number != "1":
+            msg = f"Only 1{unit} is currently supported for PyArrow."
```
MarcoGorelli, line 199: if we want to leave non-1 units with Q for a follow-up, can we add "for 'q'" to the error message? because it looks like it's supported for non-'q' units

narwhals/_duckdb/expr_dt.py
```diff
@@ -108,3 +112,22 @@ def total_microseconds(self) -> DuckDBExpr:
...
         raise NotImplementedError(msg)
+
+    def truncate(self, every: str) -> DuckDBExpr:
+        number, unit = parse_interval_string(every)
+        unit_dict = unit_to_str()
+        if unit == "ns":
+            msg = "Truncating to nanoseconds is not yet supported for DuckDB."
+            raise NotImplementedError(msg)
+        if unit in {"q", "y"} and number != "1":
+            msg = f"Only 1{unit} is currently supported for DuckDB."
```
MarcoGorelli, line 123: same comment as for pyarrow, let's specify the unit in the error message

narwhals/_spark_like/expr_dt.py
```diff
@@ -62,3 +65,16 @@ def _weekday(_input: Column) -> Column:
             return (self._compliant_expr._F.dayofweek(_input) + 6) % 7
 
         return self._compliant_expr._with_callable(_weekday)
+
+    def truncate(self, every: str) -> SparkLikeExpr:
+        number, unit = parse_interval_string(every)
+        if number != "1":
+            msg = "Only unit of 1 is currently implemented for Spark-like.\nGot {number}."
```
MarcoGorelli, line 72: missing `f` 😉

narwhals/utils.py
```diff
@@ -2000,3 +2003,35 @@ def to_tuple(self) -> tuple[_T, ...]:
...
         return it if isinstance(it, tuple) else tuple(it)
+
+
+def parse_interval_string(every: str) -> tuple[str, str]:
+    """Parse a string like "1d", "2h", "3m" into a tuple of (number, unit).
+
+    Returns:
+        A tuple of number and unit parsed from the interval string.
+    """
+    if match := _PATTERN_INTERVAL.match(every):
```
MarcoGorelli, line 2014: yeessss walrus 😍

narwhals/utils.py
```diff
@@ -2000,3 +2003,35 @@ def to_tuple(self) -> tuple[_T, ...]:
         # Collect and return as a `tuple`.
         it = self._into_iter()
         return it if isinstance(it, tuple) else tuple(it)
+
+
+def parse_interval_string(every: str) -> tuple[str, str]:
```
MarcoGorelli, line 2008: could we return the number as an int here, so the type is `tuple[int, str]`, and then we don't need to call `int(` on the result each time?

MarcoGorelli (commented):
thanks @raisadz ! 

left some minor comments

not sure i understand some of the xfails, gonna pull it to see if i can make sense of any

narwhals/utils.py
```diff
@@ -2000,3 +2002,15 @@ def to_tuple(self) -> tuple[_T, ...]:
...
+def parse_interval_string(every: str) -> tuple[int, str]:
+    """Parse a string like "1d", "2h", "3m" into a tuple of (number, unit).
+
+    Returns:
+        A tuple of multiple and unit parsed from the interval string.
+    """
+    if match := _PATTERN_INTERVAL.match(every):
+        return int(match["multiple"]), match["unit"]
+    msg = f"Invalid `every` string: {every}."
+    raise ValueError(msg)
```
dangotbanned, line 2016: @MarcoGorelli how would you feel about *pre-emptively* moving this and `_PATTERN_INTERVAL` to a new `_duration.py` module?

From your work in [`polars-time/src/windows/duration.rs`](https://github.com/pola-rs/polars/blob/ad1c7194d3f6b8be792924d76b3616e418aa4159/crates/polars-time/src/windows/duration.rs) I'm sure you're well-aware of the complexities if/when we move beyond this simple case.

I was going to suggest doing this *more generally* for `utils.py` - since I imagine we reached the current size by:
1. Add some small piece of functionality that is used in a lot of places
2. Throw it somewhere that doesn't cause circular imports (`utils`)
3. Possibly build on that functionality over time
4. Rinse/repeat

[On main](https://github.com/narwhals-dev/narwhals/blob/eb1368266cdae6d31ccfe1cde720104241232223/narwhals/utils.py) we're already at 2k LOC 😅

🥫 for 💭 😉

MarcoGorelli: sure, no objections 👍

narwhals/_arrow/series_dt.py
```diff
@@ -189,3 +191,15 @@ def total_nanoseconds(self) -> ArrowSeries:
...
+        multiple, unit = parse_interval_string(every)
+        if unit == "q" and multiple != 1:
+            msg = (
+                "For 'q' unit only multiple 1 is currently supported for PyArrow.\n"
+                f"Got multiple {multiple!s}."
+            )
+            raise NotImplementedError(msg)
+        return self.with_native(
+            pc.floor_temporal(self.native, multiple=multiple, unit=UNITS_DICT[unit])  # type: ignore[call-overload]
+        )
```
dangotbanned, line 205: The `[call-overload]` is related to the type of `unit`, an option could be using a `Literal` for `unit`.

https://github.com/zen-xu/pyarrow-stubs/blob/8fe826acfabb49a9fb84f9568d06c998a0c0455f/pyarrow-stubs/compute.pyi#L2028-L2040


So the return type [here](https://github.com/raisadz/narwhals/blob/b77110ed74d46757550bceec2abd3ada099df072/narwhals/utils.py#L2007) would change to something like:

```py
_IntervalUnit: TypeAlias = Literal[
    "ns", "us", "ms", "mo", "m", "s", "h", "d", "w", "q", "y"
]


def parse_interval_string(every: str) -> tuple[int, _IntervalUnit]: ...
```

Then we'd need to change the return statement to:

```py
        return int(match["multiple"]), cast("_IntervalUnit", match["unit"])
```

Then in `_arrow.utils`:


```py
NativeIntervalUnit: TypeAlias = Literal[
        "year",
        "quarter",
        "month",
        "week",
        "day",
        "hour",
        "minute",
        "second",
        "millisecond",
        "microsecond",
        "nanosecond",
    ]

UNITS_DICT: Mapping[_IntervalUnit, NativeIntervalUnit] = {
    "y": "year",
    "q": "quarter",
    "mo": "month",
    "d": "day",
    "h": "hour",
    "m": "minute",
    "s": "second",
    "ms": "millisecond",
    "us": "microsecond",
    "ns": "nanosecond",
}
```

Happy to hop on and add this in a commit 😄

raisadz: @dangotbanned sure, feel free to add it as a commit, thank you 😃

MarcoGorelli:
Investigating the test failures led to:

- https://github.com/apache/arrow/issues/46301
- https://github.com/pola-rs/polars/issues/22590

I think some API decision needs to be made on Polars' side, which we can choose to follow (or not)

For now, the simplest thing might be to just disallow `multiplier` being different from:
- `1`, `2`, `3`, `4`, `6`, `12` for 'month'
- `1`, `2`, `4` for 'quarter'
- `1` for 'year'

because then at least the results match independent of origin

Based on what the responses are from upstream libraries we can then choose what to do for other multipliers. I'd imagine that the ones listed here are by far the most common ones anyway

narwhals/utils.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2465/commits/6a1325662df991243e91aade5c2e7153d6e8af32)

narwhals/_arrow/series_dt.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2465/commits/f927348f5be38e53dee1aa7771d802cb6aa33ac7)

narwhals/_spark_like/expr_dt.py
```diff
@@ -62,3 +65,14 @@ def _weekday(_input: Column) -> Column:
...
         return self._compliant_expr._with_callable(_weekday)
+
+    def truncate(self, every: str) -> SparkLikeExpr:
+        multiple, unit = parse_interval_string(every)
+        if multiple != 1:
+            msg = f"Only multiple 1 is currently supported for Spark-like.\nGot {multiple!s}."
+            raise ValueError(msg)
+
+        def _truncate(_input: Column) -> Column:
+            return self._compliant_expr._F.date_trunc(UNITS_DICT[unit], _input)
```
dangotbanned, line 76: Very minor, but we can reuse the lookup across multiple calls this way

```suggestion
        format = UNITS_DICT[unit]

        def _truncate(_input: Column) -> Column:
            return self._compliant_expr._F.date_trunc(format, _input)
```

raisadz:
> Investigating the test failures led to:
> 
> * [Incorrect result for `floor_temporal` with 3 and 'year' apache/arrow#46301](https://github.com/apache/arrow/issues/46301)
> * [bug: inconsistent origin used in `dt.truncate` pola-rs/polars#22590](https://github.com/pola-rs/polars/issues/22590)
> 
> I think some API decision needs to be made on Polars' side, which we can choose to follow (or not)
> 
> For now, the simplest thing might be to just disallow `multiplier` being different from:
> 
> * `1`, `2`, `3`, `4`, `6`, `12` for 'month'
> * `1`, `2`, `4` for 'quarter'
> * `1` for 'year'
> 
> because then at least the results match independent of origin
> 
> Based on what the responses are from upstream libraries we can then choose what to do for other multipliers. I'd imagine that the ones listed here are by far the most common ones anyway

Thank you for investigating this, @MarcoGorelli ! I added validation for multiples of 'mo', 'q', and 'y' units in [616b01f](https://github.com/narwhals-dev/narwhals/pull/2465/commits/616b01f64d0fb598ddd01901b28a7bf007828a2d)

dangotbanned:
> thanks @raisadz and @dangotbanned ! i added some commit as i'm pretty keen to see this, hope it's ok
> 
> @dangotbanned any further comments?


Thanks @MarcoGorelli, I'll take a look v soon

narwhals/expr_dt.py
```diff
@@ -761,3 +761,52 @@ def timestamp(self, time_unit: TimeUnit = "us") -> ExprT:
...
         )
+
+    def truncate(self, every: str) -> ExprT:
+        """Divide the date/datetime range into buckets.
+
+        Arguments:
+            every: str
+                Length of bucket. Must be of form
+                `<multiple><unit>`, where `multiple` is a positive integer
+                and `unit` is one of:
```
dangotbanned, line 772:
```suggestion
            every: Length of bucket. Must be of form `<multiple><unit>`,
                where `multiple` is a positive integer and `unit` is one of:
```

The `str` annotation wasn't playing nice

image

narwhals/series_dt.py
```diff
@@ -653,3 +653,40 @@ def timestamp(self, time_unit: TimeUnit) -> SeriesT:
...
         )
+
+    def truncate(self, every: str) -> SeriesT:
+        """Divide the date/datetime range into buckets.
+
+        Arguments:
+            every: str
+                Length of bucket. Must be of form
+                `<multiple><unit>`, where `multiple` is a positive integer
+                and `unit` is one of:
```
dangotbanned, line 664:
```suggestion
            every: Length of bucket. Must be of form `<multiple><unit>`,
                where `multiple` is a positive integer and `unit` is one of:
```

narwhals/_duration.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import Literal
+from typing import cast
+from typing import get_args
+
+if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
+__all__ = ["IntervalUnit", "parse_interval_string"]
+
+IntervalUnit: TypeAlias = Literal["ns", "us", "ms", "s", "m", "h", "d", "mo", "q", "y"]
```
dangotbanned, line 16: Thinking this will be helpful when we're far away from the public `.dt` methods

```suggestion
IntervalUnit: TypeAlias = Literal["ns", "us", "ms", "s", "m", "h", "d", "mo", "q", "y"]
"""A Polars duration string interval unit.

- 'ns': nanosecond.
- 'us': microsecond.
- 'ms': millisecond.
- 's': second.
- 'm': minute.
- 'h': hour.
- 'd': day.
- 'mo': month.
- 'q': quarter.
- 'y': year.
"""
```

narwhals/_duration.py
```diff
@@ -0,0 +1,50 @@
...
+
+PATTERN_INTERVAL: re.Pattern[str] = re.compile(
+    r"^(?P<multiple>\d+)(?P<unit>ns|us|ms|mo|m|s|h|d|q|y)$"
```
dangotbanned, line 19: I saw this maybe an hour or two after I wrote (https://github.com/narwhals-dev/narwhals/pull/2465#discussion_r2070531991)

- https://discuss.python.org/t/proposal-add-z-as-a-synonym-for-z-in-python-res-for-standardization/90378

```suggestion
    r"^(?P<multiple>\d+)(?P<unit>ns|us|ms|mo|m|s|h|d|q|y)\Z"
```

narwhals/_pandas_like/series_dt.py
```diff
@@ -190,3 +192,39 @@ def timestamp(self, time_unit: TimeUnit) -> PandasLikeSeries:
...
+                arr_dtype = arr.dtype
+                result_arr = arr.astype(f"datetime64[{multiple}{np_unit}]").astype(
+                    arr_dtype
+                )
+            result_native = native.__class__(
+                result_arr, dtype=native.dtype, index=native.index, name=native.name
+            )
+            return self.with_native(result_native)
+        if unit == "m":  # replace "m" unit with "min"
+            every += "in"
```
dangotbanned, line 229: When comments like this were needed, it's usually a good sign to write something different 😉 

```suggestion
        if unit == "m":
            every = every.replace("m", "min", count=1)
```

narwhals/_dask/expr_dt.py
```diff
@@ -157,3 +158,15 @@ def total_nanoseconds(self) -> DaskExpr:
...
             "total_nanoseconds",
         )
+
+    def truncate(self, every: str) -> DaskExpr:
+        _, unit = parse_interval_string(every)
+        if unit in {"mo", "q", "y"}:
+            msg = f"Truncating to {unit} is not supported yet for dask."
+            raise NotImplementedError(msg)
+        if unit == "m":  # replace "m" unit with "min"
+            every += "in"
```
dangotbanned, line 168:
```suggestion
        if unit == "m":
            every = every.replace("m", "min", count=1)
```

dangotbanned (commented):
Thanks @MarcoGorelli, @raisadz!

Just a few nits and I think we're good to go 🥳

dangotbanned:
@MarcoGorelli maybe I'm getting ahead of myself, but might we wanna get a tracking issue up on? 

- Existing methods that *could accept* duration strings
- Yet-to-be-implemented methods/functionality that *require* duration strings

narwhals/_duration.py
MarcoGorelli: Not sure I follow the suggestion, sorry - is `\Z` preferred for some reason? I've never seen it, I've always only seen `^` and `$`. I'd suggest keeping this one as-is if it's ok

raisadz:
Thanks @MarcoGorelli and @dangotbanned for your suggestions and help!

narwhals/_duration.py
dangotbanned: @MarcoGorelli did you read the thread I linked? 

It is preferred and will be getting a new synonym (`\z`) - since (`$`) is commonly misused.
I linked the thread because **I did the exact misuse** 😅

dangotbanned: No worries @MarcoGorelli!

You're only 2 days behind me 😄

dangotbanned (approved):

Issue #2479, opened 2025-05-04
perf: minimise/remove overhead from `_typing_extensions_has_pep_696`


MarcoGorelli:
I was doing some profiling today:
```python
import duckdb
from duckdb import ColumnExpression, ConstantExpression
import narwhals as nw



def narwhals_soln(penguins):
    df = nw.from_native(penguins)
    return (
        df.filter(nw.col("species") == "Gentoo", nw.col("body_mass_g") > 6000)
        .with_columns(bill_length_cm = nw.col("bill_length_mm") / 10)
        .select(
            "species",
            "island",
            "bill_depth_mm",
            "flipper_length_mm",
            "body_mass_g",
            "sex",
            "bill_length_cm",
        )
        .collect()
        .to_native()
    )


path = '../scratch/penguins.parquet'

for _ in range(10):
    narwhals_soln(duckdb.read_parquet(path))
```
where `penguins.parquet` is this dataset: https://www.kaggle.com/datasets/marcogorelli/penguins

`tuna` shows this function as being surprisingly significant:

![Image](https://github.com/user-attachments/assets/fc1aaa4c-144e-4768-856b-d651a46d2f57)

PR #2480, opened 2025-05-04
perf: Dont export `stable` in ``narwhals/__init__.py``, instead require ``import narwhals.stable``

MarcoGorelli:
I noticed in https://github.com/narwhals-dev/narwhals/issues/2479 that this is another unnecessary source of overhead: even if someone just uses `import narwhals`, we still import `narwhals.stable`

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/__init__.py
```diff
@@ -160,7 +159,6 @@
     "scan_parquet",
     "selectors",
     "show_versions",
-    "stable",
```
dangotbanned, line 163: Have you tried enabling all *strictest* import related config?

- https://microsoft.github.io/pyright/#/typed-libraries?id=library-interface
- https://microsoft.github.io/pyright/#/configuration?id=type-check-rule-overrides
- https://mypy.readthedocs.io/en/stable/running_mypy.html#how-mypy-handles-imports
- https://mypy.readthedocs.io/en/stable/config_file.html#import-discovery

I'm not sure if we have downstream coverage that would catch this - but we might?

A similar (but not the same) regression from `pyright` and `pandas` recently
- https://github.com/microsoft/pyright/issues/10248

MarcoGorelli: The only case that this would break would be if someone did
```
import narwhals as nw
nw.stable
```
but there's no cases of that on GitHub https://github.com/search?q=%22nw.stable%22+language%3APython&type=code&l=Python

MarcoGorelli: I've tried running `pyright narwhals tests` with
```
reportMissingImports = "error"
reportMissingModuleSource = "none"
reportPrivateImportUsage = "error"
```
and it doesn't detected anything new

Reckon this is fine then?

dangotbanned: Yeah sounds good to me, thanks for checking @MarcoGorelli

MarcoGorelli:
thanks for your review!

PR #2493, opened 2025-05-04
fix: unify `ColumnNotFound` for `duckdb` and `pyspark`

### LINKED ISSUE #2472: error reporting: unify "column not found" error message for DuckDB / spark-like
```python
import narwhals as nw
import polars as pl

import duckdb

df = pl.DataFrame({'a': [1,2,3]})
nw.from_native(df).select(nw.col('b'))
nw.from_native(df.to_pandas()).select(nw.col('b'))
nw.from_native(df.to_arrow()).select(nw.col('b'))
nw.from_native(duckdb.table('df')).select(nw.col('b'))
```

All of the above, except the last one, raise `nw.exceptions.ColumnNotFoundError`

we should catch `duckdb.BinderError` and, if `'column cannot be referenced before it is defined'` is in the error message, raise `nw.exceptions.ColumnNotFoundError`, similarly to how we do for other backends

Then we should be able to avoid xfails in `select_test.py` `test_missing`

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2472 

## Checklist

- [x] Tests added
- [x] Documented the changes


EdAbati:
I think I can make some other clean-up of repetitive code. I'll try tomorrow morning

EdAbati:
I made a followup PR with the cleanup :)

narwhals/_duckdb/expr.py
```diff
@@ -186,7 +187,14 @@ def from_column_names(
         context: _FullContext,
     ) -> Self:
         def func(df: DuckDBLazyFrame) -> list[duckdb.Expression]:
-            return [col(name) for name in evaluate_column_names(df)]
+            col_names = evaluate_column_names(df)
+            missing_columns = [c for c in col_names if c not in df.columns]
```
MarcoGorelli, line 191: `df.columns` comes with overhead unfortunately, I think we should avoid calling it where possible. How much overhead depends on the operation

I was hoping we could do something like we do for Polars. That is to say, when we do `select` / `with_columns`, we wrap them in `try/except`, and in the `except` block we intercept the error message to give a more useful / unified one

MarcoGorelli (changes requested):
thanks for working on this! just got a comment on the `.columns` usage

narwhals/_duckdb/expr.py
EdAbati: Ah interesting, I was not aware 😕

What is happening in the background in duckdb that causes this overhead ? Do you have a link? (Just want to learn more)

Also, is it a specific caveat of duckdb? I don't think we should worry about that in spark-like but I might be wrong

I will update the code tonight anyway (but of course feel free to add commits to this branch if you need it for today's release)

dangotbanned: > `df.columns` comes with overhead unfortunately, I think we should avoid calling it where possible. How much overhead depends on the operation

@MarcoGorelli could we add that to (#805) and put more of a focus towards it? 🙏

MarcoGorelli: I don't think it's documented, but evaluating `.columns` may sometimes require doing a full scan. Example:
```python
In [48]: df = pl.DataFrame({'a': rng.integers(0, 10_000, 100_000_000), 'b': rng.integers(0, 10_000, 100_000_000)})

In [49]: rel = duckdb.table('df')
100% ▕████████████████████████████████████████████████████████████▏

In [50]: rel1 = duckdb.sql("""pivot rel on a""")

In [51]: %timeit rel.columns
385 ns ± 7.62 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [52]: %timeit rel1.columns
585 μs ± 3.8 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

Granted, we don't have `pivot` in the Narwhals lazy API, but a pivot may appear in the history of the relation which someone passes to `nw.from_native`, and the output schema of `pivot` is value-dependent (😩 )

The same consideration should apply to spark-like

MarcoGorelli: `.alias` for example is completely non-value-dependent, so that stays fast
```python
In [60]: %timeit rel.alias
342 ns ± 2.3 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [61]: %timeit rel1.alias
393 ns ± 2.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

narwhals/_spark_like/dataframe.py
```diff
@@ -274,6 +275,14 @@ def select(
     ) -> Self:
         new_columns = evaluate_exprs(self, *exprs)
         new_columns_list = [col.alias(col_name) for (col_name, col) in new_columns]
+        if not self._implementation.is_sqlframe():
```
EdAbati, line 278: `sqlframe` doesn't seem to raise any error at `select` time. I guess it gets triggered during collection

@MarcoGorelli do you have any suggestions about what to do in this case?

narwhals/_spark_like/dataframe.py
```diff
@@ -274,6 +275,14 @@ def select(
...
+            from pyspark.errors import AnalysisException
+
+            try:
+                return self._with_native(self.native.select(*new_columns_list))
+            except AnalysisException as e:
+                msg = f"Selected columns not found in the DataFrame.\n\nHint: Did you mean one of these columns: {self.columns}?"
```
EdAbati, line 284: Not 100% sure about this error message. I don't we can access the missing column names at this level, am I missing something?

narwhals/_spark_like/dataframe.py
EdAbati: I think I have an idea. I can test it like we do for polars lazy frames 🤔

narwhals/_spark_like/dataframe.py
MarcoGorelli: I think what you're written is great - even though we can't access them, we can still try to be helpful

tests/frame/select_test.py
EdAbati, line 1: I split the test into lazy and eager to simplify a bit the if-else statements. I hope it is a bit more readable ?

tests/frame/select_test.py
```diff
@@ -83,48 +84,68 @@ def test_comparison_with_list_error_message() -> None:
...
+        if constructor_id == "polars[lazy]":
+            # In the lazy case, Polars only errors when we call `collect`,
+            # and we have no way to recover exactly which columns the user
+            # tried selecting. So, we just emit their message (which varies
+            # across versions...)
+            return df.collect()
+        return df
+
+    if constructor_id == "polars[lazy]":
+        msg = r"^e|\"(e|f)\""
```
EdAbati, line 129: Before it was `"e|f"`. Now it is a bit stricter

tests/frame/select_test.py
```diff
@@ -83,48 +84,68 @@ def test_comparison_with_list_error_message() -> None:
...
+        pass
     else:
-        with pytest.raises(ColumnNotFoundError, match=msg):
-            df.select(selected_columns)
         with pytest.raises(ColumnNotFoundError, match=msg):
             df.drop(selected_columns, strict=True)
+    if "polars" in str(constructor_eager):
+        msg = r"\n\nHint: Did you mean one of these columns: \['a', 'b', 'z'\]?"
+    with pytest.raises(ColumnNotFoundError, match=msg):
+        df.select(nw.col("fdfa"))
```
EdAbati, line 106: before this was not tested for polars

tests/frame/select_test.py
```diff
@@ -83,48 +84,68 @@ def test_comparison_with_list_error_message() -> None:
...
+
+
+def test_missing_columns_lazy(
+    constructor_lazy: ConstructorLazy, request: pytest.FixtureRequest
+) -> None:
+    constructor_id = str(request.node.callspec.id)
+    if any(id_ == constructor_id for id_ in ("sqlframe", "pyspark[connect]")):
```
EdAbati, line 113: `sqlframe` raises at collect. 😕

I need to double check `pystpark.connect`. Currently cannot set it up locally... Working on it ⏳

narwhals/_duckdb/dataframe.py
```diff
@@ -184,7 +184,10 @@ def select(
         *exprs: DuckDBExpr,
     ) -> Self:
         selection = (val.alias(name) for name, val in evaluate_exprs(self, *exprs))
-        return self._with_native(self.native.select(*selection))
+        try:
+            return self._with_native(self.native.select(*selection))
+        except duckdb.BinderException as e:
+            raise ColumnNotFoundError.from_available_column_names(self.columns) from e
```
MarcoGorelli, line 190: do we risk catching other errors with this? `BinderException` might be a bit broad, shall we also match on `str(e)` before raising `ColumnNotFoundError`?

we should probably also do this for:
- with_columns
- simple_select
- filter

?

dangotbanned: @MarcoGorelli this has reminded me about an issue with `SparkLikeLazyFrame`

https://github.com/narwhals-dev/narwhals/blob/b7001e47070981d8f264de202f4d082245f61e82/narwhals/_spark_like/dataframe.py#L152-L164

This one is a bigger problem because it captures `CTRL+C`, so you can't easily stop the test suite while it's running

tests/frame/select_test.py
```diff
@@ -83,48 +83,57 @@ def test_comparison_with_list_error_message() -> None:
...
-        else:
-            with pytest.raises(ColumnNotFoundError, match=msg):
-                df.select(selected_columns)
-        if POLARS_VERSION >= (1,):
-            # Old Polars versions wouldn't raise an error
-            # at all here
-            if isinstance(df, nw.LazyFrame):
-                with pytest.raises(ColumnNotFoundError, match=msg):
-                    df.drop(selected_columns, strict=True).collect()
-            else:
```
EdAbati, line 116: `drop` should be already tested in `drop_test.py`

tests/frame/drop_test.py
```diff
@@ -55,6 +42,19 @@ def test_drop_strict(
...
 
+    context: Any
+    if strict:
+        if "polars_lazy" in str(constructor):
+            msg = r"^\"z\""
+        else:
+            msg = (
+                r"The following columns were not found: \[.*\]"
+                r"\n\nHint: Did you mean one of these columns: \['a', 'b'\]?"
+            )
```
EdAbati, line 53: We use `parse_columns_to_drop` and therefore raise our `ColumnNotFoundError.from_missing_and_available_column_names(missing_columns=missing_columns, available_columns=cols)` for every backend

narwhals/_duckdb/dataframe.py
```diff
@@ -184,7 +184,12 @@ def select(
         *exprs: DuckDBExpr,
     ) -> Self:
         selection = (val.alias(name) for name, val in evaluate_exprs(self, *exprs))
-        return self._with_native(self.native.select(*selection))
+        try:
+            return self._with_native(self.native.select(*selection))
+        except duckdb.BinderException as e:
+            if "this column cannot be referenced before it is defined" in str(e):
+                raise ColumnNotFoundError.from_available_column_names(self.columns) from e
+            raise
```
MarcoGorelli, line 192: should we do this for `simple_select` and `aggregate` too?

i think in polars we have a `catch_exception` utility, could/should we make something like that?

EdAbati: Sorry @MarcoGorelli I didn't have time to look at this yet :(

I just saw that `catch_polars_exception` it is way cleaner! I'll update later tonight

EdAbati:
I finally had a few minutes to go back to this 🥲

Few notes:
- `catch_{}_exception` made it cleaner thanks for the tip
- `sqlframe` is tricky because it raises at collect time. Also the error will be different based on the backend. Do we have a way to know which type of `sqlframe` we are dealing with. For `duckdb` we could use the `catch_duckdb_exception` but we need to make sure `duckdb` is available. Any ideas? Could we think about it in a follow up?
- regarding `simple_select`, it should already be covered by the below. We are already testing it too:
https://github.com/narwhals-dev/narwhals/blob/b5f72dd36926d8be9506754cd9e92f8a855a42b8/narwhals/dataframe.py#L165-L181
Would you like to do something else for lazy backends?

EdAbati (commented):
There are a couple of annoying errors I'll check later.

IMO they seem unrelated (mostly the `pyspark` one). The `typing` errors I cannot replicate locally 👀

dangotbanned:
> `sqlframe` is tricky because it raises at collect time. Also the error will be different based on the backend. Do we have a way to know which type of sqlframe we are dealing with?

@EdAbati IIRC there's some import-related functions in `_spark_like.utils` that may be helpful for you?

I'm assuming this would work in a similar way, where you either need to use some `Base*` class or use the string in the concrete type in an import path

EdAbati:
Found some time to update this. (sorry for the late reply)

> @EdAbati IIRC there's some import-related functions in _spark_like.utils that may be helpful for you?

The problem IMO is that since `sqlframe` lets backend raise their errors we can only intercepts the ones of the backends we also support (i.e. `pyspark` and `duckdb`)
Not sure if the best solution would be to let `sqlframe` do their error handling or to intercept the errors just for the backend we support.
Maybe it should be discussed in an issue/follow-up?

@MarcoGorelli is there anything that you think is missing now? :)

MarcoGorelli:
thanks! i think the logic looks right, the tests look a little complex but maybe that's ok. will get back to this shortly

tests/conftest.py
```diff
@@ -306,5 +312,9 @@ def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
         metafunc.parametrize(
             "constructor_eager", eager_constructors, ids=eager_constructors_ids
         )
+    elif "constructor_lazy" in metafunc.fixturenames:
+        metafunc.parametrize(
+            "constructor_lazy", lazy_constructors, ids=lazy_constructors_ids
+        )
```
MarcoGorelli: I feel slightly uneasy about adding an extra constructor just for one test. and if we need to add it, then maybe `constructors` could be a union of `eager_constructors` and `lazy_constructors`, rather than making all 3?

if possible, i'd suggest to leave as-is for now and see if it's possible to use `constructor` in the test

MarcoGorelli (commented):
thanks @EdAbati ! looks good to me

just a couple of merge conflicts and a suggestion on the tests, but then i'd say we can ship it 🚢

MarcoGorelli:
> The problem IMO is that since sqlframe lets backend raise their errors we can only intercepts the ones of the backends we also support (i.e. pyspark and duckdb)
Not sure if the best solution would be to let sqlframe do their error handling or to intercept the errors just for the backend we support.
Maybe it should be discussed in an issue/follow-up?

Yeah that's fine - I think in general it's ok to aim for "we try to unify what we can, but there may be some differences that we have no control over"

tests/conftest.py
EdAbati: I updated now. I thought `test_missing_columns` was going to be less readable, but it doesn't make a lot of difference.

tests/utils.py
```diff
@@ -42,6 +42,7 @@ def get_module_version_as_tuple(module_name: str) -> tuple[int, ...]:
 
 Constructor: TypeAlias = Callable[[Any], "NativeLazyFrame | NativeFrame | DataFrameLike"]
 ConstructorEager: TypeAlias = Callable[[Any], "NativeFrame | DataFrameLike"]
+ConstructorLazy: TypeAlias = Callable[[Any], "NativeLazyFrame"]
```
EdAbati, line 45: @MarcoGorelli do you think we should delete this too?
I think it actually makes the `LAZY_CONSTRUCTORS: dict[str, ConstructorLazy]` a bit more accurate/stricter

EdAbati (commented):
Sorry for the delay again 🥲 let me know if there is still something that looks off, I have some time today to update

and thank you for the review

tests/frame/select_test.py
```diff
@@ -86,49 +88,50 @@ def test_comparison_with_list_error_message() -> None:
 def test_missing_columns(
     constructor: Constructor, request: pytest.FixtureRequest
 ) -> None:
-    if (
-        ("pyspark" in str(constructor))
-        or "duckdb" in str(constructor)
-        or "ibis" in str(constructor)
-    ):
+    constructor_id = str(request.node.callspec.id)
+    if any(id_ == constructor_id for id_ in ("sqlframe", "ibis")):
```
MarcoGorelli, line 90: nice! should we do this pattern throughout the test suite / have some helper function for it?

tests/frame/with_columns_test.py
```diff
@@ -78,3 +78,31 @@ def test_with_columns_series_shape_mismatch(constructor_eager: ConstructorEager)
...
     with pytest.raises(ShapeError):
         df1.with_columns(second=second)
+
+
+def test_with_columns_missing_column(
+    constructor: Constructor, request: pytest.FixtureRequest
+) -> None:
+    constructor_id = str(request.node.callspec.id)
+    if any(id_ == constructor_id for id_ in ("sqlframe", "pyspark[connect]", "ibis")):
+        # These backend raise errors at collect
```
MarcoGorelli: is it an issue that they raise errors at collect?

because below, we do
```python
    if "polars_lazy" in str(constructor) and isinstance(df, nw.LazyFrame):
        # In the lazy case, Polars only errors when we call `collect`
        with pytest.raises(ColumnNotFoundError, match=msg):
            df.with_columns(d=nw.col("c") + 1).collect()
```

perhaps we could change that to just be
```
    if isinstance(df, nw.LazyFrame):
        # In the lazy case, Polars only errors when we call `collect`
        with pytest.raises(ColumnNotFoundError, match=msg):
            df.with_columns(d=nw.col("c") + 1).collect()
```
?

EdAbati: Good point! 

actually the comment here was misleading, fixed it.
At collect `sqlframe` will re-surface the error of the backend it uses, it won't be a `ColumnNotFoundError` (yet?). `ibis` was introduced after I started working on this I think (soooorry it took soo long 🥲), maybe we can work on ibis in a follow up?

Also `"pyspark[connect]"` should not have been a xfail. now it is tested. Thanks for the catch :)

tests/frame/select_test.py
EdAbati: Do you mean using `request.node.callspec.id` instead of the entire `constructor` as string?

Yeah I like it a bit more too, I will make a followup PR :)

MarcoGorelli (approved):
thanks @EdAbati !

EdAbati:
Thank you @MarcoGorelli for the review and for the patience with this one 🥲❤️

dangotbanned:
@EdAbati @MarcoGorelli does this error seem related to this PR?

- https://github.com/narwhals-dev/narwhals/actions/runs/16328848063/job/46125919696

I haven't touched anything I swear 😂

EdAbati:
> does this error seem related to this PR?
> 
> * https://github.com/narwhals-dev/narwhals/actions/runs/16328848063/job/46125919696

yeah It looks like the "stricter" regex (that start with `^`) are not ok for polars `0.20.22` :( 

I have some time to update this tonight, I'll make a PR if no one does it before me :)

PR #2494, opened 2025-05-04
chore: Simplify `PandasLikeDataFrame.pivot`

### LINKED ISSUE #2490: Simplify `PandasLikeDataFrame.pivot`
https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_pandas_like/dataframe.py#L931

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2490

## Checklist

- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below. And two tiny perf

Not sure that it makes it significantly simpler, but enough to not fail the ruff rules

### DIFF
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -928,7 +928,7 @@ class PandasLikeDataFrame(
     def gather_every(self, n: int, offset: int) -> Self:
         return self._with_native(self.native.iloc[offset::n], validate_column_names=False)
 
-    def pivot(  # noqa: C901, PLR0912
+    def pivot(
         self,
         on: Sequence[str],
         *,
@@ -960,9 +960,8 @@ class PandasLikeDataFrame(
             result = frame.pivot(columns=on, index=index, values=values)
         elif aggregate_function == "len":
             result = (
-                frame.groupby([*on, *index])
+                frame.groupby([*on, *index], as_index=False)
                 .agg(dict.fromkeys(values, "size"))
-                .reset_index()
                 .pivot(columns=on, index=index, values=values)
             )
         else:
@@ -975,22 +974,18 @@ class PandasLikeDataFrame(
             )
 
         # Put columns in the right order
-        if sort_columns and self._implementation is Implementation.CUDF:
-            uniques = {
-                col: sorted(self.native[col].unique().to_arrow().to_pylist())
-                for col in on
-            }
-        elif sort_columns:
-            uniques = {col: sorted(self.native[col].unique().tolist()) for col in on}
-        elif self._implementation is Implementation.CUDF:
-            uniques = {
-                col: self.native[col].unique().to_arrow().to_pylist() for col in on
-            }
-        else:
-            uniques = {col: self.native[col].unique().tolist() for col in on}
+        uniques = (
+            {col: self.native[col].unique().to_arrow().to_pylist() for col in on}
+            if self._implementation.is_cudf()
+            else {col: self.native[col].unique().tolist() for col in on}
+        )
+
+        if sort_columns:
+            uniques = {k: sorted(v) for k, v in uniques.items()}
+
         ordered_cols = list(product(values, *uniques.values()))
         result = result.loc[:, ordered_cols]
-        columns = result.columns.tolist()
+        columns = result.columns
 
         n_on = len(on)
         if n_on == 1:
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -942,6 +942,56 @@ class PandasLikeDataFrame(
         values = values or exclude_column_names(self, {*on, *index})
         return index, values
 
+    @staticmethod
+    def _pivot_format_single_col(col_values: tuple[str, ...]) -> str:
+        """Format single column values to match polars output in case of multiple `on` values."""
+        return '{"' + '","'.join(col_values) + '"}'
+
+    def _pivot_format_output_column_names(
+        self,
+        native_columns: pd.Index[str] | pd.MultiIndex,
+        *,
+        on: Sequence[str],
+        values: Sequence[str],
+        separator: str,
+    ) -> list[str]:
+        """Reformat output column names from a native pandas-like pivot operation.
+
+        Such ops columns are a multi-index object that we are going to manipulate and re-map
+        to match polars pivot column names.
+
+        - If `len(on) == 1` and `len(values) == 1`, then we get the unique values of the
+            columns in `values`.
+        - If `len(on) == 1` and `len(values_) > 1`, then we flatten and join all the native
+            multi-index output column names, separating them using the `separator` provided.
+        - If `len(on) > 1`, then the values of each `on` column is represented as
+            '{"v11", "v21"}', '{"v11", "v22"}', '{"v12", "v21"}', '{"v12", "v22"}' where
+            v11, v12 are the unique values of `on[0]` column and v21, v22 are the unique
+            values of the `on[1]` column (and so on..).
+            - Now, if `len(values) == 1` then no other column identifier is required
+            - if `len(values) > 1`, then those should be prefixed with the values column
+                names. E.g. `values = ["x", "y"]` would lead to the following output columns
+                'x_{"v11", "v21"}', 'x_{"v11", "v22"}', ..., 'y_{"v12", "v21"}', 'y_{"v12", "v22"}'
+                where `"_"` is the separator value.
+        """
+        n_on = len(on)
+        n_values = len(values)
+
+        if n_on == 1:
+            new_columns = [
+                separator.join(col).strip() if n_values > 1 else col[-1]
+                for col in native_columns
+            ]
+        else:
+            new_columns = [
+                separator.join([col[0], self._pivot_format_single_col(col[-n_on:])])  # type: ignore[arg-type]
+                if len(values) > 1
+                else self._pivot_format_single_col(col[-n_on:])  # type: ignore[arg-type]
+                for col in native_columns
+            ]
+
+        return new_columns
+
     def _pivot_table(
         self,
         on: Sequence[str],
@@ -1021,38 +1071,9 @@ class PandasLikeDataFrame(
         ordered_cols = list(product(values, *chain(uniques)))
         result = result.loc[:, ordered_cols]
         columns = result.columns
-
-        # Reformat output column names to match polars ones
-        # pandas-like native output columns are multi-index object that we are going to
-        # manipulate and re-map to match polars.
-        n_on = len(on)
-        if n_on == 1:
-            # If `n_on == 1` and `len(values_) == 1`, then we simply get the unique values
-            # of such `values_` column
-            # If `n_on == 1` and `len(values_) > 1`, then we flatten and join all the native
-            # multi-index output column names, separating them using the `separator` provided.
-            new_columns = [
-                separator.join(col).strip() if len(values) > 1 else col[-1]
-                for col in columns
-            ]
-        else:
-            # If `n_on > 1`, then the values of each `on` column is represented as
-            # '{"v11", "v21"}', '{"v11", "v22"}', '{"v12", "v21"}', '{"v12", "v22"}' where
-            # v11, v12 are the unique values of `on[0]` column and v21, v22 are the unique
-            # values of the `on[1]` column.
-            # Now, if len(values)==1 then no other column identifier is required; if
-            # len(values)>1, then those are prefixed with the values column names. E.g.
-            # `values = ["x", "y"]` would lead to the following output columns
-            # 'x_{"v11", "v21"}', 'x_{"v11", "v22"}', ..., 'y_{"v12", "v21"}', 'y_{"v12", "v22"}'  # noqa: ERA001
-            # where `"_"` is an example for the separator value.
-            new_columns = [
-                separator.join([col[0], '{"' + '","'.join(col[-n_on:]) + '"}'])
-                if len(values) > 1
-                else '{"' + '","'.join(col[-n_on:]) + '"}'
-                for col in columns
-            ]
-
-        result.columns = new_columns
+        result.columns = self._pivot_format_output_column_names(
+            native_columns=columns, on=on, values=values, separator=separator
+        )
         result.columns.names = [""]  # type: ignore[attr-defined]
         return self._with_native(result.reset_index())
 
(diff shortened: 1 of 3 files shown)


dangotbanned:
These two should really be a function that's called twice

https://github.com/narwhals-dev/narwhals/blob/86e7bf868beae63f3e71a1d248d95b1e28f49d19/narwhals/_pandas_like/dataframe.py#L953-L957

The only difference is where `index` and `values` are passed

dangotbanned:
Whatever this is doing (I'm lost 😅) deserves to be split out and given a name

https://github.com/narwhals-dev/narwhals/blob/86e7bf868beae63f3e71a1d248d95b1e28f49d19/narwhals/_pandas_like/dataframe.py#L988-L1002

FBruzzesi:
> Whatever this is doing (I'm lost 😅) deserves to be split out and given a name

Aligning with polars output column names is not easy task 😅

narwhals/_pandas_like/dataframe.py
```diff
@@ -938,73 +1000,58 @@ def pivot(  # noqa: C901, PLR0912
...
+            # `values = ["x", "y"]` would lead to the following output columns
+            # 'x_{"v11", "v21"}', 'x_{"v11", "v22"}', ..., 'y_{"v12", "v21"}', 'y_{"v12", "v22"}'  # noqa: ERA001
+            # where `"_"` is an example for the separator value.
             new_columns = [
                 separator.join([col[0], '{"' + '","'.join(col[-n_on:]) + '"}'])
                 if len(values) > 1
                 else '{"' + '","'.join(col[-n_on:]) + '"}'
                 for col in columns
             ]
+
```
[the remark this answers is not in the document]
dangotbanned: A starting point could be this expression

https://github.com/narwhals-dev/narwhals/blob/cd92372f3bd5a5d6339eae09ab7c476328c032cb/narwhals/_pandas_like/dataframe.py#L1051

Which is a repeat of part of the expression in the other branch

https://github.com/narwhals-dev/narwhals/blob/cd92372f3bd5a5d6339eae09ab7c476328c032cb/narwhals/_pandas_like/dataframe.py#L1049-L1049

The only difference is `col[0], `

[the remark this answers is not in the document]
FBruzzesi: Hey @dangotbanned thanks for speeding this up. I broke it down in https://github.com/narwhals-dev/narwhals/pull/2494/commits/7c10412e8d9f76f813fda9027d9cd08e4992f595 

I added to couple of `# type: ignore` due to different behaviour for `pd.Index[str]` vs `pd.MultiIndex` 🤔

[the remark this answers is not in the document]
dangotbanned: @FBruzzesi before I approve, I just wanted to check in with you on the changes in ([refactor: The great splitting part ✌️](https://github.com/narwhals-dev/narwhals/pull/2494/commits/ba99df2004d9fa027deca6c435ad1d20d37f99e4))

Another way to look at this is, which would you prefer as a starting point if asked to implement `ArrowDataFrame.pivot`?

https://github.com/narwhals-dev/narwhals/blob/b86b5a2b49c5a195f07b63553b771d6652f06cdb/narwhals/_arrow/dataframe.py#L781

If we have a huge comment/docstring explaining how the code works, then we'd likely have to copy the whole thing over.

To me, it would be more helpful to have lots of *small* named steps (methods) that the operation is composed from.
That way - we have less to think about in translating each and can add in `pyarrow` specific quirks where needed 😅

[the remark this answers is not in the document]
FBruzzesi: >To me, it would be more helpful to have lots of small named steps (methods) that the operation is composed from.

Yes it makes sense, and I can see the benefit of it, especially if someone wants to address `pivot` from scratch for other backends.

On the other hand my _very personal_ preference is to not go too granular for functionalities used only once 🤷🏼‍♀️

- Going too granular for me does not lower the overall complexity, which goes from flatten to nested (a method calling a method calling a method)
  - Namely we went from a 10 lines, which to me were not particularly complex if documented, to 4 extra methods calling each other, yet without much docstrings explaining what's happening 
- For my mental model and/or mental load, when reading what's happening the path goes like `pivot` -> `_pivot_remap_column_names` -> `_pivot_single_on_names` and `_pivot_multi_on_names` -> `_pivot_multi_on_name`. If I skip any of this, I store a checkpoint in my head to something I need  jump back to and check eventually to fully understand what's happening 😅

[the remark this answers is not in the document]
dangotbanned: @FBruzzesi maybe it would help understand my motivation a bit better, if we de-sugar (https://github.com/narwhals-dev/narwhals/pull/2494/commits/7c10412e8d9f76f813fda9027d9cd08e4992f595)?


<details><summary>De-sugared</summary>
<p>

```py
    def _pivot_format_output_column_names(
        self,
        native_columns: pd.Index[str] | pd.MultiIndex,
        *,
        on: Sequence[str],
        values: Sequence[str],
        separator: str,
    ) -> list[str]:
        n_on = len(on)
        n_values = len(values)
        new_columns: list[Any] = []

        if n_on == 1:
            for col in native_columns:
                if n_values > 1:  # noqa: SIM108
                    item = separator.join(col).strip()
                else:
                    item = col[-1]
                new_columns.append(item)
        else:
            for col in native_columns:
                if len(values) > 1:
                    item = separator.join(
                        [col[0], self._pivot_format_single_col(col[-n_on:])]  # type: ignore[arg-type]
                    )
                else:
                    item = self._pivot_format_single_col(col[-n_on:])  # type: ignore[arg-type]
                new_columns.append(item)

        return new_columns
```

</p>
</details> 

I started by doing this and made two observations:

1. There are two versions of the same predicate
    i. `n_values > 1` and `len(values) > 1`
2. The predicate is evaluated for every iteration, but is the same for all of them
    i. The second variant is *also* calling `len` every time

IMO, these things are much easier to miss when inside a listcomp/genexpr.
So the first thing was move them to be *once-per-pivot*:

https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L958

https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L970

---

> Namely we went from a 10 lines, to 4 extra methods calling each other, without docstrings explaining what's happening, except the method names themselves.

We could add docs if you want?
I thought these logically separated things:

Distinct paths for where `n_on` is needed (`single_on` or `multi_on`)
https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L992-L994

I didn't think this one needed an explanation once it was broken down this far:

https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L958-L960

Here, I was trying to turn the contents of your doc into the variable names

https://github.com/narwhals-dev/narwhals/blob/7c10412e8d9f76f813fda9027d9cd08e4992f595/narwhals/_pandas_like/dataframe.py#L972

https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L970-L974

In my head that *slight* change made the difference between these two branches really clear

https://github.com/narwhals-dev/narwhals/blob/54fc3e9c6d043fbe22d38a7279d084efcf527464/narwhals/_pandas_like/dataframe.py#L975-L977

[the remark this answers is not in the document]
dangotbanned: > For my mental model and/or mental load, when reading what's happening the path goes like `pivot` -> `_pivot_remap_column_names` -> `_pivot_single_on_names` and `_pivot_multi_on_names` -> `_pivot_multi_on_name`. If I skip any of this, I store a checkpoint in my head to something I need jump back to and check eventually to fully understand what's happening 😅

To be honest, I'm not *super happy* about the method names.
I was thinking they might be a starting point for namespacing the pivot stuff.

So if we directly translated this section, we'd have:

```py
class PandasLikePivot:
    def remap_column_names(self):...
    def _single_on_names(self):...
    def _multi_on_names(self):...
    def _multi_on_name(self):...
```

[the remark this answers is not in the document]
FBruzzesi: Please don't get me wrong, I definitly understand the reasoning behind these kinds of breakdown - it's more about the fact I find it harder to reason about hence to write. I will definitly keep that in mind for future refactoring/new features ✨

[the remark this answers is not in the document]
dangotbanned: @FBruzzesi I'm happy if you wanna make changes to get (https://github.com/narwhals-dev/narwhals/pull/2494/commits/ba99df2004d9fa027deca6c435ad1d20d37f99e4) back to somewhere you're happier with?

More than anything, (https://github.com/narwhals-dev/narwhals/pull/2494#discussion_r2073588743) was me wanting to make sure we had this discussion - more so than trying to convince you my way is best 😄

narwhals/_pandas_like/dataframe.py
```diff
@@ -928,85 +930,159 @@ def clone(self) -> Self:
...
-                col: sorted(self.native[col].unique().to_arrow().to_pylist())
+        index, values = self._pivot_into_index_values(on, index, values)
+        result = self._pivot(on, index, values, aggregate_function)
+
+        # Select the columns in the right order
+        uniques = (
+            (
+                self.get_column(col)
+                .unique()
+                .sort(descending=False, nulls_last=False)
```
FBruzzesi, line 1071: @dangotbanned I reverted the addition of default values in `sort` - I don't think one exception justifies to break the "compliant method should not have default" rile 😁

narwhals/_pandas_like/dataframe.py
```diff
@@ -928,85 +930,159 @@ def clone(self) -> Self:
...
+                .to_list()
```
FBruzzesi, line 1072: Do we need to trigger a `to_list`? I would need to check what happens in the cudf case  🤔

narwhals/_pandas_like/series.py
```diff
@@ -601,9 +601,11 @@ def cum_sum(self, *, reverse: bool) -> Self:
         )
         return self._with_native(result)
 
-    def unique(self, *, maintain_order: bool) -> PandasLikeSeries:
-        # pandas always maintains order, as per its docstring:
-        # "Uniques are returned in order of appearance"  # noqa: ERA001
+    def unique(self, *, maintain_order: bool = True) -> PandasLikeSeries:
```
FBruzzesi, line 604: Here it's not a problem to add a default as it's unused anyway

narwhals/_pandas_like/dataframe.py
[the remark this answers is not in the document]
FBruzzesi: > To be honest, I'm not super happy about the method names.

I agree, I don't think they are descriptive enough or particularly intuitive. Maybe that's a subconscious rule - if you can't name something, it's hard to factor out into a method/function 😂

narwhals/_pandas_like/dataframe.py
dangotbanned: @FBruzzesi it's the `Protocol`s themselves where we don't want the defaults.

The implementations **can** have defaults, but the issue is when you define it in the `Protocol` then they **must** have them as well

narwhals/_pandas_like/dataframe.py
dangotbanned: That was already there, I didn't really question it 🤷‍♂️

dangotbanned: https://github.com/narwhals-dev/narwhals/blob/b86b5a2b49c5a195f07b63553b771d6652f06cdb/narwhals/_pandas_like/dataframe.py#L983-L990

FBruzzesi: Yes sorry, I didn't mean to blame it on you - once refactoring I think it's a good opportunity to revisit for other improvements as well.

One other tiny example is:

```diff
(
+     self.native.groupby([*on, *index], as_index=False)
-     self.native.groupby([*on, *index])
      .agg(dict.fromkeys(values, "size"))
-     .reset_index()
      .pivot(columns=on, index=index, values=values)
)

narwhals/_pandas_like/dataframe.py
FBruzzesi: I am a bit confused on the following statement:

> The implementations can have defaults, but the issue is when you define it in the Protocol then they must have them as well

am I naive to think that we still try to avoid setting defaults as much as possible in any compliant level?

Regarding:

> This also means you now need to keep the default of Series.sort in sync with DataFrame.pivot - whereas the other way we're just saying

For me the issue arise from the assumption that `sort_columns` flag will always behave _exactly_ as `sort` with default values. I don't see why that should be necessarily the case, and in practice, we could/should not keep `sort` default values aligned with what's required in pivot (nor viceversa).

narwhals/_pandas_like/dataframe.py
dangotbanned: Maybe we'd want another issue for following up the `to_list`?

I noticed that there's also [`pandas.MultiIndex.from_product`](https://pandas.pydata.org/docs/reference/api/pandas.MultiIndex.from_product.html), which might simplify/improve the performance of this part as well?

https://github.com/narwhals-dev/narwhals/blob/9643379dd753f7104debaf23c4f88dba7f30373b/narwhals/_pandas_like/dataframe.py#L1078-L1080

I'm totally unfamiliar with that method, so might not be helpful for us 😅

FBruzzesi: > Maybe we'd want another issue for following up the to_list?

Sure

> I'm totally unfamiliar with that method, so might not be helpful for us 😅

That makes two of us

narwhals/_pandas_like/dataframe.py
dangotbanned: @FBruzzesi fantastic 😂

Okay I'm just gonna approve and merge

dangotbanned (approved):
Thanks for bearing with me on this @FBruzzesi 😍

PR #2495, opened 2025-05-05
chore: cleanup `ColumnNotFoundError`

EdAbati:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

-  Followup of #2493 

## Checklist

- [x] Tests added
- [x] Documented the changes

> [!WARNING]  
> The diff of this PR includes also the diff of #2493 . I created a separate PR because I feel these changes are beyond the scope of the other one. And to make the review easier :)

### DIFF
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -8,7 +8,7 @@ from narwhals._compliant import EagerExpr
 from narwhals._expression_parsing import evaluate_output_names_and_aliases
 from narwhals._pandas_like.group_by import PandasLikeGroupBy
 from narwhals._pandas_like.series import PandasLikeSeries
-from narwhals.exceptions import ColumnNotFoundError
+from narwhals.utils import check_columns_exists
 from narwhals.utils import generate_temporary_column_name
 
 if TYPE_CHECKING:
@@ -128,13 +128,11 @@ class PandasLikeExpr(EagerExpr["PandasLikeDataFrame", PandasLikeSeries]):
                     for column_name in evaluate_column_names(df)
                 ]
             except KeyError as e:
-                missing_columns = [
-                    x for x in evaluate_column_names(df) if x not in df.columns
-                ]
-                raise ColumnNotFoundError.from_missing_and_available_column_names(
-                    missing_columns=missing_columns,
-                    available_columns=df.columns,
-                ) from e
+                check_columns_exists(
+                    evaluate_column_names(df), available_columns=df.columns, from_error=e
+                )
+                msg = "Unreachable code, please report a bug."
+                raise AssertionError(msg) from e
 
         return cls(
             func,
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -15,6 +15,7 @@ from narwhals._spark_like.utils import import_functions
 from narwhals._spark_like.utils import import_native_dtypes
 from narwhals._spark_like.utils import import_window
 from narwhals._spark_like.utils import native_to_narwhals_dtype
+from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import InvalidOperationError
 from narwhals.typing import CompliantLazyFrame
 from narwhals.utils import Implementation
@@ -257,6 +258,13 @@ class SparkLikeLazyFrame(
         raise ValueError(msg)  # pragma: no cover
 
     def simple_select(self, *column_names: str) -> Self:
+        df_columns = self.columns
+        missing_columns = [c for c in column_names if c not in df_columns]
+        if missing_columns:
+            raise ColumnNotFoundError.from_missing_and_available_column_names(
+                missing_columns=missing_columns,
+                available_columns=df_columns,
+            )
         return self._with_native(self.native.select(*column_names))
 
     def aggregate(
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -366,7 +366,8 @@ class SparkLikeLazyFrame(
     def unique(
         self, subset: Sequence[str] | None, *, keep: LazyUniqueKeepStrategy
     ) -> Self:
-        check_column_exists(self.columns, subset)
+        if subset:
+            check_column_exists(subset, available_columns=self.columns)
         subset = list(subset) if subset else None
         if keep == "none":
             tmp = generate_temporary_column_name(8, self.columns)
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -19,7 +19,7 @@ from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import InvalidOperationError
 from narwhals.typing import CompliantLazyFrame
 from narwhals.utils import Implementation
-from narwhals.utils import check_column_exists
+from narwhals.utils import check_columns_exists
 from narwhals.utils import find_stacklevel
 from narwhals.utils import generate_temporary_column_name
 from narwhals.utils import not_implemented
@@ -367,7 +367,7 @@ class SparkLikeLazyFrame(
         self, subset: Sequence[str] | None, *, keep: LazyUniqueKeepStrategy
     ) -> Self:
         if subset:
-            check_column_exists(subset, available_columns=self.columns)
+            check_columns_exists(subset, available_columns=self.columns)
         subset = list(subset) if subset else None
         if keep == "none":
             tmp = generate_temporary_column_name(8, self.columns)
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -15,7 +15,6 @@ from narwhals._spark_like.utils import import_functions
 from narwhals._spark_like.utils import import_native_dtypes
 from narwhals._spark_like.utils import import_window
 from narwhals._spark_like.utils import native_to_narwhals_dtype
-from narwhals.exceptions import ColumnNotFoundError
 from narwhals.exceptions import InvalidOperationError
 from narwhals.typing import CompliantLazyFrame
 from narwhals.utils import Implementation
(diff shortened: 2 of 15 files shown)


narwhals/_pandas_like/expr.py
```diff
@@ -128,13 +128,11 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
-                ]
-                raise ColumnNotFoundError.from_missing_and_available_column_names(
-                    missing_columns=missing_columns,
-                    available_columns=df.columns,
-                ) from e
+                check_columns_exists(
+                    evaluate_column_names(df), available_columns=df.columns, from_error=e
+                )
+                msg = "Unreachable code, please report a bug."
+                raise AssertionError(msg) from e
```
EdAbati, line 135: I had to add this otherwise the typechecker complains.

Can you think of a cleaner/better way?

narwhals/utils.py
```diff
@@ -1541,10 +1537,19 @@ def generate_repr(header: str, native_repr: str) -> str:
...
+    available_columns: Sequence[str],
+    from_error: Exception | None = None,
+) -> None:
+    if missing := set(columns).difference(available_columns):
+        error = ColumnNotFoundError.from_missing_and_available_column_names(
+            missing_columns=sorted(missing), available_columns=list(available_columns)
+        )
+        if from_error is not None:
+            raise error from from_error
+        raise error
```
dangotbanned, line 1552: https://github.com/narwhals-dev/narwhals/pull/2495#discussion_r2072933503

> I had to add this otherwise the typechecker complains.
> 
> Can you think of a cleaner/better way?

@EdAbati one problem is that calling `check_columns_exist` gives us no information statically - since it returns `None` (but also `Never`, depending on internal logic)
If you rewrote it such that is returned `ColumnNotFoundError | None`, then on the outside you could do:

```py
if err := check_columns_exist(...):
    raise err
# continue the happy path
```

EdAbati:
Removed the dependency on #2493 , this is now ready for review :)

dangotbanned:
@EdAbati I've been experimenting with #2534 a bit locally and realised some of the changes I was making to `drop`, `parse_columns_to_drop` might make sense to use here.

- (https://github.com/narwhals-dev/narwhals/pull/2495/commits/4d00461886b72b636e67ff781f680e4ee9c5f2d2)

I was finding it difficult to make those as suggestions, so I hope collecting them into 1 commit makes it easier to revert if you prefer what was there before 🙂

I made some notes on the commit description as well

EdAbati:
Thanks for the update!

I think most of thechanges looks great to me.

I'm not 100% sure about the signature of `check_columns_exist`, I personally tend to prefer more explicit kwargs. And IMO having the 1st arg that could be either a frame or a list of cols could be a bit less straightforward to read if one doesn't look at the signature. My vote is on something like:

```python
def check_columns_exist(columns: Sequence[str], /, *, available: Sequence[str]) -> ColumnNotFoundError | None:
    ...
```

When used it will become:

```diff
- check_columns_exist(self, subset)
+ check_columns_exist(subset, available=self.columns)

- check_columns_exist(frame, to_drop)
+ check_columns_exist(to_drop, available=frame.columns)

check_columns_exist(
-     df.columns.tolist(),  # type: ignore[attr-defined]
-     column_names,  # type: ignore[arg-type]
+     column_names,
+     available=df.columns.tolist(),
)
```

But this is highly subjective, maybe @FBruzzesi or @MarcoGorelli can break the tie :)

On a similar note, don't you think forcing positional args in `ColumnNotFoundError.from_missing_and_available_column_names` could be a bit risky? One (if doesn't use `check_columns_exist`) can accidentally write `ColumnNotFoundError.from_missing_and_available_column_names(available, missing)` instead of `ColumnNotFoundError.from_missing_and_available_column_names(missing, available)`

Anyway all minor points. If the manjority likes this version let's go for it

FBruzzesi:
Hey everyone! I didn't go through the detailed history of what happened here 

> I'm not 100% sure about the signature of `check_columns_exist`, I personally tend to prefer more explicit kwargs. And IMO having the 1st arg that could be either a frame or a list of cols could be a bit less straightforward to read if one doesn't look at the signature. My vote is on something like:
> 
> ```python
> def check_columns_exist(columns: Sequence[str], /, *, available: Sequence[str]) -> ColumnNotFoundError | None:
>     ...
> ```

I strongly agree with this statement: considering how easy it would be to swap the two arguments I would prefer to have them as keyword only. And passing column names directly instead of the entire dataframe would make it more generic and re-usable.

> On a similar note, don't you think forcing positional args in `ColumnNotFoundError.from_missing_and_available_column_names` could be a bit risky? One (if doesn't use `check_columns_exist`) can accidentally write `ColumnNotFoundError.from_missing_and_available_column_names(available, missing)` instead of `ColumnNotFoundError.from_missing_and_available_column_names(missing, available)`

Same consideration - I rather have them keyword only to be explicit in this case

dangotbanned:
Hey @EdAbati, @FBruzzesi 

You'd be surprised how many thoughts your comments sparked for me 😄 

As I mentioned in (https://github.com/narwhals-dev/narwhals/pull/2495#issuecomment-2876752218), these changes are kind of a merging of ideas. 
I now have some extra context with these 3 commits:
- [#2547-9](https://github.com/narwhals-dev/narwhals/pull/2547/commits/494419c29489cd00f8395da5efe68eb72dce9fb5)
- [#2547-10](https://github.com/narwhals-dev/narwhals/pull/2547/commits/5f367a50d6e5c12c2bf35aae0322ebb44099d9d4)
- [#2547-11](https://github.com/narwhals-dev/narwhals/pull/2547/commits/ecc5c46666083e314e76d8de09a8fed33cfa1228)

Just wanted to check in and say **I hear you** 🙂 

Hopefully will be able to collect my thoughts *properly* later today.
But the main part is about encoding more into types/classes/methods vs utility functions that pass around state and have to be repeated in every implementation (which is where the concerns that lead to keyword-only args come from IMO)

MarcoGorelli:
i haven't checked everything but - so long as `.columns` doesn't called in cases where the code could pass without it - then I'm happy to trust/defer to you all for this

FBruzzesi:
Hey @dangotbanned thanks for the thoughtful reasoning walkthrough. I agree with most of the points you made and will try to add my two cents.

The reason for suggesting a verbose approach is that what we have is a utility function, which is much more generic than the use case - the answer we get is from the following question "is X a subset of Y?" - so to me it was important to make sure who is X and who is Y.

Regarding `ColumnNotFoundError.from_missing_and_available_column_names` - ok that's fair to avoid repeating ourselves since the order is already in the method name.

Now to the most important piece: "What do now?" - yes I really like that approach, and by moving from utility function to a method, it's much clearer who is what - so thank for the clarity and the suggestion, I think it's really great

dangotbanned:
Thanks @FBruzzesi 😍 (https://github.com/narwhals-dev/narwhals/pull/2495#issuecomment-2888401197)

> Now to the most important piece: "What do now?" - yes I really like that approach, and by moving from utility function to a method, it's much clearer who is what - so thank for the clarity and the suggestion, I think it's really great


Reading this back now I realised I waffled **quite a lot** and missed the point I wanted to land on 😭 

What I proposed in (https://github.com/narwhals-dev/narwhals/pull/2495#issuecomment-2884098316) was what I had in mind *while writing* (https://github.com/narwhals-dev/narwhals/pull/2495/commits/4d00461886b72b636e67ff781f680e4ee9c5f2d2)

> [!IMPORTANT]
> I didn't mean we need to do any of that in this PR


That was just the direction I wanted to move in *afterwards*.

I'm happy with the PR as-is - or we can revert some parts - but (in my mind) either way is a stepping stone to (*What do now?*) 🙂

dangotbanned (approved):

FBruzzesi:
> I didn't mean we need to do any of that in this PR
>
> That was just the direction I wanted to move in _afterwards_.

On this regard - I think it's worth addressing it here at this point since the discussion started and the idea is quite clear. 
Currently we are in a situation for which:

- The codebase is in a transitional situation (according to Dan)
- It's not verbose/clear enough (according to Edo and I)

I will try to speed this up today

narwhals/_compliant/dataframe.py
```diff
@@ -270,6 +272,9 @@ def _evaluate_aliases(self, *exprs: CompliantExprT_contra) -> list[str]:
         it = (expr._evaluate_aliases(self) for expr in exprs)
         return list(chain.from_iterable(it))
 
+    def _check_columns_exist(self, subset: Sequence[str]) -> ColumnNotFoundError | None:
+        return check_columns_exist(subset, available=list(self.columns))
```
EdAbati, line 276: added this method at compliant level as @dangotbanned suggested.

I like it way more now

narwhals/utils.py
```diff
@@ -1544,10 +1537,14 @@ def generate_repr(header: str, native_repr: str) -> str:
     )
 
 
-def check_column_exists(columns: Sequence[str], subset: Sequence[str] | None) -> None:
-    if subset is not None and (missing := set(subset).difference(columns)):
-        msg = f"Column(s) {sorted(missing)} not found in {columns}"
-        raise ColumnNotFoundError(msg)
+def check_columns_exist(
+    subset: Sequence[str], /, *, available: list[str]
+) -> ColumnNotFoundError | None:
```
EdAbati, line 1542: I kept also this function for the exeptions like `narwhals/_pandas_like/utils.py`
(Plus a function is easier to test then the method.)

Changed slightly the signature, using one short `kwargs`. 😇 a middle ground between the previous 2 solutions

narwhals/_spark_like/dataframe.py
EdAbati, line 1: to answer at @MarcoGorelli 's comment

> so long as .columns doesn't called in cases where the code could pass without it

This PR doesn't introduce additional calls to `.columns` but here and in `duckdb` we already called `.column` to check the input in `unique` and in `drop`

If they are not needed they should be removed in a separate PR. Hopefully they are easier to spot  now?

EdAbati (commented):
Ok this was suuuuper delayed. :( I implemented some of the suggestions above.

Let me know what you think and if in general you thinkthis PR improves things compared to before :D

narwhals/utils.py
dangotbanned: > Changed slightly the signature, using one short `kwargs`. 😇 a middle ground between the previous 2 solutions

No qualms from me 😄 

@EdAbati should we punt the rest of (https://github.com/narwhals-dev/narwhals/pull/2495#issuecomment-2888468992) to another issue?

I really wanna stress that **I'm happy with getting these changes merged first**! 😅

narwhals/_compliant/dataframe.py
EdAbati: great point and awesome explanation. updated :)

narwhals/utils.py
EdAbati: Happy to merge now if everyone is happy :)

PR #2501, opened 2025-05-06
test: make `polars` optional in `truncate_test`

mgorny:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
I'll try adding a workflow to test this use case (separately).

PR #2502, opened 2025-05-06
test: Add a GHA workflow testing against narrower dep set

mgorny:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Related to #2501, this is meant to help detect regressions like that. As expected, it fails with `main` but passes once #2501 is merged (e.g. https://github.com/mgorny/narwhals/actions/runs/14852314516/job/41698095253).

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -99,3 +99,28 @@ jobs:
         # reprs differ between versions, so we only run doctests on the latest Python
         if: matrix.python-version == '3.13'
         run: pytest narwhals/*.py --doctest-modules
+
+  # Test against smaller dependency set, used e.g. on Gentoo.
+  pytest-narrower-deps:
+    strategy:
+      matrix:
+        python-version: ["3.13"]
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: "true"
+          cache-suffix: ${{ matrix.python-version }}
+          cache-dependency-glob: "pyproject.toml"
+      - name: install-reqs
+        run: uv pip install -e ".[pandas, pyarrow]" --group tests --system
+      - name: show-deps
+        run: uv pip freeze
+      - name: Run pytest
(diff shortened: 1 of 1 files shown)


FBruzzesi (approved):
I am definitly ok with this. It's extremely fast to run and will safeguard to introduce top level imports aside from pandas and pyarrow in our test suite.

Thanks @mgorny ✨

NarwhalChubbiwhalGIF

.github/workflows/pytest.yml
```diff
@@ -99,3 +99,28 @@ jobs:
         # reprs differ between versions, so we only run doctests on the latest Python
         if: matrix.python-version == '3.13'
         run: pytest narwhals/*.py --doctest-modules
+
+  # Test against smaller dependency set, used e.g. on Gentoo.
+  pytest-narrower-deps:
```
dangotbanned, line 104: I'm not sure on the naming (or an alternative), but some thoughts 😅 

## Notes
- *deps* -> *dependencies*?
- Maybe a group **like** *Test Downstream Libraries*?
  - If you see a need for additional workflows for `Gentoo`, then *Test Downstream Packagers*?

dangotbanned (approved):
Thanks @mgorny, +1 to what @FBruzzesi said (https://github.com/narwhals-dev/narwhals/pull/2502#pullrequestreview-2817097906)


Just have one **non-blocking** comment

.github/workflows/pytest.yml
mgorny: Any name is fine with me, though I think "downstream libraries" would be confusing, since it's only testing narwhals themselves.

dangotbanned: > Maybe a group **like** Test Downstream Libraries?

@mgorny I guess my emphasis wasn't very helpful 🤦‍♂️ 

The thought that didn't quite escape my head was:

> Maybe a group ~~like~~ **with a name like** *Test Downstream Libraries*?
> If you see a need for additional workflows for `Gentoo`, ~~then~~ a name for that group could be *Test Downstream Packagers*?

mgorny: Ohh, sorry, I didn't realize you were referring to the existing group. To be honest, I don't think we're going to add any more workflows, so not sure if this is worth another group.

dangotbanned: All good - its my bad for being unclear

mgorny:
Thanks!

PR #2511, opened 2025-05-08
chore: Simplify `PandasLikeDataFrame|DaskLazyFrame.join` method

### LINKED ISSUE #2488: Simplify `*DataFrame.join`
https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_dask/dataframe.py#L261-L269

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_pandas_like/dataframe.py#L577-L585

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2488

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -326,7 +326,7 @@ class CompliantLazyFrame(
         self,
         other: Self,
         *,
-        how: Literal["left", "inner", "cross", "anti", "semi"],
+        how: JoinStrategy,
         left_on: Sequence[str] | None,
         right_on: Sequence[str] | None,
         suffix: str,
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -258,7 +258,161 @@ class DaskLazyFrame(
             self.native.sort_values(list(by), ascending=ascending, na_position=position)
         )
 
-    def join(  # noqa: C901
+    def _inner_join(
+        self,
+        other: Self,
+        *,
+        left_on: Sequence[str] | None,
+        right_on: Sequence[str] | None,
+        suffix: str,
+    ) -> Self:
+        return self._with_native(
+            self.native.merge(
+                other.native,
+                left_on=left_on,
+                right_on=right_on,
+                how="inner",
+                suffixes=("", suffix),
+            )
+        )
+
+    def _left_join(
+        self,
+        other: Self,
+        *,
+        left_on: Sequence[str] | None,
+        right_on: Sequence[str] | None,
+        suffix: str,
+    ) -> Self:
+        result_native = self.native.merge(
+            other.native,
+            how="left",
+            left_on=left_on,
+            right_on=right_on,
+            suffixes=("", suffix),
+        )
+        extra = []
+        for left_key, right_key in zip(left_on, right_on):  # type: ignore[arg-type]
+            if right_key != left_key and right_key not in self.columns:
+                extra.append(right_key)
+            elif right_key != left_key:
+                extra.append(f"{right_key}_right")
+        return self._with_native(result_native.drop(columns=extra))
+
+    def _full_join(
+        self,
+        other: Self,
+        *,
+        left_on: Sequence[str],
+        right_on: Sequence[str],
+        suffix: str,
+    ) -> Self:
+        # dask does not retain keys post-join
+        # we must append the suffix to each key before-hand
+
+        right_on_mapper = _remap_full_join_keys(left_on, right_on, suffix)
+        other_native = other.native.rename(columns=right_on_mapper)
+        check_column_names_are_unique(other_native.columns)
+        right_on = list(right_on_mapper.values())  # we now have the suffixed keys
+        return self._with_native(
+            self.native.merge(
+                other_native,
+                left_on=left_on,
+                right_on=right_on,
+                how="outer",
+                suffixes=("", suffix),
+            )
+        )
+
+    def _cross_join(
+        self,
+        other: Self,
+        *,
+        suffix: str,
+    ) -> Self:
+        key_token = generate_temporary_column_name(
+            n_bytes=8, columns=[*self.columns, *other.columns]
+        )
+
+        return self._with_native(
+            self.native.assign(**{key_token: 0})
+            .merge(
+                other.native.assign(**{key_token: 0}),
+                how="inner",
+                left_on=key_token,
+                right_on=key_token,
+                suffixes=("", suffix),
+            )
+            .drop(columns=key_token)
+        )
+
+    def _semi_join(
+        self,
+        other: Self,
+        *,
+        left_on: Sequence[str] | None,
+        right_on: Sequence[str] | None,
+    ) -> Self:
+        if right_on is None:  # pragma: no cover
+            msg = "`right_on` cannot be `None` in semi-join"
+            raise TypeError(msg)
+        other_native = (
+            select_columns_by_name(
+                other.native,
+                list(right_on),
+                self._backend_version,
+                self._implementation,
+            )
+            .rename(  # rename to avoid creating extra columns in join
+                columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
+            )
+            .drop_duplicates()  # avoids potential rows duplication from inner join
+        )
+        return self._with_native(
+            self.native.merge(
+                other_native, how="inner", left_on=left_on, right_on=left_on
+            )
+        )
+
+    def _anti_join(
+        self,
+        other: Self,
+        *,
+        left_on: Sequence[str] | None,
+        right_on: Sequence[str] | None,
+    ) -> Self:
+        if right_on is None:  # pragma: no cover
+            msg = "`right_on` cannot be `None` in anti-join"
+            raise TypeError(msg)
+
+        indicator_token = generate_temporary_column_name(
+            n_bytes=8, columns=[*self.columns, *other.columns]
+        )
+
+        other_native = (
+            select_columns_by_name(
+                other.native,
+                column_names=list(right_on),
+                backend_version=self._backend_version,
+                implementation=self._implementation,
+            )
+            .rename(  # rename to avoid creating extra columns in join
+                columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
+            )
+            .drop_duplicates()
+        )
+        df = self.native.merge(
+            other_native,
+            how="outer",
+            indicator=indicator_token,  # pyright: ignore[reportArgumentType]
+            left_on=left_on,
+            right_on=left_on,
+        )
+        return self._with_native(
+            df[df[indicator_token] == "left_only"].drop(columns=[indicator_token])
+        )
+
+    def join(
         self,
         other: Self,
         *,
@@ -267,123 +421,36 @@ class DaskLazyFrame(
         right_on: Sequence[str] | None,
         suffix: str,
     ) -> Self:
-        if how == "cross":
-            key_token = generate_temporary_column_name(
-                n_bytes=8, columns=[*self.columns, *other.columns]
+        if how == "inner":
+            return self._inner_join(
+                other=other, left_on=left_on, right_on=right_on, suffix=suffix
             )
 
-            return self._with_native(
-                self.native.assign(**{key_token: 0})
-                .merge(
-                    other.native.assign(**{key_token: 0}),
-                    how="inner",
-                    left_on=key_token,
-                    right_on=key_token,
-                    suffixes=("", suffix),
-                )
-                .drop(columns=key_token)
+        elif how == "cross":
+            return self._cross_join(other=other, suffix=suffix)
+
+        elif how == "anti":
+            return self._anti_join(other=other, left_on=left_on, right_on=right_on)
+
+        elif how == "semi":
+            return self._semi_join(other=other, left_on=left_on, right_on=right_on)
+
+        elif how == "left":
+            return self._left_join(
+                other=other, left_on=left_on, right_on=right_on, suffix=suffix
             )
 
-        if how == "anti":
-            indicator_token = generate_temporary_column_name(
-                n_bytes=8, columns=[*self.columns, *other.columns]
-            )
-
-            if right_on is None:  # pragma: no cover
-                msg = "`right_on` cannot be `None` in anti-join"
-                raise TypeError(msg)
-            other_native = (
-                select_columns_by_name(
-                    other.native,
-                    list(right_on),
-                    self._backend_version,
-                    self._implementation,
-                )
-                .rename(  # rename to avoid creating extra columns in join
-                    columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
-                )
-                .drop_duplicates()
-            )
-            df = self.native.merge(
-                other_native,
-                how="outer",
-                indicator=indicator_token,  # pyright: ignore[reportArgumentType]
-                left_on=left_on,
-                right_on=left_on,
-            )
-            return self._with_native(
-                df[df[indicator_token] == "left_only"].drop(columns=[indicator_token])
-            )
-
-        if how == "semi":
-            if right_on is None:  # pragma: no cover
-                msg = "`right_on` cannot be `None` in semi-join"
-                raise TypeError(msg)
-            other_native = (
-                select_columns_by_name(
-                    other.native,
-                    list(right_on),
-                    self._backend_version,
-                    self._implementation,
-                )
-                .rename(  # rename to avoid creating extra columns in join
-                    columns=dict(zip(right_on, left_on))  # type: ignore[arg-type]
-                )
-                .drop_duplicates()  # avoids potential rows duplication from inner join
-            )
-            return self._with_native(
-                self.native.merge(
-                    other_native, how="inner", left_on=left_on, right_on=left_on
-                )
-            )
-
-        if how == "left":
-            result_native = self.native.merge(
-                other.native,
-                how="left",
-                left_on=left_on,
-                right_on=right_on,
-                suffixes=("", suffix),
-            )
-            extra = []
-            for left_key, right_key in zip(left_on, right_on):  # type: ignore[arg-type]
-                if right_key != left_key and right_key not in self.columns:
-                    extra.append(right_key)
-                elif right_key != left_key:
-                    extra.append(f"{right_key}_right")
-            return self._with_native(result_native.drop(columns=extra))
-
-        if how == "full":
-            # dask does not retain keys post-join
-            # we must append the suffix to each key before-hand
-
+        elif how == "full":
             # help mypy
             assert left_on is not None  # noqa: S101
             assert right_on is not None  # noqa: S101
 
-            right_on_mapper = _remap_full_join_keys(left_on, right_on, suffix)
-            other_native = other.native.rename(columns=right_on_mapper)
-            check_column_names_are_unique(other_native.columns)
-            right_on = list(right_on_mapper.values())  # we now have the suffixed keys
-            return self._with_native(
-                self.native.merge(
-                    other_native,
-                    left_on=left_on,
-                    right_on=right_on,
-                    how="outer",
-                    suffixes=("", suffix),
-                )
+            return self._full_join(
+                other=other, left_on=left_on, right_on=right_on, suffix=suffix
             )
-
-        return self._with_native(
-            self.native.merge(
-                other.native,
-                left_on=left_on,
-                right_on=right_on,
-                how=how,
-                suffixes=("", suffix),
-            )
-        )
+        else:  # pragma: no cover
+            msg = f"Unreachable code, got unexpected join method: {how}"
+            raise AssertionError(msg)
 
     def join_asof(
         self,
--- a/narwhals/_dask/dataframe.py
+++ b/narwhals/_dask/dataframe.py
@@ -259,12 +259,7 @@ class DaskLazyFrame(
         )
 
     def _inner_join(
-        self,
-        other: Self,
-        *,
-        left_on: Sequence[str] | None,
-        right_on: Sequence[str] | None,
-        suffix: str,
+        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str], suffix: str
     ) -> Self:
         return self._with_native(
             self.native.merge(
@@ -277,12 +272,7 @@ class DaskLazyFrame(
         )
 
     def _left_join(
-        self,
-        other: Self,
-        *,
-        left_on: Sequence[str] | None,
-        right_on: Sequence[str] | None,
-        suffix: str,
+        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str], suffix: str
     ) -> Self:
         result_native = self.native.merge(
             other.native,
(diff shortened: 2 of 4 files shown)


dangotbanned:
@FBruzzesi would you be open to flipping the method names around?

Most of the private ones that are related **share a common prefix**.
That helps them show up together in autocomplete 🙂 

I do understand why e.g (`anti_join`, `left_join`) make a lot of sense in isolation, but another way to think about it is:

```py
DataFrame.join(how="anti") -> CompliantDataFrame._join_anti
DataFrame.join(how="left") -> CompliantDataFrame._join_left
```

### `DaskLazyFrame`

image

### `PandasLikeDataFrame`
image

FBruzzesi:
@dangotbanned moved naming to `_join_<method>`

dangotbanned:
> @dangotbanned moved naming to `_join_<method>`

Thanks @FBruzzesi will try to do a proper review today 🙏

narwhals/dataframe.py
```diff
@@ -237,75 +237,66 @@ def sort(
...
+                other, how=how, left_on=left_on, right_on=right_on, suffix=suffix
+            )
+        else:
+            if left_on is not None or right_on is not None:
+                msg = f"If `on` is specified, `left_on` and `right_on` should be None for {how}."
+                raise ValueError(msg)
+            result = compliant.join(
+                other, how=how, left_on=on, right_on=on, suffix=suffix
+            )
+        return self._with_compliant(result)
```
dangotbanned, line 281: We've now got 3 distinct paths, which come from:

1. `how: Literal['cross'], on: None, left_on: None, right_on: None`
2. `how: Literal['inner', 'left', 'full', 'semi', 'anti'], on: None, left_on: list[str], right_on: list[str]`
3. `how: Literal['inner', 'left', 'full', 'semi', 'anti'], on: list[str], left_on: None, right_on: None`

Maybe we can utilize this to avoid the double-dipping like in (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2083106551)?

dangotbanned: ## Edit
I need to remember my own advice 🤦‍♂️ 

me yesterday on discord 😄 
> Oh I see!
> So a few different ways to solve:
> 
> 1. Mutually exclusive `@overload`s
>   i. So you'd specify `None` is only allowed for both `*_on` when `how="cross"`
>   ii. However these can be tricky to get right, especially when there's a lot of parameters
> 2. Raise an error instead of asserting when an invariant is broken (https://github.com/narwhals-dev/narwhals/pull/2000/commits/529c196549c77f093c8a77a61e9ff26e9511eb66)
>   i. This is the **easiest solution**, but often *feels* like you're doing something unnecessary
> 3. Split incompatible signatures into distinct methods/functions
>   i. This is **my preferred solution** and what I'm usually hinting towards when I suggest adding multiple `@classmethod`s

narwhals/_pandas_like/dataframe.py
```diff
@@ -576,153 +576,185 @@ def group_by(
...
+        right_on = list(right_on_mapper.values())  # we now have the suffixed keys
+        return self._with_native(
+            self.native.merge(
+                other_native,
+                left_on=left_on,
+                right_on=right_on,
+                how="outer",
+                suffixes=("", suffix),
+            ),
+        )
```
dangotbanned, line 626:
```suggestion
        right_suffixed = list(right_on_mapper.values())
        return self._with_native(
            self.native.merge(
                other_native,
                left_on=left_on,
                right_on=right_suffixed,
                how="outer",
                suffixes=("", suffix),
            )
        )
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -576,153 +576,185 @@ def group_by(
...
+    def _join_cross(self, other: Self, *, suffix: str) -> Self:
+        implementation = self._implementation
+        backend_version = self._backend_version
+        if (implementation.is_modin() or implementation.is_cudf()) or (
+            implementation.is_pandas() and backend_version < (1, 4)
+        ):
+            key_token = generate_temporary_column_name(
+                n_bytes=8, columns=[*self.columns, *other.columns]
             )
+
```
dangotbanned, line 637:
```suggestion
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -576,153 +576,185 @@ def group_by(
...
+                self.native.assign(**{key_token: 0})
+                .merge(
+                    other.native.assign(**{key_token: 0}),
+                    how="inner",
+                    left_on=key_token,
+                    right_on=key_token,
+                    suffixes=("", suffix),
                 )
+                .drop(columns=key_token)
             )
```
dangotbanned, line 649:
```suggestion
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -576,153 +576,185 @@ def group_by(
...
+            ),
+            columns=dict(zip(right_on, left_on)),
+            implementation=implementation,
+            backend_version=backend_version,
+        ).drop_duplicates()
+
         return self._with_native(
             self.native.merge(
-                other.native,
+                other_native,
```
dangotbanned, line 712:
```suggestion
    def _join_filter_rename(
        self, other: Self, left_on: Sequence[str], right_on: Sequence[str]
    ) -> pd.DataFrame:
        """Rename to avoid creating extra columns in `"anti"`, `"semi"` join, and avoids potential rows duplication."""
        implementation = self._implementation
        backend_version = self._backend_version
        return rename(
            select_columns_by_name(
                other.native, list(right_on), backend_version, implementation
            ),
            columns=dict(zip(right_on, left_on)),
            implementation=implementation,
            backend_version=backend_version,
        ).drop_duplicates()

    def _join_semi(
        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str]
    ) -> Self:
        other_native = self._join_filter_rename(other, left_on, right_on)
        return self._with_native(
            self.native.merge(
                other_native, how="inner", left_on=left_on, right_on=left_on
            )
        )

    def _join_anti(
        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str]
    ) -> Self:
        if self._implementation.is_cudf():
            return self._with_native(
                self.native.merge(
                    other.native, how="leftanti", left_on=left_on, right_on=right_on
                )
            )
        indicator_token = generate_temporary_column_name(
            n_bytes=8, columns=[*self.columns, *other.columns]
        )
        return self._with_native(
            self.native.merge(
                self._join_filter_rename(other, left_on, right_on),
```

narwhals/_pandas_like/dataframe.py
```diff
@@ -576,153 +576,185 @@ def group_by(
...
+            left_on=left_on,
+            right_on=right_on,
+            suffixes=("", suffix),
+        )
+        extra = []
+        for left_key, right_key in zip(left_on, right_on):
+            if right_key != left_key and right_key not in self.columns:
+                extra.append(right_key)
+            elif right_key != left_key:
+                extra.append(f"{right_key}{suffix}")
```
dangotbanned, line 607: Might there be a way to do this using `set` operations?

We don't need the order in the `drop(columns=...)` call

dangotbanned (commented):
Thanks @FBruzzesi - looking a lot better! 🥳 

I've got a few suggestions on the `pandas` one, but I think the same would apply for `dask` as well

narwhals/_pandas_like/dataframe.py
dangotbanned: What on earth is a `join_filter` I hear you ask? 😉 

The last two are not like the others, which is why they require one less parameter

https://github.com/narwhals-dev/narwhals/blob/dd41833963acaf2641c6c4c8948ee7aa5f25734a/narwhals/typing.py#L239-L250

FBruzzesi:
> Thanks @FBruzzesi - looking a lot better! 🥳
> 
> I've got a few suggestions on the `pandas` one, but I think the same would apply for `dask` as well

Thanks @dangotbanned - great work you've done!
I will have time later today to go through each point

narwhals/dataframe.py
dangotbanned: A follow-up could look at option 3, which would be making some `CompliantJoin` class, then having:
- `CompliantJoin.left`
- `CompliantJoin.inner`
- ...

Would probably be easier to share code between `pandas` and `dask` that way as well 🙂

dangotbanned:
I was thinking about simplifying `BaseFrame.join_asof` in a similar way to (https://github.com/narwhals-dev/narwhals/pull/2511/commits/1c1daae80ba81597ce117fa8e055d0053354991a)

I thought I'd take a look at the `polars` version(S).

They're simpler, *and* accept more parameters and types 🤔 

- https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/dataframe/frame.py#L7614-L7630
- https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/lazyframe/frame.py#L5283-L5319

Maybe they simplfied it at some point?

narwhals/_dask/dataframe.py
```diff
@@ -258,133 +258,160 @@ def sort(
...
+                backend_version=self._backend_version,
+                implementation=self._implementation,
             )
+            # rename to avoid creating extra columns in join
+            .rename(columns=dict(zip(right_on, left_on)))
+            .drop_duplicates()
+        )
+        df = self.native.merge(
+            other_native,
+            how="outer",
```
FBruzzesi, line 370: Spotted that we are wasted some cycles - this can be a left join (with the same filter below). Fixing that

narwhals/_pandas_like/dataframe.py
FBruzzesi: Solved in https://github.com/narwhals-dev/narwhals/pull/2511/commits/ae64d74df150a27db63a6a3fc9a1e9ceecc832ac

narwhals/_pandas_like/dataframe.py
FBruzzesi: Solved in https://github.com/narwhals-dev/narwhals/pull/2511/commits/a0fca3d5c7bcfaa67c870b7e1c384bafa477ec04

narwhals/dataframe.py
FBruzzesi: > A follow-up could look at option 3, which would be making some `CompliantJoin` class, then having:
> 
> * `CompliantJoin.left`
> * `CompliantJoin.inner`
> * ...
> 
> Maybe the calls from `BaseFrame` would look like:
> 
> ```python
> compliant.join(**kwds).left()
> compliant.join(**kwds).inner()
> ```
> 
> Or just have the one `join`, which makes the other calls internally, to keep it simple from the outside:
> 
> ```python
> compliant.join(**kwds)
> self.left()
> ```
> 
> Would probably be easier to share code between `pandas` and `dask` that way as well 🙂

This seems like an interesting idea but I am not sure I get it right and/or I would end up with the expected changes.

Especially:

Where would `CompliantJoin` exist and how would it interact with other compliant classes?

and 

> Maybe the calls from `BaseFrame` would look like:
> 
> ```python
> compliant.join(**kwds).left()
> compliant.join(**kwds).inner()
> ```

Would this require some special path for polars?

dangotbanned: @FBruzzesi you could think of it like these:
- `CompliantGroupBy`
- `CompliantWhen`
- #2393

Or similar to how the `Expr`, `Series` namespace classes work.

Essentially, we define some common parts and then refine from there per-backend

> Would this require some special path for polars?

Yeah most likely, `polars` gets to skip our abstractions in a lot of places and this would be one of those cases I'd assume.

Anyways, this is very much just the kernel of an idea 😅
The goal would be sharing more code between backends 😎

FBruzzesi: Should we... keep that as a follow up then? 😅

> The goal would be sharing more code between backends 😎

Yep this is much loud and clear! Pandas-like and Dask can definitly share some functionalities eventually :)

dangotbanned: > Should we... keep that as a follow up then? 😅

Yeah for sure!

I started with that in (https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2083257048) but clearly got carried away with my rambling 😄

narwhals/_pandas_like/dataframe.py
```diff
@@ -563,153 +563,185 @@ def group_by(
...
-                right_on=right_on,
-                suffixes=("", suffix),
+    def _join_semi(
+        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str]
+    ) -> Self:
+        other_native = self._join_filter_rename(
+            other=other,
+            columns_to_select=list(right_on),
+            columns_mapping=dict(zip(right_on, left_on)),
+        )
```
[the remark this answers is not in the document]
FBruzzesi: @dangotbanned please forgive me - I didn't get your point here 🥲

narwhals/_pandas_like/dataframe.py
```diff
@@ -563,153 +563,185 @@ def group_by(
...
+                right_on=left_on,
             )
+            .loc[lambda t: t[indicator_token] == "left_only"]
+            .drop(columns=indicator_token)
         )
 
+    def _join_filter_rename(
+        self, other: Self, columns_to_select: list[str], columns_mapping: dict[str, str]
+    ) -> pd.DataFrame:
+        """Helper function to avoid creating extra columns and row duplication.
```
dangotbanned, line 689: @FBruzzesi in (https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2105813874) I'm saying that this version is more duplicated than what I had in the original suggestion, which was: 

```py
    def _join_filter_rename(
        self, other: Self, left_on: Sequence[str], right_on: Sequence[str]
    ) -> pd.DataFrame:
        """Rename to avoid creating extra columns in `"anti"`, `"semi"` join, and avoids potential rows duplication."""
        implementation = self._implementation
        backend_version = self._backend_version
        return rename(
            select_columns_by_name(
                other.native, list(right_on), backend_version, implementation
            ),
            columns=dict(zip(right_on, left_on)),
            implementation=implementation,
            backend_version=backend_version,
        ).drop_duplicates()
```

The `list(right_on)` and `dict(zip(right_on, left_on))` parts are defined in one place - rather than repeated each time you call `_join_filter_rename`

FBruzzesi: Oh I see, it has been a while since this PR, I didn't spot it in the other comment.
I see what you mean, it a similar conundrum to #2495, the signature with `columns_to_select` and `columns_mapping` in my opinion is much more intuitive to understand what happens within the method without jumping into it, on the other hand, yes, we end up creating and passing these objects externally of the method itself.
I really don't know 🤔

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2511#event-18343540555

@FBruzzesi you may have noticed I remembered about this PR in (https://github.com/narwhals-dev/narwhals/pull/2511#ref-pullrequest-3165108714) and then quickly lost track 😅 

Feel free to ping me if I forget to review tomorrow!

narwhals/_pandas_like/dataframe.py
```diff
@@ -557,153 +557,185 @@ def group_by(
...
+        other_native = self._join_filter_rename(
+            other=other,
+            columns_to_select=list(right_on),
+            columns_mapping=dict(zip(right_on, left_on)),
+        )
         return self._with_native(
             self.native.merge(
-                other.native,
+                other_native,
+                # TODO(FBruzzesi): Raise issue upstream for Modin
```
dangotbanned: What was the problem for `modin`?

narwhals/_pandas_like/dataframe.py
```diff
@@ -557,153 +557,185 @@ def group_by(
...
-                    .loc[lambda t: t[indicator_token] == "left_only"]
-                    .drop(columns=indicator_token)
-                )
+    def _join_full(
+        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str], suffix: str
+    ) -> Self:
+        # Pandas coalesces keys in full joins unless there's no collision
+        right_on_mapper = _remap_full_join_keys(left_on, right_on, suffix)
+        other_native = other.native.rename(columns=right_on_mapper)
+        check_column_names_are_unique(other_native.columns)
```
dangotbanned, line 595: Huh, I've not noticed this before, but we have two version of this function

This version is https://github.com/narwhals-dev/narwhals/blob/3aba9bbb933baa584653b6f70cc1084eda2ed378/narwhals/_pandas_like/utils.py#L595-L611


But I thought it was https://github.com/narwhals-dev/narwhals/blob/3aba9bbb933baa584653b6f70cc1084eda2ed378/narwhals/_utils.py#L1575-L1584

Considering that we don't have coverage for the main difference between the two - what do you think about removing the `_pandas_like` one?

AFAIK, `pandas` only supports `Hashable` column *"names"* - so I'm a little confused 🤔

narwhals/_compliant/dataframe.py
```diff
@@ -331,7 +331,7 @@ def join(
         self,
         other: Self,
         *,
-        how: Literal["left", "inner", "cross", "anti", "semi"],
+        how: JoinStrategy,
```
dangotbanned, line 334: ooh well spotted!

I'm not sure why there haven't been any typing issues reported about this before/since changing in this PR.
But nice to see we support all the same strategies 😄

narwhals/_dask/dataframe.py
```diff
@@ -271,7 +271,131 @@ def sort(self, *by: str, descending: bool | Sequence[bool], nulls_last: bool) ->
...
+
+    def _join_full(
+        self, other: Self, *, left_on: Sequence[str], right_on: Sequence[str], suffix: str
+    ) -> Self:
+        # dask does not retain keys post-join
+        # we must append the suffix to each key before-hand
+
+        right_on_mapper = _remap_full_join_keys(left_on, right_on, suffix)
+        other_native = other.native.rename(columns=right_on_mapper)
+        check_column_names_are_unique(other_native.columns)
```
dangotbanned, line 311: This version is `nw._utils.check_column_names_are_unique` btw

narwhals/_pandas_like/dataframe.py
```diff
@@ -557,153 +557,185 @@ def group_by(
...
 
+    def _join_filter_rename(
+        self, other: Self, columns_to_select: list[str], columns_mapping: dict[str, str]
+    ) -> pd.DataFrame:
+        """Helper function to avoid creating extra columns and row duplication.
+
+        Used in `"anti"` and `"semi`" join's.
+
+        Notice that a native object is returned.
+        """
```
dangotbanned, line 679: What if we returned a native object from all of the `_join*` methods instead?

Then we just do one `_with_native` in `join()`

`concat` and `_concat*` use that pattern btw

https://github.com/narwhals-dev/narwhals/blob/3aba9bbb933baa584653b6f70cc1084eda2ed378/narwhals/_compliant/namespace.py#L181-L198

dangotbanned:
*Mostly finished on the review, just need to add a comment* 

Long awaited review will be with you shortly 😉

dangotbanned (changes requested):
Thanks for the PR @FBruzzesi (*and bearing with me* 😄)

I've got a couple of new notes, which are what the *Request changes* are related to.

Regarding (https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2105922818), I still have a preference for my earlier suggestion.
*But*, **both** are improvements over what we have on `main` - so feel free to keep that part as-is 🙂

narwhals/_pandas_like/dataframe.py
FBruzzesi: Interesting (and very good) point. Let me check (possibly on a separate branch)

narwhals/_pandas_like/dataframe.py
FBruzzesi: Sure, I don't have a strong preference, as it's easy and lightweight to pass between one to another

narwhals/_pandas_like/dataframe.py
FBruzzesi: **repro.py**

```py
import modin.pandas as mpd

data_left = {"a": ["A", "B", "A"], "b": [1, 2, 3]}
data_right = {"a": ["A"], "c": [100]}

df_left = mpd.DataFrame(data_left)
df_right = mpd.DataFrame(data_right)

print("##### outer join #####")
try:
    print(df_left.merge(df_right, how="outer", on="a", indicator="foo"))
except Exception as exc:
    print(exc)

print("##### left join #####")
try:
    print(df_left.merge(df_right, how="left", on="a", indicator="foo"))
except Exception as exc:
    print(exc)
```

```terminal
> export MODIN_ENGINE=python
> python repro.py

##### outer join #####
UserWarning: merge is not currently supported by PandasOnPython, defaulting to pandas implementation.
   a  b      c        foo
0  A  1  100.0       both
1  A  3  100.0       both
2  B  2    NaN  left_only
##### left join #####
Internal Error. Please visit https://github.com/modin-project/modin/issues to file an issue with the traceback and the command that caused this error. If you can't file a GitHub issue, please email bug_reports@modin.org.
Internal and external indices on axis 1 do not match.
```

narwhals/_pandas_like/dataframe.py
FBruzzesi: Ok just kidding, getting headaches with the type checker (rightfully so) because of:

```diff
if how == "cross":
-     return self._join_cross(other=other, suffix=suffix)
+     result = self._join_cross(other=other, suffix=suffix)

if left_on is None or right_on is None:  # pragma: no cover
    raise ValueError(left_on, right_on)
```
now the second condition is always checked

dangotbanned: @FBruzzesi sorry!!!!! I edited your comment instead of replying by mistake 😬 

How about this (like in `concat` 😉 )
```diff
if how == "cross":
-     return self._join_cross(other=other, suffix=suffix)
+     result = self._join_cross(other=other, suffix=suffix)
- if left_on is None or right_on is None:  # pragma: no cover
+ elif left_on is None or right_on is None:  # pragma: no cover
    raise ValueError(left_on, right_on)
```

narwhals/_pandas_like/dataframe.py
dangotbanned: Thanks!

So is it that `how="left"` isn't supported for `modin`?

dangotbanned: Ah https://github.com/modin-project/modin/issues/7384

Nice one thank you ❤️

FBruzzesi: Found and linked a very similar issue: https://github.com/modin-project/modin/issues/7384

narwhals/_pandas_like/dataframe.py
FBruzzesi: 🤦🏼‍♂️🤦🏼‍♂️ of course!

FBruzzesi:
Thanks for all the support @dangotbanned 
I (should) have addressed all your points. Only discussion left is:

> Regarding (https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2105922818), I still have a preference for my earlier suggestion.

but as of:

> But, both are improvements over what we have on main - so feel free to keep that part as-is 🙂

for now yes I would rather keep it as is. My mind has not changed from the [previous comment on it](https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2105926752) 😂

dangotbanned (approved):

PR #2513, opened 2025-05-08
release: Bump version to 1.38.1

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2514, opened 2025-05-08
chore: Bump version manually only in pyproject.toml

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related [comment](https://github.com/narwhals-dev/narwhals/pull/2513#discussion_r2079333983)

## Checklist

- [x] Tests added
- [ ] Documented the changes


FBruzzesi:
I can definitly replicate the issue locally, but I hoped that it was some env issue 🥲
It looks like that pre-commit does not install narwhals when running `check-api-reference` 🤔

FBruzzesi:
> @FBruzzesi I wonder if the new `uv version` command might be helpful for us here?
> 
> * https://docs.astral.sh/uv/reference/cli/#uv-version
> * https://github.com/astral-sh/uv/releases/tag/0.7.0

@dangotbanned this PR was aiming at something else, but we could definitly replace:

https://github.com/narwhals-dev/narwhals/blob/42813b4764541e0db163c3d140c0445899318d5a/utils/bump_version.py#L45-L57

with 

```py
subprocess.run(["uv", "version", "--bump", how])
```

🤩

utils/bump_version.py
```diff
@@ -42,19 +43,8 @@
...
-    version = ".".join(version[:-1] + [str(int(version[-1]) + 1)])
-elif how == "minor":
-    version = ".".join(version[:-2] + [str(int(version[-2]) + 1), "0"])
-elif how == "major":
-    version = ".".join([str(int(version[0]) + 1), "0", "0"])
-content = content.replace(f'version = "{old_version}"', f'version = "{version}"')
-with open("pyproject.toml", "w", encoding="utf-8") as f:
-    f.write(content)
+subprocess.run(["uv", "version", "--bump", how]) 
+version = subprocess.run(["uv", "version", "--short"], capture_output=True).stdout.decode("utf-8")
```
FBruzzesi, line 47: @dangotbanned this is a pretty solid diff if you ask me! Thanks for suggesting it

.pre-commit-config.yaml
FBruzzesi, line 1: So regarding this, I am going a bit insane.
@dangotbanned was `make typing` re-using the env created from `check-api-reference`? I don't really get how that works

dangotbanned: > So regarding this, I am going a bit insane.

dead 🤣

dangotbanned: > @dangotbanned was `make typing` re-using the env created from `check-api-reference`? I am a bit confused on how that works

I honestly have no clue about the `Makefile` 😔, maybe @EdAbati could help?

EdAbati: I have some time to try to reproduce this locally so I'll get back to you with a more informed answer 

But the makefile doesn't create any env. You are supposed to run it in an activated environment.
That's also why it is `repo: local` in the `pre-commit-config`

Which error did you get?

FBruzzesi: I guess the main question is: how was `make typing` able to run in here?
We can add a github action to check for api reference, yet I think it is good to keep it as a pre-commit hook for faster feedback loop

EdAbati: aaah I think I get the question now.

So depending on when you run `pre-commit`, you have 2 different behaviours:
- locally: you have an activated env and `make typing` aka the `mypy` pre-commit step works
- ci: `mypy` from `pre-commit` is skipped https://github.com/narwhals-dev/narwhals/blob/add25453fc35fb374da87cb9d26dafb6c35e86c2/.pre-commit-config.yaml#L3
 we have a separate workflow for [typing](https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/typing.yml)

EdAbati: I don't undestand fully why this change impacts `check-api-reference` hook :D

Why do we need more dependencies now?

FBruzzesi: Ohh I am so dumb 🙈
`importlib.metadata.version` needs `narwhals` to be installed - however 

> pre-commit never installs from the repository under test as it makes caching intractable (it only uses the contents of the configuration and installs in such a way that your repository is intentionally inaccessible)

see [comment](https://github.com/narwhals-dev/narwhals/pull/2514#discussion_r2080141503)

So it makes sense to have it locally and have a dedicated GHA to check in CI - it will be a quick task anyway

dangotbanned: Oh @FBruzzesi, I'm not sure if this applies to *everything* in `importlib` - but something to keep in mind 
- https://github.com/narwhals-dev/narwhals/pull/2391#pullrequestreview-2813442234

dangotbanned: > Mmmh ok, I will rollback to a manual update then using uv version command

@FBruzzesi we might be able to do this lazily actually?
That'd address the performance concern

I feel bad for my comment causing this diff 😢 (https://github.com/narwhals-dev/narwhals/pull/2514/commits/47192aefec80f0385cf3c53d40d09c342bcee0ec#diff-37658e014e0dde149ef44679e6dedd63cbe8759af46155d7def83ed21b8dda47)

dangotbanned: Apologies for the commit suggestion (https://github.com/narwhals-dev/narwhals/pull/2514/commits/71996c343e8efd1801dcc07067c8091ed60eb9b0)

`__init__.py` wasn't in the PR anymore

FBruzzesi: Sorry I had a crazy day! Updating the `bump_version.py` accordingly if the state of `__init__` is now ok

dangotbanned (approved):
Thanks @FBruzzesi 🥳

.pre-commit-config.yaml
```diff
@@ -44,7 +44,6 @@ repos:
   hooks:
     - id: check-api-reference
       name: check-api-reference
-      pass_filenames: false
```
MarcoGorelli, line 47: Why remove this?

FBruzzesi: Some mistake while going back and forth in the yaml - fixing now

.pre-commit-config.yaml
```diff
@@ -44,7 +44,6 @@ repos:
   hooks:
     - id: check-api-reference
       name: check-api-reference
```
FBruzzesi, line 46:
```suggestion
      name: check-api-reference
      pass_filenames: false
```

narwhals/__init__.py
```diff
@@ -165,3 +167,16 @@
     "to_py_scalar",
     "when",
 ]
+
+
+def __getattr__(name: _t.Literal["__version__"]) -> str:  # type: ignore[misc]
```
MarcoGorelli, line 172: To be completely honest I'd rather just update in two places, I have no idea what this is doing

Happy to take the installation update in the docs page though 👍

MarcoGorelli (changes requested):

narwhals/__init__.py
FBruzzesi: Just to make sure I understand correctly - the concern with importing `importlib.metadata` is regarding startup time? see [comment](https://github.com/narwhals-dev/narwhals/pull/2514#discussion_r2081394946)

MarcoGorelli: i feel uneasy about the global - do you have a reference from another project that does this?

FBruzzesi: Looking for [`__version__ = version(__name__)`](https://github.com/search?q=%22__version__+%3D+version%28__name__%29%22+language%3APython&type=code&l=Python) on github you will get many results

A couple of projects I know that made it in the first few pages of the list are:

- [connector-x](https://github.com/sfu-db/connector-x/blob/57387f4df77c08ec9766b7affbebd3fbc2e13c6c/connectorx-python/connectorx/__init__.py#L28)
- [manim](https://github.com/ManimCommunity/manim/blob/f304bd93eab1ca450a86eff476c7f51bd81fe1fb/manim/__init__.py#L6)

MarcoGorelli: sure, thanks

still not sure we need it though - there's some other nice improvements in here though, happy to take those

regarding perf, `from narwhals import __version__` currently takes 0m0.156s, and `from importlib import metadata` takes 0m0.108s

dangotbanned: Some notes:

- `__version__` doesn't seem to be mentioned as a standard in the [python docs](https://docs.python.org/3/search.html?q=__version__)
  - I understand lots of packages seem to provide it anyway, but I'm not sure where the pattern is documented
- `__getattr__` is a well-defined method of deferring imports
  - https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access
- There are going to be many more examples of this pattern following `3.14`
  - https://docs.python.org/3.14/whatsnew/3.14.html#optimizations
  - https://github.com/python/cpython/issues/118761

IMO, I don't think there's a use-case for checking `narwhals.__version__`, but if people want to do that - the cost of it isn't paid by everyone else since we're providing it lazily.
Are there any search results of a package checking it? (sorry I'm replying from my phone 😅)

@MarcoGorelli I understand your caution and ultimately will *defer* 😉 to you, but I think this balances:
- Theoretical backwards-compatibility
- Avoids introducing performance overhead for those that don't need it
- Ensures that potential cost is only paid once
- Lowers our maintenance burden, by simplifying version bumps

MarcoGorelli: thanks - could you elaborate on the `global` part please?

dangotbanned: > thanks - could you elaborate on the `global` part please? i've had a look at the linked projects and they don't seem to have it
@MarcoGorelli 

I'm assuming you're referring to the projects in (https://github.com/narwhals-dev/narwhals/pull/2514#discussion_r2083054239)?
The 3 `stdlib` examples I gave in (https://github.com/narwhals-dev/narwhals/pull/2514#discussion_r2083104410) all featured use of `global`:
- https://github.com/python/cpython/pull/132060/files
- https://github.com/python/cpython/pull/128898/files
- https://github.com/python/cpython/pull/129860/files

I guess for some more info on `global`, these might be helpful:
- https://docs.python.org/3/tutorial/classes.html#python-scopes-and-namespaces
- https://docs.python.org/3/tutorial/classes.html#scopes-and-namespaces-example
- https://docs.python.org/3/reference/simple_stmts.html#global
- https://docs.python.org/3/reference/datamodel.html#modules
- https://docs.python.org/3/reference/datamodel.html#module.__dict__

As a more concrete example, we can do the following on `main`:

```py
import narwhals as nw

>>> nw.__version__
'1.38.2'

>>> "__version__" in nw.__dict__
True

>>> nw.__dict__["__version__"]
'1.38.2'

>>> nw.__version__ == nw.__dict__["__version__"] == nw.__dict__.__getitem__("__version__")
True
```

But in this PR, we can see where the `__getattr__` hook slips in:

```py
import narwhals as nw

>>> nw.__dict__["__version__"]
KeyError: '__version__'

>>> nw.__version__
'1.38.2'

>>> nw.__dict__["__version__"]
'1.38.2'
```

Without the `global`:
- `__version__` is never added to `nw.__dict__`
- Each lookup of `nw.__version__` **always** goes via `__getattr__`, which is triggered after the regular `__dict__` lookup fails

So the `global` is what provides the lazy import, since the work is only ever done once and only when a user asks for it 🙂.
All subsequent lookups behave identically to a regular lookup into the namespace.

As another example, `polars` has 2 behaviors related to `__getattr__`:
- [Deprecations](https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/__init__.py#L445-L474) are triggered on **every lookup**
  - E.g. `pl.NUMERIC_DTYPES`
- [Dependencies](https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/dependencies.py) use much more magic than I understand 😄, but they eventually resolve to replacing a module's globals
  - https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/dependencies.py#L71-L76

dangotbanned: Interestingly, `polars` seems to be fine with using `find_spec` and still considers that lazy 

- https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/dependencies.py#L134
- https://github.com/pola-rs/polars/blob/34b5729cec09a8e3619c70844a77830c0613f070/py-polars/polars/dependencies.py#L173-L194

MarcoGorelli (approved):
thanks for the detailed explanations and for bearing with me on this one, really appreciate it

thanks both!

PR #972, opened 2024-09-14
ci: add codspeed for tpch queries

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues 

- Closes #805 

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added 
- [ ] Documented the changes

PR #2516, opened 2025-05-08
ci: Add codspeed for performance monitoring

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #805

## Checklist

- [ ] Tests added
- [ ] Documented the changes


FBruzzesi:
Ok, so:

- even with 10% of data, it takes 30 mins to run
- in the codspeed website I can see that these benchmarks come with a warning:
>[!WARNING]
> This benchmark contains 32 system calls, totalling 39.1 s of execution time. Since they cannot be consistently instrumented, those calls are not included in the measure. Please switch to [the Walltime instrument to accurately measure system calls](https://docs.codspeed.io/instruments/walltime). Learn more [about measurement and system calls](https://docs.codspeed.io/instruments/cpu/#system-calls).

which to me indicates that the [numbers in the report here](https://github.com/narwhals-dev/narwhals/pull/2516#issuecomment-2863275237) are not tracking what we would like to see.

Additionally, we don't get the split by backend, which is also something I would like to see if we integrate performance tooling.
To do that we would need to have the benchmark on a lower level such as in `execute_query`

tpch/generate_data.py
```diff
@@ -8,10 +8,11 @@
 import pyarrow.csv as pc
 import pyarrow.parquet as pq
 
-if not Path("data").exists():
-    Path("data").mkdir()
-
 SCALE_FACTOR = 0.1
```
dangotbanned, line 11: @FBruzzesi (https://github.com/narwhals-dev/narwhals/issues/805#issuecomment-2849315813)

> > In [#972](https://github.com/narwhals-dev/narwhals/pull/972/files#diff-5c9ed6d149656d9e7081a1e8ab603c514441aea0f18c360de825e49c32caf79fL13-R13) we were using TPCH with 0.25 ratio and it was taking ~40mins to run IIRC. That's a bit much for what I would consider fast iteration - maybe a ratio of 0.1 is more reasonable to start with
> 
> IIRC the docs for the duckdb TPCH tests used `0.01` - so we can go lower

I found the bit in the docs that used `0.01` (https://duckdb.org/docs/0.10/extensions/tpch#listing-expected-answers)

> To produced the expected results for all queries on scale factors 0.01, 0.1, and 1, run

If we can run these with 10x less data, surely we should right?

FBruzzesi: > The current run has been going for almost 2 hours 😅

Yes I have been monitoring it - it's a bit odd, isn't it? I am not fully sure what's going on 🤔

PR #2527, opened 2025-05-10
tests: fix `collect_with_kwargs` for polars nightly

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

### DIFF
--- a/tests/frame/collect_test.py
+++ b/tests/frame/collect_test.py
@@ -158,7 +158,7 @@ def test_collect_to_invalid_backend(
 
 def test_collect_with_kwargs(constructor: Constructor) -> None:
     collect_kwargs = {
-        nw.Implementation.POLARS: {"no_optimization": True},
+        nw.Implementation.POLARS: {"engine": "in-memory"},
         nw.Implementation.DASK: {"optimize_graph": False},
         nw.Implementation.PYARROW: {},
     }


FBruzzesi:
TIL:

```terminal
(Pdb) POLARS_VERSION
(1, 29, 0)
(Pdb) POLARS_VERSION > (1, 29)
True
```

FBruzzesi:
I guess I will wait for the release for it to be green 🥦😂

dangotbanned:
@FBruzzesi it seems like the issue is [marcogorelli/variable-brink-glacier](https://www.kaggle.com/code/marcogorelli/variable-brink-glacier/notebook) builds a `.whl` for [`main`](https://github.com/pola-rs/polars/tree/main) - but it has the version of the current release 🤔 

The version seems to be [grabbed dynamically](https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/py-polars/polars/_utils/polars_version.py#L2-L4) from the `PyO3` package.

So that seems to lead here
- https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/py-polars/src/lib.rs#L406-L407

Where the version comes from a complile-time ([`!env`](https://doc.rust-lang.org/std/macro.env.html)) environment variable `CARGO_PKG_VERSION`.

Maybe we could set that to the next minor/major version before building `polars`?

@MarcoGorelli does any of that sound right and/or possible? 😂

MarcoGorelli:
thanks for looking into it - maybe, although it sounds like polars might start doing pre-releases now, so this nightly build might become unneeded

FBruzzesi:
> I'm not sure why, but CI is no longer failing on other branches 🤔

@dangotbanned I guess the changes in  #2551 👀

MarcoGorelli (approved):

dangotbanned:
> > I'm not sure why, but CI is no longer failing on other branches 🤔
> 
> @dangotbanned I guess the changes in #2551 👀

Oh nice one 🎉

PR #2530, opened 2025-05-11
chore: Simplify PySpark std and var

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -365,8 +365,8 @@ class SparkLikeLazyFrame(
             window = self._Window.partitionBy(subset or self.columns)
             df = (
                 self.native.withColumn(tmp, self._F.count("*").over(window))
-                .filter(self._F.col(tmp) == 1)
-                .drop(tmp)
+                .filter(self._F.col(tmp) == self._F.lit(1))
+                .drop(self._F.col(tmp))
             )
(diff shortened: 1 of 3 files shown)


EdAbati:
I was under the impression that `stddev` and `var` coming from `pyspark.pandas.spark.functions` could be a bit faster because they take a `pyspark.sql.Column` and are implemented directly in Scala (they are not a wrapper around the `sql` API).
e.g.:

https://github.com/apache/spark/blob/2a6ddde05af38ad70bca68006384bba406502dd2/python/pyspark/pandas/spark/functions.py#L57-L58

https://github.com/apache/spark/blob/2a6ddde05af38ad70bca68006384bba406502dd2/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/aggregate/CentralMomentAgg.scala#L342-L363

But I don't have data to back up this claim :D

Also `pyspark.pandas.spark.functions` are not in `main` anymore, so maybe it is safer and cleaner to do it this way

PR #2545, opened 2025-05-13
chore: Reuse SeriesT alias more, make more types TYPE_CHECKING-only

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


dangotbanned:
@MarcoGorelli some notes, not sure if you're aware of these:

- Trailing docstrings don't work within a `TYPE_CHECKING` block in an editor
  - I'm not sure if they work in the API reference
  - but I *could* see that being an issue if they're not runtime objects 🤔
- For the `Literal` aliases, if they're runtime objects we *could* use `get_args` for error messages
- `Protocol` defs should be fine as long as we aren't inheriting from them
  - `SupportsNativeNamespace`, and `DTypes` all good
- Any aliases that are used to parameterize a protocol/generic/another alias **at runtime** have some quirks

IMO, I *think* what you're trying to do here is already covered by omitting these types from `nw.typing.__all__`

MarcoGorelli:
thanks for your review - i checked the api reference and all looks good 👍 

> Any aliases that are used to parameterize a protocol/generic/another alias at runtime have some quirks

could you elaborate on this please?

MarcoGorelli:
i'll take some of these out and leave them for a separate discussion

dangotbanned (approved):
Thanks @MarcoGorelli, all looking good in VSCode now 😍

PR #2546, opened 2025-05-13
feat: Add `narwhals.sql` module

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -6,6 +6,7 @@ from narwhals import dependencies
 from narwhals import dtypes
 from narwhals import exceptions
 from narwhals import selectors
+from narwhals import sql
 from narwhals.dataframe import DataFrame
 from narwhals.dataframe import LazyFrame
 from narwhals.dtypes import Array
@@ -161,6 +162,7 @@ __all__ = [
     "scan_parquet",
     "selectors",
     "show_versions",
+    "sql",
     "sum",
     "sum_horizontal",
     "to_native",
--- a/docs/api-reference/sql.md
+++ b/docs/api-reference/sql.md
@@ -1,4 +1,4 @@
-# `narwhals.selectors`
+# `narwhals.sql`
 
 ::: narwhals.sql
     handler: python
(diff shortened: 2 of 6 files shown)


narwhals/__init__.py
```diff
@@ -6,6 +6,7 @@
 from narwhals import dtypes
 from narwhals import exceptions
 from narwhals import selectors
+from narwhals import sql
```
dangotbanned, line 9: @MarcoGorelli have you thought about treating `sql` like the other backend packages?

So the exported interface *could* **still be**:

```py
import narwhals as nw

nw.sql.table(...)
```

But you *could* have an internal structure **more like**:

```
narwhals/
    __init__.py
    ...
    _sql/
        __init__.py
        some_1.py
        some_2.py
        table.py
        ...
```

The main benefit I'm thinking of is hiding a lot of the utils you'll eventually need.
Even now with the `__all__` and only a single function, IIRC `pyright` will show all of these on hover of `nw.sql.<...>`:
```py
TYPE_CHECKING
Mapping
narwhals_to_native_dtype
from_native
Version
StandaloneDataFrame
LazyFrame
DType
Schema
table
annotations
```

But I guess the other one would be if you decided you wanted to *more closely* mirror `polars.sql()` - I think that would be easier if `sql` wasn't the name of a top-level module


- https://docs.pola.rs/api/python/stable/reference/sql/python_api.html
- https://github.com/pola-rs/polars/tree/py-1.29.0/py-polars/polars/sql

FWIW, I haven't used `polars.sql()` before 😅

Either way I'm interested to see where this leads 🙂

MarcoGorelli:
The main barrier I'm seeing is https://github.com/eakmanrq/sqlframe/issues/399

Alternatives could be:
- Just accept this and document the limitation. Because we'd be hiding SQLFrame behind `nw.sql`, this could lead to hard-to-debug errors for users
- Use Ibis to generate SQL, but then we're limited by what they don't support
- Transpile to SQLGlot ourselves

I can kind of see the writing on the wall for the last one

PR #1048, opened 2024-09-23
feat: add `Expr|Series` `log` and `log10` methods

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues 

Used in plotly

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [x] Documented the changes

PR #2549, opened 2025-05-14
feat: Add support for `Series|Expr.log` method

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- [Related comment](https://github.com/narwhals-dev/narwhals/pull/1048#issuecomment-2877563155)

## Checklist

- [x] Tests added
- [x] Documented the changes

**TODO**:

- [x] dask via `dask.array`
- [x] All other lazy backends but polars raise an exception for both negative number and zero, polars returns NaN and -inf respectively - might need to account for those cases with dedicated checks
- [ ] Looking for idea to have better tests


narwhals/_pandas_like/series.py
```diff
@@ -1015,6 +1015,38 @@ def hist(  # noqa: C901, PLR0912
...
         return PandasLikeDataFrame.from_native(ns.DataFrame(data), context=self)
 
+    def log(self, base: float) -> Self:
+        native = self.native
+        implementation = self._implementation
+
+        dtype_backend = get_dtype_backend(native.dtype, implementation=implementation)
+
+        if implementation.is_cudf():
+            import cupy as cp
```
FBruzzesi, line 1025: cupy is a dependency of cudf, so it's a safe import

narwhals/_arrow/series.py
```diff
@@ -1148,6 +1148,9 @@ def __contains__(self, other: Any) -> bool:
             msg = f"Unable to compare other of type {type(other)} with series of type {self.dtype}."
             raise InvalidOperationError(msg) from exc
 
+    def log(self, base: float) -> Self:
+        return self._with_native(pc.logb(self.native, base))  # type: ignore[arg-type]
```
FBruzzesi, line 1152: @dangotbanned hello 👋🏼 this might be something missing in the pyarrow stubs 🤔

dangotbanned: Hey @FBruzzesi 🙂

Have you tried doing this?

```py
pc.logb(self.native, lit(base)
```

Most of the `@overload` definitions fall back to nonsensical messages (like you have here) when you don't use `pa.scalar`.
So I aliased that to `lit` for fewer headaches 😉

dangotbanned: `logb` does have an update in

- https://github.com/zen-xu/pyarrow-stubs/pull/233/files#diff-2211aa889baf83bb2bb7829fae4d85d7c532085aeddf611f4ed235782f48af69

But (unrelated to this PR) the next `pyarrow-stubs` bump we do may be difficult - since they're no longer checked with `mypy` 😞

- https://github.com/zen-xu/pyarrow-stubs/pull/227/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9

EdAbati:
It looks 🤩
What do you mean by 'Looking for idea to have better tests'?

FBruzzesi:
Hey @EdAbati 👋🏼

> It looks 🤩 

Ahah I wanted to take a simple feature after a few long debated PRs 😂
Ok in all honesty I am having very short time to contribute in the last period

> What do you mean by 'Looking for idea to have better tests'?

Ideally I would want something using hypothesis 🤔

narwhals/_arrow/series.py
dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/2549/commits/ad504c8eca8452a47214a7a0c0da648f01fe1654)


> of course Dan had typing solution :) 

@FBruzzesi 😍

MarcoGorelli:
Tbh I think hypothesis is probably overkill for log

dangotbanned:
> Tbh I think hypothesis is probably overkill for log

I thought that sounded like a good idea to weed out any edge cases.
To me, `log` seems like something we'd want to behave identically across backends.

FWIW, I have only used `pl.Expr.log` on very rare occasions - so maybe I'm being overly cautious 😄

MarcoGorelli:
the implementation looks quite simple, i'm not so worried here (like I was with say `rolling_std`)

dangotbanned:
> the implementation looks quite simple, i'm not so worried here (like I was with say `rolling_std`)

All good - in that case @FBruzzesi maybe just try a `@pytest.mark.paramterize` that covers some different `base` values?

FBruzzesi:
Sorry for the very slow iterations these days!

> the implementation looks quite simple, i'm not so worried here (like I was with say rolling_std)

@MarcoGorelli I agree, I am more concern of having edge cases behaving the same way, but I guess negative numbers and zero it's all there is 

 > try a @pytest.mark.parameterize that covers some different base values?

@dangotbanned  added in https://github.com/narwhals-dev/narwhals/pull/2549/commits/6be341c7b2a69fd16e6b425c3a5bc51112db68dc

MarcoGorelli (approved):
thanks @FBruzzesi ! I just added a little commit to address what I'd commented on, feel free to ship it!

PR #2555, opened 2025-05-16
chore: Rename `call_kwargs` to `scalar_kwargs`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_compliant/expr.py
```diff
@@ -349,7 +349,7 @@ def __init__(
         implementation: Implementation,
         backend_version: tuple[int, ...],
         version: Version,
-        call_kwargs: dict[str, Any] | None = None,
+        scalar_kwargs: dict[str, Any] | None = None,
```
dangotbanned, line 352: Should this be typed more strictly now?

```suggestion
        scalar_kwargs: dict[str, PythonLiteral] | None = None,
```

Apologies if I've misunderstood the name change 😅

MarcoGorelli: yup thanks

it requires a handful of casts though, right?

dangotbanned: > it requires a handful of casts though, right?

Yeah I think your options are either that or a `TypedDict` that looks *something like* this:

```py

from typing import TypedDict

if TYPE_CHECKING:
    from narwhals.typing import RankMethod

class _ScalarKwargs(TypedDict, total=False):
    ddof: int
    n: int
    reverse: bool
    method: RankMethod
    descending: bool
```

You'd still be checking for the presence of keys, but when you find them - they'll be the right type

MarcoGorelli: oooooh I'd forgotten about `total=False` - thanks! 😍 that's real nice

dangotbanned: aha I see you started down that route yesterday (https://github.com/narwhals-dev/narwhals/pull/2555/commits/cba568dbe1aaf19f1d8d99697b5b898bc914a500)

narwhals/typing.py
```diff
@@ -389,6 +390,21 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+    center: int
+    ddof: int
+    descending: bool
+    limit: int | None
+    method: RankMethod
+    min_samples: int
+    n: int
+    reverse: bool
+    strategy: FillNullStrategy | None
+    window_size: int
```
dangotbanned, line 405: @MarcoGorelli what do you think about moving this here?

https://github.com/narwhals-dev/narwhals/blob/main/narwhals/_compliant/typing.py


I don't think we need it so public 🙂

MarcoGorelli:
thanks for your review!

dangotbanned:
> thanks for your review!

Thanks @MarcoGorelli ♥️

PR #2528, opened 2025-05-10
feat(DRAFT): Adds `Expr.first()`

dangotbanned:
Will close #2526

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues
- Closes #2526

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- [x] Initial naive implementations
- [ ] Decide on usage rules (https://github.com/narwhals-dev/narwhals/issues/2526#issuecomment-2869123455)
- [ ] Support in `group_by().agg(...)`?
- [ ] Add Expr.last()`?
- [ ] Add `narwhals`-level methods
- [ ] Tests, tests, tests

Issue #668, opened 2024-07-29
[Enh]: improve when-then-otherwise to include chaining


aivanoved:
### Please describe the purpose of the new feature or describe the problem to solve.

As discussed in #588, where a simple `when-then-otherwise` expression is added, the case for chaining `when-then` expressions is left to be done at some later point. This will complete and close #47

### Additional information that may help us understand your needs.

I would be willing to write this code, as looking to complete the open #588 and this is a natural extension

MarcoGorelli:
@dangotbanned in your rewrite of `when/then` i recall you'd said that this might be an easy follow-up - just flagging it in case it interests you, as there's been another request for it (#2597 )

else no worries, i'll get to it before too long

dangotbanned:
Ah thanks for the ping @MarcoGorelli!

Yeah so I might need to refresh myself a bit, but IIRC the latest version of all backends would support:

- `pyarrow` function is pretty hidden (and not annotated) in the `pc.compute` stubs
- `pandas` added support somewhere in `2.*`?
  - Assuming `dask` repeats the same
- `polars` natively has it
- `ibis.cases`
- SQL-based backends I'm sure you know 😉

I'd happily review a PR, but probably won't get around to writing one myself for a while

dangotbanned:
@MarcoGorelli I've just finished the `ExprIR` version in (https://github.com/narwhals-dev/narwhals/blob/563076d5e22ce9faccfdce5f99ada865bd13f9db/narwhals/_plan/when_then.py)

The majority of it is direct from the `rust` version - which is pretty clever **and** simple.
In particular, the way [`ChainedThen.otherwise` works](https://github.com/pola-rs/polars/blob/b9dd8cdbd6e6ec8373110536955ed5940b9460ec/crates/polars-plan/src/dsl/arity.rs#L89-L130) made me smile 😄 

I'm still expecting the *"real"* version to have more complexity - but this should be helpful whenever someone takes a shot at it 🙌

williambdean:
Should this be done in a single PR for all implementations or is there way to breakup / partially complete? What's typical?

Issue #2291, opened 2025-03-25
api (low-prio): allow for multi-output expressions on both side of binary operation if they're the same


MarcoGorelli:
Low-prio, but we should be able to allow `nw.all() - nw.all().mean()`

MarcoGorelli:
here's a sketch of how it could be done
```diff
--- a/narwhals/_sql/expr.py
+++ b/narwhals/_sql/expr.py
@@ -64,13 +64,24 @@ class SQLExpr(LazyExpr[SQLLazyFrameT, NativeExprT], Protocol[SQLLazyFrameT, Nati
             other_native_series = {
                 key: df._evaluate_expr(value)
                 if self._is_expr(value)
-                else self._lit(value)
+                else [self._lit(value)]
                 for key, value in expressifiable_args.items()
             }
-            return [
-                call(native_series, **other_native_series)
-                for native_series in native_series_list
-            ]
+            res = []
+            for key, val in other_native_series.items():
+                if len(native_series_list) == 1:
+                    for x in val:
+                        res.append(call(native_series_list[0], **{key: x}))
+                elif len(val) == 1:
+                    for y in native_series_list:
+                        res.append(call(y, **{key: val[0]}))
+                elif len(native_series_list) == len(val):
+                    for x, y in zip(val, native_series_list):
+                        res.append(call(y, **{key: x}))
+                else:
+                    msg = "invalid"
+                    raise ValueError(msg)
+            return res
 
         return func
```

MarcoGorelli:
as a precursor,
```python
    def _evaluate_expr(self, expr: CompliantExprT_contra, /) -> Any:
        result = expr(self)
        assert len(result) == 1  # debug assertion  # noqa: S101
        return result[0]
```
should be probably be amended so that it returns a list of native expressions, rather than a single one

dangotbanned:
Does `polars` allow this?

dangotbanned:
> Does `polars` allow this?

Indeed it does past-me! 😄 

Same concept as (https://github.com/narwhals-dev/narwhals/issues/3029#issuecomment-3451057779), where it is [like they are zipped](https://github.com/pola-rs/polars/issues/25022#issuecomment-3447096266).

In this case though it seems less of a footgun, so I'm on board 👍

PR #2572, opened 2025-05-18
feat(RFC): A richer `Expr` IR

### LINKED ISSUE #2571: [Enh]: A richer `Expr` internal representation
## Related
This cuts across many issues/previous discussions, these are some I've seen pretty clear links to:
- https://github.com/narwhals-dev/narwhals/pull/2483#issuecomment-2866902903
- https://github.com/narwhals-dev/narwhals/pull/2483#issuecomment-2867331343
- https://github.com/narwhals-dev/narwhals/pull/2483#issuecomment-2867446959
- https://github.com/narwhals-dev/narwhals/pull/2483#issuecomment-2869070157
- (https://github.com/narwhals-dev/narwhals/pull/2538/commits/a7eeb0d23e67cb70e7cfa73cec2c7b69a15c8bef#r2083562677)
- https://github.com/narwhals-dev/narwhals/issues/2225
- https://github.com/narwhals-dev/narwhals/issues/1848
- https://github.com/narwhals-dev/narwhals/issues/2534#issuecomment-2875676729
- https://github.com/narwhals-dev/narwhals/issues/2291
- https://github.com/narwhals-dev/narwhals/issues/2522
- https://github.com/narwhals-dev/narwhals/pull/2555

## Description
Following (https://github.com/narwhals-dev/narwhals/pull/2483#issuecomment-2866902903) I've been sneakily incubating this idea that I feel *now* has enough legs to share 😏.

The big picture is replacing [`ExprMetadata`](https://github.com/narwhals-dev/narwhals/blob/c7a6080430f12c43dc37be325ee6f467fd510796/narwhals/_expression_parsing.py#L229-L567) with a rich node-based representation - that encodes **every step** of an `Expr`.
This might seem like a huge leap to take - but we're in a unique position where a [battle-tested solution](https://github.com/pola-rs/polars/blob/4d9900b3ca22bd3a6a684411d52b11d9a66435ad/crates/polars-plan/src/dsl/expr/mod.rs) already exists 😎 

So far I've been mapping out what this could look like on ([`oh-nodes`](https://github.com/narwhals-dev/narwhals/tree/oh-nodes)).
I'm going to open a PR so anyone feel free to comment on that or on this thread 

## Notes
I wrote up a bit of guiding mantra of what I'm hoping the end result should look like:
- Each `Expr` method should be representable by a single node
  - But the node does not need to be unique to the method
- A chain of `Expr` methods should form a plan of operations
- We must be able to enforce rules on what plans are permitted:
  - Must be flexible to both eager/lazy and individual backends
  - Must be flexible to a given context (select, with_columns, filter, group_by)
- Nodes are:
  - Immutable, but
    - Can be extended/re-written at both the Narwhals & Compliant levels
  - Introspectable, but
    - Store as little-as-needed for the common case
    - Provide properties/methods for computing the less frequent metadata

## Examples
As everything is represented by different classes - this one is a bit visual 😄 

### Column selections
```py
from narwhals._plan import demo as nwd

>>> nwd.col("a"), nwd.col("b", "c"), nwd.nth(1), nwd.nth(3, 4, 5)
(Narwhals DummyExpr:
 col('a'),
 Narwhals DummyExpr:
 cols(['b', 'c']),
 Narwhals DummyExpr:
 nth(1),
 Narwhals DummyExpr:
 index_columns((3, 4, 5)))
```

### Literals
We can discern the kind of literal that's wrapped:
```py
import polars as pl

from narwhals._plan import demo as nwd
from narwhals._plan.dummy import DummySeries

series = DummySeries.from_native(pl.Series([1.1, 1.2]))
scalar = nwd.lit(5)
series = nwd.lit(series)
```
```py
>>> scalar, series
(Narwhals DummyExpr:
 lit(int: 5),
 Narwhals DummyExpr:
 lit(Series))
```

```py
>>> scalar._ir.is_scalar, series._ir.is_scalar
(True, False)
```

### Funky
How about something more complex?
```py
import narwhals as nw
from narwhals._plan import demo as nwd

>>> nwd.col("a").alias("b").cast(nw.Int8()).n_unique() + (nwd.col("c").count() * nwd.lit(10))
Narwhals DummyExpr:
[(col('a').alias('b').cast(Int8).n_unique()) + ([(col('c').count()) * (lit(int: 10))])]
```

### Order dependence
Here I'm trying to enforce the rules from (https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2083557149).

The idea would be that we allow the last two variants in lazy backends:
```py
import narwhals as nw
from narwhals._plan import demo as nwd

orderable_1 = nwd.col("a").alias("d").first()
orderable_2 = nwd.col("b").cast(nw.String()).last()
orderable_3 = nwd.col("c").sort_by(nwd.col("e")).first()
orderable_4 = nwd.col("d").last().over(order_by=nwd.col("e", "f", "g"))
```

The outputs include suggestions that use the actual `Expr`

<details><summary><code>orderable_1</code></summary>
<p>

```py
>>> nwd.ensure_orderable_rules(orderable_1)
OrderDependentExprError: first() is order-dependent and requires an ordering operation for lazy backends.
Hint:
Instead of:
    col('a').alias('d').first()

If you want to aggregate to a single value, try:
    col('a').alias('d').sort_by(...).first()

Otherwise, try:
    col('a').alias('d').first().over(order_by=...)
```

</p>
</details>

<details><summary><code>orderable_2</code></summary>
<p>

```py
>>> nwd.ensure_orderable_rules(orderable_2)
OrderDependentExprError: last() is order-dependent and requires an ordering operation for lazy backends.
Hint:
Instead of:
    col('b').cast(String).last()

If you want to aggregate to a single value, try:
    col('b').cast(String).sort_by(...).last()

Otherwise, try:
    col('b').cast(String).last().over(order_by=...)
```

</p>
</details>

<details><summary><code>orderable_3</code></summary>
<p>

```py
>>> nwd.ensure_orderable_rules(orderable_3)[0]
Narwhals DummyExpr:
col('c').sort_by(by=(col('e'),), options=SortMultipleOptions(descending=[False], nulls_last=[False])).first()
```

</p>
</details>

<details><summary><code>orderable_4</code></summary>
<p>

```py
>>> nwd.ensure_orderable_rules(orderable_4)[0]
Narwhals DummyExpr:
col('d').last().over(order_by=[cols(['e', 'f', 'g'])])
```

</p>
</details>

## References
<details><summary>Lots of links to polars source</summary>
<p>

- https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/crates/polars-python/src/lazyframe/visitor/expr_nodes.rs
- https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/crates/polars-plan/src/dsl/expr.rs
- https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/crates/polars-plan/src/dsl/function_expr/mod.rs
- https://github.com/pola-rs/polars/blob/dafd0a2d0e32b52bcfa4273bffdd6071a0d5977a/crates/polars-plan/src/dsl/options/mod.rs#L137-L172
- https://github.com/pola-rs/polars/blob/3fd7ecc5f9de95f62b70ea718e7e5dbf951b6d1c/crates/polars-plan/src/plans/options.rs#L35-L106
- https://github.com/pola-rs/polars/blob/3fd7ecc5f9de95f62b70ea718e7e5dbf951b6d1c/crates/polars-plan/src/plans/options.rs#L131-L236
- https://github.com/pola-rs/polars/blob/3fd7ecc5f9de95f62b70ea718e7e5dbf951b6d1c/crates/polars-plan/src/plans/options.rs#L240-L267
- https://github.com/pola-rs/polars/blob/6df23a09a81c640c21788607611e09d9f43b1abc/crates/polars-plan/src/plans/aexpr/mod.rs

</p>
</details>

dangotbanned:
Will close #2571

- [x] ✨ Feature

## Related issues

- Closes #2571
  - See *Related* section there for more links
- Search PRs: [`(sort:updated-desc "(expr-ir)/" in:title)`](https://github.com/narwhals-dev/narwhals/pulls?q=sort%3Aupdated-desc+is%3Apr+%22%28expr-ir%29%2F%22+in%3Atitle)

## Checklist

- [x] Tests added
- [ ] Documented the changes

**Very open to feedback**

## Tasks

<details><summary><b>Show 2025 May-July</b></summary>
<p>

- [x] Investigate the sources for [`pl.Expr.meta`](https://docs.pola.rs/api/python/stable/reference/expressions/meta.html)
  - Reproducing how some of those methods work would be huge
  - See ([exploring `pl.Expr.meta`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/98292dcae05d651722b5a924dce6043c085e74f0))
- [x] Implement iteration methods for `ExprIR`
  - Needed for `meta` methods
  - [feat: Add Expr graph iteration methods](https://github.com/narwhals-dev/narwhals/pull/2572/commits/394627dcbe3e3aa26d412e0286a3821b6cef70ea)
- [x] #2578 
  - ~~Merge another PR with (https://github.com/narwhals-dev/narwhals/pull/2391#issuecomment-2832442331) first~~
  - Wanting to use `TypeVar` defaults more
  - A lot of places should have `TypeVar("T", bound=Thing, default=Thing)`, instead of an opaque `ExprIR`
- [x] `Selector`(s)
  https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/expr.py#L336-L337
  - [x] Add a variant of [`BinaryExpr`](https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/expr.py#L87-L105) that describes the restricted set of operators that are allowed
    - [x] Some WIP-rogress (https://github.com/narwhals-dev/narwhals/pull/2572/commits/87b8402ba8ace8019f0f1cc58ce8b9981d57a709)
  - [x] Very close (https://github.com/narwhals-dev/narwhals/pull/2572/commits/b9eb25e56be22407f8c3a68860efced6fe2419a7)
  - [x] ([feat: Add selectors reprs](https://github.com/narwhals-dev/narwhals/pull/2572/commits/233c67087a7420890e3de19562d35142833966ad))
    - Deviating a bit from `polars`, since they wrap `pl.col` internally
- [x] Accept `IntoExpr` in more places (including and **beyond what `narwhals` allows now**)
  - [x] `demo.py`
    - [x] [`*_horizontal`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/0bada4851ebd08380da1d000e1d2e553229e1689)
    - [x] ([feat: Expressify `concat_str`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/4808f2da12c87d7cf5bc21c03838d674f5227854))
  - [x] `dummy.py`
    - [x] ([feat: Expressify `over`, `sort_by`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/756480225c8e6cc0e98fed75f0254dd2761db37a))
    - [x] Binary operators (https://github.com/narwhals-dev/narwhals/pull/2572/commits/af274d314b6e9002bb577c4850d7fd4fa6ec5655)
- [x] Flesh out (and properly understand) `FunctionOptions` + friends [see comment](https://github.com/narwhals-dev/narwhals/pull/2572#issuecomment-2891577685)
  - [x] Top-level (https://github.com/narwhals-dev/narwhals/pull/2572/commits/0982b3a1016a618a0f2a8cdfaafac37b8f1b274b)
  -  ~~Where does the `{flags: ...}` -> `{collect_groups: ..., flags: ...}` expansion happen?~~
  - `polars>=1.3.0` fixed the issue (see [comment](https://github.com/narwhals-dev/narwhals/pull/2572#issuecomment-2901082900))
- [x] `Ternary`
  - [x] [feat: Implement chained `when-then-otherwise` 🥳](https://github.com/narwhals-dev/narwhals/pull/2572/commits/35fb57821c3c155b862db56c92356df09799023c))
  - Related (https://github.com/narwhals-dev/narwhals/issues/668#issuecomment-2904162550)
- [x] Namespaces
  - [x] `Meta`
    - [x] Support `is_*`, `has_*`, `output_name` meta methods
      - (https://github.com/narwhals-dev/narwhals/pull/2572/commits/882eff073d05171fc809f2ded219742973ef117b)
      - (https://github.com/narwhals-dev/narwhals/pull/2572/commits/6d9b9e1078f6318bc079bf8903b56372f340006d)
      - (https://github.com/narwhals-dev/narwhals/pull/2572/commits/4f1ea19bde75a20e92950df1519e108ed6d2107d)
    - [x] [`root_names`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/ffb0e39f67ab18ab1170fea3999e91a4403270eb)
    - ~~`undo_aliases`, `pop`~~
      - Too complex
  - [x] `Name`
    - [x] Redo `name.py`  (`Expr::KeepName`, `Expr::RenameAlias`)
      - Changing to how they're represented in `polars` will help with the `meta` methods 
      - See [comments](https://github.com/narwhals-dev/narwhals/pull/2572/commits/ae8cc5c6fcac746d783e98e7eb81af8ef9c15e5f)
      - Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/4a8c2d5442ccd2ff0f64d8bc04fd9bd34ab1a3ce)
    - [x] (https://github.com/narwhals-dev/narwhals/pull/2572/commits/737c83a67a85c96acf663bdfddac488b68812ea7)
  - [x] `Cat`, `Struct`, `List` (https://github.com/narwhals-dev/narwhals/pull/2572/commits/a3e29d1c8da03ba078628772cf0d562390a335f0)
  - [x] `String` (https://github.com/narwhals-dev/narwhals/pull/2572/commits/72c33ce57d48768098ba9509604b073b891c97da)
  - [x] `DateTime` (https://github.com/narwhals-dev/narwhals/pull/2572/commits/aee0a7e8976a1de0056b58881764d70084f1ab9e)
- [x] [Port over various `_expression_parsing.py` rules](https://github.com/narwhals-dev/narwhals/blob/5eb20af2f2dbb22d9c98fdfcf19692dde4b2aee4/tests/plan/expr_parsing_test.py)
- [x] Output name expansion 
  - [x] Review how the `rust` version works
  - [x] [`ExpansionFlags`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/01f4d26346f3046db5d34c330b372356540ab9da)
  - [**Pending upstream refactor**](https://github.com/narwhals-dev/narwhals/pull/2572#discussion_r2127108991)
  - [x] `expand_function_inputs`
    - [x] `rewrite_projections`
      - [x] `replace_selector`
        -  ~~`expand_selector`~~
          -  ~~`replace_selector_inner`~~
      - [x] `replace_and_add_to_results`
        - [x] `replace_nth`
        - [x] `prepare_excluded`
        - [x] `expand_columns`
        - ~~`expand_dtypes`~~
          - ~~`replace_dtype_or_index_with_column`~~
          - ~~`dtypes_match` (probably can solve w/ existing `narwhals`)~~
        - [x] `expand_indices`
          - [x] `replace_index_with_column`
        - [x] `replace_wildcard`
           - [x] `rewrite_special_aliases`
           - [x] `replace_wildcard_with_column`
        - [x] `replace_regex`
          - [x] `expand_regex`
- [x] Vistor pattern/rewrites/`ExprIR.map_ir`
  - See (https://github.com/narwhals-dev/narwhals/pull/2572#discussion_r2134750336)
  - [x] [feat: Impl `ExprIR.map_ir` for most nodes](https://github.com/narwhals-dev/narwhals/pull/2572/commits/1d6332607ff179ba97168f558c7f9b34b3e52027)
  - [x] [feat: Impl `WindowExpr.map_ir`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/25501036e8ad531225643841bb5d75c355f0c500)
  - [x] [feat: Impl `FunctionExpr.map_ir`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/7dd7092a318bae80090a92d54de0e5387374e178)
    - `RollingExpr`, `AnonymousExpr` inherit
- [x] Add tests for `selectors`
    - See (https://github.com/narwhals-dev/narwhals/blob/6414142765b8645f57888db71c7773145162e034/tests/plan/expr_expansion_test.py)
 

</p>
</details>

- [x] #3066
  - [ ] Leftover 2025 Aug-Sep refactor ideas (https://github.com/narwhals-dev/narwhals/pull/3066#issuecomment-3242037939)
- [x] #3122
- [x] #3143
  - Had left this *pending* (#2528)
- [x] #3166
- [x] #3173
- [x] Fix `__dict__` appearing on `Immutable` subclasses ([thread](https://github.com/narwhals-dev/narwhals/pull/3194#discussion_r2423147312))
  - [x] #3201
- [x] #3206
- [x] #3215 
- [x] Review where `ColumnNameOrSelector` can be added
- [ ] Support `Expr.meta.serialize`, `Expr.deserialize`
  - Briefly played around in ([feat(expr-ir): Start sketching out (de)serde](https://github.com/narwhals-dev/narwhals/commit/d5c00bc2c7ceda094e0b5cf8054689f4c8e5c6c4))
  - Definitely made easier by using such a restricted set of types
- [ ] **(low prio)** Add optimistic-lazy predicate parser into `pc.Expression`
  - Mentioned in ([thread](https://github.com/narwhals-dev/narwhals/pull/3173#discussion_r2405642521))
  - Similar to https://github.com/apache/datafusion-python/blob/e75addfa64a91c0d91ef059917a451e04495b83a/src/pyarrow_filter_expression.rs#L96-L181
- [x] #3224
  - [#3173](https://github.com/narwhals-dev/narwhals/pull/3173) needed to build out more supporting features *first*
- [x] #3295
- [x] ✨ #3233
- [x] #3265
- [x] #3294
- [x] ✨ #3325
- [ ] Add namespaced methods
  - [x] Excluding `dt.*`, `str.to_date(time)`
  - [ ] `dt.*`
  - [ ] `str.to_date(time)`
- [ ] **(ongoing)** Keep `ExprIR` with in sync with `narwhals` 
  - [x] #3353
  - [x] #3359
  - [x] #3666
  - [ ] #3315
  - [ ] #3365
  - [ ] #3392
  - [ ] #3395
  - [ ] #3460
  - [ ] #3538
  - [ ] #3625
- [x] #3352
- [ ] Review sugar for reducing `ArrowExpr` boilerplate
  - See [*What's next?*](https://github.com/narwhals-dev/narwhals/pull/3325#issue-3664477936)
- [x] #3384
- [ ] `ArrowListNamespace` -> (`ArrowExprListNamespace`, `ArrowScalarListNamespace`)
  - They can still share a base for `len`, `get`
  - Everything else has needed a different impl for `ListScalar`
- [x] Move things up from `Eager{DataFrame,Namespace}` when the `narwhals`-level depends on them
  - See [`concat` comment](https://github.com/narwhals-dev/narwhals/pull/3371#discussion_r2645787267)
- [x] #3376
- [ ] **(high prio)** Add `LogicalPlan` 🚀
  - #3611
  - Related: [1](https://github.com/narwhals-dev/narwhals/pull/3143#discussion_r2383171446), [2](https://github.com/narwhals-dev/narwhals/pull/3173#discussion_r2404448442), [3](https://github.com/narwhals-dev/narwhals/pull/3332#discussion_r2578055061)
  - First target will be sharing `over(*partition_by)` windows between expressions in the same context
    - `main` can do this for multi-output (by *not* expanding)
    - but this could be done for expressions with a different root, inside a `LogicalPlan`
d3fc48a1a0b9291ae6bac21af27e1039e63ba3cf/python/cudf_polars/cudf_polars/dsl/ir.py)
  - [x] Review [`DataType` propagation](https://github.com/pola-rs/polars/blob/b5956c30ef578949f87189330c036f86da32c516/crates/polars-plan/src/plans/aexpr/schema.rs) and [supertyping](https://github.com/pola-rs/polars/blob/b5956c30ef578949f87189330c036f86da32c516/crates/polars-core/src/utils/supertype.rs) rules
    - [ ] 🙏 #3386
    - It's looking like some version of this will be required to correctly rewrite `LogicalPlan`s
    - Otherwise, each sub-plan has to be eager and request a new schema from the backend after processing the previous step
      - That's current `narwhals` behavior, but likely more of an issue when `Selector`s can be used in more places
  - A [little write-up](https://github.com/narwhals-dev/narwhals/blob/63dbc65bc68d6dea849f927711da46c287f1a029/narwhals/_plan/logical_plan/plan.py#L29-L85) comparing `oh-nodes` to `polars` internals
- [x] **(medium prio)** Implement missing (non-`Expr`) features for `LogicalPlan`
  - [x] Functions
    - [x] #3371
  - [x] I/O
    - [x] #3370
    - [x] #3369
  - [x] `BaseFrame`
    - [x] #3378
    - [x] #3368
    - [x] #3414
  - [x] `DataFrame`
    - [x] #3373
    - [x] #3364
- [x] #3617
- [x] #3647
- [x] #3659
  - [x] #3662


dangotbanned:
## `FunctionFlags`, `FunctionOptions`, `Function`, `FunctionExpr`

Feel like there's more I need to understand on the correct way to propagate the flags to the top-level.

@MarcoGorelli whenever you get to this - I've left loads of notes w/ references - hoping that you'd be able to demystify the `rust` magic 😄 

<details><summary><b>All 4 classes</b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/options.py#L13-L49

https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/options.py#L52-L108

https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/common.py#L149-L172

https://github.com/narwhals-dev/narwhals/blob/0bada4851ebd08380da1d000e1d2e553229e1689/narwhals/_plan/expr.py#L157-L185

</p>
</details>

narwhals/_plan/common.py
```diff
@@ -0,0 +1,206 @@
...
+
+        return DummyCompliantExpr._from_ir(self, version)
+
+    @property
+    def is_scalar(self) -> bool:
+        return False
+
+
+class Function(ExprIR):
+    """Shared by expr functions and namespace functions.
```
dangotbanned, line 150: ### TODO
Give this a different base than `ExprIR`.

Following (https://github.com/narwhals-dev/narwhals/pull/2572/commits/98292dcae05d651722b5a924dce6043c085e74f0), it is clearer that we won't want to iterator over this in the same way as `nw._plan.expr` nodes.
These should all be for use in `FunctionExpr.function`

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/e6ca72b97df070e345f99503f53a146b353052b2)

narwhals/_plan/functions.py
```diff
@@ -0,0 +1,400 @@
...
+
+class CumAgg(Function):
+    __slots__ = ("reverse",)
+
+    reverse: bool
+    """https://github.com/narwhals-dev/narwhals/pull/2555"""
+
+    @property
+    def function_options(self) -> FunctionOptions:
+        return FunctionOptions.length_preserving()
```
dangotbanned, line 202: Seems like `FunctionOptions.length_preserving` is one we need to pay more attention to

dangotbanned:
Thanks for peeking @MarcoGorelli 

> at the moment it looks like this adds a self-standing `_plan`, that's not integrated into the rest of Narwhals? if it's possible to integrate it with the rest to prove that this is feasible, i'd be very interested in taking a close look

That is definitely the eventual goal! 🤞 

Despite how quickly things have progressed, I still feel I'm a few steps behind being ready for that *just yet*.

## General overview

I'm trying to focus on modeling these structures and how they interact:
- https://github.com/pola-rs/polars/blob/b9dd8cdbd6e6ec8373110536955ed5940b9460ec/crates/polars-plan/src/dsl/mod.rs
- https://github.com/pola-rs/polars/blob/b9dd8cdbd6e6ec8373110536955ed5940b9460ec/crates/polars-plan/src/dsl/expr/mod.rs

My thought was that `narwhals` currently solves *similar* problems, but in *different* ways to `polars`. 
By getting this subset of `polars` that we need translated from `rust` -> `python` **first**, we're then in a good position to make decisions on how to bridge any gaps that `narwhals` is occupying at the moment, if that makes sense?

So like what I have in [`narwhals/_plan/dummy.py`](https://github.com/narwhals-dev/narwhals/blob/9569d99f74531f49d9498c0e57f66aec33048b3b/narwhals/_plan/dummy.py) is all about creating an accurate expression graph.
Consuming the graph (evaluating an expression) is a very important step - but the main difference I'm anticipating is:

### Current
- Expressions are based on evaluating callables
- with other callables used for output names & aliases
- They also optionally store a `ExprMetadata` object
- Some backends do other things with function names, kwargs, depth
- Lazy backends use more callables for handling windows

### `ExprIR`
- Expressions are (likely still) based on evaluating callables
- All of the remaining details are *either*:
  - Encoded into a node on the graph
    - E.g. the nodes for `Expr.var`, `Expr.std` have the name and `ddof` already
    - Depth can be computed anywhere by traversing the graph
  - *Or*, can be expressed in terms of operating on a node
    - [`meta` namespace](https://github.com/narwhals-dev/narwhals/pull/2572/commits/882eff073d05171fc809f2ded219742973ef117b), [examples](https://github.com/narwhals-dev/narwhals/pull/2572/commits/4f1ea19bde75a20e92950df1519e108ed6d2107d)
    - [`WindowExpr`](https://github.com/narwhals-dev/narwhals/blob/9569d99f74531f49d9498c0e57f66aec33048b3b/narwhals/_plan/expr.py#L292-L361) is just another *albeit complex* 😅 node

I am confident we'll end up with something that's easier to maintain - but trying to integrate the two *mid-solve* **and** maintaining that branch over time seems like it'd be a real challenge 😔 

---

## `FunctionOptions`, `FunctionFlags`

The parts mentioned in (https://github.com/narwhals-dev/narwhals/pull/2572#issuecomment-2891577685) ~~are some~~ **were one** of the main hurdles left.
However I think it is really just a skill issue *on my part* in understanding the `rust` code. That was all I was hoping for some help with 🙏 
In `narwhals` terms, it is closest to (https://github.com/narwhals-dev/narwhals/blob/1b93c0ed2dc2bf47d7b8e4b4cab4c9e9cad59800/narwhals/_expression_parsing.py) but only a subset of the rules - since it only concerns functions and how they compose.

> [!NOTE]
> Right before I was about to send this comment, I managed to fix the issue I had by updating to [`1.30.0`](https://github.com/pola-rs/polars/releases/tag/py-1.30.0) 🤦‍♂️
> See ([diff](https://github.com/pola-rs/polars/pull/22573/files#diff-e1f3f8030efb7cfcccd6da0d5121c6a6954f6d81c7a328df38c631d269a4a11b)), which removed the flags I was having trouble finding in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/0982b3a1016a618a0f2a8cdfaafac37b8f1b274b)

### Example
Now all the flags I've been using are propagated in the same way as in `polars`! 🎉 

<details><summary>Repro code</summary>
<p>


```py
import polars as pl

from narwhals._plan import demo as nwd  # noqa: F811
from narwhals._plan import meta  # noqa: F811

expr_pl = (
    pl.col("a")
    .sort()
    .fill_null(1)
    .shift(1)
    .abs()
    .drop_nulls()
    .skew()
    .alias("col->sort->fill_null->shift->abs->drop_nulls->skew->alias")
)
expr_nwd = (
    nwd.col("a")
    .sort()
    .fill_null(1)
    .shift(1)
    .abs()
    .drop_nulls()
    .skew()
    .alias("col->sort->fill_null->shift->abs->drop_nulls->skew->alias")
)
roundtrip_pl = meta.polars_expr_to_dict(expr_pl)
roundtrip_nw = str(expr_nwd._ir)
```

</p>
</details> 


<details><summary><code>roundtrip_pl</code></summary>
<p>

I haven't added `ALLOW_EMPTY_INPUTS` - so that's an expected difference between the two

```py
>>> roundtrip_pl
```
```py
{'Alias': [{'Function': {'input': [{'Function': {'input': [{'Function': {'input': [{'Function': {'input': [{'Function': {'input': [{'Sort': {'expr': {'Column': 'a'},
                   'options': {'descending': False,
                    'nulls_last': False,
                    'multithreaded': True,
                    'maintain_order': False,
                    'limit': None}}},
                 {'Literal': {'Dyn': {'Int': 1}}}],
                'function': 'FillNull',
                'options': {'check_lengths': True,
                 'flags': 'ALLOW_GROUP_AWARE | ROW_SEPARABLE | LENGTH_PRESERVING'}}},
              {'Literal': {'Dyn': {'Int': 1}}}],
             'function': 'Shift',
             'options': {'check_lengths': True,
              'flags': 'ALLOW_GROUP_AWARE | LENGTH_PRESERVING'}}}],
          'function': 'Abs',
          'options': {'check_lengths': True,
           'flags': 'ALLOW_GROUP_AWARE | ROW_SEPARABLE | LENGTH_PRESERVING'}}}],
       'function': 'DropNulls',
       'options': {'check_lengths': True,
        'flags': 'ALLOW_GROUP_AWARE | ALLOW_EMPTY_INPUTS | ROW_SEPARABLE'}}}],
    'function': {'Skew': True},
    'options': {'check_lengths': True,
     'flags': 'ALLOW_GROUP_AWARE | RETURNS_SCALAR'}}},
  'col->sort->fill_null->shift->abs->drop_nulls->skew->alias']}
```



</p>
</details> 

<details><summary><code>roundtrip_nw</code></summary>
<p>

**To produce this I removed the outer `"`, `"` and pasted back to `ruff` to format**

The overall shape is very similar and the deviations from `polars` have been documented

```py
Alias(
    expr=FunctionExpr(
        function=Skew(),
        input=[
            FunctionExpr(
                function=DropNulls(),
                input=[
                    FunctionExpr(
                        function=Abs(),
                        input=[
                            FunctionExpr(
                                function=Shift(n=1),
                                input=[
                                    FunctionExpr(
                                        function=FillNull(),
                                        input=[
                                            Sort(
                                                expr=Column(name="a"),
                                                options=SortOptions(
                                                    descending=False, nulls_last=False
                                                ),
                                            ),
                                            Literal(
                                                value=ScalarLiteral(
                                                    dtype=Unknown, value=1
                                                )
                                            ),
                                        ],
                                        options=FunctionOptions(
                                            flags="ALLOW_GROUP_AWARE | ROW_SEPARABLE | LENGTH_PRESERVING"
                                        ),
                                    )
                                ],
                                options=FunctionOptions(
                                    flags="ALLOW_GROUP_AWARE | LENGTH_PRESERVING"
                                ),
                            )
                        ],
                        options=FunctionOptions(
                            flags="ALLOW_GROUP_AWARE | ROW_SEPARABLE | LENGTH_PRESERVING"
                        ),
                    )
                ],
                options=FunctionOptions(flags="ALLOW_GROUP_AWARE | ROW_SEPARABLE"),
            )
        ],
        options=FunctionOptions(flags="ALLOW_GROUP_AWARE | RETURNS_SCALAR"),
    ),
    name="col->sort->fill_null->shift->abs->drop_nulls->skew->alias",
)
```

</p>
</details>

dangotbanned:
https://results.pre-commit.ci/run/github/760058710/1747943430.kzuOzX_8QTiUCxaCHld8Wg

How do I disable this? 🤣 

```
codespell..............................................................................Failed
- hook id: codespell
- exit code: 65

narwhals/_plan/expr.py:20: RightT ==> right
narwhals/_plan/expr.py:102: RightT ==> right
narwhals/_plan/expr.py:116: RightT ==> right
narwhals/_plan/operators.py:9: RightT ==> right
narwhals/_plan/operators.py:39: RightT ==> right
narwhals/_plan/operators.py:40: RightT ==> right
narwhals/_plan/typing.py:12: RightT ==> right
narwhals/_plan/typing.py:19: RightT ==> right
narwhals/_plan/typing.py:19: RightT ==> right
```

.pre-commit-config.yaml
```diff
@@ -19,7 +19,7 @@ repos:
   hooks:
     - id: codespell
       files: \.(py|rst|md)$
-      args: [--ignore-words-list=ser]
+      args: [--ignore-words-list=ser,rightt]
```
MarcoGorelli, line 22: maybe something like
```
      args: [--ignore-words-list="ser,rightt"]
```
?

dangotbanned: Thanks @MarcoGorelli

I just tried that and it seems to break the `ser` ignore 😢 

But also not fix this one

<details><summary>Details</summary>
<p>

```
pre-commit run codespell -a
codespell................................................................Failed
- hook id: codespell
- exit code: 65

narwhals/dependencies.py:135: ser ==> set
narwhals/dependencies.py:136: ser ==> set
narwhals/dependencies.py:137: ser ==> set
narwhals/dependencies.py:139: ser ==> set
narwhals/dependencies.py:158: ser ==> set
narwhals/dependencies.py:159: ser ==> set
narwhals/dependencies.py:160: ser ==> set
narwhals/dependencies.py:175: ser ==> set
narwhals/dependencies.py:176: ser ==> set
narwhals/dependencies.py:177: ser ==> set
narwhals/dependencies.py:222: ser ==> set
narwhals/dependencies.py:223: ser ==> set
narwhals/dependencies.py:224: ser ==> set
narwhals/dependencies.py:227: ser ==> set
narwhals/dependencies.py:228: ser ==> set
narwhals/dependencies.py:229: ser ==> set
narwhals/dependencies.py:308: ser ==> set
narwhals/dependencies.py:309: ser ==> set
narwhals/dependencies.py:313: ser ==> set
narwhals/dependencies.py:313: ser ==> set
narwhals/dependencies.py:313: ser ==> set
narwhals/dependencies.py:426: ser ==> set
narwhals/dependencies.py:427: ser ==> set
narwhals/dependencies.py:435: ser ==> set
narwhals/dependencies.py:438: ser ==> set
narwhals/dependencies.py:439: ser ==> set
narwhals/dependencies.py:439: ser ==> set
narwhals/_polars/series.py:271: ser ==> set
narwhals/_polars/series.py:277: ser ==> set
narwhals/_dask/dataframe.py:96: ser ==> set
narwhals/_dask/dataframe.py:97: ser ==> set
narwhals/_dask/dataframe.py:229: ser ==> set
narwhals/_dask/dataframe.py:230: ser ==> set
narwhals/_dask/dataframe.py:230: ser ==> set
narwhals/_dask/dataframe.py:230: ser ==> set
narwhals/_dask/dataframe.py:231: ser ==> set
narwhals/_arrow/series_dt.py:69: ser ==> set
narwhals/_arrow/series_dt.py:70: ser ==> set
narwhals/_arrow/series_dt.py:73: ser ==> set
narwhals/_arrow/series_dt.py:74: ser ==> set
narwhals/_arrow/series_dt.py:75: ser ==> set
narwhals/_arrow/series_dt.py:76: ser ==> set
narwhals/_arrow/series_dt.py:87: ser ==> set
narwhals/_compliant/selectors.py:99: ser ==> set
narwhals/_compliant/selectors.py:99: ser ==> set
narwhals/_compliant/selectors.py:111: ser ==> set
narwhals/_compliant/selectors.py:111: ser ==> set
narwhals/_compliant/selectors.py:129: ser ==> set
narwhals/_compliant/selectors.py:129: ser ==> set
narwhals/_compliant/selectors.py:138: ser ==> set
narwhals/_compliant/selectors.py:138: ser ==> set
narwhals/_compliant/selectors.py:174: ser ==> set
narwhals/_compliant/selectors.py:174: ser ==> set
narwhals/_compliant/selectors.py:186: ser ==> set
narwhals/_compliant/selectors.py:187: ser ==> set
narwhals/_compliant/selectors.py:187: ser ==> set
narwhals/_compliant/selectors.py:193: ser ==> set
narwhals/_compliant/selectors.py:194: ser ==> set
narwhals/_compliant/selectors.py:194: ser ==> set
narwhals/_arrow/series.py:211: ser ==> set
narwhals/_arrow/series.py:212: ser ==> set
narwhals/_arrow/series.py:216: ser ==> set
narwhals/_arrow/series.py:217: ser ==> set
narwhals/_arrow/series.py:220: ser ==> set
narwhals/_arrow/series.py:221: ser ==> set
narwhals/_arrow/series.py:224: ser ==> set
narwhals/_arrow/series.py:225: ser ==> set
narwhals/_arrow/series.py:228: ser ==> set
narwhals/_arrow/series.py:229: ser ==> set
narwhals/_arrow/series.py:232: ser ==> set
narwhals/_arrow/series.py:233: ser ==> set
narwhals/_arrow/series.py:236: ser ==> set
narwhals/_arrow/series.py:237: ser ==> set
narwhals/_arrow/series.py:240: ser ==> set
narwhals/_arrow/series.py:241: ser ==> set
narwhals/_arrow/series.py:244: ser ==> set
narwhals/_arrow/series.py:245: ser ==> set
narwhals/_arrow/series.py:248: ser ==> set
narwhals/_arrow/series.py:249: ser ==> set
narwhals/_arrow/series.py:252: ser ==> set
narwhals/_arrow/series.py:253: ser ==> set
narwhals/_arrow/series.py:259: ser ==> set
narwhals/_arrow/series.py:260: ser ==> set
narwhals/_arrow/series.py:266: ser ==> set
narwhals/_arrow/series.py:267: ser ==> set
narwhals/_arrow/series.py:273: ser ==> set
narwhals/_arrow/series.py:274: ser ==> set
narwhals/_arrow/series.py:277: ser ==> set
narwhals/_arrow/series.py:278: ser ==> set
narwhals/_arrow/series.py:281: ser ==> set
narwhals/_arrow/series.py:282: ser ==> set
narwhals/_arrow/series.py:285: ser ==> set
narwhals/_arrow/series.py:286: ser ==> set
narwhals/_arrow/series.py:289: ser ==> set
narwhals/_arrow/series.py:290: ser ==> set
narwhals/_arrow/series.py:293: ser ==> set
narwhals/_arrow/series.py:294: ser ==> set
narwhals/_arrow/series.py:298: ser ==> set
narwhals/_arrow/series.py:299: ser ==> set
narwhals/_arrow/series.py:304: ser ==> set
narwhals/_arrow/series.py:305: ser ==> set
narwhals/_arrow/dataframe.py:496: ser ==> set
narwhals/_arrow/dataframe.py:496: ser ==> set
narwhals/_arrow/dataframe.py:496: ser ==> set
narwhals/_arrow/dataframe.py:497: ser ==> set
narwhals/_arrow/dataframe.py:497: ser ==> set
narwhals/_arrow/dataframe.py:497: ser ==> set
narwhals/_plan/typing.py:12: RightT ==> right
narwhals/_plan/typing.py:19: RightT ==> right
narwhals/_plan/typing.py:19: RightT ==> right
tests/expr_and_series/is_in_test.py:28: ser ==> set
tests/expr_and_series/is_in_test.py:29: ser ==> set
narwhals/_plan/operators.py:9: RightT ==> right
narwhals/_plan/operators.py:39: RightT ==> right
narwhals/_plan/operators.py:40: RightT ==> right
tests/stable_api_test.py:146: ser ==> set
tests/stable_api_test.py:147: ser ==> set
narwhals/_pandas_like/series.py:247: ser ==> set
narwhals/_pandas_like/series.py:248: ser ==> set
narwhals/_pandas_like/series.py:251: ser ==> set
narwhals/_pandas_like/series.py:261: ser ==> set
narwhals/_pandas_like/series.py:339: ser ==> set
narwhals/_pandas_like/series.py:343: ser ==> set
narwhals/_pandas_like/series.py:343: ser ==> set
narwhals/_pandas_like/series.py:345: ser ==> set
narwhals/_pandas_like/series.py:345: ser ==> set
narwhals/_pandas_like/series.py:347: ser ==> set
narwhals/_pandas_like/series.py:347: ser ==> set
narwhals/_pandas_like/series.py:349: ser ==> set
narwhals/_pandas_like/series.py:349: ser ==> set
narwhals/_pandas_like/series.py:352: ser ==> set
narwhals/_pandas_like/series.py:358: ser ==> set
narwhals/_pandas_like/series.py:359: ser ==> set
narwhals/_pandas_like/series.py:359: ser ==> set
narwhals/_pandas_like/series.py:359: ser ==> set
narwhals/_pandas_like/series.py:359: ser ==> set
narwhals/_pandas_like/series.py:359: ser ==> set
narwhals/_pandas_like/series.py:368: ser ==> set
narwhals/_pandas_like/series.py:370: ser ==> set
narwhals/_pandas_like/series.py:371: ser ==> set
narwhals/_pandas_like/series.py:383: ser ==> set
narwhals/_pandas_like/series.py:384: ser ==> set
narwhals/_pandas_like/series.py:387: ser ==> set
narwhals/_pandas_like/series.py:388: ser ==> set
narwhals/_pandas_like/series.py:391: ser ==> set
narwhals/_pandas_like/series.py:392: ser ==> set
narwhals/_pandas_like/series.py:395: ser ==> set
narwhals/_pandas_like/series.py:396: ser ==> set
narwhals/_pandas_like/series.py:399: ser ==> set
narwhals/_pandas_like/series.py:400: ser ==> set
narwhals/_pandas_like/series.py:403: ser ==> set
narwhals/_pandas_like/series.py:404: ser ==> set
narwhals/_pandas_like/series.py:407: ser ==> set
narwhals/_pandas_like/series.py:408: ser ==> set
narwhals/_pandas_like/series.py:411: ser ==> set
narwhals/_pandas_like/series.py:412: ser ==> set
narwhals/_pandas_like/series.py:412: ser ==> set
narwhals/_pandas_like/series.py:413: ser ==> set
narwhals/_pandas_like/series.py:416: ser ==> set
narwhals/_pandas_like/series.py:417: ser ==> set
narwhals/_pandas_like/series.py:420: ser ==> set
narwhals/_pandas_like/series.py:421: ser ==> set
narwhals/_pandas_like/series.py:421: ser ==> set
narwhals/_pandas_like/series.py:422: ser ==> set
narwhals/_pandas_like/series.py:425: ser ==> set
narwhals/_pandas_like/series.py:426: ser ==> set
narwhals/_pandas_like/series.py:433: ser ==> set
narwhals/_pandas_like/series.py:434: ser ==> set
narwhals/_pandas_like/series.py:441: ser ==> set
narwhals/_pandas_like/series.py:442: ser ==> set
narwhals/_pandas_like/series.py:449: ser ==> set
narwhals/_pandas_like/series.py:450: ser ==> set
narwhals/_pandas_like/series.py:457: ser ==> set
narwhals/_pandas_like/series.py:458: ser ==> set
narwhals/_pandas_like/series.py:465: ser ==> set
narwhals/_pandas_like/series.py:466: ser ==> set
narwhals/_pandas_like/series.py:473: ser ==> set
narwhals/_pandas_like/series.py:474: ser ==> set
narwhals/_pandas_like/series.py:543: ser ==> set
narwhals/_pandas_like/series.py:545: ser ==> set
narwhals/_pandas_like/series.py:545: ser ==> set
narwhals/_pandas_like/series.py:555: ser ==> set
narwhals/_pandas_like/series.py:559: ser ==> set
narwhals/_pandas_like/series.py:563: ser ==> set
narwhals/_pandas_like/series.py:565: ser ==> set
narwhals/_pandas_like/series.py:763: ser ==> set
narwhals/_pandas_like/series.py:766: ser ==> set
narwhals/_plan/expr.py:20: RightT ==> right
narwhals/_plan/expr.py:102: RightT ==> right
narwhals/_plan/expr.py:116: RightT ==> right
```

</p>
</details>

dangotbanned: @MarcoGorelli you were very close!

(https://github.com/narwhals-dev/narwhals/pull/2572/commits/89c45879ba9d23b6b3a4d593595fe64e57541704)

I really would never have thought passing an argument that ends with `-list` would be accepted more than once 🤷‍♂️ 

```
args: [--ignore-words-list=ser, --ignore-words-list=RightT]
```

dangotbanned:
I should've expected this, but it was a nice suprise to find we get hashable selectors for free 😄 

```py
from narwhals._plan import selectors as ndcs

>>> ndcs.matches("[^z]a")._ir == ndcs.matches("[^z]a")._ir
True

>>> ndcs.matches("[^z]a")._ir == ndcs.matches("abc")._ir
False
```

@MarcoGorelli regarding (#2291)

```py
from narwhals._plan import selectors as ndcs

>>> ndcs.all()._ir == ndcs.all()._ir
True

lhs = ndcs.all()
rhs = ndcs.all().mean()

>>> lhs._ir == rhs._ir
False

>>> lhs._ir == rhs._ir.expr
True
```

And the same holds for the *non-selectors* `all` 🥳 

```py
from narwhals._plan import demo as nwd

lhs = nwd.all()
rhs = nwd.all().mean()
>>> lhs._ir == rhs._ir
False

>>> lhs._ir == rhs._ir.expr
True

>>> type(rhs._ir)
narwhals._plan.aggregation.Mean
```

tests/plan/expr_parsing_test.py
```diff
@@ -0,0 +1,193 @@
...
+    with pytest.raises(InvalidOperationError, match=ELEMENTWISE_ERR):
+        assert nwd.sum_horizontal(a, a.cum_sum()).over(order_by="a")
+
+    assert nwd.sum_horizontal(a, a.cum_sum().over(order_by="i"))
+    assert nwd.sum_horizontal(a.diff(), a.cum_sum().over(order_by="i"))
+    with pytest.raises(InvalidOperationError, match=ELEMENTWISE_ERR):
+        assert nwd.sum_horizontal(a.diff(), a.cum_sum()).over(order_by="i")
+
+    with pytest.raises(InvalidOperationError, match=ELEMENTWISE_ERR):
+        assert nwd.sum_horizontal(a.diff().abs(), a.cum_sum()).over(order_by="i")
```
dangotbanned, line 110: @MarcoGorelli quick question

This is adapted from an existing test:

<details><summary><code>tests.expression_parsing_test.test_window_kind</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/63c8e4771a1df4e0bfeea5559c303a4a447d5cc2/tests/expression_parsing_test.py#L10-L45

</p>
</details> 

AFAICT, all of the expressions I've needed a `InvalidOperationError` for *shouldn't* be valid.

But they aren't raising in current `narwhals` 🤔 

### 1
```py
import narwhals as nw

a = nw.col("a")
a.cum_sum().abs().over(order_by="id")
```

<details><summary>This error explicitly mentions <code>abs</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/9bd10ad47acfa042d2c51455fd5218539bc08e1a/narwhals/_expression_parsing.py#L357-L362


</p>
</details> 

### 2, 3, 4
These are all raising the same as (**1**), but the issue seems to be that horizontal functions aren't being treated as elementwise

```py
import narwhals as nw

a = nw.col("a")
nw.sum_horizontal(a, a.cum_sum()).over(order_by="a")
nw.sum_horizontal(a.diff(), a.cum_sum()).over(order_by="i")
nw.sum_horizontal(a.diff().abs(), a.cum_sum()).over(order_by="i")
```

In `polars`, they all seem to be `elementwise` but with an additional flag

https://github.com/pola-rs/polars/blob/944bf553f1111c31259b55348a7dd0a512ae51a1/crates/polars-plan/src/dsl/function_expr/mod.rs#L1388-L1392

I've done the same in this PR, but I don't think that flag would factor into this?

https://github.com/narwhals-dev/narwhals/blob/9bd10ad47acfa042d2c51455fd5218539bc08e1a/narwhals/_plan/functions.py#L291-L299

narwhals/_namespace.py
```diff
@@ -203,7 +203,9 @@ def from_backend(cls, backend: EagerAllowed, /) -> EagerAllowedNamespace: ...
 
     @overload
     @classmethod
-    def from_backend(cls, backend: ModuleType, /) -> Namespace[CompliantNamespaceAny]: ...
+    def from_backend(
+        cls, backend: IntoBackend, /
+    ) -> Namespace[CompliantNamespaceAny]: ...
```
dangotbanned, line 136: ### TODO
- [x] Split into another PR
- [x] Merge it #2622

Needed this to get the typing to chill out when parametrizing backends 😅

tests/plan/expr_parsing_test.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+    assert _is_expr_ir_binary_expr(b.unique() * a.map_batches(lambda x: x).first())
+
+
+# TODO @dangotbanned: Figure out how to fit this in
+@pytest.mark.xfail(reason="Did not raise, haven't added a check to raise for it yet")
+def test_invalid_binary_expr_shape() -> None:
+    """Cannot combine length-changing expressions with length-preserving ones or aggregations."""
+    with pytest.raises(ShapeError):
+        nwd.col("a").unique() + nwd.col("b")
```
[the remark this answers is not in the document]
dangotbanned: If we remove the aggregation part, then I think the rule can be reduced to requiring the other operand to be scalar

For that, we've got a property on every node already 😎 

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/common.py#L148-L150

A nice thing about it is that the traversal is handled by each node, for example:

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/expr.py#L75-L77

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/expr.py#L154-L156

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/expr.py#L278-L280

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/aggregation.py#L19-L21

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/literal.py#L46-L48

https://github.com/narwhals-dev/narwhals/blob/eb7605eea55e18c4a784e6dfb7404ef3dac22fb3/narwhals/_plan/common.py#L268-L270

narwhals/_plan/expr.py
```diff
@@ -88,6 +91,12 @@ def iter_right(self) -> t.Iterator[ExprIR]:
         yield self
         yield from self.expr.iter_right()
 
+    def __init__(self, *, expr: ExprIR, name: str) -> None:
+        if expr.meta.has_multiple_outputs():
+            raise alias_duplicate_error(expr, name)
+        kwds = {"expr": expr, "name": name}
+        super().__init__(**kwds)
```
dangotbanned: I'm somewhat showing off here, since we don't raise this early at the moment 😏 

Wherever the current test ends up is no longer a valid state 😎 

https://github.com/narwhals-dev/narwhals/blob/a0d82cd8f30b42930fd8b8cec2e1c38bb05ddf7b/tests/frame/select_test.py#L151-L154

narwhals/_plan/expr_expansion.py
```diff
@@ -0,0 +1,306 @@
+"""Based on [polars-plan/src/plans/conversion/expr_expansion.rs].
+
+## Notes
+- Goal is to expand every selection into a named column.
+- Most will require only the column names of the schema.
+
+## Current `narwhals`
+As of [6e57eff4f059c748cf84ddcae276a74318720b85], many of the problems
+this module would solve *currently* have solutions distributed throughout `narwhals`.
```
dangotbanned: > [!NOTE]
> Hold off on this until this new upstream PR is resolved

- [ ] https://github.com/pola-rs/polars/pull/23081

narwhals/_plan/common.py
```diff
@@ -0,0 +1,372 @@
...
+    def map_ir(self, function: Callable[[ExprIR], ExprIR], /) -> ExprIR:
+        """Apply `function` to each child node, returning a new `ExprIR`.
+
+        See [`polars_plan::plans::iterator::Expr.map_expr`] and [`polars_plan::plans::visitor::visitors`].
+
+        [`polars_plan::plans::iterator::Expr.map_expr`]: https://github.com/pola-rs/polars/blob/0fa7141ce718c6f0a4d6ae46865c867b177a59ed/crates/polars-plan/src/plans/iterator.rs#L152-L159
+        [`polars_plan::plans::visitor::visitors`]: https://github.com/pola-rs/polars/blob/0fa7141ce718c6f0a4d6ae46865c867b177a59ed/crates/polars-plan/src/plans/visitor/visitors.rs
+        """
+        msg = "Need to handle recursive visiting first!"
+        raise NotImplementedError(msg)
```
dangotbanned: > [!IMPORTANT]
> All of the [`expr_expansion`](https://github.com/narwhals-dev/narwhals/blob/f76c9ddbafd367c05d99b572140caadfba1bd1f7/narwhals/_plan/expr_expansion.py) work is dependent on getting this working

I've planned out some *very simplified* tests that are probably easier to follow than the real stuff in `expr_expansion`:


<details><summary><code>tests.plan.expr_expansion_test.py</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/f76c9ddbafd367c05d99b572140caadfba1bd1f7/tests/plan/expr_expansion_test.py#L209-L265

</p>
</details> 

Once all of this stuff is working (🤞), we should have a backend-agnostic solution for all of these parts of `narwhals`

<details><summary>Current <code>narwhals</code></summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/f76c9ddbafd367c05d99b572140caadfba1bd1f7/narwhals/_plan/expr_expansion.py#L7-L32

</p>
</details> 

And if we're *really lucky*, large parts/most of the work will be safe to cache - if we find any bottlenecks 😎

dangotbanned: Made a lot of progress here today 🎉 

The `selectors` tests show off the rewriting the best so far

- [feat: Impl `ExprIR.map_ir` for most nodes](https://github.com/narwhals-dev/narwhals/pull/2572/commits/1d6332607ff179ba97168f558c7f9b34b3e52027)
- [test: Add `test_replace_selector`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/360caec89d70f952a317da06b98406be06ce63a8)

dangotbanned: Surprised myself with how much simpler `FunctionExpr` was vs `WindowExpr` 😄 

- [feat: Impl `FunctionExpr.map_ir`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/7dd7092a318bae80090a92d54de0e5387374e178)

narwhals/_plan/expr.py
dangotbanned: > I'm somewhat showing off here, since we don't raise this early at the moment 😏
> 

Famous last words, I managed to introduce a bug 😄 

The return of `expr.meta.has_multiple_outputs` matches upstream - but that isn't the same condition `polars` uses to detect the duplicate


 

<details><summary>🐛🐛🐛</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/93033385a910424b5aecebabc1c9a9d08b45fef7/tests/plan/expr_parsing_test.py#L326-L339

</p>
</details> 


I'm not sure if this is me not understanding the definition of `has_multiple_outputs` - or it is a bug in `polars`.


## Examples
<details><summary><b>Repro</b></summary>
<p>

```py
import polars as pl


def has_multiple_outputs(*exprs: pl.Expr) -> None:
    s = "\n".join(
        f"expr={expr!s}\nhas_multiple_outputs={expr.meta.has_multiple_outputs()}\n"
        for expr in exprs
    )
    print(s)  # noqa: T201


>>> has_multiple_outputs(
    pl.col("a"),
    pl.col("a", "b"),  # The only one which actually expands to multiple **outputs**
    (pl.col("a") + pl.col("b")),
    (pl.col("a") + pl.col("b") + pl.col("c")),
    pl.sum_horizontal("a", "b"),
    pl.sum_horizontal("a", "b", "c"),
    pl.sum_horizontal(pl.col("a"), "b", "c"),
    pl.sum_horizontal(pl.col("a", "b"), "c"),
    pl.sum_horizontal(pl.col("a", "b", "c")),
    pl.sum_horizontal(pl.col("a", "b", "c")).alias("a+b+c"),
)
```


</p>
</details> 

<details><summary><b>Output</b></summary>
<p>

```py
expr=col("a")
has_multiple_outputs=False

expr=cols(["a", "b"])
has_multiple_outputs=True

expr=[(col("a")) + (col("b"))]
has_multiple_outputs=False

expr=[([(col("a")) + (col("b"))]) + (col("c"))]
has_multiple_outputs=False

expr=col("a").sum_horizontal([col("b")])
has_multiple_outputs=False

expr=col("a").sum_horizontal([col("b"), col("c")])
has_multiple_outputs=False

expr=col("a").sum_horizontal([col("b"), col("c")])
has_multiple_outputs=False

expr=cols(["a", "b"]).sum_horizontal([col("c")])
has_multiple_outputs=True

expr=cols(["a", "b", "c"]).sum_horizontal()
has_multiple_outputs=True

expr=cols(["a", "b", "c"]).sum_horizontal().alias("a+b+c")
has_multiple_outputs=True
```

</p>
</details> 

The test that they're checked against is quite simple.
We iterate over the tree until **any** node returns `True` for the inner check

https://github.com/narwhals-dev/narwhals/blob/93033385a910424b5aecebabc1c9a9d08b45fef7/narwhals/_plan/meta.py#L27-L28

https://github.com/narwhals-dev/narwhals/blob/93033385a910424b5aecebabc1c9a9d08b45fef7/narwhals/_plan/meta.py#L172-L175

But I think a horizontal reduction should return `False` **iff** all multi-selection nodes are contained within the reduction.

At runtime, this is valid so I'd argue it must hold that it doesn't have multiple outputs:

```py
import polars as pl

ldf = pl.LazyFrame(schema={"a": int, "b": int, "c": int, "d": int})
>>> ldf.with_columns(pl.sum_horizontal(pl.col(["a", "b", "c"])).alias("a+b+c")).collect()
shape: (0, 5)
┌─────┬─────┬─────┬─────┬───────┐
│ a   ┆ b   ┆ c   ┆ d   ┆ a+b+c │
│ --- ┆ --- ┆ --- ┆ --- ┆ ---   │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64   │
╞═════╪═════╪═════╪═════╪═══════╡
└─────┴─────┴─────┴─────┴───────┘
```

narwhals/_plan/expr.py
```diff
@@ -0,0 +1,837 @@
...
+def _cols(names: t.Iterable[str], /) -> Seq[Column]:
+    return tuple(_col(name) for name in names)
+
+
+class _ColumnSelection(ExprIR):
+    """Nodes which can resolve to `Column`(s) with a `Schema`."""
+
+    def expand_columns(self, schema: _Schema, /) -> Seq[Column]:
+        """Transform selection in context of `schema` into simpler nodes."""
+        raise NotImplementedError
```
dangotbanned: ## TODO
- [x] Remove `_ColumnSelection.expand_columns`
    - I added this before the `map_ir`/`expr_expansion` stuff
    - Those can handle much more complex cases - in addition to these simpler ones
- [x] Add equivalent schema checking after/at the end of `prepare_projection`
    - Only needs to deal with `Column` at that stage
    - Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/090330c521e038da54bb517ce5ad0810004c2951)
- [x] Redo the tests from (https://github.com/narwhals-dev/narwhals/commit/590af55073cad8d7a2835d53c87772f98faa20ef) to use the new path
    - *Mostly* resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/090330c521e038da54bb517ce5ad0810004c2951)

narwhals/_plan/expr.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/090330c521e038da54bb517ce5ad0810004c2951)

narwhals/_plan/expr.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/d8dcfa4398948288da1a19d57809baead63cd899)

narwhals/_plan/expr_expansion.py
dangotbanned: I went ahead and pulled the plug on implementing, since there hasn't been any movement on (https://github.com/pola-rs/polars/pull/23081).

Quite happy with all the functionality that's got tests so far 🥳 

- https://github.com/narwhals-dev/narwhals/blob/6414142765b8645f57888db71c7773145162e034/tests/plan/expr_expansion_test.py

There's a very clear path to expressifying all of `narwhals` now 😄

narwhals/_plan/expr_expansion.py
```diff
@@ -0,0 +1,506 @@
...
+    if not _all_columns_match(origin, columns):
+        msg = "expanding more than one `col` is not allowed"
+        raise ComputeError(msg)
+    result: deque[ExprIR] = deque()
+    for name in columns.names:
+        if name not in exclude:
+            expanded = replace_with_column(origin, Columns, name)
+            expanded = rewrite_special_aliases(expanded)
+            result.append(expanded)
+    return tuple(result)
```
dangotbanned: ## TODO: Plan how to handle exceptions & caching

All of these can raise currently:
- `expand_columns`
- `expand_indices`
- `replace_wildcard`
- `rewrite_special_aliases`

IIRC, the issues are:
- **Raising** an exception within a cached function 
  - ~~*I think* makes the next (matching) call return `None` instead of raising again~~
  - No idea where I got that ^ from, **seems to be working fine** 🥳  
- **Returning** the exception
  - *I think* there's an issue of raising a *previously* raised exception
  - Possible solution is introducing a result type that always reconstructs the exception before raising
    - Maybe I've been reading too much `rust` 😅

narwhals/_plan/expr_expansion.py
```diff
@@ -0,0 +1,518 @@
...
+
+
+def expand_indices(
+    origin: ExprIR, /, indices: IndexColumns, *, schema: FrozenSchema, exclude: Excluded
+) -> Seq[ExprIR]:
```
dangotbanned: ### TODO
- [x] Replace `schema: FrozenSchema` with `col_names: FrozenColumns`

narwhals/_plan/expr_expansion.py
```diff
@@ -0,0 +1,518 @@
...
+                rewrite_projections(child.input, keys=(), schema=schema)
+            )
+        return child
+
+    return origin.map_ir(fn)
+
+
+def replace_nth(origin: ExprIR, /, schema: FrozenSchema) -> ExprIR:
+    n_fields = len(schema)
+    names = tuple(schema)
```
dangotbanned: ### TODO
- [x] Replace `schema: FrozenSchema` with `col_names: FrozenColumns`

narwhals/_plan/expr_expansion.py
dangotbanned: - This and (https://github.com/narwhals-dev/narwhals/pull/2572#discussion_r2145593529) don't need the `DType`(s)
- so they can increase the chances of a cache hit by allowing a match between two schemas that only share the same column names
  - *once caching is added here*

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/708d6ac1d719d5c5437572568959832ec34b83d8)

narwhals/_plan/expr_expansion.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/708d6ac1d719d5c5437572568959832ec34b83d8)

narwhals/_plan/expr_expansion.py
dangotbanned: > [!IMPORTANT]
> Investigated and found this wasn't an issue

dangotbanned: Alright so I finally ran into the issue I was expecting to see - but fully misremembered what it was 😅 

Instead of silently not raising - we get an mangled traceback - which gets worse the more times you hit the error.

<details><summary><b>Traceback</b></summary>
<p>

```py
ComputeError                              Traceback (most recent call last)
Cell In[25], line 1
----> 1 nwd.col("a", "b").meta.output_name()

File /Users/danie/Documents/GitHub/narwhals/narwhals/_plan/meta.py:80, in IRMetaNamespace.output_name(self, raise_if_undetermined)
     77 if isinstance(ok_or_err, ComputeError):
     78     if raise_if_undetermined:
     79         # _expr_output_name.cache_clear()
---> 80         raise ok_or_err
     81     return None
     82 return ok_or_err

    [... skipping hidden 1 frame]

Cell In[25], line 1
----> 1 nwd.col("a", "b").meta.output_name()

File /Users/danie/Documents/GitHub/narwhals/narwhals/_plan/meta.py:80, in IRMetaNamespace.output_name(self, raise_if_undetermined)
     77 if isinstance(ok_or_err, ComputeError):
     78     if raise_if_undetermined:
     79         # _expr_output_name.cache_clear()
---> 80         raise ok_or_err
     81     return None
     82 return ok_or_err

    [... skipping similar frames: <module> at line 1 (4 times), IRMetaNamespace.output_name at line 80 (4 times), InteractiveShell.run_code at line 3667 (4 times)]

    [... skipping hidden 1 frame]

Cell In[25], line 1
----> 1 nwd.col("a", "b").meta.output_name()

File /Users/danie/Documents/GitHub/narwhals/narwhals/_plan/meta.py:80, in IRMetaNamespace.output_name(self, raise_if_undetermined)
     77 if isinstance(ok_or_err, ComputeError):
     78     if raise_if_undetermined:
     79         # _expr_output_name.cache_clear()
---> 80         raise ok_or_err
     81     return None
     82 return ok_or_err

ComputeError: this expression may produce multiple output names
```

</p>
</details> 

The fix is commented out, which is **clear the cache *before* raising** - **if** the cached value is an `Exception`

dangotbanned: First fix in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/357d419daf2184fbbe2dd7aca1554f71099f598f)

tests/plan/meta_test.py
```diff
@@ -54,3 +54,41 @@ def test_meta_root_names(
...
     assert nw_result == expected
     assert nw_result == pl_result
+
+
+@pytest.mark.parametrize(
+    ("nw_expr", "pl_expr", "expected"),
+    [
+        (nwd.col("a"), pl.col("a"), "a"),
+        (nwd.lit(1), pl.lit(1), "literal"),
+        (nwd.len(), pl.len(), "len"),
```
dangotbanned: https://github.com/narwhals-dev/narwhals/actions/runs/15848535214/job/44676174562?pr=2572

- [x] Figure out which version added `pl.len`
- [x] Make it conditional, like w/ `over`

tests/plan/meta_test.py
```diff
@@ -54,3 +54,41 @@ def test_meta_root_names(
...
+            "c",
+        ),
+        pytest.param(
+            nwd.col("c").alias("x").fill_null(50),
+            pl.col("c").alias("x").fill_null(50),
+            "x",
+            marks=pytest.mark.xfail(
+                reason="Incorrectly matched `Literal.name` instead of earlier `Alias.name`."
+            ),
+        ),
```
dangotbanned, line 208: ### Investigate
(https://github.com/narwhals-dev/narwhals/pull/2572/commits/7dcdf86b533f10444234efa804d13701daa3f582)
> Discovered the bug while writing tests for `NamedIR` - which lost track of the alias after enocuntering a literal

I must've got something wrong with the iteration pattern
- `"literal"` is selected as it is the last named node (inside `fill_null`)
- But the name in there shouldn't be prioritized if there was a `alias` or `col` earlier

tests/plan/meta_test.py
dangotbanned: Resolved in ([test: Backcompat `len`, `nth`](https://github.com/narwhals-dev/narwhals/pull/2572/commits/b79181a0904e090798b96b0292bf53e09e7d7ef3))

tests/plan/meta_test.py
dangotbanned: Original case fixed in (https://github.com/narwhals-dev/narwhals/pull/2572/commits/54d078113814e84e815cdbbaa380fe3b6ffca35d)

Realised the problem was pretty widespread, but feeling confident as of (https://github.com/narwhals-dev/narwhals/pull/2572/commits/bd5c33f6cf843009f36bd8feb15999e6a8d17511)

**TIL**: The first `.then(...)` clause in a `when-then-otherwise` is where the column name comes from 🤯

narwhals/_plan/dummy.py
```diff
@@ -0,0 +1,1039 @@
...
+        return self.from_native(native, self.name, version=self.version)
+
+    def alias(self, name: str) -> Self:
+        return self.from_native(self.native, name, version=self.version)
+
+    def __len__(self) -> int:
+        return len(self.native)
+
+    def to_list(self) -> list[t.Any]:
+        raise NotADirectoryError
```
dangotbanned: wow, autocomplete truly beefed it on that one 😭 

```suggestion
    def to_list(self) -> list[t.Any]:
        raise NotImplementedError
```

dangotbanned:
aha found you!

- https://github.com/apache/arrow/issues/21761
- https://github.com/narwhals-dev/narwhals/actions/runs/16103334673/job/45435305338?pr=2572

narwhals/_plan/expr_expansion.py
```diff
@@ -0,0 +1,483 @@
...
+            parent = origin.expr
+            leaf_name_or_err = meta.get_single_leaf_name(parent)
+            if not isinstance(leaf_name_or_err, str):
+                raise leaf_name_or_err
+            alias = origin.function(leaf_name_or_err)
+            return Alias(expr=parent, name=alias)
+        else:
+            msg = "`keep`, `suffix`, `prefix` should be last expression"
+            raise InvalidOperationError(msg)
+    return origin
```
dangotbanned: > [!IMPORTANT]
> Need to review specifically how this bit has changed for `polars>=1.32.0`

### Related
- https://github.com/pola-rs/polars/pull/23351
- https://github.com/narwhals-dev/narwhals/pull/2572#discussion_r2127108991
- https://github.com/pola-rs/polars/issues/23765
- #2898


### `expr_expansion.rs` changes
- https://github.com/pola-rs/polars/pull/23351/files#diff-3efddc5204834a1cc0a622cf46403be4707500b040e798bdf9e076e27570566f

narwhals/_plan/_expansion.py
```diff
@@ -0,0 +1,371 @@
...
+                multiple_columns = True
+            elif isinstance(e, Nth):
+                has_nth = True
+            elif isinstance(e, All):
+                has_wildcard = True
+            elif isinstance(e, SelectorIR):
+                has_selector = True
+            elif isinstance(e, Exclude):
+                has_exclude = True
+        return ExpansionFlags(
```
dangotbanned: ### TODO(perf): Optimize `ExpansionFlags.from_ir`

At least until switching over to (https://github.com/pola-rs/polars/pull/23351), there's a simple optimization here.

The first branch should narrow to:

```py
if isinstance(e, (_ColumnSelection, SelectorIR)):
```

Then all unrelated cases skip 4 additional branches, by moving the current 5 inside that:

```py
            if isinstance(e, (_ColumnSelection, SelectorIR)):
                if isinstance(e, (Columns, IndexColumns)):
                    multiple_columns = True
                elif isinstance(e, Nth):
                    has_nth = True
                elif isinstance(e, All):
                    has_wildcard = True
                elif isinstance(e, SelectorIR):
                    has_selector = True
                elif isinstance(e, Exclude):
                    has_exclude = True
```

narwhals/_plan/_expansion.py
```diff
@@ -0,0 +1,372 @@
...
+            raise column_index_error(index, names)
+        yield names[index]
+
+
+def expand_column_selection(
+    origin: ExprIR, tp: type[_ColumnSelection], /, names: Iterable[str], exclude: Excluded
+) -> Iterator[ExprIR]:
+    for name in names:
+        if name not in exclude:
+            yield rewrite_special_aliases(replace_with_column(origin, tp, name))
```
dangotbanned: > [!IMPORTANT]
> This is causing duplicates whenever a `_ColumnSelection` appears in a `tuple[ExprIR, ...]` and expands to 2 or more column names

- `over(*partition_by)`
- `over(order_by=...)`
- `sort_by(*by)`


https://github.com/narwhals-dev/narwhals/blob/dd62ae8ec5f9d2768916d0f138359720ca9ad3d0/tests/plan/over_test.py#L73-L99

PR #2578, opened 2025-05-20
chore: Add `_typing_compat` module

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Mentioned in (https://github.com/narwhals-dev/narwhals/pull/2391#issuecomment-2832442331)
- Needed again for (#2572)

## Checklist

- [ ] Tests added
- [x] Documented the changes

**Module doc preview**

image

---

I was [originally expecting](https://github.com/narwhals-dev/narwhals/pull/2391#issuecomment-2832442331) we'd need to add some config for `ruff` and `mypy`. 
Haven't had any issues without it, so I'm not sure we need it?

### DIFF
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -26,9 +26,9 @@ from narwhals._translate import FromNative
 from narwhals._translate import NumpyConvertible
 from narwhals._translate import ToNarwhals
 from narwhals._translate import ToNarwhalsT_co
+from narwhals._typing_compat import deprecated
 from narwhals.utils import Version
 from narwhals.utils import _StoresNative
-from narwhals.utils import deprecated
 from narwhals.utils import is_compliant_series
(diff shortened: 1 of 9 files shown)


dangotbanned:
To expand on the link to (#2572) some more ...

**Ideally**, it would be good to (*statically*) know the type of every node in an expression.
**Pragmatically**, it would be good to *preserve* the static types where we can - but fall back to a base class (default) when we can't.

There's a big benefit to knowing as much information of the graph as possible - to avoid needing `assert`, `isinstance, `cast`, etc littered everywhere

MarcoGorelli (approved):
nice, thanks @dangotbanned

PR #2585, opened 2025-05-21
chore: Use shorter imports

MarcoGorelli:
This simple ruff config change reduces package size by 2.2% 😱 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -38,8 +38,6 @@ jobs:
         run: |
           cd altair
           uv pip install -e ".[dev, all]" --system
-          # temporary pin to get ci green
-          uv pip install "numpy<2.2" --system
       - name: install-narwhals-dev
         run: |
           uv pip uninstall narwhals --system
@@ -186,10 +184,7 @@ jobs:
           UV_SYSTEM_PYTHON: 1
         run: |
           cd py-shiny
-          # temporary pin to get ci green
           make narwhals-install-shiny
-          uv pip install "pyflakes<3.3.0" "flake8<7.2.0" --system
-          uv pip install "plotly<6.1.0" --system
       - name: install-narwhals-dev
         run: |
           uv pip uninstall narwhals --system
@@ -280,6 +275,8 @@ jobs:
         run: |
           uv pip uninstall narwhals --system
           uv pip install -e . --system
+          # temporary pin to get CI green
+          uv pip install "polars<1.30"
       - name: show-deps
         run: uv pip freeze
       - name: Run pytest
@@ -356,8 +353,6 @@ jobs:
       - name: install-deps
         run: |
           cd hierarchicalforecast
-          # temporary pin to get ci green
-          uv pip install "IPython<9" --system
           uv pip install --system ".[dev]"
       - name: install-narwhals-dev
         run: |
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
     - id: ruff
       args: [--fix]
     - id: ruff
+      name: check-docstrings
       alias: check-docstrings
       entry: python utils/check_docstrings.py
 - repo: https://github.com/codespell-project/codespell
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -2,86 +2,90 @@ from __future__ import annotations
 
 import typing as _t
 
-from narwhals import dependencies
-from narwhals import dtypes
-from narwhals import exceptions
-from narwhals import selectors
-from narwhals.dataframe import DataFrame
-from narwhals.dataframe import LazyFrame
-from narwhals.dtypes import Array
-from narwhals.dtypes import Binary
-from narwhals.dtypes import Boolean
-from narwhals.dtypes import Categorical
-from narwhals.dtypes import Date
-from narwhals.dtypes import Datetime
-from narwhals.dtypes import Decimal
-from narwhals.dtypes import Duration
-from narwhals.dtypes import Enum
-from narwhals.dtypes import Field
-from narwhals.dtypes import Float32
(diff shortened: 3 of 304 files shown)


FBruzzesi:
@MarcoGorelli since you spoiled me in the first place by introducing me to the `force-single-line = true` option, I am not a big fan of this change. FWIW, it makes the diff's harder to read. As an example consider the following diffs representing the same change:

```diff
from typing import TYPE_CHECKING
from typing import Any
- from typing import Callable
from typing import Sequence
from typing import cast
```

vs 

```diff
- from typing import TYPE_CHECKING, Any, Callable, Sequence, cast
+ from typing import TYPE_CHECKING, Any, Sequence, cast
```

which one is simpler to read?

(of course it can get much more complex)

---

> This simple ruff config change reduces package size by 2.2% 😱

Is it worth it?

I think increasing the line length to something between 100-120 might lead to a similar or even larger package size reduction 😅

MarcoGorelli:
IMHO yes, as we're adding backends the size is creeping up

> I think increasing the line length to something between 100-120 might lead to a similar or even larger package size reduction 😅

sure, but that does make it harder to edit code with multiple buffers open

FBruzzesi:
> IMHO yes, as we're adding backends the size is creeping up

Should we start considering plugins/extensions?

e.g. `narwhals[daft]` installing `narwhals-daft` which implement the compliant class outside this repo, and here we simply add the entrypoint for it?

MarcoGorelli:
Possibly yes, we should consider that too

> which one is simpler to read?

Written like this, I do see your point 👍 But on GitHub, it will highlight in green the new parts, and in red the removed parts, so that really helps with readibility
I'm also not sure that the imports diff is meaningful in most cases

FBruzzesi:
Thanks for the concern @MarcoGorelli - It's all good

> How opposed are you to this change? Is it bearable, or strongly opposed?

Not sure if there is an option to have it in multiple lines anyway (could that be a compromise?) e.g.:

```py
from typing import (
    TYPE_CHECKING,
    Any,
    Sequence,
    cast,
)
```
although it would fit on a single line. Or would this neutralize the entire size reduction? 🙈

> though arguably they make it harder to keep multiple tabs open at the same time. Anyway, we can do that separately if desired

On a laptop that's true - if you have a bigger screen I find 100-120 line length to be the right size typically

MarcoGorelli:
cool, thanks! shipping then before merge conflicts 😄 🙈

MarcoGorelli:
The only manual changes here are:
```diff
$ git diff shorter-imports 
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index 9e7a69299..4dabb99b4 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,6 @@ repos:
     - id: ruff
       args: [--fix]
     - id: ruff
-      name: check-docstrings
       alias: check-docstrings
       entry: python utils/check_docstrings.py
 - repo: https://github.com/codespell-project/codespell
diff --git a/narwhals/_arrow/dataframe.py b/narwhals/_arrow/dataframe.py
index fda195805..3785fe9fd 100644
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -49,10 +49,10 @@ if TYPE_CHECKING:
     from narwhals._arrow.expr import ArrowExpr
     from narwhals._arrow.group_by import ArrowGroupBy
     from narwhals._arrow.namespace import ArrowNamespace
-    from narwhals._arrow.typing import (  # type: ignore[attr-defined]
+    from narwhals._arrow.typing import (
         ChunkedArrayAny,
-        Mask,
-        Order,
+        Mask,  # type: ignore[attr-defined]
+        Order,  # type: ignore[attr-defined]
     )
     from narwhals._compliant.typing import CompliantDataFrameAny, CompliantLazyFrameAny
     from narwhals._translate import IntoArrowTable
diff --git a/narwhals/_arrow/group_by.py b/narwhals/_arrow/group_by.py
index c656cba4b..61a20e6ef 100644
--- a/narwhals/_arrow/group_by.py
+++ b/narwhals/_arrow/group_by.py
@@ -14,9 +14,9 @@ from narwhals.utils import generate_temporary_column_name
 if TYPE_CHECKING:
     from narwhals._arrow.dataframe import ArrowDataFrame
     from narwhals._arrow.expr import ArrowExpr
-    from narwhals._arrow.typing import (  # type: ignore[attr-defined]
-        AggregateOptions,
-        Aggregation,
+    from narwhals._arrow.typing import (
+        AggregateOptions,  # type: ignore[attr-defined]
+        Aggregation,  # type: ignore[attr-defined]
         Incomplete,
     )
     from narwhals._compliant.group_by import NarwhalsAggregation
diff --git a/narwhals/_arrow/series.py b/narwhals/_arrow/series.py
index 0f173eb3c..ed6f82e49 100644
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -52,14 +52,14 @@ if TYPE_CHECKING:
 
     from narwhals._arrow.dataframe import ArrowDataFrame
     from narwhals._arrow.namespace import ArrowNamespace
-    from narwhals._arrow.typing import (  # type: ignore[attr-defined]
+    from narwhals._arrow.typing import (
         ArrayAny,
         ArrayOrChunkedArray,
         ArrayOrScalar,
         ChunkedArrayAny,
         Incomplete,
         NullPlacement,
-        Order,
+        Order,  # type: ignore[attr-defined]
         TieBreaker,
         _AsPyType,
         _BasicDataType,
```

Issue #2165, opened 2025-03-07
api: check whether we should get the connection time zone for duckdb/pyspark


MarcoGorelli:
Currently, for pyspark, we map `TimestampType` to `Datetime(time_zone='UTC')`. Is this right? In DuckDB, you can get the time zone of the connection. Is there anything similar for PySpark? Should we be using that?

MarcoGorelli:
We really need to make a decision on this one

Currently, for duckdb, it's possible to get the time zone with:
```
duckdb.sql("""select current_setting('timezone')""").fetchone()[0]
```
This is a bit more expensive than I'd like it to be:
```python
In [14]: %timeit duckdb.sql("""select current_setting('timezone')""").fetchone()[0]
749 μs ± 27.6 μs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

---

## Proposed solution

When computing `.schema`, we don't compute the time zone. We just either store `None` or `Some(Deferred)`. We only evaluate the time zone if the user:
- asks for it, e.g. they execute something like `df.collect_schema()['a'].time_zone`
- prints the schema. In that case, it would be good to show it, and it's a bit surprising to me that DuckDB / PySpark don't show it by default

Note that we can't cache this value, as it's possible for users to change the time zone
```python
In [16]: duckdb.sql("""set timezone = 'Europe/Amsterdam'""")

In [17]: duckdb.sql("""select current_setting('timezone')""").fetchone()[0]
Out[17]: 'Europe/Amsterdam'
```

MarcoGorelli:
This would be a better way to check https://github.com/duckdb/duckdb/discussions/17578#discussioncomment-13216782

```
rel1.query('duckdb_settings()', 'select value from duckdb_settings() where name = \'TimeZone\'').fetchone()[0]
```
as then the sql query would just be
```
In [13]: rel2.query('duckdb_settings()', 'select value from duckdb_settings() where name = \'TimeZone\'').sql_query()
Out[13]: 'SELECT "value" FROM duckdb_settings() WHERE ("name" = \'TimeZone\')'
```
without it depending in any way on the original relation

It's still a bit annoying that, given that users can modify the configuration's time zone, we can't cache this

MarcoGorelli:
To make this even more uncomfortable, we currently cache the schema, but if the users are allowed to change it mid-query (as duckdb allows), then our cached schema is no longer accurate

uurrgghh

[...]

PR #2590, opened 2025-05-22
enh: reflect connection time zone in `Datetime` for DuckDB, support `convert_time_zone` and `replace_time_zone` 

MarcoGorelli:
one step towards https://github.com/narwhals-dev/narwhals/issues/2165

Still need to do the same for PySpark and Ibis

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -9,7 +9,7 @@ import duckdb
 from duckdb import FunctionExpression, StarExpression
 
 from narwhals._duckdb.utils import (
-    CachedTimeZone,
+    DeferredTimeZone,
     col,
     evaluate_exprs,
     generate_partition_by_sql,
@@ -219,14 +219,13 @@ class DuckDBLazyFrame(
             # due to Python3.13 failures.
             self._cached_native_schema = dict(zip(self.columns, self.native.types))
 
-        nw_schema = {}
-        cached_time_zone = CachedTimeZone()
-        for column_name, duckdb_dtype in self._cached_native_schema.items():
-            nw_schema[column_name] = native_to_narwhals_dtype(
+        cached_time_zone = DeferredTimeZone()
+        return {
+            column_name: native_to_narwhals_dtype(
                 duckdb_dtype, self._version, self.native, cached_time_zone
             )
-
-        return nw_schema
+            for column_name, duckdb_dtype in zip(self.native.columns, self.native.types)
+        }
 
     @property
     def columns(self) -> list[str]:
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -219,10 +219,10 @@ class DuckDBLazyFrame(
             # due to Python3.13 failures.
             self._cached_native_schema = dict(zip(self.columns, self.native.types))
 
-        cached_time_zone = DeferredTimeZone()
+        deferred_time_zone = DeferredTimeZone(self.native)
         return {
             column_name: native_to_narwhals_dtype(
-                duckdb_dtype, self._version, self.native, cached_time_zone
+                duckdb_dtype, self._version, deferred_time_zone
             )
             for column_name, duckdb_dtype in zip(self.native.columns, self.native.types)
         }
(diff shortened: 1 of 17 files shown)


narwhals/_duckdb/dataframe.py
```diff
@@ -212,23 +214,40 @@ def filter(self, predicate: DuckDBExpr) -> Self:
...
     def schema(self) -> dict[str, DType]:
-        if self._cached_schema is None:
+        if self._cached_native_schema is None:
             # Note: prefer `self._cached_schema` over `functools.cached_property`
             # due to Python3.13 failures.
-            self._cached_schema = {
-                column_name: native_to_narwhals_dtype(duckdb_dtype, self._version)
-                for column_name, duckdb_dtype in zip(
-                    self.native.columns, self.native.types
+            self._cached_native_schema = dict(zip(self.columns, self.native.types))
```
MarcoGorelli, line 220: rather than caching the Narwhals schema (which may change, if the user modifies the connection time zone! !!!), we cache the native schema (which is the more potentially expensive one to calculate)

tests/translate/from_native_test.py
```diff
@@ -547,4 +547,4 @@ def test_pyspark_connect_deps_2517() -> None:  # pragma: no cover
 
     spark = SparkSession.builder.getOrCreate()  # pyright: ignore[reportAttributeAccessIssue]
     # Check this doesn't raise
-    nw.from_native(spark.createDataFrame([{"a": 1}]))
+    nw.from_native(spark.createDataFrame([{"a": 1}]))  # pyright: ignore[reportCallIssue]
```
MarcoGorelli, line 550:
```suggestion
    nw.from_native(spark.createDataFrame([{"a": 1}]))  # pyright: ignore[reportCallIssue, reportArgumentType]
```

narwhals/_duckdb/expr_dt.py
```diff
@@ -124,13 +125,52 @@ def _truncate(expr: Expression) -> Expression:
...
+                    raise NotImplementedError(msg)
+                return native_series_list
+
+            return self._compliant_expr.__class__(
+                func,
+                evaluate_output_names=self._compliant_expr._evaluate_output_names,
+                alias_output_names=self._compliant_expr._alias_output_names,
+                backend_version=self._compliant_expr._backend_version,
+                version=self._compliant_expr._version,
+            )
```
FBruzzesi, line 174: Pretty sad that the only difference with `convert_time_zone` is the name of the function in the error message 👀

FBruzzesi (commented):
Thanks @MarcoGorelli - as an idea to avoid increasing the arguments to pass to `native_to_narwhals_dtype`, `_time_zone` could be a DuckDBLazyFrame (cached) property and avoid passing `df` as an argument.

It would kinda go into the direction that we are trying to think/refactor in other PRs as well. WDYT?

MarcoGorelli:
Thanks for your review!

> _time_zone could be a DuckDBLazyFrame (cached) property and avoid passing df as an argument.

I did think about this, but the issue is that users can change the connection time zone, so then that cached property would no longer be accurate:
```python
df = nw.from_native(rel)
print(df.collect_schema())
rel.query("set timezone = 'Asia/Kolkata'")
print(df.collect_schema())  # should change to reflect new time zone
df.select(nw.col('a').dt.replace_time_zone(None))
```

FBruzzesi:
> I did think about this, but the issue is that users can change the connection time zone, so then that cached property would no longer be accurate:

Oh I see, the underlying duckdb object would remain the same, the change happens as a "config" 🤔 Thanks for the quick example

MarcoGorelli:
> and avoid passing df as an argument.

however, we can still avoid passing `df` around as an argument by making a `DeferredTimeZone` class 😎  Happy with this approach?

FBruzzesi (approved):
>  we can still avoid passing df around as an argument by making a DeferredTimeZone class 😎 Happy with this approach?

Thanks for bearing with me - pretty happy with this change 👌🏼

BooGIF

MarcoGorelli:
thanks for your review, much better now!

PR #2598, opened 2025-05-23
feat: Add support for `Expr.str.zfill`

### LINKED ISSUE #2596: [Enh]: str zfill

I would like to use narwhals for the latent-calendar project. However, it uses zfill

https://github.com/williambdean/latent-calendar/blob/07ffd832188098af21a469470bb988560d7b6e80/latent_calendar/transformers.py#L155

pad a string

use the pandas, polars, spark, etc functionality

looked through the docs

_No response_

williambdean:
This adds `zfill` for both Series and Expr

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2596

## Checklist

- [x] Tests added
- [x] Documented the changes


williambdean:
Working through this: 

```
check-api-reference....................................................................Failed
- hook id: check-api-reference
- exit code: 1

Series.str: not documented
{'zfill'}
Expr.str: not documented
{'zfill'}
```

MarcoGorelli:
ah yeah you need to add it to the `expr_str.md` / `series_str.md` pages

williambdean:
Thanks. On my way now. 
Will add some tests when I log on later

williambdean:
Is this a bug in pandas? 

```python
import pandas as pd
import pyarrow as pa

pd.Series(["A", "AB", "ABC"], dtype=pd.ArrowDtype(pa.string())).str.zfill(3)
```

Trackback:

```
  File "/Users/will/GitHub/various/narwhals/.venv/lib/python3.12/site-packages/pandas/core/strings/accessor.py", line 137, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/will/GitHub/various/narwhals/.venv/lib/python3.12/site-packages/pandas/core/strings/accessor.py", line 1818, in zfill
    result = self._data.array._str_map(f)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ArrowExtensionArray' object has no attribute '_str_map'. Did you mean: '_str_pad'?
```

tests/expr_and_series/str/zfill_test.py
FBruzzesi, line 1: Could we add a test to check the behavior for dashes?

From [Polars API docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.zfill.html):

_A sign prefix (-) is handled by inserting the padding after the sign character rather than before._

```py
df = pl.DataFrame({"a": [-1, 123, 999999, None]})
df.with_columns(zfill=pl.col("a").cast(pl.String).str.zfill(4))
shape: (4, 2)
┌────────┬────────┐
│ a      ┆ zfill  │
│ ---    ┆ ---    │
│ i64    ┆ str    │
╞════════╪════════╡
│ -1     ┆ -001   │
│ 123    ┆ 0123   │
│ 999999 ┆ 999999 │
│ null   ┆ null   │
└────────┴────────┘
```

FBruzzesi (commented):
Thanks for the contribution @williambdean 🙏🏼

> Is this a bug in pandas?

It looks like one 😅
We might be able to workaround by accessing the underlying pyarrow series - one example of that is the following:

https://github.com/narwhals-dev/narwhals/blob/35ec9fec57a6a932c0435ea6e210303bc4e34bb0/narwhals/_pandas_like/series.py#L1026-L1033)

it might still be worth to report it upstream

tests/expr_and_series/str/zfill_test.py
williambdean: I've created a pandas issue here https://github.com/pandas-dev/pandas/issues/61485

FBruzzesi:
> How do y'all do the CASE WHEN. Are there any examples?

@williambdean sure there are a lot of cases and differ from backend to backend - #2549 has examples for lazy backends

> Or maybe you have a different way of implementing in mind

Are you referring to the [trailing dash comment](https://github.com/narwhals-dev/narwhals/pull/2598#discussion_r2105303786)?
If so, I guess it depends 😅 pandas already does that out of the box - I have no clue for the other backends though

williambdean:
In the cases with no implementation of zfill (which seems like a few), I was using lpad. I was thinking on aiming for 
CASE WHEN starts_with(col, "+") THEN ... WHEN starts_with("-") THEN ... ELSE <lpad-implementation> END

williambdean:
Awesome. Thanks for the lead. Will look into that example

MarcoGorelli:
IMHO it's OK to xfail pandas here, it looks like the rest is mostly ready? we can always make a follow-up to work around the pandas bug if necessary

williambdean:
Sounds good. The case that I will finish with is the behavior with leading + or - for the implementations that are using lpad

williambdean:
I've added  a test case for the leading +. zfill from python does "+1".zfill(3) = "+01". But it seems like the polars eager does something different 

These are the current failures with the latest commits: 

```

===================================================== short test summary info =====================================================
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[pyarrow] - AssertionError: Mismatch at index 0: 0-1 != -01
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[pyarrow] - AssertionError: Mismatch at index 0: 0-1 != -01
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[pandas[pyarrow]] - AttributeError: 'ArrowExtensionArray' object has no attribute '_str_map'. Did you mean: '_str_pad'?
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[polars[eager]] - AssertionError: Mismatch at index 1: 0+1 != +01
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[polars[eager]] - AssertionError: Mismatch at index 1: 0+1 != +01
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[pandas[pyarrow]] - AttributeError: 'ArrowExtensionArray' object has no attribute '_str_map'. Did you mean: '_str_pad'?
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[ibis] - AssertionError: Mismatch at index 0: 0-1 != -01
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[sqlframe] - AssertionError: Mismatch at index 0: 0-1 != -01
```

Which behavior is suppose to match?

MarcoGorelli:
Probably good to match polars

williambdean:
well pandas and polars don't match 😆

MarcoGorelli:
I don't really understand the pandas output, but we can standardise on what polars does and note the pandas difference in the docstring? This feels like really edge case behavior anyway

williambdean:
I've adjusted the tests and the implementations to match polars behavior. I have a case for pandas.

I'm new to arrow and still working through that implmenetation. Open to pointers!

tests/expr_and_series/str/zfill_test.py
```diff
@@ -0,0 +1,39 @@
...
+data = {"a": ["-1", "+1", "1", "12", "123", "99999", "+9999", None]}
+polars_expected = {"a": ["-01", "0+1", "001", "012", "123", "99999", "+9999", None]}
+pandas_expected = {"a": ["-01", "+01", "001", "012", "123", "99999", "+9999", None]}
+
+
+def skip_pandas_pyarrow(constructor: Constructor | ConstructorEager) -> None:
+    name: str = constructor.__name__
+    if name == "pandas_pyarrow_constructor":
+        # Skip this test for pandas with pyarrow, as it doesn't support str.zfill
+        raise pytest.skip()
```
FBruzzesi, line 17:
```suggestion
    if name == "pandas_pyarrow_constructor":
        reason = (
            "pandas with pyarrow backend doesn't support str.zfill, see "
            "https://github.com/pandas-dev/pandas/issues/61485"
        )
        raise pytest.skip(reason=reason)
```

narwhals/expr_str.py
```diff
@@ -447,3 +447,19 @@ def to_lowercase(self) -> ExprT:
...
+        """Transform string to zero-padded variant.
+
+        Arguments:
+            width: The desired length of the string after padding. If the length of the
+                string is greater than `width`, no padding is applied.
+                If `width` is less than 0, no padding is applied.
+
+        Returns:
+            A new expression.
+
```
FBruzzesi, line 461: Could you add an example for expr case as well?

FBruzzesi (commented):
Thanks @williambdean 

I really like the workarounds to align with polars behavior 👌🏼 (Although at this point I wonder if polars is the odd one).

I have a couple of nitpicks:

1. Could we document in both Expr and Series the difference/unique behavior of pandas? Here is an [example of a similar warning](https://github.com/narwhals-dev/narwhals/blob/46eb2a8cf42f73ca024c11ea20dc8d0742ab587b/narwhals/series.py#L2117-L2120)
2. In duckdb and spark-like the inner function if called `func`, in ibis is called `_zfill`. I think we don't have any rule and we are definitly not strict on that, but ... can we make them the same here 😂?

tests/expr_and_series/str/zfill_test.py
FBruzzesi: Think a second longer on this one, I think we should explicitly xfail (not xskip) it, so that eventually in pandas 3.x it would be fixed and we will automatically spot it

williambdean:
Thanks for the review. I will add those changes. 

Do you know how to do the arrow implementation? 

<details>

<summary>Arrow test failure</summary>

```
        length = pc.utf8_length(self.native)
        less_than_width = pc.less(length, width)
        starts_with_minus = pc.equal(self.slice(0, 1).native, lit("-"))

        result = pc.case_when(
            [
                (
>                   starts_with_minus & less_than_width,
                    pc.utf8_lpad(
                        self.slice(1, None).native, width - 2, padding="0"
                    ).concat(lit("-")),
                ),
                (less_than_width, pc.utf8_lpad(self.native, width=width, padding="0")),
            ],
            self.native,
        )
E       TypeError: unsupported operand type(s) for &: 'pyarrow.lib.ChunkedArray' and 'pyarrow.lib.ChunkedArray'

narwhals/_arrow/series_str.py:72: TypeError
===================================================== short test summary info =====================================================
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[pyarrow] - TypeError: unsupported operand type(s) for &: 'pyarrow.lib.ChunkedArray' and 'pyarrow.lib.ChunkedArray'
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[pyarrow] - TypeError: unsupported operand type(s) for &: 'pyarrow.lib.ChunkedArray' and 'pyarrow.lib.ChunkedArray'
```

</details>

williambdean:
Thanks for the help @dangotbanned !

williambdean:
Ha, seems like the pandas behavior has even changed in the versions that are being tested? Confusing. https://github.com/narwhals-dev/narwhals/actions/runs/15283097790/job/42986670950?pr=2598

The edits you made worked for me locally, but seeing some pyarrow failures: https://github.com/narwhals-dev/narwhals/actions/runs/15283097790/job/42986670950?pr=2598

And would need to implement the dask behavior as well: 
https://github.com/narwhals-dev/narwhals/actions/runs/15283097790/job/42986670953?pr=2598

williambdean:
It seems to just be the https://github.com/narwhals-dev/narwhals/actions/workflows/extremes.yml workflow now.

MarcoGorelli (commented):
thanks!

i've left some comments

will open an issue with Polars to check if the behaviour is intentional. re-reading this, I don't think it is, it might just be a bug

MarcoGorelli:
i'm not too bothered about what happens for `'++1'`, but for `'+1'` at least i think the Polars output doesn't look correct

MarcoGorelli:
ok this is marked as "accepted" in Polars https://github.com/pola-rs/polars/issues/22983

So I'd say:
- for `'+1'.zfill(3)`, the output should be `'+01'`. We may need to xfail for Polars for now
- for `'++1'.zfill(4)`, i'd say that the behaviour is unspecified and we shouldn't worry about what happens

sorry for contradictory instructions from me @williambdean , my bad here

MarcoGorelli:
Fixed in Polars now https://github.com/pola-rs/polars/pull/22985#event-17849747949

MarcoGorelli:
Yeah if it's feasible, sure!

williambdean:
I have some todos in the comment above. Is it fine to skip some of the old pandas and pyarrow versions in test suite. Thoughts?

williambdean:
I am seeing some failures that look unrelated in my eyes. Whats happening with those smoke tests?

MarcoGorelli:
thanks for updating! the unrelated test failures have been resolved, it should all be green now

some failures seem related though:
```
__________ [doctest] narwhals.series_str.SeriesStringNamespace.zfill ___________
413 
414 Returns:
415     A new Series with strings padded with zeros on the left.
416 
417 Examples:
418     >>> import pandas as pd
419     >>> import narwhals as nw
420     >>> s_native = pd.Series(["+1", "-23", "456", "123456"])
421     >>> s = nw.from_native(s_native, series_only=True)
422     >>> s.str.zfill(5).to_native()
Differences (unified diff with -expected +actual):
    @@ -1,5 +1,5 @@
    -0    000+1
    -1    00-23
    -2    00456
    -3   123456
    +0     +0001
    +1     -0023
    +2     00456
    +3    123456
     dtype: object
```

williambdean:
Changed the example in the docstring! Thank you for the help with that.

williambdean:
Last thing: How do we want to handle the previous pandas behavior? Would I xfail for certain versions?

tests/expr_and_series/str/zfill_test.py
```diff
@@ -0,0 +1,39 @@
...
+    if name in {"pandas_pyarrow_constructor", "modin_pyarrow_constructor"}:
+        reason = (
+            "pandas with pyarrow backend doesn't support str.zfill, see "
+            "https://github.com/pandas-dev/pandas/issues/61485"
+        )
+        raise pytest.skip(reason=reason)
+
+    if "polars" in name:
+        reason = "Polars zfill behavior is different from pandas at the moment."
+        raise pytest.skip(reason=reason)
```
MarcoGorelli: can we in-line this (like we do for other tests)?

and also, for Polars, we should test it. in `PolarsExprStringNamespace`, we could implement this for Polars with version checks, kind of like we do for `PolarsExprListNamespace.len`, where for old Polars you replicate the logic with `pad` / `when/then`, and for newer versions you just use `str.zfill`

MarcoGorelli (commented):
thanks @williambdean for updating, this is getting closer 💪 

how old are the pandas versions that fail? if they're from say early 1.* versions, then i think it's ok to just skip those versions in the tests

MarcoGorelli:
currently CI doesn't run automatically because this is your first pr, if you wanted to make a separate little pr first to fix a typo or something so that ci runs automatically, feel freet o

williambdean:
The zfill bug fix was add with pandas==1.5.0
See the release notes here: https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#strings

tests/expr_and_series/str/zfill_test.py
williambdean: What do you mean by in-line? Can you share an example

MarcoGorelli:
thanks - i'd say fine to skip the test then for pandas pre 1.5

williambdean:
Starting with the edits. Thanks for hanging in there with me

I'm getting this error while running tests locally now:

```

    from duckdb import FunctionExpression

    from narwhals._duckdb.utils import lit, when
>   from narwhals.utils import not_implemented
E   ImportError: cannot import name 'not_implemented' from 'narwhals.utils' (/Users/will/github/various/narwhals/narwhals/utils.py)

narwhals/_duckdb/expr_str.py:8: ImportError
================================================== short test summary info ===================================================
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[ibis] - ImportError: cannot import name '_is_naive_format' from 'narwhals.utils' (/Users/will/github/various/narwhals/narwhals/ut...
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[duckdb] - ImportError: cannot import name 'not_implemented' from 'narwhals.utils' (/Users/will/github/various/narwhals/narwhals/uti...
```

EDIT: Changed the import!

narwhals/expr_str.py
williambdean: Good call. Done!

narwhals/_polars/expr.py
```diff
@@ -335,6 +335,26 @@ class PolarsExprStringNamespace:
     def __init__(self, expr: PolarsExpr) -> None:
         self._compliant_expr = expr
 
+    def zfill(self, width: int) -> PolarsExpr:
+        native_expr = self._compliant_expr.native
+        native_result = native_expr.str.zfill(width)
+
+        if self._compliant_expr._backend_version <= (1, 30, 0):
```
williambdean, line 346: This number is based on [this PR](https://github.com/pola-rs/polars/pull/22985) being merged and the [current version in release](https://github.com/pola-rs/polars/releases/tag/py-1.30.0) is 1.30.0

Let me know if it should be changed!

narwhals/expr_str.py
```diff
@@ -447,3 +447,40 @@ def to_lowercase(self) -> ExprT:
         return self._expr._with_elementwise_op(
             lambda plx: self._expr._to_compliant_expr(plx).str.to_lowercase()
         )
+
+    def zfill(self, width: int) -> ExprT:
+        """Transform string to zero-padded variant.
+
+        Warning:
+            Different backends handle strings that start with "+" and "-" signs differently:
```
williambdean: Can this be removed? Or altered?

williambdean:
Working through the latest errors: 

```
Using Python 3.8.18 environment at: /opt/hostedtoolcache/Python/3.8.18/x64
pandas==0.25.3
polars==0.20.3
numpy==1.17.5
pyarrow==11.0.0
scipy==1.5.0
scikit-learn==1.1.0
duckdb==1.0.0
```

leads to 

https://github.com/narwhals-dev/narwhals/actions/runs/15656337682/job/44119480876#step:9:663


```
Using Python 3.8.18 environment at: /opt/hostedtoolcache/Python/3.8.18/x64
pandas==1.1.5
polars==0.20.3
numpy==1.17.5
pyarrow==11.0.0
scipy==1.5.0
scikit-learn==1.1.0
duckdb==1.0.0
```

leads to 

https://github.com/narwhals-dev/narwhals/actions/runs/15656337682/job/44119480877#step:10:659


length is not working if it is not an int. Will trying wrapping in `pl.lit`

And the pyarrow one, I'm not sure about:
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[pyarrow] - pyarrow.lib.ArrowInvalid: Negative buffer resize: -64

williambdean:
EDIT:

The polars series error is coming from here: 

https://github.com/narwhals-dev/narwhals/blob/4d70d6afb7a89c347a2b872bc18dd44a7a7328cd/narwhals/_polars/series.py?plain=1#L700

So `pl.lit` would not work and I think the version would have to be skips for polars unless there is an alternative implementation. Please let me know if you have any ideas

williambdean:
Would love to get this to finish line. Thank you all for your time!

williambdean:
Thanks for that feedback. I've gone ahead and address that. Let me know your thoughts. The _arrow is a bit tough and has me running into "too many local variables"

williambdean:
Need to digest the typing failure still. Any tips would be appreciated!

FBruzzesi:
Hey @williambdean I did a commit (https://github.com/narwhals-dev/narwhals/pull/2598/commits/cb0c30ee24e0c3680c8616d3a3d20951e2bcc332) to address both the following:

> The _arrow is a bit tough and has me running into "too many local variables"

I restructured a bit and renamed the variables

> Need to digest the typing failure still. Any tips would be appreciated!

The issue was that `pl.when` generates an expression rather than a series. Since the test was calling that in a select context, it was not failing. The result is of the operation is:

```py
(Pdb) native_result
<Expr ['.when(Series[a]).then(Series[a…'] at 0x7F599DD65430>
```

I moved to `pl.Series.zip_with` to make sure a Series is maintained in the output and changed the series test slightly to check that

narwhals/expr_str.py
FBruzzesi: Behavior should be consistent across all backends now, isn't it?

williambdean: Yes with the * for *before* 1.5 pandas. I will remove

williambdean:
Thanks for those edits @FBruzzesi

narwhals/expr_str.py
williambdean: done

FBruzzesi:
@MarcoGorelli and/or @dangotbanned would you mind giving it another pass? To me it looks ready to be merged 👌🚀

dangotbanned:
> @MarcoGorelli and/or @dangotbanned would you mind giving it another pass? To me it looks ready to be merged 👌🚀

Sure, I'll take a look in a bit @FBruzzesi

narwhals/_arrow/series_str.py
```diff
@@ -60,3 +62,40 @@ def to_uppercase(self) -> ArrowSeries:
...
+
+        conditions = pc.make_struct(
+            pc.and_(
+                starts_with_hyphen, less_than_width
+            ),  # starts with hyphen and less than width
+            pc.and_(
+                starts_with_plus, less_than_width
+            ),  # starts with plus and less than width
+            less_than_width,  # less than width
+        )
```
dangotbanned: I think we're fine without these comments, they just said the same thing again in a different order

```suggestion
        conditions = pc.make_struct(
            pc.and_(starts_with_hyphen, less_than_width),
            pc.and_(starts_with_plus, less_than_width),
            less_than_width,
        )
```

narwhals/_ibis/expr_str.py
```diff
@@ -100,4 +101,30 @@ def to_datetime(self, format: str | None) -> IbisExpr:
         fn = self._to_datetime_naive if _is_naive_format(format) else self._to_datetime
         return self._compliant_expr._with_callable(fn(format))
 
+    def zfill(self, width: int) -> IbisExpr:
+        def func(expr: ir.StringColumn) -> ir.Value:
+            length = expr.length()
+            less_than_width = expr.length() < lit(width)
```
dangotbanned:
```suggestion
            length = expr.length()
            less_than_width = length < lit(width)
```

narwhals/_ibis/expr_str.py
```diff
@@ -100,4 +101,30 @@ def to_datetime(self, format: str | None) -> IbisExpr:
...
+                    starts_with_minus & less_than_width,
+                    (substring.lpad(width - 1, zero).lpad(width, hyphen)),
+                ),
+                (
+                    starts_with_plus & less_than_width,
+                    (substring.lpad(width - 1, zero).lpad(width, plus)),
+                ),
+                (less_than_width, expr.lpad(width, zero)),
+                else_=expr,
+            )
```
dangotbanned:
```suggestion
            substring = expr.substr(one, sub_length).lpad(width - 1, zero)
            return ibis.cases(
                (starts_with_minus & less_than_width, (substring.lpad(width, hyphen))),
                (starts_with_plus & less_than_width, (substring.lpad(width, plus))),
                (less_than_width, expr.lpad(width, zero)),
                else_=expr,
            )
```

narwhals/_polars/series.py
```diff
@@ -686,6 +686,25 @@ class PolarsSeriesStringNamespace:
...
+            less_than_width = length < width
+
+            other = (
+                native_series.str.slice(1, length)
+                .str.zfill(width - 1)
+                .str.pad_start(width, "+")
+            )
+            native_result = native_result.zip_with(
+                mask=~(starts_with_plus & less_than_width), other=other
+            )
```
dangotbanned: Might we be able to do this a bit lazier?

There's a lot of `pl.Series` methods that dispatch to `pl.Expr` on `pl.DataFrame | pl.LazyFrame` under the hood.
I feel like we could do something like that for performance, but also re-use our `_backend_version` branching logic from `PolarsExpr`.

Currently we evaluate at least 6 intermediate `pl.Series`, which to me seems like too much for an edge case 🤔

dangotbanned (commented):
Thanks all, (especially @williambdean!)

I've got a few suggestions and an idea for `PolarsSeries`.
I think we're *almost* there 🎉

williambdean:
I've merged in those suggestions. Thanks @dangotbanned

narwhals/_polars/series.py
dangotbanned: I guess we'll see if this holds up in CI 🤞 

[perf: Reuse `PolarsExpr` impl in `PolarsSeries`](https://github.com/narwhals-dev/narwhals/pull/2598/commits/4842eb79a69ca7a6a7d0ceb31c6652a165ba93e7)

dangotbanned (approved):
Thanks for the PR @williambdean 🥳

tests/expr_and_series/str/zfill_test.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def skip_pandas_pyarrow(constructor: Constructor | ConstructorEager) -> None:
+    name: str = constructor.__name__
+    if name in {"pandas_pyarrow_constructor", "modin_pyarrow_constructor"}:
+        reason = (
+            "pandas with pyarrow backend doesn't support str.zfill, see "
+            "https://github.com/pandas-dev/pandas/issues/61485"
+        )
+        raise pytest.skip(reason=reason)
```
MarcoGorelli: we usually use `skip` for things which don't and will never pass

here, however, we can expect that pandas[pyarrow] will eventually support zfill, so it think we should use `request.applymarker(pytest.mark.xfail)`

MarcoGorelli (commented):
really solid work, thanks @williambdean for sticking to this, and others for reviewing!

just a minor comment, then we'll get this into the next release

tests/expr_and_series/str/zfill_test.py
dangotbanned: @MarcoGorelli I keep forgetting to suggest we add more tools like this in `tests.utils.py`


But trimmed down to just the constructor test, like

https://github.com/narwhals-dev/narwhals/blob/99222f675dfe874033ac837340a98e7641de5f3b/tests/utils.py#L154-L156

williambdean: I will add in a similar fashion to is_pyarrow_windows_no_tzdata

EDIT: updated

FBruzzesi:
Thanks again @williambdean 🙌🏼

UpVoteDownVoteGIF (2)

williambdean:
Thank you all for the assistance!

Issue #2492, opened 2025-05-04
Simplify misc


dangotbanned:
Some odds and ends that aren't in a category (yet?)

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/utils.py#L345

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/utils.py#L1148

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/translate.py#L363

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/translate.py#L634-L636

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/translate.py#L776

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_pandas_like/utils.py#L599-L601

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_namespace.py#L289-L292

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_expression_parsing.py#L363

dangotbanned:
@FBruzzesi for the `_namespace.py` one I have an API change idea, but haven't got around to typing it up.

Let me know if you start on it and I'll pull my act together 😂

FBruzzesi:
Hey @dangotbanned - no I didn't take a look at it yet, so feel free to go ahead with your idea :)

Issue #1940, opened 2025-02-04
Make `import_optional_dependency` util


MarcoGorelli:
We should make such a util, which can take an optional argument specifying the minimum version. Then we can raise unified and consistent error messages

mikeweltevrede:
Sounds interesting @MarcoGorelli - could you explain a bit more on this?

MarcoGorelli:
hey @mikeweltevrede ! good to hear from you again!

we have some places where we import optional dependencies inline, e.g.

https://github.com/narwhals-dev/narwhals/blob/5f1f19f542e13dd3efc217f56291b16db88b88d6/narwhals/dataframe.py#L503-L507

We should probably have a helper function which does this, and a custom narwhals error for missing optional dependencies

pandas has something similar (`import_optional_dependency` in their source code) which we could probably use (with attribution, of course)

dangotbanned:
> pandas has something similar (`import_optional_dependency` in their source code) which we could probably use (with attribution, of course)

Here's the `polars` one - which is pretty clean and well-documented

https://github.com/pola-rs/polars/blob/0ca485a88a217eab2d55efbd981dfd68709e63be/py-polars/polars/dependencies.py#L225-L287

dangotbanned:
AFAIK, if we want the imported modules to be understood by a type checker - we'll need *1* function per-import.

# Example
<details><summary><code>_module_not_found</code></summary>
<p>

```py
from __future__ import annotations

from importlib import import_module
from importlib.util import find_spec

from types import ModuleType
from typing import Any
from typing import Literal

def _module_not_found(module_name: str, **kwds: Any) -> ModuleNotFoundError:
    some_logic_transforming_kwds = repr(kwds)
    msg = f"{module_name!r} not found.\n{some_logic_transforming_kwds!r}"
    return ModuleNotFoundError(msg)
```

</p>
</details> 


## `importlib`-only
By default, we know nothing about the the contents of the module.

```py
def import_optional_1(module_name: Literal["polars", "pandas"], **kwds: Any):
    if find_spec(module_name):
        return import_module(module_name)
    raise _module_not_found(module_name=module_name, **kwds)
```

![Image](https://github.com/user-attachments/assets/36efa59e-29e8-493a-b39f-ff79aff058c3)

## Single function
Better than the first option.
But we [can't capture `@overload`(s)](https://github.com/narwhals-dev/narwhals/pull/2051#discussion_r1969753588)

```py
[...]

PR #2620, opened 2025-05-26
refactor: Simplify `Implementation.to_native_namespace`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Another piece of #2492

## Checklist

- [x] Tests added
- [ ] Documented the changes

Not sure how likable this is (see [discord](https://discord.com/channels/1235257048170762310/1276640597284884490/1379466092099145821))


dangotbanned:
> > The phrasing of this tickled me 😂
> 
> I personally don't mind but who knows you guys don't have strong opinions on these hacks 😂
> 
> > @FBruzzesi what if we solved this and closed a related issue in one go?
> > 
> > * [Make `import_optional_dependency` util #1940 (comment)](https://github.com/narwhals-dev/narwhals/issues/1940#issuecomment-2780636862)
> 
> Sure, the polars implementation seems easy to adapt as well. What I am not too keen on is that `MIN_VERSIONS` is defined based on implementation, so that feels a bit circular, but definitly fixable 😂
> 
> > We'd also be able to get typing working (at least for the individual cases)
> 
> To my knowledge, `ModuleType` is not a generic, so how would that work? i.e. what should the following simplified version return?
> 
> ```python
> @overload
> def import_optional(module_name: Literal["pandas"]) -> ?
>     return __import__(module_name)
> ```

At the end of (https://github.com/narwhals-dev/narwhals/issues/1940#issuecomment-2780636862) ...

So you leave each `import_*` function return **un-annotated**, and let the the `Module(...)` part get inferred 😎 

> Image


Then you can have a mapping like:


```py
# or replace `Implementation` with the literal you resolve in `Implementation.value`
_IMPORT_FUNCTIONS: Mapping[Implementation, Callable[..., ModuleType]] = {
    Implementation.PANDAS: import_pandas,
    Implementation.POLARS: import_polars,
    ...: ...,
}
```

narwhals/dependencies.py
```diff
@@ -439,6 +442,105 @@ def is_narwhals_series_int(ser: Any | Series[IntoSeriesT]) -> TypeIs[Series[Into
     return is_narwhals_series(ser) and ser.dtype.is_integer()
 
 
+def import_polars():  # type: ignore[no-untyped-def] # noqa: ANN202
+    if find_spec("polars"):
+        import polars as pl
+
+        return pl
+
+    raise module_not_found(module_name="polars")
```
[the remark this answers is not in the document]
dangotbanned: I took a look at `pyright-cov` and it seems like supporting exclusions would complicate things a bit

https://github.com/MarcoGorelli/pyright-cov/blob/592084eae57d037c1a070b594c7897aa74f8deb5/src/pyright_cov.py#L42-L45

@MarcoGorelli here's a very basic solution 😄 

```py

if TYPE_CHECKING:
    from pathlib import Path
    from typing import Any, Callable, Sequence


class Exclude:
    _fn: Callable[[str], bool]

    def __call__(self, name: str, /) -> bool:
        return self._fn(name)

    @classmethod
    def from_patterns(cls, *patterns: str) -> Exclude:
        if patterns:
            return cls._from_patterns(patterns)
        return cls._skip()

    @staticmethod
    def _from_patterns(patterns: Sequence[str], /) -> Exclude:
        import re

        searchers = tuple(re.compile(p).search for p in patterns)

        def fn(name: str, /) -> bool:
            return any(search(name) for search in searchers)

        obj = Exclude.__new__(Exclude)
        obj._fn = fn
        return obj

    @staticmethod
    def _skip() -> Exclude:
        def fn(_: str, /) -> bool:
            return False

        obj = Exclude.__new__(Exclude)
        obj._fn = fn
        return obj


def calculate_coverage_percentage(
    pyright_data: dict[str, Any], *exclude_patterns: str
) -> float:
    symbols = pyright_data["typeCompleteness"]["symbols"]
    if symbols_unknown := [sym for sym in symbols if not sym["isTypeKnown"]]:
        excludes = Exclude.from_patterns(*exclude_patterns)
        excluded = [sym for sym in symbols_unknown if excludes(sym["name"])]
        n_excluded = len(excluded)
        n_symbols = len(symbols) - n_excluded
        n_symbols_unknown = len(symbols_unknown) - n_excluded
        score = (1 - (n_symbols_unknown / n_symbols)) * 100
    else:
        score = 100
    return round(score, 1)


def read_pyright_data(source: str | Path) -> dict[str, Any]:
    import json
    from pathlib import Path

    with Path(source).open("rb") as f:
        return json.load(f)


if __name__ == "main":
    data = read_pyright_data("pyright-cov.json")
    without_exclude = calculate_coverage_percentage(data)
    with_exclude = calculate_coverage_percentage(data, r".*import_.*")

    print(f"Original coverage: {without_exclude}\nAdjusted coverage: {with_exclude}")  # noqa: T201
```

```
Original coverage: 99.3
Adjusted coverage: 100.0
```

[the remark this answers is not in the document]
FBruzzesi: TIL: `pyright-cov` is from Marco 😂😂 I guess I completely missed such news before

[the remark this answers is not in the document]
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2620/commits/ab7161f7f632b41b348f6e637629042fa26a8469)

dangotbanned (approved):
I'm approving now that (https://github.com/narwhals-dev/narwhals/pull/2620#discussion_r2115556375) is resolved - but I'll leave merging *for now* in case anyone else has any thoughts

Thanks @FBruzzesi 🥳

dangotbanned:
The only way I've found to fix this was by removing the `sys.modules` monkeypatching.
But I don't really like that as it means `find_spec` isn't being tested against the case where the module hasn't been imported yet 🤔

https://github.com/narwhals-dev/narwhals/actions/runs/15491689078/job/43618476234?pr=2620

narwhals/utils.py
```diff
@@ -352,10 +356,14 @@ def to_native_namespace(self) -> ModuleType:
...
-            return importer()
-        msg = f"Not supported Implementation: {self}"  # pragma: no cover
-        raise AssertionError(msg)  # pragma: no cover
+        if self is Implementation.UNKNOWN:  # pragma: no cover
+            msg = f"Not supported Implementation: {self}"
+            raise AssertionError(msg)
+
+        importer = _IMPORT_FUNCTIONS.get(
+            self, lambda: import_optional_dependency(self.value, MIN_VERSIONS[self])
+        )
```
FBruzzesi: The default is somehow a hack, yet avoids to bloat the `_IMPORT_FUNCTIONS` dict with same code

FBruzzesi:
@MarcoGorelli @dangotbanned I tried to implement the idea I had mentioned in discord in https://github.com/narwhals-dev/narwhals/pull/2620/commits/96540c329c60c489e20404a7a8ad41d118f1a639, yet apparently everything breaks apart with ibis, so I am happy to revert unless you guys have some other ideas.

The reason to use `from importlib.metadata import version` was mostly because `sqlframe` does not have `__version__` that we can access (that's what polars does).
However, 

- `import_module("ibis")` -> ok
- `import_module("ibis-framework")` -> errors
- `version("ibis")` -> errors
- `version("ibis-framework")` -> ok

dangotbanned:
@FBruzzesi here's a platter of thoughts 😄 

## This PR

### `importlib`
There's a very new [PEP 794 – Import Name Metadata](https://peps.python.org/pep-0794/) that seems relevant to the issue you had here

> * `version("ibis")` -> errors
> * `import_module("ibis-framework")` -> errors

`ibis-framework` is the name of the **distribution**, whereas `ibis` is the name of the **package**.
They're *usually* the same, but sadly not in this case 😞 

https://github.com/python/cpython/blob/24069fbca861a5904ee7718469919e84828f22e7/Lib/importlib/metadata/__init__.py#L980-L987
https://github.com/python/cpython/blob/24069fbca861a5904ee7718469919e84828f22e7/Lib/importlib/__init__.py#L71-L88

### `__version__`
> The reason to use `from importlib.metadata import version` was mostly because `sqlframe` does not have `__version__` that we can access (that's what polars does).

So this detail is already handled in here, you might notice it starts with a call to `to_native_namespace`

<details><summary><code>Implementation._backend_version</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/deaaf76b2a63dd6ad37d20714564712101b4a9f0/narwhals/utils.py#L617-L635

</p>
</details> 


## Minimum versions (*generally*)
I totally forgot to follow up on (https://discord.com/channels/1235257048170762310/1276640597284884490/1379470754051002389)
> I'll continue this over on that other issue so you've got something to come back to

We have two other methods for enforcing version constraints:
- [`utils.validate_backend_version`](https://github.com/narwhals-dev/narwhals/blob/deaaf76b2a63dd6ad37d20714564712101b4a9f0/narwhals/utils.py#L653-L658)
  - For the absolute minimum
- [`@requires.backend_version`](https://github.com/narwhals-dev/narwhals/blob/deaaf76b2a63dd6ad37d20714564712101b4a9f0/narwhals/utils.py#L1838-L1894)
  - For method-specific support

*If* there is another use-case these don't cover, I think it would make the most sense to **improve** those tools.
This PR would be adding a third way of handling package versions - leading to 3 different error messages for the same issue 😔

FBruzzesi:
Thanks @dangotbanned, I am aware of all the code parts you mentioned.

I think `import_optional_dependency` is a much more generic function, for a much more generic use case - at this point we should restrict to be able to import and validate the backends we support, at least for now to address `to_native_namespace`.

> utils.validate_backend_version

Yes, but it requires and implementation & the version already

> So this detail is already handled in here, you might notice it starts with a call to to_native_namespace

We can abstract the body of this

---

I guess we can proceed with one of the two following options:

- Revert to backend specific import functions, but add version validation
- Restrict `import_optional_dependency` to the backends we have, to deal with case by case if needed

in this way we could be able to re-use the above mentioned functions/methods

narwhals/_utils.py
```diff
@@ -615,24 +586,23 @@ def is_sqlframe(self) -> bool:
         return self is Implementation.SQLFRAME  # pragma: no cover
 
     def _backend_version(self) -> tuple[int, ...]:
```
FBruzzesi, line 573: Should this be a property? 🤔

FBruzzesi:
@dangotbanned I am quite happy with the state of this now. Let me know your thoughts whenever you have the time 🤞🏼

tests/dependencies/imports_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+    assert isinstance(result, dd.DataFrame)
+
+
+def test_import_pandas(monkeypatch: pytest.MonkeyPatch) -> None:
+    pytest.importorskip("pandas")
+    if sys.version_info >= (3, 9):
+        monkeypatch.delitem(sys.modules, "pandas")
+    else:  # pragma: no cover
+        # NOTE: AttributeError: partially initialized module 'pandas' has no attribute 'compat' (most likely due to a circular import)
```
FBruzzesi: This condition might not be enough, I am trying to figure out which would suffice

narwhals/_utils.py
```diff
@@ -615,24 +574,10 @@ def is_sqlframe(self) -> bool:
...
-        elif self is Implementation.DASK:
-            into_version = get_dask()
-        else:
-            import sqlframe._version
+        from importlib.metadata import version
 
-            into_version = sqlframe._version
-        return parse_version(into_version)
+        distribution_name = _IMPLEMENTATION_TO_DISTRIBUTION_NAME.get(self, self.value)
+        return parse_version(version=version(distribution_name))
```
dangotbanned: We might end up triggering this downstream issue again with `importlib.metadata.version`:

- https://github.com/narwhals-dev/narwhals/pull/2391#issuecomment-2810496010
- https://github.com/pyinstaller/pyinstaller-hooks-contrib/issues/902#issuecomment-2810594839
- https://github.com/pyinstaller/pyinstaller-hooks-contrib/commit/18caba0fb9eb0f46353db63f614a2173ac41f522

It does have a solution, but I guess we'd need to keep the metadata requirements in sync somehow?

narwhals/_utils.py
```diff
@@ -615,24 +574,10 @@ def is_sqlframe(self) -> bool:
         return self is Implementation.SQLFRAME  # pragma: no cover
 
     def _backend_version(self) -> tuple[int, ...]:
```
dangotbanned, line 573: > Should this be a property? 🤔

I thought the same, @MarcoGorelli wanted a method when we added it (https://github.com/narwhals-dev/narwhals/pull/2324#discussion_r2051061025)

tests/dependencies/imports_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if impl.is_pyarrow() or (sys.version_info >= (3, 10) and impl.is_pandas()):
+        # NOTE: Can't do `monkeypatch.delitem` safely for other implementations
+        monkeypatch.delitem(sys.modules, module_name)
+
+    module = impl.to_native_namespace()
+    df = getattr(module, frame_constructor)(data, **kwargs)
+    result = _roundtrip_query(df)
+
+    assert isinstance(result, getattr(module, frame_type))
```
dangotbanned: In the *previous tests*, I forgot to add the step that I was eyballing in my IDE - but was actually the motivation for them 🤦‍♂️ 

<details><summary>Previous tests</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/214e15adbb114f6c605dcd3c813b739c53b6376b/tests/dependencies/imports_test.py#L30-L73

</p>
</details> 

If we used `polars` as an example, I was trying to make sure the **typing survived a roundtrip**.

I could see that out on hover, but the test should've been:

```py
def test_import_polars(data: dict[str, Any]) -> None: 
    pytest.importorskip("polars") 
    df = dependencies.import_polars().DataFrame(data) 
    result = _roundtrip_query(df) 
    import polars as pl

    if TYPE_CHECKING:
        from typing_extensions import assert_type
        
        assert_type(result, pl.DataFrame)
    assert isinstance(result, pl.DataFrame) 
```

That style is what I wrote in the ...


<details><summary><code>Namespace</code> tests</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/e831be1d95c2b37fd9b0951b7ac39e1893a9af0a/tests/namespace_test.py#L120-L132

</p>
</details> 


Since the PR doesn't have typing anymore, I'm not sure the test makes much sense as we're effectively just checking:
- Does `find_spec` work?
- Can we construct a dataframe via `import_module`?
- Once we have a dataframe, can we use `from_native` and `to_native`?

I think the other tests you've added give us the coverage for the new behavior

narwhals/_utils.py
```diff
@@ -649,6 +594,20 @@ def _backend_version(self) -> tuple[int, ...]:
...
 
+_IMPLEMENTATION_TO_MODULE_NAME: dict[Implementation, str] = {
+    Implementation.DASK: "dask.dataframe",
+    Implementation.MODIN: "modin.pandas",
+    Implementation.PYSPARK: "pyspark.sql",
+    Implementation.PYSPARK_CONNECT: "pyspark.sql.connect",
+}
+"""Stores non default mapping from Implementation to module name"""
+
+_IMPLEMENTATION_TO_DISTRIBUTION_NAME: dict[Implementation, str] = {
```
dangotbanned: Could we annotate these as `Mapping[Implementation, str]`?

I was thinking of doing the same for `MIN_VERSIONS`, but held off when it was in `utils` - should be fine now though

It'd mean we get a warning in CI if anyone submits a PR that mutates them at runtime.
Always good to try and prevent mutable global state 🙂

narwhals/_utils.py
FBruzzesi: Understand, never mind then!

FBruzzesi: Understand, never mind then!

narwhals/_utils.py
FBruzzesi: I can revert to use `__version__` attribute where applicable, `_version` for sqlframe, and `from importlib.metadata import version` only for ibis (I couldn't spot an attribute storing the library version)

As we discussed in the past, `__version__` is not a standard, so not guaranteed

dangotbanned: > As we discussed in the past, `__version__` is not a standard, so not guaranteed

Yeah that's a good point, personally I'd want use to be using `importlib.metadata.version` for that reason.
I wonder if the `pyinstaller` folks could collect metadata for all dependencies in our `pyproject.toml`?

It does seem a bit off to me that we'd need to do things in a non-standardized way for what I can only assume is a downstream perf optimization 🤔

FBruzzesi: The main "issue" is that it would require to import the module in the first place, which is somewhat a looping with `to_native_namespace` 🤔

An idea would be to move the following block into `_backend_version`:

```py
if self is Implementation.UNKNOWN:
    msg = "Cannot return native namespace from UNKNOWN Implementation"
    raise AssertionError(msg)

from importlib import import_module

module_name = _IMPLEMENTATION_TO_MODULE_NAME.get(self, self.value)
module = import_module(module_name)
```

and _cache_ the module once loaded to be reused in `to_native_namespace`

FBruzzesi: @dangotbanned is `from importlib import import_module` ok to use?

tests/dependencies/imports_test.py
FBruzzesi: So you would you be ok with just leave the tests `test_to_native_namespace` and `test_to_native_namespace_unknown`?

narwhals/_utils.py
dangotbanned: > and cache the module once loaded to be reused in `to_native_namespace`

I'm definitely in favor of adding some caching into the mix *somewhere* (https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-12211098), we're checking `backend_version` waaaaaaay too many times for something that doesn't change.
But I think we can look at that in another issue.

> and `from importlib.metadata import version` only for ibis (I couldn't spot an attribute storing the library version)

`ibis` seems to have always had a `__version__`?

- `ibis==1.0.0` https://github.com/ibis-project/ibis/blob/33201ac9cdbdafde0ae0dd99960657bc827f056a/ibis/__init__.py#L138
- `ibis==6.0.0` https://github.com/ibis-project/ibis/blob/97173935961231bac391943ba56ab499f4f9c2fe/ibis/__init__.py#L4
- `ibis==10.5.0` https://github.com/ibis-project/ibis/blob/cfa5768b6fb6020f712aadc842c9c35966e2acac/ibis/__init__.py#L5

dangotbanned: > @dangotbanned is `from importlib import import_module` ok to use?

Yeah should be - you'll get the same `ModuleNotFoundError` if it isn't installed

```py
>>> import ibis2
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[32], line 1
----> 1 import ibis2

ModuleNotFoundError: No module named 'ibis2'
```

```py
from importlib import import_module

>>> import_module("ibis2")
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[31], line 3
      1 from importlib import import_module
----> 3 import_module("ibis2")

File ~\AppData\Roaming\uv\python\cpython-3.13.2-windows-x86_64-none\Lib\importlib\__init__.py:88, in import_module(name, package)
     86             break
     87         level += 1
---> 88 return _bootstrap._gcd_import(name[level:], package, level)

File <frozen importlib._bootstrap>:1387, in _gcd_import(name, package, level)

File <frozen importlib._bootstrap>:1360, in _find_and_load(name, import_)

File <frozen importlib._bootstrap>:1324, in _find_and_load_unlocked(name, import_)

ModuleNotFoundError: No module named 'ibis2'
```

tests/dependencies/imports_test.py
dangotbanned: Yeah as long as they're all we need for coverage

FBruzzesi:
What's up with pyarrow nightly 🥲

dangotbanned:
> What's up with pyarrow nightly 🥲

Same!

I'm seeing that recently in #2572 as well

narwhals/_utils.py
```diff
@@ -574,13 +569,40 @@ def is_sqlframe(self) -> bool:
...
+        under the `_native_namespace` name.
+        """
+        if self is Implementation.UNKNOWN:  # pragma: no cover
+            msg = "Cannot return backend version from UNKNOWN Implementation"
+            raise AssertionError(msg)
+
+        from importlib import import_module
+
+        module_name = _IMPLEMENTATION_TO_MODULE_NAME.get(self, self.value)
+        self._native_namespace = import_module(module_name)
```
dangotbanned: I don't think we need to store modules on enums (or any state for that matter)

To be super clear, my comment in (https://github.com/narwhals-dev/narwhals/pull/2620#discussion_r2140114096) was meant to mean *let's defer caching to another issue*.

I was just about to merge and was pretty surprised to see this added in a commit titled [rollback _backend_verson](https://github.com/narwhals-dev/narwhals/pull/2620/commits/a847650435d977dcd7b19f0f90f107077ea51daa).

If we **need** to add caching of module loading, I would much prefer to just lift the `_LazyModule` stuff from `polars` - since that has typing support built in (https://github.com/pola-rs/polars/blob/7cb263a6dc4e3a1893b6ad4af963733c83268b60/py-polars/polars/dependencies.py)

FBruzzesi: This was to avoid repeating the exact same imports to check the backend version and return the module.

I am a bit in a rush but I just made a [commit](https://github.com/narwhals-dev/narwhals/pull/2620/commits/c4983669a9a0824eb5f7f72051c1c125145a0fa1) to factor out and cache in a helper function.

dangotbanned (approved):

dangotbanned:
oop I forgot to add a message 🤦‍♂️ 

@FBruzzesi thanks for the PR, bearing with me, and updating (#2492)! 😍

PR #2632, opened 2025-06-02
refactor: _duckdb runtime type checks to front end

### LINKED ISSUE #2583: chore: move `ensure_type` checks out from `_duckdb`
We have some checks like
```python
        ensure_type(window_size, int, type(None))
        ensure_type(min_samples, int)
```
which we should be doing at a higher level, not within `_duckdb`

camriddell:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2583 

## Checklist

- [x] Tests added
- [ ] Documented the changes

If runtime type checks are required, they should be handled at the superficial most level. Note that this was already being done for the various `Expr.rolling_*` and `Series.rolling_*` methods via `narwhals.utils._validate_rolling_arguments`, so the check at the `_duckdb` level was redundant.

This PR:
- removes `narwhals._duckdb.utils.ensure_type`
- adds runtime type checking to the `Expr.shift` and `Series.shift` (and tests for these)

### DIFF
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -993,6 +993,10 @@ class Expr:
             |└─────┴─────────┘ |
             └──────────────────┘
         """
+        if not isinstance(n, int):
+            msg = f"argument `n` must be of type {int!r}, got type {type(n)!r}"
+            raise TypeError(msg)
+
         return self._with_orderable_window(
             lambda plx: self._to_compliant_expr(plx).shift(n)
         )
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -15,7 +15,6 @@ from narwhals._duckdb.expr_str import DuckDBExprStringNamespace
 from narwhals._duckdb.expr_struct import DuckDBExprStructNamespace
 from narwhals._duckdb.utils import (
     col,
-    ensure_type,
     generate_order_by_sql,
     generate_partition_by_sql,
     lit,
@@ -546,8 +545,6 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "Expression"]):
 
     @requires.backend_version((1, 3))
     def shift(self, n: int) -> Self:
-        ensure_type(n, int)
-
         def func(inputs: DuckDBWindowInputs) -> Expression:
             order_by_sql = generate_order_by_sql(*inputs.order_by, ascending=True)
             partition_by_sql = generate_partition_by_sql(*inputs.partition_by)
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -125,8 +125,6 @@ class DuckDBExpr(LazyExpr["DuckDBLazyFrame", "Expression"]):
         min_samples: int,
         ddof: int | None = None,
     ) -> DuckDBWindowFunction:
-        ensure_type(window_size, int, type(None))
-        ensure_type(min_samples, int)
         supported_funcs = ["sum", "mean", "std", "var"]
         if center:
             half = (window_size - 1) // 2
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -284,12 +284,3 @@ def generate_order_by_sql(*order_by: str, ascending: bool) -> str:
     else:
         by_sql = ", ".join([f"{col(x)} desc nulls last" for x in order_by])
     return f"order by {by_sql}"
-
-
-def ensure_type(obj: Any, *valid_types: type[Any]) -> None:
-    # Use this for extra (possibly redundant) validation in places where we
-    # use SQLExpression, as an extra guard against unsafe inputs.
-    if not isinstance(obj, valid_types):  # pragma: no cover
-        tp_names = " | ".join(tp.__name__ for tp in valid_types)
-        msg = f"Expected {tp_names!r}, got: {type(obj).__name__!r}"
-        raise TypeError(msg)
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -1017,6 +1017,10 @@ class Series(Generic[IntoSeriesT]):
             2    4.0
             dtype: float64
         """
+        if not isinstance(n, int):
+            msg = f"argument `n` must be of type {int!r}, got type {type(n)!r}"
+            raise TypeError(msg)
+
         return self._with_compliant(self._compliant_series.shift(n))
 
     def sample(
--- a/tests/expr_and_series/shift_test.py
+++ b/tests/expr_and_series/shift_test.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+from contextlib import nullcontext
+from typing import Any
+
 import pyarrow as pa
 import pytest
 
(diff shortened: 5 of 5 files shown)


narwhals/expr.py
```diff
@@ -993,6 +993,10 @@ def shift(self, n: int) -> Self:
             |└─────┴─────────┘ |
             └──────────────────┘
         """
+        if not isinstance(n, int):
+            msg = f"argument `n` must be of type {int!r}, got type {type(n)!r}"
+            raise TypeError(msg)
```
MarcoGorelli: it is kinda nice to write `ensure_type(n, int)`, shall we move this `ensure_type` to utils.py ?

camriddell: Would it also be worth using the `ensure_type` function here (for the `TypeError`s specifically): https://github.com/narwhals-dev/narwhals/blob/b5a32ddc09b62a6c7d68c4e37b446aa0091308ca/narwhals/utils.py#L1473-L1506?

MarcoGorelli: Yeah probs

dangotbanned: @camriddell I'm not quite sure on passing a string parameter name around, but we *could* give a better error message with something like:

```py
def ensure_type(obj: Any, *valid_types: type[Any], param_name: str = "") -> None:
    if not isinstance(obj, valid_types):  # pragma: no cover
        tp_names = " | ".join(tp.__name__ for tp in valid_types)
        msg = f"Expected {tp_names!r}, got: {type(obj).__name__!r}"
        if param_name:
            left_pad = " " * 4
            val = repr(obj)
            assign = f"{left_pad}{param_name}="
            underline = (" " * len(assign)) + ("^" * len(val))
            msg = f"{msg}\n{assign}{val}\n{underline}"
        raise TypeError(msg)
```

```py
>>> ensure_type("world", int)
TypeError: Expected 'int', got: 'str'

>>> ensure_type("world", int, param_name="hello")
TypeError: Expected 'int', got: 'str'
    hello='world'
          ^^^^^^^
>>> ensure_type(5, int, param_name="window_size")
<no error>

>>> ensure_type(5.9, int, param_name="window_size")
TypeError: Expected 'int', got: 'float'
    window_size=5.9
                ^^^


```

narwhals/utils.py
```diff
@@ -1957,6 +1945,41 @@ def decorate(init_child: _Constructor[_T, P, R2], /) -> _Constructor[_T, P, R2]:
...
+        TypeError: Expected 'int', got: 'str'
+            test='hello'
+                 ^^^^^^^
+    """
+    if not isinstance(obj, valid_types):  # pragma: no cover
+        tp_names = " | ".join(tp.__name__ for tp in valid_types)
+        msg = f"Expected {tp_names!r}, got: {type(obj).__name__!r}"
+        if param_name:
+            left_pad = " " * 4
+            val = repr(obj)
```
dangotbanned, line 1989: My bad on this part, but probably a good idea to set some limit on the repr

```suggestion
            val = repr(obj)
            truncate_limit = 20
            if len(val) > truncate_limit:
                val = f"{type(obj).__name__}(...)"
```

We don't want this 😄 

```py
import polars as pl

data = {"a": [1, 2, 3], "b": [4, 5, 6]}
df = pl.DataFrame(data)

val = repr(df)
underline = "^" * len(val)
>>> print(f"{val}\n{underline}")
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 2   ┆ 5   │
│ 3   ┆ 6   │
└─────┴─────┘
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

camriddell:
@dangotbanned 

If we're considering non-builtin types for checking then we likely want to use factor in the `type(obj).__module__` here in addition to the `repr(type(obj).__name__)` to disambiguate types that share the same name but come from different modules.

Current behavior
```python
>>> df = pl.DataFrame([[1],[2],[3],[4],[5]], schema=[*'abcde'])
>>> ensure_type(df, pl.Series, param_name="df")
    ...
TypeError: Expected 'Series', got: 'DataFrame'
    df=DataFrame(...)
       ^^^^^^^^^^^^^^
```

Where we'd end up with some odd ends

```python
>>> import polars as pl
>>> import pandas as pd
>>> df = pl.DataFrame([[1],[2],[3],[4],[5]], schema=[*'abcde'])
>>> ensure_type(df, pd.DataFrame, param_name="df")
    ...
TypeError: Expected 'DataFrame', got: 'DataFrame'
    df=DataFrame(...)
       ^^^^^^^^^^^^^^
```

It ends up being a bit more verbose, but improves debugging
```python
>>> import polars as pl
>>> import pandas as pd
>>> df = pl.DataFrame([[1], [2], [3], [4], [5]], schema=[*"abcde"])
>>> ensure_type(df, pd.DataFrame, param_name="df")
    ...
TypeError: Expected 'pandas.core.frame.DataFrame', got: 'polars.dataframe.frame.DataFrame'
    df=polars.dataframe.frame.DataFrame(...)
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2632#issuecomment-2936096118

Good point @camriddell I agree!

That reminds me of a somewhat recent PR in `polars` that added a utility named something like `qualified_name`

If you're eager you might want to track it down to see how it solved this issue - but if not I'll take a look later 😅

camriddell:
> [#2632 (comment)](https://github.com/narwhals-dev/narwhals/pull/2632#issuecomment-2936096118)

https://github.com/narwhals-dev/narwhals/blob/e864628c33aa2d7b08ff5d659c9678295c0653f3/narwhals/utils.py#L1982-L1997

I already rolled my own. I'll check out the `polars` utility implementation though to see if there are corners I may have missed though.

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2632#issuecomment-2936394090

Haha nice one, okay I found it anyway

- https://github.com/pola-rs/polars/pull/22390

camriddell:
> [#2632 (comment)](https://github.com/narwhals-dev/narwhals/pull/2632#issuecomment-2936394090)
> 
> Haha nice one, okay I found it anyway
> 
> * [feat(python): When reporting unexpected types in errors, module-qualify the typename pola-rs/polars#22390](https://github.com/pola-rs/polars/pull/22390)

Yep the implementation is the same idea, grab the module and type name then stitch them together (removing the module if it is a built-in). The Polars version handles the cases where one passes an instance and optionally removes the prefix from types defined in Polars, but I don't think we need/should include either of those cases for our purposes.

```python
def qualified_type_name(obj: Any, *, qualify_polars: bool = False) -> str:
    """
    Return the module-qualified name of the given object as a string.

    Parameters
    ----------
    obj
        The object to get the qualified name for.
    qualify_polars
        If False (default), omit the module path for our own (Polars) objects.
    """
    if isinstance(obj, type):
        module = obj.__module__
        name = obj.__name__
    else:
        module = obj.__class__.__module__
        name = obj.__class__.__name__


    if (
        not module
        or module == "builtins"
        or (not qualify_polars and module.startswith("polars."))
    ):
        return name

    return f"{module}.{name}"
```

camriddell:
@MarcoGorelli I think this is ready for a final pass!

MarcoGorelli (approved):
cool, no objections from my end, thanks!

dangotbanned (approved):
Thanks for this @camriddell!

Provided I haven't broken anything in (https://github.com/narwhals-dev/narwhals/pull/2632/commits/711db0545435cb599314dd8e065b1f7c370efe10) - I think we're ready to ship it 🥳

Issue #1868, opened 2025-01-26
enh: allow for scalars in `sum_horizontal`


MarcoGorelli:
e.g.
```python
import duckdb
import narwhals as nw
rel = duckdb.sql('select * from values (1, 4), (1, 2), (2, 3), (2, 4) df(a, b)')
df = nw.from_native(rel)
 df.select(nw.sum_horizontal(1, 'a', 'b'))
```
throws
```python
InvalidIntoExprError: Expected an object which can be converted into an expression, got <class 'int'>

Hint:
- if you were trying to select a column which does not have a string
  column name, then you should explicitly use `nw.col`.
  For example, `df.select(nw.col(0))` if you have a column named `0`.
- if you were trying to create a new literal column, then you
  should explicitly use `nw.lit`.
  For example, `df.select(nw.lit(0))` if you want to create a new
  column with literal value `0`.
```

I think we should have a way of allowing for anything which can't be parsed as an expression to be parsed as `nw.lit` in these cases

FBruzzesi:
Are we aiming for the same behavior for boolean's (e.g. in `any_horizontal` and `all_horizontal`)?

Additionally, polars allow to pass numbers to `any|all_horizontal` and booleans to `sum|mean|max|min_horizontal`. Thoughts on this?

I have a way to address it, which is failing for dask only (I believe due to some broadcasting issues for how we create the `lit`)

MarcoGorelli:
thanks for looking into it!

not sure, will think about it, but this probably is quite low-prio for now

FBruzzesi:
> this probably is quite low-prio for now

I agree, using `nw.lit(value)` should work anyway, and that's what the error message suggest to do.

> may require a bit of a refactor if we want to do it properly (and not just special-case a few functions)

I was able to tackle it by adding a few extra checks/cases in `extract_compliant` function, however for different methods and backend we might require casting to the correct type

MarcoGorelli:
somehow we've regressed and now get a rubbish error message
```
    depth=max(x._depth for x in exprs) + 1,
              ^^^^^^^^
AttributeError: 'int' object has no attribute '_depth'
```

we should disallow non-expression input here at the narwhals level

FBruzzesi:
@MarcoGorelli I was going through the list of issues. This is now fixed. I would say that adding a test is all we need.

Update: I tried to check for other horizontal ops and... it's a bit messy. I will come back to this with more time and calm

PR #2647, opened 2025-06-05
feat: Add `coalesce` for all backends

### LINKED ISSUE #2639: [Enh]: add coalesce top level function

_No response_

https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.coalesce.html

This is a very common operation when joining tables or combining two columns into a single column.

_No response_

Instead of coalesce you can re-implement this with when-then-otherwise but it can be tedious to handle a variadic number of parameters.

_No response_

camriddell:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2639

## Checklist

- [x] Tests added
- [x] Documented the changes

Adds a `narwhals.coalesce` function for use with Narwhals Expressions.


camriddell:
I'll probably need to write more tests, at the very least this should serve as a functional MVP.

camriddell:
Hold off on merging for now, I need some tests for coalescing across disparate datatypes.

camriddell:
@MarcoGorelli thoughts on the following considerations for `coalesce`-ing across different datatypes:
1. Enforce a type promotion consistent with Polars
    - This will be tricky for pandas since it naturally promotes to object dtype, not to mention standard pandas `str` dtype is non-nullable and casting a numeric type to `str` type results in `'nan'`.
    - I can't think of issues with other backends, but this would likely involve a fair bit more code.
2. Restrict input types: you can only use `coalesce` if all of the datatypes fall within the same subtype (e.g. numeric, string, datetime, ...)
3. Don't worry about it we accept that cross type promotion will break and will mark those backends to xfail this type of test.

narwhals/functions.py
```diff
@@ -1774,3 +1775,61 @@ def concat_str(
...
+    |  │ -1  │         |
+    |  └─────┘         |
+    └──────────────────┘
+    """
+    flat_exprs = flatten(
+        [
+            expr if isinstance(expr, (Expr, str)) else lit(expr)
+            for expr in chain(flatten([exprs]), more_exprs)
+        ]
+    )
```
MarcoGorelli: this looks a bit suspicious, can we reuse anything? check `concat_str` or any of the other functions in this file

MarcoGorelli (commented):
thanks @camriddell !

i think what you've done is fine and we shouldn't do any casting on behalf of the user, let  the backends handle that as they wish

narwhals/functions.py
camriddell: The main idea was to coerce any inputs that are not strings/expressions into literals. Though this would also mean that any valid `nw.Series` objects would be inappropriately coerced as well.

Is there a helper somewhere that does this? If not, I've made the change to remove this capability- reverting is easy.

tests/expr_and_series/coalesce_test.py
```diff
@@ -0,0 +1,65 @@
...
+            "pandas_pyarrow",
+            "pandas_nullable",
+            "modin_pyarrow",
+            "pyarrow_table",
+            "ibis",
+        ]
+    ):
+        # backends handle upcasting independently, most of them fail
+        #   due to strict preferences against automatic upcasting
+        request.applymarker(pytest.mark.xfail)
```
MarcoGorelli: tbh i'd be partial to removing this test, if it's backend dependent anyway

narwhals/_duckdb/namespace.py
```diff
@@ -207,6 +207,19 @@ def func(_df: DuckDBLazyFrame) -> list[Expression]:
...
+            cols = (c for _expr in exprs for c in _expr(df))
+            return [CoalesceOperator(*cols)]
+
+        return self._expr(
+            call=func,
+            evaluate_output_names=combine_evaluate_output_names(*exprs),
+            alias_output_names=combine_alias_output_names(*exprs),
+            backend_version=self._backend_version,
+            version=self._version,
+        )
```
MarcoGorelli: this should be elementwise, right? can we use the new `_with_elementwise` here for duckdb and pyspark?

camriddell:
@MarcoGorelli think this is in its final iteration unless there's something you catch!

narwhals/_arrow/namespace.py
```diff
@@ -266,6 +267,35 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
...
+                    version=self._version,
+                )
+            ]
+
+        exprs = tuple(
+            expr
+            if self._expr._is_expr(expr)
+            else self.lit(expr, dtype=None).broadcast(ExprKind.LITERAL)
+            for expr in exprs
+        )
```
MarcoGorelli: can we assume that they're all expressions, and raise the same (rubbish) error message we get in `sum_horizontal` for now?

we can then address them together with a more general solution 

https://github.com/narwhals-dev/narwhals/issues/1868

narwhals/_dask/namespace.py
```diff
@@ -278,6 +278,29 @@ def func(df: DaskLazyFrame) -> list[dx.Series]:
...
+        def func(df: DaskLazyFrame) -> list[dx.Series]:
+            series = align_series_full_broadcast(
+                df, *(s for _expr in exprs for s in _expr(df))
+            )
+            return [reduce(lambda x, y: x.fillna(y), series)]
+
+        exprs = tuple(
+            expr
+            if self._expr._is_expr(expr)
+            else self.lit(expr, dtype=None).broadcast(ExprKind.LITERAL)
```
MarcoGorelli: same comment

narwhals/_pandas_like/namespace.py
```diff
@@ -82,6 +104,7 @@ def _lit_pandas_series(df: PandasLikeDataFrame) -> PandasLikeSeries:
                 index=df._native_frame.index[0:1],
                 context=self,
             )
+            pandas_series._broadcast = True
```
MarcoGorelli: this...kinda scares me 😄  do we need it?

camriddell:
@MarcoGorelli ready for another pass!

narwhals/functions.py
```diff
@@ -1847,3 +1847,60 @@ def concat_str(
...
+
+
+def coalesce(
+    exprs: IntoExpr | Iterable[IntoExpr], *more_exprs: IntoExpr | NonNestedLiteral
+) -> Expr:
+    """Folds the columns from left to right, keeping the first non-null value.
+
+    Arguments:
+        exprs: Columns to coalesce. Strings are parsed as column names, other non-expression inputs are
+            parsed as literals.
```
MarcoGorelli: is this true? i think other non-expression inputs are disallowed and people should use `lit`?

narwhals/stable/v1/__init__.py
```diff
@@ -1448,6 +1448,20 @@ def concat_str(
     )
 
 
+def coalesce(exprs: IntoExpr | Iterable[IntoExpr], *more_exprs: IntoExpr) -> Expr:
+    """Folds the columns from left to right, keeping the first non-null value.
+
+    Arguments:
+        exprs: Columns to coalesce. Strings are parsed as column names, other non-expression inputs are
+            parsed as literals.
```
MarcoGorelli: same

narwhals/functions.py
```diff
@@ -1847,3 +1847,60 @@ def concat_str(
...
+    |  │ 4   │         |
+    |  │ -1  │         |
+    |  └─────┘         |
+    └──────────────────┘
+    """
+    flat_exprs = flatten([*flatten([exprs]), *more_exprs])
+
+    non_exprs = [expr for expr in flat_exprs if not isinstance(expr, (str, Expr, Series))]
+    if non_exprs:
+        msg = f"All arguments must be of type {str!r}, {Expr!r}, or {Series!r}.\nGot: {', '.join(repr((type(e), e)) for e in non_exprs)}"
```
MarcoGorelli: can we write "All arguments to `coalesce`"?

MarcoGorelli (approved):
thanks @camriddell ! just left two really minor comments then feel free to merge

narwhals/functions.py
camriddell: Thanks, forgot to update the docstring after I removed the `lit` coercing.

PR #2649, opened 2025-06-06
chore: Refactor window functions for lazy backends

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


dangotbanned:
@MarcoGorelli I forgot to tag you on this 🤦‍♂️ 🤦‍♂️ 

- https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2123605800

Would you be able to add that bit back in for `SparkLikeExpr`, or explain the removal please? 🙏

MarcoGorelli:
hey - sure, it just felt simpler without it, in general i think too many abstractions and too small functions make it harder to rewrite things (and inevitably things need rewriting)

dangotbanned:
> make it harder to rewrite things (and inevitably things need rewriting)

Interesting, so it seems like we both had the same motivation but from different angles 😄 

I thought it would be easier to refactor, since we reduce the exposure of `*WindowInputs` internals to fewer places

## Example
I feel like this is clearer that the change this PR made was the removal of `WindowInputs.expr`

### Where this PR has `is_first_distinct`
```py
    def is_first_distinct(self) -> Self:
        def func(df: SparkLikeLazyFrame, inputs: SparkWindowInputs) -> Sequence[Column]:
            return [
                self._F.row_number().over(
                    self.partition_by(*inputs.partition_by, expr).orderBy(
                        *self._sort(*inputs.order_by)
                    )
                )
                == 1
                for expr in self(df)
            ]

        return self._with_window_function(func)
```

### What is could have looked like

```py
def is_first_distinct(self) -> Self:
        def func(df: SparkLikeLazyFrame, inputs: SparkWindowInputs) -> Sequence[Column]:
            row_number = self._F.row_number()
            order = list(self._sort(inputs))
            return [
                row_number.over(self.partition_by(inputs, expr).orderBy(*order)) == 1
                for expr in self(df)
            ]

        return self._with_window_function(func)
```

narwhals/_ibis/expr.py
```diff
@@ -57,9 +56,23 @@ def __init__(
...
+        ) -> list[ir.Value]:
+            assert not window_inputs.order_by  # noqa: S101
+            return [
+                expr.over(ibis.window(group_by=window_inputs.partition_by))
+                for expr in call(df)
+            ]
 
-        # This can only be set by `_with_window_function`.
-        self._window_function: IbisWindowFunction | None = None
+        return window_func
```
dangotbanned: If `*WindowFunction` were a class, you could have something like:

```py
        self._window_function: IbisWindowFunction = (
            window_function or IbisWindowFunction.from_callable(call)
        )
```

Or this and replace the `call(df)` with `expr(df)`:
```py
        self._window_function: IbisWindowFunction = (
            window_function or IbisWindowFunction.from_expr(self)
        )
```

dangotbanned: Whatever the `@classmethod` is, it can be enforced in a `WindowFunction` protocol, - rather than `LazyExpr._default_window_function` (which should be added if sticking with this pattern IMO)

MarcoGorelli: dask doesn't have this function, so not sure it can be added to lazyexpr?

dangotbanned: > dask doesn't have this function, so not sure it can be added to lazyexpr?

What if we had something like this?

```py
class LazyExpr(
    CompliantExpr[CompliantLazyFrameT, NativeExprT],
    Protocol38[CompliantLazyFrameT, NativeExprT],
):
    @property
    def window_function(self) -> WindowFunction: ...


class IbisExpr(LazyExpr):
    _window_function: WindowFunction | None

    @property
    def window_function(self) -> WindowFunction:
        return self._window_function or WindowFunction.from_expr(self)


class DaskExpr(LazyExpr):
    window_function = not_implemented()  # pyright: ignore[reportAssignmentType]
```

Another benefit is that we don't initialize the default when it isn't needed - very lazy indeed! 😎

MarcoGorelli:
that's nice - thanks @dangotbanned for your review!

gonna ship this so we can work towards standardising empty sums, but definitely happy to take further improvements based on what you're suggesting 🙏

Issue #2486, opened 2025-05-04
[Enh]: Add `DType.to_<backend>()`


dangotbanned:
# Tracking
- [x] #2626
- [ ] #2264
- [x] #2957

# Description
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/2460#discussion_r2072360670_

For all of the `narwhals_to_native_dtype` functions, I'd recommend the same solution.

> [!IMPORTANT]
> *It'd be nice* to also do this for `native_to_narwhals`, but the caching may be a challenge

It is a bit much (to implement) for (#2460), but hopefully it gives an example of where these kinds of changes can lead 🙂 

## General
We could *instead* have methods on each `DType` class like:

```py
to_arrow
to_dask
to_duckdb
to_polars
# NOTE: Could be more granular for these
to_pandas_like
to_spark_like
```

- Each would be much shorter (the goal of the rules) and very easy to digest.
- They'd be **more performant** as we'd skip all of the `isinstance` checks 
  - By just calling the same method for every single case
- We could also make these public very easily
  - I imagine they'd be helpful in the same way as (#1912) 😉 
- ~~Unsupported conversions are handled by the abscence of methods~~

## Implementation
We would want to borrow/adapt the [`classinstmethod`](https://github.com/pola-rs/polars/blob/479b9d24da685fecad3d5682b7bc82ffe0f52a45/py-polars/polars/datatypes/classes.py#L32-L37) descriptor from `polars`.
They use that for [`to_python`](https://github.com/pola-rs/polars/blob/479b9d24da685fecad3d5682b7bc82ffe0f52a45/py-polars/polars/datatypes/classes.py#L230-L246) which is the same kind of operation - and is something we might want in the future anyway (#2264) 😏 

## `V1`
Another big benefit of this is that we'd move the `MAIN` vs `V1` logic into the `V1` data types.
I think this would make it really clear for us what the behavior of `Enum` should be:

https://github.com/narwhals-dev/narwhals/blob/162dd5fbf9878155ed64d16043040f6b075f6101/narwhals/_polars/utils.py#L184-L191

- [x] #2550

dangotbanned:
If anyone takes a stab at this feel free to ping me 🙂 

[...]

PR #2654, opened 2025-06-07
feat(typing): Add `IntoDType` alias

### LINKED ISSUE #2626: Add `IntoDType` alias, replace `DType | type[DType]`
Most places that we currently use `DType | type[DType]` will eventually raise an error if the `type[...]` is nested.

But this isn't reflected in the typing or in the docstrings, which I think isn't helpful

## Example

```py
import polars as pl
import narwhals as nw

data = {"a": ["A", "B", "A"], "b": [1, 2, 3], "c": [9, 2, 4], "d": [8, 7, 8]}
df = pl.DataFrame(data)
frame = nw.from_native(df)

good = nw.col("a").cast(nw.Struct([nw.Field("a", nw.String)]))
bad = nw.col("a").cast(nw.Struct)
```
```py
>>> frame.select(good)
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|  shape: (3, 1)   |
|  ┌───────────┐   |
|  │ a         │   |
|  │ ---       │   |
|  │ struct[1] │   |
|  ╞═══════════╡   |
|  │ {"A"}     │   |
|  │ {"B"}     │   |
|  │ {"A"}     │   |
|  └───────────┘   |
└──────────────────┘
```

The error we get *kinda* hints towards the issue in the individual case.
But if we hit this path when dealing with *multiple* `DType`s - I imagine the error wouldn't help much at all

```py
>>> frame.select(bad)
AttributeError: type object 'Struct' has no attribute 'fields'
```

<details><summary>Traceback</summary>
<p>

```py
/narwhals/narwhals/_polars/expr.py:65, in PolarsExpr.cast(self, dtype)
     64 def cast(self, dtype: DType | type[DType]) -> Self:
---> 65     dtype_pl = narwhals_to_native_dtype(dtype, self._version, self._backend_version)
     66     return self._with_native(self.native.cast(dtype_pl))

/narwhals/narwhals/_polars/utils.py:214, in narwhals_to_native_dtype(dtype, version, backend_version)
    207     return pl.List(narwhals_to_native_dtype(dtype.inner, version, backend_version))
    208 if isinstance_or_issubclass(dtype, dtypes.Struct):
    209     fields = [
    210         pl.Field(
    211             field.name,
    212             narwhals_to_native_dtype(field.dtype, version, backend_version),
    213         )
--> 214         for field in dtype.fields
    215     ]
    216     return pl.Struct(fields)
    217 if isinstance_or_issubclass(dtype, dtypes.Array):  # pragma: no cover

AttributeError: type object 'Struct' has no attribute 'fields'
```

</p>
</details> 

## Proposal
We should narrow the `type[DType]` side of the union to something like:

```py

from narwhals import dtypes

if TYPE_CHECKING:

# NOTE: Needs to use forward references in the real thing    
NonNestedDType: TypeAlias = (
    dtypes.NumericType
    | dtypes.TemporalType
    | dtypes.String
    | dtypes.Boolean
    | dtypes.Binary
    | dtypes.Unknown
    | dtypes.Object
)

IntoDType: TypeAlias = dtypes.DType | type[NonNestedDType]
```

That would align nicely with the python type aliases we have already:

https://github.com/narwhals-dev/narwhals/blob/a0d82cd8f30b42930fd8b8cec2e1c38bb05ddf7b/narwhals/typing.py#L355-L360

And we now get some helpful red squiggles on the failing case 🥳:

Image

dangotbanned:
Closes #2626

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2626
- Will be expanded by (https://github.com/narwhals-dev/narwhals/issues/2264#issuecomment-2930792917)

## Checklist

- [x] Tests added
- [x] Documented the changes

I *think* `selectors` are safe as-is, but will follow that up separately if needed


dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2654#commits-pushed-b2bcb45

@MarcoGorelli sorry I butchered a rebase 😭

narwhals/typing.py
```diff
@@ -359,6 +359,38 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+    |│ a   ┆ b        │|
+    |│ --- ┆ ---      │|
+    |│ i32 ┆ list[i8] │|
+    |╞═════╪══════════╡|
+    |│ 1   ┆ [4, 0]   │|
+    |│ 2   ┆ [5, 0]   │|
+    |│ 3   ┆ [6, 0]   │|
+    |└─────┴──────────┘|
+    └──────────────────┘
+"""
```
dangotbanned, line 391: The logic I'm following for this is that we use `IntoDType` for anything that can trigger a `cast`.

Or another way to look at it is if the call reaches into `_<backend>.utils.<thing>_to_<other>_dtype` then this should warn when you're likely to raise

tests/series_only/cast_test.py
```diff
@@ -125,7 +125,7 @@ def test_cast_to_enum_vmain(
     with pytest.raises(
         ValueError, match="Can not cast / initialize Enum without categories present"
     ):
-        df_nw.select(col_a.cast(nw.Enum))
+        df_nw.select(col_a.cast(nw.Enum))  # type: ignore[arg-type]
```
dangotbanned, line 128: Not sure if this got lost, but needing these ignores is a positive outcome 🙂

FBruzzesi: Nice one

tests/expr_and_series/cast_test.py
```diff
@@ -360,3 +360,20 @@ def test_cast_binary(request: pytest.FixtureRequest, constructor: Constructor) -
...
+        a.cast(nw.Field)  # type: ignore[arg-type]
+    a.cast(nw.List)  # type: ignore[arg-type]
+    a.cast(nw.Array)  # type: ignore[arg-type]
+    a.cast(nw.Enum)  # type: ignore[arg-type]
+    a.cast(nw.Struct([nw.Field]))  # type: ignore[list-item]
+    a.cast(nw.Struct({"a": nw.Int16, "b": nw.Enum}))  # type: ignore[dict-item]
+    with pytest.raises(TypeError):
+        a.cast(nw.Field("a", nw.Array))  # type: ignore[arg-type]
+    a.cast(nw.List(nw.Struct))  # type: ignore[arg-type]
+    a.cast(nw.Array(nw.List, 2))  # type: ignore[arg-type]
```
FBruzzesi: Shouldn't (some of) these fail? Nested data type without inner type?

dangotbanned: > Shouldn't (some of) these fail? Nested data type without inner type?

@FBruzzesi they should do when they're evaluated with a backend!
But as the errors aren't consistent, I opted for typing-only tests *for now*.

I sort of mentioned that in (https://github.com/narwhals-dev/narwhals/issues/2264#issuecomment-2930792917) 
> I'd started some very early work on that a while back, but needing to handle the nested case everywhere was getting repetitive.
> I imagine we'd end up with some utilities/error handling for that throughout the work on the other issues 🙏

and also in (https://github.com/narwhals-dev/narwhals/issues/2486#issuecomment-2868828840)

> But these are **invalid**
> ```py
> sad_dtypes = (
>     nw.Enum,
>     nw.List,
>     nw.Struct,
>     nw.Array,
>     nw.Decimal,
>     nw.Decimal(),
>     nw.Field("hi", nw.String()),
>     nw.Field,
> )
> 
> for dtype in sad_dtypes:
>     try:
>         converted = to_polars(dtype)
>         print(f"Converted: {dtype!r} -> {converted!r}")
>     except (ValueError, NotImplementedError, AttributeError) as err:
>         print(f"{err!r}")
> 
> ValueError('Can not cast / initialize Enum without categories present')
> AttributeError("type object 'List' has no attribute 'inner'")
> AttributeError("type object 'Struct' has no attribute 'fields'")
> AttributeError("type object 'Array' has no attribute 'size'")
> NotImplementedError('Casting to Decimal is not supported yet.')
> NotImplementedError('Casting to Decimal is not supported yet.')
> AttributeError("type object 'Float64' has no attribute 'name'")
> Converted: <class 'narwhals.dtypes.Field'> -> Unknown
> ```
> 
> I think I might have revealed a bug or two for those 😄

FBruzzesi: > they should do when they're evaluated with a backend!

Makes a lot of sense! Thanks @dangotbanned 👌 I will approve for now, it would be nice to have an umbrella test that catches even a general exception for these

FBruzzesi (approved):
Thanks @dangotbanned 🚢🚢

It has been quite some time since last time I approved a PR 😂

NarwhalChubbiwhalGIF

tests/expr_and_series/cast_test.py
dangotbanned: Thanks @FBruzzesi, resolved in (https://github.com/narwhals-dev/narwhals/pull/2654/commits/eaca4b35aeb514a6af41fd1d79849d3f3674fcd9)

MarcoGorelli:
nice one, thanks both!

Issue #2652, opened 2025-06-07
"push down" `over` in `when/then/otherwise`


MarcoGorelli:
similar to #2623 , we should be able to support
```
import polars as pl
import narwhals as nw
import duckdb
import daft
import ibis

data = {
    "a": [1,2,4],
    "b": [1,1,2],
    'i': [0,1,2]
}
print(
    nw.from_native(pl.DataFrame(data))
    .lazy('duckdb')
    .select('i', nw.when(nw.col('a')>1).then(nw.col('a').sum()).otherwise(2).over('b'))
    .sort('i')
)
```
by rewriting it as
```
    .select('i', nw.when(nw.col('a')>1).then(nw.col('a').sum().over('b')).otherwise(2))
```

MarcoGorelli:
https://github.com/narwhals-dev/narwhals/pull/2663#issuecomment-2967129371

Issue #2645, opened 2025-06-05
[Bug]: When-Then doesn't broadcast predicates or result (but Polars does)


camriddell:
### Describe the bug

Narwhals when/then is more strict than Polars is. Based on the behavior, it seems that Polars does not explicitly check the result size of the predicates and instead broadcasts the evaluated predicates against the returned value.

### Steps or code to reproduce the bug

```python
import polars as pl
import narwhals as nw

df = pl.DataFrame({'a': [1,2,3]})
print(
    nw.from_native(df)
    .select(
        nw.when(nw.col('a').sum() == 6).then(100)
    )
)
# Traceback (most recent call last):
#   File "/home/cameron/.vim-excerpt", line 9, in <module>
#     nw.when(nw.col('a').sum() == 6).then(100)
#     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "/home/cameron/.pyenv/versions/dutc-site-new/lib/python3.12/site-packages/narwhals/functions.py", line 1569, in when
#     return When(*predicates)
#            ^^^^^^^^^^^^^^^^^
#   File "/home/cameron/.pyenv/versions/dutc-site-new/lib/python3.12/site-packages/narwhals/functions.py", line 1486, in __init__
#     check_expressions_preserve_length(self._predicate, function_name="when")
#   File "/home/cameron/.pyenv/versions/dutc-site-new/lib/python3.12/site-packages/narwhals/_expression_parsing.py", line 576, in check_expressions_preserve_length
#     raise ShapeError(msg)
# narwhals.exceptions.ShapeError: Expressions which aggregate or change length cannot be passed to 'when'.
```

### Expected results

```python
import polars as pl

df = pl.DataFrame({'a': [1,2,3]})

print(
    df.select(
        pl.when(pl.col('a').sum() == 6).then(100)
    )
)
# shape: (1, 1)
# ┌─────────┐
# │ literal │
# │ ---     │
# │ i32     │
# ╞═════════╡
# │ 100     │
# └─────────┘

# it seems that Polars evaluates the when/then first
#   the broadcasts the result against the rest of the frame.
print(
    df.with_columns(
        pl.when(pl.col('a').sum() == 6).then(100),
    )
)
# shape: (3, 2)
# ┌─────┬─────────┐
# │ a   ┆ literal │
# │ --- ┆ ---     │
# │ i64 ┆ i32     │
# ╞═════╪═════════╡
[...]

PR #2662, opened 2025-06-10
chore: refactor `EagerWhen._if_then_else`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_compliant/namespace.py
```diff
@@ -140,7 +140,7 @@ def _dataframe(self) -> type[EagerDataFrameT]: ...
     def _series(self) -> type[EagerSeriesT]: ...
     def when(
         self, predicate: EagerExprT
-    ) -> EagerWhen[EagerDataFrameT, EagerSeriesT, EagerExprT, NativeSeriesT]: ...
+    ) -> EagerWhen[EagerDataFrameT, EagerSeriesT, EagerExprT]: ...
```
MarcoGorelli, line 142: I'm struggling to follow this

If I remove `NativeSeriesT` from `EagerWhen`, and hence from the return type, then `mypy` / `pyright` tell me:
```
   /home/runner/work/narwhals/narwhals/narwhals/_compliant/namespace.py:133:7 - warning: Type variable "NativeSeriesT" used in generic Protocol "EagerNamespace" should be covariant (reportInvalidTypeVarUse)
```

But, I can't make it covariant, because it's used as an argument to `from_native`. If I did, type checkers would complain. I can't tell what they're expecting me to do 😄 

@dangotbanned sorry for the ping, any tips would be appreciated 🙏

MarcoGorelli:
I think we may just need to remove `NativeSeriesT` from `EagerNamespace`? It's used as an argument, but due to how `EagerNamespace` is used (probably?), pyright/mypy expect `EagerNamespace` to be covariant in `NativeSeriesT`. But, it can't be covariant in `NativeSeriesT` because `NativeSeriesT` appears as a function argument, so it should be contravariant.

One trick could be to add `def _no_op(self, s: NativeSeriesT) -> NativeSeriesT: return s` to trick type checkers into thinking that `EagerNamespace` is invariant in `NativeSeriesT`. But, I think that would look extremely confusing. IMHO this is a chance to just remove `NativeSeriesT` from `EagerNamespace` protocol and simplify things a little

narwhals/_compliant/namespace.py
dangotbanned: Hey @MarcoGorelli, I'll try to take a look at this today 🙏 

Would you be able to give some more background on what you're trying to accomplish with this PR please?

I'm not following the motivation from either the title or commit messages 🤔

narwhals/_compliant/namespace.py
```diff
@@ -132,31 +131,17 @@ def from_native(self, data: NativeFrameT_co | Any, /) -> CompliantLazyFrameT:
...
-    # - Using the guards there introduces `_NativeModin`, `_NativeCuDF`
-    # - These types haven't been integrated into the backend
-    # - Most of the `pandas` stuff is still untyped
-    @overload
-    def from_native(
-        self, data: NativeFrameT | NativeSeriesT | Any, /
-    ) -> EagerDataFrameT | EagerSeriesT: ...
-    def from_native(
-        self, data: NativeFrameT | NativeSeriesT | Any, /
-    ) -> EagerDataFrameT | EagerSeriesT:
```
dangotbanned, line 159: I don't think this should have been merged

narwhals/_compliant/dataframe.py
```diff
@@ -450,14 +449,10 @@ def _numpy_column_names(
...
     def _gather_slice(self, rows: _SliceIndex | range) -> Self: ...
-    def _select_multi_index(
-        self, columns: SizedMultiIndexSelector[NativeSeriesT]
-    ) -> Self: ...
-    def _select_multi_name(
-        self, columns: SizedMultiNameSelector[NativeSeriesT]
-    ) -> Self: ...
+    def _select_multi_index(self, columns: SizedMultiIndexSelector[Any]) -> Self: ...
+    def _select_multi_name(self, columns: SizedMultiNameSelector[Any]) -> Self: ...
     def _select_slice_index(self, columns: _SliceIndex | range) -> Self: ...
```
dangotbanned, line 456: Yeah can you revert this PR?

dangotbanned: A refactor on `when` shouldn't have to remove typing from other parts of `narwhals`

narwhals/_compliant/namespace.py
MarcoGorelli: sure, it's to enable #2652 and #2645

i spent about 2 hours trying different things and wasn't able to get anywhere with this 😳 , based on https://github.com/narwhals-dev/narwhals/pull/2662#issuecomment-2962008537 it looks to me like there were already issues with these protocols. happy for some parts to be brought back, so long as we're able to move towards the linked issues

PR #2663, opened 2025-06-10
feat: when-then broadcasting

camriddell:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2645 

## Checklist

- [x] Tests added
- [ ] Documented the changes

This adds broadcasting behavior to the `when/then` pattern similar to Polars (see #2645) for our eager backends and dask.
The implementation for the Lazy backends will require #2652 unless we want to introduce some hacky workarounds (e.g. we may be able to do this by joining after an aggregation to "broadcast" results; however I haven't fully thought through how this would look in the codebase).

### DIFF
--- a/narwhals/_compliant/when_then.py
+++ b/narwhals/_compliant/when_then.py
@@ -154,11 +154,18 @@ class EagerWhen(
         is_expr = self._condition._is_expr
         when: EagerSeriesT = self._condition(df)[0]
         then: EagerSeriesT
+        if (
+            self._condition._metadata is not None
+            and self._condition._metadata.is_scalar_like
+        ):
+            when._broadcast = True
+
         if is_expr(self._then_value):
             then = self._then_value(df)[0]
         else:
             then = when.alias("literal")._from_scalar(self._then_value)
             then._broadcast = True
+
         if is_expr(self._otherwise_value):
             otherwise = self._otherwise_value(df)[0]
         elif self._otherwise_value is not None:
--- a/narwhals/_compliant/when_then.py
+++ b/narwhals/_compliant/when_then.py
@@ -21,6 +21,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self, TypeAlias
 
     from narwhals._compliant.typing import EvalSeries, ScalarKwargs
+    from narwhals._compliant.window import WindowInputs
     from narwhals._utils import Implementation, Version, _FullContext
     from narwhals.typing import NonNestedLiteral
 
@@ -50,6 +51,9 @@ class CompliantWhen(Protocol38[FrameT, SeriesT, ExprT]):
     @property
     def _then(self) -> type[CompliantThen[FrameT, SeriesT, ExprT]]: ...
     def __call__(self, compliant_frame: FrameT, /) -> Sequence[SeriesT]: ...
+    def _window_function(
+        self, compliant_frame: FrameT, window_inputs: WindowInputs[Any]
+    ) -> Sequence[SeriesT]: ...
 
     def then(
         self, value: IntoExpr[SeriesT, ExprT], /
@@ -124,9 +128,7 @@ class LazyThen(
         obj = cls.__new__(cls)
         obj._call = when
 
-        # This may require more complicated logic if we want to push down the
-        # `over`: https://github.com/narwhals-dev/narwhals/issues/2652.
-        obj._window_function = None
+        obj._window_function = when._window_function
 
         obj._when_value = when
         obj._depth = 0
@@ -154,11 +156,6 @@ class EagerWhen(
         is_expr = self._condition._is_expr
         when: EagerSeriesT = self._condition(df)[0]
         then: EagerSeriesT
-        if (
-            self._condition._metadata is not None
-            and self._condition._metadata.is_scalar_like
-        ):
-            when._broadcast = True
 
         if is_expr(self._then_value):
             then = self._then_value(df)[0]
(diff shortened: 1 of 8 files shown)


camriddell:
Rebased on top of #2662 do not merge until that PR lands in main.
We should be able to merge once that one receives the green light (provided all tests pass here).

MarcoGorelli:
thanks @camriddell ! could you rebase again please?

narwhals/_compliant/when_then.py
```diff
@@ -154,11 +154,18 @@ def __call__(self, df: EagerDataFrameT, /) -> Sequence[EagerSeriesT]:
         is_expr = self._condition._is_expr
         when: EagerSeriesT = self._condition(df)[0]
         then: EagerSeriesT
+        if (
+            self._condition._metadata is not None
+            and self._condition._metadata.is_scalar_like
+        ):
+            when._broadcast = True
```
MarcoGorelli: i don't think we should ever broadcast `when`:

- either `when` preserves length, in which case we don't need to do anything
- or `when` is scalar-like, in which case `then` and `otherwise` must also be scalar-like

It's true that Polars supports `pl.when(when_scalar).then(then_column)`, but I don't think we should, just like how we don't support `df.filter(predicate_scalar)`

narwhals/_dask/namespace.py
```diff
@@ -283,25 +285,44 @@ def _then(self) -> type[DaskThen]:
...
-            then_value = self._then_value(df)[0]
-        else:
-            then_value = self._then_value
-        (then_series,) = align_series_full_broadcast(df, then_value)
-        validate_comparand(condition, then_series)
+        def _aggregates(expr: Any) -> bool:
+            if isinstance(expr, DaskExpr):
+                return expr._metadata is not None and expr._metadata.is_scalar_like
+            elif isinstance(expr, dd.Series):
+                return len(expr) == 1
```
MarcoGorelli: this kinda scares me 😄 if `dd.Series` is lazy, does `len(expr)` force compute somewhere?

tests/expr_and_series/when_test.py
```diff
@@ -115,12 +115,6 @@ def test_when_then_otherwise_into_expr(constructor: Constructor) -> None:
     assert_equal_data(result, expected)
 
 
-def test_when_then_invalid(constructor: Constructor) -> None:
-    df = nw.from_native(constructor(data))
-    with pytest.raises(ShapeError):
-        df.select(nw.when(nw.col("a").sum() > 1).then("c"))
```
MarcoGorelli: this should still be invalid, because the argument to `when` is scalar-like but the argument to `then` isn't

tests/expr_and_series/when_test.py
```diff
@@ -144,3 +138,63 @@ def test_when_then_otherwise_multi_output(constructor: Constructor) -> None:
...
+    ("condition", "then", "otherwise", "expected"),
+    [
+        (nw.col("a").sum() == 6, 100, 200, [100]),
+        (nw.col("a").sum() == 6, nw.col("a"), 200, [1, 2, 3]),
+        (nw.col("a").sum() == 6, nw.col("a"), nw.col("b"), [1, 2, 3]),
+        (nw.col("a").sum() == 6, 100, nw.col("b"), [100, 100, 100]),
+        (nw.col("a").sum() == 5, 100, 200, [200]),
+        (nw.col("a").sum() == 5, nw.col("a"), 200, [200, 200, 200]),
+        (nw.col("a").sum() == 5, nw.col("a"), nw.col("b"), [4, 5, 6]),
+        (nw.col("a").sum() == 5, 100, nw.col("b"), [4, 5, 6]),
```
MarcoGorelli: similarly here, we should remove test cases: 2, 3, 4, 6, 7, 8 (i started counting at 1 😳 )

narwhals/_dask/namespace.py
camriddell: Ah, I believe you are correct here. Based on your above comment we don't need to know if all expressions aggregate- just whether the predicate aggregates and can enforce behavior from there.

narwhals/_dask/namespace.py
```diff
@@ -283,23 +285,33 @@ def _then(self) -> type[DaskThen]:
...
+        )
+        otherwise_value = (
+            self._otherwise_value(df)[0]
+            if isinstance(self._otherwise_value, DaskExpr)
+            else self._otherwise_value
+        )
 
         if self._otherwise_value is None:
-            return [then_series.where(condition)]
+            otherwise_value = get_dask_expr()._expr.Where._defaults["other"]
```
MarcoGorelli: this is the only part that jumps out to me as suspicious 🤔

MarcoGorelli (approved):
thanks @camriddell ! really solid work here! really impressed you figured out all the details, and I'm excited that we can support `nw.when(nw.col('a').count()>0).then(nw.col('a').sum())` now 🥳

I just made two updates:
- not using private methods from Dask. In general, doing that kind of thing gives a very high chance of future failures, i'd advising against doing that whenever possible
- replace a couple of `if x is True` with just `if x` (unless there was some reason i missed for using `if x is True`?

camriddell:
- closes #2652 
- closes #2645

Issue #2556, opened 2025-05-16
enh: More math(s) functions: sqrt, exp, kurtosis


MarcoGorelli:
We're adding `log` in https://github.com/narwhals-dev/narwhals/pull/2549. We already have skew

We should also add:

- [x] #2668
- [x] #2667
- [ ] kurtosis (using Polars' defaults)

PR #2668, opened 2025-06-12
feat: add `Expr.sqrt` and `Series.sqrt`

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2556

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/tests/expr_and_series/sqrt_test.py
+++ b/tests/expr_and_series/sqrt_test.py
@@ -45,6 +45,6 @@ def test_sqrt_dtype_pandas_pyarrow() -> None:
     s = pd.Series([1.0, None, 2.0], name="a", dtype="Float32[pyarrow]", index=[8, 7, 6])
     result = nw.from_native(s, series_only=True).sqrt().to_native()
     expected = pd.Series(
-        [1, None, 1.414213], name="a", dtype="Float32[pyarrow]", index=[8, 7, 6]
+        [1, None, 1.414213], name="a", dtype="Float64[pyarrow]", index=[8, 7, 6]
     )
     pd.testing.assert_series_equal(result, expected)


narwhals/_duckdb/expr.py
```diff
@@ -878,6 +878,14 @@ def _exp(expr: Expression) -> Expression:
 
         return self._with_elementwise(_exp)
 
+    def sqrt(self) -> Self:
+        def _sqrt(expr: Expression) -> Expression:
+            return when(expr < lit(0), lit(float("nan"))).otherwise(
+                FunctionExpression("sqrt", expr)
+            )
+
+        return self._with_elementwise(_sqrt)
```
MarcoGorelli, line 887: i think this might throw a syntax error if we do `nw.col('a').sum().sqrt().over('b')`. which admittedly probably isn't very common 😄 we may need to define `_window_function` here too, so that the `over` gets pushed down appropriately

sorry if this isn't clear, i can provide details tomorrow

MarcoGorelli: Actually, this is OK as a follow-up, ignore my comment for now

narwhals/_pandas_like/series.py
```diff
@@ -1081,6 +1081,42 @@ def exp(self) -> Self:
...
+                self._implementation,
+                self._backend_version,
+                self._version,
+            )
+            result_native = native.__class__(
+                result_arr, dtype=out_dtype, index=native.index, name=native.name
+            )
+        else:
+            result_native = np.sqrt(native)
+        return self._with_native(result_native)
```
MarcoGorelli: would `self._with_native(self.native.pow(.5))` work here?

raisadz: yes seems to work 😅 thank you!

tests/expr_and_series/sqrt_test.py
```diff
@@ -0,0 +1,50 @@
...
+    expected = pd.Series([1, None, 1.414213], name="a", dtype="Float32", index=[8, 7, 6])
+    pd.testing.assert_series_equal(result, expected)
+
+
+@pytest.mark.skipif(PANDAS_VERSION < (2, 1, 0), reason="nullable types require pandas2+")
+def test_sqrt_dtype_pandas_pyarrow() -> None:
+    s = pd.Series([1.0, None, 2.0], name="a", dtype="Float32[pyarrow]", index=[8, 7, 6])
+    result = nw.from_native(s, series_only=True).sqrt().to_native()
+    expected = pd.Series(
+        [1, None, 1.414213], name="a", dtype="Float32[pyarrow]", index=[8, 7, 6]
```
MarcoGorelli: looks like the expected here needs changing to `Float64[pyarrow]`

I think that's fine, so long as the `[pyarrow]` part is preserved

MarcoGorelli:
```suggestion
        [1, None, 1.414213], name="a", dtype="Float64[pyarrow]", index=[8, 7, 6]
```

PR #2458, opened 2025-04-30
chore: bump `pyarrow-stubs==19.2`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #2124

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- [x] Bump to (https://github.com/zen-xu/pyarrow-stubs/releases/tag/19.2)
- [x] Resolve all new issues
- [ ] Do another pass for hidden workarounds that aren't needed

Issue #2385, opened 2025-04-14
[enh]: More permissive `group_by().agg(...)` for `pyarrow`


dangotbanned:
## Related
- Focused on **introducing** expressions in `group_by(...)`.
  - #1385 
  - #2325
- General
  - #2225
  - #1848
  - #1076
- https://github.com/zen-xu/pyarrow-stubs/pull/197
- https://github.com/zen-xu/pyarrow-stubs/pull/197#issuecomment-2781379291

## Description
While working on (https://github.com/zen-xu/pyarrow-stubs/pull/197), I learned some new things about [`pyarrow.Table.group_by.aggregate`](https://arrow.apache.org/docs/python/compute.html#py-grouped-aggrs).

### We *could* allow a wider range of functions
Currently we allow **10** different `Expr` aggregation methods:
https://github.com/narwhals-dev/narwhals/blob/16427440e6d74939c403083b52ce3fb0af7d63c7/narwhals/_compliant/group_by.py#L52

For `pyarrow`, they utilize *only* **9/21** native functions, that are allowed in a the native context.

<details><summary>9 native used</summary>
<p>

- [pc.approximate_median](https://arrow.apache.org/docs/python/generated/pyarrow.compute.approximate_median.html)
- [pc.count](https://arrow.apache.org/docs/python/generated/pyarrow.compute.count.html)
- [pc.count_distinct](https://arrow.apache.org/docs/python/generated/pyarrow.compute.count_distinct.html)
- [pc.max](https://arrow.apache.org/docs/python/generated/pyarrow.compute.max.html)
- [pc.mean](https://arrow.apache.org/docs/python/generated/pyarrow.compute.mean.html)
- [pc.min](https://arrow.apache.org/docs/python/generated/pyarrow.compute.min.html)
- [pc.stddev](https://arrow.apache.org/docs/python/generated/pyarrow.compute.stddev.html)
- [pc.sum](https://arrow.apache.org/docs/python/generated/pyarrow.compute.sum.html)
- [pc.variance](https://arrow.apache.org/docs/python/generated/pyarrow.compute.variance.html)

</p>
</details> 

<details><summary>How are they mapped over?</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/16427440e6d74939c403083b52ce3fb0af7d63c7/narwhals/_arrow/group_by.py#L29-L41

</p>
</details> 

*Some* also accept options for a small amount of flexibility. 
[...]

PR #2010, opened 2025-02-13
chore: fast path for pandas-like group by with single expression

FBruzzesi:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Addresses the [discord thread](https://discord.com/channels/1235257048170762310/1237430779584446546/1339395919799648266)

## Checklist

- [x] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below

I was able to see a slight improvement on the example provided, but definitly performance is still a bit behind.
Using py-spy I was not able to point it to something narwhals specifically does 🤔

PR #2680, opened 2025-06-15
chore: Simplify `PandasLikeGroupBy`

### LINKED ISSUE #2489: Simplify `PandasLikeGroupBy`
This method is over 200 LOC :flushed:

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_pandas_like/group_by.py#L82

dangotbanned:
Will close #2489

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2489
- **Partially unblocks** https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2085080744

- Hoping to make adding `Expr.(first|last)` less complicated

> [!NOTE]
> If you're a `pandas` expert, please feel free to hop on and add some commits 🙂 

## Tasks
- [x] Get all tests passing using `**named_aggs` [see thread](https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2157421442)
- [x] Clean up all the patchwork fixes to bugs (https://github.com/narwhals-dev/narwhals/pull/2680/commits/c2df42015ed3eae82aa9e1ec133b0e09069ae4d3)


dangotbanned:
So I've done some git archaeology ⛏️ and *I think* these PRs are where the complexity slowly crept in

### PRs
No criticisms from me, I've got the benefit of time and a change in scope on my side 😄 

- #917
- #1613
- #1645
- #1866
- #2010


### *This* PR
- I'd like to make the task of adding paths for new aggregations *less daunting*
- Currently, there's a lot of context to keep in your head while reading through `.agg()`
  - but no typing on the native side to ease the burden
- `std` and `var` could probably share more code
  - there's a fair bit of duplication that just needs to parameterize `self._grouped.<method_name>`
  - the complication is how the aggregations are collected and aggregated in multiple passes 
    - which loses the context of the `PandasLikeExpr`
- The counts in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/294c6de9659965661e3b5461883221cd20a71b12) are just a guide - purely reducing them isn't the goal

### Related
In the (#2680) version of `ArrowGroupBy`, there are some changes that might be helpful


<details><summary><code>ArrowGroupBy</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/5446095ef1d98cd97d4cd748888a59d52c083645/narwhals/_arrow/group_by.py#L28-L146

</p>
</details>

MarcoGorelli:
thanks for looking into this

i'd like to rewrite this to use namedagg, so it's probably worth doing that first https://github.com/narwhals-dev/narwhals/issues/1661

dangotbanned:
> thanks for looking into this
> 
> i'd like to rewrite this to use namedagg, so it's probably worth doing that first #1661

Thanks @MarcoGorelli, I've linked the two issues

No objections from me on how we get there 🙂 

Did you know `NamedAgg` is just a `NamedTuple`? 

- https://pandas.pydata.org/docs/dev/reference/api/pandas.NamedAgg.html
- https://github.com/pandas-dev/pandas/blob/c067bcd701e6cb4125e869a2802ef867d8395800/pandas/core/groupby/generic.py#L112-L114


## Edit
Yeah I've taken more of a look and I get it now. It's the **combination** of using `NamedAgg` and `**kwargs` that we'd aim for

- https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#named-aggregation

narwhals/_pandas_like/group_by.py
```diff
@@ -2,23 +2,104 @@
...
-    _REMAP_AGGS: ClassVar[Mapping[NarwhalsAggregation, Any]] = {
+    Ignoring most special-casing for now, just trying to work out the right shape.
+
+    The idea would be using this like:
+
+        df.groupby(...).agg(**named_aggs(..., ..., exclude=...)).reset_index()
+
+    Looks entirely different to the current `PandasLikeGroupBy` 🤔
+    """
+    return dict(chain.from_iterable(_named_aggs(gb, expr, exclude) for expr in exprs))
```
dangotbanned: Alright this is seeming promising now 😏 

Great shout on `NamedAgg` @MarcoGorelli!


<details><summary>test_all_kind_of_aggs</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/9f1600c7fa963a12b34fd24079a4f921c4c722e8/tests/frame/group_by_test.py#L353-L378

</p>
</details>

We'd be doing *roughly* this internally:

```py
from operator import methodcaller

import pandas as pd

data = {"a": [1, 1, 1, 2, 2, 2], "b": [4, 5, 6, 0, 5, 5]}
df = pd.DataFrame(data)
grouped = df.groupby("a", sort=False, as_index=True, dropna=False, observed=True)
result = (
    grouped.agg(
        c=pd.NamedAgg("b", "mean"),
        d=pd.NamedAgg("b", "mean"),
        e=pd.NamedAgg("b", "std"),
        f=pd.NamedAgg("b", methodcaller("std", ddof=2)),
        g=pd.NamedAgg("b", methodcaller("var", ddof=2)),
        h=pd.NamedAgg("b", methodcaller("var", ddof=2)),
        i=pd.NamedAgg("b", methodcaller("nunique", dropna=False)),
    )
    .reset_index()
    .sort_values("a")
)

>>> result.to_dict("list")
{'a': [1, 2],
 'c': [5.0, 3.3333333333333335],
 'd': [5.0, 3.3333333333333335],
 'e': [1.0, 2.8867513459481287],
 'f': [1.4142135623730951, 4.08248290463863],
 'g': [2.0, 16.666666666666668],
 'h': [2.0, 16.666666666666668],
 'i': [3, 2]}
```

dangotbanned: The `_agg_func` here is looking a lot easier to extend than the current `PandasLikeGroupBy.agg`

For (https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2085080744) I can quite easily adapt it for some branching on the backend version - to patch in functions that are consistent across `pandas` versions

dangotbanned: I'm *down* to **21** failures now locally, will probably push them mostly as-is tomorrow

Suprised how many tests passed without any special-casing

dangotbanned:
@MarcoGorelli maybe a longshot, but was wondering if you were interested in a fix for the `DataFrameGroupBy.aggregate` `@overload`(s)?

Both `mypy` and `pyright` aren't happy with them 😞 

<details><summary><code>mypy</code></summary>
<p>

```py
narwhals\_pandas_like\group_by.py:176: error: No overload variant matches argument type
"dict[str, tuple[str, Literal['any', 'all', 'count', 'first', 'idxmax', 'idxmin', 'last', 'max', 'mean', 'median', 'min', 'nunique', 'prod', 'quantile', 'sem', 'size', 'std', 'sum', 'var', 'cov', 'skew'] | Callable[..., Any]]]"  [call-overload]
                result = self._grouped.agg(**into_agg)  # type: ignore[arg-type]
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
narwhals\_pandas_like\group_by.py:176: note: Error code "call-overload" not covered by "type: ignore" comment
narwhals\_pandas_like\group_by.py:176: note: Possible overload variants:
narwhals\_pandas_like\group_by.py:176: note:     def aggregate(self, func: Literal['size']) -> Series[Any]
narwhals\_pandas_like\group_by.py:176: note:     def aggregate(self, func: Callable[..., Any] | str | ufunc | list[Callable[..., Any] | str | ufunc] | Mapping[Any, Callable[..., Any] | str | ufunc | list[Callable[..., Any] | str | ufunc]] | None = ..., *args: Any, engine: Literal['cython', 'numba'] | None = ..., engine_kwargs: _WindowingNumbaKwargs | None = ..., **kwargs: Any) -> DataFrame
```

</p>
</details> 


<details><summary><code>pyright</code></summary>
<p>

image

I'm pretty sure the gaps in **red** are causing the issues.


image

</p>
</details>

`**kwargs` could be typed using [`Unpack`](https://typing.python.org/en/latest/spec/callables.html#unpack-kwargs) and [PEP 728 – `TypedDict` with Typed Extra Items](https://peps.python.org/pep-0728/)

<details><summary><b>Example Fix</b></summary>
<p>


You can also specify the `pyright` setting as:

```toml
# pyproject.toml
[tool.pyright]
enableExperimentalFeatures = true
```

But below should be easier to copy/paste 🙂 

```py

# pyright: enableExperimentalFeatures=true

import numpy as np
import pandas as pd

if TYPE_CHECKING:
    from collections.abc import Hashable
    from typing import Any

    import typing_extensions as te
    from pandas._typing import AggFuncTypeFrame, WindowingEngine, WindowingEngineKwargs
    from pandas.core.groupby.generic import AggScalar, NamedAgg

    Aggregation: te.TypeAlias = NamedAgg | tuple[Hashable, AggScalar]

    class AggKwargs(te.TypedDict, extra_items=Aggregation, total=False):
        engine: WindowingEngine
        engine_kwargs: WindowingEngineKwargs


def df_groupby_agg(
    func: AggFuncTypeFrame | None = None, *args: Any, **kwargs: te.Unpack[AggKwargs]
) -> tuple[AggFuncTypeFrame | None, tuple[Any, ...], AggKwargs]:
    return func, args, kwargs


# NOTE: `DataFrameGroupBy._agg_examples_doc`
a = df_groupby_agg("min")
b = df_groupby_agg(["min", "max"])
d = df_groupby_agg(lambda x: sum(x) + 2)
e = df_groupby_agg({"B": ["min", "max"], "C": "sum"})
f = df_groupby_agg(
    b_min=pd.NamedAgg(column="B", aggfunc="min"),
    c_sum=pd.NamedAgg(column="C", aggfunc="sum"),
)
g = df_groupby_agg(lambda x: x.astype(float).min())

# NOTE: More fancy variants
m1 = df_groupby_agg("size")
m2 = df_groupby_agg(engine=None)
m3 = df_groupby_agg(engine_kwargs={"nogil": True}, alias_1=("column_1", "first"))
m4 = df_groupby_agg(
    column_1=("column_1", "last"),
    column_2=pd.NamedAgg("column_1", aggfunc=lambda x: sum(x) + 2),
)
m5 = df_groupby_agg(
    alias_2=pd.NamedAgg("column_2", "max"),
    alias_3=("column_3", np.min),
    engine="cython",
    alias_4=("column_4", "quantile"),
)
```

</p>
</details> 

And now none of these are causing yelling 😎 

image

narwhals/_pandas_like/group_by.py
dangotbanned: As of (https://github.com/narwhals-dev/narwhals/pull/2680/commits/b50456de9204f951a70857417f12d44d72c7305a) I'm down to **4** errors locally.

Haven't tracked down the issue for `tpch` yet, but the other one's are the `bool` columns - which might require a dedicated path

dangotbanned: `tpch` **Q1** is fixed with (https://github.com/narwhals-dev/narwhals/pull/2680/commits/12ae9af0b5868cf21326d002f6e7859c6e81b4ed)

> [!NOTE]
> The `pyarrow` tpch failures are unrelated to this PR
> [main](https://github.com/narwhals-dev/narwhals/actions/runs/15685085154/job/44186112369), [pr=2680](https://github.com/narwhals-dev/narwhals/actions/runs/15685101580/job/44186167111?pr=2680)


Now on to non-string columns (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2150539774) and then whatever else is turning up in CI 🤞

narwhals/_pandas_like/group_by.py
```diff
@@ -72,200 +123,97 @@ def agg(self, *exprs: PandasLikeExpr) -> PandasLikeDataFrame:  # noqa: C901, PLR
...
+            msg = (
+                "Unpacking the mapping of named aggregations fails "
+                "when non-`str` column names exist.\n"
+                f"{self._keys=}\n{new_names=}\n"
+            )
+            raise NotImplementedError(msg)
+        if all_aggs_are_simple:
+            result: pd.DataFrame
+            if named_aggs := self._named_aggs(*exprs, exclude=exclude):
+                result = self._grouped.agg(**named_aggs)  # type: ignore[call-overload]
```
dangotbanned: @FBruzzesi I was wondering if you had any ideas on a good way to solve this? 🙏 

## Problem
The complication is the tests which use `bool` - which I remember you dealt with in (#2325).

[`DataFrameGroupBy.agg(**kwargs)`](https://github.com/pandas-dev/pandas/blob/c067bcd701e6cb4125e869a2802ef867d8395800/pandas/core/groupby/generic.py#L1740) works for everything **provided all aliases are `str`**.

There's not really a way to work around this syntax issue *directly*:

<details><summary><b>Code block</b></summary>
<p>

```py
def agg(**named_aggs):
    return named_aggs


agg(**{"a": "hello"})  # noqa: PIE804
agg(**{True: "hello"})
```

</p>
</details> 

Even if we managed to hack our way past the first `**kwargs`, there's [another unpacking right after](https://github.com/pandas-dev/pandas/blob/c067bcd701e6cb4125e869a2802ef867d8395800/pandas/core/groupby/generic.py#L1915).

<details><summary><b><i>Somewhere</i> we'd always run into this</b></summary>
<p>


```py
Cell In[28], line 6
      2     return named_aggs
      5 agg(**{"a": "hello"})  # noqa: PIE804
----> 6 agg(**{True: "hello"})

TypeError: keywords must be strings
```

</p>
</details>


## Paths forward
The two *high-level* options I see (*but I'm very interested in alternatives* 🙏):

1. Intercept the expressions -> alias them with `str` names -> then pass to `.agg(**named_aggs)` -> unalias
2. Do something like the existing `PandasLikeGroupBy.agg`, where each aggregation is evaluated independently

### **(1)**
Generally I'd want to do that, but there's already a step like it for the `group_by(...)`.
It seems like it could get complicated to add a third `pd.DataFrame` into the mix 🤔 


<details><summary><b><i>Even more</i> code block</b></summary>
<p>


```py
from typing import cast

import pandas as pd

import narwhals as nw
from narwhals._pandas_like.group_by import PandasLikeGroupBy  # noqa: TC001

data = {True: [1, 2], False: [3, 4]}
native = pd.DataFrame(data)
df = nw.from_native(native)
nw_grouped = df.group_by(True)  # type: ignore  # noqa: FBT003
# boom = nw_grouped.agg(nw.col(False).max())
compliant_grouped = cast("PandasLikeGroupBy", nw_grouped._grouped)

original = compliant_grouped._df.native.columns
after_parse_exprs = compliant_grouped.compliant.native.columns

>>> original, after_parse_exprs
(Index([True, False], dtype='bool'),
 Index([True, False, '_True_tmp'], dtype='object'))
```

</p>
</details> 

### **(2)**
If we went down this route - I 100% need a `pandas` expert to babysit me 😄

FBruzzesi: Hey @dangotbanned , tomorrow I will try to go deeper in this PR.

AFAIK we didn't even try to fix that in the past and:
1. We provide the compliant (but actually wrong) type hint in `group_by` method, namely: `def group_by(self, keys: Sequence[str] | Sequence[PandasLikeExpr], ...)`
2. Then we _silence type checkers and linter big time_ when we explicitly do `group_by(True)` to test for pandas - [`test_select_boolean_cols`](https://github.com/narwhals-dev/narwhals/blob/2c8c25aa316e898f888f7b97c1a4ed0435bf7414/tests/frame/select_test.py#L58)

> There's not really a way to work around this syntax issue directly:

I guess not, once you (by which I mean pandas) allow anything hashable as column name, you shoot yourself in the foot for pinning down more specific type hinting 🙈

I understand that 1. (providing the wrong type hint in the signature) is not a _good_ practice, yet pragmatically if someone is using narwhals in the first place, it's very likely that they either:

- standardize/allow columns to be `str`
- pass whatever comes in wrapped around `nw.col(name)` - which correctly falls into `PandasLikeExpr`

dangotbanned: Thanks for getting to this so quickly @FBruzzesi!

Just to clarify, the problem I'm having is a **runtime** `TypeError` being raised (apologies if you already understood 😅)

FBruzzesi: > the problem I'm having is a runtime TypeError being raised

Whops then no I didn't realize that! Thanks for clarifying

dangotbanned: Okay @FBruzzesi so this is what I meant in (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2150539774) for **Paths forward (1)**

- [fix(DRAFT): Slap on non `str` support](https://github.com/narwhals-dev/narwhals/pull/2680/commits/3a905a6ffd617a26a60e9245db69c83730603b84)

I'm not really happy with:
- how many different places we'd need to iterate over the column names
- the 2 different methods of aliasing
- it only seems to solve *our* tests, but not the failures in
  - [marimo](https://github.com/narwhals-dev/narwhals/actions/runs/15704950591/job/44248454519?pr=2680)
  - [plotly](https://github.com/narwhals-dev/narwhals/actions/runs/15704950591/job/44248454601?pr=2680)

I feel like I should be using *more* native operations, I know there was something here before with using the `pd.DataFrame.columns` setter - maybe that's relevant?

narwhals/_pandas_like/group_by.py
```diff
@@ -1,24 +1,104 @@
...
-    _REMAP_AGGS: ClassVar[Mapping[NarwhalsAggregation, Any]] = {
+@lru_cache(maxsize=32)
+def _agg_func(
+    name: NativeAggregation, /, **kwds: Unpack[ScalarKwargs]
+) -> _AggFunc:  # pragma: no cover
+    if name == "nunique":
+        return methodcaller(name, dropna=False)
+    if not kwds or kwds.get("ddof") == 1:
+        return name
+    return methodcaller(name, **kwds)
```
dangotbanned: ## Investigate
- [`modin` doesn't understand `methodcaller`](https://github.com/narwhals-dev/narwhals/actions/runs/15704950607/job/44248454473?pr=2680)
- Need to figure out what they accept in `NamedAgg(aggfunc=...)`

narwhals/_pandas_like/group_by.py
```diff
@@ -1,24 +1,104 @@
...
+
 
-class PandasLikeGroupBy(EagerGroupBy["PandasLikeDataFrame", "PandasLikeExpr", str]):
-    _REMAP_AGGS: ClassVar[Mapping[NarwhalsAggregation, Any]] = {
+@lru_cache(maxsize=32)
+def _agg_func(
+    name: NativeAggregation, /, **kwds: Unpack[ScalarKwargs]
+) -> _AggFunc:  # pragma: no cover
+    if name == "nunique":
+        return methodcaller(name, dropna=False)
```
dangotbanned, line 64: ## Investigate
- Thinking [this `plotly` failure](https://github.com/narwhals-dev/narwhals/actions/runs/15704950591/job/44248454601?pr=2680) may be related to `nunique`
- It only seems to be an issue for `pandas_pyarrow_constructor`

Maybe there's some relation back to the same computation in `pyarrow`?

https://github.com/narwhals-dev/narwhals/blob/2c8c25aa316e898f888f7b97c1a4ed0435bf7414/narwhals/_arrow/group_by.py#L88-L89

narwhals/_pandas_like/group_by.py
dangotbanned: Possibly related?

- [ ] https://github.com/modin-project/modin/issues/7096

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/eea8e642c18454d511e9cc604df32b30c0d4c2ed)

tests/frame/group_by_test.py
```diff
@@ -540,3 +540,20 @@ def test_renaming_edge_case(constructor: Constructor) -> None:
...
+    """
+    data = {"a": [1, 2, 1, 2, 3, 4]}
+    expected = {"a": [1, 2, 3, 4], "len": [2, 2, 1, 1], "len_a": [2, 2, 1, 1]}
+    result = (
+        nw.from_native(constructor(data))
+        .group_by("a")
+        .agg(nw.len(), nw.len().alias("len_a"))
+        .sort("a")
+    )
+    assert_equal_data(result, expected)
```
dangotbanned: ### FIXME
Raises here as there isn't an available column

https://github.com/narwhals-dev/narwhals/blob/243f6cf3d72e193052770d6e47d770ec3184f851/narwhals/_pandas_like/group_by.py#L224-L226

Not sure how to resolve this as `pd.NamedAgg` needs `(column, aggfunc)` but there aren't any columns in scope 🤔

dangotbanned: Resolved-ish

- (https://github.com/narwhals-dev/narwhals/pull/2680/commits/80cfb1b95212ec80e602ee0c6934b7bffac9bf9a)
- (https://github.com/narwhals-dev/narwhals/pull/2680/commits/8fe49c73a0fb0b4407e84d76c9f3646e86a911df)

dangotbanned: Well `marimo` is passing, so I'll take that win 😄 

https://github.com/narwhals-dev/narwhals/actions/runs/15715899194/job/44285432927?pr=2680

narwhals/_pandas_like/group_by.py
dangotbanned: Okay so *I think* I've narrowed it down to `px._core.process_dataframe_hierarchy` in `plotly`
- https://github.com/plotly/plotly.py/blob/1ec864bbed58ee2834a879d323ade4b53bfde364/tests/test_optional/test_px/test_px_functions.py#L301-L303
- https://github.com/plotly/plotly.py/blob/1ec864bbed58ee2834a879d323ade4b53bfde364/plotly/express/_core.py#L1902-L2112

There's some pretty wild stuff going on, so I don't know *yet* how to repro it.

I do know that this initial `data` dataset is **supposed** to produce `9` unique colors - but *currently* has only `1`:

```py
data = {
    "vendors": ["A", "B", "C", "D", "E", "F", "G", "H"],
    "sectors": [
        "Tech",
        "Tech",
        "Finance",
        "Finance",
        "Tech",
        "Tech",
        "Finance",
        "Finance",
    ],
    "regions": ["North", "North", "North", "North", "South", "South", "South", "South"],
    "values": [1, 3, 2, 4, 2, 2, 1, 4],
    "total": ["total", "total", "total", "total", "total", "total", "total", "total"],
    "calls": [8, 2, 1, 3, 2, 2, 4, 1],
    "hover": ["a", "b", "c", "d", "e", "f", "g", "h"],
}
```

@FBruzzesi here we are again 😄 

Any wisdom would be **super** appreciated!

- https://github.com/plotly/plotly.py/commit/0389591c8714737e43d3f4285c7c50ee3de6a457
- https://github.com/plotly/plotly.py/commit/519cc680fe5c694682c0b41ce7672a4e31574f1e
- https://github.com/plotly/plotly.py/pull/4790

dangotbanned: I'm getting a bit closer to building up the repro for this today.
The full query is going to be huge, but once we have it as a `narwhals` test hopefully it'll expose whatever the edge case is 🙏

dangotbanned: Okay I've got a repro now

- [test: Add failing `plotly` repro](https://github.com/narwhals-dev/narwhals/pull/2680/commits/2317ff4511dec17fac6189c07572f1796ad81d59)

It's more minimal than the downstream code, but still waaaaaay too big to narrow down yet.

The issue is related to `n_unique` and `pyarrow`-backed pandas-like dataframes

FBruzzesi: Thanks @dangotbanned , I will try to take a look soon. Yet I am not so comfortable with pandas internals to know what's going on 😂 trial & error will be then

dangotbanned: Thank you!

> Yet I am not so comfortable with pandas internals to know what's going on 😂 trial & error will be then

Literally me this whole PR 😂

FBruzzesi: I pinned it down to the following comparison: `nw.when(nw.col(f"{name}{n_unique_token}") == 1)` 

For pandas[pyarrow] it fails to evaluate to True even when it should 🤔

From the debugger:

```py
(Pdb) discrete_aggs
['vendors', 'sectors', 'total', 'regions']

(Pdb) frame
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
|                                                                               Narwhals LazyFrame                                                                               |
|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|  vendors_path_copy sectors_path_copy regions_path_copy total_path_copy  count vendors vendors_N_UNIQUE regions regions_N_UNIQUE  total total_N_UNIQUE  sectors sectors_N_UNIQUE|
|0                 A              Tech             North           total      1       A                1   North                1  total              1     Tech                1|
|1                 B              Tech             North           total      1       B                1   North                1  total              1     Tech                1|
|2                 C           Finance             North           total      1       C                1   North                1  total              1  Finance                1|
|3                 D           Finance             North           total      1       D                1   North                1  total              1  Finance                1|
|4                 E              Tech             South           total      1       E                1   South                1  total              1     Tech                1|
|5                 F              Tech             South           total      1       F                1   South                1  total              1     Tech                1|
|6                 G           Finance             South           total      1       G                1   South                1  total              1  Finance                1|
|7                 H           Finance             South           total      1       H                1   South                1  total              1  Finance                1|
└────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

(Pdb) frame.with_columns(*[(nw.col(f"{name}{n_unique_token}") == 1).alias(name) for name in discrete_aggs])
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
|                                                                              Narwhals LazyFrame                                                                               |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|  vendors_path_copy sectors_path_copy regions_path_copy total_path_copy  count vendors vendors_N_UNIQUE regions regions_N_UNIQUE  total total_N_UNIQUE sectors sectors_N_UNIQUE|
|0                 A              Tech             North           total      1   False                1   False                1  False              1   False                1|
|1                 B              Tech             North           total      1   False                1   False                1  False              1   False                1|
|2                 C           Finance             North           total      1   False                1   False                1  False              1   False                1|
|3                 D           Finance             North           total      1   False                1   False                1  False              1   False                1|
|4                 E              Tech             South           total      1   False                1   False                1  False              1   False                1|
|5                 F              Tech             South           total      1   False                1   False                1  False              1   False                1|
|6                 G           Finance             South           total      1   False                1   False                1  False              1   False                1|
|7                 H           Finance             South           total      1   False                1   False                1  False              1   False                1|
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

FBruzzesi: It looks like they are all string columns, that's why:

```py
(Pdb) pprint(frame.schema)
Schema([('vendors_path_copy', String),
        ('sectors_path_copy', String),
        ('regions_path_copy', String),
        ('total_path_copy', String),
        ('count', Int64),
        ('vendors', String),
        ('vendors_N_UNIQUE', String),
        ('regions', String),
        ('regions_N_UNIQUE', String),
        ('total', String),
        ('total_N_UNIQUE', String),
        ('sectors', String),
        ('sectors_N_UNIQUE', String)])

(Pdb) pprint(frame.to_native().dtypes)
vendors_path_copy    string[pyarrow]
sectors_path_copy    string[pyarrow]
regions_path_copy    string[pyarrow]
total_path_copy      string[pyarrow]
count                 int64[pyarrow]
vendors              string[pyarrow]
vendors_N_UNIQUE     string[pyarrow]
regions              string[pyarrow]
regions_N_UNIQUE     string[pyarrow]
total                string[pyarrow]
total_N_UNIQUE       string[pyarrow]
sectors              string[pyarrow]
sectors_N_UNIQUE     string[pyarrow]
dtype: object
```

dangotbanned: @FBruzzesi nice one!

> It looks like they are all string columns, that's why:

But isn't everything supposed to be string columns?
There's loads of casts to `nw.String` in the downstream code (https://github.com/plotly/plotly.py/blob/1ec864bbed58ee2834a879d323ade4b53bfde364/plotly/express/_core.py#L1902-L2112)

Really weird that this didn't show up with the previous `group_by` 🤔

FBruzzesi: The output of this group by aggregation should/would not be a string:

```pypy
agg_f[f'{args["color"]}{n_unique_token}'] = (
    nw.col(args["color"])
    .n_unique()
    .alias(f'{args["color"]}{n_unique_token}')
)
```
🤔

dangotbanned: Ohh  I get it now, this is a much smaller native repro.

Why would we want string typed counts? 😂


```py
import pandas as pd

data = {
    "labels": [
        "A",
        "B",
        "C",
        "D",
        "E",
        "F",
        "G",
        "H",
        "Tech",
        "Finance",
        "Tech",
        "Finance",
        "North",
        "South",
        "total",
    ],
    "vendors": [
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
        "(?)",
    ],
}
```

```py
>>> pd.DataFrame(data).groupby("labels").agg(
    n_unique=("vendors", lambda x: x.nunique(dropna=False))
).reset_index().dtypes
labels      object
n_unique     int64
dtype: object
```

```py
>>> pd.DataFrame(data).convert_dtypes(dtype_backend="pyarrow").groupby("labels").agg(
    n_unique=("vendors", lambda x: x.nunique(dropna=False))
).reset_index().dtypes
labels      string[pyarrow]
n_unique    string[pyarrow]
dtype: object
```

FBruzzesi: > ```python
> >>> pd.DataFrame(data).groupby("labels").agg(
>     n_unique=("vendors", lambda x: x.nunique(dropna=False))
> ).reset_index().dtypes
> labels      object
> n_unique     int64
> dtype: object
> ```
> 
> ```python
> >>> pd.DataFrame(data).convert_dtypes(dtype_backend="pyarrow").groupby("labels").agg(
>     n_unique=("vendors", lambda x: x.nunique(dropna=False))
> ).reset_index().dtypes
> labels      string[pyarrow]
> n_unique    string[pyarrow]
> dtype: object
> ```

WTF mind blown!

dangotbanned: > WTF mind blown!

Even more confusing is this 😞:

```py
>>> pd.DataFrame(data).convert_dtypes(dtype_backend="pyarrow").groupby("labels").agg(
    n_unique=("vendors", "nunique") # lambda x: x.nunique(dropna=False))
).reset_index().dtypes
labels      string[pyarrow]
n_unique              int64
dtype: object
```

dangotbanned: No idea how we solve this yet - but ([test: Rewrite as minimal repro](https://github.com/narwhals-dev/narwhals/pull/2680/commits/82c31ba5a06646087e8fed8c0a46258c7b1870e3))

FBruzzesi: I didn't look at the changes (at all) but, you ended up using `lambda x: x.nunique(dropna=False)` to avoid having a dedicated path for `nunique` case as before and pass arguments?

dangotbanned: > I didn't look at the changes (at all) but, you ended up using `lambda x: x.nunique(dropna=False)` to avoid having a dedicated path for `nunique` case as before and pass arguments?

Yeah it's what the [`pd.NamedAgg`](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#named-aggregation) suggest for non-default keywords:

> When using named aggregation, additional keyword arguments are not passed through to the aggregation functions; only pairs of `(column, aggfunc)` should be passed as `**kwargs`.
> If your aggregation functions require additional arguments, apply them partially with `functools.partial()`

I started out (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2148818140) using `operator.methodcaller` - as `functools.partial` would be backend dependent (uses `SeriesGroupBy.nunique`).

There was an issue with `modin` not understanding that (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2151945563) so I've got a path which is essentially the same as a `lambda` but with typing.
But `pandas[pyarrow]` seems to do this weird string preservation on **any** callable

dangotbanned: Alright so the issue is here https://github.com/pandas-dev/pandas/blob/8a1d5a06f9fb3c232249e3ed301932053efb06d8/pandas/core/groupby/ops.py#L941-L972

`BaseGrouper.agg_series` is called with `preserve_dtype=True` - which isn't the default.
The actual result is an `int` until the conversion that happens there

dangotbanned: Looks like this was raised before and closed without a real solution 😔 

- https://github.com/pandas-dev/pandas/issues/57317

dangotbanned: Okay seems to have done the trick!

- [fix: Cast `string[pyarrow]` back to int](https://github.com/narwhals-dev/narwhals/pull/2680/commits/6fd3b4c4425ed7ec090a5a5f8eb1c0a98e53f69a)

Now I need to get back to *"simplifying"* - this is a mess currently 😄

narwhals/_pandas_like/dataframe.py
```diff
@@ -452,8 +452,10 @@ def filter(
     def with_columns(
         self: PandasLikeDataFrame, *exprs: PandasLikeExpr
     ) -> PandasLikeDataFrame:
+        if not exprs:
+            return self
         columns = self._evaluate_into_exprs(*exprs)
-        if not columns and len(self) == 0:
+        if not columns and len(self) == 0:  # pragma: no cover
```
dangotbanned: Hmm, I'm realising now that this was similar to the no-op fastpath I added in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/3a905a6ffd617a26a60e9245db69c83730603b84)

I'm not sure why the `len(self)` part was there - so I'll leave this here in case

narwhals/_pandas_like/group_by.py
```diff
@@ -274,9 +310,55 @@ def __iter__(self) -> Iterator[tuple[Any, PandasLikeDataFrame]]:
...
+    frame: PandasLikeDataFrame, subset: Sequence[str], /
+) -> Iterator[bool]:
+    version = frame._version
+    impl = frame._implementation
+    dtypes = version.dtypes
+    for col in subset:
+        native = frame.native.dtypes[col]
+        if str(native) != "object":
+            dtype = native_to_narwhals_dtype(native, version, impl)
+            yield isinstance(dtype, dtypes.String) and is_pyarrow_dtype_backend(
```
dangotbanned: I think this should actually be wider

```suggestion
            yield not (dtype.is_integer()) and is_pyarrow_dtype_backend(
```

In (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2157266811), the same thing happened for categorical types.
Probably a good idea to add tests for more than strings

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/46e5ae0bfea82cbd858f575ad77bba5a556f3772), (https://github.com/narwhals-dev/narwhals/pull/2680/commits/5a540ff6b94b8d5f1c28725c3e5c119c43b1ae63)

narwhals/_pandas_like/group_by.py
dangotbanned: Things are starting to shape up now 🥳 

- [refactor: Move all state into `AggExpr`, docs](https://github.com/narwhals-dev/narwhals/pull/2680/commits/c2df42015ed3eae82aa9e1ec133b0e09069ae4d3)

narwhals/_pandas_like/group_by.py
```diff
@@ -1,24 +1,219 @@
...
+        return lambda x: getattr(x, name)(**kwds)
+    return methodcaller(name, **kwds)
+
+
+def _n_unique(self: NativeSeriesGroupBy) -> pd.Series[Any]:
+    """`modin` gets confused by `operator.methodcaller`.
+
+    See https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2151945563
+    """
+    return self.nunique(dropna=False)
```
dangotbanned: > [!IMPORTANT]
> This section and `AggExpr` are the most relevant to (#2528)

In this new model, adding support for more kinds of aggregation (like `first`, `last`) can be achieved without making any *significant* changes to `PandasLikeGroupBy`.

The *only* change would be adding to `_REMAP_AGGS`, but that's needed for all `DepthTrackingGroupBy` anyway

<details><summary><code>_REMAP_AGGS</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/bff83c8e3a8bf5ce1006faf18818b23f37bc4040/narwhals/_pandas_like/group_by.py#L216-L227

</p>
</details> 

A nice thing is we've now got the all the upstream supported aggregations spelled out in `NativeAggregation`.
`any` and `all` might be easy wins for us in the future 😎 

There might also be some overlap with (#2385)

<details><summary><code>NativeAggregation</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/bff83c8e3a8bf5ce1006faf18818b23f37bc4040/narwhals/_pandas_like/group_by.py#L39-L62


</p>
</details> 

The specifics in (https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2085080744) can therefore be handled either in the cached `_native_agg` or that can make a call to another function if anything gets too complex.

narwhals/_utils.py
```diff
@@ -1191,7 +1191,7 @@ def generate_unique_token(
     return generate_temporary_column_name(n_bytes=n_bytes, columns=columns)
 
 
-def generate_temporary_column_name(n_bytes: int, columns: Sequence[str]) -> str:
+def generate_temporary_column_name(n_bytes: int, columns: Container[str]) -> str:
```
dangotbanned: I've split this into (#2712), but still need it here until that's merged

dangotbanned:
I've been thinking about this some more.

Although I prefer most aspects of the new API, two parts have been bugging me a lot:

1. The weirdness around needing to undo the casting that `pandas` does
  i. https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2157095151
  ii. https://github.com/pandas-dev/pandas/issues/57317
2. That [`.agg(**named_aggs)`](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#named-aggregation) doesn't support keyword arguments **natively**
  i. I'm ~80% sure that comes with a performance cost


I'm quite open to suggestions on how to adapt parts of this into using the `__getitem__` API instead

```py
DataFrameGroupBy[list(columns)].method(**kwargs)  # then renaming, etc
```

I really just wanted something *more structured* and easier to extend in the future, as we expand the kinds of expressions that can be used in aggregations 🙂

narwhals/_pandas_like/group_by.py
```diff
@@ -1,24 +1,218 @@
...
+    - Proxy all incoming expressions through a rename mapper
+    - At the end, rename back to the original
+    - An empty result follows a no-op path in `with_columns.
+    """
+    original = group_by._original_columns
+    exclude = set(group_by.exclude)
+    if remaining := set(original).difference(exclude):
+        union = exclude.union(original)
+        return {
+            name: generate_temporary_column_name(8, union)
```
FBruzzesi: In realistic terms, could `generate_temporary_column_name(8, union)`generate the same token twice in a row?

narwhals/_pandas_like/group_by.py
```diff
@@ -40,231 +249,121 @@ def __init__(
...
+        self._original_columns = tuple(df.columns)
         self._drop_null_keys = drop_null_keys
-        self._compliant_frame, self._keys, self._output_key_names = self._parse_keys(
-            df, keys=keys
+        ns = df.__narwhals_namespace__()
+        frame, self._keys, self._output_key_names = self._parse_keys(df, keys=keys)
+        self._exclude: tuple[str, ...] = (*self._keys, *self._output_key_names)
+        self._remap_non_str_columns = _remap_non_str(self)
+        self._compliant_frame = frame.with_columns(
+            *(ns.col(old).alias(new) for old, new in self._remap_non_str_columns.items())
```
FBruzzesi: By using rename, we should avoid a horizontal concat and other ops called in `with_columns`:

```suggestion
        self._remap_non_str_columns = _remap_non_str(self)
        self._compliant_frame = frame.rename(self._remap_non_str_columns)
```

narwhals/_pandas_like/group_by.py
```diff
@@ -1,24 +1,218 @@
...
     from narwhals._pandas_like.expr import PandasLikeExpr
+    from narwhals._utils import Implementation
+    from narwhals.dtypes import DType
+
+    NativeGroupBy: TypeAlias = "_NativeGroupBy[tuple[str, ...], Literal[True]]"
+    NativeSeriesGroupBy: TypeAlias = "_SeriesGroupBy[Any, tuple[str, ...]]"
+
+NativeApply: TypeAlias = "Callable[[pd.DataFrame], pd.Series[Any]]"
+InefficientNativeAggregation: TypeAlias = Literal["cov", "skew"]
+NativeAggregation: TypeAlias = Literal[
```
FBruzzesi, line 29: I assume this is taken from pands docs. I wonder if it makes sense to keep only those that we can realistically hit from Narwhals, and add them as we increase the expressions supported

FBruzzesi (commented):
Thanks @dangotbanned I just did a first pass. Will continue later and/or tomorrow.
By this first look it seems quite promising, I can definitely see how more modular and extensible it is

narwhals/_pandas_like/group_by.py
```diff
@@ -40,231 +249,121 @@ def __init__(
...
 
-        # dict of {output_name: root_name} that we count n_unique on
-        # We need to do this separately from the rest so that we
-        # can pass the `dropna` kwargs.
-        nunique_aggs: dict[str, str] = {}
-        simple_aggs: dict[str, list[str]] = collections.defaultdict(list)
-        simple_aggs_functions: set[str] = set()
+        if all_aggs_are_simple:
+            result: pd.DataFrame
+            if named_aggs := self._named_aggs(agg_exprs):
```
FBruzzesi: To contrast it with `_agg_complex`, I would rename `_named_aggs` to `_agg_simple`

narwhals/_pandas_like/group_by.py
```diff
@@ -40,231 +249,121 @@ def __init__(
...
-            native_frame = self.compliant.native.reset_index(drop=True)
-        else:
-            native_frame = self.compliant.native
-
-        self._grouped = native_frame.groupby(
+        native = self.compliant.native
+        if set(native.index.names).intersection(self.compliant.columns):
+            native = native.reset_index(drop=True)
+        self._grouped: NativeGroupBy = native.groupby(
             list(self._keys),
```
FBruzzesi: Nitpick: if we need a copy (which I don't recall why that's the case), I would prefer `self._keys.copy()`. I think it makes more explicit

narwhals/_pandas_like/group_by.py
```diff
@@ -40,231 +249,121 @@ def __init__(
...
+
+    def _apply_exprs(self, exprs: Iterable[PandasLikeExpr]) -> NativeApply:
+        ns = self.compliant.__narwhals_namespace__()
+        into_series = ns._series.from_iterable
+
+        def fn(df: pd.DataFrame) -> pd.Series[Any]:
             out_group = []
             out_names = []
             for expr in exprs:
                 results_keys = expr(self.compliant._with_native(df))
```
FBruzzesi, line 275: Can we move `self.compliant._with_native(df)` outside the for loop?

narwhals/_pandas_like/group_by.py
```diff
@@ -273,9 +372,67 @@ def __iter__(self) -> Iterator[tuple[Any, PandasLikeDataFrame]]:
...
+    return any(_has_non_int_nullable_dtype(frame, subset))
+
+
+def _has_non_int_nullable_dtype(
+    frame: PandasLikeDataFrame, subset: Sequence[str], /
+) -> Iterator[bool]:
+    version = frame._version
+    impl = frame._implementation
+    for col in subset:
+        native = frame.native.dtypes[col]
```
FBruzzesi: Nitpick: can we move `native_dtypes = frame.native.dtypes` outside of the for loop?

narwhals/_pandas_like/group_by.py
```diff
@@ -40,231 +249,121 @@ def __init__(
...
+            if named_aggs := self._named_aggs(agg_exprs):
+                result = self._grouped.agg(**named_aggs)  # type: ignore[call-overload]
+            else:
+                result = self.compliant.__native_namespace__().DataFrame(
+                    list(self._grouped.groups), columns=self._keys
+                )
+        elif self.compliant.native.empty:
+            raise empty_results_error()
+        else:
+            result = self._agg_complex(exprs)
```
FBruzzesi: Disclaimer 1: I know that this behavior was not introduced by this refactoring
Disclaimer 2: I am happy to keep that as a follow up 

We end up running even simple aggregations via apply, aren't we? I think we can (and eventually should) use apply only for those that are actually complex

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2680#pullrequestreview-2991562953)

Thanks for the review @FBruzzesi!

I didn't manage to get back to this today, hopefully will later this week 🙂

narwhals/_pandas_like/group_by.py
dangotbanned: Huh, for some reason I was assuming `self._keys` was a `Sequence[str]` - but it turns out I can't read my own documentation 🤣 

https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/group_by.py#L231-L232

narwhals/_pandas_like/group_by.py
dangotbanned: > [!NOTE]
> Most of this is just trying to explain the current behavior and how I landed on it

## Background
(https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2150539774) was where I first ran into the problem that needed solving.

So what you're suggesting *I think* aligns with how I was initially planning to solve this.

I ended up using `with_columns`, based on what I saw in:

https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_compliant/group_by.py#L80-L117


## Performance
> By using rename, we should avoid a horizontal concat and other ops called in `with_columns`

Coming from the perspective of the costs **I understand** in `python` and `narwhals` - here's what I've done:

- When **all** columns are `str`
  - `ns.col(old).alias(new)` is iterating over an empty `dict`
  - Since no `Expr`s are produced, `with_columns` is a noop 
  - https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/dataframe.py#L440-L443
- For everything else (which my *assumption* is that it is uncommon)
  - *Everything is relative to the number of non-`str` columns*
  - There's the initial cost of `ns.col(old).alias(new)` into a `with_columns`
  - This line https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/group_by.py#L304-L306
  - And then we finally use the names in a rename op at the end here https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/group_by.py#L334-L338

## What now?
If we can use some alternative (more performant) `pandas` API, while *either*:
- Preserving these noop paths
- Removing the need for them entirely

then I'm open to it! 🙂 

I'm happy to defer any decisions on stuff where you have a better understanding the `pandas` data model

narwhals/_pandas_like/group_by.py
dangotbanned: So the name `_agg_complex` has managed to survive since the 2nd commit of the PR (https://github.com/narwhals-dev/narwhals/pull/2680/commits/c61e09e21891edf34ca0e637754dff34c71b9ce7) 😅 

The *complex* part is coming entirely from the text of the warning message on main:

https://github.com/narwhals-dev/narwhals/blob/6bf173100b040670fbf6823d647d968e332ea59d/narwhals/_pandas_like/group_by.py#L236-L238

By comparison, `_named_aggs` was *named* 😉 later and is derived from the equivalent `pandas` API that we're abstracting over
- https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#named-aggregation
- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.NamedAgg.html#pandas.NamedAgg

---

I do agree that the two **should be aligned**, but my preference would be renaming `_agg_complex` to reference that we're using `apply`:
- https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#flexible-apply
- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html

All of these concerns are very specific to `pandas`, so the terms *simple* and *complex* seem less helpful to me; rather than what API we're using.
I think that leans into the idea that we're adapting `polars` concepts to the `pandas` backend.

I hope that made sense and I'm not just rambling 😄

narwhals/_pandas_like/group_by.py
FBruzzesi: Thanks for the full details! It makes a lot of sense, and indeed we should not lose performance of `with_columns` if all columns are already with string names.

Maybe the compromise would be to add a fast path in `rename`: if mapping is empty, return self.
That's what polars does anyway, but for pandas it would add an overhead running `.rename(columns={})`, so it might be worth to manage it ourselves. WDYT?

dangotbanned: Yeah a fast path in `PandasDataFrame.rename` (and others I suppose?) sounds like a good idea to me!

There's probably a few things that'll need to be *untangled* from the current logic to make this work - but if we *can* rename instead then the end result may be simpler as well 🙂

narwhals/_pandas_like/group_by.py
dangotbanned: Well spotted!

Turns out that ends up here https://github.com/pandas-dev/pandas/blob/d5f97ed21a872c2ea0bbe2a1de8b4242ec6a58d1/pandas/core/internals/managers.py#L337-L339
I assumed it would've at least been cached but I stand corrected 😄 

Resolved in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/da7ea98251b8715aaf15817c40567442ccf386fe)

narwhals/_pandas_like/group_by.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/b7e7b37b67217f96ead93cc4fbf8b245758aea08)

narwhals/_pandas_like/group_by.py
dangotbanned: Probably yeah

I hadn't changed much on this part, I think it was just moving those two lines outside of the function?


## `main`
https://github.com/narwhals-dev/narwhals/blob/6bf173100b040670fbf6823d647d968e332ea59d/narwhals/_pandas_like/group_by.py#L246-L255

## (#2680)
https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/group_by.py#L352-L366

narwhals/_pandas_like/group_by.py
FBruzzesi: > Yeah a fast path in PandasDataFrame.rename (and others I suppose?) sounds like a good idea to me!

Yes the annoying part is that it would be nice and easy to do it at the Narwhals level. Yet, when we need it here, we are operating at the compliant level.

Since for now this is something specific to pandas, I would address it only in https://github.com/narwhals-dev/narwhals/blob/6bf173100b040670fbf6823d647d968e332ea59d/narwhals/_pandas_like/dataframe.py#L460

---

I can take it and address it separately, but first run some benchmark to validate the hypothesis

narwhals/_pandas_like/group_by.py
FBruzzesi: Nice one! 👌🏼

narwhals/_pandas_like/group_by.py
dangotbanned: > I assume this is taken from pands docs.

I can confirm your assumption 😉 

https://github.com/narwhals-dev/narwhals/blob/de7347e0ec46cc555b7a9c1e254fb5c9b7bf5cd2/narwhals/_pandas_like/group_by.py#L61

For this part in the [`pandas` docs](https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html#built-in-aggregation-methods), I grouped them under `InefficientNativeAggregation`

> Of the methods listed below, those with a `*` do *not* have an efficient, GroupBy-specific, implementation.


---

> I wonder if it makes sense to keep only those that we can realistically hit from Narwhals, and add them as we increase the expressions supported

So there are a few benefits to the current situation, but they're all fairly minor:

1. Less legwork for whoever needs to keep the alias updated in the future
2. We have a clear target of what *could* be used
3. We know exactly what *can't* be used
4. The `Arrow` equivalent also contains things we haven't implemented yet
https://github.com/narwhals-dev/narwhals/blob/6bf173100b040670fbf6823d647d968e332ea59d/narwhals/_arrow/group_by.py#L28-L29
https://github.com/narwhals-dev/narwhals/pull/2458#discussion_r2068673611

If we keep it as-is, we get independent diagnostics for incorrect keys and values, which could be helpful if one forgets that `pandas` doesn't understand snake_case 🙂 

image

narwhals/_pandas_like/group_by.py
dangotbanned: I agree and *want* to say we leave this to a follow-up.
At the same time, I think if we're not doing **all aggregations** in a single path - then using the `agg(**named_aggs)` API is the wrong approach and we should try to do (https://github.com/narwhals-dev/narwhals/pull/2680#issuecomment-3015969184) instead 🤔 


In this block, one of the nice things is that everything is collected into a `DataFrame` for us

https://github.com/narwhals-dev/narwhals/blob/da7ea98251b8715aaf15817c40567442ccf386fe/narwhals/_pandas_like/group_by.py#L282-L289

Then in the `apply` case, we also have everything going into a `DataFrame`

https://github.com/narwhals-dev/narwhals/blob/da7ea98251b8715aaf15817c40567442ccf386fe/narwhals/_pandas_like/group_by.py#L342-L350

If we want to branch per-`PandasLikeExpr`, then having multiple `Series` outputs would be *conceptually* easier to put back together.

We also wouldn't need to handle non-`str` column names, since they aren't rejected by `__getitem__`


Also we could just do all of the operations in order of appearance in a single pass, rather than an initial pass for complexity and a second one for execution

https://github.com/narwhals-dev/narwhals/blob/da7ea98251b8715aaf15817c40567442ccf386fe/narwhals/_pandas_like/group_by.py#L274-L282

narwhals/_pandas_like/group_by.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/532d6b3cbaf9db004dae53574e5e4cea45f4173d)

I renamed `_apply_exprs` as well to try and hint towards the fact it produces a function from expressions

narwhals/_pandas_like/group_by.py
dangotbanned: So the source of randomness seems to be platform-dependent.

The only part the [docs](https://docs.python.org/3/library/secrets.html#how-many-bytes-should-tokens-use) state that I could link to a collision would be:

> As of 2015, it is believed that 32 bytes (256 bits) of randomness is sufficient for the typical use-case expected for the [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module.

I would say if a collision is a concern, we should be using **32** instead of **8**.

Or maybe have a function specifically for multiple names, where 
- we use `token_hex` for a prefix that no columns have
- then enumerate for a suffix, so `"{unique_prefix}_{index}"`

narwhals/_pandas_like/dataframe.py
FBruzzesi: In both these condition, I am not sure we should return `self`.

```py
import polars as pl

data = {"a": [1,2,3]}
frame = pl.DataFrame(data)

frame is frame.with_columns()
False
```

but here it would be:

```py
import narwhals as nw
import pandas as pd

data = {"a": [1,2,3]}
frame = pd.DataFrame(data)

frame is nw.from_native(frame).with_columns().to_native()
True
```

narwhals/_pandas_like/group_by.py
```diff
@@ -340,12 +340,23 @@ def _select_results(
...
+
+        - [User guide] mentions `include_groups` 4 times without deprecation.
+        - [`DataFrameGroupBy.apply`] doc says the default value of `True` is deprecated since `2.2.0`.
+        - `False` is explicitly the only *non-deprecated* option, but entirely omitted since [pandas-dev/pandas-stubs#1268].
+
+        [pandas-dev/pandas-stubs#1270]: https://github.com/pandas-dev/pandas-stubs/issues/1270
+        [User guide]: https://pandas.pydata.org/pandas-docs/stable/user_guide/groupby.html
+        [`DataFrameGroupBy.apply`]: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html
+        [pandas-dev/pandas-stubs#1268]: https://github.com/pandas-dev/pandas-stubs/pull/1268
+        """
```
dangotbanned, line 257: @FBruzzesi just some extra context on how I got here 😭 

For some unknown reason, I *suddenly* needed to switch my `pylance` to pre-release (https://github.com/microsoft/pylance-release/releases/tag/2025.6.101) to enable `pandas-stubs`.

Before I got to that solution, I tried reinstalling `pandas-stubs` - which I guess ignored our pin and went straight to 

https://github.com/narwhals-dev/narwhals/blob/5fd4160b7b098e64ddbdaf15a70a5a565dde1af1/pyproject.toml#L68

Finally, once I had `pandas-stubs` working again - I then ran into the rabbit hole I've documented here 🤦‍♂️

I considered just rolling everything back, but the situation was so frustrating that  I don't want to punt it to someone else to discover in the future 😅

narwhals/_pandas_like/group_by.py
dangotbanned: So I made the smallest change I could in ([perf: Use `rename` instead of `with_columns`](https://github.com/narwhals-dev/narwhals/pull/2680/commits/cfeac25b3dab9b64330cec1bfce07adb8fe0493d)) which seems to have passed everything except `scikit-lego`

https://github.com/narwhals-dev/narwhals/actions/runs/16168806351/job/45636975071?pr=2680

It looks like of the **141** failures, most are some version of `ColumnNotFoundError` with a list of `int` as being missing

## Update 1
Not had luck tracking down the issue

@FBruzzesi I could really use some help here 🙏 

An example would be:
- https://github.com/koaning/scikit-lego/blob/a39e6d7a33a9c6fdd16faa2b3a23aa2dc58ab0e7/tests/test_meta/test_grouped_transformer.py#L129-L150

which seems to depend on:
- https://github.com/koaning/scikit-lego/blob/a39e6d7a33a9c6fdd16faa2b3a23aa2dc58ab0e7/sklego/meta/grouped_transformer.py#L12-L265

And the error is:

```py
FAILED tests/test_meta/test_grouped_transformer.py::test_group_correlation_minmaxscaler[dataset_with_single_grouping7] - 

narwhals.exceptions.ColumnNotFoundError: The following columns were not found: [1]

Hint: Did you mean one of these columns: [0, 'nwfea1527928747e', '__sklego_target__', '_0_tmp____________']?
```

Like I can see that renaming has failed, but I don't know where to begin with making a minimal repro out of the `scikit-lego` tests 😔

FBruzzesi: Hey @dangotbanned in scikit-lego we heavily use group_by's, so that's a good test to have. I will take a look tomorrow. Is it possible to be related with returning the same object vs a copy of it? Similar to the comment I made here: https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2193400231

dangotbanned: > Is it possible to be related with returning the same object vs a copy of it?

I had a similar thought in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/24eb87301c07ba4255c5d0ee0090a9a84620a8cc) but I assumed these were all cases where we *actually* rename.
Whereas (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2193400231) is for the inverse IIUC?

FBruzzesi: Nevermind, the issue in is `__iter__`. As `_grouped` has the columns renamed already, we are unable to find the original ones

FBruzzesi: Regarding returning a copy vs original object: I am not sure what's the best way of achieving it.

- `rename` allows empty mapping (and returns a copy)
- `assign` allows empty args as well (and returns a copy) (could this be the fastpath in `with_columns`? or just use `native.copy()`?)

They seem to run with similar speed - I assume both rename and assign without args just return a copy without other operations (aside checking the args):

```py
import numpy as np
import pandas as pd

frame = pd.DataFrame(np.random.randn(1_000_000, 100))

%timeit frame.rename(columns={})
458 ms ± 3.18 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit frame.assign()
461 ms ± 6.79 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit frame.copy()
467 ms ± 8.89 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


frame is frame.rename(columns={})
False

frame is frame.assign()
False
```

dangotbanned: > Regarding returning a copy vs original object: I am not sure what's the best way of achieving it.

> They seem to run with similar speed - I assume both rename and assign without args just return a copy without other operations (aside checking the args):

I must say, I am tempted to discuss how we can optimize this 😅 

But I think the priorities we should have at the moment are:

1. Getting the [`scikit-lego` tests](https://github.com/narwhals-dev/narwhals/actions/runs/16169431080/job/45639010549?pr=2680) green again
2. Adding a *more minimal* test case to `narwhals` that would pass on `main` and in whatever fix we arrived at in step **1**
3. Removing the parts of this PR that still assume `PandasLikeGroupBy.__init__` is adding columns aliased columns
  i. This might be the issue itself, but I haven't tracked it down yet 😔 

Then we can figure out any perf bottlenecks and see if taking any of the paths you've mentioned cause anything to regress 🙂

FBruzzesi: Hey @dangotbanned yes I agree. Those two issue should be addressed separately 👌🏼
For a minimal repro for a test case, I think it could be enough to have a column with a numeric name, that is not part of the keys.

```py
import narwhals as nw
import pandas as pd

frame = pd.DataFrame({0: [1, 2], 1: ["a", "b"]})
list(nw.from_native(frame).group_by(1).__iter__())
```

> ColumnNotFoundError: The following columns were not found: [0]
>
> Hint: Did you mean one of these columns: ['nw88d47f91c139ef', np.int64(1), '_1_tmp']?

dangotbanned: ohhh that was what you meant by

> Nevermind, the issue in is `__iter__`. As `_grouped` has the columns renamed already, we are unable to find the original ones

🤦‍♂️ 🤦‍♂️ 🤦‍♂️


So what we need is coverage here for the non-`str` columns case?

https://github.com/narwhals-dev/narwhals/blob/46282e8146e60deec5084595d684f54967831c4c/narwhals/_pandas_like/group_by.py#L377-L386

FBruzzesi: Yes apologies for being vague without a repro! That's exactly what we should target for coverage and fixing scikit-lego issue

dangotbanned: Ahhh thank you @FBruzzesi, we're now unbroken 😄 

- https://github.com/narwhals-dev/narwhals/actions/runs/16201526949/job/45741527522?pr=2680
- (https://github.com/narwhals-dev/narwhals/pull/2680/commits/9eb672e74a461e5fb65e6ee362206ee0629b1785)

Just need to clean everything up now

narwhals/_pandas_like/group_by.py
dangotbanned: @FBruzzesi I'm most of the way finished on converting everything back to the `__getitem__` api

All tests are passing, and I even managed to get `modin` to support some extra ops 😄

FBruzzesi: Amazing work! I closed/resolved some threads to clean up conversation. Ping me here or on discord when PR is ready for a final pass of review 😉

narwhals/_pandas_like/group_by.py
dangotbanned: Marking as resolved, as of (https://github.com/narwhals-dev/narwhals/pull/2680/commits/91b5800e292266dc83c1c30f5f2fe416141b1de4) original column names are preserved throughout

narwhals/_pandas_like/group_by.py
dangotbanned: Same as (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2200616113), this part was removed

narwhals/_pandas_like/group_by.py
dangotbanned: Just saying that I agree, but recommend doing as a follow-up

dangotbanned:
@FBruzzesi feel free to delete any docs that seem unnecessary

The API has shrunk a lot in (https://github.com/narwhals-dev/narwhals/pull/2680/commits/91b5800e292266dc83c1c30f5f2fe416141b1de4), so if what's left is easier enough to understand without docs - then I'm not precious on keeping them 😄

FBruzzesi (approved):
Thanks for all the back and forth @dangotbanned - I have no more objections! This looks 🔥  

It covers so many edge cases that a few times I found myself thinking "wait a minute, let me change this line, I don't think we need it", well actually we do 😂

<div></div>

MarcoGorelli:
```
______________________ test_double_same_aggregation[cudf] ______________________
[XPASS(strict)] 
_________________________ test_all_kind_of_aggs[cudf] __________________________
[XPASS(strict)] 
```

nice one, well done both!

dangotbanned:
> ```
> ______________________ test_double_same_aggregation[cudf] ______________________
> [XPASS(strict)] 
> _________________________ test_all_kind_of_aggs[cudf] __________________________
> [XPASS(strict)] 
> ```
> 
> nice one, well done both!

Ahh I was curious about if this was fixed for all of them 🥳

PR #2683, opened 2025-06-15
feat: Add `order_by` option in `with_row_index` method

### LINKED ISSUE #2307: [Enh]: Allow `LazyFrame.with_row_index(..., order_by=...)`
# Description

Proposing to allow the following: `LazyFrame.with_row_index(name, order_by=into_exprs)`

This would be a deviation from polars itself.
In a way this is similar to #2174

## Suggested implementation

I am missing some detail for sure, but it would be something close to:

```py
self.with_column(**{name, nw.lit(1).cum_sum().over(order_by=order_by) - 1})
```

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2307

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/tests/frame/with_row_index_test.py
+++ b/tests/frame/with_row_index_test.py
@@ -32,7 +32,9 @@ def test_with_row_index_eager(constructor_eager: ConstructorEager) -> None:
 def test_with_row_index_lazy(
     constructor: Constructor, order_by: str | Sequence[str], expected_index: list[int]
 ) -> None:
-    if "pandas" in str(constructor) and PANDAS_VERSION < (1, 3) and order_by == "abc":
+    if (
+        "pandas" in str(constructor) and PANDAS_VERSION < (1, 3) and order_by == "abc"
+    ):  # pragma: no cover
         reason = "ValueError: first not supported for non-numeric data."
         pytest.skip(reason=reason)
 


narwhals/_arrow/dataframe.py
```diff
@@ -478,14 +478,17 @@ def to_dict(
...
-        cols = self.columns
-
-        row_indices = pa.array(range(df.num_rows))
-        return self._with_native(
-            df.append_column(name, row_indices).select([name, *cols])
-        )
+    def with_row_index(self, name: str, order_by: str | Sequence[str] | None) -> Self:
+        if order_by is None:
+            df = self.native
+            row_index = pa.array(range(df.num_rows))
```
FBruzzesi: Should we use 

```suggestion
            row_index = pa.array(np.arange(df.num_rows))
```

instead?

narwhals/_pandas_like/dataframe.py
```diff
@@ -415,13 +415,21 @@ def estimated_size(self, unit: SizeUnit) -> int | float:
...
-        namespace = self.__narwhals_namespace__()
-        row_index = namespace._series.from_iterable(
-            range(len(frame)), context=self, index=frame.index
-        ).alias(name)
-        return self._with_native(namespace._concat_horizontal([row_index.native, frame]))
+        index = frame.index
+        plx = self.__narwhals_namespace__()
+        if order_by is None:
+            row_index = plx._series.from_iterable(
+                range(len(frame)), context=self, index=index
```
FBruzzesi: Should we use `np.arange/cupy.arange` instead?

narwhals/_polars/dataframe.py
```diff
@@ -440,10 +440,40 @@ def group_by(
...
+    def _with_row_index_order_by_single(self, name: str, order_by: str) -> Self:
+        result = self.native.select(
+            pl.col(order_by).rank(method="ordinal").alias(name) - 1, pl.all()
+        )
+        return self._with_native(result)
+
+    @requires.backend_version((1, 9))
+    def _with_row_index_order_by_multi(self, name: str, order_by: Sequence[str]) -> Self:
+        # Before polars 1.9.0, either `.over(partition_by=...)` is not implemented or it does
+        # not properly break ties with multiple columns in partition_by
```
FBruzzesi: Empirically true 🙈

narwhals/_compliant/dataframe.py
```diff
@@ -364,7 +364,30 @@ def unpivot(
...
+        )
+
+    def _with_row_index_order_by_multi(self, name: str, order_by: Sequence[str]) -> Self:
+        plx = self.__narwhals_namespace__()
+        columns = self.columns
+        const_expr = plx.lit(value=1, dtype=None).alias(name).broadcast(ExprKind.LITERAL)
+        row_index_expr = (
+            plx.col(name).cum_sum(reverse=False).over(partition_by=[], order_by=order_by)
+            - 1
+        )
```
FBruzzesi: I moved from `rank` to `cum_sum` because it was failing in (for me...) mysterious ways.

As in polars, even using `rank` requires to create a temporary column (even if constant)

FBruzzesi:
I let https://github.com/narwhals-dev/narwhals/pull/2683/commits/87ebad0ac438942235945b2718bc2d9e50f0b195 sneak in - it avoids using `range(size)` whenever possible. For any large sized dataset this might matter

narwhals/dataframe.py
```diff
@@ -2431,29 +2436,50 @@ def drop_nulls(self, subset: str | list[str] | None = None) -> Self:
...
 
-    def with_row_index(self, name: str = "index") -> Self:
+    def with_row_index(
+        self, name: str = "index", order_by: str | Sequence[str] | None = None
+    ) -> Self:
         """Insert column which enumerates rows.
 
         Arguments:
             name: The name of the column as a string. The default is "index".
+            order_by: Column(s) to order by when computing the row index. Must be not None.
```
FBruzzesi: I think it's good to specify that `order_by` must be not None although it's actually allowed for backward compat for polars and dask - all other lazy backends will raise an exception

narwhals/_arrow/dataframe.py
```diff
@@ -478,14 +478,19 @@ def to_dict(
...
+    def with_row_index(self, name: str, order_by: str | Sequence[str] | None) -> Self:
+        if order_by is None:
+            import numpy as np  # ignore-banned-import
 
-        row_indices = pa.array(range(df.num_rows))
-        return self._with_native(
-            df.append_column(name, row_indices).select([name, *cols])
-        )
+            df = self.native
+            row_index = pa.array(np.arange(df.num_rows, dtype=np.int64))
```
EdAbati: out of curiosity, why do we need to switch to `np.arange` vs `range` (as we had before)? is it more efficient?

narwhals/_compliant/dataframe.py
```diff
@@ -364,7 +364,36 @@ def unpivot(
...
+        return self.with_columns(const_expr).select(row_index_expr, plx.col(*columns))
+
+    def with_row_index(self, name: str, order_by: str | Sequence[str] | None) -> Self:
+        if order_by is None:
+            msg = (
+                "`LazyFrane.with_row_index` requires `order_by` to be specified as it is an "
```
EdAbati:
```suggestion
                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
```

narwhals/dataframe.py
EdAbati: random idea: what if we allow `None` for dask and polars in v1 only?

at the moment the docstring contradicts the type of the args. I can see it may be confusing

tests/frame/with_row_index_test.py
```diff
@@ -1,20 +1,60 @@
...
+    result = (
+        nw.from_native(constructor(data)).with_row_index(order_by=order_by).sort("xyz")
+    )
+    expected = {"index": expected_index, **data}
     assert_equal_data(result, expected)
+
+
+def test_with_row_index_lazy_exception(constructor: Constructor) -> None:
+    msg = (
+        "`LazyFrane.with_row_index` requires `order_by` to be specified as it is an "
```
EdAbati:
```suggestion
        "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
```

narwhals/_polars/dataframe.py
```diff
@@ -695,10 +725,40 @@ def group_by(
...
+        result = self.native.with_columns(
+            pl.lit(value=True, dtype=pl.Boolean()).alias(name)
+        ).select(
+            pl.col(name)
+            .rank(method="ordinal", descending=False)
+            .over(partition_by=None, order_by=order_by)
+            - 1,
+            pl.all().exclude(name),
+        )
+        return self._with_native(result)
```
EdAbati: Maybe in a follow up we should create a parent `PolarsBaseFrame` class that has all common methods (to avoid repetitions)

narwhals/_arrow/dataframe.py
FBruzzesi: I remember this coming up in the past as well and especially for pyarrow it had a large impact:

```py
In [5]: %timeit pa.array(np.arange(100_000))
35 µs ± 220 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [6]: %timeit pa.array(range(100_000))
9.36 ms ± 675 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

For a size of 100k, using `np.arange` is 200+ times faster 😂

For pandas-like:

- cudf -> I would rather keep everything on gpu
- other -> empirically I cannot see a large difference, maybe on the ballpark of 10 to 20% in favor of `np.arange`

narwhals/dataframe.py
FBruzzesi: > random idea: what if we allow None for dask and polars in v1 only?

Yes that's probably the best way to go

> at the moment the docstring contradicts the type of the args. I can see it may be confusing

One tiny issue is that we have another argument `name` which has a default already, so the next one also requires a default

narwhals/_polars/dataframe.py
FBruzzesi: YES

FBruzzesi: @EdAbati [here we go](https://github.com/narwhals-dev/narwhals/pull/2689)

FBruzzesi:
@MarcoGorelli all change requests should be addressed in https://github.com/narwhals-dev/narwhals/pull/2683/commits/bfbdc52c844c0e50e3e8d4eec5ae4c2b2916b975

@EdAbati regarding v1 vs main namespace, this commit takes care of it: https://github.com/narwhals-dev/narwhals/pull/2683/commits/17ef19511926624b9057f3c139496f2ced689ddc

narwhals/dataframe.py
MarcoGorelli: you can make it non-required by making it keyword-only, 
```python
    def with_row_index(
        self, name: str = "index", *, order_by: str | Sequence[str]
    ) -> Self:
```

narwhals/_spark_like/dataframe.py
```diff
@@ -522,11 +522,25 @@ def unpivot(
             unpivoted_native_frame = unpivoted_native_frame.drop(*ids)
         return self._with_native(unpivoted_native_frame)
 
+    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
```
MarcoGorelli: anything which previously had this as `not_implemented` has no backcompat issues, so we can just remove the `None` default right away

MarcoGorelli (changes requested):
thanks, this is getting closer!

can we also add a note in `docs/backcompat.md`?

narwhals/dataframe.py
MarcoGorelli: you can make it required by making it keyword-only, 
```python
    def with_row_index(
        self, name: str = "index", *, order_by: str | Sequence[str]
    ) -> Self:
```

narwhals/_spark_like/dataframe.py
FBruzzesi: It would still be possible to reach `LazyFrame.with_row_index(..., order_by=None)` of each compliant frame from the v1 namespace though 🤔
Shouldn't we control  and raise explicitly for such situation?

MarcoGorelli: i wouldn't bother, the idea with v1 is "code you've written today will keep running in the future", if there's something which it's not possible to have already written then it's not our concern

narwhals/_arrow/expr.py
```diff
@@ -129,8 +129,11 @@ def shift(self, n: int) -> Self:
         return self._reuse_series("shift", n=n)
 
     def over(self, partition_by: Sequence[str], order_by: Sequence[str]) -> Self:
-        assert self._metadata is not None  # noqa: S101
-        if partition_by and not self._metadata.is_scalar_like:
+        if (
+            partition_by
+            and self._metadata is not None
```
MarcoGorelli, line 134: can we keep the debug assertion?

narwhals/_duckdb/dataframe.py
```diff
@@ -504,10 +505,21 @@ def unpivot(
...
+    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
+        if order_by is None:
+            msg = (
+                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
+                "order-dependent operation."
+            )
+            raise ValueError(msg)
+        order_by_sql = generate_order_by_sql(*order_by, ascending=True)
+        rel = self.native  # noqa: F841
+        query = f"select row_number() over ({order_by_sql}) -1 as {name}, * from rel"  # noqa: S608
```
MarcoGorelli: we may want to use `SQLExpression`, `rel.select`, and `.alias`? this would require duckdb>=1.3, but i think that's ok

we can also use `FunctionExpression('row_number')`

MarcoGorelli (commented):
really nice, just some comments

narwhals/_spark_like/dataframe.py
dangotbanned: From a typing perspective, the `None` is an issue - since it's defined in `CompliantLazyFrame`

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_compliant/dataframe.py#L368

But only valid for:
- `PolarsLazyFrame`
- `DaskLazyFrame`
- `EagerDataFrame`

## Alternative
We can just test for truthiness on `order_by` when it is `Sequence[str]`.

This definition will be falsy for any of `""`, `()`, `[]`, etc

```py
class CompliantLazyFrame:
    def with_row_index(self, name: str, order_by: Sequence[str]) -> Self: ...
```

That would remove the need to use `assert` or ignoring the `None` case, we'd just defer to the backend raising here 

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_spark_like/dataframe.py#L534-L535


But in the places where it is valid, we just do:

```py
if order_by:
    ...
else:
    ...
```

narwhals/_compliant/dataframe.py
```diff
@@ -401,6 +402,10 @@ def __narwhals_namespace__(
     def to_narwhals(self) -> DataFrame[NativeFrameT]:
         return self._version.dataframe(self, level="full")
 
+    def _with_native(
+        self, df: NativeFrameT, *, validate_column_names: bool = True
+    ) -> Self: ...
```
dangotbanned, line 406: Huh, I really missed this one 😄 

@FBruzzesi I think I'm gonna add a mini PR for `Compliant(Data|Lazy)Frame._with_native` and this one as well.

We can move them after (#2713), but it's a pretty core method so it should've been here for you already!

narwhals/_arrow/expr.py
dangotbanned: I prefer w/o the assert, but you can also shorten it with a walrus:

```py
        if partition_by and (metadata := self._metadata) and not metadata.is_scalar_like:
```

It might be worth (separate to this PR), to document under what conditions it can be `None` in:

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_compliant/expr.py#L81-L82

If the `assert` fails, there isn't a meaningful traceback or any docs to fall back on what the failure means

narwhals/dataframe.py
dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160018155)

That's still got a syntax error 

image


Kinda related to (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160323551)

You can do this and remove the mention of `None` in the docs (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2151498986)

```py
    def with_row_index(
        self, name: str = "index", order_by: str | Sequence[str] = ()
    ) -> Self:
        if not order_by: ...
```

If you want something more explicit, you could use a sentinel like in https://github.com/pola-rs/polars/blob/2494fcc9f9d2f34db2875e485bb4ce10bf8d97a5/py-polars/polars/expr/expr.py#L10419-L10426

narwhals/_arrow/expr.py
MarcoGorelli: it's not meant to be `None` when we get here, hence my preference for the debug assertion

dangotbanned: > it's not meant to be `None` when we get here

I understand that it's not meant to be, but in (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160335652) I'm saying it isn't documented why and the assertion alone doesn't help:

```py
import pyarrow as pa

data = {"a": ["A", "B", "A"], "b": [1, 2, 3]}
table = pa.table(data)
arrow_ns = Version.MAIN.namespace.from_backend("pyarrow").compliant
arrow_df = arrow_ns.from_native(table)
>>> arrow_df.select(arrow_ns.col("b").max().over("a", ()))
```

AFAICT, any metadata-dependent `CompliantExpr` cannot be created without starting as an `Expr`.
But I think we can do better than this in terms of explaining

```py
AssertionError                            Traceback (most recent call last)
Cell In[1], line 9
      7 arrow_ns = Version.MAIN.namespace.from_backend("pyarrow").compliant
      8 arrow_df = arrow_ns.from_native(table)
----> 9 arrow_df.select(arrow_ns.col("b").max().over("a", ()))

File c:\Users\danie\Documents\GitHub\narwhals\narwhals\_arrow\expr.py:132, in ArrowExpr.over(self, partition_by, order_by)
    131 def over(self, partition_by: Sequence[str], order_by: Sequence[str]) -> Self:
--> 132     assert self._metadata is not None  # noqa: S101
    133     if partition_by and not self._metadata.is_scalar_like:
    134         msg = "Only aggregation or literal operations are supported in grouped `over` context for PyArrow."

AssertionError:
```

narwhals/_arrow/dataframe.py
```diff
@@ -481,14 +481,28 @@ def to_dict(
...
+        else:
+            order_by_ = [order_by] if isinstance(order_by, str) else order_by
+            plx = self.__narwhals_namespace__()
+
+            row_index_expr = (
+                plx.col(order_by_[0])
+                .rank(method="ordinal", descending=False)
+                .over(partition_by=[], order_by=order_by_)
+                - 1
+            ).alias(name)
```
dangotbanned: Apologies if this seems a bit nit :face_with_peeking_eye:

I'm not sure why, but when there's a binary op I find it easier to read like this:

```suggestion
            rank = plx.col(order_by_[0]).rank("ordinal", descending=False)
            row_index_expr = (rank.over([], order_by=order_by_) - 1).alias(name)
```

narwhals/_compliant/dataframe.py
```diff
@@ -365,7 +364,9 @@ def unpivot(
         value_name: str,
     ) -> Self: ...
     def with_columns(self, *exprs: CompliantExprT_contra) -> Self: ...
-    def with_row_index(self, name: str) -> Self: ...
+
+    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self: ...
+
     def _evaluate_expr(self, expr: CompliantExprT_contra, /) -> Any:
```
dangotbanned:
```suggestion
    def with_columns(self, *exprs: CompliantExprT_contra) -> Self: ...
    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self: ...
    def _evaluate_expr(self, expr: CompliantExprT_contra, /) -> Any:
```

narwhals/_ibis/dataframe.py
```diff
@@ -413,10 +413,22 @@ def unpivot(
...
+            msg = (
+                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
+                "order-dependent operation."
+            )
+            raise ValueError(msg)
+        to_select = [
+            ibis.row_number().over(ibis.window(order_by=order_by)).name(name),
+            ibis.selectors.all(),
+        ]
+        return self._with_native(self.native.select(*to_select))
```
dangotbanned, line 421: Really like how nice and simple this is 🎉

narwhals/_pandas_like/dataframe.py
```diff
@@ -416,23 +416,37 @@ def estimated_size(self, unit: SizeUnit) -> int | float:
...
+            plx = self.__narwhals_namespace__()
+
+            row_index_expr = (
+                plx.col(order_by[0])
+                .rank(method="ordinal", descending=False)
+                .over(partition_by=[], order_by=order_by)
+                - 1
+            ).alias(name)
+            result = self.select(row_index_expr, plx.col(*self.columns)).native
+            return self._with_native(result)
```
dangotbanned: I know this suggestion isn't typing correctly, but I think we should change that - rather than using `_concat_horizontal`.


Also the same thing as (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160359256) could apply here - but I've left that as-is:


```suggestion
    def with_row_index(
        self, name: str, order_by: str | Sequence[str] | None
    ) -> PandasLikeDataFrame:
        plx = self.__narwhals_namespace__()
        if order_by is None:
            size = len(self)
            if self._implementation.is_cudf():
                import cupy as cp  # ignore-banned-import  # cuDF dependency.

                data = cp.arange(size)
            else:
                import numpy as np  # ignore-banned-import

                data = np.arange(size)

            row_index = plx._series.from_iterable(
                data, context=self, index=self.native.index, name=name
            )
            # TODO @dangotbanned: Add typing support for this in *at-least* `pandas`
            return plx.concat([row_index, self], how="horizontal")  # type: ignore[list-item]
        else:
            row_index_expr = (
                plx.col(order_by[0])
                .rank(method="ordinal", descending=False)
                .over(partition_by=[], order_by=order_by)
                - 1
            ).alias(name)
            return self.select(row_index_expr, plx.col(*self.columns))
```

narwhals/_polars/dataframe.py
```diff
@@ -491,11 +506,6 @@ def group_by(
 
         return PolarsGroupBy(self, keys, drop_null_keys=drop_null_keys)
 
-    def with_row_index(self, name: str) -> Self:
-        if self._backend_version < (0, 20, 4):
-            return self._with_native(self.native.with_row_count(name))
-        return self._with_native(self.native.with_row_index(name))
-
     def drop(self, columns: Sequence[str], *, strict: bool) -> Self:
```
dangotbanned, line 510: Great to see these two reduced to one path 🥳

dangotbanned (changes requested):
Thanks @FBruzzesi, just left a few suggestions on bits that weren't covered by (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160323551), (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160335652)

I think this is pretty close to ready, happy to approve once there's consensus on the `order_by=None` stuff

narwhals/_arrow/expr.py
FBruzzesi: > it's not meant to be None when we get here, hence my preference for the debug assertion

I am from mobile and not sure if today or tomorrow I can manage to do anything on this PR - anyway, I was actually encountering this AssertionError when `partition_by` was None and/or some other issue.

Will try to get you a repro, but it should be "easy" from the current main

narwhals/_compliant/dataframe.py
FBruzzesi: I added it when I was doing all the computation at the complaint protocol(s) level. Now it's a nice to have but unrelated to this PR

narwhals/_spark_like/dataframe.py
FBruzzesi: > i wouldn't bother, the idea with v1 is "code you've written today will keep running in the future", if there's something which it's not possible to have already written then it's not our concern

I think having a good and consistent error message would not hurt to be honest, as it also suggests how to fix it for all lazy backends

> From a typing perspective, the None is an issue - since it's defined in CompliantLazyFrame

Yes I was also sensing either some blood bath or some ignore(s) to be required

narwhals/dataframe.py
FBruzzesi: My understanding for a sentinel like that is that it's useful when None can be a valid parameter. Which is not the case here 🙈

narwhals/_pandas_like/dataframe.py
FBruzzesi: Hey @dangotbanned , this diff is not rendering as expected I think. Is the main point using `plx.concat(..., how="horizontal") instead of the private one?

narwhals/_arrow/expr.py
FBruzzesi: Never mind, @dangotbanned has the repro above! (I finally opened a laptop since the start of the weekend)

narwhals/_pandas_like/dataframe.py
dangotbanned: Ahh weird, does this work?

There were a few tweaks but yeah `plx.concat` is probably the most important

```py
    def with_row_index(
        self, name: str, order_by: str | Sequence[str] | None
    ) -> PandasLikeDataFrame:
        if order_by is None:
            size = len(self)
            if self._implementation.is_cudf():
                import cupy as cp  # ignore-banned-import  # cuDF dependency.

                data = cp.arange(size)
            else:

                data = np.arange(size)

            row_index = plx._series.from_iterable(
                data, context=self, index=self.native.index, name=name
            )
            # TODO @dangotbanned: Add typing support for this in *at-least* `pandas`
            return plx.concat([row_index, self], how="horizontal")  # type: ignore[list-item]
        else:
            row_index_expr = (
                plx.col(order_by[0])
                .rank(method="ordinal", descending=False)
                .over(partition_by=[], order_by=order_by)
                - 1
            ).alias(name)
            return self.select(row_index_expr, plx.col(*self.columns))
```

FBruzzesi: Thanks a ton! This is much simpler to read!

For context: I see this 🤯

image

narwhals/dataframe.py
dangotbanned: > My understanding for a sentinel like that is that it's useful when None can be a valid parameter. Which is not the case here 🙈

Yeah good point, forget I brought sentinels into the mix 😄 

`()` would be my preference - but ultimately we're just balancing trade-offs of backcompat vs a new default that raises

narwhals/_dask/dataframe.py
```diff
@@ -204,12 +205,28 @@ def drop(self, columns: Sequence[str], *, strict: bool) -> Self:
...
-    def with_row_index(self, name: str) -> Self:
+    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
         # Implementation is based on the following StackOverflow reply:
         # https://stackoverflow.com/questions/60831518/in-dask-how-does-one-add-a-range-of-integersauto-increment-to-a-new-column/60852409#60852409
-        return self._with_native(
-            add_row_index(self.native, name, self._backend_version, self._implementation)
-        )
+        if order_by is None:
+            return self._with_native(
+                add_row_index(
```
dangotbanned, line 213: I've just been playing around w/ `dask` while looking at (#2722)

Maybe we could switch to this instead of `add_row_index`?

```py
import dask
import dask.array as da
import dask.dataframe as dd

data = {"a": [4, 2, 6, 7, 3, 2, 6, 7, 3, 3]}
df: dd.DataFrame = dd.from_dict(data, npartitions=1)
name = "index"
with_row_index = df.assign(**{name:da.arange(df.size.compute())})

>>> with_row_index.compute()
   a  index
0  4      0
1  2      1
2  6      2
3  7      3
4  3      4
5  2      5
6  6      6
7  7      7
8  3      8
9  3      9
```

### Edit

Actually, `df.size.compute()` could be arbitrarily expensive.
Looks nice but may be a footgun

narwhals/_pandas_like/dataframe.py
dangotbanned: Ahh (https://github.com/narwhals-dev/narwhals/actions/runs/15809029085/job/44557840612?pr=2683)

```py
 narwhals/_pandas_like/dataframe.py:436: error: Incompatible return value type
(got "PandasLikeDataFrame", expected "Self")  [return-value]
                return plx.concat([row_index, self], how="horizontal")  # ...
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
narwhals/_pandas_like/dataframe.py:436: note: Error code "return-value" not covered by "type: ignore" comment
Found 1 error in 1 file (checked 386 source files)
```

The suggestion I gave (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160408679) was after I also ran into `mypy` telling me off 😉

narwhals/dataframe.py
```diff
@@ -2431,29 +2437,56 @@ def drop_nulls(self, subset: str | list[str] | None = None) -> Self:
...
+            |  b: [[5,4]]      |
             └──────────────────┘
         """
-        return super().with_row_index(name)
+        if order_by is None:
+            msg = (
+                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
+                "order-dependent operation."
+            )
+            raise ValueError(msg)
```
MarcoGorelli: If it's not allowed to be `None` and we raise immediately, shall we just remove it from the signature altogether?

Something like
```diff
diff --git a/narwhals/dataframe.py b/narwhals/dataframe.py
index 43d864b1e..3576f84b8 100644
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -138,12 +138,6 @@ class BaseFrame(Generic[_FrameT]):
     ) -> R:
         return function(self, *args, **kwargs)
 
-    def with_row_index(self, name: str, order_by: str | Sequence[str] | None) -> Self:
-        order_by_ = [order_by] if isinstance(order_by, str) else order_by
-        return self._with_compliant(
-            self._compliant_frame.with_row_index(name, order_by=order_by_)
-        )
-
     def drop_nulls(self, subset: str | list[str] | None) -> Self:
         subset = [subset] if isinstance(subset, str) else subset
         return self._with_compliant(self._compliant_frame.drop_nulls(subset=subset))
@@ -1096,7 +1090,10 @@ class DataFrame(BaseFrame[DataFrameT]):
             a: [[1,2]]
             b: [[4,5]]
         """
-        return super().with_row_index(name=name, order_by=order_by)
+        order_by_ = [order_by] if isinstance(order_by, str) else order_by
+        return self._with_compliant(
+            self._compliant_frame.with_row_index(name, order_by=order_by_)
+        )
 
     @property
     def schema(self) -> Schema:
@@ -2480,7 +2477,10 @@ class LazyFrame(BaseFrame[FrameT]):
             |  b: [[5,4]]      |
             └──────────────────┘
         """
-        return super().with_row_index(name=name, order_by=order_by)
+        order_by_ = [order_by] if isinstance(order_by, str) else order_by
+        return self._with_compliant(
+            self._compliant_frame.with_row_index(name, order_by=order_by_)
+        )
 
     @property
     def schema(self) -> Schema:
```
?

(splitting up the implementations for `nw.DataFrame` and `nw.LazyFrame` is to allow for the signatures to be slightly different, which is fine IMO)

narwhals/_arrow/expr.py
MarcoGorelli: Ah, I see! Thanks for explaining - happy with what you've written then, thanks 🙏

narwhals/_spark_like/dataframe.py
MarcoGorelli: I just tried
```diff
diff --git a/narwhals/_compliant/dataframe.py b/narwhals/_compliant/dataframe.py
index 471ec6a59..1e2537de7 100644
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -253,6 +253,7 @@ class CompliantDataFrame(
         value_name: str,
     ) -> Self: ...
     def with_columns(self, *exprs: CompliantExprT_contra) -> Self: ...
+    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self: ...
     @overload
     def write_csv(self, file: None) -> str: ...
     @overload
@@ -364,7 +365,7 @@ class CompliantLazyFrame(
         value_name: str,
     ) -> Self: ...
     def with_columns(self, *exprs: CompliantExprT_contra) -> Self: ...
-    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self: ...
+    def with_row_index(self, name: str, order_by: Sequence[str]) -> Self: ...
     def _evaluate_expr(self, expr: CompliantExprT_contra, /) -> Any:
         result = expr(self)
         assert len(result) == 1  # debug assertion  # noqa: S101
diff --git a/narwhals/_duckdb/dataframe.py b/narwhals/_duckdb/dataframe.py
index 77db33322..4267e32c8 100644
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -507,13 +507,7 @@ class DuckDBLazyFrame(
         )
 
     @requires.backend_version((1, 3))
-    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
-        if order_by is None:
-            msg = (
-                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
-                "order-dependent operation."
-            )
-            raise ValueError(msg)
+    def with_row_index(self, name: str, order_by: Sequence[str]) -> Self:
         order_by_sql = generate_order_by_sql(*order_by, ascending=True)
         row_index_expr = SQLExpression(
             f"{FunctionExpression('row_number')} over ({order_by_sql}) - 1"
diff --git a/narwhals/_ibis/dataframe.py b/narwhals/_ibis/dataframe.py
index 5907d427e..0c96b08a4 100644
--- a/narwhals/_ibis/dataframe.py
+++ b/narwhals/_ibis/dataframe.py
@@ -413,13 +413,7 @@ class IbisLazyFrame(
         )
         return self._with_native(unpivoted.select(*final_columns))
 
-    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
-        if order_by is None:
-            msg = (
-                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
-                "order-dependent operation."
-            )
-            raise ValueError(msg)
+    def with_row_index(self, name: str, order_by: Sequence[str]) -> Self:
         to_select = [
             ibis.row_number().over(ibis.window(order_by=order_by)).name(name),
             ibis.selectors.all(),
diff --git a/narwhals/_spark_like/dataframe.py b/narwhals/_spark_like/dataframe.py
index 7f502f7c8..010b6fd79 100644
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -522,13 +522,7 @@ class SparkLikeLazyFrame(
             unpivoted_native_frame = unpivoted_native_frame.drop(*ids)
         return self._with_native(unpivoted_native_frame)
 
-    def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
-        if order_by is None:
-            msg = (
-                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
-                "order-dependent operation."
-            )
-            raise ValueError(msg)
+    def with_row_index(self, name: str, order_by: Sequence[str]) -> Self:
         row_index_expr = (
             self._F.row_number().over(
                 self._Window.partitionBy(self._F.lit(1)).orderBy(*order_by)
diff --git a/narwhals/dataframe.py b/narwhals/dataframe.py
index 21d596a95..3576f84b8 100644
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -138,12 +138,6 @@ class BaseFrame(Generic[_FrameT]):
     ) -> R:
         return function(self, *args, **kwargs)
 
-    def with_row_index(self, name: str, order_by: str | Sequence[str] | None) -> Self:
-        order_by_ = [order_by] if isinstance(order_by, str) else order_by
-        return self._with_compliant(
-            self._compliant_frame.with_row_index(name, order_by=order_by_)
-        )
-
     def drop_nulls(self, subset: str | list[str] | None) -> Self:
         subset = [subset] if isinstance(subset, str) else subset
         return self._with_compliant(self._compliant_frame.drop_nulls(subset=subset))
@@ -1096,7 +1090,10 @@ class DataFrame(BaseFrame[DataFrameT]):
             a: [[1,2]]
             b: [[4,5]]
         """
-        return super().with_row_index(name=name, order_by=order_by)
+        order_by_ = [order_by] if isinstance(order_by, str) else order_by
+        return self._with_compliant(
+            self._compliant_frame.with_row_index(name, order_by=order_by_)
+        )
 
     @property
     def schema(self) -> Schema:
@@ -2438,7 +2435,7 @@ class LazyFrame(BaseFrame[FrameT]):
         return super().drop_nulls(subset=subset)
 
     def with_row_index(
-        self, name: str = "index", order_by: str | Sequence[str] | None = None
+        self, name: str = "index", *, order_by: str | Sequence[str]
     ) -> Self:
         """Insert column which enumerates rows.
 
@@ -2480,13 +2477,10 @@ class LazyFrame(BaseFrame[FrameT]):
             |  b: [[5,4]]      |
             ÔööÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÇÔöÿ
         """
-        if order_by is None:
-            msg = (
-                "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
-                "order-dependent operation."
-            )
-            raise ValueError(msg)
-        return super().with_row_index(name=name, order_by=order_by)
+        order_by_ = [order_by] if isinstance(order_by, str) else order_by
+        return self._with_compliant(
+            self._compliant_frame.with_row_index(name, order_by=order_by_)
+        )
 
     @property
     def schema(self) -> Schema:
diff --git a/narwhals/stable/v1/__init__.py b/narwhals/stable/v1/__init__.py
index cf6f8a076..626baf702 100644
--- a/narwhals/stable/v1/__init__.py
+++ b/narwhals/stable/v1/__init__.py
@@ -271,7 +271,7 @@ class LazyFrame(NwLazyFrame[IntoFrameT]):
         """
         order_by_ = [order_by] if isinstance(order_by, str) else order_by
         return self._with_compliant(
-            self._compliant_frame.with_row_index(name=name, order_by=order_by_)
+            self._compliant_frame.with_row_index(name=name, order_by=order_by_)  # type: ignore[arg-type]
         )
```
and it passes pyright and mypy, and only requires a single `type: ignore` (in `stable.v1`, but we can add a little comment to explain why)

FBruzzesi: @MarcoGorelli I am still not fully sure about removing the 

```py
if order_by is None:
    msg = (
        "`LazyFrame.with_row_index` requires `order_by` to be specified as it is an "
        "order-dependent operation."
    )
    raise ValueError(msg)
```

I will push a commit doing that so that's easier to discuss about it

FBruzzesi: Here you go: https://github.com/narwhals-dev/narwhals/pull/2683/commits/ce3fc57b2d54dad00e5de0178e3bcd3f03126be8

narwhals/dataframe.py
```diff
@@ -2438,13 +2435,13 @@ def drop_nulls(self, subset: str | list[str] | None = None) -> Self:
         return super().drop_nulls(subset=subset)
 
     def with_row_index(
-        self, name: str = "index", order_by: str | Sequence[str] | None = None
+        self, name: str = "index", *, order_by: str | Sequence[str]
```
dangotbanned: Well I guess I learned something new today 😄 

Never seen a default before a non-default - quite surprised that the `*` avoids the syntax error!

narwhals/_pandas_like/dataframe.py
```diff
@@ -431,7 +429,7 @@ def with_row_index(
                 data, context=self, index=self.native.index, name=name
             )
             # TODO @dangotbanned: Add typing support for this in *at-least* `pandas`
-            return plx.concat([row_index, self], how="horizontal")  # type: ignore[list-item]
+            return plx.concat([row_index, self], how="horizontal")  # type: ignore[return-value, list-item]
```
dangotbanned: I *think* this one needs to stay as `PandasLikeDataFrame`

Because `EagerNamespace.concat` is using a `TypeVar` that is resolved to `PandasLikeDataFrame`

(Regarding the new `return-value` ignore)

dangotbanned: I forgot to mention - I did something different in `ArrowDataFrame`, which lets you use `select` instead of `concat` - maybe that avoids both of these ignores?

FBruzzesi: oh that's a nice trick. Pushing it now

tests/frame/with_row_index_test.py
```diff
@@ -1,20 +1,66 @@
...
+        .with_row_index(name="foo bar", order_by=order_by)
+        .sort("xyz")
+    )
+    expected = {"foo bar": expected_index, **data}
     assert_equal_data(result, expected)
+
+
+@pytest.mark.parametrize("namespace", [nw, nw_v1])
+def test_with_row_index_lazy_exception(
+    constructor: Constructor, namespace: ModuleType
```
MarcoGorelli: really minor, but i think it'd be better to split this into a test in `tests/v1_test.py` too, so that `tests/v1_test.py` shows and tests all the places where behaviour differs

MarcoGorelli:
nice one, thanks for sticking to it!

Issue #2666, opened 2025-06-11
typ: restore (if possible) `NativeSeriesT` to `EagerNamespace`


MarcoGorelli:
https://github.com/narwhals-dev/narwhals/pull/2662#issuecomment-2962008537 removed `NativeSeriesT`. Reason:
- the signature changed, so it no longer appeared as a return value, so type checkers no longer assumed it to be invariant
- type checkers raised saying that it must be covariant, and wouldn't allow `NativeSeriesT_contra`
- but, it was used as an argument type, so type checkers would also raise if we used `NativeSeriesT_co` instead

I removed this in order to enable some highish prio work to get through, sorry about that, though I'm open to keeping `NativeSeriesT` there if possible

PR #2693, opened 2025-06-17
typing: restore NativeSeriesT to EagerDataFrame and EagerNamespace, avoid unnecessary broadcast in eager when/then/otherwise

MarcoGorelli:
This addresses part of https://github.com/narwhals-dev/narwhals/issues/2666

## Related
- #2666
- [remove NativeSeriesT from EagerDataFrame / EagerNamespace ](https://github.com/narwhals-dev/narwhals/pull/2662/commits/e19d5e17ce089ecea4944eae6e2753832bb8c730)

## Description

`EagerNamespace` leaves me in covariance/contravariance hell, I'm trying to put together a smaller repro to isolate the issue i'm facing

OK, I've finally fully (I think) made sense of the issues with `EagerNamespace` and why I wasn't able to put `NativeSeriesT` there too

- If you define a protocol `P` generic in `T` and `T` doesn't appear in any of `P`'s method's arguments, then type checkers expect `P` to be covariant in `T`
- If you define a method `foo(self, a: T) -> None`, then type checkers expect `P` to be contravariant in `T`. But, if you instead define it to be `foo(self, a: T | Any) -> None`, then that turns off the type checking for `a` and so type checkers expect `P` to be covariant in `T`. 
- If you then overload `foo` and one of the overloads matches just `T`, then you'll run into a bad place, as type checkers simultaneously expect `P` to be contravariant in `T` and also covariant

In this case, `EagerNamespace.from_native` accepts `NativeSeriesT | Any`, and it's the `| Any` that's the really problematic part. It might as well just be `Any`, at which point we don't need `EagerNamespace` to be generic in `NativeSeriesT` if it's not used anywhere

Having said that, there might also be a bug in MyPy? https://gist.github.com/mypy-play/021d9f12a5e02e1e837767f0e6fd0e4a

I think possible solutions on that side are:
- keep `EagerNamespace` non-generic in `NativeSeriesT` (i.e. status-quo)
- make `EagerNamespace` generic in `NativeSeriesT_contra`, and then `type: ignore` a couple of lines in `_from_native_impl` when `ModinSeries` can't be assigned to `NativeSeries`


dangotbanned:
Thanks @MarcoGorelli - love to see all this detail in the description! 😍

Will do my best to review tomorrow, just wanted to say I really appreciate this

dangotbanned:
I haven't tried these out, but two ideas I had

## Modin
Change these defs

https://github.com/narwhals-dev/narwhals/blob/f7d94519c73d05993d15195ce1cdb671262b8c86/narwhals/_namespace.py#L113-L117

To this, which *should* make them compatible with the `Native*` protocols

```py
    class _ModinDataFrame(NativeFrame, Protocol):
        _pandas_class: type[pd.DataFrame]

    class _ModinSeries(NativeSeries, Protocol):
        _pandas_class: type[pd.Series[Any]]
```

## Variance
For the `TypeVar`(s) or `Protocol`(s) that are problematic, trying defining the `TypeVar`(s) using [`infer_variance`](https://docs.python.org/3/library/typing.html#typing.TypeVar)

```py
from typing import Protocol
from narwhals._typing_compat import TypeVar

class CompliantWhat(Protocol):
    def hello(self) -> str: ...

CompliantWhatT = TypeVar("CompliantWhatT", bound=CompliantWhat, infer_variance=True)
```

MarcoGorelli:
thanks - i tried the first one but it didn't resolve the issue

dangotbanned:
> thanks - i tried the first one but it didn't resolve the issue

No worries!

I'm just moving this to draft while I'm experimenting (https://github.com/narwhals-dev/narwhals/pull/2693#event-18210901998)


I really should've followed up this **TODO** I left (https://github.com/narwhals-dev/narwhals/pull/2662/commits/e19d5e17ce089ecea4944eae6e2753832bb8c730#r2139770243)
A large chunk of the issue is that

narwhals/_compliant/when_then.py
```diff
@@ -148,8 +149,15 @@ def from_when(
...
     CompliantWhen[EagerDataFrameT, EagerSeriesT, EagerExprT],
-    Protocol38[EagerDataFrameT, EagerSeriesT, EagerExprT],
+    Protocol38[EagerDataFrameT, EagerSeriesT, EagerExprT, NativeSeriesT],
 ):
+    def _temp_invariant(self, _: NativeSeriesT, /) -> NativeSeriesT:
+        """**DO NOT MERGE**.
+
+        Using as a placeholder until there's real usage of `NativeSeriesT`.
+        """
+        return _
```
dangotbanned: What I'm planning to do is de-deduplicate the broadcasting semantics that we have in these two,  that was introduced in (#2662)

https://github.com/narwhals-dev/narwhals/blob/f7d94519c73d05993d15195ce1cdb671262b8c86/narwhals/_pandas_like/namespace.py#L321-L327

https://github.com/narwhals-dev/narwhals/blob/f7d94519c73d05993d15195ce1cdb671262b8c86/narwhals/_arrow/namespace.py#L274-L282


AFAICT, all we needed to do in (#2662) was replace the use of `_extract_comparand` - with some generic version of `align_series_full_broadcast`

https://github.com/narwhals-dev/narwhals/blob/f7d94519c73d05993d15195ce1cdb671262b8c86/narwhals/_compliant/dataframe.py#L432-L434

Before we had something like this:

```py
        return pc.if_else(when, then, otherwise)
```

So the step can just happen in the implemented `EagerWhen.__call__` and depend on the **uninplemented** `Eager*._align_series_full_broadcast`

dangotbanned: I'm leaving the coverage to fail, since I **really** don't want to forget about not merging that method 😅 

https://github.com/narwhals-dev/narwhals/actions/runs/15735541100/job/44347151400?pr=2693

Hopefully will be able to dip back in today

dangotbanned: Alright I'm pretty much done, just missed an import somewhere that's causing https://github.com/narwhals-dev/narwhals/actions/runs/15738449057/job/44357345930?pr=2693

Probably just something slipped out of `TYPE_CHECKING`

narwhals/_arrow/series.py
```diff
@@ -4,6 +4,7 @@
 
 import pyarrow as pa
 import pyarrow.compute as pc
+from typing_extensions import Self
```
dangotbanned: That'll be it 🫣

https://github.com/narwhals-dev/narwhals/pull/2693#discussion_r2155071081

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2693/commits/ed96beef340c380f1fc1b57a90934809c98112da)

narwhals/_compliant/when_then.py
dangotbanned: Went down a bit of a rabbit hole, but hopefully we've now got:

- More functional typing
- Better docs on broadcasting
- More shared code
- *Hopefully* more performant broadcasting (by using `(np|pa).repeat`)

narwhals/_compliant/when_then.py
```diff
@@ -148,31 +149,39 @@ def from_when(
...
         is_expr = self._condition._is_expr
         when: EagerSeriesT = self._condition(df)[0]
         then: EagerSeriesT
+        align = when._align_full_broadcast
 
         if is_expr(self._then_value):
             then = self._then_value(df)[0]
         else:
             then = when.alias("literal")._from_scalar(self._then_value)
-            then._broadcast = True
```
MarcoGorelli: can we revert this please? i don't like how you've moved `._broadcast = True` into `_from_scalar`, are we sure the logic is correct in other places where `_from_scalar` is used, like in `_compliant/expr.py`?

narwhals/_compliant/series.py
```diff
@@ -280,8 +280,28 @@ class EagerSeries(CompliantSeries[NativeSeriesT], Protocol[NativeSeriesT]):
...
+            nw.col("a") - nw.col("a").mean()
+
+        because then it's more efficient to extract the right-hand-side's single element as a scalar.
+        """
+        ...
+
     def _from_scalar(self, value: Any) -> Self:
-        return self.from_iterable([value], name=self.name, context=self)
+        ser = self.from_iterable([value], name=self.name, context=self)
+        ser._broadcast = True
```
MarcoGorelli: same

narwhals/_compliant/when_then.py
dangotbanned: Is there a situation where we'd have a scalar that shouldn't be broadcast?

I moved it there because every use was doing the same dance of calling that method and then setting an attribute.
It seems like an easy thing to miss when there's post-constructor logic that the caller is responsible for

To be honest, I think that line should really be:

```py
EagerSeries.from_scalar(self._then_value)
```

Where we do these internally:
- `name = "literal"`
  - rather than "aliasing" `when` for a value being `then`
- `_broadcast = True`

MarcoGorelli: > Is there a situation where we'd have a scalar that shouldn't be broadcast?

I think currently in `_reuse_series_inner` we call `_from_scalar` with then following it without setting `broadcast=True`?

> Is there a situation where we'd have a scalar that shouldn't be broadcast?

I think so, yes. But I'd have to think carefully about this. Let's leave it for a separate discussion and only resolve the typing issues in this pr

dangotbanned: No worries 🙂

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2693/commits/26bcd19c80c01cfb862a9abf59561d7073ebf39c)

narwhals/_compliant/series.py
dangotbanned: Ditto https://github.com/narwhals-dev/narwhals/pull/2693/commits/26bcd19c80c01cfb862a9abf59561d7073ebf39c

MarcoGorelli:
thanks @dangotbanned , awesome additions and rewrites!

this helped notice that there was a further chance to improve by avoiding an unnecessary broadcast if `otherwise_value` is a scalar

narwhals/_compliant/when_then.py
```diff
@@ -175,14 +175,9 @@ def __call__(self, df: EagerDataFrameT, /) -> Sequence[EagerSeriesT]:
...
             result = self._if_then_else(when.native, then.native, otherwise.native)
-        elif self._otherwise_value is not None:
-            otherwise = when._from_scalar(self._otherwise_value)
-            otherwise._broadcast = True
-            when, then, otherwise = align(when, then, otherwise)
-            result = self._if_then_else(when.native, then.native, otherwise.native)
         else:
             when, then = align(when, then)
-            result = self._if_then_else(when.native, then.native, None)
+            result = self._if_then_else(when.native, then.native, self._otherwise_value)
```
dangotbanned, line 180: Yes please 😍

dangotbanned:
The PR probably needs a rename, but other than that 🚢 🚢 🚢

narwhals/_arrow/dataframe.py
```diff
@@ -330,7 +332,8 @@ def select(self: ArrowDataFrame, *exprs: ArrowExpr) -> ArrowDataFrame:
                 self.native.__class__.from_arrays([]), validate_column_names=False
             )
         names = [s.name for s in new_series]
-        reshaped = align_series_full_broadcast(*new_series)
+        align = new_series[0]._align_full_broadcast
```
dangotbanned, line 335: I did this kind of aliasing on *almost* every use.
Haven't come up with a nice short name, that still conveys the message well

In this PR we make this change, which I like because the `series` part is embedded in the type you call it on

```py
utils.align_series_full_broadcast
EagerSeries._align_full_broadcast
```

I dunno maybe something like one of these could work:

```py
EagerSeries.align_all
EagerSeries().align_with  # <-- instance method
EagerSeries.broadcast
```

I wonder if the `rust` impl of `polars` has any similar utils with a name we can steal? 😄

MarcoGorelli:
thanks Dan!

Issue #2690, opened 2025-06-17
tests: update tests for cudf 25.08.00a


MarcoGorelli:
There's 1 cudf failure, and 2 cudf-polars ones, nothing major https://www.kaggle.com/code/marcogorelli/narwhals-pre-release-test-cudf/notebook?scriptVersionId=245939985

EdAbati:
I can have a look at this

EdAbati:
~it may be a silly question but I don't see the 2 cudf-polars errors.~

~I see 33 from `NARWHALS_DEFAULT_CONSTRUCTORS=pandas python -m cudf.pandas -m pytest`
and 1 from `pytest --constructors=cudf`~

~which ones are you referring to?~

Ignore me I was looking at the wrong version of hte notebook

MarcoGorelli:
not at all a silly question, i didn't provide much detail 😄 

if you look at https://www.kaggle.com/code/marcogorelli/narwhals-pre-release-test-cudf/notebook?scriptVersionId=245939985 you see that `!cd narwhals && pytest --constructors=cudf` has one test failure, and `!cd narwhals && NARWHALS_POLARS_GPU=1 pytest --constructors=polars[lazy]` has two

FBruzzesi:
Should this be closed by #2700?

MarcoGorelli:
thanks!

PR #2700, opened 2025-06-19
fix: update tests for `cuDF` `25.08.00a`

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2690

## Checklist

- [x] Tests added
- [x] Documented the changes

TODOs:

- [ ] fix `tests/expr_and_series/lit_test.py::test_date_lit[cudf] - NotImplementedError: cuDF does not yet support date32[day][pyarrow]`

### DIFF
--- a/narwhals/_polars/utils.py
+++ b/narwhals/_polars/utils.py
@@ -217,6 +217,14 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
     return pl.Unknown()  # pragma: no cover
 
 
+def _is_polars_exception(exception: Exception, backend_version: tuple[int, ...]) -> bool:
+    if backend_version >= (1,):
+        # Old versions of Polars didn't have PolarsError.
+        return isinstance(exception, pl.exceptions.PolarsError)
+    # Last attempt, for old Polars versions.
+    return "polars.exceptions" in str(type(exception))  # pragma: no cover
+
+
 def catch_polars_exception(
     exception: Exception, backend_version: tuple[int, ...]
 ) -> NarwhalsError | Exception:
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
thanks! i presume `test_date_lit` can just be xfailed?

EdAbati:
> thanks! i presume test_date_lit can just be xfailed?

Yeah I am a bit torn. It could be xfailed, but it would a bit awkard to say that `nw.Date`only works (i.e. doesn't raise an Exception) with older `cudf` versions right?

Before cudf (<25.08.00) was doing this conversion:

```python
df = nw.from_native(cudf.DataFrame({"a": [1]}))
result = df.with_columns(nw.lit(date(2020, 1, 1), dtype=nw.Date))
result.to_native().info()
# <class 'cudf.core.dataframe.DataFrame'>
# RangeIndex: 1 entries, 0 to 0
# Data columns (total 2 columns):
#  #   Column   Non-Null Count  Dtype
# ---  ------   --------------  -----
#  0   a        1 non-null      int64
#  1   literal  1 non-null      datetime64[ms]
# dtypes: datetime64[ms](1), int64(1)
# memory usage: 16.0 bytes
```

What do you think if we fix it with something like this?

```python
def narwhals_to_native_dtype(...):
    ...
    if isinstance_or_issubclass(dtype, dtypes.Date):
            return "datetime64[ms]"
        try:
            import pyarrow as pa  # ignore-banned-import
        except ModuleNotFoundError:  # pragma: no cover
            msg = "'pyarrow>=11.0.0' is required for `Date` dtype."
        return "date32[pyarrow]"
 ```
 
 (I'm not an expert of time/date types I may not know edge cases)

MarcoGorelli:
> it would a bit awkard to say that nw.Dateonly works (i.e. doesn't raise an Exception) with older cudf versions right?

sure, but if it's a decision cudf made, then i'd say that the awkwardness is on them 😄

EdAbati:
Well that makes my life easier :D

MarcoGorelli (approved):
thanks @EdAbati !

PR #2709, opened 2025-06-20
fix: Set Ibis time_unit and time_zone for Datetime dtype conversion

rwhitten577:
Casting to/from Ibis Datetime dtype did not pass the `time_unit` or `time_zone` args. 

Seems like this isn't supported by most other lazy backends either, so haven't modified the tests to explicitly check for the unit & timezone conversion. 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_ibis/utils.py
+++ b/narwhals/_ibis/utils.py
@@ -112,7 +112,9 @@ def native_to_narwhals_dtype(ibis_dtype: IbisDataType, version: Version) -> DTyp
     if ibis_dtype.is_date():
         return dtypes.Date()
     if ibis_dtype.is_timestamp():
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
thanks @rwhitten577 ! time zone things are notoriously annoying to test, but this is simple enough and aligned enough with what we do for other backends that i'm happy to ship it

PR #2710, opened 2025-06-20
fix: raise NotImplementedError instead of returning it

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/utils.py
+++ b/narwhals/_arrow/utils.py
@@ -237,14 +237,10 @@ def extract_native(
     If one of the two sides has a `_broadcast` flag, then extract the scalar
     underneath it so that PyArrow can do its own broadcasting.
     """
-    from narwhals._arrow.dataframe import ArrowDataFrame
     from narwhals._arrow.series import ArrowSeries
 
     if rhs is None:  # pragma: no cover
         return lhs.native, lit(None, type=lhs._type)
-
-    if isinstance(rhs, ArrowDataFrame):
-        return NotImplemented
 
     if isinstance(rhs, ArrowSeries):
         if lhs._broadcast and not rhs._broadcast:
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -1770,7 +1770,8 @@ class not_implemented:  # noqa: N801
         # NOTE: Prefer not exposing the actual class we're defining in
         # `_implementation` may not be available everywhere
         who = getattr(instance, "_implementation", self._name_owner)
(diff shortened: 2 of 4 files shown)


narwhals/_arrow/utils.py
```diff
@@ -237,15 +237,11 @@ def extract_native(
...
     underneath it so that PyArrow can do its own broadcasting.
     """
-    from narwhals._arrow.dataframe import ArrowDataFrame
     from narwhals._arrow.series import ArrowSeries
 
     if rhs is None:  # pragma: no cover
         return lhs.native, lit(None, type=lhs._type)
 
-    if isinstance(rhs, ArrowDataFrame):
-        return NotImplemented
```
MarcoGorelli, line 247: this doesn't seem to affect any test

it would only be used if we supported binary operations between series and dataframes (e.g. `s - df`), which we don't anyway (and I don't think we should)

narwhals/_utils.py
```diff
@@ -1793,13 +1794,13 @@ def deprecated(cls, message: LiteralString, /) -> Self:
...
 
-def _not_implemented_error(what: str, who: str, /) -> NotImplementedError:
+def _raise_not_implemented_error(what: str, who: str, /) -> NotImplementedError:
     msg = (
         f"{what!r} is not implemented for: {who!r}.\n\n"
         "If you would like to see this functionality in `narwhals`, "
         "please open an issue at: https://github.com/narwhals-dev/narwhals/issues"
     )
-    return NotImplementedError(msg)
+    raise NotImplementedError(msg)
```
MarcoGorelli, line 1803: this one doesn't strictly need changing, but doing so means we can enforce a pre-commit check for `return NotImplemented` globally and avoid this common-ish error

dangotbanned: This isn't like the others

When you `return` an exception, and `raise` it elsewhere - the `msg` assignment doesn't show up in the traceback

I haven't checked out the branch, but the traceback this produces will have the message repeated twice

MarcoGorelli: thanks - just checked and it only shows up once, the exception is only being raised here (and not returned and re-raised)

PR #2716, opened 2025-06-21
chore: use _with_elementwise in IbisNamespace

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_ibis/namespace.py
+++ b/narwhals/_ibis/namespace.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import operator
 from functools import reduce
 from itertools import chain
-from typing import TYPE_CHECKING, Any, cast
+from typing import TYPE_CHECKING, Any, Callable, cast
 
 import ibis
 import ibis.expr.types as ir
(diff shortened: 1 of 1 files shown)


narwhals/_spark_like/namespace.py
```diff
@@ -73,7 +73,7 @@ def _native_dtypes(self):  # type: ignore[no-untyped-def] # noqa: ANN202
         else:
             return import_native_dtypes(self._implementation)
 
-    def _with_elementwise(
+    def _expr_from_callable(
```
MarcoGorelli, line 76: i'd misnamed this one, as what's returned isn't a namespace but an expr

narwhals/_duckdb/namespace.py
```diff
@@ -49,7 +49,7 @@ def _expr(self) -> type[DuckDBExpr]:
     def _lazyframe(self) -> type[DuckDBLazyFrame]:
         return DuckDBLazyFrame
 
-    def _with_elementwise(
+    def _expr_from_elementwise(
```
dangotbanned, line 52: Could this be a `@classmethod` on `LazyExpr` (or the children directly)?

So from `*Namespace` we call:

```py
self._expr.from_elementwise(...)
```

That seems a bit more consistent *to me*, with what we have elsewhere

https://github.com/narwhals-dev/narwhals/blob/a9879b4101915bb1e29018ba8ecd8af4383c2508/narwhals/_compliant/namespace.py#L49-L68

MarcoGorelli: thanks for the suggestion

we'd then get some circular imports which we'd need to resolve

can consider it as a follow-up though

dangotbanned: A follow up is fine, but what do you mean by 

> we'd then get some circular imports which we'd need to resolve

The `DuckDBNamespace._expr` is called inside `DuckDBNamespace._expr_from_elementwise`?

https://github.com/MarcoGorelli/narwhals/blob/de40f3448a5bb4c8ee79cce4d124f881fe2363f1/narwhals/_duckdb/namespace.py#L67

Issue #2715, opened 2025-06-21
docs: `is_between` doesn't show `ClosedInterval` values. Remove type aliases


MarcoGorelli:
tbh i find these type aliases a bit frustrating. if I look at the `is_between` docs, i shouldn't need an extra click to `ClosedInterval` to know what values it can take

i'd prefer is the user-facing docstrings just showed `Literal["left", "right", "none", "both"]` and kept it simple

dangotbanned:
Is this an issue you're having only in the API Reference [`Expr.is_between`](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.is_between)? 

I don't have any objections to including the literals in the docstring (e.g. like [`DataFrame.join`](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.join)

But I feel like the IDE experience is much worse with the `Literal`(s) as annotations (see #2352).
Personally, I'd want to prioritize IDE vs static docs since I think the latter is used much less frequently

MarcoGorelli:
> But I feel like the IDE experience is much worse with the Literal(s) as annotations (see https://github.com/narwhals-dev/narwhals/pull/2352).

I've clicked but don't see why it's much worse could you elaborate please?

dangotbanned:
> > But I feel like the IDE experience is much worse with the Literal(s) as annotations (see [#2352](https://github.com/narwhals-dev/narwhals/pull/2352)).
> 
> I've clicked but don't see why it's much worse could you elaborate please?

Sure thing

## Hover

For the aliases that have a docstring, we lose that context that's available on hover.
(#2352) is displaying that *not only* in `DataFrame.join`, but also in all of the `Compliant*` classes for free

<details><summary>Screenshot 1</summary>
<p>

> ### IDE Hover
> 
> ![image](https://github.com/user-attachments/assets/a56ccf62-6e31-4131-9759-a2ba2285167d)


</p>
</details> 


## Typing

If we use a `Literal`, making this change doesn't produce any type checker diagnostics - despite now having a new unhandled case

<details><summary>Screenshot 2</summary>
<p>


![Image](https://github.com/user-attachments/assets/90e7b97e-1641-4865-b3a8-875665359474)

</p>
</details> 


I was expecting this to produce a diagnostic, but it seems the `else: raise AssertionError` is hiding it.
[...]

Issue #1894, opened 2025-01-29
chore: review `AssertionError` internally


MarcoGorelli:
I think we should only be raising `AssertionError` for cases when safety assertions fail, in code that we don't expect users to encounter

It would be good to review how we're using AssertionError and raise better error types where appropriate

mikeweltevrede:
Good point, I would say that an `AssertionError` should usually not be raised manually. According to [the Python docs](https://docs.python.org/3/library/exceptions.html#AssertionError), these are raised when an `assert` fails. Instead, we should indeed check which errors to raise. For example, in https://github.com/narwhals-dev/narwhals/blob/5550ad897992e3a639cea3fb0cca5bede98da6d7/narwhals/utils.py#L333 you would probably expect a `NotImplementedError` to be raised. Or in https://github.com/narwhals-dev/narwhals/blob/5550ad897992e3a639cea3fb0cca5bede98da6d7/narwhals/functions.py#L583, I would probably expect a `RuntimeError` or a custom `UnreachableCodeReachedError`.

Luckily, there are only a few occurrences in the code base: https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals%20AssertionError%20&type=code.

PS: Of course, this would be a breaking change where users who would expect to be catching an `AssertionError` now would see their code fail.

PR #2717, opened 2025-06-21
chore(typing): Add `_typing_compat.assert_never`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Mentioned in https://github.com/narwhals-dev/narwhals/issues/2715#issuecomment-2993628074
- Kinda related #1894

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_typing_compat.py
+++ b/narwhals/_typing_compat.py
@@ -33,6 +33,11 @@ if TYPE_CHECKING:
     else:
         from typing_extensions import TypeVar, deprecated
 
+    if sys.version_info >= (3, 11):
+        from typing import Never, assert_never
+    else:
+        from typing_extensions import Never, assert_never
+
     _Fn = TypeVar("_Fn", bound=Callable[..., Any])
 
 
@@ -65,9 +70,21 @@ else:  # pragma: no cover
 
             return wrapper
 
+    if sys.version_info >= (3, 11):
+        from typing import assert_never
+    else:
+        _ASSERT_NEVER_REPR_MAX_LENGTH = 100
+
+        def assert_never(arg: Never, /) -> Never:
+            value = repr(arg)
+            if len(value) > _ASSERT_NEVER_REPR_MAX_LENGTH:
+                value = value[:_ASSERT_NEVER_REPR_MAX_LENGTH] + "..."
+            msg = f"Expected code to be unreachable, but got: {value}"
+            raise AssertionError(msg)
+
     # TODO @dangotbanned: Remove after dropping `3.8` (#2084)
     # - https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1965921386
     from typing import Protocol as Protocol38
 
 
-__all__ = ["Protocol38", "TypeVar", "deprecated"]
+__all__ = ["Protocol38", "TypeVar", "assert_never", "deprecated"]
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -23,7 +23,7 @@ from narwhals._translate import (
     ToNarwhals,
     ToNarwhalsT_co,
 )
-from narwhals._typing_compat import deprecated
+from narwhals._typing_compat import assert_never, deprecated
 from narwhals._utils import (
     Version,
     _StoresNative,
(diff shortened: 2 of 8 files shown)


FBruzzesi (commented):
Thanks @dangotbanned - should we replace all the other "unreachable" code paths as well ([search](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+%22unreachable+code%22&type=code))?

One nice thing that we are losing is the suggestion "Please report an issue at https://github.com/narwhals-dev/narwhals/issues/new". I wonder if it's worth it to have our version of `assert_never` independently of the python version

dangotbanned:
> Thanks @dangotbanned 

Thanks for peeking @FBruzzesi 

> should we replace all the other "unreachable" code paths as well ([search](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+%22unreachable+code%22&type=code))?

If you can find any (besides https://github.com/narwhals-dev/narwhals/pull/2717/commits/db14ab1c94623238d33ee0b3986661e5e0afdf13) that both `pyright` and `mypy` agree are unreachable - I'm happy for those to use `assert_never` as well 🙂 

We can only use it when we can prove that every case has been checked, like this

image


It isn't just limited to `Literal`, this one is handling some pretty complex types 

image


Here for example, the type hasn't been narrowed to `Never` - so if we used `assert_never` it'd produce an error during type checking

image


---

> One nice thing that we are losing is the suggestion "Please report an issue at [narwhals-dev/narwhals/issues/new](https://github.com/narwhals-dev/narwhals/issues/new)". I wonder if it's worth it to have our version of `assert_never` independently of the python version

I'm open to using a custom message - only caveat is it has to be the same for every use.

We need to have the signature as this to be able to link whatever we use back to [`typing.assert_never`](https://docs.python.org/3/library/typing.html#typing.assert_never)

```py
def assert_never(arg: Never, /) -> Never:
```

I was playing around with the idea of different runtime behavior than the `stdlib` in (#2705).
We'd basically need to do something like this here though:

```py

import sys

if TYPE_CHECKING:
    from typing import assert_never

    if sys.version_info >= (3, 11):
        from typing import Never
    else:
        from typing_extensions import Never

else:
    def assert_never(arg: Never, /) -> Never:
        msg = "Hello"
        raise AssertionError(msg)
```

FBruzzesi:
> If you can find any (besides [db14ab1](https://github.com/narwhals-dev/narwhals/commit/db14ab1c94623238d33ee0b3986661e5e0afdf13)) that both `pyright` and `mypy` agree are unreachable - I'm happy for those to use `assert_never` as well 🙂

I will give it a shot 🤗

> I'm open to using a custom message - only caveat is it has to be the same for every use.
> 
> We need to have the signature as this to be able to link whatever we use back to [`typing.assert_never`](https://docs.python.org/3/library/typing.html#typing.assert_never)

I would suggest to simply append the link at the end of the message:

```python
def assert_never(arg: Never, /) -> Never:
    value = repr(arg)
    if len(value) > _ASSERT_NEVER_REPR_MAX_LENGTH:
        value = value[:_ASSERT_NEVER_REPR_MAX_LENGTH] + "..."
    
    msg = (
        f"Expected code to be unreachable, but got: {value}.\n"
        "Please report an issue at [narwhals-dev/narwhals/issues/new](https://github.com/narwhals-dev/narwhals/issues/new)"
    )

    raise AssertionError(msg)
```

narwhals/_typing_compat.py
```diff
@@ -65,9 +70,24 @@ def wrapper(func: _Fn, /) -> _Fn:
 
             return wrapper
 
+    _ASSERT_NEVER_REPR_MAX_LENGTH = 100
+    _BUG_URL = (
+        "https://github.com/narwhals-dev/narwhals/issues/new?template=bug_report.yml"
+    )
```
dangotbanned, line 76: https://github.com/narwhals-dev/narwhals/pull/2717#issuecomment-3003679679

@FBruzzesi I tried using the link you gave *first*, but I think this one might be more handy?

- https://github.com/narwhals-dev/narwhals/issues/new
- https://github.com/narwhals-dev/narwhals/issues/new?template=bug_report.yml

FBruzzesi: Yes nice one! Thank you 👌

MarcoGorelli (approved):
cool, thanks @dangotbanned !

Issue #2713, opened 2025-06-21
Refine public `Compliant*` methods


MarcoGorelli:
We're currently using the compliant classes to share some code between implementations

This is fine, but for protocols with other 3rd party backends should implement, I don't think this is ok:
- it will make it harder for us to change some internals
- it forces them to implement methods like `_evaluate_expr` which they really shouldn't need to care about

I'd like to suggest therefore that we aim to have:
- `CompliantDataFrame`, `CompliantLazyFrame`, `CompliantSeries`, `CompliantExpr`, `CompliantNamespace`: these should be `Protocol`s which _only_ implement the necessary functionality which we call from `narwhals.DataFrame` / ...
- we can subclass these to preserve some common functionality within Narwhals. For example, `DuckDBExpr` / `SparkLikeExpr` / ... all inherit from `LazyExpr`, which in turn inherits from `CompliantExpr`. It's fine to have some implementations in `LazyExpr`

## PRs
- [x] #2934
- [x] #3041
- [x] #3045
- [x] #3054
- [x] #3055 
- [x] #3056 
- [x] #3062 
- [ ] #3140

dangotbanned:
Sounds good, agreed!

I recall @FBruzzesi and I discussing something similar and there was a mermaid diagram with an idea
But it seems we forgot 🤦‍♂ 

I'll try to find it since it might help with where to move stuff


## Update
Found it (https://github.com/narwhals-dev/narwhals/pull/2563#issuecomment-2888428494)

So we'd just be prefixing the new `Protocol`s with something like `Impl`

<details><summary>Hidden, see the updated version in <a href="https://github.com/narwhals-dev/narwhals/issues/2713#issuecomment-3012874025"</a> comment </i></summary>

<p>

> > @MarcoGorelli @dangotbanned I think we start to have a few piece of the code base (such as this one) which can be abstracted for duckdb, ibis and spark-like (dask tbd). I am having a hard time to come up with a name for such compliant class, mostly because `EagerDataFrame` inherits from `CompliantLazyFrame` 😂 `LazyDataFrame` does not sound right for some reason.
> 
> Well spotted @FBruzzesi!
> 
> The main blocker is that `EagerDataFrame` inherits both protocols.
> 
[...]

PR #2718, opened 2025-06-21
refactor: Enforce `LazyExpr.from_elementwise`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- See https://github.com/narwhals-dev/narwhals/pull/2716#discussion_r2160078141

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -32,7 +32,7 @@ from narwhals._utils import _StoresCompliant, not_implemented
 from narwhals.dependencies import get_numpy, is_numpy_array
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping, Sequence
+    from collections.abc import Iterable, Mapping, Sequence
 
     from typing_extensions import Self, TypeIs
 
@@ -893,6 +893,11 @@ class LazyExpr(
 
     @classmethod
     def _alias_native(cls, expr: NativeExprT, name: str, /) -> NativeExprT: ...
+
+    @classmethod
+    def from_elementwise(
+        cls, func: Callable[[Iterable[NativeExprT]], NativeExprT], *exprs: Self
+    ) -> Self: ...
 
     @property
     def name(self) -> LazyExprNameNamespace[Self]:
--- a/narwhals/_duckdb/expr.py
+++ b/narwhals/_duckdb/expr.py
@@ -21,11 +21,15 @@ from narwhals._duckdb.utils import (
     narwhals_to_native_dtype,
     when,
 )
-from narwhals._expression_parsing import ExprKind
+from narwhals._expression_parsing import (
+    ExprKind,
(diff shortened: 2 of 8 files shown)


narwhals/_compliant/expr.py
```diff
@@ -894,6 +894,11 @@ def fn(names: Sequence[str]) -> Sequence[str]:
     @classmethod
     def _alias_native(cls, expr: NativeExprT, name: str, /) -> NativeExprT: ...
 
+    @classmethod
+    def from_elementwise(
+        cls, func: Callable[[Iterable[NativeExprT]], NativeExprT], *exprs: Self
+    ) -> Self: ...
```
dangotbanned: @MarcoGorelli would you be open to adding a little doc here? 🙏

narwhals/_duckdb/expr.py
```diff
@@ -213,6 +217,32 @@ def func(df: DuckDBLazyFrame) -> list[Expression]:
             version=context._version,
         )
 
+    @classmethod
+    def from_elementwise(
+        cls, func: Callable[[Iterable[Expression]], Expression], *exprs: Self
+    ) -> Self:
+        def call(df: DuckDBLazyFrame) -> list[Expression]:
+            cols = (col for _expr in exprs for col in _expr(df))
```
dangotbanned, line 229: I think these could just be:

```py
return [func(chain.from_iterable(expr(df) for expr in exprs))]
```

narwhals/_compliant/expr.py
dangotbanned: Actually, another thought.

What about `LazyExpr.from_horizontal_op`?

That would mirror the same constructor these all use for `ExprMetadata`

https://github.com/narwhals-dev/narwhals/blob/f17acfcb17b9ebb3e8267fa6cb31da87113d1005/narwhals/functions.py#L1369-L1374

https://github.com/narwhals-dev/narwhals/blob/f17acfcb17b9ebb3e8267fa6cb31da87113d1005/narwhals/_expression_parsing.py#L477-L481

I think the detail that we should use this for `*_horizontal` (and `concat_str`) is more relevant than the `elementwise` part

dangotbanned:
> thanks @dangotbanned
> 
> i've renamed and prefixed with an underscore as I don't think this should be part of the compliant protocols #2713, but fine for now


Thanks @MarcoGorelli!

One thing though - the name in the PR title doesn't match the code 😉

I prefer the one without elementwise

MarcoGorelli:
thanks! same, but it does need to be elementwise if we use `window_function` the way it's defined inside it

PR #2719, opened 2025-06-22
feat: add `ignore_nulls` to `any_horizontal` and `all_horizontal`

### LINKED ISSUE #2678: API: align `any_horizontal` and `all_horizontal` with Polars' future behaviour?
The future default behaviour of `{any,all}_horizontal` will likely change in Polars

I think the most future-proof way to handle this in Narwhals would be to add `ignore_nulls` to `{any,all}_horizontal`, and to make it a required argument in the main namespace

If later we want to make it non-required, then that could be done in a backwards-compatible manner. But we should wait for Polars 2.0 to do that.

So, plan:
- introduce `ignore_nulls` in `{any, all}_horizontal`
- in `stable.v1`, it defaults to `False` (backwards-compatibility)
- in `main`, it has no default (it's a required argument)
- once Polars 2.0 comes out, we consider making the default `True` in the main Narwhals namespace / v2. Easier to start strict and loosen later

MarcoGorelli:
Just a start, and curious to see the effect on downstream libraries (presumably, nothing?)

~~Still needs doing for `all_horizontal`~~

closes #2678 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


tests/expr_and_series/all_horizontal_test.py
```diff
@@ -8,38 +9,75 @@
 from tests.utils import POLARS_VERSION, Constructor, ConstructorEager, assert_equal_data
 
 
-@pytest.mark.parametrize("expr1", ["a", nw.col("a")])
-@pytest.mark.parametrize("expr2", ["b", nw.col("b")])
-def test_allh(constructor: Constructor, expr1: Any, expr2: Any) -> None:
+def test_allh(constructor: Constructor) -> None:
```
MarcoGorelli, line 12: let's cut down on unnecessary test times a little..

Issue #267, opened 2024-06-07
[Enh]: Registering extensions via narwhals to downstream dataframe libraries


FBruzzesi:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

Direct use case I have in mind is a library I created and maintain: [iso-week-date](https://github.com/FBruzzesi/iso-week-date)

### Please describe the purpose of the new feature or describe the problem to solve.

The objective is to have a way to register an extension namespace in narwhals once, which is then propagated to the downstream libraries.

Here is the problem/headache I face: as a library creator I want to provide both a [pandas](https://pandas.pydata.org/docs/development/extending.html) and a [polars](https://docs.pola.rs/py-polars/html/reference/api.html) extension by registering namespaces.

In the current scenario, I am duplicating explicitly writing every method for each library ([pandas](https://github.com/FBruzzesi/iso-week-date/blob/main/iso_week_date/pandas_utils.py#L324) and [polars](https://github.com/FBruzzesi/iso-week-date/blob/main/iso_week_date/polars_utils.py#L328)), which is far from ideal.

### Suggest a solution if possible.

The ideal scenario would be to write the extension once using narwhals and decorate it for each downstream library.

Maybe this is already possible, but I would like to double check if that's the case (and if the answer is positive I think it would be awesome to document). Here is the closest I got:

**_core.py**
```py
import narwhals as nw

class PlusOneSeriesNW:
    def __init__(self, series: nw.Series):
        self._series: nw.Series = series

    def plus_one(self):
         return self.series + 1
```

**pandas_extension.py**
```py
from ._core import PlusOneSeriesNW

@pd.api.extensions.register_series_accessor("p1")
class PlusOneSeries:
    def __init__(self, series: pd.Series):
        self._series = PlusOneSeriesNW(nw.from_series(series, ...))
    
    def plus_one(self):
        return nw.to_series(self._series.plus_one(), ...)
```
Now if someone imports `PlusOneSeries` from the library, it will be possible to run: `series.p1.plus_one()`

**polars_extension.py**
> Ok just same as above

### Would you want to open a pull request?

yes

wd60622:
Would be an awesome addition. I believe I have the same use-case as described for my package latent-calendar where my main API is [a pandas extension](https://wd60622.github.io/latent-calendar/examples/cal-attribute/). I would like generalize this if I could!

MarcoGorelli:
[...]

PR #2721, opened 2025-06-22
feat: Custom namespace registration

Excidion:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #267 
- Enabler for [alteryx/featuretools/issues/1967](https://github.com/alteryx/featuretools/issues/1967#issuecomment-2973605932)

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Adds ability to register new namespaces like so (I was surprised narwhals didn't have ceil and floor yet):
```py
import narwhals as nw
import pandas as pd

@nw.api.register_expr_namespace("rounding")
class Rounding:
    def __init__(self, expr: nw.Expr) -> None:
        self._expr = expr

    def ceil(self, dtype=nw.Int64) -> nw.Expr:
        return (self._expr + 1).cast(dtype=dtype)   # i know this doesn't actually work for all cases

    def floor(self, dtype=nw.Int64) -> nw.Expr:
        return self._expr.cast(dtype=dtype)


df = nw.from_native(pd.DataFrame([1.4, 24.3, 55.7, 64.001], columns=["n"]))
print(df.select(
    nw.col("n"),
    nw.col("n").rounding.floor().alias("floor"),
    nw.col("n").rounding.ceil().alias("ceil"),
).to_native().to_markdown())
"""
|    |      n |   floor |   ceil |
|---:|-------:|--------:|-------:|
|  0 |  1.4   |       1 |      2 |
|  1 | 24.3   |      24 |     25 |
|  2 | 55     |      55 |     56 |
|  3 | 64.001 |      64 |     65 |
"""
```

Heavily "inspired" by the polars implementation. Still needs tests and also updates to the docstrings. **Just wanted to know if this is even going in the right direction before proceeding?**

Checklist:
+ [x] Tested for `Expr`
+ [ ] Tested for `Series` ?
+ [x] Tested for `DataFrame` ?
+ [ ] Tested for `LazyFrame` ?

This is my first time contributing to FOSS, so please forgive if I missed anything. :)

### DIFF
--- a/narwhals/__init__.py
+++ b/narwhals/__init__.py
@@ -2,11 +2,6 @@ from __future__ import annotations
 
 import typing as _t
 
-from narwhals.api import (
-    register_dataframe_namespace,
-    register_expr_namespace,
-    register_lazyframe_namespace,
-)
 from narwhals import dependencies, dtypes, exceptions, selectors
 from narwhals._utils import (
     Implementation,
@@ -17,6 +12,11 @@ from narwhals._utils import (
     maybe_get_index,
     maybe_reset_index,
     maybe_set_index,
+)
+from narwhals.api import (
+    register_dataframe_namespace,
+    register_expr_namespace,
+    register_lazyframe_namespace,
 )
 from narwhals.dataframe import DataFrame, LazyFrame
 from narwhals.dtypes import (
--- a/narwhals/_reexport.py
+++ b/narwhals/_reexport.py
@@ -1,4 +1,5 @@
 """Re-export Narwhals functionality to avoid cyclical imports."""
+from __future__ import annotations
 
 from narwhals.dataframe import DataFrame, LazyFrame
 from narwhals.expr import Expr
--- a/narwhals/api.py
+++ b/narwhals/api.py
@@ -51,7 +51,7 @@ def _create_namespace(
         if name in _reserved_namespaces:
             msg = f"cannot override reserved namespace {name!r}"
             raise AttributeError(msg)
-        elif hasattr(cls, name):
+        if hasattr(cls, name):
             warn(
                 f"Overriding existing custom namespace {name!r} (on {cls.__name__!r})",
                 UserWarning,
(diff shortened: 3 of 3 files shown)


MarcoGorelli:
thanks for the pr!

it's not clear to me why this is needed - could you clarify why this is more desirable than just
```python
import narwhals as nw
import pandas as pd

def ceil(expr, dtype=nw.Int64) -> nw.Expr:
    return (expr + 1).cast(dtype=dtype)

def floor(expr, dtype=nw.Int64) -> nw.Expr:
    return expr.cast(dtype=dtype)


df = nw.from_native(pd.DataFrame([1.4, 24.3, 55.7, 64.001], columns=["n"]))
print(df.select(
    nw.col("n"),
    floor(nw.col("n")).alias("floor"),
    ceil(nw.col("n")).alias("ceil"),
).to_native().to_markdown())


```
please?

Excidion:
Sure! So the original starting point is my wish to [use narwhals as a backend for featuretools](https://github.com/alteryx/featuretools/issues/1967#issuecomment-2973582281).

I looked into the featuretools code and found out that it heavily relies on [woodwork](https://github.com/alteryx/woodwork) for it's logic. So as I understand it, the first step would have to be, to narwhalify [woodworks WoodworkTableAccessor class](https://woodwork.alteryx.com/en/stable/generated/woodwork.table_accessor.WoodworkTableAccessor.html#).

For now, woodwork only supports pandas and registers itself to a `pd.DataFrame`'s Namspace  [like that](https://github.com/alteryx/woodwork/blob/main/woodwork/table_accessor.py#L1864):
```py
# woodwork/woodwork/table_accessor.py
@pd.api.extensions.register_dataframe_accessor("ww")
class PandasTableAccessor(WoodworkTableAccessor):
    pass
```
Therefore my idea was, that if narwhals had a similar capability, we could add something like this:
```py
# woodwork/woodwork/table_accessor.py
@nw.api.register_dataframe_accessor("ww")
class NarwhalsTableAccessor(WoodworkTableAccessor):
    pass  # TODO
```
Therefore I looked into the polars repo to see how it's implemented there as inspiration. The result is this PR.
(In the meantime I also tested the `nw.api.register_dataframe_accessor` function with positive results.)

Hope this gives a bit better context. Love to hear your thoughts if you think this could benefit narwhals generally.

Excidion:
@MarcoGorelli  if you think this is not within scope (maybe also related to https://github.com/narwhals-dev/narwhals/issues/267) let me know, then I'll investigate other options between woodwork and narwhals :)

PR #2729, opened 2025-06-23
chore: use `window_expression` for DuckDB

MarcoGorelli:
This should help reduce churn when they introduce `duckdb.WindowExpression`

It also reduces how much sql we need to write by concentrating it all in `window_expression`, reducing the potential points of failure

~~`rank` still needs doing~~

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_duckdb/utils.py
```diff
@@ -280,8 +282,41 @@ def generate_partition_by_sql(*partition_by: str | Expression) -> str:
...
+    descending: bool = False,
+    ignore_nulls: bool = False,
+) -> Expression:
+    # TODO(unassigned): Replace with `duckdb.WindowExpression` when they release it.
+    # https://github.com/duckdb/duckdb/discussions/14725#discussioncomment-11200348
+    try:
+        from duckdb import SQLExpression
+    except ModuleNotFoundError as exc:
+        msg = f"DuckDB>=1.3.0 is required for this operation. Found: DuckDB {duckdb.__version__}"
+        raise NotImplementedError(msg) from exc
```
dangotbanned: This is a bit of a downgrade from the message we had before - since we lose the context of what the operation was 🤔 

Was there an issue with having `@requires.backend_version((1, 3))` on the methods?

https://github.com/narwhals-dev/narwhals/blob/b656c4ef5ff325c777984ffab8ee013aeaeffa7d/narwhals/_utils.py#L1806-L1827

MarcoGorelli: it just seemed redundant at this point, but sure, no objections to keeping the `requires` in

PR #2737, opened 2025-06-26
chore: Add `_constants` module

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [ ] Tests added
- [x] Documented the changes

The new module is the `narwhals`-equivalent to [`polars/_utils/constants.py`](https://github.com/pola-rs/polars/blob/2f53652d5e6c6f57800db38804b7eb5948a815fe/py-polars/polars/_utils/constants.py)

I was drawn to this mainly because of the temporal constants - which I always have trouble remembering 😄
To track down the others, I've enabled [magic-value-comparison (`PLR2004`)](https://docs.astral.sh/ruff/rules/magic-value-comparison/) which helps with some - but not all constants.

*Beyond* these changes, I haven't scoured everywhere to find others - but no objections from me if anyone else can find some that could use a meaningful name 🙂
Quite often, there will be a comment next to the value/calculation explaining it

### DIFF
--- a/narwhals/_constants.py
+++ b/narwhals/_constants.py
@@ -1,5 +1,22 @@
 from __future__ import annotations
 
+import datetime as dt
+
 MAX_REPR_LENGTH = 40
 MAX_WHILE_REPEAT = 100
 NON_ELEMENTARY_DEPTH = 2
+
+# Temporal (from `polars._utils.constants`)
+SECONDS_PER_DAY = 86_400
+SECONDS_PER_HOUR = 3_600
+NS_PER_SECOND = 1_000_000_000
+"""Nanoseconds per second."""
+US_PER_SECOND = 1_000_000
+"""Microseconds per second."""
+MS_PER_SECOND = 1_000
+"""Milliseconds per second."""
+
+EPOCH_YEAR = 1970
+EPOCH_DATE = dt.date(EPOCH_YEAR, 1, 1)
+EPOCH = dt.datetime(EPOCH_YEAR, 1, 1).replace(tzinfo=None)
+EPOCH_UTC = dt.datetime(EPOCH_YEAR, 1, 1, tzinfo=dt.timezone.utc)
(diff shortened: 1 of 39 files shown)


narwhals/_constants.py
```diff
@@ -0,0 +1,37 @@
+from __future__ import annotations
+
+import datetime as dt
+
+MAX_REPR_LENGTH = 40
+"""Length to begin truncating an object repr in an error message."""
+MAX_WHILE_REPEAT = 100
+"""Loop count to trigger an error if we can't generate a unique column name."""
+NON_ELEMENTARY_DEPTH = 2
+"""Threshold for complex group by aggregations in `pandas`, `pyarrow`, `dask`."""
```
MarcoGorelli: i like the temporal ones, but would prefer to revert these

dangotbanned: No worries, should be resolved in (https://github.com/narwhals-dev/narwhals/pull/2737/commits/7365e1be2ee4a62c346808d0da60c7864bd4865a)

pyproject.toml
```diff
@@ -169,7 +169,6 @@ ignore = [
   "PD901",   # pandas-df-variable-name (This is a auxiliary library so dataframe variables have no concrete business meaning)
   "PLC0415", # `import` should be at the top-level of a file
   "PLR0913", # too-many-arguments
-  "PLR2004", # magic-value-comparison
```
MarcoGorelli: let's keep this turned off globally, tbh i think this one's too noisy

dangotbanned: Happy for you to revert it, or I can later today?

dangotbanned: Okay now we have arrived 😄 (https://github.com/narwhals-dev/narwhals/pull/2737/commits/b6424850023ed33aa5e36012e31a53c1b3199d90)

PR #2008, opened 2025-02-13
chore: fix pandas-stubs issues

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2740, opened 2025-06-26
RFC, patch: Preserves dtype_backend in `str.to_datetime`

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

So I was looking at how to implement `str.to_date` and for pandas I was falling back to `.str.to_datetime().dt.date()`, and then noticed that we are not preserving the dtype backend.

This is an issue upstream: [pandas#58220](https://github.com/pandas-dev/pandas/issues/58220)
But in the meanwhile we can convert it back (not sure this is the best way of achieving it)

---

Disclaimer: the test I added is failing for `numpy_nullable`, I will comment down below in the codebase

### DIFF
--- a/tests/expr_and_series/str/to_datetime_test.py
+++ b/tests/expr_and_series/str/to_datetime_test.py
@@ -2,13 +2,14 @@ from __future__ import annotations
 
 from contextlib import nullcontext as does_not_raise
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Literal
 
 import pyarrow as pa
 import pytest
 
 import narwhals as nw
 from narwhals._arrow.utils import parse_datetime_format
+from narwhals._pandas_like.utils import get_dtype_backend
 from tests.utils import (
     PYARROW_VERSION,
     assert_equal_data,
@@ -231,3 +232,22 @@ def test_to_datetime_tz_aware(
             "b": [datetime(2020, 1, 1, 0, 2, 3, tzinfo=timezone.utc)],
         }
         assert_equal_data(result, expected)
+
+
+@pytest.mark.parametrize("dtype_backend", ["pyarrow", "numpy_nullable"])
+def test_to_datetime_pd_preserves_pyarrow_backend_dtype(
+    dtype_backend: Literal["pyarrow", "numpy_nullable"],
+) -> None:
+    pytest.importorskip("pandas")
+    import pandas as pd
+
+    df = nw.from_native(
+        pd.DataFrame({"a": ["2020-01-01T12:34:56", None]}).convert_dtypes(
+            dtype_backend=dtype_backend
+        )
+    )
+    result = df.with_columns(b=nw.col("a").str.to_datetime()).to_native()
+    result_dtype = get_dtype_backend(
+        result["b"].dtype, df._compliant_frame._implementation
+    )
+    assert result_dtype == dtype_backend
--- a/narwhals/_pandas_like/series_str.py
+++ b/narwhals/_pandas_like/series_str.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any
 from narwhals._compliant.any_namespace import StringNamespace
 from narwhals._pandas_like.utils import (
     PandasLikeSeriesNamespace,
+    get_dtype_backend,
     is_pyarrow_dtype_backend,
 )
 
@@ -68,8 +69,14 @@ class PandasLikeSeriesStringNamespace(
         return result
 
     def _to_datetime(self, format: str | None, *, utc: bool) -> Any:
-        return self.implementation.to_native_namespace().to_datetime(
+        dtype_backend = get_dtype_backend(self.native.dtype, self.implementation)
+        result = self.implementation.to_native_namespace().to_datetime(
             self.native, format=format, utc=utc
+        )
+        return (
+            result.convert_dtypes(dtype_backend=dtype_backend)
+            if dtype_backend is not None
+            else result
         )
 
     def to_uppercase(self) -> PandasLikeSeries:


tests/expr_and_series/str/to_datetime_test.py
```diff
@@ -231,3 +232,22 @@ def test_to_datetime_tz_aware(
...
+    pytest.importorskip("pandas")
+    import pandas as pd
+
+    df = nw.from_native(
+        pd.DataFrame({"a": ["2020-01-01T12:34:56", None]}).convert_dtypes(
+            dtype_backend=dtype_backend
+        )
+    )
+    result = df.with_columns(b=nw.col("a").str.to_datetime()).to_native()
+    result_dtype = get_dtype_backend(
```
FBruzzesi, line 254: `get_dtype_backend` returns `None` for numpy_nullable. I wonder if that's correct for the datetime case

dangotbanned: I just started debugging ...

Did you know on L249 the dtype is `string[python]`?

```py
native = df.to_native()
>>> native.dtypes
a    string[python]
dtype: object
```

FBruzzesi: Yes, yet `get_dtype_backend` doesn't pick up any difference between that and `object`:

 ```py
import narwhals as nw
import pandas as pd

from narwhals._pandas_like.utils import get_dtype_backend

data = ["a", "b"]

s = pd.Series(data)
print(s.dtype)
print(get_dtype_backend(s.dtype, nw.Implementation.PANDAS))

s_np_null = pd.Series(data).convert_dtypes(dtype_backend="numpy_nullable")
print(s_np_null.dtype)
print(get_dtype_backend(s_np_null.dtype, nw.Implementation.PANDAS))
```
```terminal
object
None
string
None
```

FBruzzesi: Is this the issue 👀?

```diff
diff --git a/narwhals/_pandas_like/utils.py b/narwhals/_pandas_like/utils.py
index 4cdd979de..7460cf8ec 100644
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -329,7 +329,7 @@ def get_dtype_backend(dtype: Any, implementation: Implementation) -> DTypeBacken
         sentinel = object()
         if (
             isinstance(dtype, pd.api.extensions.ExtensionDtype)
-            and getattr(dtype, "base", sentinel) is None
+            and getattr(dtype, "base", sentinel) is sentinel
         ):
             return "numpy_nullable"
     return None
```

FBruzzesi: Ok I am going on a tangent but - with that change:

```py
import narwhals as nw
import pandas as pd

from narwhals._pandas_like.utils import get_dtype_backend

data = ["a", "b"]

s = pd.Series(data)
print(s.dtype)
print(get_dtype_backend(s.dtype, nw.Implementation.PANDAS))

s_np_null = pd.Series(data).convert_dtypes(dtype_backend="numpy_nullable")
print(s_np_null.dtype)
print(get_dtype_backend(s_np_null.dtype, nw.Implementation.PANDAS))
```
```terminal
object
None
string
numpy_nullable
```

However for datetime dtype, nothing changes, and it goes undetected 🤔

```py
from datetime import datetime
import pandas as pd

s = pd.Series([datetime(2020, 1, 1)])

print(s.dtype)
# datetime64[ns]

print(s.convert_dtypes(dtype_backend="numpy_nullable").dtype)
# datetime64[ns]
```

dangotbanned: > Is this the issue?

~~No~~ (Actually I think we're saying the same thing 😄) 
I think the problem is that `string[python]` isn't a numpy dtype

This is what it's being converted to 

https://github.com/pandas-dev/pandas/blob/7096342bb91c4ac21cf498567d5b09eee8534543/pandas/core/arrays/string_.py#L90-L122

But we're only recognising these as backends

https://github.com/narwhals-dev/narwhals/blob/7cfa873bfd32eb3f3963f58b36ed0e50239c6331/narwhals/typing.py#L244

dangotbanned: There's apparently all of these options 🤔

- `"string"`
- `"string[python]"`
- `"string[pyarrow]"`
- `"string[pyarrow_numpy]"`

https://github.com/pandas-dev/pandas/blob/7096342bb91c4ac21cf498567d5b09eee8534543/pandas/core/arrays/string_.py#L140-L181

MarcoGorelli: > However for datetime dtype, nothing changes, and it goes undetected 🤔

pandas doesn't have a numpy-nullable datetime dtype, so `.convert_dtypes(dtype_backend="numpy_nullable")` is a no-op there

FBruzzesi: Thanks for the info @MarcoGorelli - as follow up questions

- What are your thoughts on the [comment on sentinel value](https://github.com/narwhals-dev/narwhals/pull/2740#discussion_r2169904874)?
- Should I go ahead and remove the check for numpy-nullable then, and keep it only for pyarrow?

MarcoGorelli: tbh i don't remember the `and getattr(dtype, "base", sentinel) is None` change, @dangotbanned do you remember the logic behind it?

dangotbanned: > tbh i don't remember the `and getattr(dtype, "base", sentinel) is None` change, @dangotbanned do you remember the logic behind it?

If I did that, I hope I left something on the commit description 🫣

dangotbanned: > If I did that, I hope I left something on the commit description 🫣

Thanks earlier dan, here we go (https://github.com/narwhals-dev/narwhals/pull/2008/commits/bad8f97d083d16da53df7d1cc7ef90ae54e0fb31)

The `None` check was to get here, without importing the `pandas` internals - which `pyright` wasn't happy with

https://github.com/pandas-dev/pandas/blob/7096342bb91c4ac21cf498567d5b09eee8534543/pandas/core/dtypes/dtypes.py#L1533-L1538

```py
class BaseMaskedDtype(ExtensionDtype):
    """
    Base class for dtypes for BaseMaskedArray subclasses.
    """

    base = None
```


@FBruzzesi this wouldn't get what we want sadly.

`sentinel` is there because we **want** the value `None`, and if we didn't find it - we need to treat `None` as a positive

```diff
-            and getattr(dtype, "base", sentinel) is None
+            and getattr(dtype, "base", sentinel) is sentinel
```


See also (https://github.com/narwhals-dev/narwhals/pull/2683#discussion_r2160394655) 😉 


Hopefully this should help going forward
- #2744

FBruzzesi: Thanks @dangotbanned 👌Is it worth to add a comment in the code for our future selves? 

>  pandas doesn't have a numpy-nullable datetime dtype, so .convert_dtypes(dtype_backend="numpy_nullable") is a no-op there

I will keep the check only for pyarrow for now then - pushing in a minute

FBruzzesi:
@MarcoGorelli I am catching up with the live stream now.

The reason for:

```py
dtype_backend: Literal["pyarrow", "numpy_nullable"] = "pyarrow"
```

is to be compliant with the type checker(s) (due to pandas-stubs). Alternatively I could have `type: ignore[...]` in `.convert_dtypes(...)`

mattharrison:
I'm confused. Is `to_datetime` deprecated? If I want to be arrow backend native but have oddly formatted date strings, I have to use `to_datetime` with `format`, then convert the numpy dates to pyarrow? I don't think `convert_dtypes` handles format dates.

MarcoGorelli:
@mattharrison could you open a new issue with a code example please and Expected output?

PR #2679, opened 2025-06-14
chore: Drop support for python `3.8`

dangotbanned:
Closes #2084

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2084

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
Let's see how CI feels about it now ... 🤞

PR #2744, opened 2025-06-27
chore: Split out, doc `is_dtype_numpy_nullable`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/2740#discussion_r2171163397
- https://github.com/narwhals-dev/narwhals/commit/bad8f97d083d16da53df7d1cc7ef90ae54e0fb31
- #2008

## Checklist

- [x] Tests added
- [x] Documented the changes

Since we all had a hard time understanding the check, seems best to give it a name and typing 🙂


narwhals/_pandas_like/utils.py
```diff
@@ -334,11 +345,7 @@ def get_dtype_backend(dtype: Any, implementation: Implementation) -> DTypeBacken
     if hasattr(pd, "ArrowDtype") and isinstance(dtype, pd.ArrowDtype):
         return "pyarrow"
     with suppress(AttributeError):
-        sentinel = object()
-        if (
-            isinstance(dtype, pd.api.extensions.ExtensionDtype)
-            and getattr(dtype, "base", sentinel) is None
-        ):
+        if is_dtype_masked(dtype):
```
dangotbanned: I mentioned before about the `suppress` maybe not being needed in (https://github.com/narwhals-dev/narwhals/pull/2008#discussion_r1958289318)

**If** `pd.api.extensions.ExtensionDtype` exists in our minimum `pandas` - then we can remove the `suppress`

`pandas` (itself) is always available, since we import it at the top-level

https://github.com/narwhals-dev/narwhals/blob/2e382405abf601b05b72c5169e2fa30b49ad0830/narwhals/_pandas_like/utils.py#L8

FBruzzesi: As #2679 bumped pandas min version to 1.1.3, then yes `pd.api.extensions.ExtensionDtype` is always available 👌

dangotbanned: Ohh nice one, thanks for checking @FBruzzesi!

I'll try this out later

MarcoGorelli:
thanks - could you leave a comment about the sentinel part please?

narwhals/_pandas_like/utils.py
dangotbanned: Yeah seems to have worked in (https://github.com/narwhals-dev/narwhals/pull/2744/commits/bc9d619444ea8a08d0fc74f8bd9039cf590ff380)

narwhals/_pandas_like/utils.py
```diff
@@ -327,6 +327,8 @@ def native_to_narwhals_dtype(
 
 def is_dtype_masked(dtype: Any) -> TypeIs[BaseMaskedDtype]:
     """Return `True` if `dtype` is `"numpy_nullable"`."""
+    # NOTE: We need a sentinel as the positive case is `BaseMaskedDtype.base = None`
+    # See https://github.com/narwhals-dev/narwhals/pull/2740#discussion_r2171667055
     sentinel = object()
```
MarcoGorelli: stil not sure i get it, what's the issue if `.base` is `None`? doesn't look like there's any tests which fail without this?

MarcoGorelli: and yes, i've read the comment

> sentinel is there because we want the value None, and if we didn't find it - we need to treat None as a positive

dangotbanned:
I've just marked this as draft as I've noticed something, maybe this would make more sense:


The aliasing for `is_pyarrow_dtype_backend` is just to keep the diff in this file, but Ideally `is_pyarrow_dtype_backend` would be replaced 

```diff
diff --git a/narwhals/_pandas_like/utils.py b/narwhals/_pandas_like/utils.py
index d5555649d..5b126e4be 100644
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -325,7 +325,7 @@ def native_to_narwhals_dtype(
     raise AssertionError(msg)
 
 
-def is_dtype_masked(dtype: Any) -> TypeIs[BaseMaskedDtype]:
+def is_dtype_numpy_nullable(dtype: Any) -> TypeIs[BaseMaskedDtype]:
     """Return `True` if `dtype` is `"numpy_nullable"`."""
     # NOTE: We need a sentinel as the positive case is `BaseMaskedDtype.base = None`
     # See https://github.com/narwhals-dev/narwhals/pull/2740#discussion_r2171667055
@@ -343,14 +343,17 @@ def get_dtype_backend(dtype: Any, implementation: Implementation) -> DTypeBacken
     """
     if implementation is Implementation.CUDF:
         return None
-    if hasattr(pd, "ArrowDtype") and isinstance(dtype, pd.ArrowDtype):
+    if is_dtype_pyarrow(dtype):
         return "pyarrow"
-    return "numpy_nullable" if is_dtype_masked(dtype) else None
+    return "numpy_nullable" if is_dtype_numpy_nullable(dtype) else None
 
 
 @functools.lru_cache(maxsize=16)
-def is_pyarrow_dtype_backend(dtype: Any, implementation: Implementation) -> bool:
-    return get_dtype_backend(dtype, implementation) == "pyarrow"
+def is_dtype_pyarrow(dtype: Any) -> TypeIs[pd.ArrowDtype]:
+    return hasattr(pd, "ArrowDtype") and isinstance(dtype, pd.ArrowDtype)
+
+
+is_pyarrow_dtype_backend = is_dtype_pyarrow
 
 
 def narwhals_to_native_dtype(  # noqa: C901, PLR0912, PLR0915

```

MarcoGorelli (changes requested):
tbh this sentinel trick looks really brittle, and i'd prefer to go back to checking for `BaseMaskedArray` please, as it was in https://github.com/narwhals-dev/narwhals/pull/2008/files

we're testing against nightlies anyway, we'll be alerted if that moves

in any case, can we add a unit test for this function? create a series with `pd.Series([1,2,3], dtype='Int64')` and check that `is_dtype_masked` returns `True`

narwhals/_pandas_like/utils.py
MarcoGorelli: i think i get it, and don't like how we're using an implementation detail like this. `pandas.core` may well be semi-private, but it likely won't change in the same way

this was also brought up in the pandas slack as a recommend way to check for nullable types, in the absense of something fully public https://pandas-dev-community.slack.com/archives/C03PH1SU1M1/p1709823580407469?thread_ts=1709822239.679639&cid=C03PH1SU1M1

dangotbanned: Does this help illustrate it better?

```py
import pandas as pd

data = {
    "a": [1, 2],
    "b": ["one", "two"],
    "c": ["one", None],
    "d": pd.Categorical(["A", "B"]),
}
df = pd.DataFrame(data).convert_dtypes()
all_dtypes = pd.concat(
    [
        df.dtypes,
        df.convert_dtypes(dtype_backend="numpy_nullable").dtypes,
        df.convert_dtypes(dtype_backend="pyarrow").dtypes,
    ]
).to_list()

sentinel = object()
results = [
    {
        "dtype": dtype,
        "is_ext": isinstance(dtype, pd.api.extensions.ExtensionDtype),
        "None default": getattr(dtype, "base", None) is None,
        "sentinel default": getattr(dtype, "base", sentinel) is None,
    }
    for dtype in all_dtypes
]
>>> results
```

Essentially, using `None` gives false positives

<details><summary>Output</summary>
<p>


```py
[{'dtype': Int64Dtype(),
  'is_ext': True,
  'None default': True,
  'sentinel default': True},
 {'dtype': string[python],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': string[python],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': CategoricalDtype(categories=['A', 'B'], ordered=False, categories_dtype=object),
  'is_ext': True,
  'None default': False,
  'sentinel default': False},
 {'dtype': Int64Dtype(),
  'is_ext': True,
  'None default': True,
  'sentinel default': True},
 {'dtype': string[python],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': string[python],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': CategoricalDtype(categories=['A', 'B'], ordered=False, categories_dtype=object),
  'is_ext': True,
  'None default': False,
  'sentinel default': False},
 {'dtype': int64[pyarrow],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': string[pyarrow],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': string[pyarrow],
  'is_ext': True,
  'None default': True,
  'sentinel default': False},
 {'dtype': CategoricalDtype(categories=['A', 'B'], ordered=False, categories_dtype=object),
  'is_ext': True,
  'None default': False,
  'sentinel default': False}]
```

</p>
</details>

dangotbanned: > this was also brought up in the pandas slack as a recommend way to check for nullable types, in the absense of something fully public [pandas-dev-community.slack.com/archives/C03PH1SU1M1/p1709823580407469?thread_ts=1709822239.679639&cid=C03PH1SU1M1](https://pandas-dev-community.slack.com/archives/C03PH1SU1M1/p1709823580407469?thread_ts=1709822239.679639&cid=C03PH1SU1M1)

@MarcoGorelli I don't seem to be able to access that 😢 

<details><summary>Screenshot</summary>
<p>

image

</p>
</details> 

But I'm happy for this to be switched to a recommended alternative.
The original change was just seeking to use something *more* public to satisfy `pyright` (https://github.com/narwhals-dev/narwhals/pull/2008#discussion_r1958246132)

dangotbanned:
> in any case, can we add a unit test for this function? create a series with `pd.Series([1,2,3], dtype='Int64')` and check that `is_dtype_masked` returns `True`

Agreed, but I think we should be testing `get_dtype_backend`, since we've exposed the return type in https://github.com/narwhals-dev/narwhals/blob/15c6086e65248f50e09928c9c2144daddcdec3bd/narwhals/typing.py#L244

Probably against a wide range of `pd.api.extensions.ExtensionDtype` and then see if both versions do the same thing? (https://github.com/narwhals-dev/narwhals/pull/2744#discussion_r2172072807), (https://github.com/narwhals-dev/narwhals/pull/2744#discussion_r2172094691)

dangotbanned:
> tbh this sentinel trick looks really brittle, and i'd prefer to go back to checking for `BaseMaskedArray` please, as it was in [#2008 (files)](https://github.com/narwhals-dev/narwhals/pull/2008/files)

Is there any chance this could lead to an update in (https://github.com/pandas-dev/pandas-stubs)? 🙏 

All of this really just stems from the decision to not make the `Dtype` public - which seems strange as the `Array` equivalent is 🤔 

image

MarcoGorelli:
> Agreed, but I think we should be testing get_dtype_backend, since we've exposed the return type in 

yup, sure

tests/get_dtype_backend_test.py
```diff
@@ -0,0 +1,170 @@
...
+    if isinstance(dtype, np.dtype):
+        return pd.core.dtypes.dtypes.NumpyEADtype(dtype)  # type: ignore[attr-defined, no-any-return]
+    else:
+        return dtype
+
+
+# NOTE: Option 2
+def generate_pandas_dtypes_mostly_public(aliases: Iterable[str]) -> list[ExtensionDtype]:
+    """Ideally we'd use this, but `NumpyEADtype` is the ugly duckling."""
+    return [construct_dtype(alias) for alias in aliases]
```
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2744#issuecomment-3013424771

@MarcoGorelli I could definitely do with your opinion on this 😅 

I managed to come up with 2 methods of generating the `Dtype`s - but I don't really like either.
Which of these is the least bad? 😭

narwhals/_pandas_like/utils.py
```diff
@@ -324,28 +325,52 @@ def native_to_narwhals_dtype(
...
+        # `1.1`: https://github.com/pandas-dev/pandas/blob/b5958ee1999e9aead1938c0bba2b674378807b3d/pandas/core/arrays/masked.py#L37
+        # `1.2`: https://github.com/pandas-dev/pandas/blob/7c48ff4409c622c582c56a5702373f726de08e96/pandas/core/arrays/masked.py#L41
+        # `1.5`: https://github.com/pandas-dev/pandas/blob/35b0d1dcadf9d60722c055ee37442dc76a29e64c/pandas/core/dtypes/dtypes.py#L1609
+        if isinstance(dtype, pd.api.extensions.ExtensionDtype):
+            from pandas.core.arrays.masked import (  # type: ignore[attr-defined]
+                BaseMaskedDtype as OldBaseMaskedDtype,  # pyright: ignore[reportAttributeAccessIssue]
+            )
+
+            return isinstance(dtype, OldBaseMaskedDtype)
+        return False
```
dangotbanned, line 353: So this wasn't covered with either of the two methods in (https://github.com/narwhals-dev/narwhals/pull/2008/files)

My version would return `False` because `BaseMaskedDtype.base` wasn't added until `pandas>=1.2.*`

The older version would return `False` because the suppressed `AttributeError` was trying to say

- new `pd.core.dtypes.dtypes.BaseMaskedDtype`
- old `pd.core.arrays.masked.BaseMaskedDtype`

dangotbanned: In either case, we need to branch on the `pandas` version here to avoid https://github.com/narwhals-dev/narwhals/actions/runs/15935001958/job/44952753166?pr=2744

dangotbanned:
FWIW, this PR really wasn't what I set out to do in (https://github.com/narwhals-dev/narwhals/pull/2744/commits/2e382405abf601b05b72c5169e2fa30b49ad0830) and it hasn't been much fun 😞 

I'm happy for anyone to take this off my hands 🙏

MarcoGorelli (approved):
thanks for looking into this so carefully

sure, happy to ship this if you're happy with it, I think the comments you've left are clear enough that we can change this in the future if needed

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2744#pullrequestreview-2968460496

Thanks for the review @MarcoGorelli 😍 

The test code has changed *a little bit* since (https://github.com/narwhals-dev/narwhals/pull/2744#discussion_r2172665738), but I still think we should only merge **one** of the options I gave.

<details><summary><b>Option 1</b> (Currently in use)</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/85d551e74d4b17dfb046e39e5c8461f3d96919ba/tests/get_dtype_backend_test.py#L102-L137

</p>
</details> 

<details><summary><b>Option 2</b> (Not used)</summary>
<p>

I may need to add some extra error handling for this on earlier versions

https://github.com/narwhals-dev/narwhals/blob/85d551e74d4b17dfb046e39e5c8461f3d96919ba/tests/get_dtype_backend_test.py#L140-L153

</p>
</details> 

Essentially, **one** of those two would be the `Dtype` source for the `pandas_dtype` fixture

https://github.com/narwhals-dev/narwhals/blob/85d551e74d4b17dfb046e39e5c8461f3d96919ba/tests/get_dtype_backend_test.py#L156-L165

PR #2747, opened 2025-06-27
feat: Add `Expr|Series.str.to_date`

### LINKED ISSUE #2728: [Enh]: Ability to convert string to date

There is ability to convert string into datetime, however, to date is missing, request is to add this feature.

Current work around: 
`nw.col("my_col").str.to_datetime(format="%Y%m").cast(nw.Date)`

To simplify the conversion from string to date without the use of datetime.

_No response_

`nw.col("my_col").str.to_datetime(format="%Y%m").cast(nw.Date)`

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2728 

## Checklist

- [x] Tests added
- [x] Documented the changes

Whenever some `to_date`-like functionality exists, I fallback to `to_datetime(fmt).dt.date()`


narwhals/_dask/expr_str.py
```diff
@@ -101,3 +103,5 @@ def zfill(self, width: int) -> DaskExpr:
         return self._compliant_expr._with_callable(
             lambda expr, width: expr.str.zfill(width), "zfill", width=width
         )
+
+    to_date = not_implemented()
```
FBruzzesi, line 106: I am investigating this - I implemented it as `.str.to_datetime(format).dt.date()`

However, the output dtype is String, probably because `.str.to_datetime` does not return a pyarrow backed datetime.

narwhals/_duckdb/expr_str.py
```diff
@@ -88,6 +88,9 @@ def to_datetime(self, format: str | None) -> DuckDBExpr:
             lambda expr: F("strptime", expr, lit(format))
         )
 
+    def to_date(self, format: str | None) -> DuckDBExpr:
+        return self.to_datetime(format=format).dt.date()
```
MarcoGorelli: if `format` is `None` you can use `.cast('date')`
```
In [30]: rel
Out[30]:
┌────────────┐
│     a      │
│  varchar   │
├────────────┤
│ 2020-01-01 │
└────────────┘

In [31]: rel.select(duckdb.ColumnExpression('a').cast('date'))
Out[31]:
┌─────────────────┐
│ CAST(a AS DATE) │
│      date       │
├─────────────────┤
│ 2020-01-01      │
└─────────────────┘
```

MarcoGorelli (commented):
nice, thanks!

narwhals/_duckdb/expr_str.py
MarcoGorelli: in fact, we should do that for `to_datetime` too (cast to timestamp)

FBruzzesi: Just committed the change to cast to date to keep it in scope with this PR

dangotbanned:
FYI, the typing issues are from #2756

Should just need an extra `to_date: Method[PolarsExpr]`

Oh and you could trim some lines by replacing:

```py
self._compliant_expr
self.compliant
```

PR #2749, opened 2025-06-28
chore: Remove pandas-like specific `check_column_names_are_unique` util

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Spotted by @dangotbanned, see [comment](https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2173209870)

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -1573,7 +1573,12 @@ def check_columns_exist(
 
 
 def check_column_names_are_unique(columns: Collection[str]) -> None:
-    len_unique_columns = len(set(columns))
+    try:
+        len_unique_columns = len(set(columns))
+    except TypeError as exc:  # pragma: no cover
+        msg = f"Expected hashable (e.g. str or int) column names, got: {columns}"
+        raise TypeError(msg) from exc
+
     if len(columns) != len_unique_columns:
         from collections import Counter
 
--- a/narwhals/_pandas_like/dataframe.py
+++ b/narwhals/_pandas_like/dataframe.py
@@ -10,7 +10,6 @@ from narwhals._compliant import EagerDataFrame
 from narwhals._pandas_like.series import PANDAS_TO_NUMPY_DTYPE_MISSING, PandasLikeSeries
 from narwhals._pandas_like.utils import (
     align_and_extract_native,
-    check_column_names_are_unique,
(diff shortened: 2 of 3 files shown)


narwhals/_utils.py
```diff
@@ -1573,7 +1573,12 @@ def check_columns_exist(
 
 
 def check_column_names_are_unique(columns: Collection[str]) -> None:
-    len_unique_columns = len(set(columns))
+    try:
+        len_unique_columns = len(set(columns))
+    except TypeError as exc:  # pragma: no cover
```
FBruzzesi, line 1578: Set on unhashable type raises a `TypeError`:

```py
set(([1], [2]))
Cell In[1], line 1
----> 1 set(([1], [2]))

TypeError: unhashable type: 'list'
```

Should _we_ raise a `ValueError` as before instead?

dangotbanned: Could you give an example of how we'd get columns like that in the places the function is currently used?

```py
import pandas as pd

>>> pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=([1], [2], [3]))
TypeError: unhashable type: 'list'
```

> AFAIK, `pandas` only supports `Hashable` column "names" - so I'm a little confused 🤔

We're checking an **existing** `NativeFrame.columns` - so I would have thought that condition is not reachable

MarcoGorelli (approved):

narwhals/_utils.py
MarcoGorelli: it is, there's a test which gets there

dangotbanned: > it is, there's a test which gets there

Is there?

Why do we have a `# pragma: no cover`?

dangotbanned: If this is a `pandas` version-specific thing - then a comment explaining the `no cover` would help

I made the suggestion in (https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2173209870) based on the fact this appears unreachable

Issue #2197, opened 2025-03-13
Different error-ing behaviour with `native_namespace=pd` compared to the orginal pandas?


StefanieSenger:
I am not sure if I do something wrong or if this is on purpose, but it seems that narwhals has a different error-ing behaviour with `native_namespace=pd` as pandas itself (at least in my little use case):

I also wonder why narwhals doesn't raise their own errors, so that libraries using narwhals don't need to check for all the possible error types and messages in their tests?

Reproducible:

```python
import narwhals.stable.v1 as nw
import pandas as pd
```

```python
# using pandas:
first = pd.Series([1, 2, 3], name="first")
second = pd.Series([1, 2, 3, 4], name="second")

df = pd.DataFrame(first)
df["second"] = second 
```

This does not error: it aligns itself
`print(df)`

>    first  second
> 0      1       1
> 1      2       2
> 2      3       3

```python
# using narwhals:
first = nw.new_series(name="first", values=[1,2,3], native_namespace=pd)
second = nw.new_series(name="second", values=[1,2,3,4], native_namespace=pd)

dict = nw.from_dict({"first": first}, backend=pd)
dict = dict.with_columns(**{"second": second})
```

This raises a pandas error:

>  ValueError: Length mismatch: Expected axis has 4 elements, new values have 3 elements

Extract from Traceback:

```
fairlearn/reductions/_moments/moment.py:63: in load_data
    self.tags = self.tags.with_columns(**{_GROUP_ID: sensitive_features})
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/dataframe.py:1286: in with_columns
    return super().with_columns(*exprs, **named_exprs)
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/dataframe.py:146: in with_columns
    self._compliant_frame.with_columns(*compliant_exprs),
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/_pandas_like/dataframe.py:488: in with_columns
    to_concat.extend(
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/_pandas_like/dataframe.py:489: in <genexpr>
    extract_dataframe_comparand(index, new_column_name_to_new_column_map[s])
[...]

PR #2752, opened 2025-06-28
enh!: simplify exceptions, consistently raise InvalidOperationError instead of ShapeError/OrderDependentExprError/LengthChangingExprError

### LINKED ISSUE #2751: api: simplify narwhals exceptions
i'm a bit concerned by how many exceptions we have, and would like that refactors don't end up changing which exception gets raised

i think we can simplify a lot of them to just be `InvalidOperationError`

MarcoGorelli:
closes #2751 

I've done a GitHub search, and there's no public repos using these exceptions

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


dangotbanned:
I'm mostly on board here - but could we keep `ShapeError`?


> i'm a bit concerned by how many exceptions we have, and would like that refactors don't end up changing which exception gets raised

That one is inherited from `polars` and I think will be safe from refactors - considering the situations it is raised in

MarcoGorelli:
we might change the order in which we check expressions' attributes which might mean that we no longer raise `ShapeError`

tbh i think this one's too finely scoped in polars too

dangotbanned:
> we might change the order in which we check expressions' attributes which might mean that we no longer raise `ShapeError`
> 
> tbh i think this one's too finely scoped in polars too

How about if we restricted it to only the errors that look like this?

```py
msg = f"Expected object of length {len_idx}, got: {len_other}."
raise ShapeError(msg)
```

Those are the ones that I'm expecting won't change, and they're focused on `Series`.

Generally I was thinking about errors like (https://github.com/narwhals-dev/narwhals/issues/2197#issuecomment-2721195705)

MarcoGorelli:
Seems reasonable, thanks!

narwhals/functions.py
```diff
@@ -1475,7 +1475,7 @@ def then(self, value: IntoExpr | NonNestedLiteral | _1DArray) -> Then:
                 "If you pass a scalar-like predicate to `nw.when`, then "
                 "the `then` value must also be scalar-like."
             )
-            raise ShapeError(msg)
+            raise InvalidOperationError(msg)
```
dangotbanned, line 1478: Maybe this could be something like:

```py
raise InvalidOperationError.from_then(...)
```

And then one a bit lower as:

```py
raise InvalidOperationError.from_otherwise(...)
```

They use the same message, just with one word changed
```py
"the {method} value must also ..."
```

tests/frame/group_by_test.py
```diff
@@ -471,21 +466,18 @@ def test_group_by_expr(
...
         ),  # Transform and Filtration
         (
             [nw.col("a").alias("foo"), nw.col("a").max()],
             pytest.raises(ComputeError),
         ),  # Transform and Aggregation
         (
             [nw.col("a").alias("foo"), nw.col("a").cum_max()],
-            pytest.raises(OrderDependentExprError),
+            pytest.raises(InvalidOperationError),
         ),  # Transform and Window
```
dangotbanned, line 481: I think these comments are based on the old `ExprKind`:

https://github.com/narwhals-dev/narwhals/blob/89448231342acda3358efea20419203e6ea35c56/narwhals/_expression_parsing.py#L117-L150


The current version is a bit different:
https://github.com/narwhals-dev/narwhals/blob/f29fa5f03a0e1c01aad313d849611b79474ad6a2/narwhals/_expression_parsing.py#L121-L155

dangotbanned (approved):
Thanks @MarcoGorelli!

Seems like a nice simplification - despite the [extra work I need to do now](https://github.com/narwhals-dev/narwhals/blob/a4d90d88a16dc6d98e55a33879e548de8c58e215/tests/plan/expr_parsing_test.py) 😉 

I've left some *non-blocking* comments, but looks great otherwise

MarcoGorelli:
thanks @dangotbanned for your review!

tests/frame/group_by_test.py
MarcoGorelli: sure, no objections to updating this (i'd actually updated the comments locally, pushed, then merged, but hadn't noticed the push hadn't gone through due to the merge commit here)

Issue #2722, opened 2025-06-22
[Enh]: Support `int_range`


dangotbanned:
## Related
- https://github.com/narwhals-dev/narwhals/issues/2307#issuecomment-2832422364
- https://github.com/narwhals-dev/narwhals/issues/2307#issuecomment-2849307655
- #2683

## Description
[`polars.int_range`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.int_range.html) was considered in (#2307) but ultimately wasn't the right fit.

I still think it's got a lot of utility on it's own and could be a model for a future [`date_range`](https://github.com/narwhals-dev/narwhals/issues/2307#issuecomment-2832454886) 

## Eager Support

If you stare at (#2683) close enough - we've got most of it implemented for **eager** backends already.
We'd just need to expose `start`, `step` - where all we've used is `stop`/`end`:

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_arrow/dataframe.py#L491

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_pandas_like/dataframe.py#L426-L433

https://github.com/narwhals-dev/narwhals/blob/84a1d817084aa42df68df9e1693ff60c7d26e209/narwhals/_polars/dataframe.py#L249

## Lazy Support
`polars` is a given, but I mentioned `ibis.range` in (https://github.com/narwhals-dev/narwhals/issues/2307#issuecomment-2855235654)

https://github.com/ibis-project/ibis/blob/58dea60ad1c3f2f4b9288df703129d7aa503e1be/ibis/expr/api.py#L2282-L2420

I still need to do a deep-dive on that, but it has all the features we'd need for lazy support:

```py
import ibis

>>> ibis.range(5, 600, 46)
IntegerRange(5, 600, 46): IntegerRange(start=5, stop=600, step=46)

>>> ibis.options.interactive = True
>>> ibis.range(5, 600, 46)
┌──────────────────┐
│ [5, 51, ... +11] │
└──────────────────┘
```

`duckdb` [seems to support ranges](https://duckdb.org/docs/stable/sql/functions/list.html#range-functions), with two variants:
- [`range`](https://duckdb.org/docs/stable/sql/functions/list.html#range)
- [`generate_series`](https://duckdb.org/docs/stable/sql/functions/list.html#generate_series)

[...]

PR #2757, opened 2025-06-29
chore: Add pandas-like `import_array_module` util function

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Similar to spark-like `import_functions`

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -14,6 +14,7 @@ from narwhals._pandas_like.series_struct import PandasLikeSeriesStructNamespace
 from narwhals._pandas_like.utils import (
     align_and_extract_native,
     get_dtype_backend,
+    import_array_module,
     narwhals_to_native_dtype,
     native_to_narwhals_dtype,
     object_native_to_narwhals_dtype,
@@ -386,14 +387,8 @@ class PandasLikeSeries(EagerSeries[Any]):
     def arg_true(self) -> Self:
         ser = self.native
         size = len(ser)
-        if self._implementation.is_cudf():
-            import cupy as cp  # ignore-banned-import  # cuDF dependency.
-
-            data = cp.arange(size)
-        else:
-            import numpy as np  # ignore-banned-import
-
-            data = np.arange(size)
+        array_funcs = import_array_module(self._implementation)
+        data = array_funcs.arange(size)
         result = ser.__class__(data, name=ser.name, index=ser.index).loc[ser]
         return self._with_native(result)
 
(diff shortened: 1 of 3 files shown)


narwhals/_pandas_like/utils.py
```diff
@@ -149,7 +151,7 @@ def set_index(
     ) != (actual_len := len(obj)):
         msg = f"Expected object of length {expected_len}, got length: {actual_len}"
         raise ShapeError(msg)
-    if implementation is Implementation.CUDF:  # pragma: no cover
+    if implementation is Implementation.CUDF:
```
FBruzzesi, line 154: Covered in pyproject.toml

narwhals/_pandas_like/utils.py
```diff
@@ -623,6 +625,21 @@ def is_non_nullable_boolean(s: PandasLikeSeries) -> bool:
     )
 
 
+def import_array_module(implementation: Implementation, /) -> ModuleType:
```
FBruzzesi, line 628: Should we cache?

dangotbanned:
@FBruzzesi this would only apply to the `arange` cases (I think 2 methods?)
but I was thinking (#2722) might be a more complete alternative for that?

We'd be able to write this, which *I think* is pretty clean 🙂 

```py
        if order_by is None:
            return self.select(plx.int_range(len(self), eager=False), plx.all())
```


> [!NOTE]
> I'm not saying we couldn't use the generalized dispatch under-the-hood in `int_range` though

FBruzzesi:
Hey @dangotbanned, thanks for taking a look.

I was getting annoyed by having to split the logic _again_ in this commit https://github.com/narwhals-dev/narwhals/commit/34cfce9da5811e8cbc7a12ca01056f35e78858a6

which however I rolled back for now 🙈

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2757#issuecomment-3016505128

oh I see, there's more coming!

In that case, would you be able to mirror how the typing works for `_spark_like` please? 🙏 

- `_utils.import_*` is typed mostly like what you have now
- Then we have accessors on the `PandasLike*` classes, which infer the return type
  - Probably using `numpy` where the `_spark_like` ones use `sqlframe`

narwhals/_pandas_like/series.py
```diff
@@ -1087,28 +1072,26 @@ def log(self, base: float) -> Self:
                 self._backend_version,
                 self._version,
             )
-            result_native = native.__class__(
+            result_native = native_cls(
```
dangotbanned, line 1081: We *could* write this as `type(native)(...)`

FBruzzesi: Sure, I will change that! Any reason one is preferable?

narwhals/_pandas_like/utils.py
dangotbanned: I'd say leave out caching for now since there's an exception being raised inside the function

I left some notes and a traceback in (https://github.com/narwhals-dev/narwhals/pull/2572#discussion_r2143443059)

narwhals/_pandas_like/series.py
dangotbanned: I generally prefer `type` as it is public api - although it is unlikely to change 😄 - and is shorter

To turn the tables slightly, would you ever prefer the second option here?

```py
len(obj)
obj.__len__()

iter(obj)
obj.__iter__()

obj_1 + obj_2
obj_1.__add__(obj_2)

obj_1 & obj_2
obj_1.__class__.__and__(obj_1, obj_2)
```

dangotbanned: I feel like the `__dunders__` often get used in a effort to lower performance overhead.
I haven't seen any data to prove that works, but there's way a lot more going on behind the scenes after the first lookup anyway

Not sure if you've seen this series before by a cpython core dev?
https://snarky.ca/unravelling-attribute-access-in-python/

dangotbanned (approved):
Thanks @FBruzzesi 🥳

Issue #2748, opened 2025-06-28
[Enh]: Align floordiv behavior across different backends


FBruzzesi:
As follow up of #2636, we should aim to align floordiv behavior across backends whenever possible

PR #2761, opened 2025-06-29
fix: Align division by zero behavior across all backends

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related #2748 : solved all cases but pandas-like

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/utils.py
+++ b/narwhals/_arrow/utils.py
@@ -258,7 +258,7 @@ def extract_native(
 def floordiv_compat(left: ArrayOrScalar, right: ArrayOrScalar, /) -> Any:
     # The following lines are adapted from pandas' pyarrow implementation.
     # Ref: https://github.com/pandas-dev/pandas/blob/262fcfbffcee5c3116e86a951d8b693f90411e68/pandas/core/arrays/arrow/array.py#L124-L154
-
+    # We modify it to add `safe_mask` so as to align with Polars' behaviour when dividing by zero.
     safe_mask = pc.not_equal(right, lit(0, type=right.type))
     safe_right = pc.if_else(safe_mask, right, lit(1, type=right.type))  # Dummy value
     non_safe_result = lit(None, type=left.type)


FBruzzesi:
@MarcoGorelli what's the current status of pandas-like broadcasting? 

I am having a very bad time having to distinguish between python scalar vs  `nw.lit(value)` and deal with all the situations that might arise.

As an example, we would _enter_ `__floordiv__` in all the following cases:

```py
frame.<with_columns|select>(
    x1 = nw.lit(value) / scalar,
    x2 = nw.lit(value) / nw.lit(another_value),
    x3 = nw.lit(value) / nw.col(name),
    x4 = nw.col(name) / scalar,
    x5 = nw.col(name) / nw.lit(value),
    x6 = nw.col(name) / nw.col(another_name),
)
```

and they need special attentions 😂

On top of that, there is specific logic given the dtype backend 🙈

narwhals/_dask/expr.py
```diff
@@ -209,13 +210,57 @@ def __rtruediv__(self, other: Any) -> Self:
...
+        ) -> dx.Series:
+            series, other = align_series_full_broadcast(df, series, other)
+            return (other.__floordiv__(series)).where(series != 0, None)
+
+        def func(df: DaskLazyFrame) -> list[dx.Series]:
+            if isinstance(other, type(self)):
+                if len(other_ := other(df)) > 1:  # pragma: no cover
+                    msg = "Expected expression with single output, found multiple"
+                    raise ValueError(msg)
+                other_series = other_[0]
```
FBruzzesi: Is this case even possible? It shouldn't be, right? As in if other is also a DaskExpr, then it would trigger `other.__floordiv__` ops.

FBruzzesi:
@MarcoGorelli I rolled backed the "not-fully functional" idea I had for pandas.
This PR would end up aligning the behavior for pyarrow and the lazy backends, which I still think it's a good achievement.

I do see tests failing for "not so old" polars due to PanicExpection's and different results of `__rfloordiv__`. I need to check until which version the issue exists (and either fix it or xfail/skip it).

Finally, for the missing coverage, the https://github.com/narwhals-dev/narwhals/pull/2761#discussion_r2183743937 above is relevant, and I would like to make sure that I am not missing anything

narwhals/_duckdb/expr.py
```diff
@@ -99,6 +99,18 @@ def __narwhals_namespace__(self) -> DuckDBNamespace:  # pragma: no cover
 
         return DuckDBNamespace(version=self._version)
 
+    def __floordiv__(self, other: Any) -> Self:
+        def func(expr: Expression, other: Expression) -> Expression:
+            return when(other != lit(0), expr // other).otherwise(lit(None))
+
+        return self._with_binary(func, other=other)
```
MarcoGorelli: I think this may need you to define `window_func`, like we do for `Expr.sum` - otherwise would `nw.col('a') / nw.col('a').sum()` work?

(ok to keep as a follow-up, but if it doesn't work, we should track it)

narwhals/_dask/expr.py
MarcoGorelli: tbh i can't remember off the top of my head, will try to take a look next week

narwhals/_duckdb/expr.py
FBruzzesi: I am not sure I follow the logic - divisions are never an aggregation.
The example you shared works fine for both true and floor div:

```py
import duckdb
import narwhals as nw
import polars as pl

_df = pl.DataFrame({"a": [2, 0, -2]})
rel = duckdb.table("_df")

(
    nw.from_native(rel)
    .select(
        truediv=nw.col("a")/nw.col("a").sum(),
        floordiv=nw.col("a")//nw.col("a").sum(),
    )
)

┌──────────────────────┐
|  Narwhals LazyFrame  |
|----------------------|
|┌─────────┬──────────┐|
|│ truediv │ floordiv │|
|│ double  │  int128  │|
|├─────────┼──────────┤|
|│     inf │     NULL │|
|│     nan │     NULL │|
|│    -inf │     NULL │|
|└─────────┴──────────┘|
└──────────────────────┘
```

MarcoGorelli:
thanks - will take a closer look, i'd like to be quite careful with this one 😄

narwhals/_arrow/utils.py
```diff
@@ -260,8 +260,14 @@ def floordiv_compat(left: ArrayOrScalar, right: ArrayOrScalar, /) -> Any:
     # The following lines are adapted from pandas' pyarrow implementation.
     # Ref: https://github.com/pandas-dev/pandas/blob/262fcfbffcee5c3116e86a951d8b693f90411e68/pandas/core/arrays/arrow/array.py#L124-L154
 
+    safe_mask = pc.not_equal(right, lit(0, type=right.type))
```
MarcoGorelli, line 262: shall we just add a comment noting what we've changed from pandas' pyarrow impl?

narwhals/_dask/expr.py
```diff
@@ -209,13 +210,49 @@ def __rtruediv__(self, other: Any) -> Self:
...
+            return (series.__floordiv__(other)).where(other != 0, None)
+
+        def func(df: DaskLazyFrame) -> list[dx.Series]:
+            if isinstance(other, type(self)):
+                if len(other_ := other(df)) > 1:  # pragma: no cover
+                    msg = "Expected expression with single output, found multiple"
+                    raise ValueError(msg)
+                other_series = other_[0]
+            else:
+                other_series = other
```
MarcoGorelli: 🤔 this looks right, but imma have to think about it, not sure about repeating the broadcasting logic here..

narwhals/_sql/expr.py
```diff
@@ -178,6 +178,8 @@ def _coalesce(self, *expr: NativeExprT) -> NativeExprT:
 
     def _count_star(self) -> NativeExprT: ...
     def _when(self, condition: NativeExprT, value: NativeExprT) -> NativeExprT: ...
+    def __floordiv__(self, other: Any) -> Self: ...
+    def __rfloordiv__(self, other: Any) -> Self: ...
```
MarcoGorelli: the implementation looks quite similar between duckdb and pyspark, wonder if the implementation can be shared here?

FBruzzesi: Disclaimer: I didn't invest more than 3 minutes looking at this

---

Yes that's true, but there is a limitation: the new abstracted `self._when` does not allow for chaining with `.otherwise`. Understandably so, for ibis it would be complex since it's _inside_ the `ibis.cases` rather than a chained ops

MarcoGorelli: it should be possible by putting statemnts within each other, like here

https://github.com/narwhals-dev/narwhals/blob/c22b58210bdd41fda8134ac97c7065275a28210b/narwhals/_sql/expr.py#L515-L528

FBruzzesi: With the latest changes in `when`, I am able to just write:

```py
def __floordiv__(self, other: Self) -> Self:
    
    def func(expr: NativeExprT, other: NativeExprT) -> NativeExprT:
        return self._when(other != self._lit(0), expr // other, self._lit(None))
    
    return self._with_binary(func, other=other)

def __rfloordiv__(self, other: Self) -> Self:
    
    def func(expr: NativeExprT, other: NativeExprT) -> NativeExprT:
        return self._when(expr != self._lit(0), other // expr, self._lit(None))
     
    return self._with_binary(func, other=other).alias("literal")
```

at the `_sql/` level. This would work for ibis and duckdb, but not for pyspark since we have a different dedicated implementation.

However, this results in all sort of typing issues. At sql level says that `NativeExprT` does not implement floordiv and rfloordiv. If I go add them these methods than it's even worse.

Honestly it's a 10 lines difference, but we should probably start doing some workaround for the typing.. Personally I don't think it's worth it

MarcoGorelli: You can use `op.floordiv` / `op.truediv` in `_sql/expr`, that'll resolve the typing issues

Personally I'd have preferred to implement arithmetic ops for `NativeExpr` and ignore Ibis' half-baked typing, but @dangotbanned seemed quite keen on typing Ibis, which seems fair enough

MarcoGorelli:
thanks for updating! generally on-board, though still hoping for a simpler dask way, will try something out

narwhals/_arrow/utils.py
MarcoGorelli: gentle ping

MarcoGorelli (changes requested):
one comment

for Dask, can we just use
```
other_series = maybe_evaluate_expr(df, other)
```
instead of lines 238-244?

PR #2765, opened 2025-06-30
chore: Move `_reuse_series` to complaint level as much as possible

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

While doing something else completely unrelated, I spotted that:

- We are not exploiting `_reuse_series` as much as we could
- We are improperly using or rather not using `scalar_kwargs` when we should. Most likely negligible, but using it properly avoids an extra function call (`df._evaluate_expr(value) if self._is_expr(value) else value`)

### DIFF
--- a/narwhals/_compliant/typing.py
+++ b/narwhals/_compliant/typing.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 from typing import TYPE_CHECKING, Any, Callable, TypedDict, TypeVar
 
 if TYPE_CHECKING:
@@ -21,21 +21,47 @@ if TYPE_CHECKING:
     from narwhals._compliant.namespace import CompliantNamespace, EagerNamespace
     from narwhals._compliant.series import CompliantSeries, EagerSeries
     from narwhals._compliant.window import WindowInputs
-    from narwhals.typing import FillNullStrategy, NativeFrame, NativeSeries, RankMethod
+    from narwhals.typing import (
+        FillNullStrategy,
+        IntoDType,
+        NativeFrame,
+        NativeSeries,
+        RankMethod,
+        RollingInterpolationMethod,
+    )
 
     class ScalarKwargs(TypedDict, total=False):
         """Non-expressifiable args which we may need to reuse in `agg` or `over`."""
 
+        adjust: bool
+        alpha: float | None
+        base: float
         center: int
+        com: float | None
         ddof: int
+        decimals: int
         descending: bool
+        fraction: float | None
+        half_life: float | None
+        ignore_nulls: bool
+        interpolation: RollingInterpolationMethod
         limit: int | None
+        maintain_order: bool
         method: RankMethod
         min_samples: int
-        n: int
+        n: int | None
+        new: Sequence[Any]
+        nulls_last: bool
+        offset: int
+        old: Sequence[Any] | Mapping[Any, Any]
+        quantile: float
+        return_dtype: IntoDType | None
         reverse: bool
+        seed: int | None
+        span: float | None
         strategy: FillNullStrategy | None
         window_size: int
+        with_replacement: bool
 
 
 __all__ = [
--- a/narwhals/_arrow/expr.py
+++ b/narwhals/_arrow/expr.py
@@ -23,7 +23,6 @@ if TYPE_CHECKING:
     from narwhals._compliant.typing import AliasNames, EvalNames, EvalSeries, ScalarKwargs
     from narwhals._expression_parsing import ExprMetadata
     from narwhals._utils import Version, _FullContext
-    from narwhals.typing import RankMethod
 
 
 class ArrowExpr(EagerExpr["ArrowDataFrame", ArrowSeries]):
@@ -122,12 +121,6 @@ class ArrowExpr(EagerExpr["ArrowDataFrame", ArrowSeries]):
     ) -> dict[str, Any]:
         return {"_return_py_scalar": False} if returns_scalar else {}
 
-    def cum_sum(self, *, reverse: bool) -> Self:
-        return self._reuse_series("cum_sum", reverse=reverse)
-
-    def shift(self, n: int) -> Self:
-        return self._reuse_series("shift", n=n)
-
     def over(self, partition_by: Sequence[str], order_by: Sequence[str]) -> Self:
         if (
             partition_by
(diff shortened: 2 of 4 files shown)


narwhals/_compliant/typing.py
```diff
@@ -21,21 +21,47 @@
...
+        offset: int
+        old: Sequence[Any] | Mapping[Any, Any]
+        quantile: float
+        return_dtype: IntoDType | None
         reverse: bool
+        seed: int | None
+        span: float | None
         strategy: FillNullStrategy | None
         window_size: int
+        with_replacement: bool
```
FBruzzesi: Well I know this is a lot, but that's what we use 😂

FBruzzesi:
Passing the type checker GHA first try 

PatOnTheBackGoMeGIF

narwhals/_compliant/typing.py
FBruzzesi: I wonder if it's better to split into many or at least a few classes to finally combine into one:

```py
class _RollingKwargs(TypedDict, total=False):
    center: int
    ddof: int
    min_samples: int
    window_size: int

class _EWMKwargs(TypedDict, total=False):
    adjust: bool
    alpha: float | None
    com: float | None
    half_life: float | None
    ignore_nulls: bool
    min_samples: int
    span: float | None

class _FillNullKwargs(TypedDict, total=False):
    limit: int | None
    strategy: FillNullStrategy | None

class _ReplaceStrictKwargs(TypedDict, total=False):
    new: Sequence[Any]
    old: Sequence[Any] | Mapping[Any, Any]
    return_dtype: IntoDType | None

class _SampleKwargs(TypedDict, total=False):
    n: int | None
    fraction: float | None
    seed: int | None
    with_replacement: bool

class _RankKwargs(TypedDict, total=False):
    method: RankMethod
    descending: bool

class _QuantileKwargs(TypedDict, total=False):
    interpolation: RollingInterpolationMethod
    quantile: float

class _SortKwargs(TypedDict, total=False):
    descending: bool
    nulls_last: bool

class ScalarKwargs(
    _RollingKwargs,
    _EWMKwargs,
    _FillNullKwargs,
    _ReplaceStrictKwargs,
    _SampleKwargs,
    _RankKwargs,
    _QuantileKwargs,
    _SortKwargs,
    total=False,
):
    """Non-expressifiable args which we may need to reuse in `agg` or `over`."""

    base: float
    decimals: int
    maintain_order: bool
    offset: int
    reverse: bool
```

---

@dangotbanned what are your thought in here?
I don't _really_ like any of the two 😂

dangotbanned: > """Non-expressifiable args which we may need to reuse in `agg` or `over`."""

@FBruzzesi do we need any of these in `agg` or `over`?

I have a feeling that changing where these args are stored will mean this wouldn't be safe anymore

https://github.com/narwhals-dev/narwhals/blob/cf5285c6685d6bc31a841492324d9efe240219f2/narwhals/_pandas_like/group_by.py#L78-L88

dangotbanned:
@FBruzzesi unless you feel strongly about this change, my suggestion would be to hold off on this for now.

IMO the real performance wins come from replacing all of this with a model that has a better separation of what needs parsing as `Expr`.
In (#2572), the `expressifiable_args` and `scalar_kwargs` are stored in different layers - and both are immutable so they can be cached safely

https://github.com/narwhals-dev/narwhals/blob/5c80033a6c923f3f36648dabba911cc47d791bb1/narwhals/_plan/dummy.py#L250-L295

In that model, everything is always separated and anything that needs parsing is done immediately - due to needing to fit into an exact shape like:

```py
    # De-sugared
    def rank(self, method: RankMethod = "average", *, descending: bool = False) -> Self:
        rank = F.Rank(options=RankOptions(method=method, descending=descending))
        ir = expr.FunctionExpr(
            input=(self._ir,), function=rank, options=rank.function_options
        )
        return DummyExpr._from_ir(ir)
```

There's definitely more we can do to improve/simplify `EagerExpr` - and I don't want to dissuade anyone from trying new things out!
This is just me realising I need to start shouting about (#2572) when I see a similar problem being addressed 😅

narwhals/_compliant/typing.py
FBruzzesi: Mmmh I see, I think there are two points to be made here:

- There are cases that can end up in `**kwds` that we are not currently capturing (e.g. quantile kwargs)
- There are cases that definitly would not end up in a group by context

I wonder if it makes sense to split it into two dicts:

- One for the kwargs of `NativeAggregation`
- One for the remaining

and then have:

```diff
    def _reuse_series(
        self,
        method_name: str,
        *,
        returns_scalar: bool = False,
-         scalar_kwargs: ScalarKwargs | None = None,
+         scalar_kwargs: ReturnScalarKwargs | OtherScalarKwargs | None = None,
        **expressifiable_args: Any,
    ) -> Self:
```

---

For[other comment](https://github.com/narwhals-dev/narwhals/pull/2765#issuecomment-3023552316), the answer is also two folded:

- Regarding moving `_reuse_series` at compliant level, then yes, I don't see why we shouldn't
- Regarding the `ScalarKwargs`, I don't feel particularly strong on this, but we are somewhat misusing our own API, aren't we?

dangotbanned: I've just had a look back at (https://github.com/narwhals-dev/narwhals/pull/2149/files#r1986103446) when I moved everything into `EagerExpr`.
It seems like the only change since then has been (#2555)

I think the difference between `(Arrow|PandasLike)Expr` is essentially the difference between what they support in `over()`

For example, `cum_sum` is using `scalar_kwargs` as documented for `pandas` - but there isn't an equivalent for `pyarrow` to use

https://github.com/narwhals-dev/narwhals/blob/cea997fb5aac0867e8ba261c6a97153b7be3f952/narwhals/_arrow/expr.py#L125-L126

https://github.com/narwhals-dev/narwhals/blob/cea997fb5aac0867e8ba261c6a97153b7be3f952/narwhals/_pandas_like/expr.py#L199-L200

https://github.com/narwhals-dev/narwhals/blob/cea997fb5aac0867e8ba261c6a97153b7be3f952/narwhals/_pandas_like/expr.py#L252-L254

I guess the *"may"* is doing the heavy lifting here 😄 

https://github.com/narwhals-dev/narwhals/blob/cea997fb5aac0867e8ba261c6a97153b7be3f952/narwhals/_compliant/expr.py#L393-L394

FBruzzesi: > I guess the "may" is doing the heavy lifting here 😄

Yes indeed 😂😂 saved on the corner!

I guess that's not used the same way for pyarrow because the `over` support in terms of which ops are allowed is much lower (e.g. `cum_sum` is not supported there, if it was, we would need the argument).

Also, some of the arguments that in principle we could use in group by, are not tracked at the moment (e.g. quantile is the first that comes to mind) - for others I fully agree that the description of `ScalarKwargs` is unrelated.

---

To move forward with this I would propose:

- Keep what was actually missing in `ScalarKwargs`
- Rollback the other addition
- Keep the logic of `_reuse_series` in `EagerExpr`

dangotbanned: > To move forward with this I would propose:
> 
> - Keep what was actually missing in `ScalarKwargs`
> - Rollback the other addition
> - Keep the logic of `_reuse_series` in `EagerExpr`

I *think* I understand what you mean and agree, but it might just be easier for you to push the changes - rather than try explaining (for my sake) 😂

dangotbanned: There is probably still room for improvement in the docs/parameter names for:
- `EagerExpr._reuse_series`
- `EagerExpr._reuse_series_inner`
- `EagerExpr._reuse_series_namespace`

FBruzzesi: > I think I understand what you mean and agree, but it might just be easier for you to push the changes - rather than try explaining (for my sake) 😂

Yes! but... actually I realized that it's something completely new to support, so I will split it into two PRs. For now I will keep `ScalarKwargs` as it was and only moving `_reuse_series` in `EagerExpr`. Will follow up shortly after we merge this with the enhancement I have in mind

MarcoGorelli (approved):
very nice, thanks @FBruzzesi !

PR #2773, opened 2025-07-02
chore: Re-sync `_compliant` exports

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

Generally aiming to have all modules outside of the `_compliant` package using imports like

```py

from narwhals._compliant import ...

if TYPE_CHECKING:
    from narwhals._compliant.typing import ...
```

The exception that remains is `any_namespace`.
I thought it was a bit ambiguous when alongside 2 *unrelated* `Namespace` hierarchies and the `EagerSeries*` stuff:

https://github.com/narwhals-dev/narwhals/blob/da0c1c2f39355e6a8aad45f8dc0be0039a976154/narwhals/_compliant/__init__.py#L13-L42

### DIFF
--- a/narwhals/_compliant/typing.py
+++ b/narwhals/_compliant/typing.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Sequence
-from typing import TYPE_CHECKING, Any, Callable, TypedDict, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Literal, TypedDict, TypeVar
 
 if TYPE_CHECKING:
     from typing_extensions import TypeAlias
@@ -48,6 +48,7 @@ __all__ = [
     "EvalNames",
     "EvalSeries",
     "IntoCompliantExpr",
+    "NarwhalsAggregation",
     "NativeFrameT_co",
     "NativeSeriesT_co",
 ]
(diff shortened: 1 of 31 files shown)


narwhals/_compliant/typing.py
```diff
@@ -154,3 +155,16 @@ class ScalarKwargs(TypedDict, total=False):
     "Callable[[CompliantFrameT, WindowInputs[NativeExprT]], Sequence[NativeExprT]]"
 )
 """A function evaluated with `over(partition_by=..., order_by=...)`."""
+
+NarwhalsAggregation: TypeAlias = Literal[
+    "sum", "mean", "median", "max", "min", "std", "var", "len", "n_unique", "count"
+]
+"""`Expr` methods we aim to support in `DepthTrackingGroupBy`.
+
+Be sure to update me if you're working on one of these:
```
FBruzzesi, line 164: 😂 I am sure the type checker would complain anyway but good to mention issues as well

FBruzzesi (approved):
I couldn't spot any issue, and always happy to lower mental effort/load when moving across the codebase! Thanks @dangotbanned 🚀

WooblesTheWooblesGIF

PR #1702, opened 2025-01-01
feat: show native object (if possible) in repr

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2776, opened 2025-07-03
feat: Use `_repr_html_` when native supports it

dangotbanned:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related #1702
- https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display
- https://github.com/pandas-dev/pandas/blob/22f12fc5d3f7fda3f198760204e7c13150c78581/pandas/core/frame.py#L1189-L1232
- https://github.com/pola-rs/polars/blob/8011fa34e0c5f1270ef52e2d3b0b2946bb2faa72/py-polars/polars/dataframe/frame.py#L1580-L1605

## Checklist

- [ ] Tests added
- [ ] Documented the changes

I discovered this method in (#2572), when I was trying to work out why `polars.Expr` looked so much better that what I had 😅

Thinking we can get more immediate benefits *now* by allowing this option when a backend supports it for `DataFrame`, `LazyFrame`, `Series`.

(screenshot)

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -1562,6 +1562,41 @@ def generate_repr(header: str, native_repr: str) -> str:
     )
 
 
+def generate_repr_html(header: str, native_html: str) -> str | None:
+    import io
+    import xml.etree.ElementTree as ET
+
+    style_css = (
+        ".dataframe caption { "
+        "caption-side: bottom; "
+        "text-align: center; "
+        "font-weight: bold; "
+        "padding-top: 8px;"
+        "}"
+    )
+    try:
+        tree = ET.parse(io.StringIO(native_html.replace("<style scoped>", "<style>")))  # noqa: S314
+    except (SyntaxError, TypeError):
+        return None
+    table = tree.find("table")
+    if table is None:
+        return None
+    caption = ET.Element("caption")
+    caption.text = header
+    table.insert(0, caption)
+    style = tree.find("style")
+    if style is not None:
+        text_existing = style.text or ""
+        style.text = f"{text_existing}\n{style_css}"
+    else:
+        style = ET.Element("style")
+        style.text = style_css
+        tree.getroot().insert(0, style)
+    buf = io.BytesIO()
+    tree.write(buf, "utf-8", method="html")
+    return buf.getvalue().decode()
+
+
 def check_columns_exist(
     subset: Collection[str], /, *, available: Collection[str]
 ) -> ColumnNotFoundError | None:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -22,9 +22,11 @@ from narwhals._expression_parsing import (
 )
 from narwhals._utils import (
     Implementation,
+    _hasattr_static,
     find_stacklevel,
     flatten,
     generate_repr,
+    generate_repr_html,
     is_compliant_dataframe,
     is_compliant_lazyframe,
     is_index_selector,
@@ -489,6 +491,12 @@ class DataFrame(BaseFrame[DataFrameT]):
 
     def __repr__(self) -> str:  # pragma: no cover
         return generate_repr("Narwhals DataFrame", self.to_native().__repr__())
+
+    def _repr_html_(self) -> str | None:  # pragma: no cover
+        native: Any = self.to_native()
+        if _hasattr_static(native, "_repr_html_") and (html := native._repr_html_()):
+            return generate_repr_html("Narwhals DataFrame", html)
+        return None
 
     def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
         """Export a DataFrame via the Arrow PyCapsule Interface.


narwhals/_utils.py
```diff
@@ -1579,6 +1579,41 @@ def generate_repr(header: str, native_repr: str) -> str:
...
+    import xml.etree.ElementTree as ET
+
+    style_css = (
+        ".dataframe caption { "
+        "caption-side: bottom; "
+        "text-align: center; "
+        "font-weight: bold; "
+        "padding-top: 8px;"
+        "}"
+    )
```
dangotbanned: If anyone has any suggestions for styling - feel free to experiment/comment 🙂 

The only decision I'd made so far was putting the [`<caption>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/caption) **below** the table

With the default `polars` formatting, it appeared between the table and the shape tuple when **above** - which I thought looked odd

image

narwhals/_utils.py
```diff
@@ -1579,6 +1579,47 @@ def generate_repr(header: str, native_repr: str) -> str:
     )
 
 
+# NOTE: Unsure on how to test this reliably
+def generate_repr_html(
+    header: Literal["Narwhals DataFrame", "Narwhals LazyFrame"], native_html: str
+) -> str | None:  # pragma: no cover
+    if header == "Narwhals LazyFrame" and "LazyFrame" in native_html:
+        html = native_html.replace("LazyFrame", "LazyFrame.to_native()")
+        return f"{html}<p><b>{header}</b></p>"
```
dangotbanned: Had to add this branch for `pl.LazyFrame` as it wasn't parsing with my naive wrapper:


```py
import io
import xml.etree.ElementTree as ET

import polars as pl

data = {"a": [1, 2, 3], "b": ["fdaf", "fda", "cf"]}
ldf = pl.LazyFrame(data)

>>> ET.parse(io.StringIO(ldf._repr_html_()))
ParseError: junk after document element: line 1, column 25
```

Seems to fail on the first `<p>` in https://github.com/pola-rs/polars/blob/dfa5efe71156c654a1ba3a54b865eae723a818e9/py-polars/polars/lazyframe/frame.py#L783

dangotbanned:
## Possible follow-ups
Just some loose ideas, nothing I'm planning to work on any time soon 😅 

- Support `dask.dataframe.DataFrame._repr_html_`
  It has one, but it is a bit strange and I wasn't able to parse it (https://github.com/narwhals-dev/narwhals/pull/2776/commits/3769326ec3e4435f8e3815b70996f70e384e6678)
image
- Normalize the styling
  - The code in `polars` is fairly short - vendoring it so we can render all the eager backends the same isn't *that unrealistic* and would be a nice DX improvement
  - https://github.com/pola-rs/polars/blob/dfa5efe71156c654a1ba3a54b865eae723a818e9/py-polars/polars/dataframe/_html.py

narwhals/_utils.py
```diff
@@ -1579,6 +1579,48 @@ def generate_repr(header: str, native_repr: str) -> str:
...
 
 
+# NOTE: Unsure on how to test this reliably
+def generate_repr_html(
+    header: Literal["Narwhals DataFrame", "Narwhals LazyFrame", "Narwhals Series"],
+    native_html: str,
+) -> str | None:  # pragma: no cover
+    if header == "Narwhals LazyFrame" and "LazyFrame" in native_html:
+        html = native_html.replace("LazyFrame", "LazyFrame.to_native()")
+        return f"{html}<p><b>{header}</b></p>"
```
FBruzzesi: I am mostly nitpicking here but... isn't the `header` actually a `footer`? 😂

narwhals/_utils.py
```diff
@@ -1579,6 +1579,48 @@ def generate_repr(header: str, native_repr: str) -> str:
...
+    if style is not None:
+        text_existing = style.text or ""
+        style.text = f"{text_existing}\n{style_css}"
+    else:
+        style = ET.Element("style")
+        style.text = style_css
+        tree.getroot().insert(0, style)
+    buf = io.BytesIO()
+    tree.write(buf, "utf-8", method="html")
+    return buf.getvalue().decode()
```
FBruzzesi, line 1587: Everything else in this function is a new language to me - I am not very helpful

dangotbanned:
@MarcoGorelli, @FBruzzesi 

Thought I'd do one last check before closing this one, here's a few options to choose from:

1. Don't support this
2. Do it, but less (defer entirely to `polars`, `pandas`)
3. Do it, but style differently (https://github.com/narwhals-dev/narwhals/pull/2776#discussion_r2183599508)
4. Do it, but increase the scope and make `pyarrow` + `pd.Series` look pretty too (https://github.com/narwhals-dev/narwhals/pull/2776#issuecomment-3053349066)

No worries if we don't want it 🙂

narwhals/_utils.py
FBruzzesi: That's very reasonable!

FBruzzesi (commented):
@dangotbanned thank to your ping - I got reminded that I _once_ started to look at this, and never finished. I am _not_ against having a good support, yet I am not very useful nor opinionated about this.

I would try to aim for a pareto optimum that balances usefulness and maintainability 😂

narwhals/_utils.py
dangotbanned: You're quite right 😂 

It started as a `header` until I ran into (https://github.com/narwhals-dev/narwhals/pull/2776#discussion_r2183599508)

I should've updated that to `footer` or [`caption`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/caption)

narwhals/_utils.py
dangotbanned: Ah yeah [`xml.etree.elementtree`](https://docs.python.org/3/library/xml.etree.elementtree.html#tutorial) is a bit of a strange one

I had to learn a bit of [`lxml`](https://lxml.de/tutorial.html) once to fix a particularly broken file.
The API of that is based on this stdlib module, but was more ergonoic than this mess :smile:

To simplify this:

- **Element**: Is a [HTML Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements)
- **Tree**: Refers to a document/webpage, but in this case it is just a table

So I'm essentially doing a fancy find/replace, but trying to preserve the structure of the document

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2776#issuecomment-3146715372

> 4. Do it, but increase the scope and make `pyarrow` + `pd.Series` look pretty too

I've started #2925 and came up against the `pyarrow.ChunkedArray` repr again, while writing an example 🤦‍♂️ 

```py
import pyarrow as pa

import narwhals as nw

>>> nw.Series.from_iterable("a", [4, 1, 3, 2], dtype=nw.UInt32, backend=pa)
┌───────────────────────────────────────────────────────┐
|                    Narwhals Series                    |
|-------------------------------------------------------|
|<pyarrow.lib.ChunkedArray object at 0x0000017129497880>|
|[                                                      |
|  [                                                    |
|    4,                                                 |
|    1,                                                 |
|    3,                                                 |
|    2                                                  |
|  ]                                                    |
|]                                                      |
└───────────────────────────────────────────────────────┘
```

Even if we don't go ahead with `_repr_html_` - I'd really like to be displaying `Series.name` and `Series.dtype` in `__repr__`

The `polars` one manages to fit in both of those + `shape`, while taking up waaaay less horizontal space and 1 fewer lines:

```py

>>> nw.Series.from_iterable("a", [4, 1, 3, 2], dtype=nw.UInt32, backend="polars")
┌─────────────────┐
| Narwhals Series |
|-----------------|
|shape: (4,)      |
|Series: 'a' [u32]|
|[                |
|        4        |
|        1        |
|        3        |
|        2        |
|]                |
└─────────────────┘
```

narwhals/_utils.py
dangotbanned: Oh I forgot, the name `header` actually came from `generate_repr`

https://github.com/narwhals-dev/narwhals/blob/e32e139d07ac56295a30828e6067b299e87fcba8/narwhals/_utils.py#L1515


Anyway - updated it in (https://github.com/narwhals-dev/narwhals/pull/2776/commits/57e333d20ec1f51b3b044b2eedce01e40ac6e6f3)

narwhals/_utils.py
dangotbanned: Realized I never followed this up with an example

Now that I'm looking at it again, maybe above isn't so bad?

image

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2776#issuecomment-3148403247

If we just wanted `pa.ChunkedArray` to look nicer, I've got a very naive *new* repr (not html) for `nw.Series`:

```py
shape: (365,)
dtype: Datetime(time_unit='us', time_zone=None)
name: 'time series'
nw.Series[pyarrow]
[
	2009-01-02 00:00:00
	2009-01-03 00:00:00
	2009-01-04 00:00:00
	2009-01-05 00:00:00
	2009-01-06 00:00:00
	…
	2009-12-28 00:00:00
	2009-12-29 00:00:00
	2009-12-30 00:00:00
	2009-12-31 00:00:00
	2010-01-01 00:00:00
]
```


```py
shape: (30,)
dtype: UInt32
name: 'lower max rows'
nw.Series[pyarrow]
[
	0
	1
	2
	…
	27
	28
	29
]
```


```py
shape: (30,)
dtype: Int16
name: 'oh pandas too???'
nw.Series[pandas]
[
	29
	28
	27
	26
	25
	24
	…
	5
	4
	3
	2
	1
	0
]
```

Would be nicer-er if we used the short type codes from `polars`

PR #2805, opened 2025-07-08
chore: fix some missing overload defaults

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_compliant/namespace.py
+++ b/narwhals/_compliant/namespace.py
@@ -178,7 +178,7 @@ class EagerNamespace(
         self,
         data: _2DArray,
         /,
-        schema: Mapping[str, DType] | Schema | Sequence[str] | None,
+        schema: Mapping[str, DType] | Schema | Sequence[str] | None = ...,
     ) -> EagerDataFrameT: ...
 
     def from_numpy(
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -270,7 +270,7 @@ def from_native(
     pass_through: bool,
     eager_only: bool,
     series_only: bool,
-    allow_series: bool | None,
+    allow_series: bool | None = ...,
 ) -> Any: ...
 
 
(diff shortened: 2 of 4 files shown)


narwhals/_compliant/namespace.py
```diff
@@ -178,7 +178,7 @@ def from_numpy(
         self,
         data: _2DArray,
         /,
-        schema: Mapping[str, DType] | Schema | Sequence[str] | None,
+        schema: Mapping[str, DType] | Schema | Sequence[str] | None = ...,
```
dangotbanned, line 181: `schema` is required for the `DataFrame` case (even if it is `None`)

I think that was what I was going for with these overloads anyway 😄

narwhals/translate.py
```diff
@@ -270,7 +270,7 @@ def from_native(
     pass_through: bool,
     eager_only: bool,
     series_only: bool,
-    allow_series: bool | None,
+    allow_series: bool | None = ...,
```
dangotbanned, line 273: Nice catch!

Would you be able to add a (*maybe typing only?*) test that fails without this default?

This comment made me think we didn't need any defaults, but it could be outdated?

```py
# All params passed in as variables
```

narwhals/_compliant/namespace.py
dangotbanned: Sorry I'm not sure that was clear, here's a second attempt...

The `schema` parameter is **always unused** for a `Series` input

https://github.com/narwhals-dev/narwhals/blob/6bf173100b040670fbf6823d647d968e332ea59d/narwhals/_compliant/namespace.py#L173-L192

MarcoGorelli:
looking back i don't think this is really necessary - closing then, and i hadn't appreciated that some overloads intentionally don't include some defaults because of what they want to communicate

Issue #2806, opened 2025-07-08
[Bug]: cuDF backend fails on native python date filtering


jp-gr:
### Describe the bug

cuDF backend fails when filtering by a native python date object:
```
lf.filter(nw.col("today") >= dt.date(2025, 1, 1))
```
With the following error:
```
TypeError: '>=' not supported between instances of 'DatetimeColumn' and 'datetime.date'
```

Note that this works fine on the polars backend

### Steps or code to reproduce the bug

```
import datetime as dt
import os
import tempfile

import cudf
import narwhals as nw


def build_parquet(tmp_dir: str) -> str:
    """Create a tiny Parquet file with one date row."""
    df = cudf.DataFrame(
        {
            "today": cudf.to_datetime(["2025-01-01"]),
            "value": [42],
        }
    )
    path = os.path.join(tmp_dir, "sample.parquet")
    df.to_parquet(path)
    return path


def repro(path: str, backend):
    lf = nw.scan_parquet(path, backend=backend)
    try:
        df = (
            lf.filter(
                nw.col("today") >= dt.date(2025, 1, 1)  # <-- native `datetime.date`
            ).collect()
        )
        print(df)
    except TypeError as err:
        # Show the exact exception that demonstrates the bug
        print("Caught exception:", err)
        raise  # re-raise so the script exits non-zero



if __name__ == "__main__":
    with tempfile.TemporaryDirectory() as tmp:
        parquet_path = build_parquet(tmp)

        print("=== Polars backend (works) ===")
        repro(parquet_path, "polars")

        print("=== Cudf backend (fails) ===")
        repro(parquet_path, backend="cudf")
```
### Expected results

Returns the same as the polars df

### Actual results

TypeError: '>=' not supported between instances of 'DatetimeColumn' and 'datetime.date'

### Please run narwhals.show_version() and enter the output below.

```shell
System:
    python: 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0]
executable: .venv/bin/python3
   machine: Linux-5.15.0-91-generic-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 1.46.0
       pandas: 2.2.2
       polars: 1.31.0
         cudf: 
        modin: 
      pyarrow: 19.0.1
        numpy: 1.26.4
```

### Relevant log output

```shell

```

jp-gr:
[...]

PR #2809, opened 2025-07-08
fix: Use `Distribution` for `show_versions` 

### LINKED ISSUE #2807: bug: `show_versions` doesn't show cudf version
see https://github.com/narwhals-dev/narwhals/issues/2806#issuecomment-3049555129

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2807

## Checklist

- [x] Tests added
- [ ] Documented the changes

**Breaking change**: It removes numpy from the list, but it automatically handles all the upstream backends


narwhals/functions.py
```diff
@@ -562,18 +561,21 @@ def _get_deps_info() -> dict[str, str]:
...
+    for impl in Implementation:
+        if impl in {Implementation.UNKNOWN, Implementation.PYSPARK_CONNECT}:
+            continue
+
         try:
-            deps_info[modname] = version(modname)
-        except PackageNotFoundError:  # noqa: PERF203
-            deps_info[modname] = ""
+            deps_info[impl.name.lower()] = ".".join(
+                str(t) for t in impl._backend_version()
```
FBruzzesi: For context: if I try this on kaggle, the cudf version I get from "25.08.00a474" becomes "25.8.474". Thoughts on this?

EdAbati: I _think_ it is because `_backend_version()` (that uses `parse_version`) returns a `tuple[int,...]`
`parse_version` does `tuple(int(re.sub(r"\D", "", v)) for v in version_str.split("."))`

If we want to keep the "full" string version we may need to stick with `importlib.metadata.version` :/

FBruzzesi: The issue is that cudf installation comes with the cuda one, so for example the name would be `"cudf-cu12"`. I am not sure we can automatically detect the suffix `"cu12"`.

On the other hand, `_backend_version()` hacks into `__version__` attribute of the library if it can.
I think two options are:

- Decouple the retrieving version and parsing version parts in `_backend_version()`
- Change the regex

I have slight preference for the former.

dangotbanned: How about casting a wider net with [`importlib.metadata.distributions`](https://github.com/python/cpython/blob/6c81051aa8813665f1d3f651da90a146cb07a8bf/Lib/importlib/metadata/__init__.py#L963-L968)?

Would something like this give you more flexibility on the name to pick?
```py
from importlib.metadata import distributions

>>> [(d.name, d.version) for d in distributions()]
```

<details><summary>Show bigly output</summary>
<p>


```py
[('asttokens', '3.0.0'),
 ('atpublic', '6.0.1'),
 ('attrs', '25.3.0'),
 ('babel', '2.17.0'),
 ('backrefs', '5.9'),
 ('black', '25.1.0'),
 ('certifi', '2025.7.9'),
 ('cfgv', '3.4.0'),
 ('charset-normalizer', '3.4.2'),
 ('click', '8.2.1'),
 ('cloudpickle', '3.1.1'),
 ('colorama', '0.4.6'),
 ('comm', '0.2.2'),
 ('covdefaults', '2.3.0'),
 ('coverage', '7.9.2'),
 ('dask', '2025.5.1'),
 ('debugpy', '1.8.14'),
 ('decorator', '5.2.1'),
 ('distlib', '0.3.9'),
 ('duckdb', '1.3.0'),
 ('executing', '2.2.0'),
 ('filelock', '3.18.0'),
 ('fsspec', '2025.5.1'),
 ('ghp-import', '2.1.0'),
 ('griffe', '1.7.3'),
 ('hypothesis', '6.135.26'),
 ('ibis-framework', '10.6.0'),
 ('identify', '2.6.12'),
 ('idna', '3.10'),
 ('iniconfig', '2.1.0'),
 ('ipykernel', '6.29.5'),
 ('ipython', '9.3.0'),
 ('ipython_pygments_lexers', '1.1.1'),
 ('jedi', '0.19.2'),
 ('Jinja2', '3.1.6'),
 ('jupyter_client', '8.6.3'),
 ('jupyter_core', '5.8.1'),
 ('locket', '1.0.0'),
 ('Markdown', '3.8.2'),
 ('markdown-exec', '1.11.0'),
 ('markdown-it-py', '3.0.0'),
 ('MarkupSafe', '3.0.2'),
 ('matplotlib-inline', '0.1.7'),
 ('mdurl', '0.1.2'),
 ('mergedeep', '1.3.4'),
 ('mkdocs', '1.6.1'),
 ('mkdocstrings', '0.29.1'),
 ('mkdocstrings-python', '1.16.12'),
 ('mkdocs-autorefs', '1.4.2'),
 ('mkdocs-get-deps', '0.2.0'),
 ('mkdocs-material', '9.6.15'),
 ('mkdocs-material-extensions', '1.3.1'),
 ('modin', '0.34.0'),
 ('mypy', '1.15.0'),
 ('mypy_extensions', '1.1.0'),
 ('narwhals', '1.46.0'),
 ('nest-asyncio', '1.6.0'),
 ('nodeenv', '1.9.1'),
 ('numpy', '2.3.1'),
 ('packaging', '25.0'),
 ('paginate', '0.5.7'),
 ('pandas', '2.2.3'),
 ('pandas-stubs', '2.2.3.250308'),
 ('parso', '0.8.4'),
 ('parsy', '2.1'),
 ('partd', '1.4.2'),
 ('pathspec', '0.12.1'),
 ('platformdirs', '4.3.8'),
 ('pluggy', '1.6.0'),
 ('polars', '1.30.0'),
 ('prettytable', '3.16.0'),
 ('pre_commit', '4.2.0'),
 ('prompt_toolkit', '3.0.51'),
 ('psutil', '7.0.0'),
 ('pure_eval', '0.2.3'),
 ('pyarrow', '20.0.0'),
 ('pyarrow-hotfix', '0.7'),
 ('pyarrow-stubs', '19.2'),
 ('Pygments', '2.19.2'),
 ('pygments-ansi-color', '0.3.0'),
 ('pymdown-extensions', '10.16'),
 ('pyright', '1.1.403'),
 ('pytest', '8.4.1'),
 ('pytest-cov', '6.2.1'),
 ('pytest-env', '1.1.5'),
 ('pytest-randomly', '3.16.0'),
 ('python-dateutil', '2.9.0.post0'),
 ('pytz', '2025.2'),
 ('pywin32', '310'),
 ('PyYAML', '6.0.2'),
 ('pyyaml_env_tag', '1.1'),
 ('pyzmq', '27.0.0'),
 ('requests', '2.32.4'),
 ('rich', '14.0.0'),
 ('ruff', '0.12.2'),
 ('six', '1.17.0'),
 ('sortedcontainers', '2.4.0'),
 ('sqlframe', '3.36.3'),
 ('sqlglot', '26.33.0'),
 ('stack-data', '0.6.3'),
 ('toolz', '1.0.0'),
 ('tornado', '6.5.1'),
 ('traitlets', '5.14.3'),
 ('types-pytz', '2025.2.0.20250516'),
 ('typing_extensions', '4.14.1'),
 ('tzdata', '2025.2'),
 ('urllib3', '2.5.0'),
 ('uv', '0.7.19'),
 ('virtualenv', '20.31.2'),
 ('watchdog', '6.0.0'),
 ('wcwidth', '0.2.13')]
```


</p>
</details> 


> Decouple the retrieving version and parsing version parts in `_backend_version()`

I'm not sure how this is related, since on `main` we aren't using `Implementation` or anything from `_utils` in these functions:

https://github.com/narwhals-dev/narwhals/blob/cc223788d5e33fe30221d4a1526c74fe15d254d0/narwhals/functions.py#L552-L596

FBruzzesi: Thanks @dangotbanned to taking a look into this :)

> How about casting a wider net with [`importlib.metadata.distributions`](https://github.com/python/cpython/blob/6c81051aa8813665f1d3f651da90a146cb07a8bf/Lib/importlib/metadata/__init__.py#L963-L968)?
> 
> Would something like this give you more flexibility on the name to pick?

It's nice that everything is listed, yet we would need to manually update the list of what we look for.

> since on `main` we aren't using `Implementation` or anything from `_utils` in these functions:

Mmmh, https://github.com/narwhals-dev/narwhals/pull/2809/commits/a9a092a8b6044778d65f31507c7bf56978ead207 is what I meant: i.e. display the version without parsing it into a tuple and then constructing the string back

dangotbanned: > > Would something like this give you more flexibility on the name to pick?
> It's nice that everything is listed, yet we would need to manually update the list of what we look for.

I was thinking it would mean we could use a `re.Pattern[str]` on the `Distribution.name`, possibly just for `cudf`?


> For context: if I try this on kaggle, the cudf version I get from "25.08.00a474" becomes "25.8.474". Thoughts on this?

Could you link a `kaggle` notebook I could test this out on please? 🙂 

I can't tell if the `Distribution.name` for the op (https://github.com/narwhals-dev/narwhals/issues/2806#issuecomment-3049555129) would be either:
- `"cudf-cu12"`
- `"cudf"`

But for both of those cases we could even keep it simple and use `str.startswith`:

```py

from importlib.metadata import Distribution, distributions
from collections.abc import Iterable, Iterator, Sequence
from itertools import chain

extra_names = ["narwhals"]
member_names = Implementation._member_names_
exclude = {"PYSPARK_CONNECT", "UNKNOWN"}
target_names = [
    name.lower() for name in (*extra_names, *member_names) if name not in exclude
]


def dep_info(target: str, dists: Iterable[Distribution]) -> Iterator[tuple[str, str]]:
    def gen() -> Iterator[tuple[str, str]]:
        for dist in dists:
            if dist.name.startswith(target):
                yield dist.name, dist.version

    yield next(gen(), (target, ""))


def deps_info(target_names: Sequence[str]) -> dict[str, str]:
    dists = list(distributions())
    return dict(chain.from_iterable(dep_info(name, dists) for name in target_names))


>>> deps_info(target_names)
{'narwhals': '1.46.0',
 'pandas': '2.2.3',
 'modin': '0.34.0',
 'cudf': '',
 'pyarrow': '20.0.0',
 'pyspark': '',
 'polars': '1.30.0',
 'dask': '2025.5.1',
 'duckdb': '1.3.0',
 'ibis-framework': '10.6.0',
 'sqlframe': '3.36.3'}
```

tests/system_info_test.py
```diff
@@ -29,7 +29,6 @@ def test_get_deps_info() -> None:
     assert "cudf" in deps_info
     assert "modin" in deps_info
     assert "pyarrow" in deps_info
-    assert "numpy" in deps_info
```
MarcoGorelli, line 32: can we keep numpy? it'd be useful to have it to reproduce certain pandas / pyarrow issues

narwhals/functions.py
```diff
@@ -562,19 +573,16 @@ def _get_deps_info() -> dict[str, str]:
...
-        except PackageNotFoundError:  # noqa: PERF203
-            deps_info[modname] = ""
-    return deps_info
+    extra_names = ("narwhals", "numpy")
+    member_names = Implementation._member_names_
+    exclude = {"PYSPARK_CONNECT", "UNKNOWN"}
+    target_names = tuple(
+        name.lower() for name in (*extra_names, *member_names) if name not in exclude
+    )
+    dists = sorted([(d.name, d.version) for d in distributions()], key=lambda t: t[0])
```
FBruzzesi: Reason for sorting is to avoid matching "pandas" with "pandas-stubs" and similar other cases

dangotbanned: Ah yeah nice one!

When I wrote that, I was originally thinking of handling cases for the different `cudf` names

But I wasn't able to test what they were or if they showed as different distributions in the first place 😅

https://github.com/narwhals-dev/narwhals/pull/2809#discussion_r2197108316

FBruzzesi:
@MarcoGorelli sorry I had some second thoughts on a double for loop 😂

I just pushed a commit that avoids that as much as possible: https://github.com/narwhals-dev/narwhals/pull/2809/commits/e6277a850ea496062fd9675d0221146ec1754f3e

The output does not _fully_ match the distribution names, rather it matches _our_ names. What I mean by this is that "ibis-framework" version is displayed under "ibis", "cudf-cu12" is displayed under "cudf" (see screenshot below)

Let me know what you think about this. cc: @dangotbanned please also share your opinion.

image

MarcoGorelli:
looks fine to me, and useful enough for debugging

PR #2814, opened 2025-07-10
enh: Start `narwhals.stable.v2` and enforce deprecations

MarcoGorelli:
From the downstream tests, the only breakage is in:
- hierarchical forecast is using `native_namespace` in `from_dict`

That's...all, from what I can tell.

I've made a PR to them to update this: https://github.com/Nixtla/hierarchicalforecast/pull/381

I don't want to be disruptive towards them, so perhaps we could take a bit of a middle-ground approach:
- for deprecated functions/params which don't show up in any downstream test, just enforce the deprecation
- for `native_namespace` in `from_dict`, keep it around longer in the main Narwhals namespace at least until there's been a new hierarchicalforecast

todo:
- [x] for `from_native_test`, separate out some tests to just `v1`, and see what we need to test for `v2` as well

Aiming to merge this (and make a Narwhals 2.0 release) by August

### DIFF
--- a/tests/v2_test.py
+++ b/tests/v2_test.py
@@ -332,11 +332,3 @@ def test_imports() -> None:
     from narwhals.stable.v2.dtypes import Enum  # noqa: F401
     from narwhals.stable.v2.selectors import datetime  # noqa: F401
     from narwhals.stable.v2.typing import IntoDataFrame  # noqa: F401
-
-
-def test_interchange_protocol() -> None:
-    class DataFrameLike:
-        def __dataframe__(self) -> None:
-            return None
-
-    nw_v2.from_native(DataFrameLike())


narwhals/stable/v1/__init__.py
```diff
@@ -373,7 +373,7 @@ def gather_every(self, n: int, offset: int = 0) -> Self:
             A new expression.
         """
         return self._with_orderable_filtration(
-            lambda plx: self._to_compliant_expr(plx).gather_every(n=n, offset=offset)
+            lambda plx: self._to_compliant_expr(plx).gather_every(n=n, offset=offset)  # type: ignore[attr-defined]
```
dangotbanned, line 368: Hey @MarcoGorelli 👋 

IIUC, would the motivation for removing deprecated members from `Compliant*` be that *extenders* of `narwhals` aren't required to implement them?

If so, this is a tricky problem to solve 🤔 

> [!NOTE]
> TL;DR: I'd recommend keeping deprecations in the protocols

## Typing

The main downside for us is that the loss of typing information makes it more likely that we *unknowingly* regress on the `v1` promise in the future.
Despite the deprecation, we can clearly see which things are deprecated atm, and have some IDE help with:

<details><summary>Finding all references</summary>
<p>

image

</p>
</details> 

<details><summary>Finding all implementations</summary>
<p>

This one probably should've had a `@deprecated` 🤦‍♂️ 

image

</p>
</details> 

New `[attr-defined]`s make me think that this link between layers is now (*statically*) murky?

If the goal were just to keep a type checker happy, then the two options are:

1. Add `Compliant*V1` protocols, use them in `*V1` `narwhals`-level and implement both in the backends that support `V1`
2. Keep deprecations in `Compliant*` protocols, punt the issue for now


## Runtime
This is the trickier one for me ...

Would a new *internal* backend be expected to support `v1`? 

For example, in the `daft` PR, we can see the *typing* distinction between things that *aren't yet* and *won't ever* be implemented.

https://github.com/MarcoGorelli/narwhals/blob/f918282f14feb43e0a9687d82e16eed79b3832f7/narwhals/_daft/dataframe.py#L333-L340


```py
    gather_every = not_implemented.deprecated(
        "`LazyFrame.gather_every` is deprecated and will be removed in a future version."
    )
    join_asof = not_implemented()
    tail = not_implemented.deprecated(
        "`LazyFrame.tail` is deprecated and will be removed in a future version."
    )
    explode = not_implemented()
```


At runtime we give the same message to users (I don't know why I made that decision (https://github.com/narwhals-dev/narwhals/pull/2232#issuecomment-2733035625) 😅)

A new backend won't get a type checker yelling to add a `not_implemented`, which at runtime will mean we just get an `AttributeError` 😢

MarcoGorelli: we've got a file which specifically tests all the functionality from v1 (`tests/v1_test.py`), if we're removing the functionality then I think it looks a bit odd to keep it but marked as `deprecated`

> Would a new internal backend be expected to support v1?

no, we only need to care about preserving functionality for code that's already been written

narwhals/stable/v2/__init__.py
```diff
@@ -0,0 +1,1257 @@
...
+    R = TypeVar("R")
+
+IntoSeriesT = TypeVar("IntoSeriesT", bound="IntoSeries", default=Any)
+
+
+class DataFrame(NwDataFrame[IntoDataFrameT]):
+    @inherit_doc(NwDataFrame)
+    def __init__(self, df: Any, *, level: Literal["full", "lazy", "interchange"]) -> None:
+        assert df._version is Version.V2  # noqa: S101
+        super().__init__(df, level=level)
```
dangotbanned, line 96: We can move `level` to `v1`-only can't we?

Or at least `"interchange"` I thought was on the chopping block?

MarcoGorelli: yes true, i just think it'd require a larger refactor

it's not user-facing though, so not necessarily pressing?

MarcoGorelli:
Just tried v2 out in Bokeh, and it all passes as-is

In Altair it requires some code changes because `eager_or_interchange_only` was used there, but with those in it seems fine

narwhals/stable/v2/__init__.py
dangotbanned: Yeah that seems fair

How about in this PR, try removing [`v2.typing.DataFrameLike`](https://github.com/MarcoGorelli/narwhals/blob/eb60fe72d621bf63dee6c151b772a31598e77bc6/narwhals/stable/v2/typing.py#L29-L32) and adding some tests using a mock w/ `__dataframe__`?

I think making sure the typing on `v2` rejects interchange only *should* be doable 🙂 

We can punt the rest to another issue

PR #2836, opened 2025-07-16
refactor: Simplify `combine_metadata`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2492

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Overall, this is not a real change. What this PR achieves:

1. It flips the logic on the positive point of view: from "result is not <xyz>" to "result is <xyz>"
2. It uses boolean operators instead of if-statements/conditionals to "reduce" complexity

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -45,7 +45,7 @@ else:
     )
 
 
-def pytest_addoption(parser: Any) -> None:
+def pytest_addoption(parser: pytest.Parser) -> None:
     parser.addoption(
         "--runslow", action="store_true", default=False, help="run slow tests"
     )


narwhals/_expression_parsing.py
```diff
@@ -549,24 +549,19 @@ def combine_metadata(  # noqa: C901, PLR0912
...
-                result_is_not_scalar_like = True
-            if not metadata.is_literal:
-                result_is_not_literal = True
-            if metadata.is_filtration:
-                n_filtrations += 1
+            result_preserves_length |= metadata.preserves_length
+            result_is_elementwise &= metadata.is_elementwise
+            result_is_scalar_like &= metadata.is_scalar_like
+            result_is_literal &= metadata.is_literal
+            n_filtrations += int(metadata.is_filtration)
```
dangotbanned, line 564: I'm pretty impressed by this 🥳 

I hadn't seen a real world use-case for `__iand__` before, but this looks like it!

PR #2839, opened 2025-07-16
refactor: Simplify compliant `Series.hist`

### LINKED ISSUE #2487: Simplify `*Series.hist`
All three implementations are breaking some rules.

I haven't spent much time thinking on a good solution, but we should probably:

- Identify common behavior
  - E.g. are they all following similar/the same steps?
- See if we can pattern this into a `CompliantHist` (*working title*) protocol

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_arrow/series.py#L1030-L1037

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_pandas_like/series.py#L940-L946

https://github.com/narwhals-dev/narwhals/blob/018d92356d1418bf448a28408f90bafeb07e96e7/narwhals/_polars/series.py#L523-L529

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2487

## Checklist

- [ ] Tests added
- [ ] Documented the changes


FBruzzesi:
Ok I am tagging it as ready for review, but I am cheating on type hints.

Will drop some inline comments for the daily struggle with them 😭

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1017,85 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+        data: dict[str, Any]
+        if bin_count == 0 or (bins is not None and len(bins) <= 1):
+            data = {"breakpoint": [], "count": []}
+        elif pc.sum(
             pc.invert(pc.or_(pc.is_nan(self.native), pc.is_null(self.native))).cast(
-                pa.uint8()
+                pa.uint64()
             ),
             min_count=0,
+        ) == pa.scalar(0, type=pa.uint64()):
```
FBruzzesi: 1. I tried to put `pc.sum(...)` into its own method, but mypy kept complaining that the return type should have been `pa.Scalar[...]` where `...` was `Uint8Type` which does not exist? at least I could not find it in the docs, nor in `pa` namespace, not `pa.dtypes`
2. A different cheat would have been to return after moving it to a python int via `.as_py()`
3. The point above would also make the comparison simpler - I don't know how it was not complaining before, but now comparing uint8 with uint64 was raising another issue

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1017,85 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
-            counts[np.isin(obj_cats, obs_cats)] = obs_counts[np.isin(obs_cats, obj_cats)]
-
-            bin_right = bins[1:]
-            return counts, bin_right
-
-        counts: Sequence[int | float | pa.Scalar[Any]] | np.typing.ArrayLike
-        bin_right: Sequence[int | float | pa.Scalar[Any]] | np.typing.ArrayLike
-
-        data_count = pc.sum(
+        data: dict[str, Any]
```
FBruzzesi: I don't really know where to start with this. My best guess was `dict[str, Sequence[int | float] | _1DArray | ArrayAny]` but still not a happy ending from `_hist_from_bins` below

narwhals/_polars/series.py
```diff
@@ -473,84 +473,134 @@ def hist(  # noqa: C901, PLR0912
...
+
+            lower = cast("float", series.min())
+            upper = cast("float", series.max())
+
+            if lower == upper:
+                lower -= 0.5
+                upper += 0.5
+
+            width = (upper - lower) / bin_count
+            bins = pl.int_range(0, bin_count + 1, eager=True) * width + lower  # type: ignore[assignment]
```
FBruzzesi: Previous version was casting to list, which is a bit of a missed opportunity to keep everything native

narwhals/_polars/series.py
```diff
@@ -473,84 +473,134 @@ def hist(  # noqa: C901, PLR0912
...
+        self,
+        series: pl.Series,
+        bins: list[float | int] | None,
+        bin_count: int | None,
+        backend_version: tuple[int, ...],
+    ) -> tuple[list[float] | None, int | None]:
+        """Prepare bins based on backend version compatibility.
+
+        polars <1.15 does not adjust the bins when they have equivalent min/max
+        polars <1.5 with bin_count=...
```
FBruzzesi, line 526: I am a bit confused by this line comment, but I kept it as it was

narwhals/_polars/series.py
dangotbanned: Have you thought about reimplementing the whole of the `polars` one using expressions?

I think this is the most complicated `Series` method and I'd guess could benefit the most *performance-wise* 🫣

FBruzzesi: Not really! As the main issue with polars is "just" managing all the different versions before and after the call to `.hist(...)`, I didn't invest much time on it 🙈

narwhals/_arrow/series.py
dangotbanned: > I tried to put `pc.sum(...)` into its own method, but mypy kept complaining that the return type should have been `pa.Scalar[...]` where ... was `Uint8Type` which does not exist? at least I could not find it in the docs, nor in `pa` namespace, not `pa.dtypes`


This was corrected in a more recent version of `pyarrow-stubs` (https://github.com/zen-xu/pyarrow-stubs/pull/230)

I've been putting off upgrading our pinned version, as:
- they aren't tested with `mypy` anymore
  - (https://github.com/zen-xu/pyarrow-stubs/pull/227)
- the author prefers to ignore `pyright` rules that I mentioned caused us downstream issues
  - https://github.com/zen-xu/pyarrow-stubs/pull/208
  - https://github.com/zen-xu/pyarrow-stubs/pull/209
  - https://github.com/zen-xu/pyarrow-stubs/pull/233

narwhals/_compliant/series.py
```diff
@@ -345,6 +345,18 @@ def __getitem__(self, item: MultiIndexSelector[Self]) -> Self:
...
+        self, bins: list[float | int] | None, bin_count: int | None
+    ) -> dict[str, Any]:
+        """Create histogram result if self is an empty series."""
+        from numpy import linspace, zeros  # ignore-banned-import
+
+        if bins is not None:
+            return {"breakpoint": bins[1:], "count": zeros(shape=len(bins) - 1)}
+
+        count = bin_count if bin_count is not None else 1
+        return {"breakpoint": linspace(0, 1, count + 1)[1:], "count": zeros(count)}
```
dangotbanned: Okay so I've got a couple notes on this one 🙂 

1. Even though it is an inline import that is technically not on `CompliantSeries` itself - I'd prefer if we didn't have a `numpy` import here
2. For `cudf` (if that supports `hist`), wouldn't we want to prefer `cupy` over `numpy`?
3. For `pyarrow`, we couldn't we use `pa.repeat(0, len(bins) - 1)` instead of `np.zeros`? https://github.com/narwhals-dev/narwhals/blob/3fbb2b30c4b3322447a455ceeb4bca548da9101c/narwhals/_arrow/series.py#L188

But something more interesting to me is that it looks like we *could* expose both of these as: 

- `nw.linear_space` ([`pl.linear_space`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.linear_space.html))
  - could pair nicely with (#2722)
  - I guess we'd do `datetime_range` in same way as well maybe?
- `nw.zeros` ([pl.zeros](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.zeros.html))
  - is sugar around [`pl.repeat`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.repeat.html) - which it looks like all the eager backends can do natively 😅 

I wouldn't have thought of this if you hadn't opened the PR, and it is exactly the kind of thinking I wanted to come out of refactoring like in (#2486) 🥳

narwhals/_polars/series.py
```diff
@@ -473,84 +474,127 @@ def hist(  # noqa: C901, PLR0912
...
+
+        # Handle version-specific bin preparation
+        bins, bin_count = self._prepare_bins(
+            series=series, bins=bins, bin_count=bin_count
+        )
+
+        # Polars inconsistently handles NaN values when computing histograms
+        #   against predefined bins: https://github.com/pola-rs/polars/issues/21082
+        if BACKEND_VERSION < (1, 15) or bins is not None:
+            series = series.set(series.is_nan(), None)
```
dangotbanned: I'm thinking this was adapted from `pandas`, I didn't even know this method existed!

https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.set.html

> Use of this function is frequently an anti-pattern, as it can block optimisation (predicate pushdown, etc). Consider using `pl.when(predicate).then(value).otherwise(self)` instead.

narwhals/_compliant/series.py
FBruzzesi: You are right! Addressed in https://github.com/narwhals-dev/narwhals/pull/2839/commits/591533b0c1c78e8b89b08be6dea9316bf47dc096

dangotbanned: Great!

I followed up with [refactor: Add `zeros`](https://github.com/narwhals-dev/narwhals/pull/2839/commits/5e78ec625d67c24c9dd297516a5d995eabcb523d) and [perf: Use `is_null(nan_is_null=True)`](https://github.com/narwhals-dev/narwhals/pull/2839/commits/9b8682b2fa6d3873376d328dbe2d13f5b54cdc23)

Fairly related, I guess I've collected a a cheatsheet for the `pyarrow`/`compute` functions here 😂 


<details><summary><b>Show aliased/adapted functions</b></summary>
<p>




https://github.com/narwhals-dev/narwhals/blob/168e930d560a62b882dbefce7ea2d7f7f3a5ce12/narwhals/_plan/arrow/functions.py#L11-L14

https://github.com/narwhals-dev/narwhals/blob/168e930d560a62b882dbefce7ea2d7f7f3a5ce12/narwhals/_plan/arrow/functions.py#L51-L79

https://github.com/narwhals-dev/narwhals/blob/168e930d560a62b882dbefce7ea2d7f7f3a5ce12/narwhals/_plan/arrow/functions.py#L138-L163

https://github.com/narwhals-dev/narwhals/blob/168e930d560a62b882dbefce7ea2d7f7f3a5ce12/narwhals/_plan/arrow/functions.py#L202-L221


</p>
</details>

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1018,95 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+                    pc.and_(
+                        pc.greater_equal(self.native, pa.scalar(float(bins[0]))),
+                        pc.less_equal(self.native, pa.scalar(float(bins[1]))),
+                    ).cast(pa.uint8())
+                )
+            ]
+            return {"breakpoint": [bins[-1]], "count": counts}
+
+        # Handle multiple bins
+        bin_indices = np.searchsorted(bins, self.native, side="left")
```
dangotbanned: The doc for `np.searchsorted` says

> Find indices where elements should be inserted to maintain order.

Isn't that one of these?

- https://arrow.apache.org/docs/python/generated/pyarrow.compute.sort_indices.html
- https://arrow.apache.org/docs/python/generated/pyarrow.compute.array_sort_indices.html

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1018,95 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+        """Calculate histogram from bins."""
+        import numpy as np  # ignore-banned-import
+
+        # Handle single bin case
+        if len(bins) == 2:
+            counts = [
+                pc.sum(
+                    pc.and_(
+                        pc.greater_equal(self.native, pa.scalar(float(bins[0]))),
+                        pc.less_equal(self.native, pa.scalar(float(bins[1]))),
```
dangotbanned: I'm a bit confused on this `float` conversion -> wrapping as a `pa.scalar`

- If we have `bins: list[float]`, then the conversion isn't needed
- If we have another type, we should use the `pyarrow` equivalent in `pa.scalar(type=...)`

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1018,95 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
-        )
+        from numpy import linspace  # ignore-banned-import
+
+        count = bin_count if bin_count is not None else 1
+        return {"breakpoint": linspace(0, 1, count + 1)[1:], "count": zeros(count)}
+
+    def _prepare_bins(
+        self, bins: list[float | int] | None, bin_count: int | None
+    ) -> list[float]:
+        """Prepare bins for histogram calculation."""
```
dangotbanned: I think this would be simpler if we didn't pass in the mutually exclusive cases in the first place

https://github.com/narwhals-dev/narwhals/blob/9b8682b2fa6d3873376d328dbe2d13f5b54cdc23/narwhals/_arrow/series.py#L1065-L1068

FBruzzesi: Sure it makes sense and it looks better! However, even by explicitly declaring the following overloads:

```py
@overload
def hist(
    self, bins: list[float | int], *, bin_count: None, include_breakpoint: bool,
) -> DataFrame[Any]: ...

@overload
def hist(
    self, bins: None, *, bin_count: int, include_breakpoint: bool,
) -> DataFrame[Any]: ...
```

The following is flagged as an issue by the type checker:

```py
final_bins = (
    bins if bins is not None else self._prepare_bins(bin_count=bin_count)
)
```

narwhals/_polars/series.py
FBruzzesi: We could do:

> series = series.zip_with(~series.is_nan(), pl.repeat(None, len(series), eager=True))

Not sure if it's more likable

narwhals/_arrow/series.py
FBruzzesi: How those two incorporate the bin values? To me it looks like those two functions are to sort indices on one array, not to find indexes where another array would go. Am I missing something?

narwhals/_arrow/series.py
dangotbanned: Ah sorry my bad, I didn't give much detail there 🤦

I'll try to put together a more complete example of what I mean later today.

The issue is something that's come up a few times in #2572 - but we can learn from the rust-side of `polars` in how to solve it 😎

narwhals/_arrow/series.py
dangotbanned: IIRC, it was something like use one of those + `take` maybe?

dangotbanned: I thought there was a `numpy` version of that pattern being used a few lines later

narwhals/_polars/series.py
dangotbanned: How about `fill_nan(None)`?

https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.fill_nan.html

narwhals/_arrow/series.py
dangotbanned: > [!NOTE]
> I've written a tutorial/blog post (sorry) 😂  

## Problem

Okay so, if we step back from `hist` for a moment; the general problem we're trying to solve is a function with mutually exclusive signatures.

Using `@overload`(s), we can identify the individual cases - but these are only for the benefit of the **caller** when we have multiple parameters that differ.

### Examples
First, what do we get with no overloads?

We can't say for sure, so the only reasonable return type is what we started with:

```py
import typing as t

def hist_1_no_overload(bins: list[float] | None, bin_count: int) -> list[float] | None:
    return bins  # nothing was narrowed
```

If we only need to narrow `bins`, then a pair of `@overload`(s) can do the trick:

```py
@t.overload
def hist_1_overload(bins: list[float], bin_count: int) -> list[float]: ...
@t.overload
def hist_1_overload(bins: None, bin_count: int) -> int: ...
def hist_1_overload(bins: list[float] | None, bin_count: int) -> list[float] | int:
    if bins is None:
        return bin_count  # this can **only** be `int`
    return bins  # we've narrowed this to `list[float]`
```


But what we're dealing with in `hist` is more complex.

We  2 parameters, which **can't be used together**:

```py
@t.overload
def hist_2_overload_bad_1(bins: list[float], bin_count: None) -> list[float]: ...
@t.overload
def hist_2_overload_bad_1(bins: None, bin_count: int) -> int: ...
def hist_2_overload_bad_1(
    bins: list[float] | None, bin_count: int | None
) -> list[float] | int:
# ...
```

^^^^ is perfectly fine for the caller:

```py
hist_2_overload_bad_1([1], None)  # okay
hist_2_overload_bad_1(None, 1)  # okay
hist_2_overload_bad_1(None, None)  # warns
hist_2_overload_bad_1([1], 1)  # warns
```

But when we're inside the function, we **still** need to treat it like any combination could have been passed:

```py
@t.overload
def hist_2_overload_bad_1(bins: list[float], bin_count: None) -> list[float]: ...
@t.overload
def hist_2_overload_bad_1(bins: None, bin_count: int) -> int: ...
def hist_2_overload_bad_1(
    bins: list[float] | None, bin_count: int | None
) -> list[float] | int:
    if bins is None:
        return bin_count  # still hasn't been narrowed from `int | None`
    return bins  # we've narrowed this to `list[float]`
```

Okay I hear you say, but what about ...

```py
@t.overload
def hist_2_overload_bad_2(bins: list[float], bin_count: None) -> list[float]: ...
@t.overload
def hist_2_overload_bad_2(bins: None, bin_count: int) -> int: ...
def hist_2_overload_bad_2(
    bins: list[float] | None, bin_count: int | None
) -> list[float] | int:
    if bins is None and bin_count is not None:
        return bin_count  # oooh look we narrowed to `int` in one arm!
    return bins  # but now I'm `list[float] | None` 😭
```
#### Fixed

So how we need to handle it is like this instead:

```py
@t.overload
def hist_2_overload_good(bins: list[float], bin_count: None) -> list[float]: ...
@t.overload
def hist_2_overload_good(bins: None, bin_count: int) -> int: ...
def hist_2_overload_good(
    bins: list[float] | None, bin_count: int | None
) -> list[float] | int:
    if bins is None:
        if bin_count is not None:  # we have to check both like this
            return bin_count  # i can only be `int`!
        raise TypeError
    return bins  # we've narrowed this to `list[float]`
```

## Summary
This stuff is **suuuuuuuuuuuuuper** tedious, especially if we know that the validation happened elsewhere.

I'm gonna cut this one short for today (I had planned to show other places like `fill_null` where this happens)😅

> [!TIP]
> As soon as we've validated which version of `hist` we're doing - that is where we need to split the signatures

### `oh-nodes`

I haven't implemented `hist` yet, but this is what I had so far

https://github.com/narwhals-dev/narwhals/blob/37d77095c547e658ceef7e298acb2442a75cd54d/narwhals/_plan/dummy.py#L222-L241

https://github.com/narwhals-dev/narwhals/blob/37d77095c547e658ceef7e298acb2442a75cd54d/narwhals/_plan/functions.py#L33-L66

I guess that would translate to add two methods like this, which get called by `nw.Series` - narrowing at the source:

```py
CompliantSeries.hist_with_bins
CompliantSeries.hist_with_bin_count
```

dangotbanned: > I guess that would translate to add two methods like this, which get called by `nw.Series` - narrowing at the source:
> 
> ```python
> CompliantSeries.hist_with_bins
> CompliantSeries.hist_with_bin_count
> ```

Just for reference, you can see where `polars` does this in some places like this:

image

FBruzzesi: Thanks @dangotbanned - that's an amazing walkthrough!

As you also mentioned, we are already doing the value checks somewhere else, namely at the narwhals level, so if I understand you correctly, your preferred solution forward would be to call compliant series' `hist_with_bins` and `hist_with_bin_count` from there, without an _actual_ implementation of `hist` itself at the compliant level?

dangotbanned: Yeah I think that would simplify the paths once we're inside those compliant methods 

Ofc, they can still be implemented using shared private methods - but the key thing is to **preserve** the narrowing that we did when possible 😎

So if we go from `T1 | T2` -> `T1`, then we shouldn't pass that to another function that widens it again to another `Union`

FBruzzesi: @dangotbanned this is a quite large commit https://github.com/narwhals-dev/narwhals/pull/2839/commits/24b6d2ead86582b28672d4091be3f1868b333206 but I wanted to keep it together so that's easy to revert if not what you intended!

I like the looks of it, looking forward to hear your opinion 😉

dangotbanned: Looking much better! 

I'll do a full pass of it later today, but I've spotted at least 1 new `pyarrow` optimization - that I couldn't see before it was split up 🥳

dangotbanned: The typing is a bit frustrating, but ...
[perf: `ArrowSeries._is_empty_series` w/ fewer ops](https://github.com/narwhals-dev/narwhals/pull/2839/commits/33b7123e1446c756383a0b4996f76a7ae3fc4c54)


Ignoring the stubs, this is how much we've simplified things

### `main`
```py
pc.sum(
    pc.invert(pc.or_(pc.is_nan(self.native), pc.is_null(self.native))).cast(pa.uint8()),
    min_count=0,
) == pa.scalar(0, type=pa.uint64())
```

### Current
```py
self.native.is_null(nan_is_null=True).combine_chunks().false_count == 0
```

narwhals/_arrow/series.py
dangotbanned: I *think* I'm close to having fully unraveled this 🫰

dangotbanned: > I _think_ I'm close to having fully unraveled this 🫰

Famous last words...

The only thing that's unraveled is my brain 😫 

See (https://github.com/narwhals-dev/narwhals/pull/2839/files#r2217327217)

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1018,102 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+        bin_indices = np.searchsorted(bins, self.native, side="left")
+        bin_indices = pc.if_else(  # lowest bin is inclusive
+            pc.equal(self.native, pa.scalar(bins[0])), 1, bin_indices
+        )
+
+        # Align unique categories and counts appropriately
+        obs_cats, obs_counts = np.unique(bin_indices, return_counts=True)
+        obj_cats = np.arange(1, len(bins))
+        counts = np.zeros_like(obj_cats)
+        counts[np.isin(obj_cats, obs_cats)] = obs_counts[np.isin(obs_cats, obj_cats)]
```
dangotbanned: @FBruzzesi @camriddell 

I can translate **almost all** of this in `pyarrow`, but the last line has stumped me 🤔 


This is fine, and we could replace `np.arange` starting from `pyarrow>=21` 
- https://github.com/apache/arrow/pull/46778
```py
series = self.native
bin_indices_np = np.searchsorted(bins, series, side="left")
bin_indices_pa: pa.ChunkedArray[Any] = pc.if_else(
    pc.equal(series, pa.scalar(bins[0])), 1, bin_indices_np
)

unique = bin_indices_pa.value_counts()
obj_cats = pa.array(np.arange(1, len(bins)), pa.int64())
obs_cats, obs_counts = unique.field("values"), unique.field("counts")
```


The second line here is doing something funky with predicates that I'm not understanding at all 😭 


```py
counts = np.zeros_like(obj_cats)
counts[np.isin(obj_cats, obs_cats)] = obs_counts[np.isin(obs_cats, obj_cats)]
```


Specifically, here where the `0` is the falsy case 

https://github.com/narwhals-dev/narwhals/blob/33b7123e1446c756383a0b4996f76a7ae3fc4c54/tests/series_only/hist_test.py#L30

dangotbanned: The variable names are also not that easy to follow

narwhals/_polars/series.py
```diff
@@ -464,69 +465,99 @@ def __contains__(self, other: Any) -> bool:
...
+
+        lower = cast("float", self.native.min())
+        upper = cast("float", self.native.max())
+
+        if lower == upper:
+            lower -= 0.5
+            upper += 0.5
+
+        width = (upper - lower) / bin_count
+        return pl.int_range(0, bin_count + 1, eager=True) * width + lower  # type: ignore[return-value]
```
dangotbanned: This is returning a `pl.Series` and not a `list[float]` right?

https://github.com/narwhals-dev/narwhals/blob/33b7123e1446c756383a0b4996f76a7ae3fc4c54/narwhals/_polars/series.py#L525

dangotbanned:
Hey @FBruzzesi, little mid-review note
I know that you haven't touched these tests so far 🫣 

([test: Start trying to clean up](https://github.com/narwhals-dev/narwhals/pull/2839/commits/9fc76d3507491d0340168864667fdb10cb265124))

While I'm trying things out locally, I've been having a tough time:
- figuring out which cases were failing
- understanding what the expected cases actually are

So these changes are *mainly* to try and make that part easier, so we get ids like:


```py
hist_test.py::test_hist_bin[pandas-[-inf, 2.5, 5.5, inf]-[3, 3, 1]-breakpoint-False]
```

In the process of that, I discovered
- most of the typing has been `Any`
- there's a lot of repetition
I'm gonna call it for now and just change that one test - since it seems like another rabbit hole 😅 

Back to work for me 😉

narwhals/_arrow/series.py
```diff
@@ -1019,103 +1018,102 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+                    ).cast(pa.uint8())
+                )
+            ]
+            return {"breakpoint": [bins[-1]], "count": counts}
+
+        # Handle multiple bins
+        bin_indices = np.searchsorted(bins, self.native, side="left")
+        bin_indices = pc.if_else(  # lowest bin is inclusive
+            pc.equal(self.native, pa.scalar(bins[0])), 1, bin_indices
+        )
```
dangotbanned: - defers `numpy` import
- using `lit` more
- determined to reduce the line count 😅


```suggestion
    def _hist_data_from_bins(self, bins: list[float]) -> dict[str, Any]:
        """Calculate histogram from bins."""
        ser = self.native
        # Handle single bin case
        if len(bins) == 2:
            is_between = pc.and_(
                pc.greater_equal(ser, lit(bins[0])), pc.less_equal(ser, lit(bins[1]))
            )
            count = pc.sum(is_between.cast(pa.uint8()))
            return {"breakpoint": [bins[-1]], "count": [count]}

        # Handle multiple bins
        import numpy as np  # ignore-banned-import

        bin_indices = np.searchsorted(bins, ser, side="left")
        # lowest bin is inclusive
        bin_indices = pc.if_else(pc.equal(ser, lit(bins[0])), 1, bin_indices)
```

narwhals/series.py
```diff
@@ -2587,24 +2587,29 @@ def hist(
...
+            if any(bins[i - 1] >= bins[i] for i in range(1, len(bins))):
+                msg = "bins must increase monotonically"
+                raise ComputeError(msg)
+            if bin_count is None:
+                result = self._compliant_series._hist_from_bins(
+                    bins=bins, include_breakpoint=include_breakpoint
+                )
+            else:  # bin_count is not None
+                msg = "can only provide one of `bin_count` or `bins`"
+                raise ComputeError(msg)
```
dangotbanned: Any time I see a comment, I always want to try and incorporate it into the code 🙂 

```suggestion
            if bin_count is not None:
                msg = f"can only provide one of `bin_count` or `bins`, got: {bin_count=}, {bins=}"
                raise ComputeError(msg)
            result = self._compliant_series._hist_from_bins(
                bins=bins, include_breakpoint=include_breakpoint
            )
```

narwhals/series.py
```diff
@@ -2587,24 +2587,29 @@ def hist(
...
 
-        return self._dataframe(
-            self._compliant_series.hist(
-                bins=bins, bin_count=bin_count, include_breakpoint=include_breakpoint
-            ),
-            level=self._level,
-        )
+            result = self._compliant_series._hist_from_bin_count(
+                bin_count=bin_count_, include_breakpoint=include_breakpoint
+            )
```
dangotbanned: I couldn't come up with a better short name, so `default` will have to do!

```suggestion
            # polars (v1.20) sets bin=10 if neither are provided.
            default = 10
            bin_count = default if bin_count is None else bin_count
            result = self._compliant_series._hist_from_bin_count(
                bin_count=bin_count, include_breakpoint=include_breakpoint
            )
```

narwhals/_compliant/series.py
```diff
@@ -269,7 +261,14 @@ def value_counts(
...
     def zip_with(self, mask: Any, other: Any) -> Self: ...
-
+    @unstable
+    def _hist_from_bins(
+        self, bins: list[float | int], *, include_breakpoint: bool
+    ) -> CompliantDataFrame[Self, Any, Any, Any]: ...
+    @unstable
+    def _hist_from_bin_count(
+        self, bin_count: int, *, include_breakpoint: bool
+    ) -> CompliantDataFrame[Self, Any, Any, Any]: ...
```
dangotbanned: 1. I think these should be named `hist_*`, since they're called by `nw.Series` 
    i. I don't know if we're *always* consistent with that, but that's the general idea I hope
    ii. Also I kept confusing the other method names like `_bins_from_bin_count` 🤦‍♂️ 
2. I've noticed some of the other new methods have a docstring.
    i. These are the one's I would aim to document though, since it isn't part of the public `polars` API

Combining those points, maybe just something short like this, as a reminder for us on IDE hover?

```py
    @unstable
    def hist_from_bins(
        self, bins: list[float | int], *, include_breakpoint: bool
    ) -> CompliantDataFrame[Self, Any, Any, Any]:
        """`Series.hist(bins=...)`."""
        ...

    @unstable
    def hist_from_bin_count(
        self, bin_count: int, *, include_breakpoint: bool
    ) -> CompliantDataFrame[Self, Any, Any, Any]:
        """`Series.hist(bin_count=...)`."""
        ...


```

narwhals/_pandas_like/series.py
```diff
@@ -945,77 +944,84 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
+        data: dict[str, list[int | float] | _1DArray]
+
+        if bin_count == 0:
+            data = {"breakpoint": [], "count": []}
+        elif self.native.count() < 1:
+            array_funcs = self._array_funcs
+
+            data = {
+                "breakpoint": array_funcs.linspace(0, 1, bin_count + 1)[1:],
+                "count": array_funcs.zeros(bin_count),
```
dangotbanned: Ah very nice use of #2757

narwhals/_pandas_like/series.py
```diff
@@ -945,77 +944,84 @@ def rank(self, method: RankMethod, *, descending: bool) -> Self:
...
         )
 
-    def hist(  # noqa: C901, PLR0912
-        self,
-        bins: list[float | int] | None,
-        *,
-        bin_count: int | None,
-        include_breakpoint: bool,
+    def _hist_from_bins(
+        self, bins: list[float | int], *, include_breakpoint: bool
```
dangotbanned: This one is in a few places both before/after this PR

Any time we write `float | int`, it can be replaced with `float`.


If you're interested in *why* - feel free to lose some brain cells by reading through these 😂 

- https://typing.python.org/en/latest/spec/special-types.html#special-cases-for-float-and-complex
- https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018
- https://discuss.python.org/t/options-for-a-long-term-fix-of-the-special-case-for-float-int-complex/54289

I think the new wording in the spec is going to be:

> we treat annotations of `float` as if the user had written `float | int`


Also for reference, `polars` uses `list[float]` instead of `list[float | int]`

https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/series/series.py#L2552-L2615

FBruzzesi:
Disclaimer: I was out enjoying some sun in Berlin after 10 days of rain 😂

---

Hey @dangotbanned 

> Hey @FBruzzesi, little mid-review note I know that you haven't touched these tests so far 🫣

Yep I possibly didn't even look at them 🙈 

> In the process of that, I discovered
> 
> * most of the typing has been `Any`
> * there's a lot of repetition
>   I'm gonna call it for now and just change that one test - since it seems like another rabbit hole 😅
>
> Back to work for me 😉

Are you refactoring them? Is there something wrong you spotted (complexity and repetition aside)?

dangotbanned (commented):
Really great job @FBruzzesi, I'm finally starting to wrap my head around these implementations 😄 

I've got a few suggestions and also made a start on expressifying `PolarsSeries.hist` in (https://github.com/narwhals-dev/narwhals/pull/2839#commits-pushed-41b5fbf)

narwhals/_compliant/series.py
FBruzzesi: Renamed and added the one line documentation for the compliant level methods. Documented the semi-private ones with a one liner

narwhals/series.py
FBruzzesi: It will do the job 😂

narwhals/_arrow/series.py
FBruzzesi: Thanks! Well spotted!

narwhals/_polars/series.py
FBruzzesi: Ignored type one level above. In practice, this is an issue in polars, isn't it? I mean, `Series.hist` can accept bins that is another series

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3092459534

> Are you refactoring them? Is there something wrong you spotted (complexity and repetition aside)?

In (https://github.com/narwhals-dev/narwhals/pull/2839#discussion_r2217327217) I was trying to limit the how much `pyarrow` was using `numpy`.
But whenever I broke something I was struggling to find the test case, so I just refactored `test_hist_bin` so that the `id` would show what the arguments were:

> So these changes are _mainly_ to try and make that part easier, so we get ids like:
> 
> ```python
> hist_test.py::test_hist_bin[pandas-[-inf, 2.5, 5.5, inf]-[3, 3, 1]-breakpoint-False]
> ```
> 


Before I think it was showing up like this maybe?:

```python
hist_test.py::test_hist_bin[pandas-param0-False]
```

dangotbanned:
- https://github.com/narwhals-dev/narwhals/pull/2839#discussion_r2216918146
- https://github.com/narwhals-dev/narwhals/pull/2839#discussion_r2217361430

Ooh something I forgot to mention - that we need to resolve before merging - is how this links into the API completeness.
I think at the moment `hist` will not show as implemented anywhere?

If we go back to the `fill_null` case again, having multiple methods can actually let us document the subset of parameters each backend supports:

https://github.com/narwhals-dev/narwhals/blob/6dac8f79d5e7220d573f0933d49fdac07d6f8700/narwhals/_ibis/expr.py#L541-L554

There I could imagine a `fill_null_with_strategy = not_implemented` could be translated into an `*`/tooltip/footnote describing what level of partial support we can do

---

For `hist`, all eager backends implement both so we don't have the same concern - but I think from a doc perspective, they both require the same tooling to be added?

FBruzzesi:
@dangotbanned I think api-completeness requires a complete re-work. ibis and dask do not even appear in the pages (screenshot).

We should have a different issue to track that.
I am not sure how far we are from this but:

- everything could be flagged as available by default
- spot not implemented methods via the `not_implemented` flag.

image

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3094499540

> We should have a different issue to track that.

Sure, happy to kick it to another issue!

I think I wrote about (https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3094466305) in the original PR for `@requires`, which might be useful to refer to?

I'll try my best to do another pass on this PR today so we can merge soon 🤞

dangotbanned:
Hey @FBruzzesi 👋 

I've been experimenting with `PandasLikeSeries.hist` and I think there could be something to this?

[refactor(suggestion): Try out generalizing everything](https://github.com/narwhals-dev/narwhals/pull/2839/commits/29263cda39dec1c7092630835428e4269b9cd462)


So we should be able to add:

- `nw.linear_space`
  - Both `pandas` and `pyarrow` are already using `numpy` for this
- `nw.Series.cut`
  - Assuming that `pyarrow` is utilizing `numpy` to do this currently
- `nw.zeros` (and by extension, `nw.repeat`)
  - `pandas` uses `numpy`
  - `pyarrow` can do both natively 


And there's a lot in there that I think could either move up into:

- `nw.Series.hist` (for the empty cases)
- Some shared compliant class (for the general "algorithm" 😄)

But *most importantly*, this is only a suggestion that came out of finally understanding `hist` 😍

FBruzzesi:
Hey @dangotbanned thanks for suggesting https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3096758332 

From a first look, I really like it. I have a few open questions also considering the Discord exchange we had about `hist`:

- Do you think this is better/more extensible if we were to eventually include also `Expr.hist`? Consider that the return type is a struct in such case.
- Do you feel like that the current lack of features, such as `nw.linear_space` and others, are a blocker?
- On the other hand, if we are not going to develop `Expr.hist`, would this be an overkill?
- How should we decide what to keep at the narwhals/compliant level and what to move to this class?

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3109991016

Thanks for all the great questions 😍!

> 1. Do you think this is better/more extensible if we were to eventually include also Expr.hist? Consider that the return type is a struct in such case.


My eyes are about fall out - but for just this part - I was thinking the shortest path to `Expr.hist` would be something like this?

At least for `pyarrow` it is fairly easy, I'd guess `pandas` would need to defer to the same thing as well?

```py
import polars as pl

import narwhals as nw

data = {"a": [1, 3, 8, 8, 2, 1, 3]}
bins: list[float] = [1, 2, 3]
df = pl.DataFrame(data)
nw_ser = nw.from_native(df).get_column("a")
nw_hist = nw_ser.hist(bins, include_breakpoint=True)

>>> nw_hist
┌──────────────────────┐
|  Narwhals DataFrame  |
|----------------------|
|shape: (2, 2)         |
|┌────────────┬───────┐|
|│ breakpoint ┆ count │|
|│ ---        ┆ ---   │|
|│ f64        ┆ u32   │|
|╞════════════╪═══════╡|
|│ 2.0        ┆ 3     │|
|│ 3.0        ┆ 2     │|
|└────────────┴───────┘|
└──────────────────────┘


pl_hist_expr = df.select(pl.col("a").hist(bins, include_breakpoint=True))

nw_pl_hist_struct = nw.from_native(
    nw_hist.to_polars().to_struct("a"), series_only=True
).to_frame()

nw_pa_hist_struct = (
    nw.from_native(nw_hist.to_arrow().to_struct_array(), series_only=True)
    .alias("a")
    .to_frame()
)

>>> pl_hist_expr, nw_pl_hist_struct, nw_pa_hist_struct
(shape: (2, 1)
 ┌───────────┐
 │ a         │
 │ ---       │
 │ struct[2] │
 ╞═══════════╡
 │ {2.0,3}   │
 │ {3.0,2}   │
 └───────────┘,
 ┌──────────────────┐
 |Narwhals DataFrame|
 |------------------|
 |  shape: (2, 1)   |
 |  ┌───────────┐   |
 |  │ a         │   |
 |  │ ---       │   |
 |  │ struct[2] │   |
 |  ╞═══════════╡   |
 |  │ {2.0,3}   │   |
 |  │ {3.0,2}   │   |
 |  └───────────┘   |
 └──────────────────┘,
 ┌────────────────────────────────────────────┐
 |             Narwhals DataFrame             |
 |--------------------------------------------|
 |pyarrow.Table                               |
 |a: struct<breakpoint: double, count: uint32>|
 |  child 0, breakpoint: double               |
 |  child 1, count: uint32                    |
 |----                                        |
 |a: [                                        |
 |  -- is_valid: all not null                 |
 |  -- child 0 type: double                   |
 |[2,3]                                       |
 |  -- child 1 type: uint32                   |
 |[3,2]]                                      |
 └────────────────────────────────────────────┘)
```

> [!NOTE]
> Proposed API 172729: `nw.DataFrame.to_struct`
> https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_struct.html#polars.DataFrame.to_struct

dangotbanned:
> Hey @dangotbanned thanks for suggesting [#2839 (comment)](https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3096758332)
> 
> From a first look, I really like it. I have a few open questions also considering the Discord exchange we had about `hist`:


Thanks @FBruzzesi!

Okay easiest one out of the way first:


> * Do you feel like that the current lack of features, such as `nw.linear_space` and others, are a blocker?

To merging this PR? Absolutely not! 😄 

I have a few thoughts on the `polars` bits that look suspiciously *pandas-esque*, but if you're happy with the PR - then I think we've got plenty of improvements to merge already!

---


If you meant, are things like `nw.linear_space` a blocker for *adopting* `narwhals`?
Probably not, but I've been noticing `np.linspace` being used downstream ([code search](https://github.com/search?q=repo%3Aplotly%2Fplotly.py+OR+repo%3Abokeh%2Fbokeh+OR+repo%3Amarimo-team%2Fmarimo+OR+repo%3Akoaning%2Fscikit-lego+OR+repo%3Aunit8co%2Fdarts+np.linspace&type=code))

I think it would be a great candidate for exposing in `narwhals` to avoid the `numpy` dependency when possible, and provide the better `polars` performance that's on table 🙂

dangotbanned:
## `Expr.hist`

> * Do you think this is better/more extensible if we were to eventually include also `Expr.hist`? Consider that the return type is a struct in such case.
> * On the other hand, if we are not going to develop `Expr.hist`, would this be an overkill?

I guess I'm viewing adding `Expr.hist` as unrelated to these changes, so I'll explain my thought process a bit more 😅 

The path I see to that is:

### `pandas`, `pyarrow`
They can do something like (https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3110332853), which leaves their `Series` implementation unchanged.

I'm cutting a lot of corners to get here (including the public API), but this is a quick-and-dirty one for `ArrowExpr`

> [!NOTE]
> There are no changes needed in either `ArrowExpr` or `ArrowSeries`



<details><summary>Arrow version diff</summary>
<p>


```diff
diff --git a/narwhals/_arrow/dataframe.py b/narwhals/_arrow/dataframe.py
index 120d11f3f..a160d4311 100644
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -466,6 +466,11 @@ class ArrowDataFrame(
             return {ser.name: ser for ser in it}
         return {ser.name: ser.to_list() for ser in it}
 
+    def to_struct(self, name: str = "") -> ArrowSeries:
+        return ArrowSeries.from_native(
+            self.native.to_struct_array(), context=self, name=name
+        )
+
     def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
         plx = self.__narwhals_namespace__()
         if order_by is None:
diff --git a/narwhals/_compliant/dataframe.py b/narwhals/_compliant/dataframe.py
index d93b11b71..583871514 100644
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -486,3 +486,5 @@ class EagerDataFrame(
                 assert_never(rows)
 
         return compliant
+
+    def to_struct(self, name: str = "") -> EagerSeriesT: ...
diff --git a/narwhals/_compliant/expr.py b/narwhals/_compliant/expr.py
index 8e6e8b5fb..0c78def12 100644
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -511,6 +511,28 @@ class EagerExpr(
     def cast(self, dtype: IntoDType) -> Self:
         return self._reuse_series("cast", dtype=dtype)
 
+    def hist_from_bins(self, bins: list[float], *, include_breakpoint: bool) -> Self:
+        method_name = "hist_from_bins"
+
+        def fn(df: EagerDataFrameT) -> Sequence[EagerSeriesT]:
+            series_in = self(df)
+            if len(series_in) > 1:
+                raise NotImplementedError
+            df_result = series_in[0].hist_from_bins(
+                bins=bins, include_breakpoint=include_breakpoint
+            )
+            aliases = self._evaluate_aliases(df)
+            return [df_result.to_struct(aliases[0])]
+
+        return self._from_callable(
+            fn,
+            depth=self._depth + 1,
+            function_name=f"{self._function_name}->{method_name}",
+            evaluate_output_names=self._evaluate_output_names,
+            alias_output_names=self._alias_output_names,
+            context=self,
+        )
+
     def __eq__(self, other: Self | Any) -> Self:  # type: ignore[override]
         return self._reuse_series("__eq__", other=other)
 
diff --git a/narwhals/_compliant/series.py b/narwhals/_compliant/series.py
index 22f2f5f5a..4324fb503 100644
--- a/narwhals/_compliant/series.py
+++ b/narwhals/_compliant/series.py
@@ -34,7 +34,7 @@ if TYPE_CHECKING:
     import pyarrow as pa
     from typing_extensions import Self
 
-    from narwhals._compliant.dataframe import CompliantDataFrame
+    from narwhals._compliant.dataframe import CompliantDataFrame, EagerDataFrame
     from narwhals._compliant.expr import CompliantExpr, EagerExpr
     from narwhals._compliant.namespace import CompliantNamespace, EagerNamespace
     from narwhals._utils import Implementation, Version, _LimitedContext
@@ -350,6 +350,13 @@ class EagerSeries(CompliantSeries[NativeSeriesT], Protocol[NativeSeriesT]):
         else:
             assert_never(item)
 
+    @unstable
+    def hist_from_bins(
+        self, bins: list[float], *, include_breakpoint: bool
+    ) -> EagerDataFrame[Self, Any, Any, NativeSeriesT]:
+        """`Series.hist(bins=..., bin_count=None)`."""
+        ...
+
     @property
     def str(self) -> EagerSeriesStringNamespace[Self, NativeSeriesT]: ...
     @property
diff --git a/narwhals/expr.py b/narwhals/expr.py
index 595fde8c1..0f278ffaf 100644
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -2563,6 +2563,15 @@ class Expr:
         """
         return self._with_elementwise(lambda plx: self._to_compliant_expr(plx).sqrt())
 
+    def hist_from_bins(
+        self, bins: list[float], *, include_breakpoint: bool = True
+    ) -> Self:
+        return self._with_orderable_filtration(
+            lambda plx: self._to_compliant_expr(plx).hist_from_bins(  # type: ignore[attr-defined]
+                bins=bins, include_breakpoint=include_breakpoint
+            )
+        )
+
     @property
     def str(self) -> ExprStringNamespace[Self]:
         return ExprStringNamespace(self)

```

</p>
</details> 

And that looks like:

```py
import narwhals as nw
import pyarrow as pa

data: dict[str, Any] = {"a": [1, 3, 8, 8, 2, 1, 3]}
bins: list[float] = [1, 2, 3]

df = pa.table(data)
nw_df = nw.from_native(df)
>>> nw_df.select(nw.col("a").hist_from_bins(bins)) # ignore the name for now!
┌──────────────────────────────────────────┐
|            Narwhals DataFrame            |
|------------------------------------------|
|pyarrow.Table                             |
|a: struct<breakpoint: int64, count: int64>|
|  child 0, breakpoint: int64              |
|  child 1, count: int64                   |
|----                                      |
|a: [                                      |
|  -- is_valid: all not null               |
|  -- child 0 type: int64                  |
|[2,3]                                     |
|  -- child 1 type: int64                  |
|[3,2]]                                    |
└──────────────────────────────────────────┘
```

### `polars`
For this, I would rip out the version branching from `PolarsSeries.hist`, and use it for `PolarsExpr.hist`.

`PolarsSeries.hist` would work in the **opposite direction** to the eager-only ones.
Pretty much the same idea as (https://github.com/narwhals-dev/narwhals/pull/2598#discussion_r2156998509). We should be doing more of that anyway 🙂

dangotbanned:
> How should we decide what to keep at the narwhals/compliant level and what to move to this class?

([refactor(suggestion): Try out generalizing everything](https://github.com/narwhals-dev/narwhals/pull/2839/commits/29263cda39dec1c7092630835428e4269b9cd462)) collects everything into that class to make the diff easier 😅 

I think these parts are the same across backends, so to `narwhals`-level

https://github.com/narwhals-dev/narwhals/blob/1334849272c21b6903270d247b6c37a7b7aa2cf0/narwhals/_pandas_like/series.py#L1103-L1106

https://github.com/narwhals-dev/narwhals/blob/1334849272c21b6903270d247b6c37a7b7aa2cf0/narwhals/_pandas_like/series.py#L1112-L1115


Anywhere I left a note like this, *could* become that method:

https://github.com/narwhals-dev/narwhals/blob/1334849272c21b6903270d247b6c37a7b7aa2cf0/narwhals/_pandas_like/series.py#L1125-L1126

https://github.com/narwhals-dev/narwhals/blob/1334849272c21b6903270d247b6c37a7b7aa2cf0/narwhals/_pandas_like/series.py#L1143-L1144


Things like this are dependent on the above

https://github.com/narwhals-dev/narwhals/blob/1334849272c21b6903270d247b6c37a7b7aa2cf0/narwhals/_pandas_like/series.py#L1095-L1097

I think a good example of splitting things across layers to implement is the work ~~we~~ *mostly Marco* did on slicing.

Its all about trying to spot the cases where we're repeating ourselves 😄 

<details><summary><code>DataFrame.__getitem__</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/80c5fc2a001bc9cdc990ad4f188708eea36f810e/narwhals/dataframe.py#L877-L926

</p>
</details> 


<details><summary><code>EagerDataFrame.__getitem__</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/80c5fc2a001bc9cdc990ad4f188708eea36f810e/narwhals/_compliant/dataframe.py#L439-L489

</p>
</details>

narwhals/_arrow/series.py
camriddell: Apologies for the delay in reviewing here. In case this is still an open question:
- `obs_*` refers to "observed" values, whereas `obj_*` refers to "objective" values. In this context observed values are ones that are encountered in the data, whereas objective values are ones that are requested to be generated. (e.g. if I have bins from `[1, 3), [3, 5), [5, 7]` and have values of `[1, 2, 3]` then the final bin range of `[5, 7]` would be objective but not observed.

```python
counts = np.zeros_like(obj_cats)
counts[np.isin(obj_cats, obs_cats)] = obs_counts[np.isin(obs_cats, obj_cats)]
```

Here we're trying to align the counts against the categories that we actually observed (thus not incrementing the counts for unobserved categories). This approach works because `obs_counts` and `obs_cats` have the same shape and are similarly unsorted, so I am simply locating the observed counts of categories that will be included in the output.

Honestly, it is probably more straightforward to convert these values to an Arrow Table and execute a semi-join between `bin_indices_pa.value_counts()` and `obj_cats`, unpack the struct result, and rename the fields. I tried to keep everything in "array-land" and it make some code that is tricky to maintain.

narwhals/_polars/series.py
```diff
@@ -535,19 +559,21 @@ def hist(  # noqa: C901, PLR0912
             include_breakpoint=include_breakpoint,
         )
 
+        # Apply post-processing corrections
+
+        # Handle column naming
         if not include_breakpoint:
             df.columns = ["count"]
```
dangotbanned:
```suggestion
            df = df.select(pl.nth(0).alias("count"))
```

narwhals/_polars/series.py
```diff
@@ -535,19 +559,21 @@ def hist(  # noqa: C901, PLR0912
...
+        # polars<1.15 implicitly adds -inf and inf to either end of bins
+        if BACKEND_VERSION < (1, 15) and bins is not None:  # pragma: no cover
             r = pl.int_range(0, len(df))
             df = df.filter((r > 0) & (r < len(df) - 1))
 
-        # polars<1.27 makes the lowest bin a left/right closed interval.
-        if self._backend_version < (1, 27) and bins is not None:
+        # polars<1.27 makes the lowest bin a left/right closed interval
+        if BACKEND_VERSION < (1, 27) and bins is not None:  # pragma: no cover
             df[0, "count"] += (series == bins[0]).sum()
```
dangotbanned: This is half suggestion, half question.

### Question
What inputs do you need to use for the `1.15` path to return anything?

Everything I've tried returned a empty dataframe 🤔

### Suggestion
Sure the `1.27` path is longer here, but the current combo makes for a challenging read for me 😔

- Row + column indexing
- Setting inplace
- Masking with a `Series`

```suggestion
        if bins is not None:
            # polars<1.15 implicitly adds -inf and inf to either end of bins
            if BACKEND_VERSION < (1, 15):  # pragma: no cover
                r = pl.int_range(0, len(df))
                df = df.filter((r > 0) & (r < len(df) - 1))

            # polars<1.27 makes the lowest bin a left/right closed interval
            if BACKEND_VERSION < (1, 27):  # pragma: no cover
                df = (
                    df.slice(0, 1)
                    .with_columns(pl.col("count") + ((pl.lit(series) == bins[0]).sum()))
                    .vstack(df.slice(1))
                )
```

narwhals/_polars/series.py
dangotbanned: Expressify as much as possible!

narwhals/_polars/series.py
FBruzzesi: I could reproduce it with polars<1.0 (see [docs example](https://docs.pola.rs/api/python/version/0.20/reference/series/api/polars.Series.hist.html)). We should figure out when this behavior change. I don't think it's 1.15 though.

Update: switch happened from 1.5 to 1.6

narwhals/_polars/series.py
FBruzzesi: I just checked the min version 0.20.4, and that is consistent, so it's even better: we don't need to rename.

Edit: or not! WTF is going one lol

dangotbanned (approved):
Thanks for addressing everything in (https://github.com/narwhals-dev/narwhals/pull/2839#pullrequestreview-3052304501) @FBruzzesi 

hungry narwhal

narwhals/_arrow/series.py
dangotbanned: @camriddell 

> Apologies for the delay in reviewing here. In case this is still an open question:

Apologies right back at ya for losing this 😅 

> Honestly, it is probably more straightforward to convert these values to an Arrow Table and execute a semi-join between `bin_indices_pa.value_counts()` and `obj_cats`, unpack the struct result, and rename the fields. I tried to keep everything in "array-land" and it make some code that is tricky to maintain.


Okay I'll bite 😄

I'd be very interested in a lil PR with that!
Personally I find it much easier to think in terms of tables than arrays and had no idea this was similar to a `pa.Table(join_type='left semi' | 'right semi')`

PR #2856, opened 2025-07-19
feat: Allow `all` and `any` in group by context

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2385

## Checklist

- [x] Tests added
- [ ] Documented the changes

Opening as draft: I am expecting some old versions to break

### DIFF
--- a/narwhals/_dask/group_by.py
+++ b/narwhals/_dask/group_by.py
@@ -44,6 +44,26 @@ def n_unique() -> dd.Aggregation:
     return dd.Aggregation(name="nunique", chunk=chunk, agg=agg)
 
 
+def _all() -> dd.Aggregation:
+    def chunk(s: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s.all(skipna=True)
+
+    def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s0.all(skipna=True)
+
+    return dd.Aggregation(name="all", chunk=chunk, agg=agg)
+
+
+def _any() -> dd.Aggregation:
+    def chunk(s: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s.any(skipna=True)
+
+    def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s0.any(skipna=True)
+
+    return dd.Aggregation(name="any", chunk=chunk, agg=agg)
+
+
 def var(ddof: int) -> _AggFn:
     return partial(_DaskGroupBy.var, ddof=ddof)
 
@@ -65,6 +85,8 @@ class DaskLazyGroupBy(DepthTrackingGroupBy["DaskLazyFrame", "DaskExpr", Aggregat
         "n_unique": n_unique,
         "count": "count",
         "quantile": "quantile",
+        "all": _all,
+        "any": _any,
     }
 
     def __init__(
--- a/tests/frame/group_by_test.py
+++ b/tests/frame/group_by_test.py
@@ -141,6 +141,39 @@ def test_group_by_depth_1_agg(
 
 
 @pytest.mark.parametrize(
+    ("values", "expected"),
+    [
+        (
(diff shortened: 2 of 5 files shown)


narwhals/_dask/group_by.py
```diff
@@ -44,6 +44,26 @@ def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
...
+
+
+def _any() -> dd.Aggregation:
+    def chunk(s: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s.any(skipna=True)
+
+    def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
+        return s0.any(skipna=True)
+
+    return dd.Aggregation(name="any", chunk=chunk, agg=agg)
```
FBruzzesi, line 64: I was _so_ surprised when all and any resulted in "unknown aggregate"

```py
import dask.dataframe as dd

data = {"a": [1,1, 2, 2], "b": [True, True, True, False], "c": [True, False, False, False]}
frame = dd.from_dict(data, 2)

frame.groupby("a").agg({"b": "all"})
> ValueError: unknown aggregate all
```

dangotbanned: I think that's raising here?

https://github.com/dask/dask/blob/3f6051ae242f894af0f43fd9de7ad7131fb3e70d/dask/dataframe/groupby.py#L764-L805

If so, these would be the one's we can use:

```py
simple_impl = {
        "sum": (M.sum, M.sum),
        "min": (M.min, M.min),
        "max": (M.max, M.max),
        "count": (M.count, M.sum),
        "size": (M.size, M.sum),
        "first": (M.first, M.first),
        "last": (M.last, M.last),
        "prod": (M.prod, M.prod),
        "median": (
            None,
            M.median,
        ),  # No chunk func for median, we can only take it when aggregating
    }
```

FBruzzesi: > I think that's raising here?

Yep!

> Otherwise, we might need to use [dd.Aggregation](https://docs.dask.org/en/stable/generated/dask.dataframe.Aggregation.html#dask.dataframe.Aggregation)

That's what I did 😉 `_any` and `_all` return a `dd.Aggregation`

dangotbanned: >>  Otherwise, we might need to use [dd.Aggregation](https://docs.dask.org/en/stable/generated/dask.dataframe.Aggregation.html#dask.dataframe.Aggregation)

> That's what I did 😉

haha yeah I realised that almost right after I commented 🫣

tests/frame/group_by_test.py
```diff
@@ -140,6 +140,39 @@ def test_group_by_depth_1_agg(
...
+@pytest.mark.parametrize(
+    ("values", "expected"),
+    [
+        (
+            {"x": [True, True, True, False, False, False]},
+            {"all": [True, False, False], "any": [True, True, False]},
+        ),
+        (
+            {"x": [True, None, False, None, None, None]},
+            {"all": [True, False, True], "any": [True, False, False]},
```
FBruzzesi, line 152: The reason to split into two tests (instead of just multiple groups) is mainly to have better control over different backend behavior. Honestly I am surprised that the only one that does not match polars behavior for the nullable case is Dask (which we could actually control since we are implementing the aggregation ourselves)

dangotbanned: > Honestly I am surprised that the only one that does not match polars behavior for the nullable case is Dask


I'm very jealous of you right now 😅 (https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2217491423)

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #2288, opened 2025-03-25
feat: Add `DataFrame.clone` for `pyarrow`

dangotbanned:
Closes #2207
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2207
- Mentioned in https://github.com/narwhals-dev/narwhals/pull/1644#issuecomment-2750891356

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [ ] Documented the changes

PR #2861, opened 2025-07-21
fix: pyarrow clone returned same native object

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -591,7 +591,7 @@ class ArrowDataFrame(
         raise AssertionError(msg)  # pragma: no cover
 
     def clone(self) -> Self:
-        return self._with_native(self.native, validate_column_names=False)
+        return self._with_native(pa.table(self.native), validate_column_names=False)
 
     def item(self, row: int | None, column: int | str | None) -> Any:
         from narwhals._arrow.series import maybe_extract_py_scalar
--- a/tests/frame/clone_test.py
+++ b/tests/frame/clone_test.py
@@ -11,6 +11,7 @@ def test_clone(constructor_eager: ConstructorEager) -> None:
     df_clone = df.clone()
     assert df is not df_clone
     assert df._compliant_frame is not df_clone._compliant_frame
+    assert df.to_native() is not df_clone.to_native()
     assert_equal_data(df_clone, expected)
     df_clone_mod = df_clone.with_columns((nw.col("a") + nw.col("b")).alias("c"))
     assert_equal_data(df, expected)


narwhals/_arrow/dataframe.py
```diff
@@ -591,7 +591,14 @@ def collect(
         raise AssertionError(msg)  # pragma: no cover
 
     def clone(self) -> Self:
-        return self._with_native(self.native, validate_column_names=False)
+        native = self.native
+        if self._backend_version < (14, 0, 0):
+            cloned_table = pa.table(
+                {col: pa.array(native.column(col)) for col in native.column_names}
```
FBruzzesi: Doing:

```suggestion
                {col: native.column(col) for col in native.column_names}
```

triggers the type-checker 😩

>Argument of type "dict[str, ChunkedArray[Any]]" cannot be assigned to parameter "data" of type "Collection[ArrayOrChunkedArray[Any]] | DataFrame | SupportArrowArray | SupportArrowStream | SupportArrowDeviceArray" in function "table"

MarcoGorelli:
i don't think pyarrow table is mutable anyway, so i don't think this should have an effect? or is there some case where it makes a difference?

FBruzzesi:
> i don't think pyarrow table is mutable anyway, so i don't think this should have an effect? or is there some case where it makes a difference?

That's a fair point! I am looking at great-tables codebase and they have a check in their test suite along the lines of `df is clone(df)`, which would fail.

Taking a deeper look, I could not find any similar check in the library itself 🧐

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2861#issuecomment-3095550672

Agreed, I did mention this in the PR 😉

- https://github.com/narwhals-dev/narwhals/pull/2288

dangotbanned:
> > i don't think pyarrow table is mutable anyway, so i don't think this should have an effect? or is there some case where it makes a difference?
> 
> That's a fair point! I am looking at great-tables codebase and they have a check in their test suite along the lines of `df is clone(df)`, which would fail.
> 

Maybe this was something they added thinking `pyarrow` was more similar to `pandas`?

FBruzzesi:
If the check performed is `x is nw.from_native(x).clone().to_native()`, then in our current implementation we would have a different result for pyarrow vs pandas and polars. But as I said, I could not find any such check in the library itself, but it's one test in their test suite

FBruzzesi:
Closing this as there is no complete agreement on what clone should return.

As mentioned in https://github.com/narwhals-dev/narwhals/pull/2861#issuecomment-3095578760, there is one possible downstream case for which the `x is x.clone()` is performed and the current implementation for pyarrow would fail.

PR #2873, opened 2025-07-22
feat: Deprecate `Expr.arg_min` and `Expr.arg_max` from the main namespace

### LINKED ISSUE #2863: API: deprecate `Expr.arg_max` and `Expr.arg_min`, only keep `Series.arg_min` and `Series.arg_max`
Been thinking about https://github.com/narwhals-dev/narwhals/issues/2526

I don't think orderable aggregations make much sense. There's two cases for them:
- returning an aggregate, e.g. in groupby-agg
- make a new column, e.g. with `over`

If people are forced to use these in combination with `over` in the lazy case, then the first one becomes impossible

So, my proposal is:
- first, deprecate `Expr.arg_max` and `Expr.arg_min`. Seeing as they're eager-only, `Series.arg_max` and `Series.arg_min` should be fine
- later, we could consider adding `arg_max_by` / `arg_min_by`. Polars doesn't have these, but it does have several methods in this spirit (`rolling_max_by`, `interpolate_by`, ...), so `min_by` / `max_by` / `arg_min_by` / `arg_max_by` still feel idiomatic

MarcoGorelli:
closes #2863 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


dangotbanned:
Would there be more of a benefit to keeping these if they worked with `over(order_by=...)`?

https://github.com/narwhals-dev/narwhals/blob/9edd5ea975c7d8358ad779737a7b0fe202eba140/tests/expr_and_series/arg_min_test.py#L42-L55

The equivalent to that is still an issue for `first`, `last` (https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2222598488) - but we can fix that *somewhere* in the broadcasting logic 🙂 

I've never used these methods before, so no love lost either way 😄

MarcoGorelli:
I don't think so, if we had `arg_min_by` it could be used in both contexts. Likewise, `min_by` could replace `first`

PR #2876, opened 2025-07-23
chore: factor out sqrt from SparkLike, DuckDB, and Ibis

ym-pett:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

I've left the commented out code in as it's a draft. All the tests pass with both of the return statements in the sqrt() functions.

however, IDE linting indicates there's a problem with the `_where` statement, could it be that I'm not defining the return type for NULL correctly?

thanks for looking at this as I've hit max confusion. 

```

### CODE REVIEW

File: narwhals/_sql/expr.py [hunk 1]
```diff
@@ -386,6 +386,12 @@ def round(self, decimals: int) -> Self:
             lambda expr: self._function("round", expr, self._lit(decimals))
         )
 
+    def sqrt(self) -> Self:
+        # return self._with_elementwise(lambda expr: self._function("sqrt", expr))
+        return self._when(
+            self < 0, NULL, self._function(lambda expr: self._function("sqrt", expr))
+        )
```
MarcoGorelli: you need to define an inner function (either with `lambda expr: `, or by defining a function within this function's body), and then use that in the condition you pass to `_when`

also, you should use `float('nan')` here, not `NULL`, they're different. there's quite a few resources on this, but this one https://www.dontusethiscode.com/blog/2025-01-08_null_vs_nan.html is by narwhals dev @camriddell so i'll recommend that one!

MarcoGorelli: if you take the function you commented out from `DuckDBExpr`, you can make something very similar here, you'll just need to use `self._when` / `self._lit` instead of the duckdb-specific when/lit ones

ym-pett: thanks will have a go!

narwhals/_sql/expr.py
```diff
@@ -385,7 +385,29 @@ def round(self, decimals: int) -> Self:
...
+        return self._with_elementwise(lambda expr: self._function("isnull", expr))
+    
+    # Self: because we're giving it a column and return a column?
+    def sqrt(self) -> Self: 
+        # Native.. type because we're passing it a native function?, my instinct was to pass 'self' too, but that didn't work either
+        def _sqrt(expr: NativeExprT) -> NativeExprT: 
+            # why _lit below? is it because they're fixed. I can see that 'self' is wrong, but didn't find how to fix
+            return self._when(self < self._lit(0), self._lit(float("nan")), self._with_elementwise(lambda expr: self._function("sqrt", expr)))
+
+        return self
```
MarcoGorelli: `Self` means that it returns an object of the same class as the input. So for example if you have `expr` and `expr` is of type `SQLExpr`, then `expr.sqrt()` is also of type `SQLExpr`

> I can see that 'self' is wrong, but didn't find how to fix

I'd suggest taking a closer look at the code in `DuckDBExpr` which you commented out

MarcoGorelli (commented):
thanks for updating! left another comment, we'll get there 💪

narwhals/_sql/expr.py
```diff
@@ -385,6 +385,19 @@ def round(self, decimals: int) -> Self:
...
+    #     def _sqrt(expr: Expression) -> Expression:
+    #         return when(expr < lit(0), lit(float("nan"))).otherwise(F("sqrt", expr))
+
+    #     return self._with_elementwise(_sqrt)    
+    
+    def sqrt(self) -> Self:
+        def _sqrt(expr: NativeExprT) -> NativeExprT:
+            return self._when(expr < self._lit(0), self._lit(float("nan")), self._with_elementwise(lambda expr: self._function("sqrt", expr)))
+
+        return self._with_elementwise(_sqrt)
```
MarcoGorelli: you're changing more than you need to, note how the model from duckdb only has a single `_with_elementwise`, but you have two

narwhals/_sql/expr.py
```diff
@@ -386,6 +386,21 @@ def round(self, decimals: int) -> Self:
...
+
+    #     return self._with_elementwise(_sqrt)
+
+    def sqrt(self) -> Self:
+        def _sqrt(expr: NativeExprT) -> NativeExprT:
+            return self._when(
+                self < self._lit(0), self._lit(float("nan"))
+                )
+
+        return self._with_elementwise(lambda expr: self._function("_sqrt", expr))
```
MarcoGorelli: this still differs too much from the `sqrt_duckdb` one you've pasted above

let's fix the type hints later

first, I'd suggest:
- copy-and-paste the duckdb one as-is
- change `lit` to `self._lit`
- change `when` to `self._when`
- change `F(` to `self._function(`

ym-pett: thanks Marco! Followed your suggestion but It doesn't recognise the 'otherwise' - could you give me an idea of where I could find an equivalent statement, simple 'else' doesn't work here

narwhals/_duckdb/expr.py
```diff
@@ -73,6 +73,16 @@ def __init__(
...
 
+    def _when(
+        self,
+        condition: Expression,
+        value: Expression,
+        otherwise: Expression | None = None,
+    ) -> Expression:
+        if otherwise is None:
+            return when(condition, value)
+        return when(condition, value).otherwise(otherwise)
```
MarcoGorelli: `_when` is defined in `_duckdb/namespace, can you move this there?

narwhals/_spark_like/expr.py
```diff
@@ -72,6 +72,13 @@ def __init__(
     def _count_star(self) -> Column:
         return self._F.count("*")
 
+    def _when(
+        self, condition: Column, value: Column, otherwise: ExprKind | None = None
+    ) -> Column:
+        if otherwise is None:
+            return self._F.when(condition, value)
+        return self._F.when(condition, value).otherwise(otherwise)
```
MarcoGorelli: same

narwhals/_spark_like/expr.py
```diff
@@ -387,13 +394,11 @@ def _log(expr: Column) -> Column:
...
         return self._with_elementwise(_log)
 
-    def sqrt(self) -> Self:
-        def _sqrt(expr: Column) -> Column:
-            return self._F.when(expr < 0, self._F.lit(float("nan"))).otherwise(
-                self._F.sqrt(expr)
-            )
+    def exp(self) -> Self:
+        def _exp(expr: Column) -> Column:
+            return self._F.exp(expr)
```
MarcoGorelli: not sure why `exp` is being added back here, perhaps an error while resolving merge conflicts?

narwhals/_sql/expr.py
```diff
@@ -181,14 +181,18 @@ def _function(self, name: str, *args: NativeExprT | PythonLiteral) -> NativeExpr
...
         return self.__narwhals_namespace__()._coalesce(*expr)
 
     def _count_star(self) -> NativeExprT: ...
 
+    def _when(
+        self,
+        condition: NativeExprT,
+        value: NativeExprT,
+        otherwise: NativeExprT | None = None,
+    ) -> NativeExprT: ...
```
MarcoGorelli: instead of `: ...`, can we make this one return
```
return self.__narwhals_namespace__()._when(condition, value, otherwise)
```

and then add update the definition in `_sql/namespace.py` to have `otherwise`?

narwhals/_sql/expr.py
```diff
@@ -492,6 +496,16 @@ def round(self, decimals: int) -> Self:
...
 
+    def sqrt(self) -> Self:
+        def _sqrt(expr: NativeExprT) -> NativeExprT:
+            return self._when(
+                expr < self._lit(0),  # type: ignore[operator]
+                self._lit(float("nan")),
+                self._function("sqrt", expr),
+            )
+
+        return self._with_elementwise(_sqrt)
```
MarcoGorelli, line 508: nice!

narwhals/_spark_like/expr.py
ym-pett: oh yep that crept back in! will do

MarcoGorelli:
thanks for updating and resolving conflicts, nice!

i think all that remains is to update `_when` in `_ibis/namespace` to also follow the same signature (even though we're not overwriting `sqrt` for ibis, `_when` is used in other places)

ym-pett:
> thanks for updating and resolving conflicts, nice!
> 
> i think all that remains is to update `_when` in `_ibis/namespace` to also follow the same signature (even though we're not overwriting `sqrt` for ibis, `_when` is used in other places)

great will do!

narwhals/_ibis/namespace.py
```diff
@@ -51,8 +51,12 @@ def _function(self, name: str, *args: ir.Value | PythonLiteral) -> ir.Value:
...
         return lit(value)
 
-    def _when(self, condition: ir.Value, value: ir.Value) -> ir.Value:
-        return ibis.cases((condition, value))
+    def _when(
+        self, condition: ir.Value, value: ir.Value, otherwise: ir.Expr | None = None
+    ) -> ir.Value:
+        if otherwise is None:
+            return ibis.cases((condition, value))
+        return ibis.cases((condition, value), else_=otherwise)
```
MarcoGorelli: looks like we don't (yet) have a test case which covers this for ibis yet, so you could add `# pragma: no cover` to the end this line (59)

narwhals/_spark_like/namespace.py
```diff
@@ -73,8 +74,12 @@ def _function(self, name: str, *args: Column | PythonLiteral) -> Column:
     def _lit(self, value: Any) -> Column:
         return self._F.lit(value)
 
-    def _when(self, condition: Column, value: Column) -> Column:
-        return self._F.when(condition, value)
+    def _when(
+        self, condition: Column, value: Column, otherwise: ExprKind | None = None
```
MarcoGorelli: `ExprKind` looks like a typo?

MarcoGorelli (commented):
looks close!

narwhals/_spark_like/namespace.py
ym-pett: i'll check!

MarcoGorelli:
In fact, now that you've made a common interface for `_when` which includes `otherwise`...perhaps it'll also work to remove the `_sqrt` implementation from `_ibis/expr.py` too? Then you wouldn't need the `pragma: no cover`

ym-pett:
> In fact, now that you've made a common interface for `_when` which includes `otherwise`...perhaps it'll also work to remove the `_sqrt` implementation from `_ibis/expr.py` too? Then you wouldn't need the `pragma: no cover`

oh cool, I'll try that out

ym-pett:
was able to remove ibis sqrt() from the ibis/expr file.

I still have an outstanding type error in the type checking for the sparklike namespace, currently trying to understand

PR #2879, opened 2025-07-24
feat: Accept kwargs in `{DataFrame,Series}.to_pandas`

dangotbanned:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2123
- Originated from https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3110332853
- Extra context (https://github.com/narwhals-dev/narwhals/issues/2123#issuecomment-3120531391)

<details><summary><b>Show original description</b></summary>
<p>

## Description
While I was looking into (https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3110332853) I noticed that the default (and only) behavior for `nw.DataFrame.to_pandas` when backed by either `pa.Table`, `pl.DataFrame` isn't great for nested datatypes:

```py
import pyarrow as pa

data = {"breakpoint": [2, 3], "count": [3, 2]}

>>> pa.table(data).to_struct_array().to_pandas()
0    {'breakpoint': 2, 'count': 3}
1    {'breakpoint': 3, 'count': 2}
dtype: object
```

We **can** do better than this, and *currently* this PR will instead give us something that works with [`pd.Series.struct`](https://pandas.pydata.org/docs/reference/series.html#struct-accessor):

```py
import pyarrow as pa
import narwhals as nw

data = {"breakpoint": [2, 3], "count": [3, 2]}

>>> nw.from_native(pa.table(data).to_struct_array(), series_only=True).to_pandas()
0    {'breakpoint': 2, 'count': 3}
1    {'breakpoint': 3, 'count': 2}
Name: , dtype: struct<breakpoint: int64, count: int64>[pyarrow]
```

## Questions
1. Would we want this for any other `DType`s e.g. `List`?
2. Would it be worthwhile to expose this as the [`to_pandas(use_pyarrow_extension_array=...)`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.to_pandas.html) parameter from `polars`?
  i. Note that this defaults to `False`, which is equivalent to our current behavior *and* `pyarrow`'s

</p>
</details> 

Adds optional keyword-only arguments to both of `DataFrame.to_pandas`, `Series.to_pandas`.

This aligns us with the same methods found in `polars` and `pyarrow`:

- [`polars.Series.to_pandas`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.to_pandas.html)
- [`pyarrow.ChunkedArray.to_pandas`](https://arrow.apache.org/docs/python/generated/pyarrow.ChunkedArray.html#pyarrow.ChunkedArray.to_pandas)
- [`polars.DataFrame.to_pandas`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_pandas.html)
- [`pyarrow.Table.to_pandas`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas)

As mentioned in (#2123), this can reduce the memory overhead if configured correctly.

**But**, the main benefit I'm excited about is that we can preserve `pyarrow` data types (nulls, nested data) see (https://github.com/narwhals-dev/narwhals/issues/2123#issuecomment-3120531391) - which were previously lost **unconditionally**

## Example

```py
import pyarrow as pa

import narwhals as nw

data = {"breakpoint": [2, 3], "count": [3, 2], "what": [None, 1]}
native = pa.table(data).to_struct_array()
series = nw.from_native(native, series_only=True)
```

### Before
```py
>>> series.to_pandas()
0    {'breakpoint': 2, 'count': 3, 'what': None}
1     {'breakpoint': 3, 'count': 2, 'what': 1.0}
Name: , dtype: object
```

### After
```py
>>> series.to_pandas(use_pyarrow_extension_array=True)
0    {'breakpoint': 2, 'count': 3, 'what': None}
1     {'breakpoint': 3, 'count': 2, 'what': 1.0}
Name: , dtype: struct<breakpoint: int64, count: int64, what: int64>[pyarrow]
```

## Tasks
- [x] Update `Compliant*` signatures
- [x] Update `Compliant*` runtime behavior
  - [x] `Arrow(DataFrame|Series)`
  - [x] `Polars(DataFrame|Series)`
  - [x] `PandasLike(DataFrame|Series)`
    - https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.dataframe.to_pandas/#cudf.DataFrame.to_pandas
    - https://docs.rapids.ai/api/cudf/stable/user_guide/api_docs/api/cudf.series.to_pandas/#cudf.Series.to_pandas
    - [feat: Partially utilize for `cuDF`](https://github.com/narwhals-dev/narwhals/pull/2879/commits/7d6ec2ab515995fa77be6ae4467b927d9de0f079)
- [x] Update `DataFrame.to_pandas`
- [x] Update `Series.to_pandas`
  - [x] Signature, pass down to compliant
  - [x] Doc 
    - [x] Description (https://github.com/narwhals-dev/narwhals/pull/2879/commits/49cb5c6f513b6d62f70d01056585d26f32d07816)
    - [x] Example
- [x] Replace earlier tests (https://github.com/narwhals-dev/narwhals/pull/2879/commits/0951477069b771e1affd2cb054050e2e94fa6b3f), (https://github.com/narwhals-dev/narwhals/pull/2879/commits/9e3d2950a785704ac577aba4f898a0236db8695a)
  - [x] [test: Update `DataFrame.to_pandas`](https://github.com/narwhals-dev/narwhals/pull/2879/commits/9e7e7762ba5ddce59f72de14834557446c274d34)
  - [x] `Series.to_pandas`
  - [x] Find which `pyarrow` added `Table.to_struct_array`
    - https://github.com/narwhals-dev/narwhals/actions/runs/16556200639/job/46817960571?pr=2879
    - [fix: Add `to_struct_array` compat for `pyarrow<15`](https://github.com/narwhals-dev/narwhals/pull/2879/commits/636b4acd67e6b36272051479155230e42e3eda3c)
- [x] Test `polars` and `pandas` as well
  - [x] `Series`
  - [x] `DataFrame`
- [x] Update PR description to match the updated PR since (https://github.com/narwhals-dev/narwhals/pull/2879#pullrequestreview-3056885164)
- [x] Investigate `PandasLike` alternatives (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2249323516)


narwhals/_arrow/utils.py
```diff
@@ -226,6 +227,33 @@ def narwhals_to_native_dtype(dtype: IntoDType, version: Version) -> pa.DataType:
...
+    import pandas as pd  # ignore-banned-import
+
+    if Implementation.PANDAS._backend_version() < (1, 5):
+        # `pd.ArrowDType` added https://github.com/pandas-dev/pandas/pull/46774
+        return None
+    # NOTE: 2.2.0 added the accessors, but probably usable earlier than that?
+    # `.struct` added https://github.com/pandas-dev/pandas/pull/54977
+    # `.list` added https://github.com/pandas-dev/pandas/pull/55777
+    if pa.types.is_struct(dtype):
+        return pd.ArrowDtype(dtype)
```
FBruzzesi: Should we consider all these?

https://github.com/narwhals-dev/narwhals/blob/f8f7f65e5bb82580f373cf8c2200e89e8febbdc1/narwhals/_pandas_like/utils.py#L547-L548

FBruzzesi (commented):
Thanks for spotting this @dangotbanned , it seems reasonable to me!
Yet if we do it, I would rather keep it consistent for all the dtypes that would only be supported with ArrowDtype's. WDYT?

narwhals/_arrow/utils.py
dangotbanned: Yeah I would actually rewrite this all more closely following the `polars` version now

https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/dataframe/frame.py#L2531-L2611

I guess the most important bit is this line:

```py
types_mapper=lambda pa_dtype: pd.ArrowDtype(pa_dtype)
```

I started the PR discovering the argument in the `pyarrow` docs, but then found the `polars` version and #2123

narwhals/typing.py
```diff
@@ -411,6 +413,74 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+    strings_to_categorical: Literal[True]
+    """Cast all `String`, `Binary` types to `pandas.Categorical`."""
+    zero_copy_only: Literal[True]
+    """Raise an `ArrowException` if this function call would require copying the underlying data."""
+    integer_object_nulls: Literal[True]
+    """Cast `Int`, `UInt` types with nulls to `numpy.dtypes.ObjectDType`."""
+    date_as_object: Literal[False]
+    """Cast `Date` types to `numpy.dtypes.ObjectDType`.
+
+    If False, convert to datetime64 dtype with the equivalent time unit (if supported).
```
dangotbanned: ### Previews
<details><summary>By default, <code>to_pandas</code> will only show this</summary>
<p>

image

</p>
</details> 


<details><summary>Once you go beyond <code>use_pyarrow_extension_array</code>, then each description gets its own context</summary>
<p>

image

image

</p>
</details> 

<details><summary>API Reference</summary>
<p>

image

</p>
</details>

narwhals/typing.py
```diff
@@ -411,6 +413,72 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
 
+class ToPandasArrowKwds(TypedDict, total=False):
+    """Keyword arguments to be passed to [`pyarrow.Table.to_pandas`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas).
+
+    Important:
+        Only *default-overriding* values are documented.
+    """
+
+    memory_pool: pa.MemoryPool
+    categories: list[str]
```
dangotbanned: I'm not sure if the way I've typed this is intuitive, no issue with just sticking with what was there before (https://github.com/narwhals-dev/narwhals/pull/2879/commits/bc09ebffb2cf2b570dec3b04465b29fd1c0df3c8)


The idea was to warn when you're passing in arguments that are equivalent to passing no arguments.

So in this case, only `strings_to_categorical` and `zero_copy_only` will do anything **because** the default has been overriden

image

narwhals/_arrow/utils.py
```diff
@@ -230,6 +231,22 @@ def narwhals_to_native_dtype(dtype: IntoDType, version: Version) -> pa.DataType:
...
+    Should be passed as `types_mapper` in [`Table.to_pandas`] and [`ChunkedArray.to_pandas`].
+
+    [`Table.to_pandas`]: https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas
+    [`ChunkedArray.to_pandas`]: https://arrow.apache.org/docs/python/generated/pyarrow.ChunkedArray.html#pyarrow.ChunkedArray.to_pandas
+    """
+    import pandas as pd  # ignore-banned-import
+
+    if Implementation.PANDAS._backend_version() < (1, 5):
+        # `pd.ArrowDType` added https://github.com/pandas-dev/pandas/pull/46774
+        return None
```
MarcoGorelli: 🤔 should this raise?

narwhals/dataframe.py
```diff
@@ -817,12 +818,28 @@ def to_native(self) -> DataFrameT:
         """
         return self._compliant_frame._native_frame
 
-    def to_pandas(self) -> pd.DataFrame:
+    def to_pandas(
+        self,
+        *,
+        use_pyarrow_extension_array: bool = False,
+        **kwds: Unpack[ToPandasArrowKwds],
```
MarcoGorelli: not sure about these tbh, the user doesn't know when `pyarrow.Table.to_pandas` is used and when it's not

would it work for this to be like the `**kwds` in `collect`, that just get passed down and are backend-specific?

ok with adding `use_pyarrow_extension_array` though

dangotbanned: Thanks for taking a look @MarcoGorelli

> not sure about these tbh, the user doesn't know when `pyarrow.Table.to_pandas` is used and when it's not

Yeah I agree, I'm not happy about them getting swallowed in the `PandasLike` cases either 🤔 

I'm gonna move this back to draft as I've got a couple ideas I wanna try out

dangotbanned: Okay I've got two options that seem like an improvement to me.

### 1 - Only allow `use_pyarrow_extension_array` for `PandasLike`
Keep roughly the same behavior, but raise if `kwds` is non-empty.

We can add another example to the docstring(s) to demonstrate how to pass `**kwds` in a safe way, like:

```py
if df.implementation.is_pandas_like():
    result = df.to_pandas(use_pyarrow_extension_array=True)
else:
    result = df.to_pandas(
        use_pyarrow_extension_array=True, self_destruct=True, zero_copy_only=True
    )
```

The only *marginal* downside is that it is less backend-agnostic.
But really, the caller is converting to a specfic backend - so that might not be a concern anyway

### 2 - Roundtrip through `pa.Table.to_pandas`
This is roughly what `polars` is doing, but we don't have the luxury of pushing the first conversion to `rust`

https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/dataframe/frame.py#L2600-L2608

```py
    def to_arrow(self, *, preserve_index: bool = False) -> pa.Table:
        if self._implementation is Implementation.CUDF:
            return self.native.to_arrow(preserve_index=preserve_index)

        import pyarrow as pa  # ignore-banned-import()

        return pa.Table.from_pandas(self.native, preserve_index=preserve_index)

    def _to_pandas(self) -> pd.DataFrame:
        if self._implementation is Implementation.PANDAS:
            return self.native
        elif self._implementation is Implementation.CUDF:
            return self.native.to_pandas()
        return self.native._to_pandas()

    def to_pandas(
        self,
        *,
        use_pyarrow_extension_array: bool = False,
        **kwds: Unpack[ToPandasArrowKwds],
    ) -> pd.DataFrame:
        if use_pyarrow_extension_array or kwds:
            if kwds:
                ns_arrow = self._version.namespace.from_backend("pyarrow").compliant
                return ns_arrow.from_native(self.to_arrow(preserve_index=True)).to_pandas(
                    use_pyarrow_extension_array=use_pyarrow_extension_array, **kwds
                )
            if self._implementation is Implementation.CUDF:
                return self.native.to_pandas(arrow_type=True)
            return self._to_pandas().convert_dtypes(dtype_backend="pyarrow")
        return self._to_pandas()
```

The main upsides here are:
- *we* eat the backend-branching, rather than offloading that to users
- `**kwds` will always end up in a (`pyarrow`) `to_pandas` call

**However**, if the motivation for using something like `zero_copy_only` was due to memory constraints - then the first `to_arrow` might make that promise impossible to keep.

dangotbanned: I'd say option **1** would be the safest set of trade-offs for us, but **2** is still a better option than silently ignoring args

narwhals/_arrow/utils.py
dangotbanned: Yeah, resolved in (https://github.com/narwhals-dev/narwhals/pull/2879/commits/db278a10d8f9c878dc8f51dd01d505e3265d864f)

narwhals/typing.py
FBruzzesi, line 1: I am probably missing something but: all the boolean flags are marked just with a single value. Does that mean that a user cannot use/change them? If that's the case, I would rather not expose them

narwhals/dataframe.py
FBruzzesi: I would also prefer option 1. If we were to write an example for how to use it, I think the best way would be similarly for how we do in our tests. Namely mapping implementation to kwargs:

```py
kwargs = {
    Implementation.PANDAS: {"use_pyarrow_extension_array": True},
    Implementation.POLARS: {"use_pyarrow_extension_array": True, "self_destruct": True, "zero_copy_only": True},
}

result = df.to_pandas(**kwargs.get(df.implementation, {}))
```

---

Matter of fact, we might want to document this pattern explicitly somewhere 😂

FBruzzesi (commented):
Thanks @dangotbanned - I left one question regarding the `ToPandasArrowKwds` typed dict, and one suggestion for how to document the changes for a user

narwhals/typing.py
dangotbanned: Ah I was somewhat anticipating confusion when I wrote (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2240316966) 😅 

## Explanation

What I'm trying to capture is that since a `TypedDict` **cannot** have defaults (but the [function they're passed to](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas) *does*) - is that by removing the default will mean that anything a user provides (and passes type checking) *should* have an effect.

## Example
The aim was to avoid documenting things which do nothing, 

Each of these 4 calls produce the same result

image

> [!NOTE]
> I'm feeling more strongly now that this isn't intuitive and the typing from (https://github.com/narwhals-dev/narwhals/commit/bc09ebffb2cf2b570dec3b04465b29fd1c0df3c8) would be easier to understand instead

narwhals/dataframe.py
dangotbanned: > I would also prefer option 1

Well I'm glad to hear that @FBruzzesi, since I got impatient and went in that direction anyway 😂 

I'll add an example for each now based on (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2256486582)

dangotbanned: I couldn't get the typing to play nicely with (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2256486582)

@FBruzzesi do you have a preference for *either* ignoring typing or *one of* these two instead?


<details><summary>Following from this section defining <code>df_nested</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/04fce8fed88725093abffc7f77a3a27a743ccc37/narwhals/dataframe.py#L865-L892

</p>
</details> 

## 1
```py
import narwhals.typing

kwargs: dict[nw.Implementation, narwhals.typing.ToPandasArrowKwds] = {
    nw.Implementation.POLARS: {"zero_copy_only": True},
    nw.Implementation.PYARROW: {"zero_copy_only": True, "date_as_object": False},
}
df_nested.to_pandas(
    use_pyarrow_extension_array=True, **kwargs.get(df_nested.implementation, {})
)
```

## 2
```py
from narwhals.typing import ToPandasArrowKwds

kwargs = {
    nw.Implementation.POLARS: ToPandasArrowKwds(zero_copy_only=True),
    nw.Implementation.PYARROW: ToPandasArrowKwds(
        zero_copy_only=True, date_as_object=False
    ),
}
df_nested.to_pandas(
    use_pyarrow_extension_array=True, **kwargs.get(df_nested.implementation, {})
)
```

FBruzzesi: _Personal_ preference is having typing via option 1 😊

narwhals/typing.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2879/commits/1fa43a437e5dc7e34c4a12a6fe21046331b8280e)

I've changed the `bool` flags, but the types which are `T | None = None` are staying as `T` 😄

narwhals/dataframe.py
dangotbanned: > Personal preference is having typing via option 1 

Nice one, I'll go with that

Both versions would support this, but I'm always surprised when this works for the string keys 😄 

image

narwhals/typing.py
```diff
@@ -411,6 +413,68 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+    For example, if multiple columns share an underlying allocation, memory can't be freed until all columns are converted.
+    """
+    split_blocks: bool
+    """Generate one internal "block" for each column when creating a `pandas.DataFrame` from a `RecordBatch` or `Table`.
+
+    While this can temporarily reduce memory note that various pandas operations can trigger "consolidation" which may balloon memory use.
+    """
+    use_threads: bool
+    safe: bool
+    """For certain data types, a cast is needed in order to store the data in a `pandas.DataFrame` or `Series` (e.g. timestamps are always stored as nanoseconds in pandas).
```
MarcoGorelli: i haven't read through the whole thread yet, but i'm really not wild about have all this here in the narwhals api, especially experimental options

MarcoGorelli (changes requested):

narwhals/typing.py
dangotbanned: Thanks @MarcoGorelli 

Re: experimental, I felt the same way but found out it has been marked as experimental since `pyarrow==1` strangely 

- https://arrow.apache.org/docs/1.0/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas

`self_destruct` was also explicitly mentioned in #2123 and is used in `polars` as well

- https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/dataframe/frame.py#L2604

dangotbanned: I'm very in favor of trimming down the arguments though

There seems to be a lot of redundancy, that could just be resolved with `types_mapper`

## Update
[docs(typing): Trim keywords and docs from `ToPandasArrowKwds`](https://github.com/narwhals-dev/narwhals/pull/2879/commits/68aea9e225784201c7e75b93bdee1a4126d328e0)

narwhals/dataframe.py
```diff
@@ -826,12 +827,28 @@ def to_native(self) -> DataFrameT:
...
         """Convert this DataFrame to a pandas DataFrame.
 
+        Arguments:
+            use_pyarrow_extension_array: Use PyArrow-backed extension arrays instead of
+                NumPy arrays for the columns of the pandas DataFrame.
+                This allows zero copy operations and preservation of null values.
+                Subsequent operations on the resulting pandas DataFrame may trigger conversion to
+                NumPy if those operations are not supported by PyArrow compute functions.
+            **kwds: Additional keyword arguments to be passed to [`pyarrow.Table.to_pandas`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_pandas)
+
```
MarcoGorelli, line 901: sorry i still don't get why this is here at all

a library doesn't have to use `pyarrow.to_pandas` to do the conversion, right?

do we not want this to just be `**kwargs` and have it be backend-specific, like the IO functions are?

dangotbanned: > do we not want this to just be `**kwargs` and have it be backend-specific, like the IO functions are?

The main difference I see is that these `to_pandas` methods are significantly more aligned than the IO functions

I've documented:

<details><summary>the limited support in pandas-like</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/68aea9e225784201c7e75b93bdee1a4126d328e0/narwhals/_pandas_like/utils.py#L691-L711

</p>
</details> 


and 

<details><summary>the conflicting arg combinations for polars</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/68aea9e225784201c7e75b93bdee1a4126d328e0/narwhals/_polars/utils.py#L258-L279

</p>
</details> 

So *yes* we can just do `**kwargs`, but I think that is **too wide** of a description, considering the only extra things are 2x for `cudf`.

I find the polars conflicts easier to understand when they are documented in our typing.
I'm less attached to the the docstrings *within* `ToPandasArrowKwds` - so if this is a case of (#1942) - I'm more than happy to remove those 🙂

dangotbanned:
Hey @MarcoGorelli, would you be able to give me some boxes to tick in order to get this one over the line please? 🙏

It seems like we've been in agreement from the start on `use_pyarrow_extension_array` 🙂 

But it doesn't seem like the changes I've made in response to (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2247516506) and (https://github.com/narwhals-dev/narwhals/pull/2879#discussion_r2266937658) have gotten us closer

Really appreciate you bearing with me on this!

narwhals/dataframe.py
dangotbanned: Removed in ([revert: `ToPandasArrowKwds`](https://github.com/narwhals-dev/narwhals/pull/2879/commits/b997fe7f4f27ae92a1372c595056ae1ec42d745a))

MarcoGorelli:
sorry, didn't get round to this, perhaps we'll revisit later

dangotbanned:
> sorry, didn't get round to this, perhaps we'll revisit later

@MarcoGorelli I'm happy to reopen and make changes, the idea has been trimmed down since your last review (https://github.com/narwhals-dev/narwhals/pull/2879#pullrequestreview-3106147405)

I really just wanted this merged to start making more progress on [nested data](https://github.com/narwhals-dev/narwhals/issues?q=sort%3Aupdated-desc+state%3Aopen+label%3A%22nested+data%22).
I'd held off on proposing more because `pandas` will be complicated to do without this

PR #2880, opened 2025-07-24
Refactor descriptives

ym-pett:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

I didn't remove the `skew` function from the `DuckDBExpr` class as it has additional logic in it missing from the sparklike class, probably due to implementation differences in the two, I can see it gets the sample skew out. I'm assuming that the sparklike and sql implementations also return sample skew, but is this a dangerous assumption to make? Should I be checking for the population vs sample distinction for all the descriptives in future work?

### DIFF
--- a/narwhals/_sql/expr.py
+++ b/narwhals/_sql/expr.py
@@ -374,6 +374,9 @@ class SQLExpr(
     def min(self) -> Self:
         return self._with_callable(lambda expr: self._function("min", expr))
 
+    def kurtosis(self) -> Self:
+        return self._with_callable(lambda expr: self._function("kurtosis", expr))
+
     # Elementwise
     def abs(self) -> Self:
         return self._with_elementwise(lambda expr: self._function("abs", expr))
--- a/narwhals/_sql/expr.py
+++ b/narwhals/_sql/expr.py
@@ -377,6 +377,9 @@ class SQLExpr(
     def kurtosis(self) -> Self:
         return self._with_callable(lambda expr: self._function("kurtosis", expr))
 
+    def skew(self) -> Self:
+        return self._with_callable(lambda expr: self._function("skewness", expr))
+
     # Elementwise
     def abs(self) -> Self:
         return self._with_elementwise(lambda expr: self._function("abs", expr))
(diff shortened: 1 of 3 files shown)


ym-pett:
ah I can see a couple of tests failing on the PR, is that something I can remediate?

MarcoGorelli:
thanks for looking into this!

notice how in duckdb, the function that gets called is `F('kurtosis_pop', expr)`. If you look at `DuckDBExpr._function`, you'll see that it just forwards the call to `F`

https://github.com/narwhals-dev/narwhals/blob/80c5fc2a001bc9cdc990ad4f188708eea36f810e/narwhals/_duckdb/expr.py#L75-L78

If you call `self._function("kurtosis"` from `SQLExpr`, it'll end up calling `F('kurtosis', expr)` in DuckDB, which isn't quite correct. That's why the tests are failing

Given that the method name varies between spark and duckdb, and given that it's not supported in ibis, this one may not be a good candidate for sharing code

MarcoGorelli:
`count` is probably a good candidate if you fancy taking a look at that instead

ym-pett:
"That's why the tests are failing" - ah I've seen how to look into the failing tests on github now and can see they're indeed to do with the duckdb test for kurtosis, nice one!

narwhals/_sql/expr.py
```diff
@@ -374,6 +374,12 @@ def median(self) -> Self:
     def min(self) -> Self:
         return self._with_callable(lambda expr: self._function("min", expr))
 
+    def skew(self) -> Self:
+        return self._with_callable(lambda expr: self._function("skewness", expr))
```
MarcoGorelli: shall we leave this one as-is if it doesn't let us do any code sharing?

ym-pett: ah yes you're right! will do.

MarcoGorelli (approved):
thanks @ym-pett !

MarcoGorelli:
ah there's just a merge conflict that needs resolving

could you do resolve it please? once that's done we can merge

ym-pett:
yes was just looking into that conflict, trying to see if i can resolve it locally (i thought I had)

ym-pett:
... that worked, no more conflicts, checks passed 🥳

PR #2882, opened 2025-07-24
refactor: add compliant level parent class for `hist`

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

Tentative to align arrow to the same API, and abstract whatever possible at the moment to the compliant level

### DIFF
--- a/narwhals/_compliant/__init__.py
+++ b/narwhals/_compliant/__init__.py
@@ -34,11 +34,11 @@ from narwhals._compliant.series import (
     EagerSeries,
     EagerSeriesCatNamespace,
     EagerSeriesDateTimeNamespace,
+    EagerSeriesHist,
     EagerSeriesListNamespace,
     EagerSeriesNamespace,
     EagerSeriesStringNamespace,
     EagerSeriesStructNamespace,
-    _EagerSeriesHist,
 )
 from narwhals._compliant.typing import (
     CompliantExprT,
@@ -83,6 +83,7 @@ __all__ = [
     "EagerSeries",
     "EagerSeriesCatNamespace",
     "EagerSeriesDateTimeNamespace",
+    "EagerSeriesHist",
     "EagerSeriesListNamespace",
     "EagerSeriesNamespace",
     "EagerSeriesStringNamespace",
@@ -99,5 +100,4 @@ __all__ = [
     "NativeFrameT_co",
     "NativeSeriesT_co",
     "WindowInputs",
-    "_EagerSeriesHist",
 ]
--- a/narwhals/_compliant/series.py
+++ b/narwhals/_compliant/series.py
@@ -69,11 +69,11 @@ __all__ = [
     "EagerSeries",
     "EagerSeriesCatNamespace",
     "EagerSeriesDateTimeNamespace",
+    "EagerSeriesHist",
     "EagerSeriesListNamespace",
     "EagerSeriesNamespace",
     "EagerSeriesStringNamespace",
     "EagerSeriesStructNamespace",
-    "_EagerSeriesHist",
 ]
 
 
@@ -447,7 +447,7 @@ class EagerSeriesStructNamespace(  # type: ignore[misc]
 ): ...
 
 
-class _EagerSeriesHist(Protocol[NativeSeriesT, _CountsT_co]):
+class EagerSeriesHist(Protocol[NativeSeriesT, _CountsT_co]):
     _series: EagerSeries[NativeSeriesT]
     _breakpoint: bool
     _data: HistData[NativeSeriesT, _CountsT_co]
@@ -466,6 +466,14 @@ class _EagerSeriesHist(Protocol[NativeSeriesT, _CountsT_co]):
         return obj
 
     def to_frame(self) -> EagerDataFrameAny: ...
+    def _linear_space(  # NOTE: Roughly `pl.linear_space`
+        self,
+        start: float,
+        end: float,
+        num_samples: int,
+        *,
+        closed: Literal["both", "none"] = "both",
+    ) -> _1DArray: ...
 
     # NOTE: *Could* be handled at narwhals-level
     def is_empty_series(self) -> bool: ...
@@ -497,19 +505,6 @@ class _EagerSeriesHist(Protocol[NativeSeriesT, _CountsT_co]):
         else:
             self._data = self._calculate_hist(self._calculate_bins(bin_count))
         return self
-
-    # NOTE: Roughly `pl.linear_space`
-    def _linear_space(
-        self,
-        start: float,
-        end: float,
-        num_samples: int,
-        *,
-        closed: Literal["both", "none"] = "both",
-    ) -> _1DArray:
-        from numpy import linspace  # ignore-banned-import
-
-        return linspace(start=start, stop=end, num=num_samples, endpoint=closed == "both")  # type: ignore[no-any-return]
 
     def _calculate_breakpoint(self, arg: int | list[float], /) -> list[float] | _1DArray:
         bins = self._linear_space(0, 1, arg + 1) if isinstance(arg, int) else arg
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, cast, overload
+from typing import TYPE_CHECKING, Any, Literal, cast, overload
 
 import pyarrow as pa
 import pyarrow.compute as pc
@@ -22,7 +22,7 @@ from narwhals._arrow.utils import (
     pad_series,
     zeros,
 )
-from narwhals._compliant import EagerSeries, _EagerSeriesHist
+from narwhals._compliant import EagerSeries, EagerSeriesHist
 from narwhals._expression_parsing import ExprKind
 from narwhals._typing_compat import assert_never
 from narwhals._utils import (
@@ -1093,7 +1093,7 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
 
 
 class _ArrowHist(
-    _EagerSeriesHist["ChunkedArrayAny", "list[ScalarAny] | pa.Int64Array | list[float]"]
+    EagerSeriesHist["ChunkedArrayAny", "list[ScalarAny] | pa.Int64Array | list[float]"]
 ):
     _series: ArrowSeries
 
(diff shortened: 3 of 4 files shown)


FBruzzesi:
@dangotbanned the main struggle is to have everything nice and compliant (pun semi intended) with types. I think it's not too far, but definitely some love is needed. 

I am opening the PR just to show you how it would look like, and how by introducing features you mentioned https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3096758332 it could be simplified even more

narwhals/_compliant/series.py
```diff
@@ -56,6 +56,8 @@
         _SliceIndex,
     )
 
+    _HistData: TypeAlias = dict[str, list[float] | _1DArray | NativeSeriesT]
```
dangotbanned: Okay I've done some investigating and I think this is a source of some issues

I'm gonna redo this as a generic `TypedDict`

Very impressed you thought to make this generic though 😍

narwhals/_compliant/series.py
```diff
@@ -436,3 +439,76 @@ class EagerSeriesStructNamespace(  # type: ignore[misc]
     StructNamespace[EagerSeriesT_co],
     Protocol[EagerSeriesT_co, NativeSeriesT_co],
 ): ...
+
+
+class _EagerSeriesHist(Generic[NativeSeriesT]):
```
dangotbanned: This'll fix the `mypy` warnings like:

```py
narwhals/_compliant/series.py:469: error: Missing return statement  [empty-body]
        def is_empty_series(self) -> bool: ...
```

With `Generic`, you'd need to add [`@abc.abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) to them, but a nice thing about [`Protocol`](https://mypy.readthedocs.io/en/stable/protocols.html#defining-subprotocols-and-subclassing-protocols) is that `...`

> In particular, omitting a value for an attribute or a method body will make it implicitly abstract

```suggestion
class _EagerSeriesHist(Protocol[NativeSeriesT]):
```

dangotbanned: > [!IMPORTANT]
> Hold off on applying this for now  

I'm trying to solve all the typing stuff atm 😂

dangotbanned: @FBruzzesi all the warnings are now resolved (https://github.com/narwhals-dev/narwhals/pull/2882#commits-pushed-974369b) 🎉 

FWIW I don't actually like what I've wrote in those commits, but its a starting point for you to build on I guess 😅

narwhals/_compliant/series.py
```diff
@@ -436,3 +444,78 @@ class EagerSeriesStructNamespace(  # type: ignore[misc]
...
+        return obj
+
+    def to_frame(self) -> Any: ...  # TODO(FBruzzesi): Add annotation
+
+    # NOTE: *Could* be handled at narwhals-level
+    def is_empty_series(self) -> bool: ...
+
+    # NOTE: **Should** be handled at narwhals-level
+    def data_empty(self) -> HistData[NativeSeriesT, _CountsT_co]:
+        return {"breakpoint": [], "count": []} if self._breakpoint else {"count": []}
```
dangotbanned, line 483: Apologies for not being clearer or if I've misunderstood the PR!

By *narwhals-level*, I meant that I was hoping we could move these bits into `nw.Series.hist`

IIUC, these cases are dependent only on the parameters.
So if we handled it in `nw.Series.hist`, then even `PolarsSeries.hist` could benefit from having this done for it 🙂 

https://github.com/narwhals-dev/narwhals/blob/fe29ec3d43efe2d464bcba5c21afc67c3d93b4ce/narwhals/_compliant/series.py#L483-L485

https://github.com/narwhals-dev/narwhals/blob/fe29ec3d43efe2d464bcba5c21afc67c3d93b4ce/narwhals/_compliant/series.py#L492-L494


<details><summary><b>Show rough and super repetitive version</b></summary>
<p>

There's loads of room for improvement on this, but it passes the tests 😄 

```diff
diff --git a/narwhals/series.py b/narwhals/series.py
index 033d5f65e..789a93d40 100644
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import math
 from collections.abc import Iterator, Mapping, Sequence
-from typing import TYPE_CHECKING, Any, Callable, Generic, Literal, overload
+from typing import TYPE_CHECKING, Any, Callable, Generic, Literal, cast, overload
 
 from narwhals._utils import (
     Implementation,
@@ -33,6 +33,7 @@ if TYPE_CHECKING:
     from typing_extensions import Self
 
     from narwhals._compliant import CompliantSeries
+    from narwhals._compliant.typing import CompliantDataFrameAny, EagerNamespaceAny
     from narwhals.dataframe import DataFrame, MultiIndexSelector
     from narwhals.dtypes import DType
     from narwhals.typing import (
@@ -2587,6 +2588,7 @@ class Series(Generic[IntoSeriesT]):
             |3        8.00      2|
             └────────────────────┘
         """
+        series = self._compliant_series
         if bins is not None:
             if any(bins[i - 1] >= bins[i] for i in range(1, len(bins))):
                 msg = "bins must increase monotonically"
@@ -2594,16 +2596,37 @@ class Series(Generic[IntoSeriesT]):
             if bin_count is not None:
                 msg = f"can only provide one of `bin_count` or `bins`, got: {bin_count=}, {bins=}"
                 raise ComputeError(msg)
-            result = self._compliant_series.hist_from_bins(
-                bins=bins, include_breakpoint=include_breakpoint
-            )
+
+            if len(bins) <= 1:
+                ns = cast("EagerNamespaceAny", series.__narwhals_namespace__())
+                data: dict[str, Any] = (
+                    {"breakpoint": [], "count": []}
+                    if include_breakpoint
+                    else {"count": []}
+                )
+                result: CompliantDataFrameAny = ns._dataframe.from_dict(
+                    data, context=ns, schema=None
+                )
+            else:
+                result = series.hist_from_bins(
+                    bins=bins, include_breakpoint=include_breakpoint
+                )
         else:
             # polars (v1.20) sets bin=10 if neither are provided.
             default = 10
             bin_count = default if bin_count is None else bin_count
-            result = self._compliant_series.hist_from_bin_count(
-                bin_count=bin_count, include_breakpoint=include_breakpoint
-            )
+            if bin_count == 0:
+                ns = cast("EagerNamespaceAny", series.__narwhals_namespace__())
+                data = (
+                    {"breakpoint": [], "count": []}
+                    if include_breakpoint
+                    else {"count": []}
+                )
+                result = ns._dataframe.from_dict(data, context=ns, schema=None)
+            else:
+                result = series.hist_from_bin_count(
+                    bin_count=bin_count, include_breakpoint=include_breakpoint
+                )
 
         return self._dataframe(result, level=self._level)
 


```

</p>
</details>

FBruzzesi: Thanks @dangotbanned - I know what you meant, but let me bring up a point which I think it's worth keeping in the back our our mind: output data types.

When we create dataframes from a dictionary without values, we should be careful on the datatypes. Since I didn't want to bother within that regard as well in this PR, I kept this issue for another time 😂 I hope it makes sense. We are already ignoring that for pandas and pyarrow, but at least we let polars do what polars does

dangotbanned: Ah I half suspected this was intentional, sure sounds reasonable to keep this PR scoped to compliant-level only 🙂

FBruzzesi:
@dangotbanned I am taking a look after dinner from my laptop, but any idea why marimo hist fails? 😢

dangotbanned:
> @dangotbanned I am taking a look after dinner from my laptop, but any idea why marimo hist fails? 😢

No idea sadly, I assumed it wasn't related because it was in `test_ibis_table` - but we'll see if the re-run changes 🤔 
- https://github.com/narwhals-dev/narwhals/actions/runs/16523439183/job/46751452067?pr=2882


### Edit
> we'll see if the re-run changes

All good 🙌

FBruzzesi:
> > @dangotbanned I am taking a look after dinner from my laptop, but any idea why marimo hist fails? 😢
> 
> No idea sadly, I assumed it wasn't related because it was in `test_ibis_table` - but we'll see if the re-run changes 🤔
> 
> * https://github.com/narwhals-dev/narwhals/actions/runs/16523439183/job/46751452067?pr=2882
> 
> ### Edit
> > we'll see if the re-run changes
> 
> All good 🙌

Mmmh how weird that something related to bin count broke :|

Anyway, I added a commit, and I think we can merge?

narwhals/_compliant/series.py
```diff
@@ -436,3 +444,77 @@ class EagerSeriesStructNamespace(  # type: ignore[misc]
...
+        return self._series.native
+
+    @classmethod
+    def from_series(
+        cls, series: CompliantSeries[NativeSeriesT], *, include_breakpoint: bool
+    ) -> Self:
+        obj = cls.__new__(cls)
+        obj._series = series
+        obj._breakpoint = include_breakpoint
+        return obj
```
dangotbanned, line 466: (https://github.com/narwhals-dev/narwhals/pull/2882/commits/764f3ff0aea83c0f4590a02a1a7d16c4bbc1ab35)

oooh nice one!

narwhals/_compliant/series.py
```diff
@@ -436,3 +444,77 @@ class EagerSeriesStructNamespace(  # type: ignore[misc]
...
+        self,
+        start: float,
+        end: float,
+        num_samples: int,
+        *,
+        closed: Literal["both", "none"] = "both",
+    ) -> _1DArray:
+        from numpy import linspace  # ignore-banned-import
+
+        return linspace(start=start, stop=end, num=num_samples, endpoint=closed == "both")  # type: ignore[no-any-return]
```
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2839#discussion_r2215630696

> Even though it is an inline import that is technically not on `CompliantSeries` itself - I'd prefer if we didn't have a `numpy` import here

🙄🙄🙄

narwhals/_compliant/__init__.py
```diff
@@ -38,6 +38,7 @@
     EagerSeriesNamespace,
     EagerSeriesStringNamespace,
     EagerSeriesStructNamespace,
+    _EagerSeriesHist,
```
dangotbanned: Fully my bad, but we should have all of these new classes/protocols *without* the `_` prefix

Especially if they're being exported like this

dangotbanned (commented):
Thanks @FBruzzesi!

We're almost there now 😅

narwhals/_compliant/series.py
FBruzzesi: nice catch! I solved it, and ruined it again

FBruzzesi:
Thanks for your patience @dangotbanned , merging into #2839 then ;)

Issue #2786, opened 2025-07-04
Establish safe patterns using `Implementation.UNKNOWN`


dangotbanned:
## Background

Currently we raise an `AssertionError` in many places `Implementation.UNKNOWN` could appear:

<details><summary><code>Implementation.(to_native_namespace|_backend_version)</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/_utils.py#L363-L371

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/_utils.py#L624-L633

</p>
</details> 

However, this seems to be in conflict with how `functions.py` depends on `Implementation` methods:

<details><summary><code>narwhals.functions.py</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L229-L247

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L322-L327

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L428-L433

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L513-L519

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L678-L679

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L754-L755

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L847-L848

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/functions.py#L952-L953

</p>
</details> 

There's also `get_native_namespace`, where we transform that into a `TypeError`:

<details><summary><code>narwhals.translate.get_native_namespace</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/4d73399c0f217286b7aed2270ea27aa4441bb166/narwhals/translate.py#L607-L622

[...]

PR #2885, opened 2025-07-25
feat: Add `DataFrame.from_dict`

dangotbanned:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Part of #2116

## Checklist

- [x] Tests added
- [x] Documented the changes

- [x] Resolve `[var-annotated]` by not using `Self` (https://github.com/narwhals-dev/narwhals/pull/2885/commits/695f9ac4850e56639aa96fe5ae3a67234621e31f)

## Possible Follow-ups
- Figure out how (#2786) fits in
- Add `@overload`s on `backend`
  - Like in `Namespace`
- Defer to `DataFrame.from_dict` in `nw.from_dict`

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -6,6 +6,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    ClassVar,
     Generic,
     Literal,
     NoReturn,
@@ -28,6 +29,7 @@ from narwhals._utils import (
     generate_repr,
     is_compliant_dataframe,
     is_compliant_lazyframe,
+    is_eager_allowed,
     is_index_selector,
     is_list_of,
     is_sequence_like,
@@ -38,12 +40,13 @@ from narwhals._utils import (
 )
 from narwhals.dependencies import get_polars, is_numpy_array
 from narwhals.exceptions import InvalidIntoExprError, InvalidOperationError
+from narwhals.functions import _from_dict_no_backend
 from narwhals.schema import Schema
 from narwhals.series import Series
 from narwhals.translate import to_native
 
 if TYPE_CHECKING:
-    from collections.abc import Iterable, Iterator, Sequence
+    from collections.abc import Iterable, Iterator, Mapping, Sequence
     from io import BytesIO
     from pathlib import Path
     from types import ModuleType
@@ -55,6 +58,7 @@ if TYPE_CHECKING:
 
     from narwhals._compliant import CompliantDataFrame, CompliantLazyFrame
     from narwhals._compliant.typing import CompliantExprAny, EagerNamespaceAny
+    from narwhals.dtypes import DType
     from narwhals.group_by import GroupBy, LazyGroupBy
     from narwhals.typing import (
         AsofJoinStrategy,
@@ -409,6 +413,8 @@ class DataFrame(BaseFrame[DataFrameT]):
             ```
     """
 
+    _version: ClassVar[Version] = Version.MAIN
+
     def _extract_compliant(self, arg: Any) -> Any:
         from narwhals.expr import Expr
         from narwhals.series import Series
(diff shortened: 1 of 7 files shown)


narwhals/dataframe.py
```diff
@@ -409,6 +413,8 @@ class DataFrame(BaseFrame[DataFrameT]):
             ```
     """
 
+    _version: ClassVar[Version] = Version.MAIN
```
dangotbanned, line 416: @MarcoGorelli 

Doing this is gonna be quite handy for the rest of (#2116)

Since we're already inside the class, we can just start from the *known* version there

https://github.com/narwhals-dev/narwhals/blob/695f9ac4850e56639aa96fe5ae3a67234621e31f/narwhals/dataframe.py#L512

instead of using `_stablify` afterwards 😄 

https://github.com/narwhals-dev/narwhals/blob/f8f7f65e5bb82580f373cf8c2200e89e8febbdc1/narwhals/functions.py#L316

MarcoGorelli: nice!

narwhals/dataframe.py
```diff
@@ -452,6 +458,68 @@ def __init__(self, df: Any, *, level: Literal["full", "lazy", "interchange"]) ->
...
+            ns = cls._version.namespace.from_backend(implementation).compliant
+            compliant = ns._dataframe.from_dict(data, schema=schema, context=ns)
+            return cls(compliant, level="full")
+        # NOTE: (#2786) needs resolving for extensions
+        msg = (
+            f"{implementation} support in Narwhals is lazy-only, but `DataFrame.from_dict` is an eager-only function.\n\n"
+            "Hint: you may want to use an eager backend and then call `.lazy`, e.g.:\n\n"
+            f"    nw.DataFrame.from_dict({{'a': [1, 2]}}, backend='pyarrow').lazy('{implementation}')"
+        )
+        raise ValueError(msg)
```
MarcoGorelli, line 521: i'm confused as to why is this necessary - if we have a `DataFrame` class, aren't we already eager?

MarcoGorelli (commented):
thanks @dangotbanned !

just left a comment

narwhals/dataframe.py
dangotbanned: I was just adapting the existing error in `nw.from_dict`

https://github.com/narwhals-dev/narwhals/blob/ee06d33b500337da40ed72eb4b8c9f8c181f6c77/narwhals/functions.py#L328-L332

Did you want this version to do something different if we wrote?

```py
import narwhals as nw

nw.DataFrame.from_dict({'a': [1, 2]}, backend="pyspark")
```

MarcoGorelli: ah I see sorry, this is indeed needed and correct, thanks!

my only other comment would be - do we want to repeat the docstring, or do like numpy and say "see [...] for full docstring"? probably fine to repeat?

https://numpy.org/doc/stable/reference/generated/numpy.ndarray.std.html

dangotbanned: >  "see [...] for full docstring"? probably fine to repeat?

Definitely repeat!

I'm guessing even to get that doc you'd need to not have type stubs or have a jupyter kernel running?

image

I'm always gonna feel strongly about IDE-experience-**first**-docs, since that's how I consume them 🙂 

*Sadly* useful docs + static typing often mean we need to repeat ourselves.
But putting in that extra effort is something I really appreciate from `polars` vs `pandas`

dangotbanned:
> this needs adding to `dataframe.md` in the docs, right?

oop yeah will do that now!

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2885#event-18831991922) 
Just need to fix `check_api_reference.py` to handle `classmethod`

Preview of the doc 

image

PR #2886, opened 2025-07-25
feat: add multivalue replacement in `.str.replace[_all]`

### LINKED ISSUE #2745: [Enh]: Allow .str.replace to receive Expr as value

The project is not open source, but I have an ML problem and a table of potential configurations that I want to override with the final one

In polars, we can use expressions when replacing string values:

```python
import polars as pl
df = pl.DataFrame({'a': ['hello', 'world'], 'b': ['Your word is: <x>', 'Your word is: <x>']})
df.with_columns(pl.col('b').str.replace('<x>', pl.col('a')).alias('c'))
```

But doing the same is not supported in narwhals, as it forces the value to be a literal:
```python
nw.from_native(df).with_columns(nw.col('b').str.replace('<x>', nw.col('a')).alias('c'))
```
This line returns the following error message:
```
TypeError: cannot create expression literal for value of type Expr.

```

_No response_

I am currently using a loop in python to overcome this issue, but the polars implementation was cleaner

_No response_

camriddell:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2745 

## Checklist

- [x] Tests added
- [ ] Documented the changes

@MarcoGorelli any thoughts on the docs for this feature? Is it worth adding examples or do you think just the update to the typing here suffices?


camriddell:
Also want a review for the error propagation and the tests. There are a decent number of backends that do not support this feature so I raised an error before calling into it. In the tests I just marked as xfail, but perhaps it makes more sense to check that the underlying TypeError is being raised?

dangotbanned:
> any thoughts on the docs for this feature?

I'd say we should be pretty explicit on which backends will support it

Is it just `polars` and `duckdb`?

Apologies if I'm misreading the tests 🫣

camriddell:
> > any thoughts on the docs for this feature?
> 
> I'd say we should be pretty explicit on which backends will support it
> 
> Is it just `polars` and `duckdb`?
> 
> Apologies if I'm misreading the tests 🫣

Nope I think you've nailed it. Polars and `duckdb` support it natively. I added `pandas` via manual iteration (since pandas drops down to slow iteration for a number of ops anyways).

camriddell:
Added some more explicit error checks, and delayed triggering errors unless the backend fails (this should force our tests to begin failing if the backend does support this feature in the future). After aggregating this behavior, here is a table showing the support for each feature by backend.

| Backend        | `str.replace[scalar]` | `str.replace[multivalue]` | `str.replace_all[scalar]` | `str.replace_all[multivalue]` |
|----------------|----------------------|---------------------------|---------------------------|-------------------------------|
| pandas         | ✅                    | ✅  (pure Python)  | ✅                         | ✅ (pure Python)          |
| polars[eager]  | ✅                    | ✅                         | ✅                         | ✅                             |
| polars[lazy]   | ✅                    | ✅                         | ✅                         | ✅                             |
| duckdb         | ❌                    | ❌                         | ✅                         | ✅                             |
| spark_like     | ❌                    | ❌                         | ✅                         | ✅                             |
| ibis           | ❌                    | ❌                         | ✅                         | ✅                             |
| pyarrow        | ✅                    | ❌                         | ✅                         | ❌                             |
| dask           | ✅                    | ❌                         | ✅                         | ❌                             |

edit: @dangotbanned added Ibis support for `replace_all[multivalue]`

narwhals/_pandas_like/series_str.py
```diff
@@ -16,11 +21,32 @@ def len_chars(self) -> PandasLikeSeries:
...
+                ]
+            else:
+                regex = re.compile(pattern)
+                values = [
+                    regex.sub(repl, string, count=max(n, 0))
+                    for string, repl in zip(self.native, value)
+                ]
+            constructor = self.compliant.__native_namespace__().Series
+            series = constructor(values, dtype=self.native.dtype, name=self.native.name)
+        return self.with_native(series)
```
dangotbanned: I feel like this is going to be a footgun for performance, but could be convinced with a benchmark or two

How do the `pandas`, `polars` + `duckdb` versions scale for these lengths?

1. `1_000`
2. `10_000`
3. `100_000`
4. `1_000_000`

camriddell: This will definitely be a slower implementation since it's pure Python, however the decision point would be between this and no support at all since the built-in `.str.replace` doesn't support the multivalue feature. So I was leaning towards "let's support this at the pandas level even if it is slow (akin to the pandas object dtype, which strings still default to)"

If you're still interested I'll code up a benchmark to show you the requested values.

dangotbanned: > If you're still interested I'll code up a benchmark to show you the requested values.

Indeed I am! 🙂

If we're happy with the trade-off, then I think we can do the same for `pyarrow` as well?

Two other options:
1. Emit a warning like we do in `PandasLikeGroupBy` when using `apply`
2. Try to find some way to do this using *more* of the `pandas` API.

My main concern is the `pandas` docs are full of warnings against iterating over `Series`/`DataFrame` - and currently this pays the cost twice by using `zip` 🤔

MarcoGorelli: thanks for looking into this!

> between this and no support at all

I'm very strongly on the "no support at all" side

My general philosophy is that we should be translating to a library's own API - if they then follow slow paths internally, then whatever, that's up to them

camriddell: I'm in line with the opinions here, and I'm more than happy to declare the *multivalue* replacement in pandas as **not supported** and push up another commit removing that conditional branch to land this.

Just wanted to share my reasoning for incl. the pandas code above
1. pandas defaults to storing strings as `"object"` dtype backed, which under the hood is operating at the Python level.
2. we're adding a feature that is supported by 2 backends and not supported by 6+ other backends. Is this a problem, fine, as expected? Just thinking that as an end-user it would stink to see that a feature has such limited support.

Curious to everyone's thoughts on the above especially around point 2.?

MarcoGorelli: 1. sure, and that's pandas decision. if/when they switch to something better, then so long as we stick to the pandas api, then we'll also get those improvements. if we implement anything ourselves in python, then we won't
2. i'm quite interested in enabling narwhals to be a nice front-end to duckdb, so tbh i'm ok with it so long as these conditions are met (which, as far as i can tell, they are):
  - it's supported by at least polars lazy and duckdb
  - it's very simple to implement

dangotbanned: @camriddell
> Curious to everyone's thoughts on the above especially around point 2.?

@MarcoGorelli 
> i'm quite interested in enabling narwhals to be a nice front-end to duckdb,
> it's supported by at least polars lazy and duckdb

My preference would be to not support it, since it can only serve a narrow section of users.
I'm fine with most of this -  but the narrower we go - the more likely we are to push compatibility paths **back** to users:

👍 eager-only
👍 lazy-only
👍 eager + some lazy
👍 lazy excluding dask
👍 most of lazy
👎 only duckdb + polars

The second line in the docs is https://narwhals-dev.github.io/narwhals/

> Full API support: cuDF, Modin, pandas, Polars, PyArrow.

There should probably be a `*` in there, but generally I'd like to push for features that can meet that claim 🙂

MarcoGorelli: One way that I'm trying to think of Narwhals is "what is a good dataframe API, which doesn't make too many assumptions about its backends, meant to look like?"

To me, such a good API generally supports expressified arguments

If in this case it's just a matter of passing things down as expressified arguments, then I think it should be fine

narwhals/_spark_like/expr_str.py
```diff
@@ -20,18 +21,24 @@ class SparkLikeExprStringNamespace(
...
+        if isinstance(value, str):
 
-        return self.compliant._with_elementwise(func)
+            def func(expr: Column) -> Column:
+                return replace_all(expr, pattern_, F.lit(value))
+
+            return self.compliant._with_elementwise(func)
+        return self.compliant._with_elementwise(
+            lambda expr, value: replace_all(expr, pattern_, value), value=value
+        )
```
dangotbanned: @camriddell I had a read through the `pyspark` docs and was curious how you reached the conclusions in (https://github.com/narwhals-dev/narwhals/pull/2886#issuecomment-3129882666)?

This one was failing because [`pyspark.sql.functions.lit`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.lit.html) doesn't accept `narwhals._spark_like.expr.SparkLikeExpr`

All I've done here is rewritten it in the style of what you had for `duckdb` and it seems to work 🙂

narwhals/_polars/series.py
```diff
@@ -705,6 +705,40 @@ def zfill(self, width: int) -> PolarsSeries:
...
+        if isinstance(value, Series):
+            value = value.to_native()
+
+        return self._compliant_series._with_native(
+            self._compliant_series.native.str.replace_all(
+                pattern,
+                value,  # type: ignore[arg-type]
+                literal=literal,
+            )
+        )
```
dangotbanned: So there's a few things going on here:

1. The annotation is `str | PolarsSeries` (Compliant), but the runtime check is on `Series` (Narwhals) (https://github.com/camriddell/narwhals/blob/e422e99e3dac9e73df8fcf980b4f42ece76c4058/narwhals/_polars/series.py#L17)
    i. I can't think of a case where we would have a `nw.Series` here
2. According to [`polars.Series.str.replace_all`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.str.replace_all.html), this wouldn't be supported?
    i. It could be a documentation issue for `polars`, since it looks like this uses [`@expr_dispatch`](https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/series/string.py#L35)
    ii. Otherwise, we should do something like `PolarsExpr.from_native(pl.lit(value.native))`, and re-use the `PolarExpr*` version like in `zfill` https://github.com/narwhals-dev/narwhals/blob/277881ae00beb29c6661f5baab72892319f71f85/narwhals/_polars/series.py#L702-L706

narwhals/_spark_like/expr_str.py
camriddell: I was going to look into this today, I managed to break my local pyspark so had to first resolve that.

dangotbanned: Ahh I had `pyspark` working at some point, but haven't had much luck for a while

`sqlframe` has been fine for me though and installing with this should get you everything except `pyspark` and `cudf` 🙏 

```cmd
uv pip install -e ".[dask,modin]" --group "local-dev"  --reinstall
```

narwhals/_polars/series.py
camriddell: I believe both 1. and 2. are related to the below, let me know if this does not address your concerns.

Polars supports passing a `Series` to value in `polars.Series.str.replace_all` so I wanted to capture that behavior here as well. My intuition was that a user should be able to do the following

```python
nw.Series(…).str.replace(…, nw.Series(…))
```

However because the Polars layer is thin, we never coerce passed arguments to their native types to be handled within the `_polars.series` layer (at least when I used the function signature above, I would get a `nw.Series` passed into the function).

---

I was actually planning to open an issue on Polars because the Series objects also support more than just a replacement `str` (disagreeing with the type hint of [`value` in polars.Series.str.replace_all](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.str.replace_all.html)).

```python
>>> import polars as pl # 1.30.0
>>> import numpy as np
>>> pl.Series(['abc', 'ab def']).str.replace('ab', pl.Series(['def', 'GHI'])) # replace_all works the same
shape: (2,)
Series: '' [str]
[
	"defc"
	"GHI def"
]
>>> pl.Series(['abc', 'ab def']).str.replace('ab', np.array(['def', 'GHI']))
shape: (2,)
Series: '' [str]
[
	"defc"
	"GHI def"
]
```

dangotbanned: > i. I can't think of a case where we would have a `nw.Series` here


Okay I had a closer look and what you've written is correct from a typing perspective.

The issue is we really should've passed the `CompliantSeries` down instead, because then we can just use the `__getattr__` instead of adding new methods

I'm gonna push a fix for that now 🥳 

### Updated

[fix: Pass down `CompliantSeries` not `nw.Series`](https://github.com/narwhals-dev/narwhals/pull/2886/commits/e4e1a2450d8815423879c8db29e9d7e4700daf4f)


Weirdly, the same thing has the wrong name on `nw.Series` 🤔 

https://github.com/narwhals-dev/narwhals/blob/0e806561cd76f052fe3038a46022a9443c581568/narwhals/series.py#L376-L381

narwhals/_pandas_like/series_str.py
dangotbanned: > To me, such a good API generally supports expressified arguments

@MarcoGorelli it seems like you've been spared my thoughts of what defines a good API *for now* 😏

We now have:

- [feat: Add ~~at least `sqlframe`~~ `spark_like` support](https://github.com/narwhals-dev/narwhals/pull/2886/commits/50cd0feeeb5ee5a9ec01d18cc75dd4b942549dd7)
- [feat: `ibis` support for `replace_all`](https://github.com/narwhals-dev/narwhals/pull/2886/commits/b69fee5a1da788e50ae9d8aa0e7005197fe6bc5b)

I'm more than happy to move forward on a feature supported by everything **except**: `pandas_like`, `pyarrow`, `dask`

camriddell:
@MarcoGorelli and @dangotbanned I believe this is ready for a more formal review.

Thanks for all of your help getting this to the finish line!

MarcoGorelli:
thanks @camriddell for bearing with this one, could you resolve the conflicts please and we can try to get it in?

camriddell:
@MarcoGorelli resolved the conflicts aka incorporated the new generic `_sql` backend.

dangotbanned:
Nit: would using *elementwise* instead of *multivalue* be a better fit for the title?

camriddell:
> Nit: would using _elementwise_ instead of _multivalue_ be a better fit for the title?

I'm neutral on this, I think elementwise alone is a little ambiguous since replacement already occurs elementwise (on the input).
That said, I'm not tied to *multivalue* either because I also think its not entirely clear what is being talked about. Definitely open to refining the terms used here!

narwhals/_pandas_like/series_str.py
MarcoGorelli: i'm confused by what the first sentence means sorry

MarcoGorelli:
> Added some more explicit error checks, and delayed triggering errors unless the backend fails (this should force our tests to begin failing if the backend does support this feature in the future). After aggregating this behavior, here is a table showing the support for each feature by backend.
> 
> Backend	`str.replace[scalar]`	`str.replace[multivalue]`	`str.replace_all[scalar]`	`str.replace_all[multivalue]`
> pandas	✅	✅ (pure Python)	✅	✅ (pure Python)
> polars[eager]	✅	✅	✅	✅
> polars[lazy]	✅	✅	✅	✅
> duckdb	❌	❌	✅	✅
> spark_like	❌	❌	✅	✅
> ibis	❌	❌	✅	❌
> pyarrow	✅	❌	✅	❌
> dask	✅	❌	✅	❌

Not sure if I'm misreading this or the test (or if it's out of date), but it looks like you've got this supported for Ibis too now?

camriddell:
> > Added some more explicit error checks, and delayed triggering errors unless the backend fails (this should force our tests to begin failing if the backend does support this feature in the future). After aggregating this behavior, here is a table showing the support for each feature by backend.
> > Backend	`str.replace[scalar]`	`str.replace[multivalue]`	`str.replace_all[scalar]`	`str.replace_all[multivalue]`
> > pandas	✅	✅ (pure Python)	✅	✅ (pure Python)
> > polars[eager]	✅	✅	✅	✅
> > polars[lazy]	✅	✅	✅	✅
> > duckdb	❌	❌	✅	✅
> > spark_like	❌	❌	✅	✅
> > ibis	❌	❌	✅	❌
> > pyarrow	✅	❌	✅	❌
> > dask	✅	❌	✅	❌
> 
> Not sure if I'm misreading this or the test (or if it's out of date), but it looks like you've got this supported for Ibis too now?

It's out of date, @dangotbanned took excellent care of the Ibis support!
I'll update the table

MarcoGorelli:
gentle ping on this, i think it's really close to the finish line

camriddell:
> gentle ping on this, i think it's really close to the finish line

Thanks for the nudge, I had missed your earlier comment about `cudf`

camriddell:
@MarcoGorelli I believe this is ready for a final look through!

MarcoGorelli:
thanks, shipping it

got a feeling there's gonna be some cudf failures but we need to do some cudf fixups anyway

further expressifications welcome!

Issue #2785, opened 2025-07-04
Tracking: Making Narwhals extensible


dangotbanned:

FBruzzesi:
Not sure if this is the best issue where to leave this thought.

We might have overlooked at #2001 which has great potential and points towards a likable pattern.

In a similar way, we might need to move `Implementation` class from an enum to a registry pattern (related to #2786), in fact there is no way to extend (inject a new value) an enum without inheriting from it, but that would be _too late_ for extensions

PR #2895, opened 2025-07-26
feat: Add support for `nw.int_range` for eager backends

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

I am slightly concerned that this could turn out to be a month long PR. I will open it as draft for now. There are a couple of pain points I already know:

- I was very careful with type hints, yet it somehow doesn't pass type checker
- A custom check (importing from dtypes) will fail. That's the default used in the dtype argument. In principle we could just ignore the argument completely and always opt for Int64
- Polars can for `eager: bool`. In our case it's a bit more complex than that. Instead of adding _yet another_ argument to the function, I allowed for `eager` to be False (default), None or the backend/implementation that should back the series


narwhals/_arrow/namespace.py
```diff
@@ -284,6 +285,49 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
...
+        dtype: IntegerType | type[IntegerType],
+    ) -> ArrowExpr:
+        def func(df: ArrowDataFrame) -> list[ArrowSeries]:
+            if isinstance(start, ArrowExpr):
+                start_eval = start(df)[0]
+                name = start_eval.name
+                start_value = start_eval.item()
+            else:
+                name = "literal"
+                start_value = start
```
FBruzzesi: The reason to allow int also at this level is mostly for internal use

dangotbanned: You could add this to `_arrow.utils` to keep the two separate?


https://github.com/narwhals-dev/narwhals/blob/cf3865654ba8b731dfead17b28eb6450ad04beb1/narwhals/_plan/arrow/functions.py#L206-L219

> [!NOTE]
> The new `pyarrow.arange` requires `start` and `stop`, but it isn't documented or in the stubs yet 😅 

Just for reference, I've been mainly using that directly for internals like:

https://github.com/narwhals-dev/narwhals/blob/cf3865654ba8b731dfead17b28eb6450ad04beb1/narwhals/_plan/arrow/dataframe.py#L102-L103



Whereas the expression equivalent does some more work + a fastpath for `lit(...)`


<details><summary><code>ExprIR</code> version</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/cf3865654ba8b731dfead17b28eb6450ad04beb1/narwhals/_plan/arrow/namespace.py#L176-L211

</p>
</details>

tests/expr_and_series/int_range_test.py
```diff
@@ -0,0 +1,104 @@
...
+    [
+        (nw.col("foo", "bar").sum(), nw.col("foo", "bar").sum()),
+        (1, nw.col("foo", "bar").sum()),
+    ],
+)
+def test_int_range_multi_named(start: int | nw.Expr, end: int | nw.Expr) -> None:
+    prefix = "`start`" if isinstance(start, nw.Expr) else "`end`"
+    msg = f"{prefix} must contain exactly one value, got expression returning multiple values"
+    with pytest.raises(ComputeError, match=msg):
+        nw.int_range(start=start, end=end)  # type: ignore[arg-type]
```
FBruzzesi: How do we get: 

> error: No overloads for "int_range" match the provided arguments
>      Argument types: (Expr | int, int | Expr) (reportCallIssue)

😭 it's literally up there in the signature

FBruzzesi: Nevermind, I am fixing this

tests/v1_test.py
```diff
@@ -1037,3 +1037,15 @@ def test_dataframe_from_arrow(eager_backend: EagerAllowed) -> None:
...
+def test_int_range() -> None:
+    pytest.importorskip("pandas")
+
+    def minimal_function(data: nw_v1.Series[Any]) -> None:
+        data.is_null()
+
+    col = nw_v1.int_range(0, 3, eager="pandas")  # type: ignore[call-overload]
+    # check this doesn't raise type-checking errors
+    minimal_function(col)
+    assert isinstance(col, nw_v1.Series)
```
FBruzzesi, line 1126: This is still not fixed 😢 this shows up as `Expr | Series[Any]`

narwhals/_arrow/namespace.py
FBruzzesi: @dangotbanned do you mean the implementation in `ArrowSeries._int_range` not for the namespace one, or am I missing something?

dangotbanned:
Hey @FBruzzesi, as I've mentioned 8347839 times, I'm very keen to see this in `narwhals`! 😄

Would you be okay if we let this one marinate for a bit and maybe target the release *after* the next? 🙏

narwhals/_arrow/namespace.py
dangotbanned: > or am I missing something

I think I might need to spend some more time with the PR 🫣

I hadn't noticed at the time that there was an `ArrowSeries._int_range`. Which is pretty unfortunate as it was like 2-3 lines below this range 😂

narwhals/_arrow/series.py
```diff
@@ -161,6 +161,31 @@ def from_iterable(
...
+    ) -> Self:
+        version = context._version
+        dtype_pa = narwhals_to_native_dtype(dtype, version)
+        if cls._implementation._backend_version() < (21, 0, 0):  # pragma: no cover
+            import numpy as np  # ignore-banned-import
+
+            data = np.arange(start=start, stop=end, step=step)
+        else:
+            data = pc.cast(
+                pa.arange(start=start, stop=end, step=step),  # type: ignore[attr-defined]
```
dangotbanned: To explain the `[attr-defined]`, I guess the best we could do is add comment linking one or both of these?

- https://github.com/apache/arrow/pull/46778
- https://github.com/apache/arrow/blob/4fc0bd8a9d5d9ec63a3a84db651397a22802b6d9/python/pyarrow/array.pxi#L575-L606


For the docs, I think this would be tracking
- https://github.com/apache/arrow/issues/47127

The stubs I think would follow that and then this would be more motivation for adding another `pyarrow-stubs` bump to (#2124)

FBruzzesi:
> Hey @FBruzzesi,as I've mentioned 8347839 times, I'm very keen to see this in `narwhals`! 😄

🎉

> Would you be okay if we let this one marinate for a bit and maybe target the release _after_ the next? 🙏

Yeah I was not expecting this to be finalized by tomorrow. It has some sharp edges that need some love and work

FBruzzesi:
@MarcoGorelli any "trick" to avoid the [custom check in precommit](https://results.pre-commit.ci/run/github/760058710/1753712447.Veti8QFVReeonlh6Ux2osg)?

- For compliant namespaces the import of `IntegerType` is in the `if TYPE_CHECKING` block
- For function.py, `Int64` is used as default value (and then overwritten by the `v*.dtypes.Int64` dtype). I guess one way of doing it is to let it be None and set it inside the function given the version. Yet It's a bit less explicit to a user.

I thought about emulating that `utils/import_check.py` does, but I would need to figure out what's happening there from scratch 😂

dangotbanned:
> any "trick" to avoid the [custom check in precommit](https://results.pre-commit.ci/run/github/760058710/1753712447.Veti8QFVReeonlh6Ux2osg)?

You could add this to `nw.typing`, and then use that everywhere instead?

```py
IntegerDType: TypeAlias = "dtypes.IntegerType | type[dtypes.IntegerType]"
```

narwhals/_arrow/namespace.py
FBruzzesi: @dangotbanned WDYT about this https://github.com/narwhals-dev/narwhals/pull/2895/commits/22c52eb8495ddb964c6acfcf37203a4dafdc8007? Is it in the direction you wanted to see?

dangotbanned: > @dangotbanned WDYT about this [22c52eb](https://github.com/narwhals-dev/narwhals/commit/22c52eb8495ddb964c6acfcf37203a4dafdc8007)? Is it in the direction you wanted to see?

Yeah looking good!

.pre-commit-config.yaml
```diff
@@ -79,7 +79,8 @@ repos:
             narwhals/_utils\.py|
             narwhals/stable/v./_?dtypes.py|
             narwhals/.*__init__.py|
-            narwhals/.*typing\.py
+            narwhals/.*typing\.py|
+            narwhals/functions.py
```
FBruzzesi: Ok I have two ways to go about this one:

- Whatever is public in `functions.py` will be re-implemented in `stable/v*/__init__.py` where we are already allowed to import the types.
- The alternative is to have the `_impl_<some_function>` in `functions.py` and implement it in `narwhals/__init__.py` for the main namespace so that `Int64` is already imported

dangotbanned: I think for any `DType` where `v*` is actually `main`, we can do this:

https://github.com/narwhals-dev/narwhals/blob/cf3865654ba8b731dfead17b28eb6450ad04beb1/narwhals/_plan/demo.py#L180-L183


Although now I think I'd do something like this:

```py
from narwhals.typing import IntegerDType

dtypes = Version.MAIN.dtypes

def func(arg: int, *, arg_with_default: IntegerDType = dtypes.Int64): ...
```

FBruzzesi: Mmmh I have mixed feelings about it but I will tag along and see how it looks like 😂

dangotbanned: You could also just do this one, and add an ignore for the `NXXX` rule breaking 🙂

```py
Int64 = Version.MAIN.dtypes.Int64
```

FBruzzesi: I don't really have a preference on this - aside deciding how to pass the default dtype, I think the PR is ready for review (within its currently limited scope)

narwhals/functions.py
```diff
@@ -62,6 +63,9 @@
     _IntoSchema: TypeAlias = "Mapping[str, DType] | Schema | Sequence[str] | None"
 
 
+dtypes = Version.MAIN.dtypes
```
dangotbanned: Really sorry, I'm to blame for this! 🫣 

- https://github.com/narwhals-dev/narwhals/actions/runs/16604935514/job/46974185697?pr=2895
- https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2237523990

```py
narwhals.functions.dtypes
  /narwhals/narwhals/functions.py:66:1 - error: 
    Type is missing type annotation and could be inferred differently by type checkers
     Inferred type is "DTypes"
```

Just fixing it now 🙂

dangotbanned: Fixed in (https://github.com/narwhals-dev/narwhals/pull/2895/commits/57393ec07a130a866ec6d2b20895760ced16c622)

At least the second variant worked *mostly* https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2237557988

FBruzzesi: What dark magic is this 🧙🏼

FBruzzesi: Thanks @dangotbanned 🙏🏼

narwhals/_arrow/series.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2895/commits/704da84bb5e8bd9865a3e334457626195466c0f4)

narwhals/_arrow/utils.py
```diff
@@ -440,4 +447,17 @@ def cast_to_comparable_string_types(
     return (ca.cast(dtype) for ca in chunked_arrays), lit(separator, dtype)
 
 
+def _native_int_range(
+    start: int, end: int, step: int = 1, *, dtype: pa.DataType = int64
+) -> ArrayAny:
```
dangotbanned: nit: I'd prefer to call this `int_range`

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2240971408

> Thanks @dangotbanned 🙏🏼

@FBruzzesi It's the least I could do after goofing like that 😄 

Started the actual review now, will try to finish tomorrow
Can't seem to reply on that thread without it becoming a review comment

narwhals/typing.py
```diff
@@ -361,7 +361,8 @@ def Binary(self) -> type[dtypes.Binary]: ...
 
 NonNestedDType: TypeAlias = "dtypes.NumericType | dtypes.TemporalType | dtypes.String | dtypes.Boolean | dtypes.Binary | dtypes.Categorical | dtypes.Unknown | dtypes.Object"
 """Any Narwhals DType that does not have required arguments."""
-
+IntegerDType: TypeAlias = "dtypes.IntegerType | type[dtypes.IntegerType]"
+"""Instance or class of IntegerType"""
```
dangotbanned: Somewhat of a delicate dance around not using `IntegerType` (since we don't export it)
```suggestion
IntegerDType: TypeAlias = "dtypes.IntegerType | type[dtypes.IntegerType]"
"""Any signed or unsigned integer DType."""
```

narwhals/functions.py
```diff
@@ -1775,3 +1779,148 @@ def coalesce(
...
+        end:  End of the range (exclusive). If set to `None` (default),
+            the value of `start` is used and `start` is set to `0`.
+        step: Step size of the range.
+        dtype: Data type of the range (must be an integer data type).
+        eager: If set to `False` (default) or `None`, then an expression is returned.
+            If set to an (eager) implementation ("pandas", "polars" or "pyarrow"), then
+            a `Series` is returned.
+
+    Returns:
+        Expr or Series: Column of integer data type `dtype`.
```
dangotbanned: We might need to tweak this some more from the original

dangotbanned: It seems to render fine on the API reference, but VSCode isn't happy 😢 

> Exp
> or Series: Column of integer data type `dtype`.

<details><summary><b>Show narwhals (IDE)</b></summary>
<p>

image

</p>
</details> 

<details><summary><b>Show narwhals (API Reference)</b></summary>
<p>

image


</p>
</details> 

<details><summary><b>Show polars (API Reference)</b></summary>
<p>

image

</p>
</details>

narwhals/_compliant/series.py
```diff
@@ -122,6 +123,16 @@ def from_iterable(
...
+    @classmethod
+    def _int_range(
+        cls,
+        start: int,
+        end: int,
+        step: int,
+        dtype: IntegerDType,
+        context: _LimitedContext,
+        name: str,
+    ) -> Self: ...
```
dangotbanned: > [!IMPORTANT]
> Here's the main change I'd like to see

I want to adapt `CompliantSeries._int_range` into:

```py
class EagerNamespace(
    DepthTrackingNamespace[EagerDataFrameT, EagerExprT],
    Protocol[EagerDataFrameT, EagerSeriesT, EagerExprT, NativeFrameT, NativeSeriesT],
):
    def int_range_eager(
        self,
        start: int,
        end: int,
        step: int = 1,
        *,
        dtype: IntegerDType = Int64,
        name: str = "literal",
    ) -> EagerSeriesT: ...
```

Which would let us make this change to `functions._int_range_impl`:

```py
impl = Implementation.from_backend(eager)
if is_eager_allowed(impl):
    # ...
    ns = Version.MAIN.namespace.from_backend(impl).compliant
    return ns.int_range_eager(start, end, step, dtype=dtype).to_narwhals()
    # series = ns._series._int_range(
    #    start=start, end=end, step=step, dtype=dtype, context=ns, name="literal"
    # )
    # return series.to_narwhals()
```

*Beyond being a bit shorter*, we get:
- A consistent `name` default used everywhere
- Only the two most common arguments are required
- As a namespace instance method, we can drop `context` parameter
- A method name that is inspired by [`polars.polars.eager_int_range`](https://github.com/pola-rs/polars/blob/867443ce3875da30791021e4072e5a6fb2249d91/py-polars/polars/functions/range/int_range.py#L221)
- A more concretly defined public (for extension) `Compliant` API
  - kinda related to #2785
  - Since we're calling from `nw.functions`, I'd prefer a public name

> [!NOTE]
> Typing the return of `int_range_eager` correctly *might* force us to do a bit of (#2713)

Here's an example of me experimenting with that idea on the lazy vs eager (but not for `int_range` specifically)

<details><summary><b>Show </b><code>ExprIR</code><b> protocols</b></summary>
<p>

The important bits are where `*Series` can appear

Ignore the `*Scalar` stuff 😂 

https://github.com/narwhals-dev/narwhals/blob/cf3865654ba8b731dfead17b28eb6450ad04beb1/narwhals/_plan/protocols.py#L532-L662

</p>
</details>

dangotbanned:
@FBruzzesi I've got some more of the review done, but as an aside ...

I had another thought about `Int64` (https://github.com/narwhals-dev/narwhals/pull/2895#issuecomment-3127507076)

I'm now leaning **much more** towards making an exclusion for it in:

https://github.com/narwhals-dev/narwhals/blob/624518ca5cc3a82bffa42a9a85272e719a7f9e77/.pre-commit-config.yaml#L70

Really, these checks only make sense for the **3** `DType`s that have multiple versions:

- `Datetime`
- `Duration`
- `Enum`

**But** - for consistency - I'd propose we exclude any `DType` which **both**:
1. Has only a single version
2. Is used as an argument default


### Why?

What's swaying me more is seeing this

https://github.com/narwhals-dev/narwhals/blob/fe2bc9d06152053735bd8b7d82173ca534521917/narwhals/_arrow/dataframe.py#L472-L477


Where I'd prefer to just write with defaults, which is more readable to me:

```py
        if order_by is None:
            row_index = plx.int_range(0, len(self))
```

But currently we can't use the `DType` as a default anywhere else, despite the workarounds in (https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2240954364)

narwhals/functions.py
```diff
@@ -1775,3 +1779,148 @@ def coalesce(
...
+) -> Expr | Series[Any]:
+    """Generate a range of integers.
+
+    Arguments:
+        start: Start of the range (inclusive). Defaults to 0.
+        end:  End of the range (exclusive). If set to `None` (default),
+            the value of `start` is used and `start` is set to `0`.
+        step: Step size of the range.
+        dtype: Data type of the range (must be an integer data type).
+        eager: If set to `False` (default) or `None`, then an expression is returned.
```
dangotbanned: I think following (https://github.com/narwhals-dev/narwhals/pull/2895/commits/9ee6209e2619e47fa375471d11584e1422d72b88), `None` isn't an option anymore?

```suggestion
        eager: If set to `False` (default), then an expression is returned.
```

narwhals/functions.py
```diff
@@ -1775,3 +1779,148 @@ def coalesce(
...
+        )
+
+    impl = Implementation.from_backend(eager)
+    if is_eager_allowed(impl):
+        assert isinstance(start, int)  # noqa: S101, help mypy
+        assert isinstance(end, int)  # noqa: S101, help mypy
+        ns = Version.MAIN.namespace.from_backend(impl).compliant
+        series = ns._series._int_range(
+            start=start, end=end, step=step, dtype=dtype, context=ns, name="literal"
+        )
```
dangotbanned: Probably my most common review comment is to beg that we don't use `assert` for typing.

Currently, this is hiding the fact that we aren't handling passing expressions **and** `eager`

```suggestion
    if is_eager_allowed(impl):
        ns = Version.MAIN.namespace.from_backend(impl).compliant
        series = ns._series._int_range(
            start=start, end=end, step=step, dtype=dtype, context=ns, name="literal"
        )
```

I've added a test with two options for solving the issue.

I'm 100% fine with not supporting `Expr` for `Series`, but would like a helpful error and typing to match this difference from `polars` 🙏

dangotbanned: Those tests currently raise here

```py
        impl = Implementation.from_backend(eager)
        if is_eager_allowed(impl):
>           assert isinstance(start, int)  # noqa: S101, help mypy
                   ^^^^^^^^^^^^^^^^^^^^^^
E           AssertionError

narwhals/functions.py:1914: AssertionError
==== short test summary info ===
FAILED tests/expr_and_series/int_range_test.py::test_int_range_eager_expr[polars[eager]] - AssertionError
FAILED tests/expr_and_series/int_range_test.py::test_int_range_eager_expr[pandas] - AssertionError
FAILED tests/expr_and_series/int_range_test.py::test_int_range_eager_expr[pandas[pyarrow]] - AssertionError
FAILED tests/expr_and_series/int_range_test.py::test_int_range_eager_expr[pyarrow] - AssertionError
```

narwhals/functions.py
dangotbanned: This version works because it doesn't have a `:`

```suggestion
    Returns:
        Expr or Series of integer data type `dtype`.
```

image

dangotbanned (commented):
Thanks for working on this @FBruzzesi 😍

I've got a mix of commits, nits, suggestions, a bug (https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2242657188) and 1 larger change (https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2241049162) I wanted to discuss first

narwhals/functions.py
FBruzzesi: @dangotbanned I am addressing all the comments.

I don't think it is allowed in polars:

```py
pl.int_range(pl.col("a").min(), eager=True)
```
> ColumnNotFoundError: a

dangotbanned: Oof 😅

In that case, just this part then

> I'm 100% fine with not supporting `Expr` for `Series`, but would like a helpful error and typing

And **(1)** from https://github.com/narwhals-dev/narwhals/pull/2895/commits/4706a30f8652c143ed42a208f259f18a6b1246a9

narwhals/_compliant/series.py
FBruzzesi: Addressed all the comments :) It looks quite good I think! That's a great mental model, thanks for investigating how polars does it and suggesting it

dangotbanned:
@FBruzzesi FYI, I'm trying to work on a new regex to fix `pre-commit`

- https://results.pre-commit.ci/run/github/760058710/1753952590.XBbyHVkkQZixwog3KCnqjA
- https://github.com/narwhals-dev/narwhals/pull/2895#issuecomment-3136001189

Here's where I'm at so far https://regex101.com/r/xBMU0m/1 (strong recommend on this tool for edge cases 😄)

```py
not_quite = r"""
(?x)
    import\ narwhals(\.stable\.v\d)?\.dtypes|
    from\ narwhals(\.stable\.v\d)?\ import\ dtypes|
    from\ narwhals(\.stable\.v\d)?\.dtypes\ import\ [^I_]+
"""
```

Which actually revealed that the current pattern misses `Datetime` here:

https://github.com/narwhals-dev/narwhals/blob/466d85de1b1bc5c40c88adfa5ca4c6873e87bb41/narwhals/_arrow/series_dt.py#L30


## Updated
Some test cases here, if anyone wants to simplify 😅 https://regex101.com/r/VJqsh1/1

> **But** - for consistency - I'd propose we exclude any `DType` which both: ...

Trying to get that working is beyond my understanding of the `re` module.

[ci: Update `dtypes-import`](https://github.com/narwhals-dev/narwhals/pull/2895/commits/1575ae8ad55813b063df90e2d97a3a2c0ecb2c0e) is just focusing on the `DType`s with changes.
I do still want to keep using + recommending `Version.dtypes` everywhere except this very narrow slice of `narwhals` (https://github.com/narwhals-dev/narwhals/pull/2895#issuecomment-3136001189)

narwhals/_compliant/series.py
dangotbanned: Wonderful 😄

That helped me spot this chance to share their `int_range`

[refactor(suggestion): Move impl to `EagerNamespace.int_range`](https://github.com/narwhals-dev/narwhals/pull/2895/commits/2c52f918dfad8ae8aca460f682993b562e7716d4)

narwhals/_compliant/namespace.py
```diff
@@ -209,3 +223,48 @@ def concat(
...
+        step: int = 1,
+        *,
+        dtype: IntegerDType = Int64,
+    ) -> EagerExprT:
+        def func(df: EagerDataFrameT) -> list[EagerSeriesT]:
+            if isinstance(start, int):  # pragma: no cover
+                name = "literal"
+                start_value = start
+            else:
+                start_eval = start(df)[0]
```
dangotbanned: I kept the coverage comment, but this module is currently ignored pending (#2116)

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/pyproject.toml#L276-L277

How might we write a test that would hit this path?
Either one of `pandas` or `pyarrow` is fine 🙂

narwhals/functions.py
```diff
@@ -62,6 +63,9 @@
     _IntoSchema: TypeAlias = "Mapping[str, DType] | Schema | Sequence[str] | None"
 
 
+Int64: _Int64 = Version.MAIN.dtypes.Int64()
```
dangotbanned: Ah I forgot to remove this 🤦‍♂️ 

We need to change this line as well

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/narwhals/functions.py#L47

We can just use `Int64` directly in `int_range` (like in each backend)

dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2895/commits/1098afb2f6e064bed8e073c74188d76784809eaf)

narwhals/_compliant/namespace.py
FBruzzesi: So the idea was to allow calling this method at the compliant level.

However:

- At the narwhals level, we wrap int's with a `lit(...)`.
- At the compliant level, whenever we need int_range, we use the native implementation.

I guess it's ok to drop it?

dangotbanned: Oh I see!

Well it turns out I added a path for it in (https://github.com/narwhals-dev/narwhals/pull/2895/commits/f13d667eda841d6be643d16279b6897855a357e4) - but that's just for `pyarrow`


If we re-wrote that to *not* pass `int`s - then it simplifies all of this extra handling we'd need to do in every backend

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/narwhals/_compliant/namespace.py#L245-L262

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/narwhals/_polars/namespace.py#L225-L226

dangotbanned: Let me know how you feel about (https://github.com/narwhals-dev/narwhals/pull/2895/commits/15ae42a480fde18807a5a1fd42873869cd52d9ec), (https://github.com/narwhals-dev/narwhals/pull/2895/commits/71e98566a50f375b48befe157d407ac7832b67e5) 🙂

narwhals/functions.py
```diff
@@ -1775,3 +1777,151 @@ def coalesce(
...
+            lambda plx: apply_n_ary_operation(
+                plx,
+                lambda *args: plx.int_range(*args, dtype=dtype),
+                start,
+                end,
+                step,
+                str_as_lit=False,
+            ),
+            ExprMetadata.selector_single(),
+        )
```
FBruzzesi, line 1921: Is there a better way to return this?

narwhals/functions.py
```diff
@@ -1775,3 +1777,151 @@ def coalesce(
...
+    if is_eager_allowed(impl):
+        if not (isinstance(start, int) and isinstance(end, int)):
+            msg = (
+                f"Expected `start` and `end` to be integer values since `eager={eager}`.\n"
+                f"Found: `start` of type {type(start)} and `end` of type {type(end)}\n\n"
+                "Hint: Calling `nw.int_range` with expressions requires:\n"
+                "  - `eager=False`"
+                "  - a context such as `select` or `with_columns`"
+            )
+            raise InvalidOperationError(msg)
```
FBruzzesi, line 1933: How do we like this specific error type in here?

narwhals/_compliant/namespace.py
FBruzzesi: Thanks @dangotbanned - It definitely _looks_ much better.
I want to justify the initial implementation by saying that I try to keep the number of extra function calls as low as possible.

I am pretty happy with the overall changes. I let you the last word in the review, I left 2 comments (on the code I wrote), just to bring the attention, but non blocking to merge

narwhals/functions.py
dangotbanned: Could you be more specific?

Btw I'm not saying any these are an issue, just guessing 😄 

1. The double `lambda`?
2. The choice for `ExprMetadata`?
3. Something else?

FBruzzesi: 🙈 apologies, it's been a long long day!

> 1. The double lambda?
> 
> 2. The choice for ExprMetadata?

These should be fine as far as I understand.

> 3. Something else?

I am not the biggest fan of having to bind one argument (because of `*, dtype`) while the others are simply passed.
Also the long stack doesn't help me much. _Maybe_, this would look nicers?

```suggestion
        return Expr(
            lambda plx: apply_n_ary_operation(
                plx,
                lambda *args: plx.int_range(*args, dtype=dtype),
                *(start, end, step),
                str_as_lit=False,
            ),
            ExprMetadata.selector_single(),
        )
```

dangotbanned: > I am not the biggest fan of having to bind one argument (because of *, dtype) while the others are simply passed.

Ah yes!

Although I will raise you this instead 😉 

```suggestion
        args = start, end, step
        return Expr(
            lambda plx: apply_n_ary_operation(
                plx, partial(plx.int_range, dtype=dtype), *args, str_as_lit=False
            ),
            ExprMetadata.selector_single(),
        )
```

Just stealing an idea from https://github.com/narwhals-dev/narwhals/blob/71e98566a50f375b48befe157d407ac7832b67e5/narwhals/functions.py#L1502-L1504

narwhals/functions.py
dangotbanned: I think I'll need to sleep on it 😄 

Alone, it makes sense but in this function we have a total of:
- 3x `ComputeError`
- 1x `InvalidOperationError`
- 1x `ValueError`

I don't know the answer, but is that common in `narwhals`?

narwhals/_compliant/namespace.py
dangotbanned: Thanks for explaining @FBruzzesi!

> that I try to keep the number of extra function calls as low as possible.

I agree that aiming for fewer function calls *should* reduce overhead - and that's something we ought to be in the business of doing 😄 

Personally, I find that a challenging goal to work towards as I'd be asking *where is the logical endpoint?*
The least overhead would be to write every feature as a single function, duplicating any shared logic such that it could be used inline.
Obviously no-one is arguing for that :joy: - but *where* to draw the line is quite subjective.

When I think about performance, I like to aim for reducing **decision** points and trying to ensure an earlier **decision** doesn't need to be made again for the same path.
For this case, that meant noticing type dispatch that we do at the `narwhals`-level:

<details><summary><code>functions._int_range_impl</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/8367cc2cc7ec84d1ae9f9979c4d537af196ad54f/narwhals/functions.py#L1887-L1888

https://github.com/narwhals-dev/narwhals/blob/8367cc2cc7ec84d1ae9f9979c4d537af196ad54f/narwhals/functions.py#L1898-L1901

https://github.com/narwhals-dev/narwhals/blob/8367cc2cc7ec84d1ae9f9979c4d537af196ad54f/narwhals/functions.py#L1908

https://github.com/narwhals-dev/narwhals/blob/8367cc2cc7ec84d1ae9f9979c4d537af196ad54f/narwhals/functions.py#L1919

</p>
</details> 

Was repeated again at the `Compliant`-level (https://github.com/narwhals-dev/narwhals/pull/2895#discussion_r2246153275), both

<details><summary>on the inner function</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/narwhals/_compliant/namespace.py#L245-L253

</p>
</details>

<details><summary>and in the method itself</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/447a27666be297d7d4f2e97182cb1025d2f821e9/narwhals/_compliant/namespace.py#L258-L262

</p>
</details>

I guess what I'm trying to get at, is that while we can't reduce the cost of the *call itself* - we can reduce the cost of what happens *inside the function*.
And this is where I prefer to prematurely optimize the most 😅

narwhals/_arrow/dataframe.py
```diff
@@ -683,8 +674,9 @@ def unique(
 
             agg_func = ArrowGroupBy._REMAP_UNIQUE[keep]
             col_token = generate_temporary_column_name(n_bytes=8, columns=self.columns)
+            col_value = int_range(0, len(self))
             keep_idx_native = (
-                self.native.append_column(col_token, pa.array(np.arange(len(self))))
+                self.native.append_column(col_token, col_value)
```
dangotbanned:
```suggestion
            keep_idx_native = (
                self.native.append_column(col_token, int_range(0, len(self)))
```

narwhals/stable/v1/__init__.py
```diff
@@ -1318,6 +1319,56 @@ def scan_parquet(
...
+def int_range(
+    start: int | Expr,
+    end: int | Expr | None = None,
+    step: int = 1,
+    *,
+    dtype: IntegerDType = Int64,
+    eager: ModuleType | Implementation | str | Literal[False] = False,
+) -> Expr | Series[Any]:
+    """Generate a range of integers.
+
```
dangotbanned, line 1406: How would you feel about introducing `int_range` as unstable?

If we need to decide on some rules for (#2917) - then there's a chance the lazy support for `polars` may suddenly have restrictions

dangotbanned (approved):
Thanks again @FBruzzesi!

spinny hands narwhal

Very happy with where things are at, none of my comments are blocking  

I'm still a bit shocked by how little new backend code was needed for the feature 😄

MarcoGorelli:
this is probably fine, i'd just like to take a look before merging please, thanks 🙏

narwhals/functions.py
```diff
@@ -1798,3 +1801,152 @@ def coalesce(
...
+
+
+@unstable
+def int_range(
+    start: int | Expr,
+    end: int | Expr | None = None,
+    step: int = 1,
+    *,
+    dtype: IntegerDType = Int64,
+    eager: ModuleType | Implementation | str | Literal[False] = False,
```
dangotbanned: We'll need to update these since (#2971)

Just mentioning as it doesn't show up as a merge conflict 🙂

dangotbanned:
> this is probably fine, i'd just like to take a look before merging please, thanks 🙏

@MarcoGorelli pinging again as (https://github.com/narwhals-dev/narwhals/pull/2895#issuecomment-3146517944) was a month ago 🫣

dangotbanned:
@FBruzzesi I still have hope that eventually we'll land this 🙏 

I have an idea of how we could reuse `int_range` to implement `date_range` for `pyarrow`.

## Idea


As an example, if we convert the output of [`pl.date_range`](https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.date_range.html) to pyarrow, we can see the `pa.DataType` as:

```py
import polars as pl
import pyarrow as pa

start, end = dt.date(2000, 1, 1), dt.date(2000, 1, 5)
expr = pl.date_range(start, end).alias("date")
table = pl.select(expr).to_arrow()
table.column("date").type
```

See [`date32`](https://arrow.apache.org/docs/cpp/api/datatype.html#classarrow_1_1_date32_type)
```
DataType(date32[day])
```

Since that is just represented as a 32-bit integer, we can do things like:

```py
dates = table.column("date")
dates
```
<details><summary><b>Show Output</b></summary>
<p>

```
<pyarrow.lib.ChunkedArray object at 0x00000296A456BDC0>
[
  [
    2000-01-01,
    2000-01-02,
    2000-01-03,
    2000-01-04,
    2000-01-05
  ]
]
```

</p>
</details> 

```py
dates.cast(pa.int32())
```

<details><summary><b>Show Output</b></summary>
<p>

```
<pyarrow.lib.ChunkedArray object at 0x00000296A456B8E0>
[
  [
    10957,
    10958,
    10959,
    10960,
    10961
  ]
]
```

</p>
</details> 


```py
dates.cast(pa.int32()).cast(pa.date32())
```

<details><summary><b>Show Output</b></summary>
<p>

```
<pyarrow.lib.ChunkedArray object at 0x00000296A456BAC0>
[
  [
    2000-01-01,
    2000-01-02,
    2000-01-03,
    2000-01-04,
    2000-01-05
  ]
]
```

</p>
</details> 

### In practice

```py
import datetime as dt

import pyarrow as pa


def date_range(
    start: dt.date,
    end: dt.date,
    interval: int,  # (* assuming the `Interval` part is solved)
    *,
    closed: ClosedInterval = "both",
) -> pa.Date32Array:
    start_i = pa.scalar(start).cast(pa.int32()).as_py()
    end_i = pa.scalar(end).cast(pa.int32()).as_py()

    # call `int_range` here for the compatibility branch
    arr = pa.arange(start_i, end_i + 1, interval)
    if closed != "both":
        if closed == "left":
            arr = arr.slice(length=len(arr) - 1)
        elif closed == "none":
            arr = arr.slice(1, len(arr) - 1)
        else:
            arr = arr.slice(1)

    # the first cast would happen in `int_range(dtype=...)`
    return arr.cast(pa.int32()).cast(pa.date32())


start, end = dt.date(2000, 1, 1), dt.date(2000, 2, 1)

date_range(start, end, interval=7, closed="none")
<pyarrow.lib.Date32Array object at 0x00000296A74176A0>
[
  2000-01-08,
  2000-01-15,
  2000-01-22,
  2000-01-29
]
```
## What's the catch?
We'd need to adapt `Interval` to support a different kind of parse.

<details><summary><b>Show <code>Interval</code></b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/01aab21f872f97c1b8d4731825f3c40e0f3a1141/narwhals/_duration.py#L49-L94

</p>
</details> 

It would be restricted to only `d`, `w`, `mo`, `q`, `y` - but even within that - maybe more flexible given that we know `+1d` is equivalent to `+1` 🤔?

narwhals/functions.py
```diff
@@ -1786,3 +1789,152 @@ def format(f_string: str, *args: IntoExpr) -> Expr:
...
+
+@unstable
+def int_range(
+    start: int | Expr,
+    end: int | Expr | None = None,
+    step: int = 1,
+    *,
+    dtype: IntegerDType = Int64,
+    eager: IntoBackend[EagerAllowed] | Literal[False] = False,
+) -> Expr | Series[Any]:
```
dangotbanned, line 1824: > [!NOTE]
> I don't think this blocks anything - just a realization from me 🙂 


I hadn't been able to put my finger on what seemed off to me wrt this until just now:

> Polars can do `eager: bool`. In our case it's a bit more complex than that.
> Instead of adding yet another argument to the function, I allowed for `eager` to be False (default), None or the backend/implementation that should back the series.


This trick would work for us in all the cases where `eager` defines `Expr | Series`.
That is *most* of them, but there is [`polars.select`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.select.html) as the ugly duckling with:


<details><summary><code>DataFrame | LazyFrame</code></summary>
<p>

```py
@overload
def select(
    *exprs: IntoExpr | Iterable[IntoExpr],
    eager: Literal[True] = ...,
    **named_exprs: IntoExpr,
) -> DataFrame: ...


@overload
def select(
    *exprs: IntoExpr | Iterable[IntoExpr],
    eager: Literal[False],
    **named_exprs: IntoExpr,
) -> LazyFrame: ...


def select(
    *exprs: IntoExpr | Iterable[IntoExpr], eager: bool = True, **named_exprs: IntoExpr
) -> DataFrame | LazyFrame:
```

</p>
</details> 

If we added `select`, then we'd need two arguments to be able to say whether we want `pl.DataFrame` or `pl.LazyFrame`


```py
def select(
    *exprs: IntoExpr | Iterable[IntoExpr],
    backend: IntoBackend[Backend],
    eager: bool = True,
    **named_exprs: IntoExpr,
) -> DataFrame | LazyFrame: ...
```

But that also seems a bit footgun-y, since something like `select(..., backend="duckdb")` would have a conflicting default.

Issue #2474, opened 2025-05-03
Minimizing overhead for `polars` proxy methods


dangotbanned:
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/2451#discussion_r2066072950_

Just collecting the comments from (#2451) for now so they don't get lost.

The next steps would be:
- [ ] Decide on some benchmarks
- [ ] Try out the alternatives in comments 3 & 4
- [ ] If we see a meaningful benefit, rewrite `Polars*` methods to avoid using `__getattr__`

<details><summary>Comment 1 (@dangotbanned)</summary>
<p>

If we're defining all the cases up front, have you thought about using a descriptor instead?
`__getattr__` makes sense at the moment because we're *theoretically* allowing things *outside* of the `narwhals` API to pass through.
In *practice*, I understand that might not be possible from the user-side - but it could mean we can just rely on `python` to raise the `AttributeError` instead of checking and raising ourselves 🤔

</p>
</details> 

<details><summary>Comment 2 (@dangotbanned)</summary>
<p>

> 2. I'm not familiar with descriptors, but open to suggestions

If you're looking to include this PR in a bugfix - we can punt this to another issue?

Some useful homework though 😉 

- [datamodel>Class instances](https://docs.python.org/3/reference/datamodel.html#id4)
- [`object.__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__)
- [datamodel>descriptors](https://docs.python.org/3/reference/datamodel.html#descriptors)
- [`object.__dict__`](https://docs.python.org/3/reference/datamodel.html#object.__dict__)

In short, I think we can reduce the performance overhead of **always** looking up (and failing) in order to reach `__getattr__` - since you've defined the full set of methods that are permitted.
So the idea would be to move a cost that is currently paid *per-instance* **and** *per-method* into one that is *per-type* for the fixed set of methods 😏 

</p>
</details> 

<details><summary>Comment 3 (@camriddell)</summary>
<p>

After looking at this might I propose the use of a decorator to dynamically add these methods to the class and then do away with the dynamic look up/attribute fetching via `__getattr__` altogether?

The pattern here typically looks like

```python
def define_attributes(attributes):
    def decorator(cls):
        for attr in attributes:
            setattr(cls, attr, ...) # insert dynamically defined function here
        return cls
    return decorator
         
@define_attributes(['min', 'max'])
class T:
    pass

t = T()
[...]

PR #2897, opened 2025-07-27
refactor: Shrink `Polars*Namespace`(s)

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- #1942
- #2474

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_utils.py
```diff
@@ -2075,3 +2076,14 @@ def to_tuple(self) -> tuple[_T, ...]:
...
+
+@lru_cache(maxsize=64)
+def deep_attrgetter(attr: str, *nested: str) -> attrgetter[Any]:
+    name = ".".join((attr, *nested)) if nested else attr
+    return attrgetter(name)
+
+
+def deep_getattr(obj: Any, name_1: str, *nested: str) -> Any:
+    """Perform a nested attribute lookup on `obj`."""
+    return deep_attrgetter(name_1, *nested)(obj)
```
dangotbanned, line 2037: I've added these here since they're generally useful and can get more mileage anywhere we use `getattr`.

IIUC, this is a reimplementation:

https://github.com/camriddell/narwhals/blob/995ab38337bbd8abe174c2a158e50a3959317ed6/narwhals/_compliant/expr.py#L423-L426

But we'd miss out on:

1. Everything in [`operator`](https://docs.python.org/3/library/operator.html) has an optimized C implementation (https://github.com/python/cpython/blob/2bd4ff07001038fada17b6dbcb5ab5aabd27c26e/Modules/_operator.c) vs defining a nested function
2. `deep_attrgetter` has a cache that can be shared across the whole of `narwhals`

dangotbanned:
FYI, this is the first in a series that'll eventually have a much bigger negative diff + improved typing

narwhals/_polars/expr.py
```diff
@@ -256,9 +262,26 @@ def struct(self) -> PolarsExprStructNamespace:
...
+    ) -> Self:
+        raise NotImplementedError
+
+    @classmethod
+    def from_column_indices(cls, *column_indices: int, context: _LimitedContext) -> Self:
+        raise NotImplementedError
+
+    @staticmethod
+    def _eval_names_indices(indices: Sequence[int], /) -> EvalNames[PolarsDataFrame]:
+        raise NotImplementedError
```
MarcoGorelli, line 284: why change this? or even why not just use `= not_implemented()`?

narwhals/_polars/series.py
```diff
@@ -681,92 +687,57 @@ def list(self) -> PolarsSeriesListNamespace:
...
 
+class PolarsSeriesListNamespace(
+    PolarsSeriesNamespace, PolarsListNamespace[PolarsSeries, pl.Series]
+):
     def len(self) -> PolarsSeries:
-        native_series = self._series.native
-        native_result = native_series.list.len()
+        name = self.name
+        ns = self.__narwhals_namespace__()
+        return self.to_frame().select(ns.col(name).list.len()).get_column(name)
```
MarcoGorelli, line 738: i guess this is fine, just not sure i see why change it?

MarcoGorelli (approved):
looks good, thanks @dangotbanned , just got a couple of questions (non-blocking)

narwhals/_polars/expr.py
dangotbanned: aha take a look at the **show logs** in this (https://github.com/narwhals-dev/narwhals/pull/2918#issuecomment-3141331583) 😲 

This solution is very much temporary - I hadn't thought to try `not_implemented` on a `Polars*` class though

oh and this diff https://github.com/narwhals-dev/narwhals/commit/e203b52a85974446e5dd434c63bdf9f510be4022

narwhals/_polars/series.py
dangotbanned: Because we have `len` at home 😉 

https://github.com/narwhals-dev/narwhals/blob/7d5a457c15cdf12c8eda5bb75759b4a1e9316cb5/narwhals/_polars/expr.py#L454-L465

So now we have that duplicated version branching reduced to a single path

But, with the added benefit of not materialising any intermediate `Series` 😎

MarcoGorelli: ah right, good one

PR #2900, opened 2025-07-28
[Enh]: Add shortened aliases for supported __dunder__ methods on expr

### LINKED ISSUE #2213: [Enh]: Add shortened aliases for supported __dunder__ methods on expr

_No response_

The Polars api has shortened aliases for certain methods like __sub__, e.g. https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.sub.html. It would be nice and I think fairly low effort to add these to narwhals as well. 

I would love to take a stab at implementing these if there is interest!

_No response_

_No response_

CangyuanLi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2213
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

This PR adds shortened aliases (e.g. `sub`, `lt`) for supported binary operators (e.g. __sub__, __lt__), mimicing the Polars API. This makes method chaining much more convenient.


CangyuanLi:
Hi all, thanks for bearing with me on my first PR! Pre-commit is failing because there are methods on Expr that are not on Series. What is narwhal's policy on breaking from the polars API? In polars, these methods are only implemented on Expr, not Series.

FBruzzesi:
> Hi all, thanks for bearing with me on my first PR!

Hey @CangyuanLi thanks for taking the initiative 🙏🏼

> Pre-commit is failing because there are methods on Expr that are not on Series. What is narwhal's policy on breaking from the polars API? In polars, these methods are only implemented on Expr, not Series.

You can add these methods in https://github.com/narwhals-dev/narwhals/blob/f99778e24e2c2220af85ce8b45a7a516b323e24b/utils/check_api_reference.py#L63 so that the check should pass

MarcoGorelli:
thanks for your pr! sorry, I forgot to reply to the issue

tbh I'm not totally sure about these. You can already type out `__sub__` / etc. if you want to chain the operations, i'm not sure about expanding the api with these

on the other hand, these are nice to have documented, whereas it's a bit awkward to document the `__` methods (and I think most users wouldn't know what they do)

i'll think about this a little more

CangyuanLi:
Thank you! I agree that its more of a nice-to-have. My argument in favor would be that it makes reasoning about the order of operations much faster (at least for me). Calling the dunder methods does that as well but does introduce visual clutter.

Issue #1945, opened 2025-02-05
request for contributions: add some end-to-end tests from modern polars


MarcoGorelli:
In Narwhals we have some end-to-end TPC-H tests, which have been really useful to find some issues not covered by unit tests

Modern Polars is published here: https://kevinheavey.github.io/modern-polars/

it would be good to take some examples from there and make a `modern_polars_test` file in narwhals which executes some of them and asserts the output matches expected output

if any example from there requires a feature missing from narwhals, please open an issue to discuss it

please make sure that the tests are runnable without large data files - we should be able to generate a small test dataset ourselves or save a small test file in `tests/data`

CLWilliam:
Hi there! Looking at working at this one! are you aiming at only trying things out on the smaller datasets as saved on [the modern polars small datasets](https://github.com/kevinheavey/modern-polars/tree/master/data)? they're all quite managable size-wise!

MarcoGorelli:
Hey - yeah small datasets so we can run in ci

skritsotalakis:
@MarcoGorelli Is this still in scope? I can move it forward right away

MarcoGorelli:
yup, sure, thanks! [#2177](https://github.com/narwhals-dev/narwhals/pull/2177) is a good starting point

MarcoGorelli:
a bunch of these have been added, thanks Stelios!

PR #2903, opened 2025-07-28
tests: more modern_polars tests

skritsotalakis:
Adding tests for:
 - ewm_mean
 - filter

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [X] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #1945 

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/modern_polars/ewm_mean_test.py
+++ b/tests/modern_polars/ewm_mean_test.py
@@ -6,12 +6,13 @@ import narwhals as nw
 from tests.utils import POLARS_VERSION, ConstructorEager, assert_equal_data
 
 
-@pytest.mark.skipif(POLARS_VERSION < (1, 1), reason="requires polars 1.1+")
 def test_ew_mean(
     constructor_eager: ConstructorEager, request: pytest.FixtureRequest
 ) -> None:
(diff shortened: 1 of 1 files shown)


tests/modern_polars/ewm_mean_test.py
```diff
@@ -0,0 +1,68 @@
+from __future__ import annotations
+
+import pytest
+
+import narwhals as nw
+from tests.utils import POLARS_VERSION, ConstructorEager, assert_equal_data
+
+
+@pytest.mark.skipif(POLARS_VERSION < (1, 1), reason="requires polars 1.1+")
```
MarcoGorelli: instead of this, could we do something like
```
if 'polars' in str(constructor_eager) and POLARS_VERSION < (1,1):
    pytest.skip()
```
?

otherwise it would skip it for other backends even if their versions are fine

skritsotalakis: that's right! thanks for the review.

 I made a new commit to address this

PR #2910, opened 2025-07-29
ci: Run pyspark constructors via path

### LINKED ISSUE #2840: ci: Trigger PySpark workflow on changes to `narwhals/_spark_like/*.py`
## Related
- https://github.com/narwhals-dev/narwhals/pull/2561#issuecomment-2888338995
- https://github.com/narwhals-dev/narwhals/pull/2561#issuecomment-2888369728
- #2621
- #2562
- #2507

> > @dangotbanned 
> > 
> > These failures _seem_ related to these recent PRs that didn't run the `pyspark` ci?
> > If that is the case - might we need a more automated way of enabling those workflows? 
> > Not sure there's much we can do for (#2557), but for (#2549) if the CI could be triggered by detecting changes to `_spark_like` that could help 🙏
> > No idea if that's possible, but seems like a problem that must come up in other repos
> 
> @FBruzzesi 
> Yeah it should be possible but not sure how easy that is.

## Description
I just remembered these comments from 2 months ago and finally followed up with finding the right docs 🙂 

https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events

> When using the `push` and `pull_request` events, you can configure a workflow to run based on what file paths are changed. Path filters are not evaluated for pushes of tags.

https://github.com/narwhals-dev/narwhals/blob/bb037039f2b97419a1facefa860d2cce6965a60e/.github/workflows/pytest-pyspark.yml#L1-L7

### LINKED ISSUE #2909: ci: only run pyspark-pytest jobs on one python version
https://github.com/narwhals-dev/narwhals/blob/4a5bccb72768d5d27562d71a5f7242119fa120a0/.github/workflows/pytest-pyspark.yml#L13-L17

https://github.com/narwhals-dev/narwhals/blob/4a5bccb72768d5d27562d71a5f7242119fa120a0/.github/workflows/pytest-pyspark.yml#L41-L46

just 3.11 should be enough

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2909 and closes #2840 

---

I am not sure if we should (and even can) keep the label triggers.
Another idea could be to have a cron once a week

### DIFF
--- a/.github/workflows/pytest-pyspark.yml
+++ b/.github/workflows/pytest-pyspark.yml
@@ -2,7 +2,10 @@ name: PyTest PySpark constructor
 
 on:
   pull_request:
-    types: [opened, labeled, synchronize]  # Triggers on PR creation, updates, and label changes
+    paths:
+      - narwhals/_expression_parsing.py
+      - narwhals/_spark_like/**
+      - narwhals/_sql/**
 
 env:
   PY_COLORS: 1
@@ -11,10 +14,9 @@ env:
 jobs:
 
   pytest-pyspark-constructor:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'pyspark') || contains(github.event.pull_request.labels.*.name, 'spark-like') || contains(github.event.pull_request.labels.*.name, 'release')}}
     strategy:
       matrix:
-        python-version: ["3.10", "3.11"]
+        python-version: ["3.11"]
         os: [ubuntu-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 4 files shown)


dangotbanned:
Thanks @FBruzzesi!

Could you try adding a change to one of the paths to test it gets picked up?

Maybe something like adding a module-level doc, saying any change to this module will trigger the `pyspark` job?

FBruzzesi:
> Thanks @FBruzzesi!
> 
> Could you try adding a change to one of the paths to test it gets picked up?
> 
> Maybe something like adding a module-level doc, saying any change to this module will trigger the `pyspark` job?

Yes I am on it! I was looking for a dumb change to make, but a doc line makes sense

FBruzzesi:
WDYT about the cron idea? Maybe every Sunday to make sure we have a run before the release?

MarcoGorelli:
sounds good - this + cron and remove the labels, might be the best combination?

MarcoGorelli:
awesome, thanks!

PR #2915, opened 2025-07-31
chore: factor out log from SparkLike, DuckDB, and Ibis

ym-pett:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_sql/expr.py
+++ b/narwhals/_sql/expr.py
@@ -513,16 +513,16 @@ class SQLExpr(LazyExpr[SQLLazyFrameT, NativeExprT], Protocol[SQLLazyFrameT, Nati
     def log(self, base: float) -> Self:
         def _log(expr: NativeExprT) -> NativeExprT:
             return self._when(
-                expr < self._lit(0), # type: ignore[operator]
-                self._lit(float("nan")),  
+                expr < self._lit(0),  # type: ignore[operator]
+                self._lit(float("nan")),
                 self._when(
-                    expr == self._lit(0), # type: ignore[operator]
+                    expr == self._lit(0),  # type: ignore[operator]
                     self._lit(float("-inf")),
                     self._function("log", expr),
-                )  
+                ),
             )
 
-        return self._with_elementwise(_log)  
+        return self._with_elementwise(_log)
 
     # Cumulative
     def cum_sum(self, *, reverse: bool) -> Self:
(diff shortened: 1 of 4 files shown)


narwhals/_sql/expr.py
```diff
@@ -509,7 +509,16 @@ def _sqrt(expr: NativeExprT) -> NativeExprT:
...
+    
+    def log(self, base: float) -> Self:
+        def _log(expr: NativeExprT) -> NativeExprT:
+            return self._when(
+                expr < self._lit(0), self._lit(float("nan")), # type: ignore[operator]
+                self._when(expr == self._lit(0), self._lit(float("-inf")), self._function("log", expr)) # type: ignore[operator]
+            )
+        
+        return self._with_elementwise(lambda expr: self._function("log", expr)) # type: ignore[operator]
+
```
ym-pett: I can see why this is wrong, I need to find a way to say 

CASE
  WHEN expr < 0 THEN NaN
  WHEN expr = 0 THEN -inf
  ELSE LOG(expr)
END

the other implementations seem to work with chaining .when, but I'm not managing to make that work for the moment. to be continued!

narwhals/_sql/expr.py
```diff
@@ -510,6 +510,20 @@ def _sqrt(expr: NativeExprT) -> NativeExprT:
...
+                self._lit(float("nan"))),  
+                self._when(
+                    expr == self._lit(0),  # type: ignore[operator]
+                    self._lit(float("-inf")),
+                    self._function("log", expr),
+                ),
+            )
+
+        return self._with_elementwise(_log)
+
```
ym-pett, line 526: still think the scope of this statement is wrong & that I need to find out how to chain _when, as it is done in the other log functions.

(the otherwise needs to apply to both conditions, at the moment it applies only to the second)

narwhals/_sql/expr.py
```diff
@@ -510,6 +510,20 @@ def _sqrt(expr: NativeExprT) -> NativeExprT:
     def exp(self) -> Self:
         return self._with_elementwise(lambda expr: self._function("exp", expr))
 
+    def log(self, base: float) -> Self:
+        def _log(expr: NativeExprT) -> NativeExprT:
+            return self._when(
+                expr < self._lit(0),  # type: ignore[operator]
+                self._lit(float("nan")),
+                self._when(
+                    expr == self._lit(0),  # type: ignore[operator]
```
MarcoGorelli: type checkers think that the result of `==` is `bool`, whereas it's actually `NativeExprT`

for now I'd suggest `cast("NativeExprT", expr == self._lit(0))`

PR #2920, opened 2025-08-01
chore: fixup operator type ignores for `SQLExpr`

ym-pett:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_sql/typing.py
```diff
@@ -12,3 +12,29 @@
 SQLExprT = TypeVar("SQLExprT", bound="SQLExprAny")
 SQLExprT_contra = TypeVar("SQLExprT_contra", bound="SQLExprAny", contravariant=True)
 SQLLazyFrameT = TypeVar("SQLLazyFrameT", bound="SQLLazyFrameAny")
+
+
+# it needs to inherit from SQLExpr, but getting errors if passing SQLExprT. I have not sorted out what relation
+# (covarinant, contravariant or intravariant it needs to be yet) it needs to have to its parent class
+class NativeSQLExpr(SQLExprT):
```
MarcoGorelli: This should inherit from `NativeExpr`

narwhals/_sql/typing.py
```diff
@@ -12,3 +12,29 @@
...
+    # not sure how to initialise the class, I want it to have all the methods etc of SQLExprT and add its own operator
+    # methods in addition to those
+    def __init__(self, other: [Int | Float]) -> None:
+        super().__init__(other)
```
MarcoGorelli: this class will only be used for type-checking purposes, so there's no need to fill in the implementations for any methods, in particular, you don't need to define `__init__`

for the others, you can just return `...`

narwhals/_sql/typing.py
```diff
@@ -12,3 +14,6 @@
 SQLExprT = TypeVar("SQLExprT", bound="SQLExprAny")
 SQLExprT_contra = TypeVar("SQLExprT_contra", bound="SQLExprAny", contravariant=True)
 SQLLazyFrameT = TypeVar("SQLLazyFrameT", bound="SQLLazyFrameAny")
+
+
+class NativeSQLExpr(NativeExpr): ...
```
MarcoGorelli: nice! this is the idea

you can then implement `__eq__`, `__le__`, ...etc within this class, but for each one, just return `...`

then:
- make `NativeSQLExprT`
- whenever `NativeExprT` is used in `narwhals/_sql`, use `NativeSQLExprT` instead

narwhals/_sql/typing.py
```diff
@@ -2,13 +2,39 @@
...
     SQLLazyFrameAny = SQLLazyFrame[Any, Any, Any]
 
 SQLExprT = TypeVar("SQLExprT", bound="SQLExprAny")
 SQLExprT_contra = TypeVar("SQLExprT_contra", bound="SQLExprAny", contravariant=True)
 SQLLazyFrameT = TypeVar("SQLLazyFrameT", bound="SQLLazyFrameAny")
+
+
+class NativeSQLExpr(NativeExpr):
+    # TODO @mp: fix input type for all these!
+    def __gt__(self, value: float) -> Boolean: ...
```
MarcoGorelli: the return type should be `Self`
```python
In [11]: import duckdb

In [12]: duckdb.ColumnExpression('a') == 3
Out[12]: (a = 3)

In [13]: type(duckdb.ColumnExpression('a') == 3)
Out[13]: duckdb.duckdb.Expression
```
when you do arithmetic / comparisons between these expressions, the output is still an expression

ym-pett: ah thanks for that, now I understand why `Self`!

ym-pett:
pytests seem to suggest something going on with the arguments to NativeSQLExprT, I think I do need to define the arguments by making a NativeSQLExprTAny; looking into this!

but NativeSQLExpr doesn't take any arguments, so unlikely to be the solution..

MarcoGorelli:
> I think I do need to define the arguments by making a NativeSQLExprTAny; looking into this!

yup, that's right!

> but NativeSQLExpr doesn't take any arguments, so unlikely to be the solution..

~~`NativeSQLExpr` doesn't, but `NativeSQLExprT` does~~

EDIT: discussed over a call how to resolve this

ym-pett:
> > I think I do need to define the arguments by making a NativeSQLExprTAny; looking into this!
> 
> yup, that's right!
> 
> > but NativeSQLExpr doesn't take any arguments, so unlikely to be the solution..
> 
> `NativeSQLExpr` doesn't, but `NativeSQLExprT` does

ah just seen this, thanks!

ym-pett:
hmm i can see *why* the typing errors happen, i'm not sure how to fix though. I'll add comments to the files where the errors are flagged.

ym-pett:
ibis typing problems: 

```
/Users/mpettinato/devel/narwhals/narwhals/_ibis/expr.py:41:41 - error: Type "Value" cannot be assigned to type variable "NativeSQLExprT@SQLExpr"
    Type "Value" is not assignable to upper bound "NativeSQLExpr" for type variable "NativeSQLExprT@SQLExpr"
      "Value" is incompatible with protocol "NativeSQLExpr"
        "__add__" is not present
        "__sub__" is not present
        "__truediv__" is not present (reportInvalidTypeArguments)
```

- i'm wondering if we need to pull a similar trick here, not sure.. though the other frameworks manage to inherit fine, so unlikely we need to construct something ibis specific since the whole idea was to let it inherit..

narwhals/_sql/typing.py
```diff
@@ -1,14 +1,47 @@
...
+
+    def __lt__(self, value: Any, /) -> Self: ...
+
+    def __ge__(self, value: Any, /) -> Self: ...
+
+    # def __le__(self, value: Self) -> Self: ...
+
+    # def __eq__(self, value: Self) -> Self: ...
+
+    # def __ne__(self, value: Self) -> Self: ...
```
MarcoGorelli: you might need a `# type: ignore[override]` if you want to add these back, this is expected as here `__eq__` does something different than it does in base Python where it always returns a boolean

MarcoGorelli:
ooh nice, well done getting the other backends to pass!! 🙌  nice to see some `type: ignore`s gone

ibis' typing is pretty poor tbh, I think we can just add a `pyright: ignore[reportInvalidTypeArguments]` to that line

---

I'm also curious - what happens if you add these new methods you've added (`__add__`, `__ge__`, ...) to `NativeExpr` itself? If it passes like that, then we can avoid introducing `NativeSQLExpr`. i don't know if that'll work though, haven't tried

ym-pett:
> pyright: ignore[reportInvalidTypeArguments]

ah sorry, hadn't seen your message yet but I'm now using your ignore message.

I also realised I have a bunch of commented out (surprised ruff didn't complain) code of other operator methods in the `NativeSQLExpr`. Should I delete those or implement them, even though they're not actually called anywhere as far as I can tell. I'd say the former? 

finally, will try your suggestion tomorrow!

ym-pett:
ah shoot, new tests failing, will check tomorrow!

MarcoGorelli:
the tubular test failure is unrelated, you can ignore it

> Should I delete those or implement them, even though they're not actually called anywhere as far as I can tell. I'd say the former?

i think it's fine to keep these standard arithmetic ones, we might use them in the future and it'll be less surprising to contributors

> finally, will try your suggestion tomorrow!

cool, thanks!

ym-pett:
up to commit [27e748f](https://github.com/narwhals-dev/narwhals/pull/2920/commits/27e748f45da59852b60194e2dad2162285667746) works and is using the new `NativeSQLExpr`. 

After is experimental and not working. I'm currently trying to understand why downstream tests no longer seem to have access to `NativeExprT` after moving the operator functions there. 

TODO: reminder to self, spotted another leftover TODO @mp, will have to clean before submitting actual PR

ym-pett:
ok managed to get all the pytests to pass, but ibis type checking still to be addressed plus need to merge main back in

ym-pett:
2 ibis type errors remain - don't know if my technique is right, is there a better way?

MarcoGorelli:
I'm seriously tempted to say that we should just switch
```
class IbisExpr(SQLExpr["IbisLazyFrame", "ir.Value"])
```
to
```
class IbisExpr(SQLExpr["IbisLazyFrame", Any])
```

Afterall, Ibis doesn't export its type annotations, nor do they even check them, so I don't think it's worth bending backwards over their non-public types anyway. We're having to introduce `cast` / `type: ignore` in every other ibis function anyway... 😩 

cc @dangotbanned in case you have thoughts

ym-pett:
re going for an 'Any' type, let me know if/when I can implement or if I should try to find another fix

dangotbanned:
Thanks for the ping @MarcoGorelli 

Let me have a think on it today

dangotbanned:
> Afterall, Ibis doesn't export its type annotations, nor do they even check them, so I don't think it's worth bending backwards over their non-public types anyway. 
> We're having to introduce cast / type: ignore in every other ibis function anyway... 😩
 
> Let me have a think on it today

My preference would be to open issue(s) upstream *first*.
Back in (https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2047704276), I noticed that a lot of our `ibis` typing woes would be fixed by some methods using `Self` as the return type.

These [operator issues](https://github.com/narwhals-dev/narwhals/actions/runs/16752902329/job/47427605742?pr=2920) would also fall into that category as we want the methods defined in https://github.com/ibis-project/ibis/blob/db1a727b3c4c75e8e4af0f7ef3d0101d26d4450b/ibis/expr/types/numeric.py and not `ir.Value` itself.

From the looks of (https://github.com/ibis-project/ibis/issues/11511), I'm getting the feeling that the `ibis` maintainers *might* be open to improvements if we propose them 🙏 

We do have the benefit of an entire sub-package documenting where we've struggled 😂

---

If we don't have any success there, another option is *isolating* the problematic paths and wrapping them with our own annotations.
I've done this a lot in (https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_arrow/utils.py), (https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_arrow/typing.py), but there's stuff like this all over `narwhals`

<details><summary><b>Show examples</b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_pandas_like/utils.py#L396-L404

https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_pandas_like/utils.py#L657

https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_pandas_like/namespace.py#L374-L410

https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_namespace.py#L102-L150

https://github.com/narwhals-dev/narwhals/blob/a8b89fd268f2031b52dfe0548e728fcacbc90a04/narwhals/_spark_like/namespace.py#L53-L60

</p>
</details> 

While this is *an option*, I wouldn't suggest it until we know that the only alternative is typing everything as `Any`

MarcoGorelli:
All I can think to suggest is, in `narwhals/_ibis/typing.py`, to introduce
```
class Expression(ir.Value, Protocol):
    def __ge__(self, other: Any, /) -> Expression: ...
    def __le__(self, other: Any, /) -> Expression: ...
    # etc.
```
and fill in the missing bits

Alternatively, use `Any` when defining the classes, but keep `ir.Column` / `ir.NumericColumn` in some methods

perhaps i'll give this a go later

dangotbanned:
> All I can think to suggest is, in `narwhals/_ibis/typing.py`, to introduce
> 
> ```py
> class Expression(ir.Value, Protocol):
> ```

Sadly this won't work IIRC, as a `Protocol` can't inherit from a *non-`Protocol`* class 😔

MarcoGorelli:
🤔 i'm wondering if `Deferred` would be a better candidate than `ir.Value`, as that's closer to `Expression` / `Column`
```python
In [1]: import ibis

In [2]: ibis._['a']
Out[2]: _['a']

In [3]: type(_)
Out[3]: ibis.common.deferred.Deferred
```

dangotbanned:
@MarcoGorelli, @ym-pett 

How do you feel about this?

- [fix(suggestion): Remove problematic protocol members](https://github.com/narwhals-dev/narwhals/pull/2920/commits/dcf1b800dd514808555f81de3861b5d5a4cc4205)

I've tried to keep the terseness of using operators through aliasing.

This works due to *most* of these functions being typed more permissively

- https://github.com/python/typeshed/blob/ecf2ef48f96f596c4ff4374161675596b9c8c569/stdlib/_operator.pyi

narwhals/_compliant/expr.py
```diff
@@ -64,6 +64,17 @@ class NativeExpr(Protocol):
...
+    # NOTE: None of these are annotated for `dx.Series`, but are added imperatively
+    # Probably better to define a sub-protocol for `NativeSQLExpr`
+    # - match `dx.Series` to `NativeExpr`
+    # - match the others to `NativeSQLExpr`
+    def __gt__(self, value: Any, /) -> Self: ...
+    def __lt__(self, value: Any, /) -> Self: ...
+    def __ge__(self, value: Any, /) -> Self: ...
+    def __le__(self, value: Any, /) -> Self: ...
+    def __eq__(self, value: Any, /) -> Self: ...  # type: ignore[override]
+    def __ne__(self, value: Any, /) -> Self: ...  # type: ignore[override]
```
dangotbanned, line 75: For context:

- https://github.com/dask/dask/blob/d0a9d158ff1dc8d29145311b81365c28ee82474c/dask/dataframe/dask_expr/_collection.py#L4609-L4632
- https://github.com/dask/dask/blob/d0a9d158ff1dc8d29145311b81365c28ee82474c/dask/dataframe/dask_expr/_collection.py#L191-L280

MarcoGorelli: oh my 🤦

MarcoGorelli, line 75: looks good, thanks - if anything, I like it better if we match on these than if we match on `between` and `isin` (which, for example, Daft doesn't have)

MarcoGorelli (commented):
looks good to me, thanks @dangotbanned 

So, the idea is to put the comparison operators in `NativeExpr`, and for the arithmetic ones we just use `operator`? And just ignore for Dask, which implements these in a hacky way anyway

@ym-pett I think if you clean up the temporary comments, then I think we can ship this

narwhals/_compliant/expr.py
dangotbanned: This is the important bit

https://github.com/narwhals-dev/narwhals/blob/5bcdb6e6e1e12f430686cd5b7ea9631419a3e7bd/narwhals/_compliant/expr.py#L57-L62

> which, for example, Daft doesn't have

I had a similar issue for `ibis.Table` in (#2944)

https://github.com/narwhals-dev/narwhals/blob/0d4b6f7011ee3a70becbd8adc7ea88365d6606ff/narwhals/typing.py#L124

It's okay to have multiple protocols/sub-protocols/aliases if we need that now

I started with 1 because we *had* a common denominator, but we grow 😄

MarcoGorelli: >  had a common denominator

right, so if we add these comparison operators, we can remove `isin` and `between`?

dangotbanned: Keep `between` if possible

- https://docs.daft.ai/en/stable/api/expressions/#daft.expressions.Expression.between

Kind of like how we have `filter` for `NativeSeries` - it's something to avoid false-positives

Waaaaay more than just the types we're interested in will have comparison dunders, but *fewer* will also have `between` as well

narwhals/_compliant/expr.py
```diff
@@ -64,6 +64,17 @@ class NativeExpr(Protocol):
     def between(self, *args: Any, **kwds: Any) -> Any: ...
     def isin(self, *args: Any, **kwds: Any) -> Any: ...
```
MarcoGorelli:
```suggestion
```

PR #2922, opened 2025-08-02
test: (temporarely) xfail polars datetime selectors

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

---

I am losing track of other potential failures because of this issue

### DIFF
--- a/tests/selectors_test.py
+++ b/tests/selectors_test.py
@@ -89,7 +89,8 @@ def test_datetime(constructor: Constructor, request: pytest.FixtureRequest) -> N
         or ("pandas" in str(constructor) and PANDAS_VERSION < (2,))
         or "ibis" in str(constructor)
         # https://github.com/pola-rs/polars/issues/23767
-        or ("polars" in str(constructor) and POLARS_VERSION == (1, 32, 0, 1))
+        # TODO(FBruzzesi): Manage polars version
+        or ("polars" in str(constructor) and POLARS_VERSION >= (1, 32, 0))
     ):
         request.applymarker(pytest.mark.xfail)
     if "modin" in str(constructor):


narwhals/_utils.py
```diff
@@ -1281,11 +1279,14 @@ def is_ordered_categorical(series: Series[Any]) -> bool:
         result = False
     else:
         native = series.to_native()
-        if is_polars_series(native):
+        impl = series.implementation
+        if impl.is_polars() and impl._backend_version() < (1, 32):
+            # Since version 1.32.0, ordering parameter is ignored and
+            # it always behaves as if 'lexical' was passed.
             result = cast("pl.Categorical", native.dtype).ordering == "physical"
```
dangotbanned:
```suggestion
        if impl.is_polars():
            # NOTE: Deprecated https://github.com/pola-rs/polars/pull/23779
            # Since version 1.32.0, ordering parameter is ignored and
            # it always behaves as if 'lexical' was passed.
            return not cast("pl.Series", native).cat.uses_lexical_ordering()
```

dangotbanned: I had a look at the deprecation PR, I never knew about this `cat` method before 😄 

Turns out it isn't even remotely new https://github.com/pola-rs/polars/pull/10325

dangotbanned (approved):
cat-ptain on a narwhal

EdAbati:
Thank you @FBruzzesi ❤️ I was looking for this fix 

I was wondering if we could also update polars in the `typing` group in this PR:
https://github.com/narwhals-dev/narwhals/blob/bf5bedeeafded9be4e3954e480de0c449aee26b0/pyproject.toml#L75

Just so we get the latest polars when running `uv pip install ... --group local-dev`. `mypy` and `pyright` seem to stay happy.

narwhals/_utils.py
FBruzzesi: If the ordering will always be lexical, any chance that the flag will be removed?

dangotbanned: Sure, there's a chance anything will be removed 😉 

I'd expect that they'd also remove the `ordering` property at the same time though

Removing the flag from the rust side is a much more invasive change
- https://github.com/pola-rs/polars/blob/b98df35e81065f66f054ca22188320ad67c87221/py-polars/polars/series/categorical.py#L67
- https://github.com/pola-rs/polars/blob/b98df35e81065f66f054ca22188320ad67c87221/crates/polars-core/src/chunked_array/logical/categorical.rs#L56-L60


### Edit
Actually this would probably be more informative 🙂 
- https://github.com/pola-rs/polars/pull/23016

Issue #2847, opened 2025-07-18
Updates to `ruff` config


dangotbanned:
## Description
Just collecting some small things that we could change in our config to do less 🙂

### Less `# noqa: ...`
Mostly for conventions outside of our control like `F`, `Window`
- https://docs.astral.sh/ruff/settings/#lint_pep8-naming_extend-ignore-names
- https://docs.astral.sh/ruff/settings/#lint_flake8-builtins_allowed-modules
- https://docs.astral.sh/ruff/settings/#lint_flake8-builtins_ignorelist

### Disable
- https://docs.astral.sh/ruff/rules/snake-case-type-alias/
  - https://github.com/narwhals-dev/narwhals/blob/b9bdc34ced6eab2811efb29c06d82406faca0a52/narwhals/typing.py#L348-L349

### Enable autofix
- https://docs.astral.sh/ruff/rules/raw-string-in-exception/
  - Lets you write the error message in the exception, then fixes it by moving it to `msg`

## Related
- https://github.com/narwhals-dev/narwhals/pull/2845#issuecomment-3088971775
- #2199

FBruzzesi:
We can enable the following rules (without any change required):

-   "A005",    # stdlib-module-shadowing
-  "ISC001",  # single-line-implicit-string-concatenation
-  "PD003",   # pandas-use-of-dot-is-null

dangotbanned:
Thanks @FBruzzesi 

I hadn't checked the ignored one's this time, but they all LGTM!

(`A005`) is also configurable - if we ever need that in the future

- https://docs.astral.sh/ruff/rules/stdlib-module-shadowing/
- https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/
- https://docs.astral.sh/ruff/rules/pandas-use-of-dot-is-null/

EdAbati:
Hey both :)

One that we also enable is [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/#superfluous-else-return-ret505)
Since it reduces a bit of indentation, IMO it can really make the code cleaner

Putting a PR together for this (if you agree)

dangotbanned:
Thanks @EdAbati!

> One that we also enable is [RET505](https://docs.astral.sh/ruff/rules/superfluous-else-return/#superfluous-else-return-ret505)

I also like that rule, but IIRC it didn't interact well with `coverage` when I tried it out 😔

If you can get both working at the same time, then I'm all for it 😅

PR #2923, opened 2025-08-02
chore: enable more Ruff rules (`RET505`, `A005`, `ISC001`, `PD003`)

EdAbati:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2847

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_polars/dataframe.py
+++ b/narwhals/_polars/dataframe.py
@@ -15,7 +15,6 @@ from narwhals._polars.utils import (
 from narwhals._utils import (
     Implementation,
     _into_arrow_table,
-    check_columns_exist,
     convert_str_slice_to_int_slice,
     is_compliant_series,
     is_index_selector,
@@ -165,11 +164,6 @@ class PolarsBaseFrame(Generic[NativePolarsFrame]):
     def from_native(cls, data: NativePolarsFrame, /, *, context: _LimitedContext) -> Self:
         return cls(data, version=context._version)
 
-    def _check_columns_exist(self, subset: Sequence[str]) -> ColumnNotFoundError | None:
-        return check_columns_exist(  # pragma: no cover
-            subset, available=self.columns
-        )
-
     def simple_select(self, *column_names: str) -> Self:
         return self._with_native(self.native.select(*column_names))
 
@@ -262,9 +256,6 @@ class PolarsDataFrame(PolarsBaseFrame[pl.DataFrame]):
     # Can't do that here 😟
     write_csv: Method[Any]
     write_parquet: Method[None]
-
-    # CompliantDataFrame
-    _evaluate_aliases: Any
 
     @classmethod
     def from_arrow(cls, data: IntoArrowTable, /, *, context: _LimitedContext) -> Self:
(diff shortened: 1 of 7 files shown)


narwhals/_polars/dataframe.py
```diff
@@ -389,7 +389,7 @@ def __getitem__(  # noqa: C901, PLR0912
             selector = rows_native, columns_native
             selected = self.native.__getitem__(selector)  # type: ignore[index]
             return self._from_native_object(selected)
-        else:  # pragma: no cover
+        else:  # pragma: no cover # noqa: RET505
```
EdAbati, line 383: I am keeping this `else` here because it is quite a long piece of code and it is all not checked by coverage. Otherwise we'd have to add many `# pragma: no cover` :)

pyproject.toml
```diff
@@ -167,15 +166,12 @@ ignore = [
...
   "FIX",     # flake8-fixme
-  "ISC001",  # single-line-implicit-string-concatenation
-  "PD003",   # pandas-use-of-dot-is-null
   "PD010",   # pandas-use-of-dot-pivot-or-unstack
   "PD901",   # pandas-df-variable-name (This is a auxiliary library so dataframe variables have no concrete business meaning)
   "PLC0415", # `import` should be at the top-level of a file
   "PLR0913", # too-many-arguments
   "PLR2004", # magic-value-comparison
   "PLW1641", # Object does not implement `__hash__` method
-  "RET505",  # superfluous-else-return
```
EdAbati, line 178: this is the one responsible for all chenges above

dangotbanned:
Hey @EdAbati 

Sorry for not mentioning this before, but would you be okay with splitting `RET505` and everything else into 2 PRs?

I think the others are going to be a much smaller diff - so it would be easier to spot any issues on the bigger one 🙂

EdAbati:
Hey @dangotbanned  :) totally agree. The changes here are only caused by `RET505` though

`A005`, `ISC001`, `PD003` are also enabled here because they don't require or cause any line changes 👀 I updated the title to make it a bit clearer

dangotbanned:
> `A005`, `ISC001`, `PD003` are also enabled here because they don't require or cause any line changes 👀 I updated the title to make it a bit clearer

aha thanks, my bad 🤦‍♂️

MarcoGorelli:
nice, on board with the idea if we can get it green, and agree with the rationelle behind https://github.com/narwhals-dev/narwhals/pull/2923#discussion_r2249229310

narwhals/_compliant/series.py
```diff
@@ -354,7 +354,7 @@ def __getitem__(self, item: MultiIndexSelector[Self]) -> Self:
             return self._gather_slice(item)
         if is_compliant_series(item):
             return self._gather(item.native)
-        if is_sized_multi_index_selector(item):
+        elif is_sized_multi_index_selector(item):  # noqa: RET505
```
EdAbati, line 357: Interstingly, `mypy` seems to disregard the `if is_compliant_series(item):` above when we don't put an `elif` here.
Maybe it is a bug in `mypy`? 
Can you think of a more elegant way to fix this?
(we could also put a `#type: ignore` in the line below)

dangotbanned: > (we could also put a #type: ignore in the line below)

100% stick with the `elif` and `#noqa` vs type ignore

`mypy` struggles with a lot of things, but a `type: ignore` will disable `pyright` from checking anything 😰

dangotbanned:
Alright I'm at 22/39 files and looking great so far - will finish the rest later

dangotbanned (approved):
Thanks for this @EdAbati!


hungry narwhal

PR #2942, opened 2025-08-05
feat: add `list.contains`

### LINKED ISSUE #2936: enh: expand `list` namespace
A couple of methods I'd like to have (as I'm using them in a project, which I'm curious to see if it can be narwhalified) are:
- `list.contains`
- `list.unique`

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2936 
- `pyarrow` support (https://github.com/apache/arrow/issues/33295)

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_polars/expr.py
+++ b/narwhals/_polars/expr.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
     from narwhals._polars.dataframe import Method, PolarsDataFrame
     from narwhals._polars.namespace import PolarsNamespace
     from narwhals._utils import Version, _LimitedContext
-    from narwhals.typing import IntoDType
+    from narwhals.typing import IntoDType, NonNestedLiteral
 
 
 class PolarsExpr:
@@ -418,7 +418,7 @@ class PolarsExprListNamespace(
     def unique(self) -> PolarsExpr:
         return self.compliant._with_native(self.native.list.unique())
 
-    def contains(self, item: Any) -> PolarsExpr:
+    def contains(self, item: NonNestedLiteral) -> PolarsExpr:
         return self.compliant._with_native(self.native.list.contains(item))
 
 
--- a/narwhals/_polars/expr.py
+++ b/narwhals/_polars/expr.py
@@ -418,6 +418,9 @@ class PolarsExprListNamespace(
     def unique(self) -> PolarsExpr:
         return self.compliant._with_native(self.native.list.unique())
 
+    def contains(self, item: Any) -> PolarsExpr:
+        return self.compliant._with_native(self.native.list.contains(item))
+
 
 class PolarsExprStructNamespace(
     PolarsExprNamespace, PolarsStructNamespace[PolarsExpr, pl.Expr]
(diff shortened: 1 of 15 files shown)


narwhals/_polars/series.py
```diff
@@ -742,6 +748,11 @@ def unique(self) -> PolarsSeries:
         ns = self.__narwhals_namespace__()
         return self.to_frame().select(ns.col(name).list.unique()).get_column(name)
 
+    def contains(self, item: NonNestedLiteral) -> PolarsSeries:
+        name = self.name
+        ns = self.__narwhals_namespace__()
+        return self.to_frame().select(ns.col(name).list.contains(item)).get_column(name)
+
```
dangotbanned, line 750: 😍

narwhals/_polars/expr.py
```diff
@@ -418,6 +418,9 @@ def len(self) -> PolarsExpr:
     def unique(self) -> PolarsExpr:
         return self.compliant._with_native(self.native.list.unique())
 
+    def contains(self, item: NonNestedLiteral) -> PolarsExpr:
+        return self.compliant._with_native(self.native.list.contains(item))
```
MarcoGorelli: Polars 1.28 is really recent

I wonder if we could do something like
```python
if self._backend_version < (1,28):
    result = pl.when(self.native, self.native.list.contains(item))
else:
    result = self.native.list.contains(item)
return self.compliant._with_native(result)
```
?

raisadz: Thanks @MarcoGorelli , I changed this

narwhals/_polars/series.py
dangotbanned: @raisadz I should write this down somewhere (besides https://github.com/narwhals-dev/narwhals/pull/2897#discussion_r2249358485)
currently we're only doing this `Expr` dispatch stuff when we need to do version branching.

There's probably not much performance benefit (e.g #2946), *unless* we need to use intermediate `Series`(s)

*For this PR though*, perfect fit 🥳 

https://github.com/raisadz/narwhals/blob/87b17fcad82d60497146b4e04bcd2a6f84458609/narwhals/_polars/expr.py#L421-L428

MarcoGorelli:
it might be possible to simplify as per https://github.com/narwhals-dev/narwhals/pull/2946

narwhals/_polars/series.py
raisadz: Thanks for explaining about the version branching @dangotbanned ! I was copying this from `list.len()` and it's good to know

raisadz:
> it might be possible to simplify as per #2946

I tried to have a similar simplification @MarcoGorelli but it doesn't seem to work in this case

MarcoGorelli (approved):
thanks all

at some point we may want to consider making a `SQLExprListNamespace`, like we have for `StringExprNamespace`, to share implementations between duckdb/pyspark/ibis where possible, but for now this is fine

choucavalier:
When you need a feature, you see the PR being opened yesterday and Marco approving the PR 1h ago. Things are moving fast!

MarcoGorelli:
😄 thanks @choucavalier 

Any other features you've found missing that would be useful to you?

PR #2356, opened 2025-04-07
fix(typing): Narrow `TypeVar`(s) used in `(Data|Lazy)Frame`

dangotbanned:
Will close #2344, #2345

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #2344
- Closes #2345
- Related #2179

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below
- [x] Update `main`
  - [x] `TypeAlias`s and `@overload`s
  - [x] `DataFrame._compliant_frame`
  - [x] `LazyFrame._compliant_frame`
- [x] Update `v1`
  - [x] `LazyFrame` overload
  - [x] `DataFrame` overload
    - Really struggling (https://github.com/narwhals-dev/narwhals/pull/2356/commits/1725f36c26541b3f3147f186a070a566f308a2c9)
    - Gave up (https://github.com/narwhals-dev/narwhals/pull/2356/commits/c4bed590041026163b1968cff0f29383221243f5)

PR #2159, opened 2025-03-06
fix(typing): Narrow `NativeSeries` Protocol

dangotbanned:
Closes #2111

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue https://github.com/narwhals-dev/narwhals/pull/2110#discussion_r1973726417
- Closes #2111

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below
Of the possible common methods, this avoid `np.ndarray` matching - as well as builtins.

### Common methods
- `filter`
- `unique`
- `value_counts`
- `equals`
- `take` (numpy)
- `to_numpy`
- `__array__` (numpy)
- `__len__` (numpy)
- `__iter__` (numpy)

Issue #2932, opened 2025-08-03
Remove `level`, `DataFrameLike` from `main`, `v2


dangotbanned:
## Related 
- https://github.com/narwhals-dev/narwhals/pull/2814#discussion_r2234109699

> @dangotbanned 
> We can move `level` to `v1`-only can't we?
> 
> Or at least `"interchange"` I thought was on the chopping block?

> @MarcoGorelli 
> yes true, i just think it'd require a larger refactor
> 
> it's not user-facing though, so not necessarily pressing?

> @dangotbanned 
> Yeah that seems fair
> 
> How about in this PR, try removing [`v2.typing.DataFrameLike`](https://github.com/MarcoGorelli/narwhals/blob/eb60fe72d621bf63dee6c151b772a31598e77bc6/narwhals/stable/v2/typing.py#L29-L32) and adding some tests using a mock w/ `__dataframe__`?
> 
> I think making sure the typing on `v2` rejects interchange only _should_ be doable 🙂
> We can punt the rest to another issue

## Description
So now we need to figure out the rest 😄

PR #2944, opened 2025-08-05
fix(typing): Avoid overlapping `DataFrame`, `LazyFrame`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- 3rd in a trilogy that started with #2159
- and last left off at #2356

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Everything here (and it is a biggy) stems from this small addition:

```py
class NativeDataFrame(Sized, NativeFrame, Protocol): ...
```

The presence of `__len__` is enough to avoid overlapping `@overload`s 🤯


dangotbanned:
These `altair` failures are a sign that things are working now 🙂 

- https://github.com/narwhals-dev/narwhals/actions/runs/16760676811/job/47454833323?pr=2944

narwhals/_compliant/dataframe.py
```diff
@@ -357,7 +360,7 @@ class EagerDataFrame(
     CompliantDataFrame[
         EagerSeriesT, EagerExprT, NativeDataFrameT, "DataFrame[NativeDataFrameT]"
     ],
-    CompliantLazyFrame[EagerExprT, NativeDataFrameT, "DataFrame[NativeDataFrameT]"],
+    CompliantLazyFrame[EagerExprT, "Incomplete", "DataFrame[NativeDataFrameT]"],
```
dangotbanned, line 364: Pretty sure this `Incomplete` is unavoidable, just like the `[type-var]` warnings on `v1.DataFrameLike`

https://github.com/narwhals-dev/narwhals/blob/89095ceb6563e818e8f9c021bc76e730870145e0/narwhals/stable/v1/__init__.py#L96-L98

I'm much happier to eat those two, when it fixes sooooooooo much else 🥳

tests/frame/join_test.py
dangotbanned, line 1: This entire test module is an issue.

We can **either**:

1. Use a `type: ignore[arg-type]` on every join
2. Or, type everything as **one of** DataFrame/LazyFrame
3. Or, split the tests into eager/lazy and type them accurately

The warnings are legit, you can't join `DataFrame` with `LazyFrame` - but we've annotated `Constructor` as both 🤔

narwhals/stable/v1/typing.py
```diff
@@ -40,7 +46,8 @@ def __dataframe__(self, *args: Any, **kwargs: Any) -> Any: ...
 `nw.Expr`, e.g. `df.select('a')`.
 """
 
-IntoDataFrame: TypeAlias = Union["NativeFrame", "DataFrame[Any]", "DataFrameLike"]
+
+IntoDataFrame: TypeAlias = Union["NativeDataFrame", "DataFrameLike"]
```
dangotbanned, line 50: I guess I was forced to get started on #2932 😅

dangotbanned:
@MarcoGorelli I just wanna loop you into this one early, since it may be a challenge to keep conflict-free 🙂 

~~There is~~ There was some more tidying up to do, but wanted to check if the general idea seems a net-positive to you?

Mainly in terms of where we no longer need/now require type ignore(s)

MarcoGorelli:
i think this is fine, yes, thanks!

tests/expr_and_series/dt/datetime_attributes_test.py
```diff
@@ -119,7 +123,9 @@ def test_to_date(request: pytest.FixtureRequest, constructor: Constructor) -> No
         request.applymarker(pytest.mark.xfail)
     dates = {"a": [datetime(2001, 1, 1), None, datetime(2001, 1, 3)]}
     if "dask" in str(constructor):
-        df = nw.from_native(constructor(dates).astype({"a": "timestamp[ns][pyarrow]"}))  # type: ignore[union-attr]
+        df_dask = cast("dd.DataFrame", constructor(dates))
+        df_dask = cast("dd.DataFrame", df_dask.astype({"a": "timestamp[ns][pyarrow]"}))
+        df = nw.from_native(df_dask)
```
dangotbanned, line 128: I'm not going to do this *everywhere*, but this fix turned out to be a great test for matching `dd.DataFrame` to `NativeLazyFrame`

image

tests/frame/join_test.py
FBruzzesi, line 1: What's happening here? This seems now worse than before?

tests/frame/join_test.py
```diff
@@ -96,7 +96,11 @@ def test_full_join(
...
     df_right = nw.from_native(constructor(df2))
     result = df_left.join(
-        df_right, on=on, left_on=left_on, right_on=right_on, how="full"
+        df_right,  # type: ignore[arg-type]
+        on=on,
+        left_on=left_on,
+        right_on=right_on,
+        how="full",
     ).sort("id", nulls_last=True)
     assert_equal_data(result, expected)
```
dangotbanned: @FBruzzesi 
> What's happening here? This seems now worse than before?

All of these are the same problem, I gave some options we can look at for solving in (https://github.com/narwhals-dev/narwhals/pull/2944#discussion_r2255379424)

But I'll walk through the existing issue step-by-step 🙂 

## Minimal repro
### `main`
On `main`, this **does not** produce any warnings:

```py

import narwhals as nw

def test_full_join(constructor: Constructor) -> None:
    data = {"id": [1, 2, 3], "name": ["Alice", "Bob", "Charlie"], "age": [25, 30, 35]}
    native_1 = constructor(data)
    native_2 = constructor(data)
    frame_1 = nw.from_native(native_1)
    frame_2 = nw.from_native(native_2)
    result = frame_1.join(frame_2, on="id")
    assert result
```

Is that a good thing?

The definition of `Constructor` on `main` is:

```py
Constructor: TypeAlias = Callable[[Any], NativeLazyFrame | NativeFrame | DataFrameLike]
```

So now let's see what the types are after each call:


```py

from typing_extensions import reveal_type


import narwhals as nw

def test_full_join(constructor: Constructor) -> None:
    data = {"id": [1, 2, 3], "name": ["Alice", "Bob", "Charlie"], "age": [25, 30, 35]}
    native_1 = constructor(data)
    reveal_type(native_1)  # Type of "native_1" is "NativeLazyFrame | NativeFrame | DataFrameLike"
    
    native_2 = constructor(data)
    reveal_type(native_2)  # Type of "native_2" is "NativeLazyFrame | NativeFrame | DataFrameLike"
    
    frame_1 = nw.from_native(native_1)  
    reveal_type(frame_1)  # Type of "frame_1" is "DataFrame[NativeLazyFrame | NativeFrame | DataFrameLike]"
    
    frame_2 = nw.from_native(native_2)
    reveal_type(frame_2)  # Type of "frame_2" is "DataFrame[NativeLazyFrame | NativeFrame | DataFrameLike]"
    
    result = frame_1.join(frame_2, on="id")
    reveal_type(result)
```

- Do all of those look correct?
- Where is `LazyFrame`?
- How did `NativeLazyFrame` end up inside `DataFrame`? 

### (#2944)
The definition of `Constructor` has been updated:
```py
Constructor: TypeAlias = Callable[[Any], NativeLazyFrame | NativeDataFrame]
```

but that doesn't explain everything!

Let's try that example previous example again:

```py

from typing_extensions import reveal_type


import narwhals as nw

def test_full_join(constructor: Constructor) -> None:
    data = {"id": [1, 2, 3], "name": ["Alice", "Bob", "Charlie"], "age": [25, 30, 35]}
    native_1 = constructor(data)
    reveal_type(native_1)  # Type of "native_1" is "NativeLazyFrame | NativeDataFrame"

    native_2 = constructor(data)
    reveal_type(native_2)  # Type of "native_2" is "NativeLazyFrame | NativeDataFrame"

    frame_1 = nw.from_native(native_1)
    reveal_type(
        frame_1
    )  # Type of "frame_1" is "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]"

    frame_2 = nw.from_native(native_2)
    reveal_type(
        frame_2
    )  # Type of "frame_2" is "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]"

    result = frame_1.join(frame_2, on="id")
    # Argument of type "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]" cannot be assigned to parameter "other" of type "DataFrame[NativeDataFrame]" in function "join"
    #  Type "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]" is not assignable to type "DataFrame[NativeDataFrame]"
    #    "LazyFrame[NativeLazyFrame]" is not assignable to "DataFrame[NativeDataFrame]"
    # Argument of type "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]" cannot be assigned to parameter "other" of type "LazyFrame[NativeLazyFrame]" in function "join"
    #  Type "LazyFrame[NativeLazyFrame] | DataFrame[NativeDataFrame]" is not assignable to type "LazyFrame[NativeLazyFrame]"
    #    "DataFrame[NativeDataFrame]" is not assignable to "LazyFrame[NativeLazyFrame]"
    reveal_type(
        result
    )  # Type of "result" is "DataFrame[NativeDataFrame] | LazyFrame[NativeLazyFrame]"
```

😳 😳 😳

> [!IMPORTANT]
> We cannot join `LazyFrame` and `DataFrame`
> Our typing now understands that `nw.from_native(constructor(data))` can produce `LazyFrame | DataFrame`


Does that help explain things better?

FBruzzesi: Thanks for the walkthrough @dangotbanned it's very much appreciated 🙏🏼

I have some comments/questions/remarks:

1. If we were in the eager case `nw.from_native(..., eager_only=True)` then I hope/assume that there would be no issues. Is that the case?
2. In the above case, if we were to do: `nw.from_native(...).lazy()` everything should just play out then? If that's the case, as I imagine, maybe that's preferable to adding all those `type: ignore`?
3. I guess this goes back to the beginning of the project, but I don't love that we have a way to flag something as eager, but for a lazy input we get `DataFrame | LazyFrame`. This is probably more of a comment to support the `{DataFrame, LazyFrame, Series}.from_native(...)` to get better tuning on typing among other advantages

dangotbanned: Thanks @FBruzzesi, good questions!

### **1**
> If we were in the eager case `nw.from_native(..., eager_only=True)` then I hope/assume that there would be no issues. Is that the case?

Yep no issues 😄 

Every test (typing) change has been stemming from at least one of:
- `Constructor`
- `IntoFrame`
- Performing native operations without the typing for it
  - I found this group quite interesting as it helps to verify the new overloads + typing
  - https://github.com/narwhals-dev/narwhals/pull/2944/files#diff-ea0a1668a35ed6b0359b1359efe2047081b041c6386d7b1d2f7c576867274533
  - https://github.com/narwhals-dev/narwhals/pull/2944/files#r2282246791


### **2**
> if we were to do: `nw.from_native(...).lazy()` everything should just play out then?

Yeah that works and you can see in this later test where that is enough to avoid the (typing) issues

https://github.com/narwhals-dev/narwhals/blob/482eca8ca59323600ac85798c9d95f6aa580fa6c/tests/frame/join_test.py#L108-L117

https://github.com/narwhals-dev/narwhals/blob/482eca8ca59323600ac85798c9d95f6aa580fa6c/tests/frame/join_test.py#L128-L129

We can do this because `DataFrame | LazyFrame` is narrowed to `LazyFrame`

```py
    def lazy(self, **kwargs: Any) -> LazyFrame: ...
```

and `LazyFrame` has enough of the API that we need for those tests


### **3**
> I guess this goes back to the beginning of the project, but I don't love that we have a way to flag something as eager, but for a lazy input we get `DataFrame | LazyFrame`

In (https://github.com/narwhals-dev/narwhals/pull/2944/files#diff-4da9e38e2e0c05c4aa26936a95b843bcc1fdd326411fb3fbe563881ab0d51516), you can see lazy input doesn't result in `DataFrame | LazyFrame`.

Previously, **eager input** was resulting in `nw_v1.DataFrame[pl.DataFrame] | nw_v1.LazyFrame[pl.DataFrame]`.
But the type assertion had to be updated, since this PR fixed that issue 🙂

dangotbanned: These seem to have done the trick 

- [test: Call `.lazy()` before every join test](https://github.com/narwhals-dev/narwhals/pull/2944/commits/dbe443222edb25059d7d06077c53b2ff616c44d0)
- [test: Move into a documented helper](https://github.com/narwhals-dev/narwhals/pull/2944/commits/3c7e1cc6cf3b299b62b78ecb63e86fbe3a9e9d1d)
  - If we come up with a different solution in the future, at least the surface area is waaaaaaay smaller for the change 😄

dangotbanned: Well - it ~~has~~ had the downside of this *looking* uncovered


https://github.com/narwhals-dev/narwhals/blob/dcd67f5bff6a79b894d38d6510d4f2736248c76a/narwhals/dataframe.py#L1867-L1887

And this *actually* not being covered

https://github.com/narwhals-dev/narwhals/blob/dcd67f5bff6a79b894d38d6510d4f2736248c76a/narwhals/_polars/dataframe.py#L547-L548

PR #2945, opened 2025-08-06
feat(typing): Add `IntoSchema` alias

dangotbanned:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Part of https://github.com/narwhals-dev/narwhals/issues/2116#issuecomment-3122104794

## Checklist

- [ ] Tests added
- [x] Documented the changes

I was going to do this *after* `IntoBackend`, but it turned out to be a much simpler change

### DIFF
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -6,7 +6,7 @@ from narwhals._compliant import CompliantDataFrame, CompliantLazyFrame, Complian
 
 if TYPE_CHECKING:
     import datetime as dt
-    from collections.abc import Iterable, Sequence, Sized
+    from collections.abc import Iterable, Mapping, Sequence, Sized
     from decimal import Decimal
     from types import ModuleType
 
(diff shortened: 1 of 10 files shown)


narwhals/typing.py
```diff
@@ -390,6 +391,36 @@ def Binary(self) -> type[dtypes.Binary]: ...
     └──────────────────┘
 """
 
+# TODO @dangotbanned: fix this?
+# Constructor allows tuples, but we don't support that *everywhere* yet
+IntoSchema: TypeAlias = "Mapping[str, dtypes.DType] | Schema"
+"""Anything that can be converted into a Narwhals Schema.
```
dangotbanned, line 397: Thought I'd float this here

The typing for `from_{dict,numpy}` has never included `Iterable[tuple[str, DType]]` - but that is accepted by `Schema`.

*If* we wanted to support that, it wouldn't be too much work (especially in `from_dict`) but would need some runtime changes

E.g.

https://github.com/narwhals-dev/narwhals/blob/31512beccc77d61cf0ff246b2bdd8ab545092dd5/narwhals/_arrow/dataframe.py#L148-L153

MarcoGorelli (approved):
thanks @dangotbanned

Issue #2905, opened 2025-07-28
[Enh]: Add `list.get()` negative indices and out-of-bounds access


skritsotalakis:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

Looking at modern Polars Chapter 2.2 (https://kevinheavey.github.io/modern-polars/method_chaining.html#:~:text=extract_city_name_pl) for #1945, I noticed that narwhals is missing a method similar to `polars expr.list.get()` (https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.list.get.html).

### Please describe the purpose of the new feature or describe the problem to solve.

This feature allows users to get values by index in a sublist (`expr.list`).

### Additional information that may help us understand your needs.

Example with data from modern Polars:

```python 

import polars

data = [
    {"OriginCityName": "Raleigh/Durham, NC", "DestCityName": "Chicago, IL"},
    {"OriginCityName": "Seattle, WA", "DestCityName": "Chicago, IL"},
    {"OriginCityName": "Indianapolis, IN", "DestCityName": "Dallas/Fort Worth, TX"},
    {"OriginCityName": "Sarasota/Bradenton, FL", "DestCityName": "Newark, NJ"},
    {"OriginCityName": "Chicago, IL", "DestCityName": "Moline, IL"},
    {"OriginCityName": "Madison, WI", "DestCityName": "Washington, DC"},
    {"OriginCityName": "West Palm Beach/Palm Beach, FL", "DestCityName": "Charlotte, NC"}
]

df_pl = pl.DataFrame(data)
```

```python
In [26]: df_pl
Out[26]: 
shape: (7, 2)
┌────────────────────────────────┬───────────────────────┐
│ OriginCityName                 ┆ DestCityName          │
│ ---                            ┆ ---                   │
│ str                            ┆ str                   │
╞════════════════════════════════╪═══════════════════════╡
│ Raleigh/Durham, NC             ┆ Chicago, IL           │
│ Seattle, WA                    ┆ Chicago, IL           │
│ Indianapolis, IN               ┆ Dallas/Fort Worth, TX │
│ Sarasota/Bradenton, FL         ┆ Newark, NJ            │
│ Chicago, IL                    ┆ Moline, IL            │
│ Madison, WI                    ┆ Washington, DC        │
│ West Palm Beach/Palm Beach, FL ┆ Charlotte, NC         │
[...]

PR #2949, opened 2025-08-06
Feat: add `list.get(index: int)`

skritsotalakis:

- [ ] 💾 Refactor
- [X] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2905

## Checklist

- [X] Tests added
- [X] Documented the changes

### DIFF
--- a/tests/frame/explode_test.py
+++ b/tests/frame/explode_test.py
@@ -46,31 +46,31 @@ def test_explode_single_col(
         .with_columns(nw.col(column).cast(nw.List(nw.Int32())))
         .explode(column)
         .select("a", column)
-        .sort("a")
+        .sort("a", column, nulls_last=True)
     )
     expected = {"a": ["w", "x", "x", "y", "z"], column: expected_values}
     assert_equal_data(result, expected)
 
 
 @pytest.mark.parametrize(
-    ("columns", "more_columns", "expected"),
+    ("column", "more_columns", "expected"),
     [
         (
             "l1",
             ["l2"],
             {
-                "a": ["x", "x", "y", "z", "w"],
-                "l1": [1, 2, None, None, None],
-                "l2": [3, None, None, 42, None],
+                "a": ["w", "x", "x", "y", "z"],
+                "l1": [None, 1, 2, None, None],
+                "l2": [None, 3, None, None, 42],
             },
         ),
         (
             "l3",
             ["l4"],
             {
-                "a": ["x", "x", "y", "z", "w"],
-                "l3": [1, 2, 3, None, 1],
-                "l4": [1, 2, 3, 123, 456],
+                "a": ["w", "x", "x", "y", "z"],
+                "l3": [1, 1, 2, 3, None],
+                "l4": [456, 1, 2, 3, 123],
             },
         ),
     ],
@@ -78,7 +78,7 @@ def test_explode_single_col(
 def test_explode_multiple_cols(
     request: pytest.FixtureRequest,
     constructor: Constructor,
-    columns: str | Sequence[str],
+    column: str,
     more_columns: Sequence[str],
     expected: dict[str, list[str | int | None]],
 ) -> None:
@@ -93,9 +93,10 @@ def test_explode_multiple_cols(
 
     result = (
         nw.from_native(constructor(data))
-        .with_columns(nw.col(columns, *more_columns).cast(nw.List(nw.Int32())))
-        .explode(columns, *more_columns)
-        .select("a", columns, *more_columns)
+        .with_columns(nw.col(column, *more_columns).cast(nw.List(nw.Int32())))
+        .explode(column, *more_columns)
+        .select("a", column, *more_columns)
+        .sort("a", column, nulls_last=True)
     )
     assert_equal_data(result, expected)
 


docs/api-reference/expr_list.md
```diff
@@ -6,5 +6,4 @@
       members:
         - len
         - unique
-      show_source: false
-      show_bases: false
+        - get
```
FBruzzesi: Can we keep the 

```md
      show_source: false
      show_bases: false
```

block and sort the members alphabetically?

docs/api-reference/series_list.md
FBruzzesi, line 1: Same as per `expr_list.md`

narwhals/_compliant/any_namespace.py
```diff
@@ -53,8 +53,11 @@ def offset_by(self, by: str) -> CompliantT_co: ...
 
 class ListNamespace(_StoresCompliant[CompliantT_co], Protocol[CompliantT_co]):
     def len(self) -> CompliantT_co: ...
+
     def unique(self) -> CompliantT_co: ...
 
+    def get(self, index: int) -> CompliantT_co: ...
+
```
FBruzzesi: Nitpick to keep it alphabetically sorted here as well

```suggestion
    def get(self, index: int) -> CompliantT_co: ...
    def len(self) -> CompliantT_co: ...
    def unique(self) -> CompliantT_co: ...
```

narwhals/_polars/expr.py
FBruzzesi, line 1: For polars these can be avoided. They are handled automatically by the parent class `PolarsAnyNamespace` if there is no modification to the native behavior (e.g. backend version cases)

narwhals/_polars/series.py
FBruzzesi, line 1: Same as per `_polars/expr.py` comment

narwhals/_pandas_like/series_list.py
```diff
@@ -33,3 +33,22 @@ def len(self) -> PandasLikeSeries:
...
+        ca = native.array._pa_array
+        result_arr = pc.list_element(ca, index)
+        nw_dtype = native_to_narwhals_dtype(result_arr.type, self.version)
+        out_dtype = narwhals_to_native_dtype(
+            nw_dtype, "pyarrow", self.implementation, self.version
+        )
+        result_native = native_cls(
+            result_arr, dtype=out_dtype, index=native.index, name=native.name
+        )
+        return self.with_native(result_native)
```
FBruzzesi: AFAIK, pandas already exposes [`.list.__getitem__`](https://pandas.pydata.org/docs/reference/api/pandas.Series.list.__getitem__.html).

Yet the series name is not preserved.

Something along the following lines _might_ work:

```py
result = self.native.list[index]
result.name = self.native.name
return self.with_native(result)
```

tests/expr_and_series/list/get_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+import re
+
+import pytest
+
+import narwhals as nw
+from tests.utils import PANDAS_VERSION, Constructor, ConstructorEager, assert_equal_data
+
+data = {"int": [[1, 2], [None, 3], [None]]}
+expected = {"int": [1, None, None]}
```
FBruzzesi: What happens (and should happen) for a null element? not a list with a null
Can we add a test for that?

tests/expr_and_series/list/get_test.py
FBruzzesi, line 1: Can we add a couple of cases for:

- index out of range
- negative indexes

I don't know what should happen and which backend support negative indexes. If that's only polars, we might want to check that in the input and raise an issue (e.g. `ValueError` or `NotImplementedError`) already at the narwhals level (both for series and expr), and add it in the docstrings

narwhals/_polars/expr.py
skritsotalakis: is the simplification possible despite that here `index` only gets `int` whereas polars' `get`  `index` can also get expr, str, etc...? (https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.list.get.html)

docs/api-reference/expr_list.md
skritsotalakis: oh! that got removed accidentally during cherry picking conflicts. will update it!

tests/expr_and_series/list/get_test.py
skritsotalakis: yes!

narwhals/_polars/expr.py
FBruzzesi: Yes we can still remove it. I would argue to add a check for the data type at the narwhals level, and raise an error for non integer values. In this way by the time the polars function is called, it's guaranteed to be called with an int. WDYT?

skritsotalakis: Sounds good. I will do that and will look into the possibility of extending the accepted index types in a future PR

MarcoGorelli: Agree, let's start with just accepting integers, we can always loosen and expressify in the future

tests/expr_and_series/list/get_test.py
skritsotalakis: This is the only point i still need to address (locally) but I am not sure how.

a) Index out of range:

Different backends have different error messages for indices out of range. not sure how to test for this.

b) Negative indices:

Negative indices are only supported by `pandas series` , `polars`, and `ibis`. Anything `pyarrow` backed fails as out of bounds. `Sql_frame` and `duckdb` return `None` when negative indices are provided, and `pyspark` has its own error.
<details>
  <summary>Click to view screenshot</summary>

  Screenshot from 2025-08-13 18-32-20

</details>


Do you think raising is still a good choice? Are there other alternative?

FBruzzesi: Hey @skritsotalakis 👋🏼

Thanks for addressing these points.

a) Index out of range: yes I would expect that to be the case. Ideally we would like to align the exception raised.

An example of handling exceptions for pyspark is the following:

https://github.com/narwhals-dev/narwhals/blob/ad8f146c9f8cf28a8c1197cadaf2282aff810fb1/narwhals/_spark_like/dataframe.py#L256-L259

assuming that out of bound is the only exception that can happen in such function.
I could not find any such example for `Expr`, so actually I am not sure. @EdAbati did you work on something similar in the past?

b) I think this might be much easier

1. For now we can raise a `ValueError` exception at the narwhals level (so that it's the same for all backends)
2. As a follow up, for those backend that do not support it natively, we can do something along the lines of:

    ```py
    if index < 0:
         return expr.list.get(expr.list.len() + index)
    ```

MarcoGorelli: thanks both

tbh i don't think it's blocking that a unified exception be raised for out-of-bounds access, i think that's ok as a follow-up, just make an issue for it so we don't forget about it

skritsotalakis:
Hey @FBruzzesi ! 

I updated the PR to address the feedback. Let me know if it looks good or not.

Cheers :) 

* p.s. not sure if the 3.11 pytest failure is related to the PR

skritsotalakis:
Thanks for all the support throughout the process!

PR #2113, opened 2025-02-28
enh: Adds support for `Time` datatype

marvinl803:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [X] 🐳 Other

## Related issues

- Related issue #1989

## Checklist

- [X] Code follows style guide (ruff)
- [X] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
I have added support for the Polars Time datatype. I wanted to implement this first to ensure there are no issues before proceeding with the Binary datatype, as mentioned in the issue.

Please let me know if any changes or fixes are needed. Thank you!

Issue #2691, opened 2025-06-17
`modin` (non-pyarrow) backend isn't run in CI


dangotbanned:
AFAIK, we are only testing `modin[pyarrow]` and not `modin` :flushed: 

https://github.com/narwhals-dev/narwhals/blob/d6b2b6ec11128e17fdaf713e22844f2db5d00d1f/.github/workflows/pytest.yml#L60

https://github.com/narwhals-dev/narwhals/blob/d6b2b6ec11128e17fdaf713e22844f2db5d00d1f/.github/workflows/pytest.yml#L91

---

Discovered quite a few failures just now, as I've installed `modin` to track down the issue in (https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2151945563).

Ignoring my `group_by`-related issues, `pytest --constructors="modin"` gives me:

```py
FAILED tests/expr_and_series/any_horizontal_test.py::test_anyh_kleene[modin] - AssertionError: Mismatch at index 2: False != None
FAILED tests/expr_and_series/str/split_test.py::test_str_split[modin-_-expected0] - TypeError: This operation requires a pyarrow-backed series. Please refer to https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_convert_dtypes and ensure you are using dtype_backend='pyarrow'. Additionally, make sure you hav...
FAILED tests/expr_and_series/str/split_test.py::test_str_split[modin-,-expected1] - TypeError: This operation requires a pyarrow-backed series. Please refer to https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_convert_dtypes and ensure you are using dtype_backend='pyarrow'. Additionally, make sure you hav...
FAILED tests/expr_and_series/str/split_test.py::test_str_split_series[modin-_-expected0] - TypeError: This operation requires a pyarrow-backed series. Please refer to https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_convert_dtypes and ensure you are using dtype_backend='pyarrow'. Additionally, make sure you hav...
FAILED tests/expr_and_series/str/split_test.py::test_str_split_series[modin-,-expected1] - TypeError: This operation requires a pyarrow-backed series. Please refer to https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.maybe_convert_dtypes and ensure you are using dtype_backend='pyarrow'. Additionally, make sure you hav...
[...]

PR #2950, opened 2025-08-06
fix: Allow `modin[pyarrow]` to use more dtypes

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Discovered in https://github.com/narwhals-dev/narwhals/commit/0c5b7d5616a6efc6dcfe34143b54181c7a56547d
- `pandas`-only has been enforced (by us) since #2113

## Checklist

- [x] Tests added
- [ ] Documented the changes

I was writing a schema conversion test and was confused by the output.

`modin` just re-exports `pd.ArrowDtype` *as-is*, so maybe this was always working? 🤔

- https://github.com/modin-project/modin/blob/e3724b67c60eee972451ca10fee1c7496f7f3861/modin/pandas/__init__.py#L99
- https://modin.readthedocs.io/en/stable/supported_apis/utilities_supported.html#other-objects-structures

### DIFF
--- a/tests/expr_and_series/cast_test.py
+++ b/tests/expr_and_series/cast_test.py
@@ -270,9 +270,7 @@ def test_cast_datetime_utc(
 
 
 def test_cast_struct(request: pytest.FixtureRequest, constructor: Constructor) -> None:
-    if any(
-        backend in str(constructor) for backend in ("dask", "modin", "cudf", "sqlframe")
-    ):
+    if any(backend in str(constructor) for backend in ("dask", "cudf", "sqlframe")):
         request.applymarker(pytest.mark.xfail)
 
     if "pandas" in str(constructor) and PANDAS_VERSION < (2, 2):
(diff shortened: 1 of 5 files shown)


tests/expr_and_series/cast_test.py
```diff
@@ -328,9 +326,7 @@ def test_cast_time(request: pytest.FixtureRequest, constructor: Constructor) ->
     if "pandas" in str(constructor) and PANDAS_VERSION < (2, 2):
         pytest.skip()
 
-    if any(
-        backend in str(constructor) for backend in ("dask", "pyspark", "modin", "cudf")
```
dangotbanned, line 332: If anyone is testing `"modin"` (not `"modin[pyarrow]"`) *locally*, then that one will now start appearing in like #2691

I'm pretty sure no one is doing that.

Expanding the utils from #2879 will be a better option going forward for consistency anyway 🙂

dangotbanned:
Thanks @FBruzzesi!

> we might want to have a better way to identify constructors. Let's keep that in mind with the whole testing module conversation.

Onto the list it goes 😄

PR #2953, opened 2025-08-06
fix: Support passing `nw.Object` to pandas-like

dangotbanned:
Closes #2851

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2851

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- `polars` always supported it
- `pyarrow` never will

Issue #2851, opened 2025-07-18
[Bug]: `nw.Object` not a valid data type


ofajardo:
### Describe the bug

An error arises when trying to build a narwhals dataframe using the from_dict method if nw.Object is included in the schema

### Steps or code to reproduce the bug

```
import narwhals as nw
df_csv = nw.from_dict({"Var1": [3, "a"], "Var2":["a", "b"]}, backend="pandas", schema={"Var1":nw.Object, "Var2":nw.String})
```

### Expected results

No error should be raised

### Actual results

This error is raised

```
File "/opt/conda/lib/python3.12/site-packages/narwhals/_utils.py", line 1521, in wrapper
    return fn(*args, **kwds)
           ^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/narwhals/functions.py", line 321, in from_dict
    return ns._dataframe.from_dict(data, schema=schema, context=ns).to_narwhals()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/narwhals/_pandas_like/dataframe.py", line 167, in from_dict
    native = native.astype(Schema(schema).to_pandas(it))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/narwhals/schema.py", line 172, in to_pandas
    name: to_native_dtype(dtype=dtype, dtype_backend=backend)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/conda/lib/python3.12/site-packages/narwhals/_pandas_like/utils.py", line 558, in narwhals_to_native_dtype
    raise AssertionError(msg)
AssertionError: Unknown dtype: <class 'narwhals.dtypes.Object'>
```

### Please run narwhals.show_version() and enter the output below.

```shell
System:
    python: 3.12.10 | packaged by conda-forge | (main, Apr 10 2025, 22:21:13) [GCC 13.3.0]
executable: /opt/conda/bin/python
   machine: Linux-6.6.93-0-virt-x86_64-with-glibc2.39

Python dependencies:
     narwhals: 1.47.1
       pandas: 2.3.0
       polars:
         cudf:
        modin:
      pyarrow:
        numpy: 2.3.1
```

[...]

PR #2957, opened 2025-08-07
feat: Add `Schema.from_{native,<backend>}`

### LINKED ISSUE #2100: enh: nw.Schema.from_native ?
I'm wondering if we should allow `nw.Schema.from_native`. It would be kinda cool, and would allow you to convert any supported schema to say pyarrow schema

dangotbanned:
Will close #2100

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2100
- #2486 
- #3038

## Checklist

- [x] Tests added
- [x] Documented the changes

See notes in (https://github.com/narwhals-dev/narwhals/issues/2100#issuecomment-3160810180) for plan

- [x] Docs
  - [x] Main
    - [x] `from_native`
    - [x] `from_arrow`
    - [x] `from_pandas_like`
    - [x] `from_polars`
  - [x] [docs: Add examples (excl cudf)](https://github.com/narwhals-dev/narwhals/pull/2957/commits/cf613b7c07d293a4459d34a14a250b58bdd410d0))
- [x] Type aliases
  - [x] Move to `nw.typing` (https://github.com/narwhals-dev/narwhals/pull/2957/commits/96813ea961409c4cfca498e0cf98eeda3cbf4bcb)


tests/frame/schema_test.py
```diff
@@ -416,3 +419,128 @@ def test_schema_to_pandas_invalid() -> None:
...
+@pytest.fixture
+def target_schema() -> nw.Schema:
+    return nw.Schema(
+        {
+            "a": nw.Int64(),
+            "b": nw.String(),
+            "c": nw.Boolean(),
+            "d": nw.Date(),
+            "e": nw.Time(),
+            "f": nw.Datetime("ns"),
```
dangotbanned: Interesting (https://github.com/narwhals-dev/narwhals/actions/runs/16803764783/job/47590943862?pr=2957)

I chose `[ns]` to try and be backward compatible with `pandas`.
However, on nightly the default has changed from `[ns]` to `[us]` 🙃 

- https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#datetime-resolution-inference

narwhals/schema.py
```diff
@@ -88,6 +90,56 @@ def len(self) -> int:
...
+        backend: Literal[
+            "pandas",
+            "modin",
+            "cudf",
+            Implementation.PANDAS,
+            Implementation.MODIN,
+            Implementation.CUDF,
+        ]
+        | ModuleType = "pandas",
+    ) -> Self:
```
dangotbanned: ### ~~TODO~~ TODONE
https://github.com/narwhals-dev/narwhals/issues/2100#issuecomment-3160810180
- [x] Replace with inference somehow?

I'm hoping this issue can be worked around:

```py
import pandas as pd
import modin.pandas as mpd

>>> mpd.ArrowDtype
pandas.core.dtypes.dtypes.ArrowDtype

>>> pd.ArrowDtype
pandas.core.dtypes.dtypes.ArrowDtype
```

We'll see 😅

dangotbanned: Right, I think we have some more cases like #2950 

https://github.com/narwhals-dev/narwhals/blob/ac376acad2efcbcb929a90ff99ed66ca42ce3bda/narwhals/_pandas_like/utils.py#L503-L528


With those aligned, it means only `cuDF` is different - which solves this issue as all the deviations are defined in https://github.com/rapidsai/cudf/blob/28613267714446c29a28183522fea86db24cae1c/python/cudf/cudf/core/dtypes.py

dangotbanned: - Part 1 [feat: Add dtype/schema guards](https://github.com/narwhals-dev/narwhals/pull/2957/commits/0b4c05d41499f0561fe4e4a8635f445e5852fff6)
- Part 2 [feat(DRAFT): Re-do pandas-like](https://github.com/narwhals-dev/narwhals/pull/2957/commits/6bd8085532ea867a3cbc287c0531b336b647eadf)

narwhals/schema.py
```diff
@@ -7,21 +7,29 @@
...
     import polars as pl
     import pyarrow as pa
+    from typing_extensions import Self, TypeAlias
 
     from narwhals.dtypes import DType
     from narwhals.typing import DTypeBackend
 
+    IntoArrowSchema: TypeAlias = "pa.Schema | Mapping[str, pa.DataType]"
+    IntoPolarsSchema: TypeAlias = "pl.Schema | Mapping[str, pl.DataType]"
+    IntoPandasSchema: TypeAlias = Mapping[str, pd.api.extensions.ExtensionDtype]
```
dangotbanned: Might be a better fit to name these `From<backend>Schema`

narwhals/_pandas_like/utils.py
```diff
@@ -282,16 +282,18 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) ->
...
         # Per conversations with their maintainers, they don't support arbitrary
         # objects, so we can just return String.
         return dtypes.String()
-
+    infer = pd.api.types.infer_dtype
     # Arbitrary limit of 100 elements to use to sniff dtype.
-    inferred_dtype = pd.api.types.infer_dtype(series.head(100), skipna=True)
+    inferred_dtype = (
+        "empty" if isinstance(series, tuple) else infer(series.head(100), skipna=True)
+    )
```
dangotbanned: This was a compromise for handling the object dtype consistently (https://github.com/narwhals-dev/narwhals/issues/2100#issuecomment-3160810180)

My thought was that here we define that an empty `Series` with object dtype should be *upcast* to `nw.String`.
I'm viewing a schema as equivalent to a `DataFrame` containing **only** empty `Series`.

However, given (#2851, #2953) I'm not so sure that we should be doing this *unconditionally* and without documenting it.

But overall, I think the edge-case of *wanting* an `nw.Object` is less important than being **internally** consistent 🙂

narwhals/schema.py
```diff
@@ -88,6 +104,192 @@ def len(self) -> int:
...
+        from_native = (
+            cls.from_cudf
+            if get_cudf() and any(is_cudf_dtype(dtype) for dtype in schema.values())
+            else cls.from_pandas
+        )
+        return from_native(schema)
+
+    # NOTE: Just aliasing for consistency
+    # Only `cuDF` has unique paths
+    from_modin = from_pandas
```
dangotbanned: See also ([chore(typing): ignore mypy for now](https://github.com/narwhals-dev/narwhals/pull/2957/commits/12e6389aae2c91df6912125d7767004dc53ee4b4))

narwhals/schema.py
```diff
@@ -88,6 +104,192 @@ def len(self) -> int:
...
+    @classmethod
+    def from_cudf(cls, schema: IntoPandasSchema, /) -> Self:
+        """Construct a Schema from a cudf schema representation.
+
+        Arguments:
+            schema: A mapping of column names to cudf data types.
+
+        Returns:
+            A Narwhals Schema.
+        """
```
dangotbanned: Open to ideas on an example for `cuDF`

narwhals/schema.py
```diff
@@ -88,6 +104,192 @@ def len(self) -> int:
         """
         return len(self)
 
+    @classmethod
+    def from_arrow(cls, schema: IntoArrowSchema, /) -> Self:
```
dangotbanned, line 96: I've made this parameter *positional-only* for flexibility on our end.

I don't feel strongly about keeping it that way, but we should probably rename it to `native_schema` for consistency elsewhere.
E.g. `nw.from_native` vs `nw.Schema.__init__` distinguish native vs narwhals-levels

docs/api-reference/schema.md
```diff
@@ -6,6 +6,13 @@
       members:
         - names
         - dtypes
+        - from_arrow
+        - from_cudf
+        - from_modin
+        - from_native
+        - from_pandas
+        - from_pandas_like
```
MarcoGorelli: perhaps just `from_pandas_like` is enough?

dangotbanned: If we were starting from scratch, that would be my choice as well

It was bugging me that we wouldn't have the `from|to` pairs aligned though 

- https://narwhals-dev.github.io/narwhals/api-reference/schema/#narwhals.schema.Schema.to_pandas

MarcoGorelli: tbh i think that's fine, it seems fairly natural that `from_` can be broader than `to_`

dangotbanned: Right, just so I'm clear, is this the API you want everywhere?

```py
class Thing:
    @classmethod
    def from_native(cls, native: Any) -> Self: ...
    @classmethod
    def from_arrow(cls, native: Any) -> Self: ...
    @classmethod
    def from_pandas_like(cls, native: Any) -> Self: ...
    @classmethod
    def from_polars(cls, native: Any) -> Self: ...

    def to_native(self, *args: Any, **kwds: Any) -> Any: ...
    def to_arrow(self, *args: Any, **kwds: Any) -> Any: ...
    def to_pandas(self, *args: Any, **kwds: Any) -> Any: ...
    def to_polars(self, *args: Any, **kwds: Any) -> Any: ...
```

I'm thinking ahead to (https://github.com/narwhals-dev/narwhals/issues/2100#issuecomment-3160810180), (#2486)

MarcoGorelli: i think so yes

dangotbanned: Sure thing, should be able to get that done later 🙂 

### Update
As promised 🎁 

([revert: remove `from_{cudf,modin,pandas}`](https://github.com/narwhals-dev/narwhals/pull/2957/commits/ef107aa1309e7701a8bd823402155f3d0947a3b6))

narwhals/schema.py
```diff
@@ -161,48 +149,30 @@ def from_pandas(cls, schema: IntoPandasSchema, /) -> Self:
...
+            >>>
             >>> mapping = {
             ...     "a": pd.DatetimeTZDtype("us", "UTC"),
             ...     "b": pd.ArrowDtype(pa.date32()),
             ...     "c": pd.StringDtype("python"),
             ...     "d": np.dtype("uint8"),
             ... }
             >>>
-            >>> nw.Schema.from_pandas(mapping)
+            >>> nw.Schema.from_pandas_like(mapping)
```
dangotbanned: I wanted to keep the two examples for pandas, since it doesn't have a concept of schemas - which makes creating one not very intuitive 😢

narwhals/_pandas_like/utils.py
MarcoGorelli: i'm fairly confused about `tuple()` being here - could we address this in a separate precursor PR if there's issues that need addressing ahead of this feature?

dangotbanned: Thanks for taking a look Marco!

`tuple()` was added because it is passed in here (L373) and I did my best to explain in (https://github.com/narwhals-dev/narwhals/pull/2957#discussion_r2265281184)

https://github.com/narwhals-dev/narwhals/blob/f61c75060cbd75b1a6ffb6708ba2e273020a4d3d/narwhals/schema.py#L358-L378

--- 

I guess a shorter version is that this error implies we already have a `PandasLikeSeries`

https://github.com/narwhals-dev/narwhals/blob/f61c75060cbd75b1a6ffb6708ba2e273020a4d3d/narwhals/_pandas_like/utils.py#L362-L365

- In this case, we don't 
  - but an empty `PandasLikeSeries` will get the same result from `pd.api.types.infer_dtype` as any other empty `Iterable`.
- `object_native_to_narwhals_dtype` only needs to be restricted to `PandasLikeSeries` because of the `head(100)` call
- These changes were to reuse that `Implementation` + `Version` branching path to handle `Object` vs `String` consistently

I used `tuple[()]` so that passing e.g. `(1,2)` will continue to get rejected by a type checker.
`tuple` is unique in that sense because no other type can represent empty-ness :smile:

dangotbanned: On a second pass, this seems easier to read.
I think I went with the other option since it *can* skip the `infer_dtype` call entirely - but it was harder to follow

([refactor: Pass `()` to `infer` to be more explicit](https://github.com/narwhals-dev/narwhals/pull/2957/commits/06a4ef98841090b818ca98a00345fbbb321b9291))

MarcoGorelli: can we do this separately? it seems unrelated to the proposed feature

dangotbanned: > it seems unrelated to the proposed feature

I don't understand how this *isn't* related to this feature?

I need to be able to convert a pandas object dtype to a narwhals dtype.

@MarcoGorelli how would you propose I do that?

- I can't use `object_native_to_narwhals_dtype`, since that now (https://github.com/narwhals-dev/narwhals/pull/2957/commits/33f67ddc4521f8e28583f319b1fcada4212d5793) requires a `PandasLikeSeries`
- I can't use `native_to_narwhals_dtype` since that says I should handle it differently

MarcoGorelli: sorry i meant - if we need to refactor pandas-like utils, can we do that separately, and then this pr only deals with adding the new feature?

dangotbanned: Ohhh did you mean do that *before* this one?

If so, then sure no *object*-ions from me 😄 

### Edit
- [x] #3038

dangotbanned:
Thanks for the review 😍

Issue #629, opened 2024-07-26
ci: make official list of Narwhals gifs to use in PR reviews


MarcoGorelli:
Currently I just use: https://chriskuehl.github.io/shipit/

But people have been very creative with their gifs

We should probably have an official list, and then, when we make a "maintainers' guide", we instruct people to please use one from the list when submitting reviews

dead serious here

anopsy:
Yes, let's extend that list with some dolphin, orcas and maybe whales gifs (as narwhals gifs are almost non existent). I'll try to collect those

dangotbanned:
I've collected a bunch now that I've seen before and some **new** ones.

I clearly have too much time on my hands ... 😅 

## Seen before
![this narwhal loves you](https://media1.tenor.com/m/X-CluTTK-Y8AAAAC/narwhal-love.gif) ![shaky narwhal](https://steemitimages.com/p/LcTxR7u1XKaa3e4T1EBuBP18JezPvjFFo8gNuE9CiKHBn3j7v3eHTK6u888Hr5fwwnHbD78aU6mDLG7iCb9r2bYBaT6M9hwR3Caw6RbhKfmfbDixNbst2uNbtcktxyUSszUpXPdQFWHgxtMyfX8Mp7ddF?format=match&mode=fit)  <img src="https://media1.tenor.com/m/GpVEolKSa0cAAAAd/unicorn-whale.gif" width=304 alt="very shaky narwhal">

![cat-ptain on a narwhal](https://64.media.tumblr.com/64fb369db15c1b058a5335f8f9ae24ba/tumblr_oql8iwJISo1qak30bo1_500.gifv) ![swing narwhals](https://media.giphy.com/media/v1.Y2lkPTc5MGI3NjExdmZkaHJ2N24wdDU4NzcydmQ5bHphNWtmczRzd2FmNTJidHByOXRwayZlcD12MV9naWZzX3NlYXJjaCZjdD1n/mPzDUb0BlwCKsi0vZf/giphy.gif)

[...]

PR #2958, opened 2025-08-07
fix: More aligning `modin`, `pandas` dtypes

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Follow up to #2950
- Related https://github.com/narwhals-dev/narwhals/pull/2957#discussion_r2260122808

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Expecting to cause some xpass on `modin` w/ `pandas<2` in ci

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -129,6 +129,16 @@ UNITS_DICT: Mapping[IntervalUnit, NativeIntervalUnit] = {
     "ns": "nanosecond",
 }
 
+PANDAS_VERSION = Implementation.PANDAS._backend_version()
+"""Static backend version for `pandas`.
+
+Always available if we reached here, due to a module-level import.
+"""
+
+
+def is_pandas_or_modin(implementation: Implementation) -> bool:
+    return implementation in {Implementation.PANDAS, Implementation.MODIN}
+
 
 def align_and_extract_native(
     lhs: PandasLikeSeries, rhs: PandasLikeSeries | object
@@ -499,10 +509,9 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912, PLR0915
         # or at least, convert_dtypes(dtype_backend='pyarrow') doesn't
         # convert to it?
         return "category"
-    backend_version = implementation._backend_version()
     if isinstance_or_issubclass(dtype, dtypes.Datetime):
         # Pandas does not support "ms" or "us" time units before version 2.0
-        if implementation is Implementation.PANDAS and backend_version < (
+        if is_pandas_or_modin(implementation) and PANDAS_VERSION < (
             2,
         ):  # pragma: no cover
             dt_time_unit = "ns"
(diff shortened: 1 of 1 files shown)


dangotbanned:
@FBruzzesi I was expecting this to XPASS in CI, but was suprised to see no effect 🤔 

I think this PR introduced the condition(s) I've updated, but I can't work out where `modin` is being tested

- #960

FBruzzesi:
> @FBruzzesi I was expecting this to XPASS in CI, but was suprised to see no effect 🤔

@dangotbanned I assume it's because https://github.com/narwhals-dev/narwhals/blob/5bcdb6e6e1e12f430686cd5b7ea9631419a3e7bd/tests/expr_and_series/cast_test.py#L70-L72

i.e. one of the many dtype casting is still failing. I can take a deeper look later.

And pyarrow backed was already passing

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2958#issuecomment-3165392289

Thanks @FBruzzesi that was super helpful 🎉 

([test: Remove a bunch of modin xfails](https://github.com/narwhals-dev/narwhals/pull/2958/commits/433c49dec9c9f6f344715988d3aa06e72b2d2260)) is now working locally foir modin-only

Only the `String` + ``'<U0'`` (fixed-width numpy?) dtype is a difference from `pandas` 

This makes (#2957) waaaaaaaaaaaaaaaaaay easier 😄

tests/expr_and_series/cast_test.py
```diff
@@ -59,19 +59,19 @@
...
         15,
     ):  # pragma: no cover
         pytest.skip()
-    if "modin_constructor" in str(constructor):
-        # TODO(unassigned): in modin, we end up with `'<U0'` dtype
-        request.applymarker(pytest.mark.xfail)
 
-    if "pyspark" in str(constructor):
+    if "modin_constructor" in str(constructor):
+        incompatible_columns = MODIN_INCOMPATIBLE_COLUMNS  # pragma: no cover
```
FBruzzesi: Would you fancy keeping an unassigned TODO? I think we should cover this (_eventually_)

dangotbanned: Sure, I'm not opposed to keeping an `xfail` *if* you think we'll eventually fix it.

My *preference* though would be isolating this case/providing more detail on the `xfail`.
For example, these have more that someone could look at and say *oh, I know how to fix that ...* 

https://github.com/narwhals-dev/narwhals/blob/dc78c538253724bb9dc16f9d2a849ebc3cd90150/tests/series_only/from_iterable_test.py#L134-L161

FBruzzesi: I agree on having more granular control and having a good reason to skip/xfail something.

For the former, it might suffices doing something like:

```diff
- assert dict(result.collect_schema()) == cast_map
+ for (lkey, ltype), (rkey, rtype) in zip(result.collect_schema().items(), cast_map.items())
+    # if some condition skip/xfail
+    assert ltype == rtype, f"{lkey} have different dtypes: {ltype} and {rtype}"
```

you get the idea 😂

For the xfail/skip reason, there are cases for which attach issues as comments. I don't think it is standardized in the test suite

Update: https://github.com/narwhals-dev/narwhals/pull/2958/commits/c1a4fb8b4b76e20ac0e87e32414a59016309e5cc

dangotbanned: Thanks @FBruzzesi LGTM

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2958#pullrequestreview-3100062485

@FBruzzesi oooh add that to #629!

PR #2962, opened 2025-08-07
feat: Add  `{Expr,Series}.is_close`

### LINKED ISSUE #2911: [Enh]: Add `is_close`

I use `is_close` quite often. For example, in a dataframe validation library, I provide a built-in check https://github.com/CangyuanLi/checkedframe/blob/8efbdd08c6c576d64ebde81a5c8cb3a3f3862be1/src/checkedframe/_checks.py#L216 to assert that two columns / values are approximately equal. Although it is easy to implement yourself, it would be great to have this natively in Narwhals!

Polars has recently added `is_close` (https://github.com/pola-rs/polars/pull/23273), so it would be great to have this in Narwhals as well.

If an `is_close` operation is supported natively, Narwhals could dispatch to that. If not, Narwhals could fall back to a generic implementation. There shouldn't be any need to write backend-specific code, as all the building blocks for an `is_close` function are already implemented.

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2911 

## Checklist

- [x] Tests added
- [x] Documented the changes

Reason for "help wanted" label:

1. By implementing `_is_close_impl` is almost possible to avoid any code duplication. All we need is already in narwhals, and no backend has this natively implemented (I think). However.... typing issues (which I didn't spend too much time on as of now but will give it a try)
2. One of the test is failing for the SQL-like backends. This is `test_is_close_expr_with_scalar` with other being a finite scalar and `nans_equal=True` still returning True when the column value is nan. ~~It's been a long day and I couldn't point down the reason. It could be that logical operators are not supposed to be used with scalar booleans. If that's the case, the implementation might need to be split up between eager and lazy.~~ Logic is now fixed 😇

---

Also I keep submitting PRs which as 600 line changes with one functionality implemented 😩

### DIFF
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -27,7 +27,8 @@ from narwhals._compliant.typing import (
     LazyExprT,
     NativeExprT,
 )
-from narwhals._utils import _is_close_impl, _StoresCompliant
+from narwhals._compliant.utils import IsClose
+from narwhals._utils import _StoresCompliant
 from narwhals.dependencies import get_numpy, is_numpy_array
 
 if TYPE_CHECKING:
@@ -75,7 +76,7 @@ class NativeExpr(Protocol):
     def __ne__(self, value: Any, /) -> Self: ...  # type: ignore[override]
 
 
-class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
+class CompliantExpr(IsClose, Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
     _implementation: Implementation
     _version: Version
     _evaluate_output_names: EvalNames[CompliantFrameT]
@@ -240,18 +241,6 @@ class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
     ) -> Sequence[str]:
         names = self._evaluate_output_names(frame)
         return alias(names) if (alias := self._alias_output_names) else names
-
-    def is_close(
-        self,
-        other: Self | NumericLiteral,
-        *,
-        abs_tol: float,
-        rel_tol: float,
-        nans_equal: bool,
-    ) -> Self:
-        return _is_close_impl(
-            self, other, abs_tol=abs_tol, rel_tol=rel_tol, nans_equal=nans_equal
-        )
 
     @property
     def str(self) -> StringNamespace[Self]: ...
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -194,6 +194,15 @@ class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
         self, window_size: int, *, min_samples: int, center: bool, ddof: int
     ) -> Self: ...
 
+    def is_close(
+        self,
+        other: Self | NumericLiteral,
+        *,
+        abs_tol: float,
+        rel_tol: float,
+        nans_equal: bool,
+    ) -> Self: ...
+
     def __and__(self, other: Any) -> Self: ...
     def __or__(self, other: Any) -> Self: ...
     def __add__(self, other: Any) -> Self: ...
@@ -891,6 +900,22 @@ class EagerExpr(
 
     def sqrt(self) -> Self:
         return self._reuse_series("sqrt")
+
+    def is_close(
+        self,
+        other: Self | NumericLiteral,
+        *,
+        abs_tol: float,
+        rel_tol: float,
+        nans_equal: bool,
+    ) -> Self:
+        return self._reuse_series(
+            "is_close",
+            other=other,
+            abs_tol=abs_tol,
+            rel_tol=rel_tol,
+            nans_equal=nans_equal,
+        )
 
     @property
     def cat(self) -> EagerExprCatNamespace[Self]:
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -194,15 +194,6 @@ class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
         self, window_size: int, *, min_samples: int, center: bool, ddof: int
     ) -> Self: ...
 
-    def is_close(
-        self,
-        other: Self | NumericLiteral,
-        *,
-        abs_tol: float,
-        rel_tol: float,
-        nans_equal: bool,
-    ) -> Self: ...
-
     def __and__(self, other: Any) -> Self: ...
     def __or__(self, other: Any) -> Self: ...
     def __add__(self, other: Any) -> Self: ...
@@ -239,6 +230,65 @@ class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
     ) -> Sequence[str]:
         names = self._evaluate_output_names(frame)
         return alias(names) if (alias := self._alias_output_names) else names
+
+    def is_close(
+        self,
+        other: Self | NumericLiteral,
+        *,
+        abs_tol: float,
+        rel_tol: float,
+        nans_equal: bool,
+    ) -> Self:
+        from decimal import Decimal
+
+        other_abs: Self | NumericLiteral
+        other_is_nan: Self | bool
+        other_is_inf: Self | bool
+        other_is_not_inf: Self | bool
+
+        if isinstance(other, (float, int, Decimal)):
+            from math import isinf, isnan
+
+            other_abs, other_is_nan, other_is_inf = abs(other), isnan(other), isinf(other)  # type: ignore[assignment]
+
+            # Define the other_is_not_inf variable to prevent triggering the following warning:
+            # > DeprecationWarning: Bitwise inversion '~' on bool is deprecated and will be
+            # >     removed in Python 3.16. This returns the bitwise inversion of the
+            # >     underlying int object and is usually not what you expect from negating
+            # >     a bool. Use the 'not' operator for boolean negation or ~int(x) if you
+            # >     really want the bitwise inversion of the underlying int.
+            other_is_not_inf = not other_is_inf
+
+        else:
+            other_abs, other_is_nan = other.abs(), other.is_nan()
+            other_is_not_inf = other.is_finite() | other_is_nan
+            other_is_inf = ~other_is_not_inf
+
+        rel_threshold = self.abs().clip(lower_bound=other_abs, upper_bound=None) * rel_tol
+        tolerance = rel_threshold.clip(lower_bound=abs_tol, upper_bound=None)
+
+        self_is_nan = self.is_nan()
+        self_is_inf = ~(self.is_finite() | self_is_nan)
+
+        # Values are close if abs_diff <= tolerance, and both finite
+        is_close = ((self - other).abs() <= tolerance) & (~self_is_inf) & other_is_not_inf
+
+        # Handle infinity cases: infinities are "close" only if they have the same sign
+        self_sign, other_sign = self > 0, other > 0
+        is_same_inf = self_is_inf & other_is_inf & (self_sign == other_sign)
+        result = is_close | is_same_inf
+
+        # Handle nan cases:
+        #   * nans_equals = True => if both values are NaN, then True
+        #   * nans_equals = False => if any value is NaN, then False
+        if nans_equal:
+            both_nan = self_is_nan & other_is_nan
+            result = result | both_nan
+        else:
+            either_nan = self_is_nan | other_is_nan
+            result = result & ~either_nan
+
+        return result
 
     @property
     def str(self) -> StringNamespace[Self]: ...
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -27,7 +27,7 @@ from narwhals._compliant.typing import (
     LazyExprT,
     NativeExprT,
 )
-from narwhals._utils import _StoresCompliant
+from narwhals._utils import _is_close_impl, _StoresCompliant
 from narwhals.dependencies import get_numpy, is_numpy_array
 
 if TYPE_CHECKING:
@@ -239,56 +239,9 @@ class CompliantExpr(Protocol[CompliantFrameT, CompliantSeriesOrNativeExprT_co]):
         rel_tol: float,
         nans_equal: bool,
     ) -> Self:
-        from decimal import Decimal
-
-        other_abs: Self | NumericLiteral
-        other_is_nan: Self | bool
-        other_is_inf: Self | bool
-        other_is_not_inf: Self | bool
-
-        if isinstance(other, (float, int, Decimal)):
-            from math import isinf, isnan
-
-            other_abs, other_is_nan, other_is_inf = abs(other), isnan(other), isinf(other)  # type: ignore[assignment]
-
-            # Define the other_is_not_inf variable to prevent triggering the following warning:
-            # > DeprecationWarning: Bitwise inversion '~' on bool is deprecated and will be
-            # >     removed in Python 3.16. This returns the bitwise inversion of the
-            # >     underlying int object and is usually not what you expect from negating
-            # >     a bool. Use the 'not' operator for boolean negation or ~int(x) if you
-            # >     really want the bitwise inversion of the underlying int.
-            other_is_not_inf = not other_is_inf
-
-        else:
-            other_abs, other_is_nan = other.abs(), other.is_nan()
-            other_is_not_inf = other.is_finite() | other_is_nan
-            other_is_inf = ~other_is_not_inf
-
-        rel_threshold = self.abs().clip(lower_bound=other_abs, upper_bound=None) * rel_tol
-        tolerance = rel_threshold.clip(lower_bound=abs_tol, upper_bound=None)
-
-        self_is_nan = self.is_nan()
-        self_is_inf = ~(self.is_finite() | self_is_nan)
-
-        # Values are close if abs_diff <= tolerance, and both finite
-        is_close = ((self - other).abs() <= tolerance) & (~self_is_inf) & other_is_not_inf
-
-        # Handle infinity cases: infinities are "close" only if they have the same sign
-        self_sign, other_sign = self > 0, other > 0
-        is_same_inf = self_is_inf & other_is_inf & (self_sign == other_sign)
-        result = is_close | is_same_inf
-
-        # Handle nan cases:
-        #   * nans_equals = True => if both values are NaN, then True
-        #   * nans_equals = False => if any value is NaN, then False
-        if nans_equal:
-            both_nan = self_is_nan & other_is_nan
-            result = result | both_nan
-        else:
-            either_nan = self_is_nan | other_is_nan
-            result = result & ~either_nan
-
-        return result
+        return _is_close_impl(
+            self, other, abs_tol=abs_tol, rel_tol=rel_tol, nans_equal=nans_equal
+        )
 
     @property
     def str(self) -> StringNamespace[Self]: ...
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -62,7 +62,17 @@ class NativeExpr(Protocol):
     """
 
     def between(self, *args: Any, **kwds: Any) -> Any: ...
-    def isin(self, *args: Any, **kwds: Any) -> Any: ...
+
+    # NOTE: None of these are annotated for `dx.Series`, but are added imperatively
+    # Probably better to define a sub-protocol for `NativeSQLExpr`
+    # - match `dx.Series` to `NativeExpr`
+    # - match the others to `NativeSQLExpr`
+    def __gt__(self, value: Any, /) -> Self: ...
+    def __lt__(self, value: Any, /) -> Self: ...
+    def __ge__(self, value: Any, /) -> Self: ...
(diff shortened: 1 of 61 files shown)


dangotbanned:
> Also I keep submitting PRs which as 600 line changes with one functionality implemented 😩

😂😂😂

narwhals/_compliant/expr.py
```diff
@@ -892,6 +904,22 @@ def exp(self) -> Self:
...
+        rel_tol: float,
+        nans_equal: bool,
+    ) -> Self:
+        return self._reuse_series(
+            "is_close",
+            other=other,
+            abs_tol=abs_tol,
+            rel_tol=rel_tol,
+            nans_equal=nans_equal,
+        )
```
FBruzzesi, line 920: This might not be needed 🤔

narwhals/_utils.py
```diff
@@ -2035,3 +2036,62 @@ def deep_attrgetter(attr: str, *nested: str) -> attrgetter[Any]:
 def deep_getattr(obj: Any, name_1: str, *nested: str) -> Any:
     """Perform a nested attribute lookup on `obj`."""
     return deep_attrgetter(name_1, *nested)(obj)
+
+
+def _is_close_impl(
```
FBruzzesi: In principle, we could use this implementation also for polars pre 1.32.0, however polars has a couple of native methods which play a bit nicer (e.g. `.sign`, `.is_infinite`, `.not_`). If we were to introduce those, then we can have a single implementation function

narwhals/expr.py
```diff
@@ -2347,6 +2347,95 @@ def sqrt(self) -> Self:
...
+            |└────────┴────────┴──────────┘|
+            └──────────────────────────────┘
+        """
+        if abs_tol < 0:
+            msg = f"`abs_tol` must be non-negative but got {abs_tol}"
+            raise ComputeError(msg)
+
+        if not (0 <= rel_tol < 1):
+            msg = f"`rel_tol` must be in the range [0, 1) but got {rel_tol}"
+            raise ComputeError(msg)
```
FBruzzesi, line 2421: For lazy backends, we are not raising for non-numeric dtypes

FBruzzesi:
@dangotbanned I am quite happy with the state of this. I left a couple of comments. Main blocker is typing. In a way I feel like mypy it's not picking up the obvious, but clearly I am not doing something as it's expected to be done 😂

dangotbanned:
I've got 3 different ideas for fixing the typing, each with varying trade-offs

I'll try to write them up when I get a chance later today/tomorrow

narwhals/_utils.py
dangotbanned: `not_` and `__invert__` are identical btw

FBruzzesi:
> I've got 3 different ideas for fixing the typing, each with varying trade-offs
> 
> I'll try to write them up when I get a chance later today/tomorrow

@dangotbanned look at https://github.com/narwhals-dev/narwhals/pull/2962/commits/22e22038535b84750adf87b7f65e68bfaf87c28f 🙏🏼

dangotbanned:
> > I've got 3 different ideas for fixing the typing, each with varying trade-offs
> > I'll try to write them up when I get a chance later today/tomorrow
> 
> @dangotbanned look at [22e2203](https://github.com/narwhals-dev/narwhals/commit/22e22038535b84750adf87b7f65e68bfaf87c28f) 🙏🏼

**Bad news**: that didn't fix the typing 😭
(there's still 2x type ignore(s), probably more if not doing 3x assignments on a single line)

**Good news**: I still have 3 other ideas 😄
(tomorrow!)

dangotbanned:
> > I've got 3 different ideas for fixing the typing, each with varying trade-offs
> > I'll try to write them up when I get a chance later today/tomorrow
> **Good news**: I still have 3 other ideas 😄 (tomorrow!)

## Idea 1 - Use a constrained `TypeVar`
Some related threads
- https://github.com/narwhals-dev/narwhals/pull/2339
- https://github.com/narwhals-dev/narwhals/pull/2325#discussion_r2051473155
- https://github.com/narwhals-dev/narwhals/pull/2325#discussion_r2051548917
- [#2944/`join_test.py` diff](https://github.com/narwhals-dev/narwhals/pull/2944/files/89095ceb6563e818e8f9c021bc76e730870145e0#diff-ff8314856956318d0da461d7cc2710a6b18d3c052581be7990ae0023a9e689ee)

(https://github.com/narwhals-dev/narwhals/commit/22e22038535b84750adf87b7f65e68bfaf87c28f) changes things a bit from when I first thought of this, but it would've looked like:

```py
CompliantSeriesOrExprT = TypeVar("CompliantSeriesOrExprT", CompliantSeriesAny, CompliantExprAny)
```

### Downsides

*If* that solves most of the issues, I still expect there will be some `cast`(s) required.

My guess is here:

<details><summary><code>CompliantSeries.is_close</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/22e22038535b84750adf87b7f65e68bfaf87c28f/narwhals/_compliant/series.py#L288-L298

</p>
</details> 



*might* need to change to this, since the function would now return `CompliantSeriesAny`

```py
        from typing import cast

        result = _is_close_impl(
            self, other, abs_tol=abs_tol, rel_tol=rel_tol, nans_equal=nans_equal
        )
        return cast("Self", result)
```

## Idea 2 - Move `utils._is_close_impl` into `nw.Expr`
*and dispatch `nw.Series` to `nw.Expr`*

If we're already inside `Expr`, we have a bounded `Self` to use for `self`, `other` - rather than handling `CompliantExprT | CompliantSeriesT`.

I can explain the benefits more if needed, but overall seems quite clean to me

### Downsides

I guess the only wart is `polars`, but that has it's own version now *anyway*.

`nw.Series` has [`Series.implementation`](https://narwhals-dev.github.io/narwhals/api-reference/series/#narwhals.series.Series.implementation) but not too sure how branching could work for `nw.Expr` when evaluating for `polars` 🤔


## Idea 3 - Add a `Protocol` for `IsClose`, shared by `Compliant{Expr,Series}`
Both `mypy` and `pyright` are accepting this

You could add it as base for each `Compliant*` and remove their definitions of `is_close` + the `_utils` function

```py
class IsClose(Protocol):
    """Every member defined is a dependency of `is_close`."""

    def __and__(self, other: Any) -> Self: ...
    def __or__(self, other: Any) -> Self: ...
    def __invert__(self) -> Self: ...
    def __sub__(self, other: Any) -> Self: ...
    def __mul__(self, other: Any) -> Self: ...
    def __eq__(self, other: Self | Any) -> Self: ...  # type: ignore[override]
    def __gt__(self, other: Any) -> Self: ...
    def __ge__(self, other: Any) -> Self: ...
    def __lt__(self, other: Any) -> Self: ...
    def __le__(self, other: Any) -> Self: ...
    def abs(self) -> Self: ...
    def is_nan(self) -> Self: ...
    def is_finite(self) -> Self: ...
    def clip(
        self,
        lower_bound: Self | NumericLiteral | TemporalLiteral | None,
        upper_bound: Self | NumericLiteral | TemporalLiteral | None,
    ) -> Self: ...
    def is_close(
        self,
        other: Self | NumericLiteral,
        *,
        abs_tol: float,
        rel_tol: float,
        nans_equal: bool,
    ) -> Self:
        from decimal import Decimal

        other_abs: Self | NumericLiteral
        other_is_nan: Self | bool
        other_is_inf: Self | bool
        other_is_not_inf: Self | bool

        if isinstance(other, (float, int, Decimal)):
            from math import isinf, isnan

            other_abs = other.__abs__()
            other_is_nan = isnan(other)
            other_is_inf = isinf(other)
            other_is_not_inf = not other_is_inf

        else:
            other_abs, other_is_nan = other.abs(), other.is_nan()
            other_is_not_inf = other.is_finite() | other_is_nan
            other_is_inf = ~other_is_not_inf

        rel_threshold = self.abs().clip(lower_bound=other_abs, upper_bound=None) * rel_tol
        tolerance = rel_threshold.clip(lower_bound=abs_tol, upper_bound=None)

        self_is_nan = self.is_nan()
        self_is_not_inf = self.is_finite() | self_is_nan

        # Values are close if abs_diff <= tolerance, and both finite
        is_close = (
            ((self - other).abs() <= tolerance) & self_is_not_inf & other_is_not_inf
        )

        # Handle infinity cases: infinities are "close" only if they have the same sign
        self_sign, other_sign = self > 0, other > 0
        is_same_inf = (~self_is_not_inf) & other_is_inf & (self_sign == other_sign)

        # Handle nan cases:
        #   * nans_equals = True => if both values are NaN, then True
        #   * nans_equals = False => if any value is NaN, then False
        either_nan = self_is_nan | other_is_nan
        result = (is_close | is_same_inf) & ~either_nan

        if nans_equal:
            both_nan = self_is_nan & other_is_nan
            result = result | both_nan

        return result
```

FBruzzesi:
Hey @dangotbanned thanks for the amazing writeup in https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3170943889

It's kind of re-assuring that I somehow thought at a 2/3 of those ideas.

> Idea 1 - Use a constrained TypeVar

This was my first attempt, and yes you are correct that it leads to the issues with the return type not being `Self`:

```terminal
narwhals/_compliant/series.py:296: error: Incompatible return value type (got "CompliantSeries[Any]", expected "Self")  [return-value]
            return _is_close_impl(
                   ^
narwhals/_compliant/expr.py:252: error: Incompatible return value type (got "CompliantExpr[Any, Any]", expected "Self")  [return-value]
            return _is_close_impl(
                   ^
Found 2 errors in 2 files (checked 153 source files)
```

For better or for worse, I retained from using cast, and preferred adding one `type: ignore` in `clip` (the one for `abs` is solved by doing `.__abs__()`, thanks!)

> Idea 2 - Move utils._is_close_impl into nw.Expr

This was indeed working, yet personally I am not too keen in triggering an extra `to_frame` + `get_column` if there is no need.

...which bring us to 

> Idea 3 - Add a Protocol for IsClose, shared by Compliant{Expr,Series}

which I couldn't dream of 🤯  It works smoothly (I assume), maybe a little verbose, yet possibly it's the preferred solution? What's you opinion?

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3171711096

> Hey @dangotbanned thanks for the amazing writeup in https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3170943889
> It's kind of re-assuring that I somehow thought at a 2/3 of those ideas.

aha thanks and I agree great to see we're on the same page 😄  


> > Idea 3 - Add a Protocol for IsClose, shared by Compliant{Expr,Series}

> which I couldn't dream of 🤯 It works smoothly (I assume), maybe a little verbose, yet possibly it's the preferred solution? What's you opinion?

Yeah that would be my preference!

The verbosity is only temporary though, as it gave me an idea for a follow-up.

We could move quite a lot into a shared `Protocol` for `Compliant{Expr,Series}` (maybe named `CompliantColumn` or `CompliantColumnar`)
Essentially anything they both have, but doesn't aggregate/return a scalar could just be defined in one place 😏 

So the end result would be *shrinking* both protocols, removing `IsClose` and just defining it there 🥳 

But we can save that part for another PR (provided it sounds good to you)

narwhals/_compliant/utils.py
```diff
@@ -0,0 +1,87 @@
...
+        upper_bound: Self | NumericLiteral | TemporalLiteral | None,
+    ) -> Self: ...
+    def is_close(
+        self,
+        other: Self | NumericLiteral,
+        *,
+        abs_tol: float,
+        rel_tol: float,
+        nans_equal: bool,
+    ) -> Self:
```
dangotbanned, line 37: https://github.com/narwhals-dev/narwhals/pull/2962#discussion_r2262187495

>  however polars has a couple of native methods which play a bit nicer (e.g. `.sign`, `.is_infinite`, `.not_`).
> If we were to introduce those, then we can have a single implementation function

Okay *blue-sky*, if you had the full `polars` API (excluding `is_close`), what would this look like? 

### Edit
I'm curious as I [mentioned on discord](https://discord.com/channels/1235257048170762310/1272835922924273694/1397243168273076375) about `is_not_null` and thought it might be good to collect some of these into an issue to discuss

narwhals/_compliant/utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+        other_abs: Self | NumericLiteral
+        other_is_nan: Self | bool
+        other_is_inf: Self | bool
+        other_is_not_inf: Self | bool
+
+        if isinstance(other, (float, int, Decimal)):
+            from math import isinf, isnan
+
+            other_abs = other.__abs__()
```
dangotbanned, line 49:
```suggestion
            # NOTE: See https://discuss.python.org/t/inferred-type-of-function-that-calls-dunder-abs-abs/101447
            other_abs = other.__abs__()
```

narwhals/_compliant/utils.py
FBruzzesi: I think that if we eventually move into the direction of https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3171900789, then it could be easy to implement it once as:


```py
class CompliantColumn(Protocol):
    def __and__(self, other: Any) -> Self: ...
    def __or__(self, other: Any) -> Self: ...
    def __invert__(self) -> Self: ...
    def __sub__(self, other: Any) -> Self: ...
    def __mul__(self, other: Any) -> Self: ...
    def __eq__(self, other: Self | Any) -> Self: ...  # type: ignore[override]
    def __gt__(self, other: Any) -> Self: ...
    def __ge__(self, other: Any) -> Self: ...
    def __lt__(self, other: Any) -> Self: ...
    def __le__(self, other: Any) -> Self: ...
    def abs(self) -> Self: ...
    def is_nan(self) -> Self: ...
    def is_finite(self) -> Self: ...
    # and many others

def _is_close_impl(  # current logic
    self: CompliantColumn,
    other: CompliantColumn | NumericLiteral,
    ...
) -> CompliantColumn: ... 

class CompliantSeries(CompliantColumn):
    ...
    def is_close(self, other, ...): return _is_close_impl(self, other, ...)

class CompliantExpr(CompliantColumn):
    ...
    def is_close(self, other, ...): return _is_close_impl(self, other, ...)

class PolarsExpr:  # does not inherit, but satisfy the protocol
    ...
    def is_close(self, other, ...):
        if self._backend_version < (1, 32):
            return _is_close_impl(self, other, ...)
        else: ...
```

dangotbanned: Oop, I think because I highlighted the *signature* (https://github.com/narwhals-dev/narwhals/pull/2962#discussion_r2264907413) I made this unclear(sorry! 🤦)

> Okay blue-sky, if you had the full polars API (excluding is_close), what would this look like?

By that I meant how would the  *method body* look?
E.g. would we be able to do fewer operations/would it be shorter/easier to read/are there any other methods that could be helpful?

FBruzzesi: Whops sorry my bad!

The only _real_ simplification I see is how to get infinite values without having to first check for finiteness and then combine with nan mask.

`sign` is definitely optional, since `>0` does the job (the difference between the two is the value 0, but since we use it conditionally to the fact that the values are ±inf, it doesn't matter here).

> how would the method body look?

After parsing other (literal vs expr), it would look exactly the same as per polars pre 1.32

narwhals/_compliant/utils.py
dangotbanned: Cross ref (https://github.com/python/mypy/issues/19609)

narwhals/_compliant/typing.py
```diff
@@ -69,6 +69,7 @@ class ScalarKwargs(TypedDict, total=False):
 CompliantExprAny: TypeAlias = "CompliantExpr[Any, Any]"
 CompliantSeriesAny: TypeAlias = "CompliantSeries[Any]"
 CompliantSeriesOrNativeExprAny: TypeAlias = "CompliantSeriesAny | NativeExpr"
+CompliantSeriesOrExprAny: TypeAlias = "CompliantSeriesAny | CompliantExprAny"
```
dangotbanned: Seems like this one isn't needed now 😄
```suggestion
```

narwhals/expr.py
```diff
@@ -2347,6 +2347,95 @@ def sqrt(self) -> Self:
...
+        rel_tol: float = 1e-09,
+        nans_equal: bool = False,
+    ) -> Self:
+        r"""Check if this expression is close, i.e. almost equal, to the other expression.
+
+        Two values `a` and `b` are considered close if the following condition holds:
+
+        $$|a-b| \le max \{ \text{rel_tol} \cdot max \{ |a|, |b| \}, \text{abs_tol} \}$$
+
+        Arguments:
```
dangotbanned: This doesn't seem to be rendering in mkdocs or vscode 😭 

image

image

narwhals/series.py
```diff
@@ -2763,6 +2763,80 @@ def sqrt(self) -> Self:
...
+            >>> import pyarrow as pa
+            >>> import narwhals as nw
+            >>>
+            >>> data = [1.0, float("inf"), 1.41, None, float("nan")]
+            >>> s_native = pa.chunked_array([data])
+            >>> s = nw.from_native(s_native, series_only=True)
+            >>> s.is_close(1.4, abs_tol=0.1).to_native()  # doctest:+ELLIPSIS
+            <pyarrow.lib.ChunkedArray object at ...>
+            [
+              [
```
dangotbanned, line 2809: Is this a sneaky solution to (https://github.com/narwhals-dev/narwhals/pull/2776#issuecomment-3148403247) 😂?

tests/expr_and_series/is_close_test.py
```diff
@@ -0,0 +1,241 @@
...
+        x.is_close(y, abs_tol=abs_tol)
+
+    with pytest.raises(ComputeError, match=msg):
+        x.is_close(y, abs_tol=abs_tol, rel_tol=999)
+
+    with pytest.raises(ComputeError, match=msg):
+        df.select(nw.col("x").is_close(nw.col("y"), abs_tol=abs_tol))
+
+    with pytest.raises(ComputeError, match=msg):
+        df.select(nw.col("x").is_close(nw.col("y"), abs_tol=abs_tol, rel_tol=999))
```
dangotbanned: There's probably a clever way to paramterize the remaining cases, but this could do?

```suggestion
@pytest.mark.parametrize("rel_tol", [1e-09, 999])
def test_is_close_raise_negative_abs_tol(
    constructor_eager: ConstructorEager, rel_tol: float
) -> None:
    df = nw.from_native(constructor_eager(data), eager_only=True)
    x, y = df["x"], df["y"]
    abs_tol = -2
    context = pytest.raises(ComputeError, match=r"abs_tol.+non-negative.+-2")
    with context:
        x.is_close(y, abs_tol=abs_tol, rel_tol=rel_tol)
    with context:
        df.select(nw.col("x").is_close(nw.col("y"), abs_tol=abs_tol, rel_tol=rel_tol))
```

tests/expr_and_series/is_close_test.py
```diff
@@ -0,0 +1,241 @@
...
+
+    nulls = nw.new_series(
+        name="nulls",
+        values=[None] * len(x),
+        dtype=nw.Float64(),
+        backend=df.implementation,
+    )
+    # Tricks to generate nan's and null's for pandas with nullable backends:
+    #   * Square rooting a negative number will generate a NaN
+    #   * Replacing a value with None once the dtype is nullable will generate <NA>'s
```
dangotbanned: Rather than this comment 4 times, would moving it into the docstring of a function work? 😉

dangotbanned (approved):
Looking pretty good, thanks @FBruzzesi

Onloy a few notes/suggestions from me 🙂

narwhals/expr.py
FBruzzesi: Fixed in https://github.com/narwhals-dev/narwhals/pull/2962/commits/d61c5c77cba42a549c5a0f00862ebc8d92fd5a7a

narwhals/series.py
FBruzzesi: 😂 that's the only way to use it in doctest, _maybe_ it's an idea?

FBruzzesi:
Thanks for the additional simplification @dangotbanned

I addressed the 

> Only a few notes/suggestions from me 🙂

in https://github.com/narwhals-dev/narwhals/pull/2962/commits/e4b7a1eed846ea38c47b65100b7b257bba3c7cac and figured out how katex wants `\text{...}` in https://github.com/narwhals-dev/narwhals/pull/2962/commits/d61c5c77cba42a549c5a0f00862ebc8d92fd5a7a - I checked the latter in the docs served locally, not sure how it renders in the editor though

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3180970722)

> I checked the latter in the docs served locally, not sure how it renders in the editor though

I'll look now!

I'm pretty sure the rest was good, but will do a *thrice* over 😄

dangotbanned:
`mkdocs` seems happy now

image


No luck for me in VSCode

image


but it seems to be the same story with `ewm_mean`

https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.ewm_mean

😂 

image

MarcoGorelli:
thanks both, great feature!

Issue #2960, opened 2025-08-07
Tracking: `ibis` typing fixes


dangotbanned:
Aiming to collect issues we've had, with the hope to make some *small* suggestions upstream and make our type checkers happier 😅 

### Code Search
Should be *most* of the individual cases, but ideally we could pull out some common issues from here
- [Using `typing.cast`](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+path%3A%2F%5Enarwhals%5C%2F_ibis%5C%2F%2F+%2F%5B%5E%5C.%5Dcast%5C%28%22%2F&type=code)
- [Using `# (type|pyright): ignore`](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+path%3A%2F%5Enarwhals%5C%2F_ibis%5C%2F%2F+%2F%28type%7Cpyright%29%3A+ignore%2F&type=code)
- [Using `: Any`](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals+path%3A%2F%5Enarwhals%5C%2F_ibis%5C%2F%2F+%2F%3A+Any%2F&type=code)

### Comments
- https://github.com/narwhals-dev/narwhals/pull/2000#discussion_r2047704276
- https://github.com/narwhals-dev/narwhals/pull/2920#issuecomment-3155731605
- https://github.com/narwhals-dev/narwhals/pull/2761#discussion_r2263916976
- https://github.com/narwhals-dev/narwhals/pull/2963#issue-3304248155

### PRs
- [x] #2972

### Upstream
- [ ] https://github.com/ibis-project/ibis/issues/11680
- [ ] https://github.com/ibis-project/ibis/pull/11771
- [ ] https://github.com/ibis-project/ibis/issues/11667
- [ ] https://github.com/ibis-project/ibis/issues/11511
- [ ] https://github.com/ibis-project/ibis/issues/11302
- [ ] https://github.com/ibis-project/ibis/issues/11188
- [ ] https://github.com/ibis-project/ibis/issues/10369

dangotbanned:
@FBruzzesi, @MarcoGorelli 

It seems like I'm viewing the use of operators as *less important* when they can be worked around (https://github.com/narwhals-dev/narwhals/pull/2920#issuecomment-3164931946)

I'm happy to concede on this and make getting operator-typing working for `ibis` the first goal here 🙂

This is a pretty minimal repro to get us started:

```py
import ibis

[...]

PR #2963, opened 2025-08-08
TYP: Use `ibis.Deferred` instead of `ibis.Value` for typing

MarcoGorelli:
This enables us to actually add arithmetic operators to `NativeExpr`, rather than having to mask those operations behind `operator.add` / `operator.truediv` (which just hide typing errors)

It also greatly simplifies the internal typing. For DuckDB we just have `duckdb.Expression`, for PySpark we just have `Column`, I think it's too confusing to have so many different types for Ibis. Especially given that (I'm repeating myself here) their types aren't public nor type-checked, they're already enough of a headache to deal with


narwhals/_ibis/expr.py
```diff
@@ -220,27 +222,27 @@ def func(df: IbisLazyFrame) -> Sequence[ir.IntegerScalar]:
...
-        def _std(expr: ir.NumericColumn, ddof: int) -> ir.Value:
+        def _std(expr: ir.Deferred, ddof: int) -> ir.Deferred:
             if ddof == 0:
                 return expr.std(how="pop")
             if ddof == 1:
                 return expr.std(how="sample")
             n_samples = expr.count()
             std_pop = expr.std(how="pop")
-            ddof_lit = cast("ir.IntegerScalar", ibis.literal(ddof))
+            ddof_lit = cast("ir.Deferred", ibis.literal(ddof))
```
MarcoGorelli, line 232: many of these just look like typing bugs (missing overloads?) in ibis

```python
In [26]: import ibis

In [27]: expr = ibis._['a']

In [28]: type(expr.std())
Out[28]: ibis.common.deferred.Deferred
```

dangotbanned:
@MarcoGorelli can you leave this open for a bit please?

I'm pretty sure `Deferred` is only useful for the operator-specific case - but is essentially an `Any` for typing 🤔

dangotbanned:
I understand that freely using operators is *desirable*, but for `ibis` it isn't safe to use them everywhere 😢 

## Example
```py
import ibis
import polars as pl

import narwhals as nw

data = {"a": ["A", "B", "A"], "b": [1, 2, 3], "c": [9, 2, 4], "d": [8, 7, 8]}
df_pl = pl.DataFrame(data)
df_ibis = ibis.memtable(df_pl)

# NOTE: Cast fixed in https://github.com/narwhals-dev/narwhals/pull/2944
nw_ibis = cast("nw.LazyFrame[ibis.Table]", nw.from_native(df_ibis))
```

`"b"` refers to an `ir.IntegerColumn`, so this is fine:
```py
>>> nw_ibis.select(nw.col("b") / nw.lit(1))
┌───────────────────────────────┐
|      Narwhals LazyFrame       |
|-------------------------------|
...
```

But, `"a"` will raise since the operator isn't defined:

```py
>>> nw_ibis.select(nw.col("a") / nw.lit(1))
File c:\Users\danie\Documents\GitHub\narwhals\narwhals\_sql\expr.py:350, in SQLExpr.__truediv__.<locals>.<lambda>(expr, other)
    349 def __truediv__(self, other: Self) -> Self:
--> [350](file:///C:/Users/danie/Documents/GitHub/narwhals/narwhals/_sql/expr.py:350)     return self._with_binary(lambda expr, other: expr.__truediv__(other), other)

AttributeError: 'StringColumn' object has no attribute '__truediv__'
```

MarcoGorelli:
same argument for `duckdb.Expression` / `pyspark.sql.Column` though, no?

dangotbanned:
> same argument for `duckdb.Expression` / `pyspark.sql.Column` though, no?

I don't think they error outside of a `collect` call like (https://github.com/narwhals-dev/narwhals/pull/2963#issuecomment-3175990958) do they?

MarcoGorelli:
thanks - closing then, as other PRs have improved the typing sitaution (thanks!)

dangotbanned:
> thanks - closing then, as other PRs have improved the typing sitaution (thanks!)

Thanks @MarcoGorelli!

Feel free to pile any `ibis` issues you have into #2960 and I'll do my best to help 😄

Issue #2858, opened 2025-07-20
docs: API Completeness page


FBruzzesi:
As mentioned in https://github.com/narwhals-dev/narwhals/pull/2839#issuecomment-3094499540, the API completeness page needs some love.

Currently some backends do not even show (ibis, dask), the namespace's never made it there and we need to figure out how to manage functionalities such as `hist` which from the narwhals level we call _other_ complaint methods instead of `hist` itself

MarcoGorelli:
yup, thanks for opening the discussion

i'd be fairly inclined to remove it for now, and put it back once it's in a better state

dangotbanned:
Very on board with improving this section, thanks for raising @FBruzzesi 🙂 

To kick things off, here's some *high level* open-ended questions we could think about:

1. Who is the audience for this page?
2. What problems are we aiming help with/solve for these users?
3. What do we like about the current page?
4. Are there any examples of other projects communicating similar info that we can learn from?

FBruzzesi:
Hey @dangotbanned 

1. Potentially any user of the library 😂 
2. I think in many cases adoption goes as "We already support X, and it would be cool to support also Y and Z". When that's the case, we make it very easy to see if we support everything for X, and to see to which backends this could be extended and/or what are the blockers. Disclaimer: This is how I used to use it 
3. I really like that for a given class, all methods and backends are in the same page
4. ibis and sqlframe both have something similar to display what's available for each backend, yet I prefer our style of having everything together and then API documentation on its own, rather than one page per backend (docs: [ibis backends](https://ibis-project.org/backends/athena), [sqlframe](https://sqlframe.readthedocs.io/en/stable/))

dangotbanned:
Thanks for all the detail @FBruzzesi!

I planned to just respond to the last point but it spiralled into all of them 😂 

# (4)
https://github.com/narwhals-dev/narwhals/issues/2858#issuecomment-3094807337

> ibis and sqlframe both have something similar to display what's available for each backend
> ...
> rather than one page per backend

I thought this was the case as well, but here's a little overview of what I found

## Ibis
I've just stumbled into the `ibis` ([operation support matrix](https://ibis-project.org/backends/support/matrix)) which seems interesting:

<details><summary><b>Show screenshot</b></summary>
<p>

<img width="1380" height="1362" alt="Image" src="https://github.com/user-attachments/assets/30184a7e-6a9c-4162-baf6-6e6041b93587" />

</p>
</details> 

[...]

PR #1960, opened 2025-02-07
chore: reduce duplication, implement more methods at the Narwhals level rather than at the compliant level, fix CI from Ibis release

MarcoGorelli:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #2967, opened 2025-08-10
refactor: Add `CompliantColumn`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Fell out of #2962
- https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3170943889
- https://github.com/narwhals-dev/narwhals/pull/2962#issuecomment-3171900789

## Checklist

- [ ] Tests added
- [x] Documented the changes

Creates a common parent protocol for `Compliant{Expr,Series}`, spec-ing their shared parts

The diff isn't huge here, but ~~it'll shrink~~ it shrinks `is_close` and allow us to implement more features in a similar way


narwhals/expr.py
```diff
@@ -188,177 +189,91 @@ def cast(self, dtype: IntoDType) -> Self:
...
             lambda plx: apply_n_ary_operation(
-                plx, lambda x, y: x == y, self, other, str_as_lit=True
+                plx, function, self, other, str_as_lit=str_as_lit
             ),
             ExprMetadata.from_binary_op(self, other),
         )
 
+    def __eq__(self, other: Self | Any) -> Self:  # type: ignore[override]
+        return self._with_binary(op.eq, other)
+
```
dangotbanned, line 199: Spotted this while looking into the rhs `__dunder__`s that are missing

On that front though, I'm pretty confused on why we have these defined for `CompliantSeries`, but not `CompliantExpr`

https://github.com/narwhals-dev/narwhals/blob/4d2c4ccaa47b335efc1abc3849dc1340a50451c0/narwhals/_compliant/expr.py#L880-L883

I assumed it was *all* rhs operators, but all of these are on `EagerExpr` (but not `CompliantExpr`) 🤔 

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/expr.py#L575

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/expr.py#L584

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/expr.py#L590

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/expr.py#L596

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/expr.py#L602


I think we should decide on a single place they're all defined - this is quite odd

narwhals/_compliant/column.py
```diff
@@ -0,0 +1,144 @@
...
+    def sqrt(self) -> Self: ...
+    def fill_null(
+        self,
+        value: Self | NonNestedLiteral,
+        strategy: FillNullStrategy | None,
+        limit: int | None,
+    ) -> Self: ...
+    def is_between(
+        self, lower_bound: Any, upper_bound: Any, closed: ClosedInterval
+    ) -> Self: ...
```
dangotbanned: In addition to `is_between`, I think we should move `is_duplicated` down to the compliant-level.

We can still keep these default impls in `CompliantColumn` - but I don't see a need to remove the *option* to implement more efficiently if we can per-backend

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/expr.py#L1467

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/expr.py#L1489

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/series.py#L1747

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/series.py#L1723

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/series.py#L1749

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/series.py#L1767

dangotbanned: `is_empty` missing at the compliant-level is creating more complexity for `hist`


https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_compliant/series.py#L477-L478


https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_arrow/series.py#L1106-L1113

https://github.com/narwhals-dev/narwhals/blob/9f7db3caad043b2d9742d0d4163eb3fafb67fb06/narwhals/_pandas_like/series.py#L1066-L1068

FBruzzesi: > is_empty missing at the compliant-level is creating more complexity for hist

`is_empty` and `is_empty_series` may return different values depending on the presence of nan and nulls. I don't think they have the same purpose. _Maybe_ the latter name is misleading 🥲

narwhals/_compliant/expr.py
```diff
@@ -946,6 +877,12 @@ def fn(names: Sequence[str]) -> Sequence[str]:
     def name(self) -> LazyExprNameNamespace[Self]:
         return LazyExprNameNamespace(self)
 
+    __radd__: not_implemented = not_implemented()
+    __rand__: not_implemented = not_implemented()
+    __rmul__: not_implemented = not_implemented()
+    __ror__: not_implemented = not_implemented()
+    is_between: not_implemented = not_implemented()
```
FBruzzesi: I have mixed feelings on this. It's technically correct, yet the implementation exists and it happens directly at the narwhals level 🤔

This is also the reason why right binary ops are all marked as `# pragma: no cover` for `EagerExpr`: they are all defined in `narwhals.Expr`. I imagine that they are defined again here to use them at the compliant level somewhere in the codebase.

narwhals/expr.py
FBruzzesi, line 1: Really nice one! 🎉

FBruzzesi (commented):
Thanks @dangotbanned this looks promising 🎉  I left a couple of minor comments

narwhals/_compliant/expr.py
dangotbanned: > I have mixed feelings on this. It's technically correct, yet the implementation exists and it happens directly at the narwhals level

>  I imagine that they are defined again here to use them at the compliant level somewhere in the codebase.

This is what I found confusing in (https://github.com/narwhals-dev/narwhals/pull/2967#discussion_r2265376753)

We do this **only for `Expr`**, but not for `Series`.
I'm suggesting that we either do it for both or neither.

I think the motivation for implementing things at the `narwhals`-level (#1960) is the same reason we'd implement it here. 
The only difference is that those changes came *before* we had the `_compliant` package at all 😄 

On that basis, I want these all to be implemented in *some* protocol(s) - which reduces the duplication and overall makes `Compliant{Expr,Series}` more consistent with eachother 🙂

narwhals/_compliant/column.py
dangotbanned: > Maybe the latter name is misleading

Ah well spotted!

That'll teach me for assuming it was mirroring `pl.Series` 😂

```py
import polars as pl

a, b, c = [], [None], [float("nan")]
>>> pl.Series(a).is_empty(), pl.Series(b).is_empty(), pl.Series(c).is_empty()
(True, False, False)
```

narwhals/_compliant/column.py
```diff
@@ -0,0 +1,151 @@
...
+        if closed == "left":
+            return (self >= lower_bound) & (self < upper_bound)
+        if closed == "right":
+            return (self > lower_bound) & (self <= upper_bound)
+        if closed == "none":
+            return (self > lower_bound) & (self < upper_bound)
+        return (self >= lower_bound) & (self <= upper_bound)
+
+    def is_duplicated(self) -> Self:
+        return ~self.is_unique()
```
dangotbanned: A nice side effect is these are now fixed in the docs (related #2858)

image


image

dangotbanned: We *could* fix `Series.rename` + `Series.shape` in the same way

- https://narwhals-dev.github.io/narwhals/api-completeness/series/

narwhals/_compliant/expr.py
dangotbanned: Resolved in (https://github.com/narwhals-dev/narwhals/pull/2967/commits/8c6099b7cb7d93cb14bea46194f65a2056392fa2), (https://github.com/narwhals-dev/narwhals/pull/2967/commits/ef63995fc4aa0a593f881e71e7fd234888328f1c)

narwhals/_compliant/column.py
dangotbanned: These were all in (https://github.com/narwhals-dev/narwhals/pull/2967/commits/ef63995fc4aa0a593f881e71e7fd234888328f1c), (https://github.com/narwhals-dev/narwhals/pull/2967/commits/3695b086449d7bf6beeda602afb1822bbf436882), (https://github.com/narwhals-dev/narwhals/pull/2967/commits/893e366262b6fb915824040301e6f3e07d9d033d)

narwhals/_compliant/column.py
FBruzzesi: I think it would be a really nice side effect. I know we are not too invested at the moment with the api completeness, but these changes are low hanging fruit, and might simplify a lot the logic there (api completeness) without the need of handling additional special cases

FBruzzesi (commented):
Thanks @dangotbanned - I am pretty much onboarded with these changes. I will wait for @MarcoGorelli approval on this one though!

Just for context, are you considering this PR a requirement for #2962 or viceversa or neither 😂?

dangotbanned:
> Thanks @dangotbanned - I am pretty much onboarded with these changes. I will wait for @MarcoGorelli approval on this one though!
> Just for context, are you considering this PR a requirement for https://github.com/narwhals-dev/narwhals/pull/2962 or viceversa or neither 😂?

Well we've done it this way round now regardless 😄 

([refactor: Move `is_close` to `CompliantColumn`](https://github.com/narwhals-dev/narwhals/pull/2967/commits/d00b12cc69039b98e865151d44c086c808ff51d6))

narwhals/_compliant/column.py
dangotbanned: > We could fix `Series.rename` + `Series.shape` in the same way

Thinking back, I'm not super concerned about these two for now

MarcoGorelli:
> Thanks @dangotbanned - I am pretty much onboarded with these changes. I will wait for @MarcoGorelli approval on this one though!

from a look this seems good, i'm running out of time for the day but if you're happy with it feel free to ship it, thanks both!

PR #2969, opened 2025-08-10
feat: Add `DType.base_type`

dangotbanned:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Should make this easier as well (#2486)

## Checklist

- [x] Tests added
- [x] Documented the changes

Incredibly simple feature, but made a big refactor possible (starting from 307047099f22d720935101714312323788003972)

Based on [`polars.DataType.base_type`](https://github.com/pola-rs/polars/blob/34595af53dbdcbde52ab062746a42aa725e321a4/py-polars/polars/datatypes/classes.py#L145-L159)

See (https://github.com/narwhals-dev/narwhals/pull/2969#discussion_r2266773501) for details

### DIFF
--- a/narwhals/_ibis/utils.py
+++ b/narwhals/_ibis/utils.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 import ibis
 import ibis.expr.datatypes as ibis_dtypes
 
-from narwhals._utils import isinstance_or_issubclass
+from narwhals._utils import Version, isinstance_or_issubclass
 
 if TYPE_CHECKING:
     from collections.abc import Mapping
@@ -20,7 +20,6 @@ if TYPE_CHECKING:
     from narwhals._duration import IntervalUnit
     from narwhals._ibis.dataframe import IbisLazyFrame
     from narwhals._ibis.expr import IbisExpr
-    from narwhals._utils import Version
     from narwhals.dtypes import DType
     from narwhals.typing import IntoDType, PythonLiteral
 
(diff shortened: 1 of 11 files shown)


narwhals/dtypes.py
```diff
@@ -73,13 +73,13 @@ def base_type(cls) -> type[Self]:
         Examples:
             >>> import narwhals as nw
             >>> nw.Datetime("us").base_type()
-            narwhals.dtypes.Datetime
+            <class 'narwhals.dtypes.Datetime'>
```
dangotbanned: For reference, we *could* have the nicer repr from `polars` - but would need to add a metaclass for `DType`.
I think this is more user friendly, given that we tret the types and objects and interchangeable in most cases

```py
    @classmethod
    def base_type(cls) -> DataTypeClass:
        """
        Return this DataType's fundamental/root type class.

        Examples
        --------
        >>> pl.Datetime("ns").base_type()
        Datetime
        >>> pl.List(pl.Int32).base_type()
        List
        >>> pl.Struct([pl.Field("a", pl.Int64), pl.Field("b", pl.Boolean)]).base_type()
        Struct
        """
        return cls

```

Fair enough though if *not* having this was an intentional choice

- https://github.com/pola-rs/polars/blob/97fb4c8a49f259b3d614a6f1f3c58994bda5d7eb/py-polars/polars/datatypes/classes.py#L145-L159
- https://github.com/pola-rs/polars/blob/97fb4c8a49f259b3d614a6f1f3c58994bda5d7eb/py-polars/polars/datatypes/classes.py#L51-L58


---

A tweak that comes to mind is this:

```py
import narwhals as nw

>>> nw.List
nw.List

>>> nw.List(nw.Int32)
nw.List(nw.Int32)
```

And then we can qualify the class name a bit more for v1's `Datetime`, `Duration` and `Enum` only

narwhals/_ibis/utils.py
```diff
@@ -174,54 +173,37 @@ def is_floating(obj: IbisDataType) -> TypeIs[ibis_dtypes.Floating]:
...
-        return ibis_dtypes.UInt8()
-    if isinstance_or_issubclass(dtype, dtypes.String):
-        return ibis_dtypes.String()
-    if isinstance_or_issubclass(dtype, dtypes.Boolean):
-        return ibis_dtypes.Boolean()
-    if isinstance_or_issubclass(dtype, dtypes.Categorical):
-        msg = "Categorical not supported by Ibis"
-        raise NotImplementedError(msg)
+    if ibis_type := NW_TO_IBIS_DTYPES.get(dtype.base_type()):
+        return ibis_type
```
dangotbanned: Not sure if the diff paints a clear enough picture

## Example (`String() | String`)
On main, we need to go through **13x** failed `isinstance_or_issubclass` calls, before finally reaching the correct one.

Inside that, every time **2x** `isinstance` calls are made

```py
_ibis.utils.narwhals_to_native_dtype(nw.String(), ...)
_ibis.utils.narwhals_to_native_dtype(nw.String, ...)
```

https://github.com/narwhals-dev/narwhals/blob/9e994f3e03153932d23a6a571afd12fe9f44740e/narwhals/_utils.py#L843-L848

So either variation of `nw.String` currently always requires **26x** failed `isinstance` calls, **+2** once we get to the correct case

## Optimization
Now most cases are now handled in a single dict lookup (**16** in total for `ibis`) 🥳 

We can do this for *at least* `polars`, `pyarrow`, `ibis`, `duckdb` - and probably with modifications in the other backends

narwhals/_pandas_like/utils.py
```diff
@@ -537,8 +510,9 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912, PLR0915
         )
     if isinstance_or_issubclass(dtype, dtypes.Date):
         try:
-            import pyarrow as pa  # ignore-banned-import
+            import pyarrow as pa  # ignore-banned-import  # noqa: F401
         except ModuleNotFoundError:  # pragma: no cover
+            # BUG: Never re-raised?
```
dangotbanned, line 516: This looked suspect to me 🤔 

https://github.com/narwhals-dev/narwhals/blob/cf83c9d3a87658eec27342647dc036b3126b641f/narwhals/_pandas_like/utils.py#L511-L518

narwhals/_pandas_like/utils.py
```diff
@@ -537,8 +511,9 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912, PLR0915
...
             msg = "'pyarrow>=13.0.0' is required for `Date` dtype."
         return "date32[pyarrow]"
```
MarcoGorelli:
```suggestion
            msg = "'pyarrow>=13.0.0' is required for `Date` dtype."
            raise ModuleNotFound(msg)
        return "date32[pyarrow]"
```

MarcoGorelli (approved):
thanks @dangotbanned , really nice

narwhals/_pandas_like/utils.py
```diff
@@ -537,9 +511,11 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912, PLR0915
         )
     if isinstance_or_issubclass(dtype, dtypes.Date):
         try:
-            import pyarrow as pa  # ignore-banned-import
+            import pyarrow as pa  # ignore-banned-import  # noqa: F401
         except ModuleNotFoundError:  # pragma: no cover
+            # BUG: Never re-raised?
             msg = "'pyarrow>=13.0.0' is required for `Date` dtype."
+            raise ModuleNotFound(msg)
```
MarcoGorelli:
```suggestion
            raise ModuleNotFoundError(msg)
```

dangotbanned:
Thanks @MarcoGorelli 😍 

also 😂 at (https://github.com/narwhals-dev/narwhals/pull/2969/commits/2967c2d77b2adfb9bb83b33e227d444f2a925c44)

PR #2971, opened 2025-08-11
typing: expose IntoBackend

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related comment https://github.com/narwhals-dev/narwhals/pull/1914#issuecomment-2629460801
- Closes #1917

## Checklist

- [ ] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -70,6 +70,7 @@ if TYPE_CHECKING:
     from narwhals.group_by import GroupBy, LazyGroupBy
     from narwhals.typing import (
         AsofJoinStrategy,
+        IntoBackend,
         IntoDataFrame,
         IntoExpr,
         IntoFrame,
@@ -473,7 +474,7 @@ class DataFrame(BaseFrame[DataFrameT]):
 
     @classmethod
     def from_arrow(
-        cls, native_frame: IntoArrowTable, *, backend: ModuleType | Implementation | str
+        cls, native_frame: IntoArrowTable, *, backend: IntoBackend
     ) -> DataFrame[Any]:
         """Construct a DataFrame from an object which supports the PyCapsule Interface.
 
@@ -533,7 +534,7 @@ class DataFrame(BaseFrame[DataFrameT]):
         data: Mapping[str, Any],
         schema: IntoSchema | None = None,
         *,
-        backend: ModuleType | Implementation | str | None = None,
+        backend: IntoBackend | None = None,
     ) -> DataFrame[Any]:
         """Instantiate DataFrame from dictionary.
 
@@ -595,7 +596,7 @@ class DataFrame(BaseFrame[DataFrameT]):
         data: _2DArray,
         schema: IntoSchema | Sequence[str] | None = None,
         *,
-        backend: ModuleType | Implementation | str,
+        backend: IntoBackend,
     ) -> DataFrame[Any]:
         """Construct a DataFrame from a NumPy ndarray.
 
@@ -724,9 +725,7 @@ class DataFrame(BaseFrame[DataFrameT]):
         pa_table = self.to_arrow()
         return pa_table.__arrow_c_stream__(requested_schema=requested_schema)  # type: ignore[no-untyped-call]
 
-    def lazy(
-        self, backend: ModuleType | Implementation | str | None = None
-    ) -> LazyFrame[Any]:
+    def lazy(self, backend: IntoBackend | None = None) -> LazyFrame[Any]:
         """Restrict available API methods to lazy-only ones.
 
         If `backend` is specified, then a conversion between different backends
@@ -2470,7 +2469,7 @@ class LazyFrame(BaseFrame[FrameT]):
         raise TypeError(msg)
 
     def collect(
-        self, backend: ModuleType | Implementation | str | None = None, **kwargs: Any
+        self, backend: IntoBackend | None = None, **kwargs: Any
     ) -> DataFrame[Any]:
         r"""Materialize this LazyFrame into a DataFrame.
 
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -40,6 +40,7 @@ if TYPE_CHECKING:
     from narwhals.typing import (
         ClosedInterval,
         FillNullStrategy,
+        IntoBackend,
         IntoDType,
         NonNestedLiteral,
         NumericLiteral,
@@ -99,7 +100,7 @@ class Series(Generic[IntoSeriesT]):
         values: _1DArray,
         dtype: IntoDType | None = None,
         *,
-        backend: ModuleType | Implementation | str,
+        backend: IntoBackend,
     ) -> Series[Any]:
         """Construct a Series from a NumPy ndarray.
 
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
     from typing_extensions import TypeAlias
 
     from narwhals import dtypes
+    from narwhals._utils import Implementation
     from narwhals.dataframe import DataFrame, LazyFrame
     from narwhals.expr import Expr
     from narwhals.schema import Schema
@@ -98,6 +99,39 @@ if TYPE_CHECKING:
         def Time(self) -> type[dtypes.Time]: ...
         @property
         def Binary(self) -> type[dtypes.Binary]: ...
+
+    _Polars: TypeAlias = Literal["polars"]
+    _Arrow: TypeAlias = Literal["pyarrow"]
+    _Dask: TypeAlias = Literal["dask"]
+    _DuckDB: TypeAlias = Literal["duckdb"]
+    _PandasLike: TypeAlias = Literal["pandas", "cudf", "modin"]
+    _Ibis: TypeAlias = Literal["ibis"]
+    _SparkLike: TypeAlias = Literal["pyspark", "sqlframe", "pyspark[connect]"]
+    _EagerOnly: TypeAlias = "_PandasLike | _Arrow"
+    _EagerAllowed: TypeAlias = "_Polars | _EagerOnly"
+    _LazyOnly: TypeAlias = "_SparkLike | _Dask | _DuckDB | _Ibis"
+    _LazyAllowed: TypeAlias = "_Polars | _LazyOnly"
+
+    Polars: TypeAlias = Literal[_Polars, Implementation.POLARS]
+    Arrow: TypeAlias = Literal[_Arrow, Implementation.PYARROW]
+    Dask: TypeAlias = Literal[_Dask, Implementation.DASK]
+    DuckDB: TypeAlias = Literal[_DuckDB, Implementation.DUCKDB]
+    Ibis: TypeAlias = Literal[_Ibis, Implementation.IBIS]
+    PandasLike: TypeAlias = Literal[
+        _PandasLike, Implementation.PANDAS, Implementation.CUDF, Implementation.MODIN
+    ]
+    SparkLike: TypeAlias = Literal[
+        _SparkLike,
+        Implementation.PYSPARK,
+        Implementation.SQLFRAME,
+        Implementation.PYSPARK_CONNECT,
+    ]
+    EagerOnly: TypeAlias = "PandasLike | Arrow"
+    EagerAllowed: TypeAlias = "EagerOnly | Polars"
+    LazyOnly: TypeAlias = "SparkLike | Dask | DuckDB | Ibis"
+    LazyAllowed: TypeAlias = "LazyOnly | Polars"
+
+    BackendName: TypeAlias = "_EagerAllowed | _LazyAllowed"
 
 
 IntoExpr: TypeAlias = Union["Expr", str, "Series[Any]"]
@@ -388,6 +422,43 @@ Examples:
     |│ 2   ┆ [5, 0]   │|
     |│ 3   ┆ [6, 0]   │|
     |└─────┴──────────┘|
+    └──────────────────┘
+"""
+
+IntoBackend: TypeAlias = "BackendName | Implementation | ModuleType"
+"""Anything that can be converted into a Narwhals Implementation.
+
+It can be specified as:
+
+- a string (backend name), such as: `"pandas"`, `"pyarrow"`, `"modin"`, `"cudf"`, etc..
+- an Implementation, such as: `Implementation.POLARS`, `Implementation.DUCKDB`, `Implementation.PYSPARK`, etc..
+- a python module, such as `dask`, `ibis`, `sqlframe`, etc..
+
+Examples:
+    >>> import dask.dataframe as dd
+    >>> from sqlframe.duckdb import DuckDBSession
+    >>> import narwhals as nw
+    >>>
+    >>> nw.scan_parquet("file.parquet", backend="dask").collect()  # doctest:+SKIP
+    ┌──────────────────┐
+    |Narwhals DataFrame|
+    |------------------|
+    |        a   b     |
+    |     0  1   4     |
+    |     1  2   5     |
+    └──────────────────┘
+    >>> nw.scan_parquet(
+    ...     "file.parquet", backend=Implementation.SQLFRAME, session=DuckDBSession()
+    ... ).collect()  # doctest:+SKIP
+    ┌──────────────────┐
+    |Narwhals DataFrame|
+    |------------------|
+    |  pyarrow.Table   |
+    |  a: int64        |
+    |  b: int64        |
+    |  ----            |
+    |  a: [[1,2]]      |
+    |  b: [[4,5]]      |
     └──────────────────┘
 """
 
--- a/docs/api-reference/typing.md
+++ b/docs/api-reference/typing.md
@@ -17,6 +17,7 @@ Narwhals comes fully statically typed. In addition to `nw.DataFrame`, `nw.Expr`,
         - IntoFrameT
         - IntoSeries
         - IntoSeriesT
+        - IntoBackend
         - IntoDType
         - IntoSchema
         - SizeUnit
--- a/narwhals/_duckdb/dataframe.py
+++ b/narwhals/_duckdb/dataframe.py
@@ -51,7 +51,12 @@ if TYPE_CHECKING:
     from narwhals.dataframe import LazyFrame
     from narwhals.dtypes import DType
     from narwhals.stable.v1 import DataFrame as DataFrameV1
-    from narwhals.typing import AsofJoinStrategy, JoinStrategy, LazyUniqueKeepStrategy
+    from narwhals.typing import (
+        AsofJoinStrategy,
+        IntoBackend,
+        JoinStrategy,
+        LazyUniqueKeepStrategy,
+    )
 
 
 class DuckDBLazyFrame(
@@ -129,7 +134,7 @@ class DuckDBLazyFrame(
             yield col(name)
 
     def collect(
-        self, backend: ModuleType | Implementation | str | None, **kwargs: Any
+        self, backend: IntoBackend | None, **kwargs: Any
     ) -> CompliantDataFrameAny:
         if backend is None or backend is Implementation.PYARROW:
             from narwhals._arrow.dataframe import ArrowDataFrame
--- a/narwhals/_ibis/dataframe.py
+++ b/narwhals/_ibis/dataframe.py
@@ -37,7 +37,12 @@ if TYPE_CHECKING:
     from narwhals.dataframe import LazyFrame
     from narwhals.dtypes import DType
     from narwhals.stable.v1 import DataFrame as DataFrameV1
-    from narwhals.typing import AsofJoinStrategy, JoinStrategy, LazyUniqueKeepStrategy
+    from narwhals.typing import (
+        AsofJoinStrategy,
+        IntoBackend,
+        JoinStrategy,
+        LazyUniqueKeepStrategy,
+    )
 
     JoinPredicates: TypeAlias = "Sequence[ir.BooleanColumn] | Sequence[str]"
 
@@ -101,7 +106,7 @@ class IbisLazyFrame(
             yield self.native[name]
 
     def collect(
-        self, backend: ModuleType | Implementation | str | None, **kwargs: Any
+        self, backend: IntoBackend | None, **kwargs: Any
     ) -> CompliantDataFrameAny:
         if backend is None or backend is Implementation.PYARROW:
             from narwhals._arrow.dataframe import ArrowDataFrame
--- a/narwhals/_namespace.py
+++ b/narwhals/_namespace.py
@@ -31,7 +31,6 @@ from narwhals.dependencies import (
 
 if TYPE_CHECKING:
     from collections.abc import Collection, Sized
-    from types import ModuleType
     from typing import ClassVar
 
     import duckdb
@@ -50,45 +49,25 @@ if TYPE_CHECKING:
     from narwhals._polars.namespace import PolarsNamespace
     from narwhals._spark_like.dataframe import SQLFrameDataFrame
     from narwhals._spark_like.namespace import SparkLikeNamespace
-    from narwhals.typing import DataFrameLike, NativeFrame, NativeLazyFrame, NativeSeries
+    from narwhals.typing import (
+        Arrow,
+        Dask,
+        DataFrameLike,
+        DuckDB,
+        EagerAllowed,
+        Ibis,
+        IntoBackend,
+        NativeFrame,
+        NativeLazyFrame,
+        NativeSeries,
+        PandasLike,
+        Polars,
+        SparkLike,
+    )
 
     T = TypeVar("T")
 
     _Guard: TypeAlias = "Callable[[Any], TypeIs[T]]"
-
-    _Polars: TypeAlias = Literal["polars"]
-    _Arrow: TypeAlias = Literal["pyarrow"]
-    _Dask: TypeAlias = Literal["dask"]
-    _DuckDB: TypeAlias = Literal["duckdb"]
-    _PandasLike: TypeAlias = Literal["pandas", "cudf", "modin"]
-    _Ibis: TypeAlias = Literal["ibis"]
-    _SparkLike: TypeAlias = Literal["pyspark", "sqlframe", "pyspark[connect]"]
-    _EagerOnly: TypeAlias = "_PandasLike | _Arrow"
-    _EagerAllowed: TypeAlias = "_Polars | _EagerOnly"
-    _LazyOnly: TypeAlias = "_SparkLike | _Dask | _DuckDB | _Ibis"
-    _LazyAllowed: TypeAlias = "_Polars | _LazyOnly"
-
-    Polars: TypeAlias = Literal[_Polars, Implementation.POLARS]
-    Arrow: TypeAlias = Literal[_Arrow, Implementation.PYARROW]
-    Dask: TypeAlias = Literal[_Dask, Implementation.DASK]
-    DuckDB: TypeAlias = Literal[_DuckDB, Implementation.DUCKDB]
-    Ibis: TypeAlias = Literal[_Ibis, Implementation.IBIS]
-    PandasLike: TypeAlias = Literal[
-        _PandasLike, Implementation.PANDAS, Implementation.CUDF, Implementation.MODIN
-    ]
-    SparkLike: TypeAlias = Literal[
-        _SparkLike,
-        Implementation.PYSPARK,
-        Implementation.SQLFRAME,
-        Implementation.PYSPARK_CONNECT,
-    ]
-    EagerOnly: TypeAlias = "PandasLike | Arrow"
-    EagerAllowed: TypeAlias = "EagerOnly | Polars"
-    LazyOnly: TypeAlias = "SparkLike | Dask | DuckDB | Ibis"
-    LazyAllowed: TypeAlias = "LazyOnly | Polars"
-
-    BackendName: TypeAlias = "_EagerAllowed | _LazyAllowed"
-    IntoBackend: TypeAlias = "BackendName | Implementation | ModuleType"
 
     EagerAllowedNamespace: TypeAlias = "Namespace[PandasLikeNamespace] | Namespace[ArrowNamespace] | Namespace[PolarsNamespace]"
     EagerAllowedImplementation: TypeAlias = Literal[
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -44,7 +44,7 @@ if TYPE_CHECKING:
     from narwhals._utils import Version, _LimitedContext
     from narwhals.dataframe import LazyFrame
     from narwhals.dtypes import DType
-    from narwhals.typing import JoinStrategy, LazyUniqueKeepStrategy
+    from narwhals.typing import IntoBackend, JoinStrategy, LazyUniqueKeepStrategy
 
     SQLFrameDataFrame = BaseDataFrame[Any, Any, Any, Any, Any]
 
@@ -198,7 +198,7 @@ class SparkLikeLazyFrame(
         return self._cached_columns
 
     def _collect(
-        self, backend: ModuleType | Implementation | str | None, **kwargs: Any
+        self, backend: IntoBackend | None, **kwargs: Any
     ) -> CompliantDataFrameAny:
         if backend is Implementation.PANDAS:
             from narwhals._pandas_like.dataframe import PandasLikeDataFrame
@@ -236,7 +236,7 @@ class SparkLikeLazyFrame(
         raise ValueError(msg)  # pragma: no cover
 
     def collect(
-        self, backend: ModuleType | Implementation | str | None, **kwargs: Any
+        self, backend: IntoBackend | None, **kwargs: Any
     ) -> CompliantDataFrameAny:
         if self._implementation.is_pyspark_connect():
             try:
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -47,6 +47,7 @@ if TYPE_CHECKING:
     from narwhals.typing import (
         ConcatMethod,
         FrameT,
+        IntoBackend,
         IntoDType,
         IntoExpr,
         IntoSchema,
@@ -167,11 +168,7 @@ def concat(items: Iterable[FrameT], *, how: ConcatMethod = "vertical") -> FrameT
 
 
 def new_series(
-    name: str,
-    values: Any,
-    dtype: IntoDType | None = None,
-    *,
-    backend: ModuleType | Implementation | str,
+    name: str, values: Any, dtype: IntoDType | None = None, *, backend: IntoBackend
 ) -> Series[Any]:
     """Instantiate Narwhals Series from iterable (e.g. list or array).
 
@@ -212,11 +209,7 @@ def new_series(
 
 
 def _new_series_impl(
-    name: str,
-    values: Any,
-    dtype: IntoDType | None = None,
-    *,
-    backend: ModuleType | Implementation | str,
+    name: str, values: Any, dtype: IntoDType | None = None, *, backend: IntoBackend
 ) -> Series[Any]:
     implementation = Implementation.from_backend(backend)
     if is_eager_allowed(implementation):
@@ -246,7 +239,7 @@ def from_dict(
     data: Mapping[str, Any],
     schema: IntoSchema | None = None,
     *,
-    backend: ModuleType | Implementation | str | None = None,
+    backend: IntoBackend | None = None,
     native_namespace: ModuleType | None = None,  # noqa: ARG001
 ) -> DataFrame[Any]:
     """Instantiate DataFrame from dictionary.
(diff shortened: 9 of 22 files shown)


narwhals/series.py
```diff
@@ -99,7 +100,7 @@ def from_numpy(
         values: _1DArray,
         dtype: IntoDType | None = None,
         *,
-        backend: ModuleType | Implementation | str,
+        backend: IntoBackend,
```
dangotbanned: Just picking out a more obvious one, but this would apply in a lot of places

The two main motivators for me when I was writing (https://github.com/narwhals-dev/narwhals/issues/2116#issuecomment-3122104794) were:

1. The possibility for writing `@overload`(s)
    i. E.g. we could provide a better return type here than `Series[Any]`
2. Using typing to document what works at runtime

I'm happy for **1** to be a future PR, but **2** we can definitely think about here 


For example, here `backend="ibis"` doesn't warn the user in the same way the other failure cases do


https://github.com/narwhals-dev/narwhals/blob/ebdf1c1a82e4af5c549b45e779fe9d42bb38fc7c/tests/series_only/from_numpy_test.py#L67-L82

So for this one maybe an alias named `IntoEagerBackend` would be helpful?

We *could* reuse the term *eager-allowed* - but a short user-facing alias is nice 😄

FBruzzesi: I am on it 😉

dangotbanned: This is towards the same goal as (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2266823489), but using a generic alias instead

<p>

```diff
diff --git a/narwhals/typing.py b/narwhals/typing.py
index 5f6ed011b..bb829d61d 100644
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from types import ModuleType
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, Union
 
 from narwhals._compliant import CompliantDataFrame, CompliantLazyFrame, CompliantSeries
@@ -8,7 +9,6 @@ if TYPE_CHECKING:
     import datetime as dt
     from collections.abc import Iterable, Mapping, Sequence, Sized
     from decimal import Decimal
-    from types import ModuleType
 
     import numpy as np
     from typing_extensions import TypeAlias
@@ -131,8 +131,6 @@ if TYPE_CHECKING:
     LazyOnly: TypeAlias = "SparkLike | Dask | DuckDB | Ibis"
     LazyAllowed: TypeAlias = "LazyOnly | Polars"
 
-    BackendName: TypeAlias = "_EagerAllowed | _LazyAllowed"
-
 
 IntoExpr: TypeAlias = Union["Expr", str, "Series[Any]"]
 """Anything which can be converted to an expression.
@@ -425,6 +423,13 @@ Examples:
     └──────────────────┘
 """
 
+
+Backend: TypeAlias = "EagerAllowed | LazyAllowed"
+BackendName: TypeAlias = "_EagerAllowed | _LazyAllowed"
+_BT = TypeVar("_BT", bound=Backend)
+IntoBackend2: TypeAlias = Union[_BT, ModuleType]
+IntoBackendAny: TypeAlias = IntoBackend2[Backend]
+
 IntoBackend: TypeAlias = "BackendName | Implementation | ModuleType"
 """Anything that can be converted into a Narwhals Implementation.
 

```


</p>
</details> 

<details><summary><b>Show example defs</b></summary>
<p>

```py
from typing import Any  # noqa: F811, I001

import narwhals.typing as nwt  # noqa: F811


def from_native(obj: Any, backend: nwt.IntoBackendAny):  # noqa: ANN201, ARG001
    return backend


def from_iterable(obj: Any, backend: nwt.IntoBackend2[nwt.EagerAllowed]):  # noqa: ANN201, ARG001
    return backend


def scan_parquet(obj: Any, backend: nwt.IntoBackend2[nwt.LazyAllowed]):  # noqa: ANN201, ARG001
    return backend


from_native(1, backend="pandas")
from_native(1, backend="ibis")
from_native(1, backend="polars")
from_native(1, backend="numpy")

from_iterable(1, backend="pandas")
from_iterable(1, backend="ibis")
from_iterable(1, backend="polars")
from_iterable(1, backend="numpy")

scan_parquet(1, backend="pandas")
scan_parquet(1, backend="ibis")
scan_parquet(1, backend="polars")
scan_parquet(1, backend="numpy")
```

</p>
</details>

I think this would cover most cases, but anywhere else we allow/disallow specific backends - having it be generic could be handy 🙂  

image

tests/from_numpy_test.py
```diff
@@ -55,7 +55,7 @@ def test_from_numpy_schema_notvalid(constructor_eager: ConstructorEager) -> None
 def test_from_numpy_non_eager() -> None:
     pytest.importorskip("duckdb")
     with pytest.raises(ValueError, match="lazy-only"):
-        nw.from_numpy(arr, backend="duckdb")
+        nw.from_numpy(arr, backend="duckdb")  # type: ignore[arg-type]
```
dangotbanned, line 58: 😍😍😍

FBruzzesi:
@dangotbanned I will keep working on this after dinner. I will also go and fix the altair issue 🙏🏼

Update: Regarding the failing test for altair, the main issue seems to be the following overload returning a `LiteralString`:

```py
@overload
def _import_guarded(req: Requirement, /) -> LiteralString: ...
```

([source](https://github.com/vega/altair/blob/d30d030a89e2e2e72c6f5a4d497968b7dfb743bf/altair/datasets/_reader.py#L533C1-L534C1))

which is then fed into `nw.Implementation.from_backend`, namely:

```py
def _into_implementation(
    backend: _NwSupport | _PandasAny | Requirement, /
) -> nw.Implementation:
    primary = _import_guarded(backend)
    impl = nw.Implementation.from_backend(primary)
    if impl is not nw.Implementation.UNKNOWN:
        return impl
    msg = f"Package {primary!r} is not supported by `narwhals`."
    raise ValueError(msg)
```

([source](https://github.com/vega/altair/blob/d30d030a89e2e2e72c6f5a4d497968b7dfb743bf/altair/datasets/_reader.py#L495C1-L503C26))

narwhals/series.py
FBruzzesi: @dangotbanned I think that https://github.com/narwhals-dev/narwhals/pull/2971/commits/2ad79a92c6f6088bb83cf2a917fcabf0275ee820 is the closest to what we need for 

> The possibility for writing @overload(s)
> i. E.g. we could provide a better return type here than Series[Any]

although it would add _a lot_ (really a lot) of overloads.


But now for each IntoBackend input value, you get the correct implementation as output (as long as it's not the module itself)

**t.py**

```py
import pyarrow as pa

from narwhals import Implementation

impl1 = Implementation.from_backend("pandas")
impl2 = Implementation.from_backend(Implementation.POLARS)
impl3 = Implementation.from_backend(pa)

reveal_type(impl1)
reveal_type(impl2)
reveal_type(impl3)
```

`mypy t.py`

```terminal
mypy t.py
t.py:9: note: Revealed type is "Literal[narwhals._utils.Implementation.PANDAS]"
t.py:10: note: Revealed type is "Literal[narwhals._utils.Implementation.POLARS]"
t.py:11: note: Revealed type is "Union[Literal[narwhals._utils.Implementation.POLARS], Literal[narwhals._utils.Implementation.PYARROW], Literal[narwhals._utils.Implementation.PANDAS], Literal[narwhals._utils.Implementation.MODIN], Literal[narwhals._utils.Implementation.CUDF]]"
```

FBruzzesi:
After https://github.com/narwhals-dev/narwhals/pull/2971/commits/f14ecd76d29844a079f37575edb6ac69e27640f9, the remaining issue in altair can be fixed with the following changes in altair itself: [altair fork branch](https://github.com/vega/altair/compare/main...FBruzzesi:altair:typing/narwhals-reader)

narwhals/dataframe.py
```diff
@@ -2470,7 +2470,7 @@ def __getitem__(self, item: str | slice) -> NoReturn:
         raise TypeError(msg)
 
     def collect(
-        self, backend: ModuleType | Implementation | str | None = None, **kwargs: Any
+        self, backend: IntoEagerBackend | None = None, **kwargs: Any
```
dangotbanned: I just tried removing all of the overloads from `Implementation.from_{backend,string}` and only here produced an error for `mypy`

```py
narwhals/dataframe.py:2545: error: 
Argument "backend" to "collect" of "CompliantLazyFrame" has incompatible type "Implementation | None"; expected
"Literal[Implementation.POLARS, Implementation.PYARROW, Implementation.PANDAS, Implementation.MODIN, Implementation.CUDF] | None"  [arg-type]

                self._compliant_frame.collect(backend=eager_backend, **kwargs), level="full"
                                                      ^~~~~~~~~~~~~
```

This would be the kinda place where a generic alias (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2266954326) would be helpful.

`IntoEagerBackend` says we can pass in `"modin"` or `"cudf"`, **but** we would actually reject them at runtime:

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/dataframe.py#L2536-L2540

So a more accurate - but longer if done inline - spelling would be:

```py
    def collect(
        self,
        backend: Literal[
            "polars",
            "pyarrow",
            "pandas",
            Implementation.POLARS,
            Implementation.PYARROW,
            Implementation.PANDAS,
        ]
        | ModuleType
        | None = None,
        **kwargs: Any,
    ) -> DataFrame[Any]:
```


Making that change was a bit of a wrecking ball 😂 

```py
Found 11 errors in 6 files (checked 421 source files)
```

But a lot of those are just other places `IntoEagerBackend` is passed into `LazyFrame.collect`

### Edit
Adding the `Implementation` overloads back only brings that number down to:

```py
Found 10 errors in 5 files (checked 421 source files)
```

So *I think* they're masking an issue `mypy` has not understanding an `in {...}` guard - like [`pyright` can](https://microsoft.github.io/pyright/#/type-concepts-advanced?id=type-guards)

---

Let me know if any of that didn't make sense 🙂

FBruzzesi: Hey @dangotbanned thanks again for your review!

> I just tried removing all of the overloads from Implementation.from_{backend,string}

I hope I am not missing something obvious, but wouldn't that break the mapping happening as in https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2267823730? And as a consequence, how could we pin down the generic of the `Data|LazyFrame`?

> This would be the kinda place where a generic alias (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2266954326) would be helpful.

This idea is not very clear to me, I am sorry 🙈

> IntoEagerBackend says we can pass in "modin" or "cudf", but we would actually reject them at runtime:

This is true, yet I see it more as a shortcoming of the `collect` itself. In my opinion `collect` should allow to materialize a lazyframe into any eager backend.

---

Also it's a bit unclear to me what's the concrete suggestion moving forward 🤔

dangotbanned: > I hope I am not missing something obvious, but wouldn't that break the mapping happening as in https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2267823730? 

*Possibly* the bit that I haven't said out loud is that `Namespace` already has the overloads for this

<details><summary><code>Namespace.from_backend</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/c441460cc091c33ddb08758ba01cfd895ff0f83f/narwhals/_namespace.py#L203-L244

</p>
</details> 

The important part is the bottom case will allow the full range of inputs, without narrowing to a specific backend.
Our target for a given `Implementation` is mostly this entrypoint, which *allows* `Implementation` (not just `Implementation.<backend>`)

```py
@overload
    @classmethod
    def from_backend(
        cls, backend: IntoBackend, /
    ) -> Namespace[CompliantNamespaceAny]: ...
```

---

> And as a consequence, how could we pin down the generic of the `Data|LazyFrame`?

I'll try to work through an example of what we need for that, maybe it'll do better than my words 😄:

## Example (`Series.from_numpy`)
The overloads added to `Implementation.from_backend` mean that we get some narrowing from **(1)** `backend`  -> **(2)** `implementation`.

This is good! 🎉 

*But* we have this already in **(3)** `is_eager_allowed`.

image


If we want to annotate `Series[Any]` to match `backend`, the first step is to switch to using `compliant.to_narwhals`.

Here is what that looks like if I let `pyright` infer the return type.

The point I'm trying to make is that **we'd *still* need to add `@overload`s here to resolve this `Union`**.

image

To be 100% clear, (https://github.com/narwhals-dev/narwhals/pull/2971/files#diff-d379de34a0e309244b9a412352a0c456fb6e395460b63e7bd635304c282fe817) is correct!

But AFAICT, they aren't resolving an issue - so I feel like I need to be the bad guy and ask *maybe less typing?* 😢

dangotbanned:
> Also it's a bit unclear to me what's the concrete suggestion moving forward 🤔

Thought I'd separate this from the explanation in (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2270597447) in case there are any follow-ups in that thread 🙂 

Out of all the places `IntoEagerBackend` is used, I think the following should be changed:

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/dataframe.py#L2472-L2474

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/functions.py#L690-L692

For `IntoBackend`, I think these should be changed:

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/dataframe.py#L729-L730

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/stable/v1/__init__.py#L186-L187

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/stable/v2/__init__.py#L149-L150


My wish is for them to reflect the runtime checks, as mentioned in (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2269638057).

So for example, this test should start producing a warning:


https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/tests/frame/lazy_test.py#L72-L75

In the same way that we now do in https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2267040161

https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/tests/from_numpy_test.py#L55-L58

To be able to do this we can either:

1. Create one or more *new* aliases, like `IntoLazyBackend`
2. Do something like (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2266954326) and spell the types like:
    i. `IntoBackendAny`
    ii. `IntoBackend[Group]`
    iii. `IntoBackend[Polars | Pandas]`

narwhals/dataframe.py
FBruzzesi: > so I feel like I need to be the bad guy and ask maybe less typing? 😢

I thought this day would have never come!

Ok I totally see your point, I thought it was an initial setup for further improvements, but I agree, let's start simpler since at the moment they are not solving much

FBruzzesi:
Hey @dangotbanned 👋🏼

> Out of all the places `IntoEagerBackend` is used, I think the following should be changed:
> 
> https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/dataframe.py#L2472-L2474

This is correct, only a subset is supported

> https://github.com/narwhals-dev/narwhals/blob/f14ecd76d29844a079f37575edb6ac69e27640f9/narwhals/functions.py#L690-L692

Here I think we are wrong. We should not allow duckdb and ibis to `read_parquet`, since the output would not be a DataFrame, rather a LazyFrame. 

Now consider these two cases, in the first, we only support pandas, polars and pyarrow, in the second we would support also modin and cudf.

So in `read_parquet` we should have `IntoEagerBackend`, what should be call the type to annotate the variable in  `collect`? I don't have a good option in mind, but as mentioned previously, to me that's more of a (current) limitation of narwhals as the idea of collecting _any_ lazy to _any_ eager should support to collect to modin and cudf in principle.

---

For the `.lazy` case a similar argument applies -disclaimer: I forgot that we need to leave `backend=None` to convert something like `DataFrame[ArrowFrame]` to `LazyFrame[ArrowFrame]` - but in principle I would opt for a fully fledged `IntoLazyBackend` with the full list of lazy backends

--- 

For the second option proposed, namely:

> Do something like (https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2266954326) and spell the types like:
> i. IntoBackendAny
> ii. IntoBackend[Group]
> iii. IntoBackend[Polars | Pandas]

I am not fully grasping this approach/I am having skill issues 🙈

---

Update: https://github.com/narwhals-dev/narwhals/pull/2971/commits/0c398e495d1df593e6a7f093c323692b8842a3f9 splits into eager and lazy, and it looks quite promising, except one dirty `type: ignore` - commenting inline

narwhals/_utils.py
```diff
@@ -360,9 +365,23 @@ def from_string(
         except ValueError:
             return Implementation.UNKNOWN
 
+    @overload
+    @classmethod
+    def from_backend(  # type: ignore[overload-overlap]
```
FBruzzesi: @dangotbanned I know this is dirty. The error I am getting without it is:

> narwhals/_utils.py:370: error: Overloaded function signatures 1 and 2 overlap with incompatible return types  [overload-overlap]

which I assume it's because of `"polars"` and `Implementation.POLARS` being in both

dangotbanned: This is what the terms `*Only` and `*Allowed` were for 😂 

https://github.com/narwhals-dev/narwhals/blob/c441460cc091c33ddb08758ba01cfd895ff0f83f/narwhals/_namespace.py#L66-L69

dangotbanned:
These are good fails! 😄 

- https://github.com/narwhals-dev/narwhals/actions/runs/16966081186/job/48090269600?pr=2971


@FBruzzesi this is my top priority for reviewing btw, I've just been busy responding on all my PRs (thanks again for reviewing those 😍 )

narwhals/_arrow/dataframe.py
```diff
@@ -557,16 +559,16 @@ def lazy(self, *, backend: Implementation | None = None) -> CompliantLazyFrameAn
...
 
             return ArrowDataFrame(
                 self.native, version=self._version, validate_column_names=False
             )
 
-        if backend is Implementation.PANDAS:
+        if backend.is_pandas():
             from narwhals._pandas_like.dataframe import PandasLikeDataFrame
 
             return PandasLikeDataFrame(
```
dangotbanned: Is it okay if I revert these *for this PR*?

I prefer the shorter version using `.is_<name>`.
However, using `backend is Implementation.<name>` actually has a nice benefit **because** of this PR.

For example, after we've checked `Implementation.{PYARROW,PANDAS,POLARS}` a type checker can determine that the only remaining members of `EagerImplementation` are `Implementation.{CUDF,MODIN}`

image


*Less important*, but it also makes the diff smaller for the actual review 🙂

FBruzzesi: Sorry I should have commented that. Reason for this change is that the first annotation change I made was _wrong_ and using `is_<name>` method saved me!

I think I first passed `IntoBackend` somewhere*, and got the warning when performing `is_ibis()` (possibly with a string). On the other hand `x is y` does not bound it to anything i.e. you could do `"pandas" is Implementation.PANDAS`

*_somewhere_ being `.lazy()` for which we had the check mixed: the first few were `x is y` and then there was `x.is_ibis()`. I can see the benefits of both approaches. I want to try something before rolling back

FBruzzesi: Nevermind, I had hoped that the following was possible:

```py
def is_pandas(self) -> TypeGuard[Literal[Implementation.PANDAS]]:
```
to get the best of both words but it is not :| 

> User-defined type guard functions and methods must have at least one input

dangotbanned: ah yeah, that's why I've been using `is_eager_allowed` like in (https://github.com/narwhals-dev/narwhals/pull/2971/commits/3258d308f43361a9fcd4f91b289853d7249a1f46)

and why these exist for `ibis`

https://github.com/narwhals-dev/narwhals/blob/5ea6ee40850916f55989c185708713fb4280a329/narwhals/_ibis/utils.py#L175-L192

FBruzzesi: Reasonable PEP: add `SelfIs` 😂

narwhals/_utils.py
dangotbanned: This is how I'd resolve the issue ([fix(typing): Avoid overloads for `collect`, `lazy`](https://github.com/narwhals-dev/narwhals/pull/2971/commits/e69681ccebad851c5c9a8d79091dd5bbb229d363))

I'm still a bit confused on why spark-like isn't available for `.lazy()`, but if it were - we could just have an `is_lazy_allowed` instead

IIUC, `modin`, `cudf` aren't included for `.collect()` - since that might require roundtrip-ing through `pandas`, `pyarrow`?

In either case, the new aliases and guards I've added aren't intended to be user-facing

narwhals/_polars/dataframe.py
```diff
@@ -578,7 +580,7 @@ def func(*args: Any, **kwargs: Any) -> Any:
         return func
 
     def _iter_columns(self) -> Iterator[PolarsSeries]:  # pragma: no cover
-        yield from self.collect(self._implementation).iter_columns()
+        yield from self.collect(self._implementation).iter_columns()  # type: ignore[arg-type]
```
dangotbanned:
```suggestion
        yield from self.collect(Implementation.POLARS).iter_columns()
```

dangotbanned:
Reporting back with my findings 🫡 from (https://github.com/narwhals-dev/narwhals/compare/chore/expose-into-backend...into-backend-generic) 

### SCREAMING_SNAKE_CASE

We *could* define a third set of aliases that only use `Implementation`

I hadn't thought to scream them *before*, but it matches how enum members are spelled:

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L31-L48

They're **mainly** used internally, but I thought it was nice not to need `Implementation` as a suffix

- (https://github.com/narwhals-dev/narwhals/compare/chore/expose-into-backend...into-backend-generic#diff-d379de34a0e309244b9a412352a0c456fb6e395460b63e7bd635304c282fe817)
- (https://github.com/narwhals-dev/narwhals/compare/chore/expose-into-backend...into-backend-generic#diff-0a5c92eaae0e453688155b2ad6625e32a380dae55b1df661f6170893f17473e1)

But **also** in compositions like:

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L56

### Backend, backend, backend, backend
With how many times we've used this term, it is odd how there's not been a `Backend` alias before.

`Backend` is a superset of `BackendName` and their respective `Implementation`s

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L76-L77

### Why is `Backend` useful?
I'm glad you asked!

Here, I've used it to define `IntoBackend` as a generic alias - which is very flexible

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L81-L85

We can use that generic alias to define a more concrete ones, like these:

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L142-L144

Because of `TypeVar(bound=Backend)`, we get a nice warning if it is misused:

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L146-L147

We can also use `IntoBackend` to describe a more accurate signature, when the full set of `(Eager|Lazy)Allowed` isn't *actually* allowed 😉 

No spark-like!

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L150-L153


No cudf, modin!

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L156-L159

And on the dancefloor in 

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/tests/frame/lazy_test.py#L50-L79


### But wait ... there's docs!
With `IntoBackend` being generic, we can define the doc with *multiple examples **once** and they appear everywhere*!

<details><summary><b>Show docs</b></summary>
<p>

*everywhere that `IntoBackend` is used **directly**

https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L82-L139

</p>
</details>

FBruzzesi:
Thanks for following up on this effort @dangotbanned - I mean I don't even know why it was _me_ who started a PR focused on better typing 😂

---

On a more serious note, let me try to add some comments on multiple points brought up here

> Reporting back with my findings 🫡 from ([chore/expose-into-backend...into-backend-generic](https://github.com/narwhals-dev/narwhals/compare/chore/expose-into-backend...into-backend-generic))

From this branch, if I need to nitpick something is that I am not the biggest fan of variable names `EAGER_ALLOWED` and `LAZY_ALLOWED` then used in methods signature. Which also leads to
 
> ### SCREAMING_SNAKE_CASE
> We _could_ define a third set of aliases that only use `Implementation`
> 
> I hadn't thought to scream them _before_, but it matches how enum members are spelled:
>
> They're **mainly** used internally, but I thought it was nice not to need `Implementation` as a suffix

Since these are types/type hints I am a preference to have it capitalized lowercase as they are already in this branch. However I see the issue of having the composition of Implementation literal and the string value into yet another type.
 
> But **also** in compositions like:
> 
> https://github.com/narwhals-dev/narwhals/blob/bd8ded1c0b842acd9e6cf0ed767584d1d1a7fd93/narwhals/_typing.py#L56

---

Coming to the 

> ### Backend, backend, backend, backend

section.

I think this is the real win of your idea. Having the flexibility of adding each single backend if needed does create a perfect combination. I am full onboarded on this

---

So TL;DR: I really really like the `IntoBackend` and `BackendT` implementation and how useful they are.

I am not super excited of having uppercase types to use internally.

Also related to naming, I have mixed feeling about these types and functions: `_CanCollectInto`, `_CanLazyInto`, `can_collect_into`, `can_lazy_into`. I don't have a _strong_(*) proposal yet but I will keep thinking about it.

(*) not sure how would you like `(_S|s)upported` over `(_C|c)an`?

narwhals/_utils.py
dangotbanned: > Also related to naming, I have mixed feeling about these types and functions: `_CanCollectInto`, `_CanLazyInto`, `can_collect_into`, `can_lazy_into`. I don't have a _strong_(*) proposal yet but I will keep thinking about it.
> 
> (*) not sure how would you like `(_S|s)upported` over `(_C|c)an`?


That works too!

I was quite upset since I worked on the `collect` one first and had the great idea of `is_collectable` - but `is_lazyable` doesn't roll off the tongue as nicely 😂

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2971#issuecomment-3193667391

## Screeeeam

> From this branch, if I need to nitpick something is that I am not the biggest fan of variable names `EAGER_ALLOWED` and `LAZY_ALLOWED` then used in methods signature. Which also leads to

> Since these are types/type hints I am a preference to have it capitalized lowercase as they are already in this branch. However I see the issue of having the composition of Implementation literal and the string value into yet another type.

> I am not super excited of having uppercase types to use internally.

Agreed!

That part is really just trying to find where we can balance tradeoffs for the semi-internal stuff.

### Alternative

Another option that's a bit more involved:

Suffix the `str` aliases with `Name`
```diff
- _Polars: TypeAlias = Literal["polars"]
- _EagerAllowed: TypeAlias = Literal[_Polars, _EagerOnly]
+ _PolarsName: TypeAlias = Literal["polars"]
+ _EagerAllowedName: TypeAlias = Literal[_PolarsName, _EagerOnlyName]
```

Rename the `Implementation` aliases to use the name previously held by `str` aliases

```diff
- POLARS: TypeAlias = Literal[Implementation.POLARS]
- EAGER_ALLOWED: TypeAlias = Literal[EAGER_ONLY, POLARS]
+ _Polars: TypeAlias = Literal[Implementation.POLARS]
+ _EagerAllowed: TypeAlias = Literal[_EagerOnly, _Polars]
```

Then update `str | Implementation` aliases

```diff
- Polars: TypeAlias = Literal[_Polars, POLARS]
+ Polars: TypeAlias = Literal[_PolarsName, _Polars]
  EagerAllowed: TypeAlias = Literal[EagerOnly, Polars]
```

### Tradeoff
That removes the need for uppercase, but using *what looks like* a private type alias here seems odd


```diff
class CompliantLazyFrame:
     def collect(
-        self, backend: EAGER_ALLOWED | None, **kwargs: Any
+        self, backend: _EagerAllowed | None, **kwargs: Any
     ) -> CompliantDataFrameAny: ...
```

However, it does have precedent since these are waaaay more public

https://github.com/narwhals-dev/narwhals/blob/5309dd5a39c123020fc281f51e5c33d5818ba1fa/narwhals/typing.py#L347

https://github.com/narwhals-dev/narwhals/blob/5309dd5a39c123020fc281f51e5c33d5818ba1fa/narwhals/typing.py#L349


### TLDR
I know this isn't the meat of (https://github.com/narwhals-dev/narwhals/pull/2971#issuecomment-3193667391), but I just wanted to show an alternative in action 😎 

Either works for me

dangotbanned:
> Coming to the
> 
> >    ### Backend, backend, backend, backend
> 
> section.
> 
> I think this is the real win of your idea. Having the flexibility of adding each single backend if needed does create a perfect combination. I am full onboarded on this

Perfect!

🥳 🥳 🥳

I'll get that branch re-synced with this, then open a PR to merge it into this one?

We can continue picking our nits in a new playground 😉

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2971#event-19191908250

I've marked as draft *until* #3002 is merged back in

dangotbanned:
Will try to do another once-over tonight, but I'm feeling this is ready to go 🙌

narwhals/_typing.py
```diff
@@ -0,0 +1,141 @@
...
+Ibis: TypeAlias = Literal[_Ibis, _IbisImpl]
+PandasLike: TypeAlias = Literal[_PandasLike, _PandasLikeImpl]
+SparkLike: TypeAlias = Literal[_SparkLike, _SparkLikeImpl]
+EagerOnly: TypeAlias = Literal[PandasLike, Arrow]
+EagerAllowed: TypeAlias = Literal[EagerOnly, Polars]
+LazyOnly: TypeAlias = Literal[SparkLike, Dask, DuckDB, Ibis]
+LazyAllowed: TypeAlias = Literal[LazyOnly, Polars]
+
+BackendName: TypeAlias = Literal[_EagerAllowed, _LazyAllowed]
+Backend: TypeAlias = Literal[EagerAllowed, LazyAllowed]
```
dangotbanned: Just had a preview of the API Reference.

> [!NOTE]
> This doesn't affect IDE usage (https://github.com/narwhals-dev/narwhals/pull/3002#discussion_r2280623292)


Currently all of these appear in annotations, but don't have an entry we can link to:

- `Backend`
- `EagerAllowed`
- `Ibis`
- `DuckDB`
- `Dask`
- `Arrow`
- `Polars`
- `Pandas`

I don't know if that is *necessarily* an issue?

`Backend` and `EagerAllowed` seem more important to define - but they are also dependent on the others 🤔 

## Examples

image


image

This one seems clear enough

image

narwhals/typing.py
```diff
@@ -3,6 +3,7 @@
 from typing import TYPE_CHECKING, Any, Literal, Protocol, TypeVar, Union
 
 from narwhals._compliant import CompliantDataFrame, CompliantLazyFrame, CompliantSeries
+from narwhals._typing import IntoBackend as IntoBackend  # noqa: PLC0414
```
dangotbanned: ## Preview
image

dangotbanned:
This PR also requires some (typing-only) fixing in `altair`

- #2944

narwhals/_typing.py
FBruzzesi: @dangotbanned what would be your preferred solution? I can think of either:

- Exposing `Backend`, `EagerAllowed`, `LazyAllowed`
- Mention them in the `IntoBackend` docstrings

If you have other options I am open to it

FBruzzesi:
> This PR also requires some (typing-only) fixing in `altair`
> 
> * [fix(typing): Avoid overlapping `DataFrame`, `LazyFrame` #2944](https://github.com/narwhals-dev/narwhals/pull/2944)

Should we wait to merge both before fixing the issue(s) in altair then?

dangotbanned:
> > This PR also requires some (typing-only) fixing in `altair`
> > 
> > * [fix(typing): Avoid overlapping `DataFrame`, `LazyFrame` #2944](https://github.com/narwhals-dev/narwhals/pull/2944)
> 
> Should we wait to merge both before fixing the issue(s) in altair then?

Yeah I think so

Well #2944 at least has to be merged before the fix for it in `altair` can happen.
I was relying on the broken typing 😅 

I think the fix for this PR can happen before/after - but we might as well fix both in one PR

narwhals/_typing.py
dangotbanned: > Exposing `Backend`, `EagerAllowed`, `LazyAllowed`

Yeah I like this option!

I haven't tested out what that looks like in the docs yet though

FBruzzesi: How does this look? (Right after `IntoBackend`)

image

FBruzzesi: Another question: don't you think we should expose `IntoBackend` (and others) publicly?

dangotbanned: > Another question: don't you think we should expose `IntoBackend` (and others) publicly?

Adding them to `nw.typing.__all__`?

Yeah that probably makes a lot of sense
I can imagine it being handy for downstream to use in annotations and not need to worry about syncing if/when we add new backends 🙂

dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2283280386

> How does this look? (Right after `IntoBackend`)

I'm curious now if we could cross ref here (but keep `IntoBackend` as-is)
- https://narwhals-dev.github.io/narwhals/api-reference/utils/#narwhals.utils.Implementation

That has a node for each `Implementation`, which also lists the `_value_`

dangotbanned:
```diff
diff --git a/narwhals/_typing.py b/narwhals/_typing.py
index ddac8bd1e..f11390d4a 100644
--- a/narwhals/_typing.py
+++ b/narwhals/_typing.py
@@ -97,7 +97,7 @@ Examples:
     |     1  2  4      |
     └──────────────────┘
 
-    An Implementation, such as: `Implementation.POLARS`, `Implementation.DUCKDB`, `Implementation.PYSPARK`
+    An [`Implementation`][narwhals.Implementation], such as: [`POLARS`][narwhals.Implementation.POLARS], [`DUCKDB`][narwhals.Implementation.DUCKDB],  [`PYSPARK`][narwhals.Implementation.PYSPARK]
 
     >>> import narwhals as nw
     >>> nw.read_parquet("file.parquet", backend=nw.Implementation.PYARROW)

```

Looks terrible in VSCode, but quite helpful on the API reference

image

narwhals/_typing.py
```diff
@@ -0,0 +1,157 @@
...
+Arrow: TypeAlias = Literal[_Arrow, _ArrowImpl]
+Dask: TypeAlias = Literal[_Dask, _DaskImpl]
+DuckDB: TypeAlias = Literal[_DuckDB, _DuckDBImpl]
+Ibis: TypeAlias = Literal[_Ibis, _IbisImpl]
+PandasLike: TypeAlias = Literal[_PandasLike, _PandasLikeImpl]
+SparkLike: TypeAlias = Literal[_SparkLike, _SparkLikeImpl]
+EagerOnly: TypeAlias = Literal[PandasLike, Arrow]
+LazyOnly: TypeAlias = Literal[SparkLike, Dask, DuckDB, Ibis]
+EagerAllowed: TypeAlias = Literal[EagerOnly, Polars]
+"""A string name or [`narwhals.Implementation`][narwhals.Implementation] of an eager backend.
```
dangotbanned: You can do this funky syntax if the full name is the reference 😎

```suggestion
"""A string name or [`narwhals.Implementation`][] of an eager backend.
```

PR #2977, opened 2025-08-12
feat: add `DataFrame.top_k` and `LazyFrame.top_k`

### LINKED ISSUE #2947: enh?: `{DataFrame/LazyFrame}.top_k`
I found myself reaching for this today

https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.top_k.html#polars.LazyFrame.top_k

In some cases I think this should be more efficient than `.sort().head(...)` which people would often reach for (though i'd hope that lazy engines could optimise this away?). PyArrow unfortunately doesn't seem to have `top_k` so for now we may need to do `sort().head()` there

mock-up of how to do it for sql-like dataframes:
```python
In [12]: df.top_k(2, by='b')
Out[12]:
shape: (2, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ str ┆ i64 │
╞═════╪═════╡
│ y   ┆ 3   │
│ z   ┆ 2   │
└─────┴─────┘

In [13]: duckdb.sql("""
    ...: with cte as
    ...: (from df
    ...: select *, row_number() over (order by b desc) as row_number)
    ...: from cte
    ...: select * exclude(row_number)
    ...: where row_number <= 2
    ...: order by row_number
    ...: """)
Out[13]:
┌─────────┬───────┐
│    a    │   b   │
│ varchar │ int64 │
├─────────┼───────┤
│ y       │     3 │
│ z       │     2 │
└─────────┴───────┘
```

raisadz:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2947

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_ibis/dataframe.py
+++ b/narwhals/_ibis/dataframe.py
@@ -353,7 +353,7 @@ class IbisLazyFrame(
             reverse = [reverse] * len(list(by))
         sort_cols = []
 
-        for is_reverse, by_col in zip(reverse, by):
+        for is_reverse, by_col in zip_strict(reverse, by):
             direction_fn = ibis.asc if is_reverse else ibis.desc
             col = direction_fn(by_col, nulls_first=False)
             sort_cols.append(cast("ir.Column", col))
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -450,7 +450,7 @@ class ArrowDataFrame(
         else:
             sorting = [
                 (key, "ascending" if is_ascending else "descending")
-                for key, is_ascending in zip(by, reverse)
+                for key, is_ascending in zip_strict(by, reverse)
             ]
         return self._with_native(
             self.native.take(pc.select_k_unstable(self.native, k, sorting)),  # type: ignore[call-overload]
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -347,7 +347,7 @@ class SparkLikeLazyFrame(
         sort_funcs = (
             self._F.desc_nulls_last if not d else self._F.asc_nulls_last for d in reverse
         )
-        sort_cols = [sort_f(col) for col, sort_f in zip(by, sort_funcs)]
+        sort_cols = [sort_f(col) for col, sort_f in zip_strict(by, sort_funcs)]
         return self._with_native(self.native.sort(*sort_cols).limit(k))
 
     def drop_nulls(self, subset: Sequence[str] | None) -> Self:


narwhals/dataframe.py
```diff
@@ -239,6 +239,14 @@ def sort(
             self._compliant_frame.sort(*by, descending=descending, nulls_last=nulls_last)
         )
 
+    def top_k(
+        self, k: int, *, by: str | Iterable[str], reverse: bool | Sequence[bool] = False
+    ) -> Self:
+        flatten_by = flatten([by])
```
FBruzzesi, line 251: Can we add a check that if `reverse` is a sequence, and it's length is different than `flatten_by`, then an exception is raise? This guarantees that `zip(by, reverse)` at the compliant level is same as `zip_strict`.

From polars:

```py
df = pl.DataFrame(
    {
        "a": ["a", "b", "a", "b", "b", "c"],
        "b": [2, 1, 1, 3, 2, 1],
    }
)

df.top_k(4, by=["b", "a"], reverse=[True])
```

> ValueError: the length of `reverse` (1) does not match the length of `by` (2)

narwhals/_duckdb/dataframe.py
```diff
@@ -409,6 +409,24 @@ def sort(self, *by: str, descending: bool | Sequence[bool], nulls_last: bool) ->
             )
         return self._with_native(self.native.sort(*it))
 
+    def top_k(self, k: int, *, by: Iterable[str], reverse: bool | Sequence[bool]) -> Self:
+        df = self.native  # noqa: F841
```
FBruzzesi: If you prefix the variable name with an underscore (`_df`) you can avoid the `# noqa: F841` flag. It's hacky I know

narwhals/_ibis/dataframe.py
```diff
@@ -346,6 +346,19 @@ def sort(self, *by: str, descending: bool | Sequence[bool], nulls_last: bool) ->
...
+    def top_k(self, k: int, *, by: Iterable[str], reverse: bool | Sequence[bool]) -> Self:
+        by = list(by)
+        if isinstance(reverse, bool):
+            reverse = [reverse] * len(by)
+        sort_cols = []
+
+        for i in range(len(by)):
+            direction_fn = ibis.desc if not reverse[i] else ibis.asc
+            col = direction_fn(by[i], nulls_first=False)
+            sort_cols.append(cast("ir.Column", col))
```
FBruzzesi: Probably a personal style preference, feel free to ignore:

```suggestion
        for is_reverse, by_col in zip(reverse, by):
            direction_fn = ibis.asc is is_reverse else ibis.desc
            col = direction_fn(by_col, nulls_first=False)
            sort_cols.append(cast("ir.Column", col))
```

If you consider this, you can also probably avoid the `by=list(by)` at the top

narwhals/_duckdb/dataframe.py
```diff
@@ -409,6 +409,24 @@ def sort(self, *by: str, descending: bool | Sequence[bool], nulls_last: bool) ->
...
+        order_by = [f'"{col}" {dir}' for col, dir in zip(by, directions)]
+        query = f"""
+        WITH cte AS (
+            SELECT *, ROW_NUMBER() OVER (ORDER BY {",".join(order_by)}) as row_number
+            FROM df
+        )
+        SELECT * EXCLUDE(row_number)
+        FROM cte
+        WHERE row_number <= {k}
+        """  # noqa: S608
```
FBruzzesi: I _think_ you can use a [qualify clause](https://duckdb.org/docs/stable/sql/query_syntax/qualify.html) to avoid making a cte (also comes down to personal preference, it should have no real impact on performance).

```suggestion
        query = f"""
        SELECT *
        FROM df
        QUALIFY ROW_NUMBER() OVER (ORDER BY {",".join(order_by)}) <= {k}
        """  # noqa: S608
```

*Disclaimer: the suggestion is written without verifying details 🙈

raisadz:
Thanks for the review @FBruzzesi ! I addressed your comments and will add `zip_strict` from https://github.com/narwhals-dev/narwhals/pull/3003 after it is merged

narwhals/dataframe.py
FBruzzesi: @raisadz I would still prefer to add a check at this level to also align the error with polars (notice that the output of `flatten` is a list anyway), but feel free to merge. We can follow up on it

MarcoGorelli: i think there's some other places where this would be useful (like `sort`) so we could probably make a validation utility for this and use it in multiple places

MarcoGorelli:
merging then, i've opened https://github.com/narwhals-dev/narwhals/issues/3026 for a follow-up, thanks all for comments!

Issue #2959, opened 2025-08-07
Tracking: Overhauling the test suite


dangotbanned:
Collecting together all of the various efforts going on 😳 

### Issues
- #1726
- #2872
- #2818
- #2742
- #1893

### Comments/Discord
- https://discord.com/channels/1235257048170762310/1402677544221937705
- https://github.com/narwhals-dev/narwhals/issues/1726#issuecomment-3148623543
- https://github.com/narwhals-dev/narwhals/issues/2872#issuecomment-3146509755
- https://github.com/narwhals-dev/narwhals/pull/2950#discussion_r2257642713
- https://github.com/narwhals-dev/narwhals/pull/2950#pullrequestreview-3096363509

### Branches
- https://github.com/narwhals-dev/narwhals/compare/main...testing-fixtures
- https://github.com/narwhals-dev/narwhals/compare/main...EdAbati:draft-new-constructor

### PRs
- https://github.com/narwhals-dev/narwhals/pull/2879/files#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6
- https://github.com/narwhals-dev/narwhals/pull/2933/files#diff-9e08e8c7769db7b58089fd7659d75ca96df40bb7b5d50299d3a30abd54da9ad6
- https://github.com/narwhals-dev/narwhals/pull/2933/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128
- https://github.com/narwhals-dev/narwhals/pull/2933/files#diff-a1a637479cb57d5314c55a064184fda2df55e96627f3ef57da09b2c07e75b69f

dangotbanned:
@EdAbati @FBruzzesi 
The **PRs** section only has diffs from my open PRs, feel free to add your own 🙌

dangotbanned:
### Hypothesis
Reading the [hypothesis strategies](https://hypothesis.readthedocs.io/en/latest/reference/strategies.html) docs was interesting.

There might be some stuff we can learn from how each of these work (*in order of relevance to `narwhals`*):
- [strategies/array-api](https://hypothesis.readthedocs.io/en/latest/reference/strategies.html#array-api) | [source code](https://github.com/HypothesisWorks/hypothesis/blob/891297c6978207bc7168cc67875108aa4bc944e3/hypothesis-python/src/hypothesis/extra/array_api.py)
[...]

Issue #2939, opened 2025-08-04
[Enh]: Sorting complex types


FBruzzesi:
# Description

Polars allow to sort by a column/series of list, array and struct type. The sorting is based on the order of the lists/arrays/structs, using lexicographical (element-wise) comparison (at least according to a few experiments I run).

I am having a bit of a hard time to replicate such behavior both in pandas and pyarrow (I didn't try other backends for now). Should we aim to support this?

This issue came out when looking into #2804 as polars allows for a `check_order` flag for _any_ type.

dangotbanned:
I wasn't able to find any `pyarrow` issues/PRs relating to an explicit API for sorting nested data.
Do you get a kernel error when trying to use the general sorting functions/methods?

FBruzzesi:
Yes both for pyarrow and pandas backed by pyarrow. Funnily enough, for a pandas without pyarrow sorts the list objects

dangotbanned:
> Do you get a kernel error when trying to use the general sorting functions/methods?

> Yes both for pyarrow and pandas backed by pyarrow. 

We might be in luck!

I just tried this (on #2879), but the error *isn't* a kernel error (which is like #2835)


```py
import pyarrow as pa
from narwhals._utils import Version

data = {"a": ["A", "B", "A"], "b": [1, 2, 3], "c": [9, 2, 4], "d": [8, 7, 8]}
table = pa.table(data)
arrow_ns = Version.MAIN.namespace.from_backend("pyarrow").compliant
struct_series_pa = arrow_ns.from_native(table).to_struct("what")
>>> struct_series_pa.native.sort()
ArrowInvalid: Invalid sort key column: No match for FieldRef.Name() in a: string
b: int64
c: int64
d: int64
----
a:
  [...
```

For this case, it seems to work fine - as long as we use an `Array` and not a `ChunkedArray`.

Here it is using [`pyarrow.StructArray.sort`](https://arrow.apache.org/docs/python/generated/pyarrow.StructArray.html):

```py
>>> struct_series_pa.native.combine_chunks().sort(order="descending")
<pyarrow.lib.StructArray object at 0x000001DDA406AC80>
-- is_valid: all not null
-- child 0 type: string
  [
    "B",
    "A",
    "A"
  ]
-- child 1 type: int64
  [
    2,
    3,
    1
  ]
-- child 2 type: int64
  [
    2,
    4,
    9
  ]
-- child 3 type: int64
  [
    7,
    8,
    8
[...]

Issue #2804, opened 2025-07-07
[Enh]: Testing module


elephaint:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

At first [hierarchicalforecast](https://nixtlaverse.nixtla.io/hierarchicalforecast/index.html), in the future also our other open-source libraries in the nixtla ecosystem

### Please describe the purpose of the new feature or describe the problem to solve.

[Pandas](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html) and [Polars](https://docs.pola.rs/api/python/stable/reference/api/polars.testing.assert_frame_equal.html) support testing functionalies, like the following functions:

`pd.testing.assert_frame_equal`
`pl.testing.assert_frame_equal`

This is really convenient e.g. when unit testing functionality. Ideally, Narwhals has it too, so that we can abstract away from the dataframe provided and run the unit test on any desired dataframe engine.

### Suggest a solution if possible.

Include wrappers for this testing functionality, or include native Narwhals testing functionality, e.g.

`nw.testing.asserts.assert_frame_equal`

FBruzzesi:
Hey @elephaint thanks for the request! I had this in mind _some time ago_ already (https://github.com/narwhals-dev/narwhals/issues/739) but there was not much interest on it.
This is definitely in scope - I will bring it up in the next community call :)

dangotbanned:
@elephaint I agree this would be great to have

@FBruzzesi should we move this into a comment on (#739) and close this one as duplicate?

dangotbanned:
Wow, took me 2 months to follow-up (https://github.com/narwhals-dev/narwhals/issues/2804#issuecomment-3048336947) 😅

@elephaint I've edited the description and added it to the older issue (https://github.com/narwhals-dev/narwhals/issues/739#issuecomment-3284312500).
The good news is @FBruzzesi has been busy making some progress 😄

Issue #739, opened 2024-08-07
[Enh]: Provide (basic) testing tables to downstream libraries


FBruzzesi:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

- This is a long shot, and I am not sure which (if any) and how other libraries do it.
- For sure I could see adoption in scikit-lego

### Please describe the purpose of the new feature or describe the problem to solve.

Provide a standard set of values and functionalities to generate tables for each backend we support

### Suggest a solution if possible.

- Have this as a separate library, let's say `narwhals-testing"
- Have this in an internal module, which we can use for our testing purpose as well

### If you have tried alternatives, please describe them below.

Manually create tables in downstream library :)

MarcoGorelli:
Agree! I think @machow also wanted some testing functionality exposed for Shiny

FBruzzesi:
Adding some tracking here for what in my opinion means to close this PR:

- [x] `nw.testing.asserts` module
    - [x] `assert_series_equal`
    - [x] `assert_frame_equal`
- [ ] `nw.testing.constructors` module to provide fixture(s) to use in test suite of downstream libraries
  - [Discord dicussion](https://discord.com/channels/1235257048170762310/1402677544221937705)
  - My naive attempt: [testing-fixtures branch](https://github.com/narwhals-dev/narwhals/compare/main...testing-fixtures)
  - A more sophisticated approach by @EdAbati: [draft-new-constructor](https://github.com/narwhals-dev/narwhals/compare/main...EdAbati:draft-new-constructor)

EdAbati:
>   - A more sophisticated approach by @EdAbati: [draft-new-constructor](https://github.com/narwhals-dev/narwhals/compare/main...EdAbati:draft-new-constructor)

(There is a fine line between 'more sophisticated' and 'overly complicated' 😅 I hope to finish the draft soon)

dangotbanned:
_Originally posted by @elephaint in [#2804](https://github.com/narwhals-dev/narwhals/issues/2804#issue-3210161851)_

> [!NOTE]
> Edited by @dangotbanned 

> ### Is this feature is needed to adopt Narwhals in an open source project?
> 
> At first [`hierarchicalforecast`](https://nixtlaverse.nixtla.io/hierarchicalforecast/index.html), in the future also our other open-source libraries in the [Nixtla](https://github.com/Nixtla) ecosystem
[...]

PR #2983, opened 2025-08-13
RFC, feat: Add `assert_series_equal`

FBruzzesi:
# Description

This PR introduces the testing module (mirroring the polars structure) and `assert_series_equal`. It's a step towards: #739 and #2804.

I think it's a good structure to leave space for, let's say, a `constructors` module within it (see #2959 and in particular the discord conversation linked in the issue).

Additional comments:

- One edge case/blocker for nested dtype here is mentioned in #2939 (I am raising a `NotImplementedError`).
- I started with `assert_series_equal` since if this goes through, then `assert_frame_equal` can re-use it for the value checks. However I didn't want to bloat the PR with 1k+ lines changes
- As usual... 600+ lines changes, but in all honesty 50%+ are tests.
- If this and `assert_frame_equal` end up being implemented, I think we can have an issue to track usage for them in the test suite itself

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/testing/asserts/frame.py
+++ b/narwhals/testing/asserts/frame.py
@@ -1,0 +1,20 @@
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from narwhals.typing import FrameT
+
+
+def assert_frame_equal(
+    left: FrameT,
+    right: FrameT,
+    *,
+    check_row_order: bool = True,
+    check_column_order: bool = True,
+    check_dtypes: bool = True,
+    check_exact: bool = False,
+    rel_tol: float = 1e-05,
+    abs_tol: float = 1e-08,
+    categorical_as_str: bool = False,
+) -> None: ...
--- a/narwhals/testing/asserts/frame.py
+++ b/narwhals/testing/asserts/frame.py
@@ -3,12 +3,12 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from narwhals.typing import FrameT
+    from narwhals.typing import IntoFrameT
 
 
 def assert_frame_equal(
-    left: FrameT,
-    right: FrameT,
+    left: IntoFrameT,
+    right: IntoFrameT,
     *,
     check_row_order: bool = True,
     check_column_order: bool = True,
@@ -17,4 +17,6 @@ def assert_frame_equal(
     rel_tol: float = 1e-05,
     abs_tol: float = 1e-08,
     categorical_as_str: bool = False,
-) -> None: ...
+) -> None:
+    msg = "TODO"  # pragma: no cover
+    raise NotImplementedError(msg)
--- a/narwhals/testing/asserts/series.py
+++ b/narwhals/testing/asserts/series.py
@@ -52,13 +52,14 @@ def assert_series_equal(  # noqa: C901, PLR0912
     if (l_len := len(left_)) != (r_len := len(right_)):
         raise_assertion_error("Series", "length mismatch", l_len, r_len)
 
-    if check_dtypes and (l_dtype := left_.dtype) != (r_dtype := right_.dtype):
+    if (l_dtype := left_.dtype) != (r_dtype := right_.dtype) and check_dtypes:
         raise_assertion_error("Series", "dtype mismatch", l_dtype, r_dtype)
 
-    if check_names and (l_name := left_.name) != (r_name := right_.name):
+    if (l_name := left_.name) != (r_name := right_.name) and check_names:
         raise_assertion_error("Series", "name mismatch", l_name, r_name)
 
-    if isinstance(l_dtype, Categorical) and categorical_as_str:
+    # TODO(FBruzzesi): Add coverage
+    if isinstance(l_dtype, Categorical) and categorical_as_str:  # pragma: no cover
         left_, right_ = left_.cast(String()), right_.cast(String())
 
     if not check_order:
@@ -74,24 +75,29 @@ def assert_series_equal(  # noqa: C901, PLR0912
 
     l_vals, r_vals = left_.filter(~l_null_mask), right_.filter(~r_null_mask)
 
-    # TODO(FBruzzesi): Open points:
-    #  [x] Handle nan's and infinities for numerical
-    #  [ ] Nested values: do nested values compare "nicely"?
-    #  [ ] date vs datetime?
     if check_exact or not l_dtype.is_numeric():
-        if l_dtype.is_nested():
-            is_equal_mask = new_series(
+        if l_dtype.is_numeric():
+            # For numeric dtypes, we can use `is_close` with 0-tolerances to handle inf
+            # and nan values.
+            is_not_equal_mask = ~left_.is_close(
+                right_, rel_tol=0, abs_tol=0, nans_equal=True
+            )
+        elif l_dtype.is_nested():
+            # TODO(FBruzzesi): Handle nested with inner numerical
+            # 1. Polars raises: AssertionError: Series are different (nested value mismatch)
+            # 2. check_exact and tolerances are considered within the inner type as well
+            is_not_equal_mask = new_series(
                 l_vals.name,
-                (l_val == r_val for l_val, r_val in zip(l_vals, r_vals)),
+                (
+                    any(x != y for x, y in zip(l_val, r_val))
+                    for l_val, r_val in zip(l_vals, r_vals)
+                ),
                 backend=l_impl,
             )
-        elif l_dtype.is_numeric():
-            # Workaround via is_close with 0-tolerances to handle inf and nan values.
-            is_equal_mask = left_.is_close(right_, rel_tol=0, abs_tol=0, nans_equal=True)
         else:
-            is_equal_mask = l_vals == r_vals
+            is_not_equal_mask = l_vals != r_vals
 
-        if not is_equal_mask.all():
+        if is_not_equal_mask.any():
             raise_assertion_error("Series", "exact value mismatch", l_vals, r_vals)
 
     else:
--- a/narwhals/testing/asserts/series.py
+++ b/narwhals/testing/asserts/series.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
     CheckFn: TypeAlias = Callable[[SeriesT, SeriesT], None]
 
 
-def assert_series_equal(  # noqa: C901, PLR0912
+def assert_series_equal(
     left: IntoSeriesT,
     right: IntoSeriesT,
     *,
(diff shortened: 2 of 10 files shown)


FBruzzesi:
<details><summary> modin & cudf `clip` issue leading to `is_close` exception </summary>The modin errors are related to `is_close` failing, which is because of `clip` with pyarrow integers (I couldn't really replicate with floats). And that brought up an old memory: https://github.com/modin-project/modin/issues/7415.

Which now leads to the harsh part:

- It's my fault that we didn't catch that in `is_close`
- #2986 should fix it

---

Once that's merged, I will take care of the issues in the tests with old pandas version
</details>

is now solved with a workaround which enables `clip` to work without issues on both cudf and modin (#2986)

FBruzzesi:
@MarcoGorelli @dangotbanned gentle ping, also to bring to your attention a comment in holoviews https://github.com/holoviz/holoviews/pull/6567#issuecomment-3223295973 that might sparkle some motivation for this PR 😂

MarcoGorelli:
thanks for the ping! looks like a good idea!

dangotbanned:
### `rust` impl
Just saving this here as I've thought of and forgotten to find it 3 times now 😂

- https://github.com/pola-rs/polars/blob/cbf621add84d0a4d51b361b25132f61b61ff9126/crates/polars-testing/src/asserts/utils.rs

narwhals/testing/asserts/series.py
```diff
@@ -0,0 +1,274 @@
...
+    categorical_as_str: bool = False,
+) -> None:
+    """Assert that the left and right Series are equal.
+
+    Raises a detailed `AssertionError` if the Series differ.
+    This function is intended for use in unit tests.
+
+    Arguments:
+        left: The first Series to compare.
+        right: The second Series to compare.
```
dangotbanned, line 40: I don't think we should be accepting anything that isn't `nw.Series` here

I'm not opposed to that being another function (e.g. `assert_native_series_equal`), but this one should just be comparing - not doing conversion as well IMO

narwhals/testing/asserts/series.py
```diff
@@ -0,0 +1,274 @@
...
+
+    if check_exact or not left_.dtype.is_float():
+        _check_exact_values(
+            left_vals,
+            right_vals,
+            check_dtypes=check_dtypes,
+            check_exact=check_exact,
+            rel_tol=rel_tol,
+            abs_tol=abs_tol,
+            categorical_as_str=categorical_as_str,
```
dangotbanned, line 112: The nested cases need coverage for an inner `Categorical` + `categorical_as_str`

was surprised to see this didn't have an effect

```diff
diff --git a/narwhals/testing/asserts/series.py b/narwhals/testing/asserts/series.py
index 1521a335d..ca4e29c70 100644
--- a/narwhals/testing/asserts/series.py
+++ b/narwhals/testing/asserts/series.py
@@ -105,7 +105,6 @@ def assert_series_equal(
             check_exact=check_exact,
             rel_tol=rel_tol,
             abs_tol=abs_tol,
-            categorical_as_str=categorical_as_str,
         )
     else:
         _check_approximate_values(left_vals, right_vals, rel_tol=rel_tol, abs_tol=abs_tol)
@@ -160,7 +159,6 @@ def _check_exact_values(
     check_exact: bool,
     rel_tol: float,
     abs_tol: float,
-    categorical_as_str: bool,
 ) -> None:
     """Check exact value equality for various data types."""
     left_impl = left.implementation
@@ -182,7 +180,6 @@ def _check_exact_values(
             check_exact=check_exact,
             rel_tol=rel_tol,
             abs_tol=abs_tol,
-            categorical_as_str=categorical_as_str,
         )
         _check_list_like(left, right, left_dtype, right_dtype, check_fn=check_fn)
         # If `_check_list_like` didn't raise, then every nested element is equal
@@ -196,7 +193,6 @@ def _check_exact_values(
             check_exact=check_exact,
             rel_tol=rel_tol,
             abs_tol=abs_tol,
-            categorical_as_str=categorical_as_str,
         )
         _check_struct(left, right, left_dtype, right_dtype, check_fn=check_fn)
         # If `_check_struct` didn't raise, then every nested element is equal

```

narwhals/testing/asserts/series.py
FBruzzesi: That's fair (and thank for the commits fixing it). However, then we might need to be even _more_ e.g. also check that the narwhals namespace version is the same. In principle the check:

https://github.com/narwhals-dev/narwhals/blob/5d7040b021778a9e61c3fabc0944a84c79f094c0/narwhals/testing/asserts/series.py#L83-L90

would pass for:

```py
import narwhals.stable.v2 as nw_v2

assert_series_equal(
    nw_v1.from_native(...),
    nw_v2.from_native(...),
    ...
)
```

dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2983#discussion_r2311368534

Ah I may have been too quick suggesting this 😅

Coercing the `right` or whatever is equivalent to `expected`; might make sense for ergonomics. E.g. like our `assert_equal_data` 

But I think allowing both inputs to be native is a bit much

dangotbanned: I'm guessing you've already checked this, but `pandas` has quite a few of these

https://github.com/pandas-dev/pandas/blob/7bfef3b1cba58a6c3aa62493e0b0905bc59e6443/pandas/_testing/asserters.py

FBruzzesi: > But I think allowing both inputs to be native is a bit much

I might disagree with this. I would prefer having the API to be symmetric in both `left` and `right` arguments, so either cast both or none of them. We can have both `assert_series_equal` and `assert_native_series_equal` as you suggested. The second is quite low effort after the first

> Coercing the `right` or whatever is equivalent to `expected`; might make sense for ergonomics. E.g. like our `assert_equal_data`

We can always create a tests utility function for better ergonomics!

> https://github.com/narwhals-dev/narwhals/pull/2983#discussion_r2312063913

I actually didn't check pandas at all 🙈

dangotbanned: > I might disagree with this. I would prefer having the API to be symmetric in both left and right arguments, so either cast both or none of them.

Totally fair!

> I actually didn't check pandas at all 🙈

Now that surprised me!

I suspected the `_check_*` naming scheme and usage of `__tracebackhide__` came from there - but just an interesting coincidence it seems 😂

narwhals/testing/asserts/series.py
```diff
@@ -0,0 +1,291 @@
...
+        )
+    else:
+        _check_approximate_values(left_vals, right_vals, rel_tol=rel_tol, abs_tol=abs_tol)
+
+
+def _check_metadata(
+    left: SeriesT, right: SeriesT, *, check_dtypes: bool, check_names: bool
+) -> None:
+    """Check metadata information: implementation, length, dtype, and names."""
```
dangotbanned, line 121: How would you feel about moving every function from here onwards into `utils`?
https://github.com/narwhals-dev/narwhals/blob/9a4b605da3e7ef83555c94c413c8648b39865a92/narwhals/testing/asserts/utils.py

dangotbanned:
I do hope to get back to reviewing this soon - please nag me if I dont! 🙏

narwhals/testing/asserts/series.py
FBruzzesi: Not a strong preference but slightly in favor of keeping it like this until we are aware that something can be re-used also for dataframes. The reason for this is that say we move `_check_metadata`, it's very likely that dataframe metadata is composed of different check statements. Then we would end up with `_check_series_metadata` and `_check_frame_metadata`.

These functions are internal only, we can always move them around as we please

narwhals/testing/asserts/series.py
```diff
@@ -0,0 +1,291 @@
...
+            values in `right`.
+        abs_tol: Absolute tolerance for inexact checking.
+        categorical_as_str: Cast categorical columns to string before comparing.
+            Enabling this helps compare columns that do not share the same string cache.
+
+    Examples:
+        >>> import pandas as pd
+        >>> from narwhals.testing import assert_series_equal
+        >>> s1 = pd.Series([1, 2, 3])
+        >>> s2 = pd.Series([1, 5, 3])
```
FBruzzesi:
```suggestion
        >>> import pandas as pd
        >>> import narwhals as nw
        >>> from narwhals.testing import assert_series_equal
        >>> s1 = nw.from_native(pd.Series([1, 2, 3]), series_only=True)
        >>> s2 = nw.from_native(pd.Series([1, 5, 3]), series_only=True)
```

narwhals/testing/asserts/utils.py
```diff
@@ -1,13 +1,27 @@
...
+SeriesDetail: TypeAlias = Literal[
+    "implementation mismatch",
+    "length mismatch",
+    "dtype mismatch",
+    "name mismatch",
+    "null value mismatch",
+    "exact value mismatch",
+    "values not within tolerance",
+    "nested value mismatch",
+]
```
dangotbanned, line 22: [refactor(suggestion): Ensure consistent reasons?](https://github.com/narwhals-dev/narwhals/pull/2983/commits/a0592e1ba4d5db98ca176e187641e667f89c759f)

I noticed a couple were used more than once - so I thought this might help us keep them consistent in the future? 🙂

If as you've planned, we add `assert_frame_equal`, then we can probably share some of these between the two

*For now* it's just a lil bit of autocomplete 😄

FBruzzesi: I would partially argue that this seems a bit of an overkill?

I can see the advantage of

> For now it's just a lil bit of autocomplete

but otherwise if I was developing something and would not know any better, it would feel like I am constrained to use one of those detail value without a clear reason 🧐

narwhals/testing/asserts/frame.py
```diff
@@ -0,0 +1,22 @@
...
+    check_row_order: bool = True,
+    check_column_order: bool = True,
+    check_dtypes: bool = True,
+    check_exact: bool = False,
+    rel_tol: float = 1e-05,
+    abs_tol: float = 1e-08,
+    categorical_as_str: bool = False,
+) -> None:
+    msg = "TODO"  # pragma: no cover
+    raise NotImplementedError(msg)
```
dangotbanned: I've added (#3129) - could we remove the `assert_frame_equal` stuff from *this* PR for now?

FBruzzesi:
@MarcoGorelli gentle ping to get your review here :)

MarcoGorelli (commented):
thanks both, i haven't checked through everything but i'm on board with the idea, if you're both happy with it then feel free to ship it

narwhals/testing/asserts/utils.py
dangotbanned: You could add a `# NOTE` by the alias saying to extend it when adding new features?

PR #2985, opened 2025-08-13
fix: expr & series bool ambiguity

### LINKED ISSUE #2984: [Bug]: Expr and Series do not implement `__bool__`
### Describe the bug

All Narwhals expressions `bool` to True.

```python
import narwhals as nw

expr = nw.col('a')

print(bool(expr)) # True
```

This diverges from Polars, pandas, NumPy where Expr and Series objects where their respective `__bool__` methods are declared ambiguous.

As a specific example here is the implementation of [polars.Expr.\_\_bool\_\_](https://github.com/pola-rs/polars/blob/0031e4e21a95b2e2d23ef3f270cee64a7f4bb02b/py-polars/polars/expr/expr.py#L160-L171)

```python
import polars as pl

expr = pl.col('a')

bool(expr)
# TypeError: the truth value of an Expr is ambiguous
#
# You probably got here by using a Python standard library function instead of the native expressions API.
# Here are some things you might want to try:
# - instead of `pl.col('a') and pl.col('b')`, use `pl.col('a') & pl.col('b')`
# - instead of `pl.col('a') in [y, z]`, use `pl.col('a').is_in([y, z])`
# - instead of `max(pl.col('a'), pl.col('b'))`, use `pl.max_horizontal(pl.col('a'), pl.col('b'))`
```

And similar to Series

```python
import polars as pl

series = pl.Series('a')

bool(series)
# TypeError: the truth value of a Series is ambiguous
#
# Here are some things you might want to try:
# - instead of `if s`, use `if not s.is_empty()`
# - instead of `s1 and s2`, use `s1 & s2`
# - instead of `s1 or s2`, use `s1 | s2`
# - instead of `s in [y, z]`, use `s.is_in([y, z])`
```

By not raising an exception we introduce the possibility of mistakenly
triggering the objects default `__bool__` and receiving unintuitive results.
Here, I highlight examples using our Expr, but the Series also have similar issues.

```python
# literals are not representatives of their underlying values
import narwhals as nw

expr = nw.lit(False)
if expr: # → True; so the following is indeed printed
    print('But it’s a literal representing False')

# inappropriate flow control
if nw.col('a') == nw.col('b'):
    print('a and b all have the same values') # this is a lie

# combining with logical operators returns the logical short-circuited value
import pandas as pd
df = pd.DataFrame({'a': [True, False, False], 'b': [True, False, True]})
print(
    nw.from_native(df).filter(nw.col('a') and nw.col('b')) # logically reduces to .filter(nw.col('b'))
)

# Misleading equality checks
#   is a sub-issue of  `all([nw.col('a')])` → True; IIRC dataclasses implement all([left.{field} == right.{field}])
from dataclasses import dataclass

@dataclass
class T:
    expr: nw.Expr

print(
    f"{T(nw.col('a')) == T(nw.col('a')) = }", # True; alledging that the underlying expressions are equivalent
    f"{T(nw.col('a')) == T(nw.col('b')) = }", # True; alledging that the underlying expressions are equivalent
    sep='\n'
)
```

Lastly, our tests (in their current form) have some bugs from this:
- [tests/series_only/sort_test.py](https://github.com/narwhals-dev/narwhals/blob/64472d2048742e1b0d42ef50ff4c040c7157cf9a/tests/series_only/sort_test.py#L24-L31) the `assert` evaluates to `bool(result == expected)` which always returns True.
- [tests/frame/columns_test.py](https://github.com/narwhals-dev/narwhals/blob/64472d2048742e1b0d42ef50ff4c040c7157cf9a/tests/frame/columns_test.py#L24-L28) has a similar issue, relying on a regular `assert`

N/A

### Expected results

N/A

### Actual results

N/A

```shell
System:
    python: 3.12.7 (main, Oct 16 2024, 04:37:19) [Clang 18.1.8 ]
executable: /home/cameron/repos/opensource/narwhals-dev/.venv/bin/python
   machine: Linux-6.12.41-1-lts-x86_64-with-glibc2.42

Python dependencies:
     narwhals: 2.1.1
        numpy: 2.3.2
       pandas: 2.3.1
        modin: 
         cudf: 
      pyarrow: 21.0.0
      pyspark: 
       polars: 1.32.2
         dask: 
       duckdb: 1.3.0
         ibis: 10.8.0
     sqlframe: 3.38.2
```

### Relevant log output

```shell

```

camriddell:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2984
- Closes #2984

## Checklist

- [x] Tests added
- [ ] Documented the changes


FBruzzesi:
Thanks @camriddell this is definitely well spotted.

Regarding failures:

- marimo does the following check to verify that (narwhals) dataframes are equals: `a.to_dict() == b.to_dict()`. Passing the `as_series=False` flag should be enough to fix the issue ([source](https://github.com/marimo-team/marimo/blob/7fa89e77ad799c82ac0eee983bacca6e9d29d1ca/tests/_plugins/ui/_impl/tables/test_narwhals.py#L48-L49)). Update: submitted a PR https://github.com/marimo-team/marimo/pull/6035
- tubular they perform an `or` which (I assume) should be a `|` instead: `(nw.col(self.columns[0]).is_null()) or (nw.col(self.columns[1]).is_null())` ([source](https://github.com/azukds/tubular/blob/7678e0db3024e958bfe6fce8ecbc1216cd75dd11/tubular/dates.py#L441C17-L444C18)). Update: submitted a PR https://github.com/azukds/tubular/pull/469

So I would say that marimo has no harm, since it's only in their test suite, for tubular you found the bug you came here to fix in the first place

MarcoGorelli:
amazing, thanks for spotting this, i'd say ship it, clearly a bug

camriddell:
Given the strong movement behind this, I'll move this out of draft.

Also: @MarcoGorelli the pytest random versions failure for Python 3.9 w/ the following versions

```
 + numpy==1.20.3
 + pandas==1.2.5
 + polars==1.1.0
 + pyarrow==13.0.0
 + python-dateutil==2.9.0.post0
 + pytz==2025.2
 + six==1.17.0
```

Should be fixed, but perhaps opened in a separate PR?

FBruzzesi:
I was a victim: https://github.com/narwhals-dev/narwhals/pull/2998/commits/724729f42c8e068f25d8b6e7d2e1eb9ff2f8f3fc

TL;DR: I added a check `if self.len(): ...` with `self` being an expr 😭

FBruzzesi:
Both PRs mentioned in https://github.com/narwhals-dev/narwhals/pull/2985#issuecomment-3185392668 have been merged! Let's ship this. Thanks @camriddell 🔥

PR #2972, opened 2025-08-11
fix(typing): Add overloads for ibis `lit`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Major culprit in #2960

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
This is an example of a *small*, targeted upstream PR we could do.

For now, it covers a lot of our cases and makes the backend code easier to read - without needing to wait on acceptance elsewhere

PR #2986, opened 2025-08-13
fix: Workaround for `clip` with modin[pyarrow] backend

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related comment: https://github.com/narwhals-dev/narwhals/pull/2983#issuecomment-3185834687

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -814,6 +814,7 @@ class PandasLikeSeries(EagerSeries[Any]):
         lower_bound: Self | NumericLiteral | TemporalLiteral | None,
         upper_bound: Self | NumericLiteral | TemporalLiteral | None,
     ) -> Self:
+        native = self.native
         _, lower = (
             align_and_extract_native(self, lower_bound)
             if lower_bound is not None
@@ -824,8 +825,17 @@ class PandasLikeSeries(EagerSeries[Any]):
             if upper_bound is not None
             else (None, None)
         )
-        kwargs = {"axis": 0} if self._implementation is Implementation.MODIN else {}
-        return self._with_native(self.native.clip(lower, upper, **kwargs))
+        if self._implementation.is_modin():
+            result = native
+            if lower is not None:
+                result = result.where(result >= lower, lower)
+            if upper is not None:
+                result = result.where(result <= upper, upper)
+
+        else:
+            result = native.clip(lower, upper)
+
+        return self._with_native(result)
 
     def to_arrow(self) -> pa.Array[Any]:
         if self._implementation is Implementation.CUDF:
--- a/tests/expr_and_series/clip_test.py
+++ b/tests/expr_and_series/clip_test.py
@@ -28,8 +28,6 @@ def test_clip_expr(
 def test_clip_expr_expressified(
     request: pytest.FixtureRequest, constructor: Constructor
 ) -> None:
-    if "modin_pyarrow" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
     if "cudf" in str(constructor):
         # https://github.com/rapidsai/cudf/issues/17682
         request.applymarker(pytest.mark.xfail)
@@ -66,8 +64,6 @@ def test_clip_series(
 def test_clip_series_expressified(
     request: pytest.FixtureRequest, constructor_eager: ConstructorEager
 ) -> None:
-    if "modin_pyarrow" in str(constructor_eager):
-        request.applymarker(pytest.mark.xfail)
     if "cudf" in str(constructor_eager):
         # https://github.com/rapidsai/cudf/issues/17682
         request.applymarker(pytest.mark.xfail)


tests/expr_and_series/clip_test.py
```diff
@@ -28,8 +28,6 @@ def test_clip_expr(
 def test_clip_expr_expressified(
     request: pytest.FixtureRequest, constructor: Constructor
 ) -> None:
-    if "modin_pyarrow" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
     if "cudf" in str(constructor):
         # https://github.com/rapidsai/cudf/issues/17682
         request.applymarker(pytest.mark.xfail)
```
FBruzzesi: I would assume that the same trick might work for cudf. I will try it tomorrow on kaggle and update it here of that's the case

FBruzzesi: Yep that's indeed the case

narwhals/_pandas_like/series.py
```diff
@@ -824,8 +825,24 @@ def clip(
...
+        else:
+            native_cls = type(native)
+            if isinstance(lower, native_cls) or isinstance(upper, native_cls):
+                # Workaround for both cudf and modin when clipping with a series
+                #   * cudf: https://github.com/rapidsai/cudf/issues/17682
+                #   * modin: https://github.com/modin-project/modin/issues/7415
+                result = native
+                if lower is not None:
+                    result = result.where(result >= lower, lower)
+                if upper is not None:
```
FBruzzesi: I find the missing coverage for these lines to be odd. The `test_clip_expr_expressified` and  `test_clip_series_expressified` tests definitely enter these conditionals (I tried locally with a breakpoint) 🤔

dangotbanned: Weird

Can confirm the same for me when running 

```cmd
pytest tests --cov=narwhals --cov=tests --cov-fail-under=100 --runslow --constructors="pandas,pandas[nullable],pandas[pyarrow],pyarrow,modin[pyarrow],polars[eager],polars[lazy],dask,duckdb,sqlframe,ibis"
```

<details><summary>Definitely hit</summary>
<p>

image


</p>
</details>

FBruzzesi: Solved in https://github.com/narwhals-dev/narwhals/pull/2986/commits/8f5d08519bef7b1d36c8e6d2f32f3f201b9f999f, which also takes care of the mixed case of having one series and one literal (which I think we don't cover?)

narwhals/_pandas_like/series.py
```diff
@@ -814,6 +814,7 @@ def clip(
         lower_bound: Self | NumericLiteral | TemporalLiteral | None,
         upper_bound: Self | NumericLiteral | TemporalLiteral | None,
     ) -> Self:
+        native = self.native
```
dangotbanned: Needs to be done with the other suggestion, but out of range 😢 
```suggestion
```

narwhals/_pandas_like/series.py
```diff
@@ -824,8 +825,24 @@ def clip(
...
+            if isinstance(lower, native_cls):
+                result = result.where(result >= lower, lower)
+                lower = None
+            if isinstance(upper, native_cls):
+                result = result.where(result <= upper, upper)
+                upper = None
+            kwargs = {"axis": 0} if self._implementation.is_modin() else {}
+            result = result.clip(lower, upper, **kwargs)
+
+        return self._with_native(result)
```
dangotbanned: lil tweaks 🙂 

```suggestion
        kwargs: dict[str, Any] = {}
        impl = self._implementation
        result = self.native
        if not impl.is_pandas():
            # Workaround for both cudf and modin when clipping with a series
            #   * cudf: https://github.com/rapidsai/cudf/issues/17682
            #   * modin: https://github.com/modin-project/modin/issues/7415
            if impl.is_modin():
                kwargs = {"axis": 0}
            if self._is_native(lower):
                result = result.where(result >= lower, lower)
                lower = None
            if self._is_native(upper):
                result = result.where(result <= upper, upper)
                upper = None
        return self._with_native(result.clip(lower, upper, **kwargs))
```

dangotbanned (approved):
Thanks @FBruzzesi!

The two suggestions I've left are dependent - but GH wouldn't let me use a single suggestion

hungry narwhal

narwhals/_pandas_like/series.py
FBruzzesi: oh yes it's so much simpler! Thanks for suggesting. I will change together in the editor and add you before merging

dangotbanned: ahh thanks!

> and add you before merging

Did you mean to do one of these?

(https://github.com/narwhals-dev/narwhals/pull/2972/commits/9fea0a7bae2588db6267f62fe9c4e29180a24a11)

PR #2989, opened 2025-08-14
feat: add `separator` argument in `read_csv`/`scan_csv`

### LINKED ISSUE #2930: enh: add `separator` argument to `read_csv` / `scan_csv`
I don't really want to align all `read_csv` arguments, but `separator` is probably a pretty fundamental one which we could align between backends
- duckdb: `delimiter`
- polars: `separator`
- pandas: `sep`
- pyspark: `sep`

raisadz:
Closes #2930

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2930
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/functions.py
```diff
@@ -671,33 +683,39 @@ def scan_csv(
...
         Implementation.PANDAS,
         Implementation.MODIN,
         Implementation.CUDF,
         Implementation.DASK,
-        Implementation.DUCKDB,
         Implementation.IBIS,
     }:
-        native_frame = native_namespace.read_csv(source, **kwargs)
+        native_frame = native_namespace.read_csv(source, sep=separator, **kwargs)
+    elif implementation in {Implementation.DUCKDB}:
```
FBruzzesi:
```suggestion
    elif implementation is Implementation.DUCKDB:
```

narwhals/functions.py
```diff
@@ -594,17 +599,19 @@ def read_csv(
...
     elif impl is Implementation.PYARROW:
+        if separator is not None and "parse_options" in kwargs:
+            msg = "Can't pass both `separator` and `parse_options`."
+            raise TypeError(msg)
         from pyarrow import csv  # ignore-banned-import
 
-        native_frame = csv.read_csv(source, **kwargs)
+        native_frame = csv.read_csv(
+            source, parse_options=csv.ParseOptions(delimiter=separator), **kwargs
+        )
```
FBruzzesi: I think this is a bit odd:

1. `separator` is not typed to be `None`
2. Even if that was the case, the following would not error in line 607: 
    ```py
    nw.read_csv(..., separator=None, parse_option=csv.ParseOptions(...), backend=nw.Implementation.PYARROW)
    ```
    However, then in line 613, we would call
    ```py
    csv.read_csv(
            source, parse_options=csv.ParseOptions(delimiter=None), parse_options=parse_options, ...
        )
    ```
    which would end up raising an exception at this point

3. Should we handle the same for other backends? i.e. pandas-like check that `sep` is not passed, and below for lazy backends

FBruzzesi (commented):
Thanks @raisadz - I left a comment regarding the special check for pyarrow. I am afraid that it would not fully achieve the goal.

Maybe what we could do is:

1. check that separator is not None at the top of the function and raise otherwise
2. then at each backend level, check if the separator was passed with it's specific backend argument name and if that's the case raise a more informative error specifying to use `separator` instead of `sep|parse_options`.
3. Disclaimer: I am expecting no one to use this feature so far. Yet the only problem with that is that this is actually a regression: `parse_options` (together with `read_options`) is the only way for pyarrow to specify arguments in `read_csv`. Therefore we are enabling to pass separator in a standard way but basically disallowing to pass any other argument.
4. The _long way_ to do this is something along the following lines I think

```py
elif impl is Implementation.PYARROW:
    if "parse_options" in kwargs:
        passed_options = kwargs.pop("parse_options")
        fields = (
            'quote_char',
            'double_quote',
            'escape_char',
            'newlines_in_values',
            'ignore_empty_lines',
            'invalid_row_handler',
        )
        parse_options = csv.ParseOptions(
            delimiter=separator, **{field: getattr(passed_options, field) for field in fields}
        )
    else:
        parse_options = csv.ParseOptions(delimiter=separator)
            
    native_frame = csv.read_csv(
        source, parse_options=parse_options, **kwargs
    )
```

raisadz:
@FBruzzesi thank you for the review! I agree about the separators validation. I added some support functions that should check the passed `kwargs` now. In PyArrow, maybe we shouldn't type all the fields:
```
fields = (
            'quote_char',
            'double_quote',
            'escape_char',
            'newlines_in_values',
            'ignore_empty_lines',
            'invalid_row_handler',
        )
``` 

as they might go out of date if PyArrow changes them and we just need to check `delimiter`? Please, let me know what you think

narwhals/functions.py
```diff
@@ -559,8 +559,36 @@ def show_versions() -> None:
         print(f"{k:>13}: {stat}")  # noqa: T201
 
 
+def validate_separator(separator: str, native_separator: str, **kwargs: Any) -> None:
+    if native_separator in kwargs and kwargs[native_separator] != separator:
+        msg = (
+            "`separator` and `parse_options.delimiter` do not match: "
+            f"`separator`={separator} and `{native_separator}`={kwargs[native_separator]}."
+        )
+        raise TypeError(msg)
```
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/2989#issuecomment-3196763681
> in PyArrow, maybe we shouldn't type all the fields: ... as they might go out of date if PyArrow changes them and we just need to check delimiter? Please, let me know what you think

This was the kind of thing I was hoping to solve with

https://github.com/narwhals-dev/narwhals/issues/2930#issuecomment-3187996578
> Would this be easier to maintain if we moved `{read,scan}_csv` to be defined on `*Namespace`?

I *think* it would mean adding:

```py
CompliantNamespace.scan_csv
EagerNamespace.read_csv
PolarsNamespace.scan_csv
PolarsNamespace.read_csv
```

narwhals/functions.py
```diff
@@ -559,8 +559,36 @@ def show_versions() -> None:
         print(f"{k:>13}: {stat}")  # noqa: T201
 
 
+def validate_separator(separator: str, native_separator: str, **kwargs: Any) -> None:
+    if native_separator in kwargs and kwargs[native_separator] != separator:
+        msg = (
+            "`separator` and `parse_options.delimiter` do not match: "
```
FBruzzesi:
```suggestion
            f"`separator` and `{native_separator}` do not match: "
```

narwhals/functions.py
```diff
@@ -671,33 +704,40 @@ def scan_csv(
...
         Implementation.DASK,
-        Implementation.DUCKDB,
         Implementation.IBIS,
     }:
-        native_frame = native_namespace.read_csv(source, **kwargs)
+        validate_separator(separator, "sep", **kwargs)
+        native_frame = native_namespace.read_csv(source, sep=separator, **kwargs)
+    elif implementation is Implementation.DUCKDB:
+        validate_separator(separator, "delimiter", **kwargs)
+        validate_separator(separator, "delim", **kwargs)
```
FBruzzesi: Oh wow! TIL: duckdb and pyspark have two ways to pass the separator

narwhals/functions.py
```diff
@@ -559,8 +559,36 @@ def show_versions() -> None:
...
+        if parse_options.delimiter != separator:
+            msg = (
+                "`separator` and `parse_options.delimiter` do not match: "
+                f"`separator`={separator} and `delimiter`={parse_options.delimiter}."
+            )
+            raise TypeError(msg)
+        return kwargs
+    from pyarrow import csv  # ignore-banned-import
+
+    return {"parse_options": csv.ParseOptions(delimiter=separator)}
```
FBruzzesi, line 633: Nevermind I completely misread this

<details><summary> Fake panic review </summary>

The issue I have with this is that if any other argument was provided in `parse_options` then it will be silently ignored.

Say someone is calling the following:

```py
nw.read_csv(file, separator=",", parse_options=csv.ParseOptions(ignore_empty_lines=False)
```

Then at the end of `validate_separator_pyarrow`, we will end up with `csv.ParseOptions(delimiter=separator, ignore_empty_lines=True)` (i.e. the default value), silently.

I agree that hardcoding fields as suggested in https://github.com/narwhals-dev/narwhals/pull/2989#pullrequestreview-3126022407 is not ideal, yet pyarrow does not provide much else we can use. We could dynamically lookup its `__dir__` or use `inspect.get_members`, exclude dunder methods, but for example we would end up with `validate` and `equals`, which are not attribute to set at instantiation.

Unsuccessful tentatives I tried:

## [inspect.signature](https://docs.python.org/3/library/inspect.html#inspect.signature)

```py
from inspect import signature
from pyarrow import csv

print(signature(csv.ParseOptions.__init__))
```
> (self, /, *args, **kwargs)

## [dataclasses.fields](https://docs.python.org/3/library/dataclasses.html#dataclasses.fields)

From the stubs I got tricked into thinking that is a dataclass:

```py
from dataclasses import fields
from pyarrow import csv

print(fields(csv.ParseOptions))
```
> TypeError: must be called with a dataclass type or instance

</details>

I have mixed feelings as now a pyarrow user should pass both `separator=xyz, parse_options=csv.ParseOptions(delimiter=xyz)`

narwhals/functions.py
raisadz: Thanks @dangotbanned, it might be good to do the refactor you suggested but it probably should be separate and out of scope for this PR?

dangotbanned: Yeah that's fine don't worry! 

I just came here to comment that for the same reason as (https://github.com/narwhals-dev/narwhals/issues/2116#issuecomment-3197109427), we shouldn't do that until after #2944 anyway 😄

narwhals/functions.py
raisadz: Thanks @FBruzzesi!

> I have mixed feelings as now a pyarrow user should pass both separator=xyz, parse_options=csv.ParseOptions(delimiter=xyz)

A user won't need to pass both `separator` and `delimiter` as `if "parse_options" not in kwargs:` then we 
 `return {"parse_options": csv.ParseOptions(delimiter=separator)}`

narwhals/functions.py
```diff
@@ -564,8 +564,36 @@ def show_versions() -> None:
...
+def validate_separator(separator: str, native_separator: str, **kwargs: Any) -> None:
+    if native_separator in kwargs and kwargs[native_separator] != separator:
+        msg = (
+            f"`separator` and `{native_separator}` do not match: "
+            f"`separator`={separator} and `{native_separator}`={kwargs[native_separator]}."
+        )
+        raise TypeError(msg)
+
+
+def validate_separator_pyarrow(separator: str, **kwargs: Any) -> Any:
```
MarcoGorelli: can we mark both of these as private (prefix them with `_`)?

narwhals/functions.py
MarcoGorelli: > This was the kind of thing I was hoping to solve with
> https://github.com/narwhals-dev/narwhals/issues/2930#issuecomment-3187996578

@dangotbanned how does that comment address knowing which options are available in `csv.ParseOptions`, and them not changing in the future?

narwhals/functions.py
MarcoGorelli: right, so someone would only need to pass both `separator` and `delimiter` if they were specifying another parse option (like `double_quote`)

tbh I think this is fine

narwhals/functions.py
dangotbanned: > @dangotbanned how does that comment address knowing which options are available in `csv.ParseOptions`, and them not changing in the future?

My thought was that version branching for `pyarrow` would make more sense to be defined in `narwhals._arrow`, rather than in `narwhals.functions`

MarcoGorelli: sure, but i don't think we should version-branch on this at all. pyarrow could add an extra option and we'd have no way to know unless we closely monitored them

dangotbanned: I took at look at the docs and these options seem quite stable to me

- [v2](https://arrow.apache.org/docs/2.0/python/generated/pyarrow.csv.ParseOptions.html) was the earliest I found
- [v7](https://arrow.apache.org/docs/7.0/python/generated/pyarrow.csv.ParseOptions.html) was the only version that has ever introduced a new option
- [v21](https://arrow.apache.org/docs/python/generated/pyarrow.csv.ParseOptions.html) is the current and has identical options to our minimum [v13](https://arrow.apache.org/docs/13.0/python/generated/pyarrow.csv.ParseOptions.html)

If we *only* want the values for reconstruction we can use:

```py

from pyarrow.csv import ParseOptions

options = ParseOptions(
    delimiter="\t",
    quote_char="?",
    double_quote=False,
    escape_char="!",
    newlines_in_values=True,
)
>>> options.__getstate__()
('\t', '?', False, '!', True, True, None)
```

If we *only* want the option names we can use:

```py
import inspect
from pyarrow.csv import ParseOptions

>>> inspect.getmembers(ParseOptions, inspect.isdatadescriptor)
[('delimiter', <attribute 'delimiter' of 'pyarrow._csv.ParseOptions' objects>),
 ('double_quote',
  <attribute 'double_quote' of 'pyarrow._csv.ParseOptions' objects>),
 ('escape_char',
  <attribute 'escape_char' of 'pyarrow._csv.ParseOptions' objects>),
 ('ignore_empty_lines',
  <attribute 'ignore_empty_lines' of 'pyarrow._csv.ParseOptions' objects>),
 ('invalid_row_handler',
  <attribute 'invalid_row_handler' of 'pyarrow._csv.ParseOptions' objects>),
 ('newlines_in_values',
  <attribute 'newlines_in_values' of 'pyarrow._csv.ParseOptions' objects>),
 ('quote_char',
  <attribute 'quote_char' of 'pyarrow._csv.ParseOptions' objects>)]
```

I guess the overlap requires a bit more thinking, but it is possible 😄

MarcoGorelli: things always seem stable until they change, i'd rather not do this, let's keep it safe, it's not that taxing on the user

narwhals/functions.py
dangotbanned: Since our default is:

```py
separator: str = ","
```

and matches `pyarrow`'s default:

```py
delimiter: str = ","
```

## Alternative

`ParseOptions.delimiter` has higher precedence *unless* `separator` overrides the default.

In either case - every other argument is respected

<details><summary><b>Show definitions</b></summary>
<p>


```py

from pyarrow import csv
from typing import Any


def merge_options(separator: str = ",", **kwargs: Any) -> dict[str, Any]:
    DEFAULT = ","  # noqa: N806
    if separator != DEFAULT:
        if opts := kwargs.pop("parse_options", None):
            opts.delimiter = separator
        else:
            opts = csv.ParseOptions(delimiter=separator)
        kwargs["parse_options"] = opts
    return kwargs


def display_merge(result: dict[str, Any]) -> None:
    if result and (options := result.pop("parse_options", None)):
        print(f"{options.delimiter=}\n{options.double_quote=}")
        if result:
            print(f"Remaining: {result!r}")
    elif result:
        print(f"Unrelated: {result!r}")
    else:
        print(f"Empty: {result!r}")

```

</p>
</details> 

Would this behavior not be more ideal?

```py
# NOTE: `double_quote` default is `True`
user_options = csv.ParseOptions(delimiter="\t", double_quote=False)
```

```py
>>> display_merge(merge_options(parse_options=user_options))
options.delimiter='\t'
options.double_quote=False
```

```py
>>> display_merge(merge_options(",", parse_options=user_options))
options.delimiter='\t'
options.double_quote=False
```

```py
>>> display_merge(merge_options("?", parse_options=user_options))
options.delimiter='?'
options.double_quote=False
```

```py
>>> display_merge(merge_options())
Empty: {}
```

```py
>>> display_merge(merge_options("\t"))
options.delimiter='\t'
options.double_quote=True
```

```py
>>> display_merge(
    merge_options(
        "?",
        parse_options=csv.ParseOptions(double_quote=False),
        read_options=csv.ReadOptions(),
    )
)
options.delimiter='?'
options.double_quote=False
Remaining: {'read_options': <pyarrow._csv.ReadOptions object at 0x000001F29413AD40>}
```

Although it is `cython`, the important part is they're all properties with setters
https://github.com/apache/arrow/blob/f8b20f131a072ef423e81b8a676f42a82255f4ec/python/pyarrow/_csv.pyx#L435-L543

MarcoGorelli: > ParseOptions.delimiter has higher precedence unless separator overrides the default.

hmmm yes, that does sound better actually, thanks!

dangotbanned: Thanks Marco

with (https://github.com/narwhals-dev/narwhals/pull/2989#discussion_r2283160998) in mind ...

Not sure if this is on `duckdb` or `sqlframe`, but `sep` has higher precedence than `delim`

```py
import polars as pl
from pathlib import Path


data: Mapping[str, Any] = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}
fp = Path.cwd() / "data" / "file.csv"

pl.DataFrame(data).write_csv(fp, separator="\t")

session = DuckDBSession.builder.getOrCreate()
>>> session.read.format("csv").load(str(fp), sep="\t", delim="?").collect()
[Row(a=1, b=4.5, z='x'), Row(a=2, b=6.7, z='y'), Row(a=3, b=8.9, z='w')]
```

**Personally**, I think we're best off just defining rule(s) and documenting what we do for each backend if needed.

So instead of

```py
>>> nw.scan_csv("...", backend="sqlframe", separator=",", sep="?", delim="\t", delimiter="!")
TypeError: `separator` and `sep` do not match: `separator`=, and `sep`=?.
```

We either:
- pick one and replace it - leaving everything else unchanged
- say we'll pick ... then ... and then ...

If any backend raises on non-matching arguments - I say *let them* - as it saves us the hassle 😅

MarcoGorelli:
Hi - do you still have time / interest to work on this? If so, I think Dan's suggestion in https://github.com/narwhals-dev/narwhals/pull/2989#discussion_r2298169873 is good, and it may even simplify the implementation
If that's addressed, I think we can get it in?

dangotbanned:
The main source of conflicts will be coming from this PR

- #3024

100% recommend checking that out first, since the tests for the whole file have been rewritten (sorry 😅)

MarcoGorelli:
Looking at this again, I think I'd misunderstood @dangotbanned 's request, I'm not sure I agree with

>  unless separator overrides the default

This is Dan's suggestion, which I'd find surprising:
- `nw.scan_csv(..., separator=',', parse_options=ParseOptions(delim='\t'))` uses `'\t'` as separator
- `nw.scan_csv(..., separator='?', parse_options=ParseOptions(delim='\t'))` uses `'?'` as separator

This would the "native arguments always take precedence" approach, which would also be surprising:
- `nw.scan_csv(..., separator=',', parse_options=ParseOptions(double_quote=False))` uses `','` as separator
- `nw.scan_csv(..., separator='?', parse_options=ParseOptions(double_quote=False))` uses `','` as separator, because the parse options include `delim=','` as the default

What this PR suggests, on the other hand, is:
- `nw.scan_csv(..., separator=',', parse_options=ParseOptions(double_quote=False))` uses `','` as separator
- `nw.scan_csv(..., separator='?', parse_options=ParseOptions(double_quote=False))` raises, and the user is forced to write `nw.scan_csv(..., separator='?', parse_options=ParseOptions(double_quote=False, delim='?'))`

It's true that the current PR's approach is more verbose, but I think it's also the safest and least surprising

Sorry for the conflicting requests here - I'd suggest that if we resolve the merge conflicts then we can move forwards

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/2989#issuecomment-3421065872

@MarcoGorelli I'm gonna have to push back a little here.

Vs some I gave in (https://github.com/narwhals-dev/narwhals/pull/2989#discussion_r2298169873), your examples seem to be missing the fact that `,` is the default both for us and `pyarrow` (IIRC all the others as well?) 

I think a user is more likely to encounter the error by our default getting in the way, when writing something like:

```py
nw.scan_csv(..., parse_options=ParseOptions(another_option=..., delim='\t'))
```

Rather than *somehow* specifying this:

```py
nw.scan_csv(..., separator=',', parse_options=ParseOptions(delim='\t'))
```

---

I also don't like that (https://github.com/narwhals-dev/narwhals/pull/2989#issuecomment-3421065872) *can* break currently working code and would set a bad precedent (:wink:) for any time in the future we want to standardise on other `**kwargs`.
#2930 gave me the impression that this is something you may be interested in elsewhere

MarcoGorelli:
> your examples seem to be missing the fact that , is the default both for us and pyarrow

Not sure what you mean here, I wrote "because the parse options include delim=',' as the default"

> I also don't like that (https://github.com/narwhals-dev/narwhals/pull/2989#issuecomment-3421065872) can break currently working code

True, but you can do a search of its usage on github: https://github.com/search?q=%22nw.scan_csv%22&type=code&p=5. There's few enough cases of `nw.scan_csv` and `nw.read_csv` that it's possible to read though all of them, and none of them include `ParseOptions`

> our default getting in the way

I think this is OK. We can raise an informative error message for PyArrow that can make it very clear what the user is expected to do. For now this is the strictest and safest option, and I have general preference for starting strict and potentially relaxing later if necessary

If we eventually had Narwhals equivalents of all of `quote_char`, `double_quote`, `newline_in_values`, `ignore_empty_lines`, then that would mostly obviate the need for users passing in `ParseOptions` themselves anyway, so this issue would go away anyway

MarcoGorelli:
thinking about this again, i'm not wild on precedence being value-dependent (i.e. based on whether some default is matched or not)

i'm fairly confident that this won't break anyone's code, and that it offers the least surprising behaviour

narwhals/functions.py
```diff
@@ -711,32 +748,39 @@ def scan_csv(
...
         Implementation.DASK,
-        Implementation.DUCKDB,
         Implementation.IBIS,
     }:
-        native_frame = native_namespace.read_csv(source, **kwargs)
+        _validate_separator(separator, "sep", **kwargs)
+        native_frame = native_namespace.read_csv(source, sep=separator, **kwargs)
+    elif implementation is Implementation.DUCKDB:
+        _validate_separator(separator, "delimiter", **kwargs)
+        _validate_separator(separator, "delim", **kwargs)
```
MarcoGorelli: in light on the discussion, we also need to check against `'sep'` here

maybe `_validate_separator` should take a tuple of native delimiters?

tests/read_scan_test.py
```diff
@@ -185,3 +206,58 @@ def test_scan_fail_spark_like_without_session(
     pattern = re.compile(r"spark.+backend.+require.+session", re.IGNORECASE)
     with pytest.raises(ValueError, match=pattern):
         getattr(nw, scan_method)("unused.csv", backend=backend)
+
+
+def test_read_csv_raise_sep_multiple_lazy(csv_path: FileSource) -> None:
```
MarcoGorelli, line 212: to cut down on test times, shall we just use a single `csv_path` here rather than looping over `csv_path`? `csv_path` is already checked in other tests anyway

raisadz:
Thanks @MarcoGorelli ! I addressed your comments

MarcoGorelli:
thanks all for comments, much appreciated. at this point i'm still inclined enough to go along with it, i think the risk is minimal and that this is the easiest-to-teach behaviour

PR #2992, opened 2025-08-14
chore: simplify dask binary ops

MarcoGorelli:
it ain't much work, but it's honest work


dangotbanned:
Not sure if it improves anything, but for dask I think they define methods like `div`, `add`, etc?

(In a less forbidden way 😅)

Issue #2522, opened 2025-05-09
enh: add `is_elementwise` and `returns_scalar` to `map_batches`


MarcoGorelli:
currently tracking metadata for `map_batches` is really awkward

we should probably add these two arguments, and raise if the function output doesn't match expectations

this should probably go in _after_ https://github.com/narwhals-dev/narwhals/pull/2483

FBruzzesi:
@MarcoGorelli since #2998 would bring us half way through this issue, I wonder how `is_elementwise` would look like for pandas-like and pyarrow. Or would we let it exist just for compatibility for non polars backends?

MarcoGorelli:
it would just affect the result's ExprMetadata

PR #2998, opened 2025-08-15
feat: Add `returns_scalar` argument in `map_batches`

### LINKED ISSUE #2995: `map_batches` raises if function returns a scalar
```
FAILED tests/expr_and_series/map_batches_test.py::test_map_batches_expr_numpy[polars[eager]] - TypeError: `map` with `returns_scalar=False` must return a Series; found 'numpy.int64'.
```

This should be addressed by https://github.com/narwhals-dev/narwhals/issues/2522, but now it becomes higher prio

gonna start by xfailing

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2522
- Closes #2995

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -133,6 +133,8 @@ class CompliantExpr(
         self,
         function: Callable[[CompliantSeries[Any]], CompliantExpr[Any, Any]],
         return_dtype: IntoDType | None,
+        *,
+        returns_scalar: bool,
     ) -> Self: ...
     def broadcast(
         self, kind: Literal[ExprKind.AGGREGATION, ExprKind.LITERAL]
@@ -753,24 +755,41 @@ class EagerExpr(
         )
 
     def map_batches(
-        self, function: Callable[[Any], Any], return_dtype: IntoDType | None
+        self,
+        function: Callable[[Any], Any],
+        return_dtype: IntoDType | None,
+        *,
+        returns_scalar: bool,
     ) -> Self:
         def func(df: EagerDataFrameT) -> Sequence[EagerSeriesT]:
             input_series_list = self(df)
-            output_names = [input_series.name for input_series in input_series_list]
-            result = [function(series) for series in input_series_list]
-            if is_numpy_array(result[0]) or (
-                (np := get_numpy()) is not None and np.isscalar(result[0])
+            output_names = (input_series.name for input_series in input_series_list)
+            native_result = tuple(function(series) for series in input_series_list)
+            if is_numpy_array(native_result[0]) or (
+                (np := get_numpy()) is not None and np.isscalar(native_result[0])
             ):
                 from_numpy = partial(
                     self.__narwhals_namespace__()._series.from_numpy, context=self
                 )
                 result = [
                     from_numpy(array).alias(output_name)
-                    for array, output_name in zip(result, output_names)
+                    for array, output_name in zip(native_result, output_names)
                 ]
+            else:
+                result = native_result
             if return_dtype is not None:
                 result = [series.cast(return_dtype) for series in result]
+
+            is_scalar_result = tuple(len(r) == 1 for r in result)
+            if not returns_scalar and any(is_scalar_result) and self.len() > 1:
+                _idx = is_scalar_result.index(True)  # Index of first result with length 1
+                _type = type(native_result[_idx])
+                msg = (
+                    "`map_batches` with `returns_scalar=False` must return a Series; "
+                    f"found '{_type.__module__}.{_type.__name__}'.\n\nIf `returns_scalar` "
+                    "is set to `True`, a returned value can be a scalar value."
+                )
+                raise TypeError(msg)
             return result
 
         return self._from_callable(
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -498,6 +498,8 @@ class Expr:
         self,
         function: Callable[[Any], CompliantExpr[Any, Any]],
         return_dtype: DType | None = None,
+        *,
+        returns_scalar: bool = False,
     ) -> Self:
         """Apply a custom python function to a whole Series or sequence of Series.
 
@@ -510,6 +512,10 @@ class Expr:
             return_dtype: Dtype of the output Series.
                 If not set, the dtype will be inferred based on the first non-null value
                 that is returned by the function.
+            returns_scalar: If the function returns a scalar, by default it will be wrapped
+                in a list in the output, since the assumption is that the function always
+                returns something Series-like.
+                If you want to keep the result as a scalar, set this argument to True.
 
         Examples:
             >>> import pandas as pd
@@ -533,7 +539,9 @@ class Expr:
         # safest assumptions
         return self._with_orderable_filtration(
             lambda plx: self._to_compliant_expr(plx).map_batches(
-                function=function, return_dtype=return_dtype
+                function=function,
+                return_dtype=return_dtype,
+                returns_scalar=returns_scalar,
             )
         )
 
--- a/tests/expr_and_series/map_batches_test.py
+++ b/tests/expr_and_series/map_batches_test.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import pytest
 
 import narwhals as nw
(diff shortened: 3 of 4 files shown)


tests/expr_and_series/map_batches_test.py
```diff
@@ -28,11 +23,30 @@ def test_map_batches_expr_numpy(
...
     df = nw.from_native(constructor_eager(data))
     expected = nw.from_native(df.select(nw.all().map_batches(lambda x: x.to_numpy())))
     assert_equal_data(expected, {"a": [1, 2, 3], "b": [4, 5, 6], "z": [7.0, 8.0, 9.0]})
+
+
+def test_map_batches_exception(
+    constructor_eager: ConstructorEager, request: pytest.FixtureRequest
+) -> None:
+    if "polars" in str(constructor_eager) and POLARS_VERSION < (1, 32, 3):
+        request.applymarker(pytest.mark.xfail(strict=True))
```
FBruzzesi: Couldn't find a way to gracefully raise an exception for previous versions as even the `meta` namespace does not track if it's an aggregation

narwhals/_compliant/expr.py
```diff
@@ -753,24 +755,41 @@ def rolling_var(
...
+                result = native_result
             if return_dtype is not None:
                 result = [series.cast(return_dtype) for series in result]
+
+            is_scalar_result = tuple(len(r) == 1 for r in result)
+            if not returns_scalar and any(is_scalar_result) and self.len() > 1:
+                _idx = is_scalar_result.index(True)  # Index of first result with length 1
+                _type = type(native_result[_idx])
+                msg = (
+                    "`map_batches` with `returns_scalar=False` must return a Series; "
```
FBruzzesi, line 787: Well, polars message is "`map` with ..." which we might match but I thought it would not really make too much sense for a user

narwhals/_compliant/expr.py
```diff
@@ -753,24 +755,41 @@ def rolling_var(
...
+                    f"found '{_type.__module__}.{_type.__name__}'.\n\nIf `returns_scalar` "
```
FBruzzesi: Is there a easier way to get something like `numpy.int64`?

narwhals/_compliant/expr.py
```diff
@@ -753,24 +755,42 @@ def rolling_var(
...
             input_series_list = self(df)
-            output_names = [input_series.name for input_series in input_series_list]
-            result = [function(series) for series in input_series_list]
-            if is_numpy_array(result[0]) or (
-                (np := get_numpy()) is not None and np.isscalar(result[0])
+            output_names = (input_series.name for input_series in input_series_list)
+            native_result = tuple(function(series) for series in input_series_list)
+            result: Sequence[EagerSeriesT]
+            if is_numpy_array(native_result[0]) or (
+                (np := get_numpy()) is not None and np.isscalar(native_result[0])
```
MarcoGorelli: can we use `is_numpy_scalar`?

narwhals/_compliant/expr.py
```diff
@@ -753,24 +755,42 @@ def rolling_var(
...
-                    for array, output_name in zip(result, output_names)
-                ]
+                    for array, output_name in zip(native_result, output_names)
+                )
+            else:
+                result = native_result
             if return_dtype is not None:
                 result = [series.cast(return_dtype) for series in result]
+
+            is_scalar_result = tuple(len(r) == 1 for r in result)
```
MarcoGorelli, line 782: instead of checking it here and like this, could we do this in the `is_numpy_scalar` block, with something like
```
if is_numpy_scalar(...) and not returns_scalar:
    # raise
```
?

FBruzzesi: What if the result is not a scalar but not a numpy scalar? 

`native_result = tuple(function(series) for series in input_series_list)` is quite generic. Actually if there result is a scalar, even assuming that we can perform `len(r)` might lead to issues

MarcoGorelli: true, but it looks like we already don't support that
```python
import narwhals as nw
import pandas as pd
import numpy as np
from datetime import datetime
import pyarrow as pa

print(nw.from_native(pd.DataFrame({'a': [datetime(2020,1,1)]})).select(nw.col('a').map_batches(lambda x: x.mean(), return_dtype=nw.Int64, returns_scalar=True)))
```
```
  File "/home/marcogorelli/polars-api-compat-dev/narwhals/_compliant/expr.py", line 773, in func
    result = [series.cast(return_dtype) for series in result]
              ^^^^^^^^^^^
AttributeError: 'Timestamp' object has no attribute 'cast'
```

come to think of it though, we should support that

so this fix is probably fine for now, but i'll make a separate issue about non-numpy scalars

narwhals/expr.py
```diff
@@ -530,12 +536,18 @@ def map_batches(
...
+        def compliant_expr(plx: Any) -> Any:
+            return self._to_compliant_expr(plx).map_batches(
+                function=function,
+                return_dtype=return_dtype,
+                returns_scalar=returns_scalar,
             )
-        )
+
+        if returns_scalar:
+            return self._with_orderable_aggregation(compliant_expr)
```
FBruzzesi, line 548: @MarcoGorelli FYI I just added this commit. If `returns_scalar=True`, we should consider it an aggregation, don't we?

dangotbanned: I thought so too 🙂 

https://github.com/narwhals-dev/narwhals/blob/25549de09a392394488bb253ba8b001b6fc91952/narwhals/_plan/functions.py#L425-L440

https://github.com/narwhals-dev/narwhals/blob/25549de09a392394488bb253ba8b001b6fc91952/narwhals/_plan/options.py#L31-L32

FBruzzesi: Thanks! ~~I will merge for now and start the follow ups on this~~ just kidding, no one approved 😂

dangotbanned: If Marco doesn't beat me to it, I'll do my best to review ~~tonight/~~ tomorrow

MarcoGorelli: nice! yes, and we don't know if it's order-dependent, so orderable-aggregation seems correct, nice

narwhals/_compliant/expr.py
dangotbanned: Is this what you're looking for?

https://github.com/narwhals-dev/narwhals/blob/cea38259c7f4d55cd5fa729d1931f5b87841b9d0/narwhals/_utils.py#L1899-L1902

narwhals/_compliant/expr.py
```diff
@@ -753,24 +755,40 @@ def rolling_var(
...
         def func(df: EagerDataFrameT) -> Sequence[EagerSeriesT]:
             input_series_list = self(df)
-            output_names = [input_series.name for input_series in input_series_list]
-            result = [function(series) for series in input_series_list]
-            if is_numpy_array(result[0]) or (
-                (np := get_numpy()) is not None and np.isscalar(result[0])
-            ):
+            output_names = (input_series.name for input_series in input_series_list)
+            native_result = tuple(function(series) for series in input_series_list)
+            result: Sequence[EagerSeriesT]
```
dangotbanned: Could we rename these to make it clearer where they're used and how they're related?

E.g.
- `input_series_list` -> `udf_series_in`
- `native_result` -> `udf_series_out`

dangotbanned: I know it sounds nitpicky and the names were *already there* 😅 

I just found it difficult to keep all the context in my head while reading the method again with the new extension to it

tests/expr_and_series/map_batches_test.py
dangotbanned: Isn't this already our default?

```suggestion
        request.applymarker(pytest.mark.xfail)
```

https://github.com/narwhals-dev/narwhals/blob/44089f6d8b06a6deb8d389a5f0610234db2030ea/pyproject.toml#L250-L251

Issue #2965, opened 2025-08-09
Improving `DType` docs


dangotbanned:
## Related
- https://discord.com/channels/1235257048170762310/1237430779584446546/1390760611012546601
- https://discord.com/channels/1235257048170762310/1237430779584446546/1391015814143934545
- https://discord.com/channels/1235257048170762310/1272835535190364283/1397581137286529046
- https://github.com/narwhals-dev/narwhals/issues/2877#issuecomment-3108734301

## Description
Collecting some comments I keep forgetting to write up as an issue 🤦‍♂️ 

Generally, I think (#1077) was *too* narrowly scoped and that we could provide more on this topic:

> The `Enum` `DType` has a property named `categories`
> 
> Ah it seems like `nw.Struct` is the only `DType` that is showing members (`to_schema()`) 
> I'm thinking this was missed because there are cases like `nw.Array` that show a *Parameters* (*Arguments*) section for the constructor
> My guess is there's something to configure in either:
> - `docs/api-reference/dtypes`
> - `mdmkdocs.yml`

> (#2877) has me thinking we have a gap in the docs for `DTypes`:
> - It seems you need to go quite deep to find what they are
> - None of the methods are documented
> - I can't find any usage of them in the narrative docs

## Prior Art
Here's some links to upstream docs on data types.
For the most part I'm using the equivalent of [`nw.Struct`](https://narwhals-dev.github.io/narwhals/api-reference/dtypes/#narwhals.dtypes.Struct).

The choice is just for something with potential content to discuss. It wouldn't be a fair comparison to use [`pandas.ArrowDtype`](https://pandas.pydata.org/docs/reference/api/pandas.ArrowDtype.html) - so I just chose another `Dtype` for that instead:

- `polars`
  - [`polars.Struct`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Struct.html)
  - [concepts/data-types-and-structures](https://docs.pola.rs/user-guide/concepts/data-types-and-structures/#appendix-full-data-types-table)
  - [expressions/structs](https://docs.pola.rs/user-guide/expressions/structs/)
- `pandas` 
  - [`pandas.CategoricalDtype`](https://pandas.pydata.org/docs/reference/api/pandas.CategoricalDtype.html)
  - [user_guide/categorical](https://pandas.pydata.org/docs/user_guide/categorical.html)
- `pyarrow`
[...]

Issue #2877, opened 2025-07-23
[Enh]: Add `is_boolean` method to `DType`


kklein:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

I would like to treat a `nw.Series` differently based on its type. For some `DType`s (or groups thereof), there are convenient `is_x` methods. As far as I can tell, such a method doesn't exist yet for the `Boolean` type.

See here for a concrete, existing use case:
https://github.com/Quantco/metalearners/blob/main/metalearners/_utils.py#L380-L383

### Please describe the purpose of the new feature or describe the problem to solve.

See above.

### Suggest a solution if possible.

As was done for e.g. `is_integer` in [this commit](https://github.com/narwhals-dev/narwhals/commit/a35ea4ccc774049c3cd9b4b061b76edca9653ffb), one could simply add an `is_boolean` method which checks whether the instance at hand is an instance of [Boolean](https://github.com/narwhals-dev/narwhals/blob/main/narwhals/dtypes.py#L290).

### Additional information that may help us understand your needs.

I'd be happy to contribute this change!

MarcoGorelli:
sure, seems reasonable

> I'd be happy to contribute this change!

cool! PRs welcome, check the [contributing guide](https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md) for how to get started and/or reach out on discord

dangotbanned:
Thanks for raising!

> For some `DType`s (or groups thereof), there are convenient `is_x` methods. As far as I can tell, such a method doesn't exist yet for the `Boolean` type.

These are inherited from the methods on [`polars.DataType`](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.DataType.html#polars.datatypes.DataType) and the theme they *mostly* share is being a [shortcut](https://github.com/pola-rs/polars/blob/6e02c2067e6e36c3002339db82343d9e21232e22/py-polars/polars/datatypes/classes.py#L172-L215) for checking multiple `DataType`s.


So for example, `is_integer` is equivalent to:

```py
import narwhals as nw


dtype_int = nw.Int64()

>>> dtype_int.is_integer()
True

>>> isinstance(
    dtype_int,
    (
        nw.Int8,
        nw.Int16,
        nw.Int32,
        nw.Int64,
        nw.Int128,
        nw.UInt8,
        nw.UInt16,
        nw.UInt32,
[...]

PR #2999, opened 2025-08-15
docs: Include `DType` methods, properties in API Reference

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Part of #2965
  - See 97628 links there!

## Checklist

- [x] Tests added
- [x] Documented the changes

Did some re-aligning with [upstream](https://docs.pola.rs/api/python/version/1/reference/datatypes.html) as well

<details><summary>Outdated</summary>
<p>

What's currently in the PR I'm considering a **starting point** to chisel away at.
I'm happy with the *what*, but the repetition is waaaay too much 😔

So I'm very open to suggestions (even if they're just a link to other docs) on ways we can present it better

</p>
</details> 

Pretty happy with the new `dtypes` page, we've now got the hierarchy documented and examples of how `__eq__` works.

The main changes are shown in (https://github.com/narwhals-dev/narwhals/pull/2999#issuecomment-3193620241)

## Relevant `mkdocs` config
- https://mkdocstrings.github.io/python/usage/configuration/members/
- https://mkdocstrings.github.io/python/usage/configuration/docstrings/
- https://mkdocstrings.github.io/python/usage/configuration/signatures/

### DIFF
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -460,8 +460,26 @@ class Datetime(TemporalType, metaclass=_DatetimeMeta):
             To see valid strings run `import zoneinfo; zoneinfo.available_timezones()` for a full list.
         """
 
-    def __eq__(self, other: object) -> bool:
-        """Check if this DType is equivalent to another DType."""
+    def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
+        """Check if this Datetime is equivalent to another DType.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Datetime("s") == nw.Datetime("s")
+            True
+            >>> nw.Datetime("us") == nw.Datetime()
+            True
+            >>> nw.Datetime("us") == nw.Datetime("ns")
+            False
+            >>> nw.Datetime("us", "UTC") == nw.Datetime(time_unit="us", time_zone="UTC")
+            True
+            >>> nw.Datetime(time_zone="UTC") == nw.Datetime(time_zone="EST")
+            False
+            >>> nw.Datetime() == nw.Duration()
+            False
+            >>> nw.Datetime("ms") == nw.Datetime
+            True
+        """
         if type(other) is _DatetimeMeta:
             return True
         if isinstance(other, self.__class__):
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -615,7 +615,9 @@ class Enum(DType):
             >>> import narwhals as nw
             >>> nw.Enum(["a", "b", "c"]) == nw.Enum(["a", "b", "c"])
             True
-            >>> nw.Enum(["a", "b", "c"]) == nw.Enum(("a", "b", "c"))
+            >>> import polars as pl
+            >>> categories = pl.Series(["a", "b", "c"])
+            >>> nw.Enum(["a", "b", "c"]) == nw.Enum(categories)
             True
             >>> nw.Enum(["a", "b", "c"]) == nw.Enum(["b", "a", "c"])
             False
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -665,6 +665,7 @@ class Field:
     def __eq__(self, other: Field) -> bool:  # type: ignore[override]
         """Check if this Field is equivalent to another Field.
 
+        Two fields are equivalent if they have the same name and the same dtype.
         Examples:
             >>> import narwhals as nw
             >>> nw.Field("a", nw.String) == nw.Field("a", nw.String())
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -615,8 +615,26 @@ class Field:
         self.dtype = dtype
 
     def __eq__(self, other: Field) -> bool:  # type: ignore[override]
-        """Check if this DType is equivalent to another DType."""
-        return (self.name == other.name) & (self.dtype == other.dtype)
+        """Check if this Field is equivalent to another Field.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Field("a", nw.String) == nw.Field("a", nw.String())
+            True
+            >>> nw.Field("a", nw.String) == nw.Field("a", nw.String)
+            True
+            >>> nw.Field("a", nw.String) == nw.Field("a", nw.Datetime)
+            False
+            >>> nw.Field("a", nw.String) == nw.Field("b", nw.String)
+            False
+            >>> nw.Field("a", nw.String) == nw.String
+            False
+        """
+        return (
+            isinstance(other, Field)
+            and (self.name == other.name)
+            and (self.dtype == other.dtype)
+        )
 
     def __hash__(self) -> int:
         return hash((self.name, self.dtype))
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -467,7 +467,7 @@ class Datetime(TemporalType, metaclass=_DatetimeMeta):
             >>> import narwhals as nw
             >>> nw.Datetime("s") == nw.Datetime("s")
             True
-            >>> nw.Datetime("us") == nw.Datetime()
+            >>> nw.Datetime() == nw.Datetime("us")
             True
             >>> nw.Datetime("us") == nw.Datetime("ns")
             False
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -639,11 +639,24 @@ class Struct(NestedType):
             self.fields = list(fields)
 
     def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
-        """Check if this DType is equivalent to another DType."""
-        # The comparison allows comparing objects to classes, and specific
-        # inner types to those without (eg: inner=None). if one of the
-        # arguments is not specific about its inner type we infer it
-        # as being equal. (See the List type for more info).
+        """Check if this Struct is equivalent to another DType.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Struct({"a": nw.Int64}) == nw.Struct({"a": nw.Int64})
+            True
+            >>> nw.Struct({"a": nw.Int64}) == nw.Struct({"a": nw.Boolean})
+            False
+            >>> nw.Struct({"a": nw.Int64}) == nw.Struct({"b": nw.Int64})
+            False
+            >>> nw.Struct({"a": nw.Int64}) == nw.Struct([nw.Field("a", nw.Int64)])
+            True
+
+            If a parent type is not specific about its inner type, we infer it as equal
+
+            >>> nw.Struct({"a": nw.Int64}) == nw.Struct
+            True
+        """
         if type(other) is type and issubclass(other, self.__class__):
             return True
         if isinstance(other, self.__class__):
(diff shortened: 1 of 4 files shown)


dangotbanned:
## Previews (as of (7a47f12f389b94614487f4a781b5e00ee7eab0bb))
> I'm happy with the what, but the repetition is waaaay too much 😔

image


image


image

narwhals/dtypes.py
```diff
@@ -660,18 +686,18 @@ class List(NestedType):
...
     def __init__(self, inner: IntoDType) -> None:
         self.inner = inner
 
     def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
+        """Check if this DType is equivalent to another DType."""
         # This equality check allows comparison of type classes and type instances.
         # If a parent type is not specific about its inner type, we infer it as equal:
         # > list[i64] == list[i64] -> True
         # > list[i64] == list[f32] -> False
         # > list[i64] == list      -> True
```
dangotbanned: The comments like this are probably good content for the docstring

FBruzzesi:
Hey @dangotbanned I fully agree, that repetition is way too much. I tried locally the following changes which I think play nicely:

```diff
# `narwhals.dtypes`

::: narwhals.dtypes
    handler: python
    options:
      members:
+       - DType
        - NumericType
        - Decimal
        ...
        - Unknown
      show_root_heading: false
      show_source: false
-     show_bases: false
+     show_bases: true
      filters: ["!^_", "^__eq__$"]
-     inherited_members: false
+     inherited_members: false
```

The `__eq__` method for nested fields, datetime and duration (and other overwrites) will still display. Give it a spin, I think it's much leaner.

narwhals/dtypes.py
FBruzzesi: Yes please let's move that into either the docstring or even at the class description level

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/2999#issuecomment-3193500536)

10000000% I really like this!

## Previews ([proposed by @FBruzzesi](https://github.com/narwhals-dev/narwhals/pull/2999#issuecomment-3193500536))


> [!NOTE]
> See updated with examples in (https://github.com/narwhals-dev/narwhals/pull/2999#discussion_r2280525947)

image

image

image

image

narwhals/dtypes.py
dangotbanned: I quite like the look of this and it keeps the examples in the class and the method relatively short WDYT?

image

dangotbanned: In my head, it makes more sense to explain *overriding behavior* where it occurs 😅 

https://github.com/narwhals-dev/narwhals/blob/0c8fccba74d7cedd791839b2b9eb5cc9d7cbea8b/narwhals/dtypes.py#L694

FBruzzesi: Yes that's very nice! I agree let's keep them separate. Someone might not care about dtype equality at all, so no need to have additional spam on the class description 🙏🏼

narwhals/dtypes.py
```diff
@@ -82,41 +82,65 @@ def base_type(cls) -> type[Self]:
...
 
     @classmethod
     def is_boolean(cls: type[Self]) -> bool:
+        """Check whether the data type is a boolean type."""
         return issubclass(cls, Boolean)
 
     def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
+        """Check if this DType is equivalent to another DType.
+
+        Examples:
```
dangotbanned, line 131: Now the first thing you see, so the discoverability issue for (https://github.com/narwhals-dev/narwhals/issues/2877#issuecomment-3108734301) is fixed 🎉

image

narwhals/dtypes.py
```diff
@@ -426,10 +452,16 @@ def __init__(
...
+        Notes:
+            To see valid strings run `import zoneinfo; zoneinfo.available_timezones()` for a full list.
+        """
 
     def __eq__(self, other: object) -> bool:
-        # allow comparing object instances to class
+        """Check if this DType is equivalent to another DType."""
         if type(other) is _DatetimeMeta:
             return True
         if isinstance(other, self.__class__):
```
dangotbanned, line 467: `Datetime.__eq__` needs examples

narwhals/dtypes.py
dangotbanned: ## TODO (Examples)
- [x] `Datetime.__eq__`
- [x] `Duration.__eq__`
- [x] `Enum.__eq__`
- [x] `Field.__eq__`

dangotbanned:
## `Field`
I think this needs a correction on the description

https://github.com/narwhals-dev/narwhals/blob/78621131ef8b628f355bc58f6ac172afc04c885e/narwhals/dtypes.py#L617-L619

`Field` itself *isn't* a subclass of `DType`

https://github.com/narwhals-dev/narwhals/blob/78621131ef8b628f355bc58f6ac172afc04c885e/narwhals/dtypes.py#L592-L593

narwhals/dtypes.py
```diff
@@ -571,13 +606,16 @@ class Field:
...
+    """The name of the field within its parent `Struct`."""
     dtype: IntoDType
+    """The `DType` of the field's values."""
 
     def __init__(self, name: str, dtype: IntoDType) -> None:
         self.name = name
         self.dtype = dtype
 
     def __eq__(self, other: Field) -> bool:  # type: ignore[override]
+        """Check if this DType is equivalent to another DType."""
```
dangotbanned: RE: https://github.com/narwhals-dev/narwhals/pull/2999#issuecomment-3193821839

Lol I added it myself 🤦‍♂️

narwhals/dtypes.py
```diff
@@ -426,10 +452,34 @@ def __init__(
...
+        """Check if this Datetime is equivalent to another DType.
 
-    def __eq__(self, other: object) -> bool:
-        # allow comparing object instances to class
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Datetime("s") == nw.Datetime("s")
+            True
+            >>> nw.Datetime("us") == nw.Datetime()
+            True
```
FBruzzesi: I think this comparison might be a bit misleading. It evaluates to true _only_ because that's the `"us"` is the default value. I can see it easily being confused with `nw.Datetime("ms") == nw.Datetime`

narwhals/dtypes.py
```diff
@@ -426,10 +452,34 @@ def __init__(
...
+            >>> nw.Datetime("us") == nw.Datetime("ns")
+            False
+            >>> nw.Datetime("us", "UTC") == nw.Datetime(time_unit="us", time_zone="UTC")
+            True
+            >>> nw.Datetime(time_zone="UTC") == nw.Datetime(time_zone="EST")
+            False
+            >>> nw.Datetime() == nw.Duration()
+            False
```
FBruzzesi, line 479: The fact that Datetime and Duration have the same number of characters, start with `D` and a `t` in the middle should be illegal 😂 I was skimming through and had to double read it after the `False`

narwhals/dtypes.py
```diff
@@ -541,8 +608,24 @@ def categories(self) -> tuple[str, ...]:
...
-        # allow comparing object instances to class
+    def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
+        """Check if this Enum is equivalent to another DType.
+
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Enum(["a", "b", "c"]) == nw.Enum(["a", "b", "c"])
+            True
+            >>> nw.Enum(["a", "b", "c"]) == nw.Enum(("a", "b", "c"))
+            True
```
FBruzzesi: Duplicated 
```suggestion
            >>> nw.Enum(["a", "b", "c"]) == nw.Enum(["a", "b", "c"])
            True
```

narwhals/dtypes.py
```diff
@@ -571,14 +654,35 @@ class Field:
...
+    """The `DType` of the field's values."""
 
     def __init__(self, name: str, dtype: IntoDType) -> None:
         self.name = name
         self.dtype = dtype
 
     def __eq__(self, other: Field) -> bool:  # type: ignore[override]
-        return (self.name == other.name) & (self.dtype == other.dtype)
+        """Check if this Field is equivalent to another Field.
+
```
FBruzzesi, line 669: Examples are helpful and users will extrapolate easily, but a direct statement would also be nice. Something like:

```suggestion
        """Check if this Field is equivalent to another Field.

        Two fields are equivalent if they have the same name and the same dtype.
```

FBruzzesi (commented):
Thanks @dangotbanned - this is really great! I left a few comments but I suspect we will ship it soon 🎉

narwhals/dtypes.py
dangotbanned: Look closer! 😉

I was trying to show that the kind of iterable passed to the constructor isn't relevant

```py
["a", "b", "c"]  # left
("a", "b", "c")  # right
```

Maybe this would be clearer?

```py
import polars as pl
import narwhals as nw


>>> nw.Enum(["a", "b", "c"]) == nw.Enum(pl.Series(["a", "b", "c"]))
True
```


Or just drop it entirely

narwhals/dtypes.py
dangotbanned: What if I flip the operands?

```py
nw.Datetime() == nw.Datetime("us")
True
```

I think that reads more like

> No argument is equivalent to the default

narwhals/dtypes.py
FBruzzesi: Oh boy I really need to stop doing this late in the evening. Sorry @dangotbanned

dangotbanned: Don't be sorry!

I got halfway through fixing it before realising what I originally did that for 😂

narwhals/dtypes.py
FBruzzesi: > What if I flip the operands?

Yes I think it would be better. I imagine that a user thinks of these as `my_dtype_obj == <some target dtype>`

narwhals/dtypes.py
```diff
@@ -426,10 +452,34 @@ def __init__(
...
+    def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
+        """Check if this Datetime is equivalent to another DType.
 
-    def __eq__(self, other: object) -> bool:
-        # allow comparing object instances to class
+        Examples:
+            >>> import narwhals as nw
+            >>> nw.Datetime("s") == nw.Datetime("s")
+            True
+            >>> nw.Datetime("us") == nw.Datetime()
```
dangotbanned: self-suggest (https://github.com/narwhals-dev/narwhals/pull/2999/files/cccc9ca03e982c56e2518fe53072c5c4bd29c829#r2280627830)
```suggestion
            >>> nw.Datetime() == nw.Datetime("us")
```

narwhals/dtypes.py
dangotbanned: Done done done https://github.com/narwhals-dev/narwhals/pull/2999/files/cccc9ca03e982c56e2518fe53072c5c4bd29c829#r2280628264

narwhals/dtypes.py
dangotbanned: How about?

```suggestion
            >>> nw.Datetime() == nw.Date()
            False
```

FBruzzesi: Nah don't worry about this! I was sarcastic!

dangotbanned:
> Thanks @dangotbanned really amazing output!
> 

Thanks for the review and all of your help on this @FBruzzesi 😍

PR #3003, opened 2025-08-17
chore: replace `zip` with `zip equal`

raisadz:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue https://github.com/narwhals-dev/narwhals/pull/2977
- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -8,6 +8,7 @@ from enum import Enum, auto
 from functools import cache, lru_cache, wraps
 from importlib.util import find_spec
 from inspect import getattr_static, getdoc
+from itertools import chain
 from operator import attrgetter
 from secrets import token_hex
 from typing import (
@@ -1077,6 +1078,46 @@ def maybe_reset_index(obj: FrameOrSeriesT) -> FrameOrSeriesT:
             obj_any._compliant_series._with_native(native_obj.reset_index(drop=True))
         )
     return obj_any
+
+
+@overload
+def zip_equal(it1: Iterable[_T1], it2: Iterable[_T2], /) -> Iterable[tuple[_T1, _T2]]: ...
+
+
+@overload
+def zip_equal(*iterables: Iterable[Any]) -> Iterable[tuple[Any, ...]]: ...
+
+
+# https://stackoverflow.com/questions/32954486/zip-iterators-asserting-for-equal-length-in-python/69485272#69485272
+def zip_equal(*iterables: Iterable[Any]) -> Iterable[tuple[Any, ...]]:
+    # For trivial cases, use pure zip.
+    if len(iterables) < 2:
+        return zip(*iterables)
+
+    # Tail for the first iterable
+    first_stopped = False
+
+    def first_tail() -> Any:
+        nonlocal first_stopped
+        first_stopped = True
+        return
+        yield
+
+    # Tail for the zip
+    def zip_tail() -> Any:
+        if not first_stopped:
+            msg = "zip_equal: first iterable is longer"
+            raise ValueError(msg)
+        for _ in chain.from_iterable(rest):
+            msg = "zip_equal: first iterable is shorter"
+            raise ValueError(msg)
+            yield
+
+    # Put the pieces together
+    iterables_it = iter(iterables)
+    first = chain(next(iterables_it), first_tail())
+    rest = list(map(iter, iterables_it))
+    return chain(zip(first, *rest), zip_tail())
 
 
 def _is_range_index(obj: Any, native_namespace: Any) -> TypeIs[pd.RangeIndex]:
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -21,6 +21,7 @@ from narwhals._utils import (
     parse_columns_to_drop,
     scale_bytes,
     supports_arrow_c_stream,
+    zip_equal,
 )
 from narwhals.dependencies import is_numpy_array_1d
 from narwhals.exceptions import ShapeError
@@ -202,7 +203,7 @@ class ArrowDataFrame(
         return self.native.to_pylist()
 
     def iter_columns(self) -> Iterator[ArrowSeries]:
-        for name, series in zip(self.columns, self.native.itercolumns()):
+        for name, series in zip_equal(self.columns, self.native.itercolumns()):
             yield ArrowSeries.from_native(series, context=self, name=name)
 
     _iter_columns = iter_columns
@@ -216,7 +217,7 @@ class ArrowDataFrame(
         if not named:
             for i in range(0, num_rows, buffer_size):
                 rows = df[i : i + buffer_size].to_pydict().values()
-                yield from zip(*rows)
+                yield from zip_equal(*rows)
         else:
             for i in range(0, num_rows, buffer_size):
                 yield from df[i : i + buffer_size].to_pylist()
@@ -290,7 +291,7 @@ class ArrowDataFrame(
         schema = self.native.schema
         return {
             name: native_to_narwhals_dtype(dtype, self._version)
-            for name, dtype in zip(schema.names, schema.types)
+            for name, dtype in zip_equal(schema.names, schema.types)
         }
 
     def collect_schema(self) -> dict[str, DType]:
@@ -431,7 +432,7 @@ class ArrowDataFrame(
         else:
             sorting = [
                 (key, "descending" if is_descending else "ascending")
-                for key, is_descending in zip(by, descending)
+                for key, is_descending in zip_equal(by, descending)
             ]
 
         null_placement = "at_end" if nulls_last else "at_start"
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -28,7 +28,7 @@ from narwhals._compliant.typing import (
     LazyExprT,
     NativeExprT,
 )
-from narwhals._utils import _StoresCompliant
+from narwhals._utils import _StoresCompliant, zip_equal
 from narwhals.dependencies import get_numpy, is_numpy_array
 
 if TYPE_CHECKING:
@@ -280,13 +280,13 @@ class EagerExpr(
             if alias_output_names:
                 return [
                     series.alias(name)
-                    for series, name in zip(
+                    for series, name in zip_equal(
                         self(df), alias_output_names(self._evaluate_output_names(df))
                     )
                 ]
             return [
                 series.alias(name)
-                for series, name in zip(self(df), self._evaluate_output_names(df))
+                for series, name in zip_equal(self(df), self._evaluate_output_names(df))
             ]
 
         return self.__class__(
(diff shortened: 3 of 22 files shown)


raisadz:
@FBruzzesi I tried to replace `zip` with a stricter version of `zip_equal` but it failed for the `group_by` test `pytest tests/frame/group_by_test.py` It looks that you could have been working on it https://github.com/narwhals-dev/narwhals/pull/2325. Could you please take a look at it?

dangotbanned:
@raisadz I haven't checked out #2977 yet, but from skimming this PR - it seems like this is being used in a lot of places that have an equal length guarantee elsewhere.
For example, a `dict` has equal length keys and values

Which of these changes are required to fix an issue for #2977?

FBruzzesi:
> @FBruzzesi I tried to replace `zip` with a stricter version of `zip_equal` but it failed for the `group_by` test `pytest tests/frame/group_by_test.py` It looks that you could have been working on it #2325. Could you please take a look at it?

Hey @raisadz if I understand you correctly, you tried to replace every `zip` operation.

The issue in `group_by` (and not sure if somewhere else) is that in the following snippet `new_name` is used _only_ if the length is the same (i.e. if the key is not multi-output):

```py
safe_keys = [
    # multi-output expression cannot have duplicate names, hence it's safe to suffix
    key.name.map(_temporary_name)
    if (metadata := key._metadata) and metadata.expansion_kind.is_multi_output()
    # otherwise it's single named and we can use Expr.alias
    else key.alias(_temporary_name(new_name))
    for key, new_name in zip(keys, output_names)
]
```

The test that is failing evaluates to the first expression: `key.name.map(_temporary_name)`, since it's a multi output selector. Hence it does not really require the name to be considered in such point in time.

However this actually spots a hidden issue in case for which multiple "mixed" keys are passed. Let's consider the test case we have:

```py
def test_group_by_selector(constructor: Constructor) -> None:
    data = {"a": [1, 1, 1], "b": [4, 4, 6], "c": [7.5, 8.5, 9.0]}
    result = (
        nw.from_native(constructor(data))
        .group_by(nw.selectors.by_dtype(nw.Int64))
        .agg(nw.col("c").mean())
        .sort("a", "b")
    )
    expected = {"a": [1, 1], "b": [4, 6], "c": [8.0, 9.0]}
```

If we were to replace the group by statement with `.group_by(nw.selectors.by_dtype(nw.Int64), "c")`, then in the `safe_keys` evaluation we would have:

```py
keys = [nw.selectors.by_dtype(nw.Int64), nw.col("c")] 
output_names = ['a', 'b', 'c']
```

which is _bad_ as we would end up aliasing "c" as "b".

Let me try to address it in a dedicated PR.

---

Unrelated to the group by issue, could we rename `zip_equal` to `zip_strict`?

- From python 3.10, `zip` has a `strict` keyword
- We have a _naive_ implementation in the tests called `zip_strict` (the one in this PR is much better and more generic)

MarcoGorelli:
cool, thanks, definitely in favour of using strict zip everywhere by default, especially if it's helped uncover an issue

dangotbanned:
> cool, thanks, definitely in favour of using strict zip everywhere by default, especially if it's helped uncover an issue

I really disagree with this, the changes in (https://github.com/narwhals-dev/narwhals/pull/3003/files#diff-168a6da9cf3080d5790d562874fe82ff45411bc50b8a1bb90b6dc00ee19ea3ca) aren't safer - they just repeatedly check the same inputs are still the same length

I'm happy with strict zip when we actually don't know the length, but if we can't guarantee **anywhere** that we have equal length iterables - then there's a major design issue 🤔

MarcoGorelli:
what's the downside of using this?

dangotbanned:
> what's the downside of using this?

It adds overhead multiple times to the same operation

How many times is `zip_strict` called for this test?

https://github.com/narwhals-dev/narwhals/blob/a4d8ebf2e802ea70362004b5656ac0ea21f704e0/tests/selectors_test.py#L197-L225

MarcoGorelli:
true, but checking if it's necessary or not is a manual task and thus is error-prone. if we decide to use it everywhere, then it's a simple pre-commit check to enforce it

from what i can see, the overhead is minimal (and comparable to the rest of the Python operations we do) so I think I'd rather take this hit (if any, i'd be quite surprised if it actually showed up in a performance test, we're typically zipping over very few elements at a time anyway) and have it be safer. When we're Python3.10, we can just use `strict=True`

dangotbanned:
Again I'm not concerned about using this where we actually need to validate it

But there are a very large number of changes here that are blindly doing it
For example, why would this be helpful?

https://github.com/raisadz/narwhals/blob/de3ffa46e266ce611578f6db9f990a7f29a0209d/narwhals/_arrow/dataframe.py#L294

If `pyarrow.Schema` is in such a broken state that it doesn't have equal length names and types - `zip_strict` isn't saving us

```py
    @property
    def schema(self) -> dict[str, DType]:
        schema = self.native.schema
        return {
            name: native_to_narwhals_dtype(dtype, self._version)
            for name, dtype in zip_strict(schema.names, schema.types)
        }
```

https://github.com/raisadz/narwhals/blob/de3ffa46e266ce611578f6db9f990a7f29a0209d/narwhals/schema.py#L129-L149

This occurs directly after a length check
```py
        return {
            name: to_native_dtype(dtype=dtype, dtype_backend=backend)
            for name, dtype, backend in zip_strict(self.keys(), self.values(), backends)
        }
```

This creates things to `len`, then checks the length?

https://github.com/raisadz/narwhals/blob/de3ffa46e266ce611578f6db9f990a7f29a0209d/narwhals/_duckdb/utils.py#L329-L332

https://github.com/raisadz/narwhals/blob/de3ffa46e266ce611578f6db9f990a7f29a0209d/narwhals/_duckdb/utils.py#L298-L307

---

All I'm asking for is to do this in a **measured way** and 100% if it fixes an issue (https://github.com/narwhals-dev/narwhals/pull/3003#issuecomment-3194306197)

narwhals/_utils.py
```diff
@@ -1075,6 +1076,48 @@ def maybe_reset_index(obj: FrameOrSeriesT) -> FrameOrSeriesT:
...
+        for _ in chain.from_iterable(rest):  # pragma: no cover
+            msg = "zip_strict: first iterable is shorter"
+            raise ValueError(msg)
+            yield
+
+    # Put the pieces together
+    iterables_it = iter(iterables)
+    first = chain(next(iterables_it), first_tail())
+    rest = list(map(iter, iterables_it))
+    return chain(zip(first, *rest), zip_tail())
```
dangotbanned: This way we'd keep the C-speed for everything that supports `zip(strict=True)` 🙂

```suggestion
if TYPE_CHECKING:
    zip_strict = partial(zip, strict=True)
else:
    import sys

    if sys.version_info >= (3, 10):
        zip_strict = partial(zip, strict=True)
    else:  # https://stackoverflow.com/questions/32954486/zip-iterators-asserting-for-equal-length-in-python/69485272#69485272

        def zip_strict(*iterables: Iterable[Any]) -> Iterable[tuple[Any, ...]]:
            # For trivial cases, use pure zip.
            if len(iterables) < 2:
                return zip(*iterables)
            # Tail for the first iterable
            first_stopped = False

            def first_tail() -> Any:
                nonlocal first_stopped
                first_stopped = True
                return
                yield

            # Tail for the zip
            def zip_tail() -> Any:
                if not first_stopped:  # pragma: no cover
                    msg = "zip_strict: first iterable is longer"
                    raise ValueError(msg)
                for _ in chain.from_iterable(rest):  # pragma: no cover
                    msg = "zip_strict: first iterable is shorter"
                    raise ValueError(msg)
                    yield

            # Put the pieces together
            iterables_it = iter(iterables)
            first = chain(next(iterables_it), first_tail())
            rest = list(map(iter, iterables_it))
            return chain(zip(first, *rest), zip_tail())
```

FBruzzesi:
Adding my two cents in the conversation since this PR probably was triggered by a request I did to validate the input in `top_k`.

My understanding of the situation is the following:

- Most of the time polars signature says "iterable" (e.g. in [`top_k`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.top_k.html) we have `by: IntoExpr | Iterable[IntoExpr]`). However, does it really make sense? In this situation, I think it would make more sense to have the type as a sequence, after all it _has_ to be finite length since:
    ```py
     by
            Column(s) used to determine the top rows.
            Accepts expression input. Strings are parsed as column names.
    ```
- Therefore we should be able to perform a length check at the top level as I was suggesting in the PR. Which in my opinion would be better than raising a vague `"zip_strict: first iterable is (longer|shorter)"` message to the user at a certain point in the computation. 
- I agree with @dangotbanned that internally more often than not we have already verified sequence lengths before zipping sequences (or iterables) together. Yet considering the almost nullable overhead shown in the [stackoverflow thread](https://stackoverflow.com/questions/32954486/zip-iterators-asserting-for-equal-length-in-python/69485272#69485272) (I am impressed), it can be a good safe guard against ourselves (see the bug I introduced in `group_by`), but I would not rely on it blindly - we should be always careful in reviews and/or tests

narwhals/_utils.py
raisadz: @dangotbanned Thank you for the suggestion! If I commit this, can I use it everywhere?

dangotbanned: Yes!!!!!

@FBruzzesi, @MarcoGorelli 

I've cooled off and remembered I do actually want this 😂

Apologies everyone for the false alarm, this will do until we can have `ruff` do it for us 

- https://github.com/narwhals-dev/narwhals/issues/2084#issuecomment-2925189450

narwhals/_utils.py
```diff
@@ -1075,6 +1076,45 @@ def maybe_reset_index(obj: FrameOrSeriesT) -> FrameOrSeriesT:
     return obj_any
 
 
+if TYPE_CHECKING:
+    zip_strict = partial(zip, strict=True)
+else:
+    import sys
+
+    if sys.version_info >= (3, 10):
+        zip_strict = partial(zip, strict=True)
```
dangotbanned, line 1085: ooof 🤦‍♂️ 

```suggestion
from functools import partial

if TYPE_CHECKING:
    zip_strict = partial(zip, strict=True)
else:
    import sys

    if sys.version_info >= (3, 10):
        zip_strict = partial(zip, strict=True)
```

dangotbanned:
To get coverage, this *might* be a starting point


```py
def test_strict_zip(monkeypatch: pytest.MonkeyPatch) -> None:
    with monkeypatch.context() as mp:
        mp.setattr(sys, "version_info", (3, 9))

```

MarcoGorelli:
thanks all

we're already running the test suite with python 3.9 so i'd be ok with just pragma no covering the else branch

MarcoGorelli:
thanks all - let's ship this then!

PR #2933, opened 2025-08-03
feat: Add `Series.from_iterable`

dangotbanned:
Closes #2925

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Supersedes #2642
- Part of #2116

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## Tasks
- [x] Tests
  - [x] main
  - [x] `v1`
  - [x] `v2`
- [x] Docs
  - [x] Docstring
  - [x] API ref (https://github.com/narwhals-dev/narwhals/pull/2933/commits/d939ef75da738d9c935ecef0391eaa8e2f8a9279)
- [x] Fix `pyarrow` version
  - Found a bug with ([test: Check lots of iterables](https://github.com/narwhals-dev/narwhals/pull/2933/commits/efc5c4e1d5aad3fe6a1d5326e58a73d9ca65de72))
  - (https://github.com/narwhals-dev/narwhals/pull/2933/commits/328ecf03dc82af8cb3e732b553784b32cac7a221)
- [x] Misc fixes (https://github.com/narwhals-dev/narwhals/pull/2933#discussion_r2252225414)

PR #3007, opened 2025-08-18
feat: Warn on incompatible `TimeUnit` in `pandas<2`

### LINKED ISSUE #2964: warn or raise on explicit incompatible `TimeUnit`
## Related
- https://github.com/narwhals-dev/narwhals/pull/2933#issuecomment-3148461173
- https://github.com/narwhals-dev/narwhals/pull/2933#issuecomment-3168253108

## Description
I was caught out in a test where for `pandas<2`, we silently ignore `nw.Datetime(time_unit=...)` and always use `"ns"`.

It is *understandable* after reading up on the issue, and an [upcoming breaking change](https://pandas.pydata.org/docs/dev/whatsnew/v3.0.0.html#datetime-resolution-inference) in `pandas`.

However, we should be documenting what is happening - so that anyone trying to trace back an issue can find out the origin

benrutter:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2964
- Closes #2964

## Checklist

- [-] Tests added (this happens within a pragma no cover block - I'm happy to take off the no-cover and add a test, but I think it potentially could require a pre-v2 pandas tests, or some clever patching?)
- [x] Documented the changes

In order to tell between a default "us" and a specified "us", I've added in a `_defaulted` object variable to DateTime. This feels a little crufty, but I'm not sure I can see a better option?


benrutter:
⚠️ Note that might be important in a subtle way that I haven't realised ⚠️

In order to make this consistent type wise, I've actually *altered the narwhals/v1 type signature for dtypes*. This feels like it could be a red-flag depending how type-strict folks our? Essentially, a `None` value can now (and is by default) be passed to `time_unit`, which under the hood switched to `us` but indicates that the `time_unit` has been set by default rather than explicitly.

I'm pretty sure this behaviour doesn't actually change anything for anyone's code, and it definitely doesn't invalidate any existing code, but editing `stable/v1` makes me feel nervous 😅

narwhals/_pandas_like/utils.py
```diff
@@ -495,6 +496,13 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
         if is_pandas_or_modin(implementation) and PANDAS_VERSION < (
             2,
         ):  # pragma: no cover
+            if (not dtype._defaulted_time_unit) and dtype.time_unit != "ns":
```
dangotbanned:
```suggestion
            if isinstance(dtype, dtypes.Datetime):
```

I do understand that this would remove the distinction between:

```py
nw.Datetime()
nw.Datetime("us")
```

But there are other factors to consider here:

- We're warning, not raising an error
- This is only for `pandas<2` ([1.5.3](https://pandas.pydata.org/docs/whatsnew/v1.5.3.html))

However, the changes in `dtypes` affect all users, regardless of whether they have `pandas` installed 


I'm happy with the distinction between these two:

```py
nw.Datetime
nw.Datetime("us")  # which *includes* nw.Datetime()
```

And it means we can suggest using a bare `nw.Datetime` in the warning if the time unit is not relevant 🙂

pyproject.toml
```diff
@@ -243,6 +243,7 @@ filterwarnings = [
   'ignore:.*defaulting to pandas implementation',
   'ignore:.*implementation has mismatches with pandas',
   'ignore:.*You are using pyarrow version',
+  "ignore:.*A time unit other than 'ns' has been specified but is unsupported",
```
dangotbanned: We might end up adding this filter back, but if the warning is being triggered then we should have a test for it

```suggestion
```

narwhals/dtypes.py
```diff
@@ -438,20 +438,24 @@ class Datetime(TemporalType, metaclass=_DatetimeMeta):
         Datetime(time_unit='ms', time_zone='Africa/Accra')
     """
 
+    _defaulted_time_unit: bool = False
+
     def __init__(
-        self, time_unit: TimeUnit = "us", time_zone: str | timezone | None = None
+        self, time_unit: TimeUnit | None = None, time_zone: str | timezone | None = None
```
dangotbanned: Related to the comment in `_pandas_like.utils` - this change affects all users.

Now, the signature *alone* cannot be used to find out the default `time_unit`.

I think this isn't a good trade-off to warn a subset of users

```suggestion
        self, time_unit: TimeUnit = "us", time_zone: str | timezone | None = None
```

narwhals/_pandas_like/utils.py
```diff
@@ -495,6 +496,13 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
...
+                msg = (
+                    "A time unit other than 'ns' has been specified but is unsupported "
+                    "in Pandas versions preceding 2.0. To avoid an error, Narwhal's has "
+                    "fallen back to a time unit of 'ns'."
```
dangotbanned: In the warning I would like to see:

- **Which** `time_unit` is being ignored?
- **Why** is it being ignored?
  - *Possibly* a link to [pandas changelog](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#construction-with-datetime64-or-timedelta64-dtype-with-unsupported-resolution)
  - Or mention the required + found versions in the style of 
  https://github.com/narwhals-dev/narwhals/blob/d318ad31ba52391384eb8b409a43892c60ac95ab/narwhals/_utils.py#L1877
- **How** can I fix this?
  - Upgrading `pandas` is *an* option
  - but not the only one and probably not the easiest if they're stuck on a 2+ year old version 😄

dangotbanned (changes requested):
Thanks for the PR @benrutter!

> In order to tell between a default "us" and a specified "us", I've added in a _defaulted object variable to DateTime. This feels a little crufty, but I'm not sure I can see a better option?

What I was trying (and failed 😭) to suggest in (https://github.com/narwhals-dev/narwhals/issues/2964#issuecomment-3188811518) was intended to only change things in `_pandas_like`

Fully my bad on that one - hopefully the comments here help explain things better

narwhals/_pandas_like/utils.py
benrutter: I like this, thanks! I think I was potentially being too literal off the bat, it feels like one of those situations where in theory having a distinction between `nw.Datatime()` and `nw.Datetime("us")` is better, but in practice, the distinction isn't worthwhile in requiring wide changes to the `Datetime` type as a whole.

pyproject.toml
benrutter: I'd filtered it out because it seems like (unless it's just me) things are configured so that warnings fail the tests (meaning without this line, all the pandas<2 tests in CICD that trigger this warning fail - i.e. any tests that use, say `nw.DateTime("us")`).

I'm not sure there's a way around that, but I like the idea of adding in a test for the warning - have put one in just now.

narwhals/_pandas_like/utils.py
benrutter: Thanks, that's really helpful! I've updated the error to:
```
msg = (
    f"The time unit '{dtype.time_unit}' has been specified but is only "
    f"available in pandas>2.0, found version {pd.__version__}."
    "To avoid an error, Narwhal's has therefore fallen back to using "
    "'ns' as a time unit.\n"
    "To avoid this warning, consider upgrading pandas or otherwise "
    "specifying 'ns' time units."
)
```

I *think* that makes things clearer, but let me know if you'd change any of the wording!

pyproject.toml
dangotbanned: Thanks @benrutter!

Okay so as of [this run](https://github.com/narwhals-dev/narwhals/pull/3007/checks?check_run_id=49752246771) I'm happy with the tests

I just wanted to *warn* 😉 you that I'm going to remove that filter - fully expecting it to blow up CI 😄 
([ci: Temporarily remove warning filter](https://github.com/narwhals-dev/narwhals/pull/3007/commits/e96bac7165609a41ff8cf37d7bd0f2e66632d2a6))
Fixing those tests will *probably* be a separate issue (so don't worry!) 

A part of my motivation for this in (https://github.com/narwhals-dev/narwhals/pull/2933#issuecomment-3148461173) was that this being silent - made testing more complicated.
In my head, the ideal thing for us would be to switch any failing tests to use `nw.Datetime` (not `nw.Datetime()`) when the `time_unit` doesn't matter (thinking that'll be most cases)

narwhals/_pandas_like/utils.py
```diff
@@ -497,6 +498,16 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
         if is_pandas_or_modin(implementation) and PANDAS_VERSION < (
             2,
         ):  # pragma: no cover
+            if isinstance(dtype, dtypes.Datetime):
+                msg = (
+                    f"The time unit '{dtype.time_unit}' has been specified but is only "
+                    f"available in pandas>2.0, found version {pd.__version__}."
```
dangotbanned: Turns out my suggestion was *too* simplistic 🤦‍♂️ 

https://github.com/narwhals-dev/narwhals/actions/runs/17515585602/job/49752675800?pr=3007#step:9:490


> E       UserWarning: The time unit **'ns' has been specified** but is only available in pandas>2.0, found version 1.5.3.To avoid an error, Narwhal's has therefore fallen **back to using 'ns'** as a time unit.
> 
> E       To avoid this warning, consider upgrading pandas or otherwise **specifying 'ns'** time units.


In (https://github.com/narwhals-dev/narwhals/pull/3007#discussion_r2312385420) I focused too much on the `"us"` part and forgot about `"ns"` being fine

dangotbanned: I'll fix that goof next

## Update
Fixed in ([fix: Don't warn when `"ns"` was specified](https://github.com/narwhals-dev/narwhals/pull/3007/commits/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228))

pyproject.toml
dangotbanned: So I don't forget, (https://github.com/narwhals-dev/narwhals/pull/3007#discussion_r2327546683) produced **35** warnings but a chunk of those are false negatives I need to fix

dangotbanned: > produced **35** warnings but a chunk of those are false negatives I need to fix

Now down to **22** warnings ([run](https://github.com/narwhals-dev/narwhals/actions/runs/17515905732/job/49753350299?pr=3007))

## Update 1
That translates to **5** test functions

- **tests/v2_test.py**
  - `test_series_from_iterable` (https://github.com/narwhals-dev/narwhals/pull/2933#issuecomment-3148461173)
    - https://github.com/benrutter/narwhals/blob/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228/tests/v2_test.py#L441-L474
- **tests/expr_and_series/dt/timestamp_test.py**
  - `test_timestamp_datetimes`
    - https://github.com/benrutter/narwhals/blob/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228/tests/expr_and_series/dt/timestamp_test.py#L30-L73
  - `test_timestamp_datetimes_tz_aware`
    - https://github.com/benrutter/narwhals/blob/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228/tests/expr_and_series/dt/timestamp_test.py#L76-L140
- **tests/expr_and_series/cast_test.py**
  - `test_cast_datetime_tz_aware`
    - https://github.com/benrutter/narwhals/blob/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228/tests/expr_and_series/cast_test.py#L204-L234
  - `test_cast_datetime_utc`
    - https://github.com/benrutter/narwhals/blob/0c6c8e1401e3b668acd8933f77ff22dd7e1fe228/tests/expr_and_series/cast_test.py#L237-L265

dangotbanned: I've fixed all of those now here 🙂 

- [x] #3098

benrutter: Oh nice!

narwhals/_pandas_like/utils.py
dangotbanned: Okay with #3098 merged things are looking good

I'll take another look at the error message tomorrow

The main thing standing out is

> Narwhal's

Should be

> Narwhals

But I'll chew on the rest a bit in the meantime 😄

dangotbanned: This is what I'd suggest, was a bit more involved than a GH suggestion could be 😅 

([docs(suggestion): Tweaks to warning message](https://github.com/narwhals-dev/narwhals/pull/3007/commits/f33edbf283111667bb1f4144141c950010ed6c26))

Which comes out as:
```
UserWarning: `nw.Datetime(time_unit='us')` is only available in 'pandas>=2.0', found version '1.5'.
Narwhals has fallen back to using `time_unit='ns'` to avoid an error.

Hint: to avoid this warning, consider either:
- Upgrading pandas: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#construction-with-datetime64-or-timedelta64-dtype-with-unsupported-resolution
- Using a bare `nw.Datetime`, if this precision is not important
```

dangotbanned (approved):
Thanks for bearing with me on this @benrutter!


hungry narwhal

MarcoGorelli:
great feature, thanks all!

Issue #981, opened 2024-09-15
[Enh]: `mode` in grouped context


FBruzzesi:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

`mode` operation in group by context behave _very_ differently across pandas, polars and pyarrow. 

As for when `mode` was introduced, I am taking a mild look at skrub usecase.

### Please describe the purpose of the new feature or describe the problem to solve.

Consider the following snippet:

```py
import pandas as pd
import polars as pl
import pyarrow as pa

data = {
    "g1": [1, 1, 1, 1],
    "x1": [1, 1, 2, 3],
    "x2": [1, 1, 2, 2],
}
```

## Polars

Polars has a consistent behavior and, even if the column is unimodal, it will return a `list[T]`

```py
grp_pl = pl.DataFrame(data).group_by("g1")

print(grp_pl.agg(pl.col("x1").mode()))
print(grp_pl.agg(pl.col("x2").mode()))
```
```terminal
shape: (1, 2)
┌─────┬───────────┐
│ g1  ┆ x1        │
│ --- ┆ ---       │
│ i64 ┆ list[i64] │
╞═════╪═══════════╡
│ 1   ┆ [1]       │
└─────┴───────────┘
shape: (1, 2)
┌─────┬───────────┐
│ g1  ┆ x2        │
│ --- ┆ ---       │
│ i64 ┆ list[i64] │
╞═════╪═══════════╡
│ 1   ┆ [1, 2]    │
└─────┴───────────┘
```

## pandas

pandas fails for multi-modal results as it would not return a scalar.

```py
grp_pd = pd.DataFrame(data).groupby("g1")

print(grp_pd.agg(x1 = ("x1", pd.Series.mode)))  # uni-modal
```
```terminal
    x1
g1    
1    1
```

```py
grp_pd.agg(x2 = ("x2", pd.Series.mode))  # multi-modal
```
> ValueError: Must produce aggregated value

## pyarrow

I was not able to find a way to have the mode value directly

### Suggest a solution if possible.

Unsure on how to proceed

MarcoGorelli:
Interesting thanks - how do skrub use mode?

FBruzzesi:
> Interesting thanks - how do skrub use mode?

[...]

PR #3019, opened 2025-08-20
feat: Allow `mode(..., keep="any")`

### LINKED ISSUE #1793: feat: introduce `mode_any` aggregation
We currently have `Expr.mode` and `Series.mode`, like Polars does. What's not great about them is that they don't aggregate

I'd like to suggest a departure from Polars here. We introduce `Expr.mode_any` and `Series.mode_any`, which actually aggregate, but make no guarantees about _which_ mode will be returned. For example, if a column contains values `[1, 2, 2, 3, 3, 4]`, then it may return `2` or it may return `3`.

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #981 (_kinda_ closes it for those backend which support it)
- Closes #1793

## Checklist

- [x] Tests added
- [x] Documented the changes

This is an initial tentative work to introduce `keep` argument to `mode` operation.

- In non-grouped context it's working as expected for every backend.
    - Notice that SQL-like backend don't have an option to return all the modes if the values are multimodal
- In grouped context:
  - pyarrow does not support mode
  - pandas-like is _very very_ inconsistent. I did the minimum effort to make it work with a unimodal distribution and raise if the output is multimodal
  - In all honesty, I didn't try dask so far
- ~~For stable namespaces I kept the default to `keep="all"` so that there are no regressions, however changed it to `keep="any"` in the main namespace. For SQL-like I am failing early if `keep=="all"` since it would end up in the wrong result/expectation~~ Reverted in https://github.com/narwhals-dev/narwhals/pull/3019/commits/812e93c2614a1e028f0a26dd6172c2f948718d5b

### DIFF
--- a/tests/expr_and_series/mode_test.py
+++ b/tests/expr_and_series/mode_test.py
@@ -90,8 +90,10 @@ def test_mode_group_by_unimodal(
     df = nw.from_native(constructor(data_group))
     impl = df.implementation
 
-    if impl.is_pyarrow():
-        # https://github.com/apache/arrow/issues/20359
+    if impl.is_dask() or impl.is_pyarrow():
+        # Issue tracker:
+        #   - Dask: https://github.com/narwhals-dev/narwhals/pull/3019#issuecomment-3216649862
+        #   - PyArrow: https://github.com/apache/arrow/issues/20359
         request.applymarker(pytest.mark.xfail)
 
     result = (
@@ -122,8 +124,10 @@ def test_mode_group_by_multimodal(
     df = nw.from_native(constructor(data_group))
     impl = df.implementation
 
-    if impl.is_pyarrow():
-        # https://github.com/apache/arrow/issues/20359
+    if impl.is_dask() or impl.is_pyarrow():
+        # Issue tracker:
+        #   - Dask: https://github.com/narwhals-dev/narwhals/pull/3019#issuecomment-3216649862
+        #   - PyArrow: https://github.com/apache/arrow/issues/20359
         request.applymarker(pytest.mark.xfail)
 
     result = (
@@ -159,8 +163,10 @@ def test_mode_group_by_multiple_cols(
     df = nw.from_native(constructor(data_group))
     impl = df.implementation
 
-    if impl.is_pyarrow():
-        # https://github.com/apache/arrow/issues/20359
+    if impl.is_dask() or impl.is_pyarrow():
+        # Issue tracker:
+        #   - Dask: https://github.com/narwhals-dev/narwhals/pull/3019#issuecomment-3216649862
+        #   - PyArrow: https://github.com/apache/arrow/issues/20359
         request.applymarker(pytest.mark.xfail)
 
     result = df.group_by("grp").agg(mode_expr).sort("grp").lazy().collect()
--- a/narwhals/_dask/group_by.py
+++ b/narwhals/_dask/group_by.py
@@ -7,7 +7,7 @@ import dask.dataframe as dd
 
 from narwhals._compliant import DepthTrackingGroupBy
 from narwhals._expression_parsing import evaluate_output_names_and_aliases
-from narwhals._utils import Implementation, zip_strict
+from narwhals._utils import zip_strict
 
 if TYPE_CHECKING:
     from collections.abc import Mapping, Sequence
@@ -20,7 +20,6 @@ if TYPE_CHECKING:
     from narwhals._compliant.typing import NarwhalsAggregation
     from narwhals._dask.dataframe import DaskLazyFrame
     from narwhals._dask.expr import DaskExpr
-    from narwhals.typing import ModeKeepStrategy, NonNestedLiteral
 
     PandasSeriesGroupBy: TypeAlias = _PandasSeriesGroupBy[Any, Any]
     _AggFn: TypeAlias = Callable[..., Any]
@@ -74,40 +73,6 @@ def std(ddof: int) -> _AggFn:
     return partial(_DaskGroupBy.std, ddof=ddof)
 
 
-def mode(keep: ModeKeepStrategy) -> dd.Aggregation:
-    """Mode aggregation for Dask DataFrame grouped operations."""
-    if keep != "any":  # pragma: no cover
-        msg = (
-            f"`Expr.mode(keep='{keep}')` is not implemented in group by context for "
-            f"backend {Implementation.DASK}\n\n"
-            "Hint: Use `nw.col(...).mode(keep='any')` instead."
-        )
-        raise NotImplementedError(msg)
-
-    def chunk(s: PandasSeriesGroupBy) -> pd.Series[Any]:
-        return s.value_counts()
-
-    def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
-        # s0._selected_obj holds the concatenated counts Series
-        return s0._selected_obj.groupby(
-            level=list(range(s0._selected_obj.index.nlevels))
-        ).sum()
-
-    def finalize(s1: PandasSeriesGroupBy) -> pd.Series[Any]:
-        # s1 is a Series indexed by (group_key, value)
-        # First identify max counts per group
-        levels = list(range(s1.index.nlevels - 1))  # group levels
-
-        def choose_mode(x: pd.Series[Any]) -> NonNestedLiteral:
-            modes = x[x == x.max()]
-            # multilevel index, take the value in last position
-            return modes.idxmax()[-1]  # pyright: ignore[reportIndexIssue]
-
-        return s1.groupby(level=levels).apply(choose_mode)
-
-    return dd.Aggregation(name="mode", chunk=chunk, agg=agg, finalize=finalize)
-
-
 class DaskLazyGroupBy(DepthTrackingGroupBy["DaskLazyFrame", "DaskExpr", Aggregation]):
     _REMAP_AGGS: ClassVar[Mapping[NarwhalsAggregation, Aggregation]] = {
         "sum": "sum",
@@ -115,7 +80,6 @@ class DaskLazyGroupBy(DepthTrackingGroupBy["DaskLazyFrame", "DaskExpr", Aggregat
         "median": "median",
         "max": "max",
         "min": "min",
-        "mode": mode,
         "std": std,
         "var": var,
         "len": "size",


docs/backcompat.md
```diff
@@ -117,7 +117,11 @@ Which should you use? In general we recommend:
 
 ## `main` vs `stable.v2` differences
 
-So far, nothing, everything non-unstable from the main namespace should be available in `narwhals.stable.v2`.
+- Since Narwhals 2.x.y:
```
FBruzzesi: TODO

narwhals/_polars/series.py
```diff
@@ -518,6 +519,10 @@ def is_close(
         )
         return self._with_native(result)
 
+    def mode(self, *, keep: ModeKeepStrategy) -> Self:
+        result = self.native.mode()
+        return self._with_native(result.head(1) if keep == "any" else result)
```
FBruzzesi, line 524: Using `head` to return a Series (`first` returns a scalar)

tests/expr_and_series/mode_test.py
```diff
@@ -1,31 +1,114 @@
...
+def test_mode_expr_keep_any(constructor: Constructor) -> None:
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.col("a", "b").mode(keep="any"))
+
+    try:
+        expected = {"a": [1], "b": [3]}
+        assert_equal_data(result, expected)
+    except AssertionError:
+        expected = {"a": [2], "b": [3]}
+        assert_equal_data(result, expected)
```
FBruzzesi: Any better ideas to test non-deterministic multimodal case?

docs/backcompat.md
MarcoGorelli: are we sure we want to change the default? for compatibility with polars api, maybe we can keep the same default that polars has? (so for pyspark/etc you in effect be required to specify `keep='any'`)

FBruzzesi: ~~Definitely much simpler to keep it all the same 👌🏼~~

Changed back in https://github.com/narwhals-dev/narwhals/pull/3019/commits/812e93c2614a1e028f0a26dd6172c2f948718d5b

narwhals/_pandas_like/group_by.py
```diff
@@ -115,6 +120,29 @@ def _getitem_aggs(
...
+            nw_dtypes = (
+                native_to_narwhals_dtype(native_dtypes[col], version, impl)  # type: ignore[index, misc]
+                if native_dtypes[col] != "object"  # type: ignore[index, misc]
+                else object_native_to_narwhals_dtype(result[col], version, impl)  # type: ignore[arg-type]
+                for col in result.columns
+            )
+
+            if any(dtype == version.dtypes.Object() for dtype in nw_dtypes):
+                msg = "TODO"
+                raise NotImplementedError(msg)
```
FBruzzesi: The assumption is that if the output type is Object, then the result is multimodal. If `keep="any"` it would be nice to keep only one element per group. But consider the following two cases:

- Case 1: Multimodal numeric values:

```py
import pandas as pd

data = {"group": [1, 1, 2], "values": [1, 2, 2]}
pd.DataFrame(data).groupby("group")["values"].agg(pd.Series.mode)
```

> ValueError: Must produce aggregated value

- Case 2: Multimodal non-numeric values

```py
import pandas as pd

data = {"group": [1, 1, 2], "values": ["foo", "bar", "baz"]}
pd.DataFrame(data).groupby("group")["values"].agg(pd.Series.mode)
```

which results in a object dtype:

```terminal
group
1    [bar, foo]
2           baz
Name: values, dtype: object
```

How should we extract one value from each group? As you can see in one case we have a list, in another a string. Not sure if there is a _clean/efficient_ way of doing it

narwhals/_pandas_like/group_by.py
```diff
@@ -115,6 +120,29 @@ def _getitem_aggs(
...
                 [ns.from_native(result_single).alias(name).native for name in names]
             )
+        elif self.is_mode():
+            if (keep := self.kwargs.get("keep")) != "any":  # pragma: no cover
+                msg = f"`keep='{keep}' is not implemented in group by context"
+                raise NotImplementedError(msg)
+
+            select = names[0] if len(names) == 1 else list(names)
+            Series = self.expr.__native_namespace__().Series
+            result = group_by._grouped[list(names)].agg(Series.mode)
```
MarcoGorelli: I think `agg(Series.mode)` this is effectively going to be as slow as apply+lambda function

there's a suggestion in https://github.com/pandas-dev/pandas/issues/19254#issuecomment-778661578, does that work?
otherwise, how about the `value_counts` one followed by a filter?

narwhals/_pandas_like/group_by.py
dangotbanned: Would `mode` work with either `transform` or `filter`?

- https://pandas.pydata.org/docs/dev/user_guide/groupby.html#transformation
- https://pandas.pydata.org/docs/dev/user_guide/groupby.html#filtration

narwhals/_pandas_like/group_by.py
FBruzzesi: Thanks for the suggestion! Happy to see we are not the first ones fighting this (lack of) feature.

Committed in https://github.com/narwhals-dev/narwhals/pull/3019/commits/f2b970d8e7439f51d8c82e934b62d482ada5e717

[`DataFrameGroupBy.value_counts`](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.value_counts.html) was introduces in pandas 1.4, I left a comment mentioning that

narwhals/_pandas_like/group_by.py
```diff
@@ -115,6 +117,42 @@ def _getitem_aggs(
...
+            #     group_by._grouped.value_counts()
+            #     .sort_values(ascending=False)
+            #     .reset_index(cols)
+            #     .groupby(group_by._keys, **group_by_kwargs)[cols]
+            #     .head(1)
+            #     .sort_index()
+            # )
+            # ```
+            result = (
+                native.groupby([*group_by._keys, *cols], **group_by_kwargs)
```
FBruzzesi: TODO: This logic will yield wrong results if multiple columns are provided in `cols`. Need to address it and have a test

narwhals/_pandas_like/group_by.py
```diff
@@ -115,6 +117,42 @@ def _getitem_aggs(
...
+                raise NotImplementedError(msg)
+
+            cols = list(names)
+            native = group_by.compliant.native
+            group_by_kwargs = {
+                "sort": False,
+                "as_index": True,
+                "dropna": group_by._drop_null_keys,
+                "observed": True,
+            }
```
dangotbanned: I could be very wrong, but I vaguely remember some of these being accessible on the (native) `DataFrameGroupBy`?

If not, then besides `dropna` - you could define them as a constant to be used here and in `PandasLikeGroupBy.__init__`?

narwhals/_dask/expr.py
```diff
@@ -663,7 +672,6 @@ def dt(self) -> DaskExprDateTimeNamespace:
     gather_every: not_implemented = not_implemented()
     head: not_implemented = not_implemented()
     map_batches: not_implemented = not_implemented()
-    mode: not_implemented = not_implemented()
```
dangotbanned, line 666: 🥳

narwhals/_polars/series.py
MarcoGorelli: 🤔 maybe we'd want a scalar though, that's what other aggregations (e.g. `Series.sum`) return

MarcoGorelli (commented):
thanks @FBruzzesi !

looks good, if we can just get `Series.mode('any')` to return a scalar (like other aggregations do), then I think we can ship it

FBruzzesi:
Thanks for the review @MarcoGorelli, I will take a look at Dask in the grouped context next, but the PR is pretty much ready (added via custom aggregation in https://github.com/narwhals-dev/narwhals/pull/3019/commits/a36ff062e0546aaf9c0619d87c0919ee1da41db2)

---

I wonder if it's worth trying to apply the pandas group by _hack_ in the pyarrow case.

Edit: PyArrow does not have the "new" group by infrastructure that pandas has. I think we can park this implementation for now

narwhals/_pandas_like/group_by.py
```diff
@@ -176,6 +205,9 @@ class PandasLikeGroupBy(
     _output_key_names: list[str]
     """Stores the **original** version of group keys."""
 
+    _kwargs: dict[str, bool]
```
dangotbanned: IIUC, we only need to be able to do `**_kwargs`

```suggestion
    _kwargs: Mapping[str, bool]
```

This way, we'd get a warning if another PR tried to mutate this (which would probably break a lot 😄)

FBruzzesi: Thanks @dangotbanned

narwhals/_pandas_like/group_by.py
FBruzzesi: I preferred storing in a `_kwargs` attribute rather than accessing grouped object, thanks for suggesting

narwhals/_polars/series.py
FBruzzesi: Resolved in https://github.com/narwhals-dev/narwhals/pull/3019/commits/ff9a842c3da328bdf882d94ec10f1f836d08bf7c

MarcoGorelli:
thanks @FBruzzesi !

there's some failures for dask nightly:
```python
FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_unimodal[dask] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multiple_cols[dask-mode_expr0] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multimodal[dask-vals_multimodal_num-expected_opt_10-expected_opt_20] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multimodal[dask-vals_multimodal_str-expected_opt_11-expected_opt_21] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multiple_cols[dask-mode_expr1] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
```

tbh i'd be ok with just not supporting this for dask, not too bothered

FBruzzesi:
> thanks @FBruzzesi !
> 
> there's some failures for dask nightly:
> 
> ```python
> FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_unimodal[dask] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
> FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multiple_cols[dask-mode_expr0] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
> FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multimodal[dask-vals_multimodal_num-expected_opt_10-expected_opt_20] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
> FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multimodal[dask-vals_multimodal_str-expected_opt_11-expected_opt_21] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
> FAILED tests/expr_and_series/mode_test.py::test_mode_group_by_multiple_cols[dask-mode_expr1] - ValueError: Cannot remove 2 levels from an index with 2 levels: at least one level must be left.
> ```
> 
> tbh i'd be ok with just not supporting this for dask, not too bothered

I went to bed after the last commit yesterday! I am taking a look at it right now. Just FYI, it does not fail if pandas is not nightly aswell

FBruzzesi:
Ok I am rolling back for dask since I couldn't make it work for nightly pandas.

For future reference, this is the `mode` implementation that works for dask before pandas 3.0.0
If someone want to take a stab at it, I think it's a nice place to start.

```py
def mode(keep: ModeKeepStrategy) -> dd.Aggregation:
    """Mode aggregation for Dask DataFrame grouped operations."""
    if keep != "any":  # pragma: no cover
        msg = (
            f"`Expr.mode(keep='{keep}')` is not implemented in group by context for "
            f"backend {Implementation.DASK}\n\n"
            "Hint: Use `nw.col(...).mode(keep='any')` instead."
        )

    def chunk(s: PandasSeriesGroupBy) -> pd.Series[Any]:
        return s.value_counts()

    def agg(s0: PandasSeriesGroupBy) -> pd.Series[Any]:
        # s0._selected_obj holds the concatenated counts Series
        return s0._selected_obj.groupby(
            level=list(range(s0._selected_obj.index.nlevels))
        ).sum()

    def finalize(s1: PandasSeriesGroupBy) -> pd.Series[Any]:
        # s1 is a Series indexed by (group_key, value)
        # First identify max counts per group
        levels = list(range(s1.index.nlevels - 1))  # group levels

        def choose_mode(x: pd.Series[Any]) -> NonNestedLiteral:
            modes = x[x == x.max()]
            # multilevel index, take the value in last position
            return modes.idxmax()[-1]  # pyright: ignore[reportIndexIssue]

        return s1.groupby(level=levels).apply(choose_mode)

    return dd.Aggregation(name="mode", chunk=chunk, agg=agg, finalize=finalize)
```

---

For context, this is totally on point https://github.com/narwhals-dev/narwhals/pull/3019#issuecomment-3216659644

>  `_selected_obj` is internal so we shouldn't be using it

but that's not even the issue. The error raises at the first step in `return s.value_counts()`

MarcoGorelli:
thanks for rolling back!

yeah `_selected_obj` is internal so we shouldn't be using it

tests/expr_and_series/mode_test.py
```diff
@@ -1,31 +1,187 @@
...
+def test_mode_expr_keep_any(constructor: Constructor) -> None:
+    df = nw.from_native(constructor(data))
+    result = df.select(nw.col("a", "b").mode(keep="any")).lazy().collect()
+
+    try:
+        expected = {"a": [1], "b": [3]}
+        assert_equal_data(result, expected)
+    except AssertionError:  # pragma: no cover
+        expected = {"a": [2], "b": [3]}
+        assert_equal_data(result, expected)
```
MarcoGorelli, line 62: 😄 love this

Issue #3023, opened 2025-08-22
Support `DataFrame().lazy({'pyspark','sqlframe','pyspark[connect]'})`


dangotbanned:
## Related
- #2771
- https://github.com/narwhals-dev/narwhals/pull/3016#discussion_r2291877763

## Description
Only spark-like backends still need support:

https://github.com/narwhals-dev/narwhals/blob/b0e10af8a3b44d003d39c37c91fdafb11d2c436e/narwhals/dataframe.py#L718-L721

https://github.com/narwhals-dev/narwhals/blob/b0e10af8a3b44d003d39c37c91fdafb11d2c436e/narwhals/_utils.py#L1664-L1671

I'm guessing it would require accepting `**kwargs` in `.lazy(...)`, in a similar way to `scan_parquet` for the `session`

https://github.com/narwhals-dev/narwhals/blob/c4bceed4ef7847f6ff291ac00bbeac1dafed706f/narwhals/functions.py#L793-L811

dangotbanned:
`pyspark` can accept `pd.DataFrame` on all versions and `pa.Table` on `>=4.0`
- https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.SparkSession.createDataFrame.html

FBruzzesi:
I am working on this 👀

PR #3024, opened 2025-08-22
test: Simplify `read_scan_test`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2959
- Got sidetracked starting #3023

## Checklist

- [x] Tests added
- [ ] Documented the changes

I was looking to deduplicate the spark-like session stuff in prep for #3023.

I thought I'd use this as full-module example for #2959 on how we can write less repetitive tests 🙂

### DIFF
--- a/tests/read_scan_test.py
+++ b/tests/read_scan_test.py
@@ -182,8 +182,7 @@ def test_scan_fail_spark_like_without_session(
     backend: _SparkLike,
     scan_method: Literal["scan_csv", "scan_parquet"],
 ) -> None:
-    _ = pytest.importorskip(backend)
-
+    pytest.importorskip(backend)
     with pytest.raises(
         ValueError,
         match="Spark like backends require a session object to be passed in `kwargs`.",
--- a/tests/read_scan_test.py
+++ b/tests/read_scan_test.py
@@ -16,6 +16,9 @@ if TYPE_CHECKING:
     from collections.abc import Mapping
     from types import ModuleType
 
+    from pyspark.sql import SparkSession
+    from sqlframe.duckdb import DuckDBSession
+
     from narwhals._typing import EagerAllowed, _LazyOnly, _SparkLike
 
 data: Mapping[str, Any] = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}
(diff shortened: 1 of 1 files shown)


tests/read_scan_test.py
```diff
@@ -1,207 +1,166 @@
...
+        builder.remote(f"sc://localhost:{os.environ.get('SPARK_PORT', '15002')}")
+        if is_spark_connect
+        else builder.master("local[1]").config("spark.ui.enabled", "false")
+    )
+    return (
+        builder.config("spark.default.parallelism", "1")
+        .config("spark.sql.shuffle.partitions", "2")
+        .config("spark.sql.session.timeZone", "UTC")
+        .getOrCreate()
+    )
```
FBruzzesi: Should we consider moving these into `tests/utils.py`? They can be re-used both in `tests/conftest.py` and in #3032

dangotbanned: I do want to eventually, but I was thinking as session-scoped fixtures?

We'd need to restructure some of the existing tests though - e.g. so the same filtering that happens in `--constructors` also applies to these heavy things 🤔

We don't *have* to do that, but that was the idea I was working on when I got distracted and did this PR instead 😂

dangotbanned: Ah second thought, @FBruzzesi yeah just move them, merge and use in #3032 if they're helpful 😅

I did explicitly mention that this was prep for #3023 anyway 🤦

FBruzzesi: Regarding

> I do want to eventually, but I was thinking as session-scoped fixtures?

and

> heavy things

The pyspark session is a singleton (the `getOrCreate` part should be key) -  we create it once and use it in the pyspark constructor

https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/tests/conftest.py#L203-L213

For SQLFrame it should be quite lightweight

> so the same filtering that happens in --constructors

What do you mean by this exactly? I am not following 🙈

Anyway, we can move it as a follow up, no worries

dangotbanned: My brain has melted for the day dude 😭 

> Anyway, we can move it as a follow up, no worries

I'm happy for you to do it now if you're keen to use it the other PR?

tests/dtypes_test.py
```diff
@@ -505,15 +505,9 @@ def test_datetime_w_tz_duckdb() -> None:
...
-    if "pyspark" not in str(constructor) or "sqlframe" in str(constructor):
-        pytest.skip()
+def test_datetime_w_tz_pyspark() -> None:  # pragma: no cover
     pytest.importorskip("pyspark")
-    from pyspark.sql import SparkSession
-
-    session = SparkSession.builder.config(
-        "spark.sql.session.timeZone", "UTC"
-    ).getOrCreate()
+    session = pyspark_session()
```
dangotbanned, line 511: Christ, great call @FBruzzesi!
I had no idea we had this logic in so many places 😂

dangotbanned:
> Self approving my edits 😂 But please take a look at [267e7b7](https://github.com/narwhals-dev/narwhals/commit/267e7b70604f3be49d7661b5756ab4d343c864d7)
> 

All good thanks @FBruzzesi 😍 

> You might have already done it, but I might want to wait for tomorrow because of
> 
> > My brain has melted for the day dude 😭
>

I'm still good to watch all the lovely code fly by and appreciate it 😄

PR #3030, opened 2025-08-23
feat: Add `{Expr,Series}.fill_nan`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/_pandas_like/dataframe.py
```diff
@@ -412,6 +412,11 @@ def drop_nulls(self, subset: Sequence[str] | None) -> Self:
         mask = ~plx.any_horizontal(plx.col(*subset).is_null(), ignore_nulls=True)
         return self.filter(mask)
 
+    def fill_nan(self, value: float | None) -> Self:
+        return self._with_native(
+            self.native.fillna(value=value), validate_column_names=False
```
MarcoGorelli: ffs, pandas doesn't let us pass `value=None`

all I can think to do is:
- go column by column
- if a column is nullable and of floating dtype, fill with `native_namespace.NA` (so, if in the future pandas does distinguish them, this will be future-proof)
- else, leave it alone (don't fill)

narwhals/_dask/dataframe.py
```diff
@@ -164,6 +164,9 @@ def columns(self) -> list[str]:
             )
         return self._cached_columns
 
+    def fill_nan(self, value: float | None) -> Self:
+        return self._with_native(self.native.fillna(value))
```
MarcoGorelli: same comment as https://github.com/narwhals-dev/narwhals/pull/3030/files#r2296083204

FBruzzesi:
@MarcoGorelli I only sneak peaked into this so far. One consideration I have is that if we are **not** going to support `Expr.fill_nan` because of schema considerations, I think we should depart from polars `{Data,Lazy}Frame` signature and allow for `fill_nan(values: Mapping[str, float| None])`. For polars one can argue that all it takes is one extra step to fill different values for different columns, i.e.:

```py
frame.with_columns(pl.col(name).fill_nan(value) for name, value in values.items())
```

But in our case, if we are going to support one method and not the other I think it would make sense to have it a bit more flexible, otherwise the workaround is not very ergonomic for a user. WDYT?

In terms of code changes with respect to what you already did, I think it's pretty minimal

MarcoGorelli:
thanks for taking a look!

I think we could consider supporting `Mapping` as well, but I think it's important that `df.fill_nan(None)` at least works

MarcoGorelli:
As in, if you're suggesting `fill_nan(self, values: float | None | Mapping[str, float | None])` then I agree!

dangotbanned:
I'm not entirely following the logic for why we can't support `Expr.fill_nan`

Couldn't the same argument be made for saying we can't support `Expr.median`?

docs/api-reference/dataframe.md
```diff
@@ -14,6 +14,7 @@
         - estimated_size
         - explode
         - filter
+        - fill_nan
```
FBruzzesi: Nitpick: swap order between these two

Same in LazyFrame

MarcoGorelli:
not sure i see the similarity, does median need to check the schema?

narwhals/_pandas_like/dataframe.py
```diff
@@ -412,6 +413,24 @@ def drop_nulls(self, subset: Sequence[str] | None) -> Self:
...
+            fill_nan(
+                col.native,
+                col.dtype,
+                col.native.dtype,
+                self._implementation,
+                value_nullable,
+                value_numpy,
+            )
+            for col in self._iter_columns()
+        ]
```
FBruzzesi: Can we use the same pattern as in all other backends? Namely pull the conditional out of the `fill_nan` and let `fill_nan` only handle the _actual_ filling part.

```diff
        cols = [
            fill_nan(
                col.native,
-               col.dtype,
                col.native.dtype,
                self._implementation,
                value_nullable,
                value_numpy,
            )
+           if col.dtype.is_float()
+           else col.native
            for col in self._iter_columns()
        ]
```

The same applies to dask

tests/frame/fill_nan_test.py
FBruzzesi, line 1: Not sure what's the purpose of the "c" column here. I think a better test would be to have a None value in it and validate that it doesn't get affected at all?

dangotbanned:
> not sure i see the similarity, does median need to check the schema?

I assumed we did this everywhere, but it looks like we only do for `Series` ops:

https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/narwhals/_polars/series.py#L339-L344

https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/narwhals/_arrow/series.py#L336-L339

https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/narwhals/_pandas_like/series.py#L509-L512

---

So I guess on the flipside of that, is this not precedent for allowing unsafe operations for lazy-only?

I'm also thinking of (https://github.com/narwhals-dev/narwhals/pull/2963#issuecomment-3175990958), where we don't check types before binary operations.

I'm trying to understand what makes *this case* special-er than others 🙂

MarcoGorelli:
aah you're correct, thanks!

I was still in the "use fill-nan in `max` to align it across backends", and that indeed would have been dtype-dependent...but just `fill_nan` isn't, I think it's OK to assume that, given the name, it can only be used on float columns

will update then, thanks 🙏

MarcoGorelli:
Nightly failures for Dask / pandas nightly are unrelated, and reproduce on `main`. i'm fairly confident they'll fix themselves before too long

narwhals/_dask/expr.py
```diff
@@ -450,7 +463,7 @@ def func(expr: dx.Series) -> dx.Series:
                 )
             return res_ser
 
-        return self._with_callable(func, "fillna")
+        return self._with_callable(func, "fill_null")
```
FBruzzesi, line 472: Nice catch 🙏🏼

tests/expr_and_series/fill_nan_test.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import annotations
+
+import narwhals as nw
+from tests.utils import Constructor, ConstructorEager, assert_equal_data
+
+
+def test_is_nan(constructor: Constructor) -> None:
```
FBruzzesi: Nitpick on naming:
```suggestion
def test_fill_nan(constructor: Constructor) -> None:
```

tests/expr_and_series/fill_nan_test.py
```diff
@@ -0,0 +1,40 @@
...
+        expected = {"a": [1.1, 2.0, 3.0], "b": [3.1, 4.0, 3.0]}
+        assert int(result.lazy().collect()["b"].null_count()) == 0
+    else:
+        expected = {"a": [1.1, 2.0, 3.0], "b": [3.1, 4.0, None]}
+        assert result.lazy().collect()["b"].null_count() == 1
+    assert_equal_data(result, expected)
+    assert result.lazy().collect()["a"].null_count() == 0
+
+
+def test_is_nan_series(constructor_eager: ConstructorEager) -> None:
```
FBruzzesi: Same as above 
```suggestion
def test_fill_nan_series(constructor_eager: ConstructorEager) -> None:
```

narwhals/_arrow/series.py
```diff
@@ -651,6 +651,9 @@ def sample(
         mask = rng.choice(idx, size=n, replace=with_replacement)
         return self._with_native(self.native.take(mask))
 
+    def fill_nan(self, value: float | None) -> Self:
+        return self._with_native(pc.if_else(pc.is_nan(self.native), value, self.native))
```
FBruzzesi: Does this need `preserve_broadcast=True`?

```suggestion
        return self._with_native(
            pc.if_else(pc.is_nan(self.native), value, self.native),
            preserve_broadcast=True,
        )
```

narwhals/_pandas_like/series.py
```diff
@@ -588,6 +588,13 @@ def fill_null(
                 )
         return res_ser
 
+    def fill_nan(self, value: float | None) -> Self:
+        value_nullable = self.__native_namespace__().NA if value is None else value
+        value_numpy = float("nan") if value is None else value
+        if get_dtype_backend(self.native.dtype, self._implementation):
+            return self._with_native(self.native.fillna(value_nullable))
+        return self._with_native(self.native.fillna(value_numpy))
+
```
FBruzzesi, line 603: In `is_nan` we check that the underlying type is numeric (maybe we should check that's a float), should we do the same here? Pandas allows `fillna` for all dtypes natively if I am not wrong:

https://github.com/narwhals-dev/narwhals/blob/3c58b0ef1e67173f415d3776918811c51cc989b1/narwhals/_pandas_like/series.py#L553-L558

Also, similar to pyarrow, do we need to `preserve_broadcast=True`?

MarcoGorelli: sure, thanks, you're correct. we should make a `is_elementwise` helper here

ooh `ser != ser` doesn't look great 😳 i may have introduced that myself, but still, not sure i feel comfortable relying on such details, it might be better to use `.isna` whilst pandas doesn't distinguish nan vs null. and when they do, we can use the proper API they provide, rather than relying on `ser != ser`

MarcoGorelli: uurgh pandas[pyarrow] sucks

narwhals/_arrow/series.py
```diff
@@ -651,6 +651,12 @@ def sample(
         mask = rng.choice(idx, size=n, replace=with_replacement)
         return self._with_native(self.native.take(mask))
 
+    def fill_nan(self, value: float | None) -> Self:
+        return self._with_native(
+            pc.if_else(pc.is_nan(self.native), value, self.native),
+            preserve_broadcast=True,
+        )
```
dangotbanned:
```suggestion
        result = pc.if_else(self.native.is_nan(), value, self.native)
        return self._with_native(result, preserve_broadcast=True)
```

narwhals/dataframe.py
```diff
@@ -1977,8 +1977,7 @@ def null_count(self) -> Self:
 
         Notes:
             pandas handles null values differently from Polars and PyArrow.
-            See [null_handling](../concepts/null_handling.md/)
-            for reference.
+            See [null_handling](../concepts/null_handling.md/) for reference.
```
dangotbanned, line 1980: This was the only style I could get working in **both** API ref + IDE 😞

```suggestion
        Notes:
            pandas handles null values differently from Polars and PyArrow.
            See [null_handling](https://narwhals-dev.github.io/narwhals/concepts/null_handling/) for reference.
```

Problematic bits:
- Using relative links
- Having a named link *inside* a Notes section means you can't define the name outside it

image

narwhals/dataframe.py
```diff
@@ -2487,8 +2486,7 @@ def drop_nulls(self, subset: str | list[str] | None = None) -> Self:
...
```
dangotbanned, line 2489: Same `DataFrame.null_count`

narwhals/expr.py
```diff
@@ -1205,14 +1203,48 @@ def fill_null(
...
+    def fill_nan(self, value: float | None) -> Self:
+        """Fill floating point NaN values with given value.
+
+        Arguments:
+            value: Value used to fill NaN values.
+
+        Notes:
+            This function only fills `'NaN'` values, not null ones, except for pandas
+            which doesn't distinguish between them.
+            See [null_handling](../concepts/null_handling.md/) for reference.
```
dangotbanned, line 1215: Same as `DataFrame.null_count`, `LazyFrame.drop_nulls`

narwhals/_pandas_like/series.py
```diff
@@ -588,6 +589,19 @@ def fill_null(
                 )
         return res_ser
 
+    def fill_nan(self, value: float | None) -> Self:
+        if not self.dtype.is_numeric():  # pragma: no cover
+            msg = f"`.is_nan` only supported for numeric dtype and not {self.dtype}, did you mean `.fill_null`?"
```
FBruzzesi:
```suggestion
            msg = f"`.fill_nan` only supported for numeric dtype and not {self.dtype}, did you mean `.fill_null`?"
```

narwhals/_pandas_like/series.py
```diff
@@ -588,6 +589,19 @@ def fill_null(
...
+            raise InvalidOperationError(msg)
+        value_nullable = self.__native_namespace__().NA if value is None else value
+        value_numpy = float("nan") if value is None else value
+        # If/when pandas exposes an API which distinguishes NaN vs null, use that.
+        if get_dtype_backend(self.native.dtype, self._implementation):
+            return self._with_native(
+                self.native.fillna(value_nullable), preserve_broadcast=True
+            )
+        return self._with_native(self.native.fillna(value_numpy), preserve_broadcast=True)
```
dangotbanned: Previously there was a check for `None` twice and unconditional `get_dtype_backend`
But we only need to check it *isn't* `None` once to skip most of that.

```suggestion
    @property
    def _NA(self) -> Any:  # noqa: N802
        nullable = get_dtype_backend(self.native.dtype, self._implementation)
        return self.__native_namespace__().NA if nullable else float("nan")

    def fill_nan(self, value: float | None) -> Self:
        if not self.dtype.is_numeric():  # pragma: no cover
            msg = f"`.is_nan` only supported for numeric dtype and not {self.dtype}, did you mean `.fill_null`?"
            raise InvalidOperationError(msg)
        # If/when pandas exposes an API which distinguishes NaN vs null, use that.
        fill = self._NA if value is None else value
        return self._with_native(self.native.fillna(fill), preserve_broadcast=True)
```

narwhals/dataframe.py
MarcoGorelli: can we resolve this in a separate pr / issue?

dangotbanned: Sure thing

narwhals/_pandas_like/series.py
MarcoGorelli: cool, thanks!

narwhals/_pandas_like/series.py
MarcoGorelli: i'd forgotten that `pandas[pyarrow]` doesn't have separate APIs for NaNs vs nulls, but `fillna` / `isna` currently only apply to null values, not NaN ones (in `Float64` columns)

🤦 

so it looks like we do still need `ser != ser`

narwhals/_pandas_like/series.py
```diff
@@ -588,6 +589,23 @@ def fill_null(
                 )
         return res_ser
 
+    @property
+    def _NA(self) -> Any:  # noqa: N802
+        nullable = get_dtype_backend(self.native.dtype, self._implementation)
+        return self.__native_namespace__().NA if nullable else float("nan")
```
dangotbanned: Actually on second thought, can't we use this?

```py
import pandas as pd

ser = pd.Series([1, 2])
>>> ser.array.dtype.na_value
nan

>>> ser.convert_dtypes().array.dtype.na_value
<NA>

>>> ser.convert_dtypes(dtype_backend="pyarrow").array.dtype.na_value
<NA>
```

MarcoGorelli: wasn't aware - nice one, thanks!

dangotbanned: Glad to see my cursed knowledge from #2744 may finally pay off 😄

MarcoGorelli:
thanks all for reviews!

narwhals/_pandas_like/series.py
MarcoGorelli: this raises for cudf actually

technically this isn't a public api, i think we may need to revert

MarcoGorelli: ah sorry it is indeed part of the pandas api: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.extensions.ExtensionDtype.html#pandas.api.extensions.ExtensionDtype

🤔 maybe it's ok to raise this as a feature request in cudf then

dangotbanned: > maybe it's ok to raise this as a feature request in cudf then

Sounds good

> i think we may need to revert

How about changing it to something like this?

```py
    @property
    def _NA(self) -> Any:  # noqa: N802
        # replace `...` with whatever the cudf-friendly version is
        return ... if self._implementation.is_cudf() else s.array.dtype.na_value

    def fill_nan(self, value: float | None) -> Self:
        # stuff
        s = self.native
        fill = self._NA if value is None else value
```

MarcoGorelli: yeah that's probably fine, thanks

PR #3016, opened 2025-08-20
feat(typing): Make `Implementation` less opaque

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- #3002
- Based on (https://github.com/python/mypy/issues/9937)
- Issue raised privately by @FBruzzesi
  - Explained in (https://github.com/narwhals-dev/narwhals/pull/3016#issuecomment-3206682861)

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## Examples
This PR gives us static typing for `Implementation` at the narwhals-level:

```py
from __future__ import annotations

from typing import Any

import pandas as pd
import polars as pl
import pyarrow as pa
from typing_extensions import reveal_type

import narwhals as nw

data: dict[str, Any] = {"a": [1, 2, 3]}

impl_1 = nw.from_native(pd.DataFrame(data)).implementation
impl_2 = nw.from_native(pl.DataFrame(data)).implementation
impl_3 = nw.from_native(pa.table(data)).implementation

reveal_type(impl_1)  # Type of "impl_1" is "Literal[Implementation.PANDAS]"
reveal_type(impl_2)  # Type of "impl_2" is "Literal[Implementation.POLARS]"
reveal_type(impl_3)  # Type of "impl_3" is "Literal[Implementation.PYARROW]"
```

---

Adapting (https://github.com/narwhals-dev/narwhals#example), the fun stuff happens when we use aliases/type variables.

`IntoFrameT` expands to all valid `Implementation`'s:
```py
if TYPE_CHECKING:
    from narwhals.typing import IntoFrameT


def agnostic_function(df_native: IntoFrameT) -> IntoFrameT:
    frame = nw.from_native(df_native)
    reveal_type(frame.implementation)  # Type of "frame.implementation" is "Literal[Implementation.PANDAS, Implementation.CUDF, Implementation.MODIN, Implementation.PYARROW, Implementation.POLARS, Implementation.PYSPARK, Implementation.SQLFRAME, Implementation.PYSPARK_CONNECT, Implementation.DASK, Implementation.DUCKDB, Implementation.IBIS]"
    return frame.to_native()
```

[...]

Issue #2193, opened 2025-03-11
[Enh]: Add support for switching/converting the backend namespace


matthewwardrop:
### Please describe the purpose of the new feature or describe the problem to solve.

Hi @MarcoGorelli ,

I've been enjoying `narwhals` in formulaic and in some internal inference related tooling at Netflix... but one thing I keep coming up against is misalignment in the namespaces of multiple narwhals objects (e.g. when we need to do a merge of some standard dataframe against a user supplied dataframe that may or may not be backed by the same kind of namespace; or when we do feature generation the output type might not match the target type). I currently work around this by explicitly doing the requisite `.from_native(.from_native(...).to_<namespace>())`... but this becomes cumbersome; and so I've started writing abstract helper functions that do this alignment for me.

### Suggest a solution if possible.

I was wondering if you would be open to patches that do these transformations upstream. Something like: `[DataFrame | Series].convert_namespace(target: str | narwhals.Implementation)`. Obviously this isn't possible for all pairs (e.g. pandas to duckdb), but it would also save some annoying workarounds.

If so, perhaps we could also have an argument to join (e.g. `auto_align_namespace`) that automatically does the conversions for you.

There are lots of places where you probably want to handle conversions manually... but in my experience, and even greater set of circumstances where you really just want it to work.

### If you have tried alternatives, please describe them below.

Manually writing a whole bunch of helper functions.

MarcoGorelli:
Thanks for the request! Something similar came up in Fairlearn too. Will think about it but I'd say that this is in scope 👍

matthewwardrop:
@MarcoGorelli Is this something you would like patches for? Or something you'd prefer to mull over and implement yourself? [I'm guessing the latter, since this is a novel feature and getting the patterns right is pretty important]

MarcoGorelli:
> [@MarcoGorelli](https://github.com/MarcoGorelli) Is this something you would like patches for? Or something you'd prefer to mull over and implement yourself? [I'm guessing the latter, since this is a novel feature and getting the patterns right is pretty important]

i'll think about this and get back to you next week

dangotbanned:
@MarcoGorelli Sorry if this is too much of a nit, but I'm assuming you mean this would be an *instance* method right?

```py
DataFrame(...).with_backend(...)
```

I think one way we could do this is by adding methods like:

```py
CompliantDataFrame.from_arrow
CompliantDataFrame.from_pandas
CompliantDataFrame.from_polars
[...]

PR #3032, opened 2025-08-23
feat: Allow spark-like backends in `.lazy(backend=...)`

### LINKED ISSUE #3023: Support `DataFrame().lazy({'pyspark','sqlframe','pyspark[connect]'})`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3023

## Checklist

- [x] Tests added
- [x] Documented the changes

TODO:

- some listed inline
- coverage

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ polars = ["polars>=0.20.4"]
 dask = ["dask[dataframe]>=2024.8"]
 duckdb = ["duckdb>=1.0"]
 ibis = ["ibis-framework>=6.0.0", "rich", "packaging", "pyarrow_hotfix"]
-sqlframe = ["sqlframe>=3.22.0"]
+sqlframe = ["sqlframe>=3.22.0,!=3.39.3"]
 
 [dependency-groups]
 core = [


narwhals/schema.py
```diff
@@ -171,3 +174,40 @@ def to_polars(self) -> pl.Schema:
...
+    def to_pyspark(self, *, session: SparkSession) -> sqlframe_types.StructType:
+        return self._to_spark_like(backend=Implementation.PYSPARK, session=session)
+
+    def to_pyspark_connect(self, *, session: SparkSession) -> sqlframe_types.StructType:
+        return self._to_spark_like(
+            backend=Implementation.PYSPARK_CONNECT, session=session
+        )
+
+    def to_sqlframe(self, *, session: SparkSession) -> sqlframe_types.StructType:
+        return self._to_spark_like(backend=Implementation.SQLFRAME, session=session)
```
FBruzzesi: TODO: Add docstrings and add in the docs, but nice nice effect 😂

docs/api-reference/expr.md
FBruzzesi, line 1: Unrelated sorting

narwhals/_utils.py
```diff
@@ -1648,9 +1648,9 @@ def is_compliant_expr(
 def is_eager_allowed(impl: Implementation, /) -> TypeIs[_EagerAllowedImpl]:
     """Return True if `impl` allows eager operations."""
     return impl in {
-        Implementation.PANDAS,
-        Implementation.MODIN,
         Implementation.CUDF,
+        Implementation.MODIN,
+        Implementation.PANDAS,
         Implementation.POLARS,
         Implementation.PYARROW,
```
FBruzzesi, line 1655: OCD alphabetically sorting

narwhals/dataframe.py
```diff
@@ -724,7 +725,10 @@ def __arrow_c_stream__(self, requested_schema: object | None = None) -> object:
         return pa_table.__arrow_c_stream__(requested_schema=requested_schema)  # type: ignore[no-untyped-call]
 
     def lazy(
-        self, backend: IntoBackend[Polars | DuckDB | Ibis | Dask] | None = None
+        self,
+        backend: IntoBackend[LazyAllowed] | None = None,
+        *,
+        session: SparkSession | None = None,
```
FBruzzesi: Unlike `scan_{csv, parquet}`, we can directly use the `session` keyword. WDYT?

tests/frame/lazy_test.py
FBruzzesi, line 1: TODO: Simplify into one test

tests/frame/lazy_test.py
```diff
@@ -60,20 +64,83 @@ def test_lazy_to_default(constructor_eager: ConstructorEager) -> None:
...
-    assert result.implementation == implementation
+    assert result.implementation == impl
+    if (
+        impl.is_duckdb()
+        and df.implementation.is_pandas()
+        and df.implementation._backend_version() >= (3, 0, 0)
+    ):
+        # Reason: https://github.com/duckdb/duckdb/issues/18297
+        # > duckdb.duckdb.NotImplementedException: Not implemented Error: Data type 'str' not recognized
+        return
```
FBruzzesi: Of course I am dumb. This must be _before_ `.lazy` call, i.e. just skip the test

narwhals/_arrow/dataframe.py
```diff
@@ -570,6 +576,35 @@ def lazy(self, backend: _LazyAllowedImpl | None = None) -> CompliantLazyFrameAny
...
+        if backend.is_spark_like():
+            from narwhals._spark_like.dataframe import SparkLikeLazyFrame
+            from narwhals.schema import Schema
+
+            if session is None:
+                msg = "Spark like backends require `session` to be not None."
+                raise ValueError(msg)
+            spark_like_schema = Schema(self.schema)._to_spark_like(
+                backend=backend, session=session
+            )
```
FBruzzesi: This might fail for some dtypes, it could make sense to have a try-except block and let spark infer the types if we can't convert directly

narwhals/schema.py
dangotbanned: I'm not against adding these methods to `Schema`, but could probably do with some explanation on why it is part of this PR 😅

narwhals/schema.py
```diff
@@ -18,7 +18,10 @@
 
     import polars as pl
     import pyarrow as pa
+    import sqlframe.base.types as sqlframe_types
 
+    from narwhals._spark_like.utils import SparkSession
```
dangotbanned: *If* we are annotating `session: SparkSession` - it needs to be defined as a structural type **and** not imported from `_spark_like.utils` into the `narwhals` level

I'm replying from my phone, so haven't checked what that type actually is yet

dangotbanned: Right so this currently might produce confusing results.

`SparkSession` is from `sqlframe`


https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/narwhals/_spark_like/utils.py#L19

https://github.com/narwhals-dev/narwhals/blob/68d762a92d50cbf9d9b91c3bae792c6f3d7c3ddf/narwhals/_spark_like/utils.py#L29


If someone has ...
- Only `sqlframe` installed, this works with `sqlframe.base.session._BaseSession` and subclasses
- Both `sqlframe` + `pyspark` installed, doesn't work with `pyspark.sql.SparkSession` or `pyspark.sql.connect.session.SparkSession`
- Only `pyspark` installed, works with anything as the symbol shows as `Unknown`
- Neither installed, works with anything as the symbol shows as `Unknown`


As someone who *hasn't* got `pyspark` installed 😭 but *has* got `sqlframe`, this is what I see in `conftest.py` which has a similar case


image


image

narwhals/schema.py
FBruzzesi: I created a helper function to convert schema from `{pandas,polars,pyarrow}` to spark-like via narwhals. Exposing those 3 would be trivial. I removed it for now, it was probably more confusing than useful. Let's start by letting spark handle/infer the dtypes. We can add it back anytime if needed and/or if we want the feature independently of this PR

dangotbanned: > I removed it for now

RIP me 😂 

```diff
diff --git a/narwhals/_spark_like/dataframe.py b/narwhals/_spark_like/dataframe.py
index c1ab4ec1f..809910834 100644
--- a/narwhals/_spark_like/dataframe.py
+++ b/narwhals/_spark_like/dataframe.py
@@ -7,6 +7,7 @@ from typing import TYPE_CHECKING, Any
 from narwhals._exceptions import issue_warning
 from narwhals._namespace import is_native_spark_like
 from narwhals._spark_like.utils import (
+    SparkSession,
     catch_pyspark_connect_exception,
     catch_pyspark_sql_exception,
     evaluate_exprs,
@@ -33,6 +34,7 @@ if TYPE_CHECKING:
     from types import ModuleType
 
     import pyarrow as pa
+    import sqlframe.base.types as sqlframe_types
     from sqlframe.base.column import Column
     from sqlframe.base.dataframe import BaseDataFrame
     from sqlframe.base.window import Window
@@ -42,7 +44,7 @@ if TYPE_CHECKING:
     from narwhals._spark_like.expr import SparkLikeExpr
     from narwhals._spark_like.group_by import SparkLikeLazyGroupBy
     from narwhals._spark_like.namespace import SparkLikeNamespace
-    from narwhals._typing import _EagerAllowedImpl
+    from narwhals._typing import _EagerAllowedImpl, _SparkLikeImpl
     from narwhals._utils import Version, _LimitedContext
     from narwhals.dataframe import LazyFrame
     from narwhals.dtypes import DType
@@ -110,6 +112,49 @@ class SparkLikeLazyFrame(
     def from_native(cls, data: SQLFrameDataFrame, /, *, context: _LimitedContext) -> Self:
         return cls(data, version=context._version, implementation=context._implementation)
 
+    @classmethod
+    def from_eager(  # name this something better
+        cls,
+        data: Any,
+        /,
+        schema: Mapping[str, DType],
+        *,
+        session: SparkSession,
+        implementation: _SparkLikeImpl,
+        version: Version,
+    ) -> Self:
+        from narwhals.schema import Schema
+
+        spark_schema = Schema(schema)._to_spark_like(
+            backend=implementation, session=session
+        )
+        # TODO @dangotbanned: branching on data/add from_pandas, from_arrow, etc
+        return cls._from_eager(
+            data,
+            spark_schema,
+            session=session,
+            implementation=implementation,
+            version=version,
+        )
+
+    @classmethod
+    def _from_eager(  # name this something better
+        cls,
+        data: Any,
+        /,
+        schema: sqlframe_types.StructType,
+        *,
+        session: SparkSession,
+        implementation: Implementation,
+        version: Version,
+    ) -> Self:
+        return cls(
+            session.createDataFrame(data, schema=schema),
+            version=version,
+            implementation=implementation,
+            validate_backend_version=True,
+        )
+
     def to_narwhals(self) -> LazyFrame[SQLFrameDataFrame]:
         return self._version.lazyframe(self, level="lazy")
 

```

narwhals/schema.py
FBruzzesi: I see, yet I am not sure what's the fix for this.

In `scan_{csv,parquet}` we have `**kwargs: Any`. I think having session might be more useful/explicit. Depending how complex it is to support typing let's see if it's worth having. I imagine that a protocol might be required. Is this the right assumption 🧐?

narwhals/schema.py
dangotbanned: I was thinking we could share some of the logic by moving it to `SparkLikeLazyFrame`?

There's a path for:

1. `pandas.DataFrame`
2. `pyarrow.Table`
3. `narwhals._compliant.CompliantDataFrame`

But `polars` and `pyarrow` are duplicating the version branching logic because of where it is defined currently 😏

narwhals/_arrow/dataframe.py
```diff
@@ -570,6 +576,37 @@ def lazy(self, backend: _LazyAllowedImpl | None = None) -> CompliantLazyFrameAny
...
+                Implementation.PYSPARK_CONNECT,
+            } and backend._backend_version() >= (4, 0, 0)
+            is_pandas_installed = find_spec("pandas") is not None
+
+            data: Any = (
+                self.native
+                if can_create_from_arrow
+                else self.to_pandas()
+                if is_pandas_installed
+                else tuple(self.iter_rows(named=True, buffer_size=512))
```
dangotbanned: Side note, the default for `buffer_size` should be defined on `{PandasLike,Arrow}DataFrame`


We can use omit it for `polars` where it has an effect, but need to provide it only for no-ops?

https://github.com/narwhals-dev/narwhals/blob/747cbd802a2ca560b4641e16c00373b6b98f979e/narwhals/_polars/dataframe.py#L519-L521

That change would mean we can have the call made from the other side (https://github.com/narwhals-dev/narwhals/pull/3032#discussion_r2296237219)

narwhals/schema.py
dangotbanned: The easiest option is to type it as `session: Any | None = None` on the public stuff and then use `SparkSession` internally

I agree it is useful to have the parameter 🙂

narwhals/_arrow/dataframe.py
FBruzzesi: > We can use omit it for polars where it has an effect, but need to provide it only for no-ops?

`buffer_size` has an effect also for pyarrow. I would prefer to keep compliant level methods without defaults as much as possible 🥲

dangotbanned: > buffer_size has an effect also for pyarrow

Ah I stand corrected 😂

We can still do the rest, just provide the default in both cases and always call `to_arrow` - which arrow already does `self.native` for anyway?

E.g. add `SparkLikeLazyFrame.from_compliant_dataframe`

narwhals/schema.py
FBruzzesi: Factor out in https://github.com/narwhals-dev/narwhals/pull/3032/commits/97dae1f9b375c8240554af935e9c34f562d9766c

narwhals/_arrow/dataframe.py
FBruzzesi: Committed in https://github.com/narwhals-dev/narwhals/pull/3032/commits/97dae1f9b375c8240554af935e9c34f562d9766 and didn't even read this comment! Happy to see we converged on the same name 🫡 - yet kept it semiprivate

dangotbanned:
@FBruzzesi I was thinking about what you said yesterday (https://github.com/narwhals-dev/narwhals/pull/3024#discussion_r2296313476)

> The pyspark session is a singleton

I didn't know this before!

I only have `sqlframe` installed, but AFAICT doesn't that mean we can allow omitting `session` if one already exists?

```py

from narwhals import Implementation

if TYPE_CHECKING:
    from narwhals._spark_like.utils import SparkSession
    from narwhals._typing import _SparkLikeImpl


def import_sqlframe_session() -> SparkSession | None:
    from sqlframe.base.session import _BaseSession

    return _BaseSession._instance


def ensure_session(
    backend: _SparkLikeImpl, *, session: SparkSession | None = None
) -> SparkSession:
    if (
        backend.is_sqlframe() and (session or (session := import_sqlframe_session()))
    ) or session is not None:
        return session
    msg = "Spark like backends require `session` to be not None."
    raise ValueError(msg)
```

```py
>>> ensure_session(Implementation.SQLFRAME)
ValueError: Spark like backends require `session` to be not None.
```

```py
>>> DuckDBSession()

>>> ensure_session(Implementation.SQLFRAME)
<sqlframe.duckdb.session.DuckDBSession at 0x1f531b02270>

>>> ensure_session(Implementation.SQLFRAME)
<sqlframe.duckdb.session.DuckDBSession at 0x1f531b02270>

>>> ensure_session(Implementation.SQLFRAME) is ensure_session(Implementation.SQLFRAME)
True
```


I was reading through this when I thought of it

https://github.com/eakmanrq/sqlframe/blob/e5272091c007b04928619068fad8c3f3e6b7040e/sqlframe/base/util.py#L196-L205

> [!IMPORTANT]
> If this general idea is okay, then we could request a public API for it?
> E.g. `get_or_raise`

FBruzzesi:
Thanks @dangotbanned - I had a long day already but let me try to reply to https://github.com/narwhals-dev/narwhals/pull/3032#issuecomment-3218366076

Isn't what you are proposing the same of using the `session.builder.getOrCreate`? This will guarantee that the session is the same if already exist:

```py

s1 = DuckDBSession.builder.getOrCreate()
s2 = DuckDBSession.builder.getOrCreate()

s1 is s2
# True
```

I am not sure what's the "omit" part in 

> I only have `sqlframe` installed, but AFAICT doesn't that mean we can allow omitting `session` if one already exists?

means. Where should we omit it?

My understanding is that after we create it once, the same will be used afterwards - In case I would argue that `DuckDBSession.builder.getOrCreate()` is a much more explicit syntax than the one we are using to create the session.

Also I am not following on why we would want to raise an exception:

> If this general idea is okay, then we could request a public API for it? E.g. `get_or_raise`

narwhals/schema.py
FBruzzesi: @dangotbanned I am going to shoot myself in the foot with this comment, so please pretty please, consider it as a follow up if this is correct. Wouldn't the following be the proper typing:


```py
if TYPE_CHECKING:
    from pyspark.sql import SparkSession as PySparkSession
    from pysparl.sql.connect.session import SparkSession as PySparkConnectSession
    from sqlframe.base.session import _BaseSession

    from narwhals._typing import Dask, DuckDB, EagerAllowed, Ibis, IntoBackend, Polars, PySpark, PySparkConnect, SQLFrame
    
    SQLFrameSession = _BaseSession[Any, Any, Any, Any, Any, Any, Any]
    
@overload
def lazy(
    self,
    backend: IntoBackend[Polars | DuckDB | Ibis | Dask],
    *,
    session: None = None,
) -> LazyFrame[Any]: ...

@overload
def lazy(
    self,
    backend: IntoBackend[PySpark],
    *,
    session: PySparkSession,
) -> LazyFrame[Any]: ...

@overload
def lazy(
    self,
    backend: IntoBackend[PySparkConnect],
    *,
    session: PySparkConnectSession,
) -> LazyFrame[Any]: ...

@overload
def lazy(
    self,
    backend: IntoBackend[SQLFrame],
    *,
    session: SQLFrameSession,
) -> LazyFrame[Any]: ...
```

tests/frame/lazy_test.py
```diff
@@ -1,19 +1,23 @@
...
 
 import narwhals as nw
 from narwhals._utils import Implementation
 from narwhals.dependencies import get_cudf, get_modin
+from tests.utils import assert_equal_data, pyspark_session, sqlframe_session
 
 if TYPE_CHECKING:
-    from narwhals._typing import Dask, DuckDB, Ibis, Polars
+    from narwhals._spark_like.utils import SparkSession
+    from narwhals._typing import LazyAllowed
```
dangotbanned:
```suggestion
    from narwhals._typing import LazyAllowed, SparkLike
```

tests/frame/lazy_test.py
```diff
@@ -54,26 +58,65 @@ def test_lazy_to_default(constructor_eager: ConstructorEager) -> None:
...
+    result = df.lazy(backend=backend, session=session)
     assert isinstance(result, nw.LazyFrame)
-    assert result.implementation == implementation
+    assert result.implementation == impl
+    assert_equal_data(df.sort("a"), data)
+
+
+@pytest.mark.parametrize("backend", ["pyspark", "sqlframe"])
+def test_lazy_spark_like_requires_session(
+    constructor_eager: ConstructorEager, backend: LazyAllowed
```
dangotbanned:
```suggestion
    constructor_eager: ConstructorEager, backend: SparkLike
```

narwhals/schema.py
dangotbanned: >  am going to shoot myself in the foot with this comment,

😂 

> Wouldn't the following be the proper typing:

There are a *couple* of issues

```cmd
>>> mypy
Found 84 errors in 35 files (checked 429 source files)

>>> pyright
78 errors, 0 warnings, 0 informations
```
Part of it is what I mentioned in (https://github.com/narwhals-dev/narwhals/pull/3032#discussion_r2296214229) - which we can totally solve later (https://github.com/narwhals-dev/narwhals/pull/3016#discussion_r2297549732)

But there are also overloads missing (e.g. `df.lazy()`)

A future problem would be that using `IntoBackend` (a union with `ModuleType`) would create incompatible overlapping overloads if we ever want to refine `LazyFrame[Any]`.
I very much do! 😄 (https://github.com/narwhals-dev/narwhals/pull/3016#discussion_r2301079552)

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3032#issuecomment-3221223798

> Isn't what you are proposing the same of using the `session.builder.getOrCreate`?
> This will guarantee that the session is the same if already exist

It seems I did *quite* a bad job explaining in (https://github.com/narwhals-dev/narwhals/pull/3032#issuecomment-3218366076) 😅 

The difference between `getOrCreate` and what I called `get_or_raise` - is that the latter **would never** create a session.

Looking back, what I suggested might be a more confusing public API:

```py
import narwhals as nw

data = {"a": [1, 2, 3], "b": ["x", "y", "z"]}

df = nw.DataFrame.from_dict(data, backend="pandas")
df.lazy("sqlframe")  # error
df.lazy("sqlframe", session=DuckDBSession())  # ok
df.lazy("sqlframe")  # ok
```

**However**, here I was thinking it would be nice to resolve `SparkSession | None` -> `SparkSession`

<details><summary>Examples</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/4c86fbe59a152a43462057e6830283724f94c145/narwhals/_arrow/dataframe.py#L579-L584

https://github.com/narwhals-dev/narwhals/blob/4c86fbe59a152a43462057e6830283724f94c145/narwhals/_pandas_like/dataframe.py#L826-L831

https://github.com/narwhals-dev/narwhals/blob/4c86fbe59a152a43462057e6830283724f94c145/narwhals/_polars/dataframe.py#L497-L502

</p>
</details> 

But maybe we could just do this instead?

```py
def ensure_session(session: SparkSession | None) -> SparkSession:
    if session:
        return session
    msg = "Spark like backends require `session` to be not None."
    raise ValueError(msg)
```

I know it isn't part of the PR now, but (https://github.com/narwhals-dev/narwhals/pull/3032#discussion_r2296160359) would be introducing a `session` parameter in a lot more places.
Maybe the case for *not requiring* it then would be stronger? Happy to drop the idea for now 😄

FBruzzesi:
Thanks for clarifying @dangotbanned (https://github.com/narwhals-dev/narwhals/pull/3032#issuecomment-3224731025)! I somehow thought it was only related to our test suite. But now I get it.

Honestly I don't have enough experience with spark-like sessions to know if, while the session is the same, some settings be changed (e.g. `spark.sql.session.timeZone`).

For SQLFrame I was about to ask: "which session should we even check?", but then I got _very_ surprised to see the following:

```py
from sqlframe.standalone import StandaloneSession
from sqlframe.spark import SparkSession

StandaloneSession() is DuckDBSession() is SparkSession()
True
```

Honestly I am not sure if this is good or bad or even expected 🙈

---

For now I would hold back: for me it falls into one of those cases in which a user can branch based on the implementation. For example:

```py
# Assume I have some lazyframe
some_lazy_frame = ...

nw_lf = nw.from_native(some_lazy_frame)

# Now I receive a pandas object
pd_frame = ...

nw_pd = nw.from_native(pd_frame, eager_only=True)

# and we want to align to the lazy frame:
nw_pd.lazy(
    backend=nw_lf.implementation,
    session=getattr(some_lazy_frame, "sparkSession", None)
)
```

(based on [`pyspark.sql.DataFrame.sparkSession`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.sparkSession.html))

---

If we were to have a function to align the backend of multiple dataframes (see: https://github.com/narwhals-dev/narwhals/issues/2193), then we can definitely pick up the current session ourselves

narwhals/_typing.py
```diff
@@ -51,7 +51,6 @@
 # - https://github.com/narwhals-dev/narwhals/pull/2971#discussion_r2277137003
 # - https://github.com/narwhals-dev/narwhals/pull/3002#issuecomment-3194267667
 _LazyFrameCollectImpl: TypeAlias = Literal[_PandasImpl, _PolarsImpl, _ArrowImpl]  # noqa: PYI047
-_DataFrameLazyImpl: TypeAlias = Literal[_PolarsImpl, _DaskImpl, _DuckDBImpl, _IbisImpl]  # noqa: PYI047
```
dangotbanned, line 54: Farewell questionably-named alias 🥳

narwhals/_utils.py
```diff
@@ -1661,13 +1661,16 @@ def can_lazyframe_collect(impl: Implementation, /) -> TypeIs[_LazyFrameCollectIm
     return impl in {Implementation.PANDAS, Implementation.POLARS, Implementation.PYARROW}
 
 
-def can_dataframe_lazy(impl: Implementation, /) -> TypeIs[_DataFrameLazyImpl]:
+def can_dataframe_lazy(impl: Implementation, /) -> TypeIs[_LazyAllowedImpl]:
     """Return True if `DataFrame.lazy(impl)` is allowed."""
```
dangotbanned:
```suggestion
def is_lazy_allowed(impl: Implementation, /) -> TypeIs[_LazyAllowedImpl]:
    """Return True if `impl` allows lazy operations."""
```

narwhals/_polars/dataframe.py
```diff
@@ -489,6 +494,20 @@ def lazy(self, backend: _LazyAllowedImpl | None = None) -> CompliantLazyFrameAny
...
+            if session is None:
+                msg = "Spark like backends require `session` to be not None."
+                raise ValueError(msg)
+
+            return SparkLikeLazyFrame._from_compliant_dataframe(
+                self,  # pyright: ignore[reportArgumentType]
+                session=session,
+                implementation=backend,
+                version=self._version,
+            )
```
dangotbanned, line 509: ### TODO (@dangotbanned)
Figure out what's up with this boi

narwhals/_duckdb/dataframe.py
```diff
@@ -189,7 +190,12 @@ def drop(self, columns: Sequence[str], *, strict: bool) -> Self:
         selection = (name for name in self.columns if name not in columns_to_drop)
         return self._with_native(self.native.select(*selection))
 
-    def lazy(self, backend: _LazyAllowedImpl | None = None) -> Self:
+    def lazy(
+        self,
+        backend: _LazyAllowedImpl | None = None,
+        *,
+        session: SparkSession | None = None,
+    ) -> Self:
```
dangotbanned: - `backend` has to accept positional, but that's all
- `session` is unused and not needed for the protocol
  - the way this gets called from `DataFrame` still requires `lazy(None, session=None)` to work

Same suggestion for `ibis`

```suggestion
    def lazy(self, backend: None = None, **_: None) -> Self:
```

narwhals/_ibis/dataframe.py
```diff
@@ -165,7 +166,12 @@ def drop(self, columns: Sequence[str], *, strict: bool) -> Self:
         selection = (col for col in self.columns if col not in columns_to_drop)
         return self._with_native(self.native.select(*selection))
 
-    def lazy(self, backend: _LazyAllowedImpl | None = None) -> Self:
+    def lazy(
+        self,
+        backend: _LazyAllowedImpl | None = None,
+        *,
+        session: SparkSession | None = None,
+    ) -> Self:
```
dangotbanned: Same as `duckdb`

```suggestion
    def lazy(self, backend: None = None, **_: None) -> Self:
```

narwhals/_spark_like/dataframe.py
```diff
@@ -561,6 +562,41 @@ def with_row_index(self, name: str, order_by: Sequence[str]) -> Self:
...
+        is_pandas_installed = find_spec("pandas") is not None
+
+        data: Any = (
+            compliant_frame.to_arrow()
+            if can_create_from_arrow
+            else compliant_frame.to_pandas()
+            if is_pandas_installed
+            else tuple(compliant_frame.iter_rows(named=True, buffer_size=512))
+        )
+
```
dangotbanned: Maybe I'm picking too many nits - personal preference is to:

- Not nest ternary expressions
  - It can feel very nice when writing, but when reading later it is more difficult
- Alias an expression/statement/name if it's too long
  - But not if the new name is longer 😉
- Allow the method name and annotation to provide context for a parameter name
  - `_from_compliant_dataframe`, `CompliantDataFrameAny,` *should* be enough 🙏


```suggestion
    @classmethod
    def _from_compliant_dataframe(
        cls,
        frame: CompliantDataFrameAny,
        /,
        *,
        session: SparkSession,
        implementation: _SparkLikeImpl,
        version: Version,
    ) -> Self:
        from importlib.util import find_spec

        impl = implementation
        # pyspark.sql requires pyarrow to be installed from v4.0.0
        if not (impl.is_sqlframe()) and impl._backend_version() >= (4, 0, 0):
            data: Any = frame.to_arrow()
        elif find_spec("pandas"):
            data = frame.to_pandas()
        else:
            data = tuple(frame.iter_rows(named=True, buffer_size=512))
```

tests/frame/lazy_test.py
```diff
@@ -54,26 +58,65 @@ def test_lazy_to_default(constructor_eager: ConstructorEager) -> None:
...
     df = nw.from_native(constructor_eager(data), eager_only=True)
-    result = df.lazy(backend=backend)
+    if (
+        impl.is_duckdb()
+        and df.implementation.is_pandas()
+        and df.implementation._backend_version() >= (3, 0, 0)
+    ):  # pragma: no cover
+        # Reason: https://github.com/duckdb/duckdb/issues/18297
+        # > duckdb.duckdb.NotImplementedException: Not implemented Error: Data type 'str' not recognized
+        return
```
dangotbanned: Should this be a skip/xfail?

FBruzzesi: Sure, let's xfail so that we can see when it get fixed, thanks

FBruzzesi:
Thanks for the detailed review @dangotbanned 

I solved all the suggestions in https://github.com/narwhals-dev/narwhals/pull/3032/commits/eda88aa66271d42e1e3f14df282195a7fa21179c

Regarding https://github.com/narwhals-dev/narwhals/pull/3032#discussion_r2301210184, my suggestion was definitely partial in terms of `@overload` coverage. I will take a look later today, but I am also happy to keep it outside this PR

tests/frame/lazy_test.py
```diff
@@ -54,26 +58,56 @@ def test_lazy_to_default(constructor_eager: ConstructorEager) -> None:
...
-    pytest.importorskip(implementation.name.lower())
+def test_lazy(constructor_eager: ConstructorEager, backend: LazyAllowed) -> None:
+    impl = Implementation.from_backend(backend)
+    pytest.importorskip(impl.name.lower())
+
+    is_spark_connect = os.environ.get("SPARK_CONNECT", None)
+    if is_spark_connect is not None and impl.is_pyspark():  # pragma: no cover
+        # Workaround for impl.name.lower() being "pyspark[connect]" for
+        # Implementation.PYSPARK_CONNECT, which is never installed.
+        impl = Implementation.PYSPARK_CONNECT
```
dangotbanned, line 78: Note to self, we need a `import_or_skip_backend` test util

dangotbanned: Actually, this might be better as part of a `lazy_backend` and/or `lazy_implementation` fixture

narwhals/_polars/dataframe.py
dangotbanned: Right, so I'm gonna say it'll be best to **keep the ignore for now**

Probably with a **TODO** since the problem is only related to this PR *because* it added a `CompliantDataFrameAny` annotation.

To silence the warning (*but not fix it*) this works, but I don't want to do that 😭:

```diff
diff --git a/narwhals/_compliant/dataframe.py b/narwhals/_compliant/dataframe.py
index a3e34bbeb..99c71d479 100644
--- a/narwhals/_compliant/dataframe.py
+++ b/narwhals/_compliant/dataframe.py
@@ -169,7 +169,7 @@ class CompliantDataFrame(
         keys: Sequence[str] | Sequence[CompliantExprT_contra],
         *,
         drop_null_keys: bool,
-    ) -> DataFrameGroupBy[Self, Any]: ...
+    ) -> DataFrameGroupBy[Incomplete, Incomplete]: ...
     def head(self, n: int) -> Self: ...
     def item(self, row: int | None, column: int | str | None) -> Any: ...
     def iter_columns(self) -> Iterator[CompliantSeriesT]: ...
diff --git a/narwhals/_polars/dataframe.py b/narwhals/_polars/dataframe.py
index df79180c9..a2bfd1e4f 100644
--- a/narwhals/_polars/dataframe.py
+++ b/narwhals/_polars/dataframe.py
@@ -502,10 +502,7 @@ class PolarsDataFrame(PolarsBaseFrame[pl.DataFrame]):
                 raise ValueError(msg)
 
             return SparkLikeLazyFrame._from_compliant_dataframe(
-                self,  # pyright: ignore[reportArgumentType]
-                session=session,
-                implementation=backend,
-                version=self._version,
+                self, session=session, implementation=backend, version=self._version
             )
 
         raise AssertionError  # pragma: no cover


```

narwhals/schema.py
dangotbanned: > my suggestion was definitely partial in terms of `@overload` coverage. I will take a look later today, but I am also happy to keep it outside this PR


> [!NOTE]
> TL;DR: It'll probably double the size of the PR  

@FBruzzesi I 100% want `@overload`s here, but since we need to have them on all of:

- `nw.DataFrame.lazy`
- `nw.stable.v1.DataFrame.lazy`
- `nw.stable.v2.DataFrame.lazy`


And `backend`, `session` have defaults  - I'd expect the **diff is gonna grow a lot**

---

This is somewhat of a starting point:


```py
class SparkSession(Protocol):
    def createDataFrame(self, *args: Any, **kwds: Any) -> Any: ...  # noqa: N802

class DataFrame(BaseFrame[DataFrameT]):
    @overload
    def lazy(self) -> LazyFrame[Any]: ...
    @overload
    def lazy(self, backend: Polars | Dask | DuckDB | Ibis) -> LazyFrame[Any]: ...
    @overload
    def lazy(
        self, backend: Polars | Dask | DuckDB | Ibis, *, session: None = ...
    ) -> LazyFrame[Any]: ...
    @overload
    def lazy(self, backend: SparkLike, *, session: SparkSession) -> LazyFrame[Any]: ...
    @overload
    def lazy(
        self, backend: IntoBackend[LazyAllowed], *, session: SparkSession | None
    ) -> LazyFrame[Any]: ...

    def lazy(
        self,
        backend: IntoBackend[LazyAllowed] | None = None,
        *,
        session: SparkSession | None = None,
    ) -> LazyFrame[Any]:
```

**But** still has loads of issues 😭

It *doesn't* catch this:

https://github.com/narwhals-dev/narwhals/blob/281246352bb47cfb5cecb6d87823d5c2eabf08b1/tests/frame/lazy_test.py#L105-L106


But it *would* catch this:

```py
df.lazy(backend=backend)
```

---

This test required adding the last overload, but really `(IntoBackend[LazyAllowed], session: SparkSession | None)` shouldn't be allowed from a typing perspective

https://github.com/narwhals-dev/narwhals/blob/281246352bb47cfb5cecb6d87823d5c2eabf08b1/tests/frame/lazy_test.py#L80-L89

---

And everything inside that follows from inside the method now needs updating to use this new `SparkSession` protocol for `session`

https://github.com/narwhals-dev/narwhals/blob/281246352bb47cfb5cecb6d87823d5c2eabf08b1/narwhals/dataframe.py#L800-L805

dangotbanned (approved):
Thanks @FBruzzesi!

I'm happy with where things are at, we can tackle (https://github.com/narwhals-dev/narwhals/pull/3032#discussion_r2303919770) after (#3016)

hungry narwhal

I think the CI failures were unrelated - but I trust your judgement 🥳

PR #3046, opened 2025-08-27
fix(ci): Disallow `sqlframe==3.39.3`

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

I don't feel like risking to ignore any test we run

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ polars = ["polars>=0.20.4"]
 dask = ["dask[dataframe]>=2024.8"]
 duckdb = ["duckdb>=1.0"]
 ibis = ["ibis-framework>=6.0.0", "rich", "packaging", "pyarrow_hotfix"]
-sqlframe = ["sqlframe>=3.22.0"]
+sqlframe = ["sqlframe>=3.22.0,!=3.39.3"]
 
 [dependency-groups]
 core = [


dangotbanned (approved):

MarcoGorelli:
tbh this bug is so bad, and their response so disappointing (they didn't even yank it) that i think we should de-emphasise sqlframe on https://narwhals-dev.github.io/narwhals/generating_sql/

FBruzzesi:
> tbh this bug is so bad, and their response so disappointing (they didn't even yank it) that i think we should de-emphasise sqlframe on https://narwhals-dev.github.io/narwhals/generating_sql/

I am sure he will get to the bottom of it, he's most likely busy at work and it's mostly a single person project. In the meanwhile I think it's good for us to skip such version and wait what happens next

MarcoGorelli:
arguably that's even more of a reason to de-emphasise it 😄 

i'm not suggesting removing it or anything, i still think we should recommend it - but at the top of our "Narwhals and SQL" page, we should probably consider having "Narwhals + DuckDB" at the top instead

PR #3066, opened 2025-08-31
refactor(expr-ir): Shrinking main

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor

## Related issues

- Target for (#2572) refactors

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- gonna collect a series of PRs in here
- eventually getting (#2572) to a size that is reviewable 🤞
- **DO NOT MERGE**

Issue #3058, opened 2025-08-30
Untangling `CompliantLazyFrame._iter_columns`


dangotbanned:
https://github.com/narwhals-dev/narwhals/blob/04dfd110c6f193bec15c19d846ab3ff7af029eea/narwhals/_compliant/dataframe.py#L278-L284

> Untangling this one will need to be it's own PR

_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/3045#discussion_r2307201728_

## Note
In `ExprIR`, **all** multi-column selections (including selectors) are backend-independent and not even part of the `Compliant*`-level.

So this method, every selectors module and anything added to support them - *can* be avoided 😏 

- #2572

At the very least, some ideas can be adapted into main 😄

PR #3068, opened 2025-08-31
refactor(expr-ir): Generate dispatch methods

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Child of #3066

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- Probably going to iterate on this idea a few times
- Generally looking for something that is concise, but doesn't forgo *all* type safety

> [!NOTE]
> Focused a bit too hard on getting a negative diff
> What's more important is that this **slows future growth**
> [`ExprDispatch`](https://github.com/narwhals-dev/narwhals/pull/3068/files#diff-c1cfa69f5ae914f1985bc19695b33e209ba72ef1dba79aa4664c46371bf9a3b5) was only going to get bigger, and didn't include most of the functions yet


narwhals/_plan/protocols.py
```diff
@@ -200,34 +205,27 @@ class ExprDispatch(StoresVersion, Protocol[FrameT_contra, R_co, NamespaceT_co]):
...
-        name: self.__narwhals_namespace__().mean_horizontal(node, frame, name),
-        strings.ConcatHorizontal: lambda self,
-        node,
-        frame,
-        name: self.__narwhals_namespace__().concat_str(node, frame, name),
+        boolean.AnyHorizontal: lambda self, node, frame, name: namespace(
+            self
+        ).any_horizontal(node, frame, name),
+        boolean.AllHorizontal: lambda self, node, frame, name: namespace(
+            self
```
dangotbanned: - Could probably trim a lot of the dispatch mapping code down with some default config in `ExprIR`, `Function`, `FunctionExpr`
- Can be dynamic like the function reprs
- The lambdas here are barely type-able anyway

dangotbanned: ## Main Variations
1. `FunctionExpr`
    1. Mostly dispatches on `type(self.function)`
        - Subclasses of `FunctionExpr` are special
    2. Some will use `namespace` first
        - Mainly horizontal
    3. Everything with an `_accessor` has that added to path
        - Except `strings.ConcatHorizontal`
2. `Column`, `Literal`, `Len`, `RangeExpr`
    1. Use `namespace`
3. All other `ExprIR`
    1. Mostly snake_case-d class name

dangotbanned: ## Non-examples
Subclasses of:
- `_ColumnSelection` (5x)
- `SelectorIR` (3x)

Concrete nodes:
- `Alias`
- `KeepName`
- `RenameAlias`


These are *expanded*-away or *evaulated* prior to reaching the `Compliant*`-level.

Since there are fewer of them, they should opt-out of generating a method caller

### Related
- #3058
- https://github.com/narwhals-dev/narwhals/blob/192de642a2850ce9d2dfb98dcf4c4fb1a29f7ca3/narwhals/_plan/expr_expansion.py

dangotbanned: Will eventually be replacing most of this (#3073 was just `cast` 😄):

https://github.com/narwhals-dev/narwhals/blob/c72c3e4f0000eb4bbb34ce03bdaad675db29a8c3/narwhals/_plan/protocols.py#L149-L200

And then something similar for this group as well:

https://github.com/narwhals-dev/narwhals/blob/c72c3e4f0000eb4bbb34ce03bdaad675db29a8c3/narwhals/_plan/protocols.py#L201-L222

PR #3084, opened 2025-09-04
tests: fix timezone aware test for Polars>=1.33.0

EdAbati:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

Hi everyone, this should fix CI for `polars>=1.33.0` that now raises an if we have tz-aware string but we don't specify the timezone in the format (i.e. ``narwhals.exceptions.ComputeError: `strptime` / `to_datetime` was called with no format and no time zone, but a time zone is part of the data. This was previously allowed but led to unpredictable and erroneous results. Give a format string, set a time zone or perform the operation eagerly on a Series instead of on an Expr.``)

Bigger questions: should we worry about making this behaviour consistent across backends?

### DIFF
--- a/tests/expr_and_series/str/to_datetime_test.py
+++ b/tests/expr_and_series/str/to_datetime_test.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
-from contextlib import nullcontext as does_not_raise
+from contextlib import AbstractContextManager, nullcontext as does_not_raise
 from datetime import datetime, timezone
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
 import pyarrow as pa
 import pytest
@@ -214,7 +214,7 @@ def test_to_datetime_tz_aware(
     if "cudf" in str(constructor):
         # cuDF does not yet support timezone-aware datetimes
         request.applymarker(pytest.mark.xfail)
-    context = (
+    context: AbstractContextManager[Any] = (
         pytest.raises(NotImplementedError)
         if any(x in str(constructor) for x in ("duckdb", "ibis")) and format is None
         else does_not_raise()


MarcoGorelli:
thanks @EdAbati !

> Bigger questions: should we worry about making this behaviour consistent across backends?

for now i'd just ensure that they all raise some exception. pandas might be the only one that doesn't?

then later if we want we can ensure a consistent error message is raised if we want, but it's less pressing

dangotbanned:
Thanks for the PR @EdAbati!

Ignoring the [very minor typing issue](https://github.com/narwhals-dev/narwhals/actions/runs/17455227739/job/49567340595?pr=3084) - I think we're good to go here?

> for now i'd just ensure that they all raise some exception. pandas might be the only one that doesn't?

@MarcoGorelli could this be a follow-up?
The ci has been pretty noisy and reducing that is a big win already IMO 😄

FBruzzesi:
Thanks @EdAbati 

I fully agree to prioritize this PR, either as is and following up or by expanding its scope to align backends behavior (I don't have a strong preference!)

The CI is unusable otherwise 😭

MarcoGorelli:
yeah sure

EdAbati:
Hi everyone, thank you for merging! There is some sort of a curse: everytime I make a PR to narwhals, work gets busier.
So, sorry I didn't reply yesterday

Anyway, if I have time (and if no one does it before me), tomorrow or on Sunday I'll try to implement what @MarcoGorelli suggested above :)

PR #2978, opened 2025-08-12
feat: being able to read daft dataframes

ym-pett:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below

started trying to adapt `nw` so that it can read daft dataframes, but I might be barking up the wrong tree: I installed daft in the repo so that it would be available to nw. I have a feeling we want to avoid that, and I'm working as if I were adding an actual `_daft` module rather than prep for a plugin. 

note to self: if installing daft was correct, need to add this to an install file, think it's pyproject.toml

PR #2714, opened 2025-06-21
chore(typing): Improving `_pandas_like.utils`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Somewhat of a resurrection of #2227
- But this time building on #2693

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- Noticed a lot of type ignores in (https://github.com/narwhals-dev/narwhals/pull/2712/commits/bb1b0842a7f1255785da1759014fef6a38909a2f)
- We can build on `_BasePandasLike` when we need it
- Another follow-up could look at where `dask` fits in to this
  - https://github.com/narwhals-dev/narwhals/pull/2511#discussion_r2083295496
  - #2713

PR #3086, opened 2025-09-04
refactor(typing): Add `_native.py`

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Following-up https://github.com/narwhals-dev/narwhals/pull/3016#discussion_r2296651319
- https://github.com/narwhals-dev/narwhals/pull/3016#discussion_r2296651319

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- [x] Add module
- [x] Replace old types
- [x] Deduplicate stable typing
- [ ] Add module doc


narwhals/_native.py
```diff
@@ -0,0 +1,380 @@
+"""The home for *mostly* [structural] counterparts to [nominal] native types.
+
+If you find yourself being yelled at by a typechecker and ended up here - **do not fear!**
+
+### (1) `Native(*Frame|Series)`
+Minimal `Protocol`(s) for matching *almost any* supported native type of that group:
+
+    class NativeThing(Protocol):
+        def something_common(self, *args: Any, **kwargs: Any) -> Any: ...
```
dangotbanned, line 29: > [!NOTE]
> I'm using VSCode as my target, and writing *specifically* with markdown that works well there on hover

image

narwhals/_native.py
```diff
@@ -0,0 +1,384 @@
...
+
+    IntoThingT = TypeVar("IntoThingT", bound=IntoThing)
+
+Important:
+    In most situations, you'll want to use these as they **do preserve** the original native type.
+
+Putting it all together, we can now add a *narwhals-level* wrapper:
+
+    class Thing(Generic[IntoThingT]):
+        def to_native(self) -> IntoThingT: ...
```
dangotbanned, line 57: @FBruzzesi I mentioned this [on discord](https://discord.com/channels/1235257048170762310/1272835535190364283/1413452120543461428)

The block I've got here is a bit more finished than the rest - no idea if this kind of thing is helpful 😅 

The general thing I'm hoping to convey is:

- What are the conventions to use when updating?
- When/why might I need to use each group?
- Where can I go to demystify some of the terms `pyright` or `mypy` spit out in error messages?
  - Specifically when using these types and something goes wrong


I still need some time to think of a good way to explain the `Native<Backend>` protocols/aliases 🤔
The problems they solve are tricky to distil into a few lines 😭

dangotbanned:
For the eagle-eyed reviewer ...

image

... yes, I did *almost* fully negate my negative diff with new docs 😂 


### Updated
Aged like milk 😭 

image

narwhals/_native.py
```diff
@@ -0,0 +1,384 @@
...
+
+[structural]: https://typing.python.org/en/latest/spec/glossary.html#term-structural
+[nominal]: https://typing.python.org/en/latest/spec/glossary.html#term-nominal
+[`Protocol`]: https://typing.python.org/en/latest/spec/protocol.html
+[`TypeAlias`]: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#type-aliases
+[`TypeVar`]: https://mypy.readthedocs.io/en/stable/generics.html#type-variables-with-upper-bounds
+[generic class]: https://docs.python.org/3/library/typing.html#user-defined-generic-types
+[`@overload`s]: https://typing.python.org/en/latest/spec/overload.html
+[Type guards]: https://typing.python.org/en/latest/spec/narrowing.html
+"""
```
ym-pett, line 103: Coming to this very much a novice:

I like the clarity of the examples in this block, plus it's great to have the links. Is the block intended to give a general, abstracted idea of the principles, with concrete examples given below?

Generally I quite like having text of the structure: 
1. abstract principles explained
1.1. now I give you some concrete instances and maybe the situations in which those occur.

So as I reader, I kept jumping back from the examples, say below l.249, e.g. `
IntoDataFrame: TypeAlias = NativeDataFrame` and thinking "ok, so I _think_ this is an example of `(1) Native(*Frame|Series)`. 

Am I correct in doing this? Is this what you'd like readers to do? If yes, I was wondering if it's worth adding a few words to this effect, like 'see e.g. `IntoDataFrame` below'

I'm also aware this isn't a tutorial but a docstring, and you have to be a lot more terse here, and maybe this just isn't done in docstrings?

Generally the whole file reads very clearly, and I find the examples in the docstrings of types below l. 249 particularly helpful. At the moment I'm not coming to it searching for a specific bit of information. I think the true test will be when that happens :)

dangotbanned: Thanks @ym-pett, really appreciate you taking a look!

> Is the block intended to give a general, abstracted idea of the principles, with concrete examples given below?

This is tricky, I guess the answer is *kinda yes*?
(https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2328357102) has the goals I'm hoping to achieve here. 

For sections **(1)-(3)** I was definitely trying to work through an abstract example.
They're all related and progressively build on eachother to arrive at the most common use-case for the module:

> How do I define these types, and what do they mean?

```py
dataframe: DataFrame[IntoDataFrameT]
lazyframe: LazyFrame[IntoLazyFrameT]
series: Series[IntoSeriesT]
```

---

*However*, section **(4)** I found much more challenging to stay in the abstract.

I started writing an example, but ...

<details><summary><b>Show cutting room floor example</b></summary>
<p>

```py

from typing import TYPE_CHECKING, Any, Literal, overload

if TYPE_CHECKING:
    import dask.dataframe as dd
    import pandas as pd
    import polars as pl

KNOWN_PACKAGES = ("polars", "dask", "pandas")


@overload
def reveal_package(native: pl.DataFrame) -> Literal["polars"]: ...
@overload
def reveal_package(native: dd.DataFrame) -> Literal["dask"]: ...
@overload
def reveal_package(native: pd.DataFrame) -> Literal["pandas"]: ...
@overload
def reveal_package(native: Any) -> Literal["unknown"]: ...


def reveal_package(native: Any) -> Literal["polars", "dask", "pandas", "unknown"]:
    package_name = type(native).__module__.split(".")[0]
    if package_name not in KNOWN_PACKAGES:
        return "unknown"
    return package_name
```

</p>
</details>

1. It is already 27 lines (vs current docstring is 74 w/ links)
2. The full explanation requires 2-5x more code
3. ... and some text *explaining* it 😅

For reference, this old comment (https://github.com/narwhals-dev/narwhals/pull/2839#discussion_r2216918146) discusses `@overload`s in an **unrelated** way - but gives an idea of how much extra content im talking about 😳

dangotbanned: > [!NOTE]
> I wrote most of (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2336844038) before realising something


> So as I reader, I kept jumping back from the examples, say below l.249,

> Generally the whole file reads very clearly, and I find the examples in the docstrings of types below l. 249 particularly helpful.

I would love to take credit for those, but in ([refactor: Deduplicate most `Into*` typing](https://github.com/narwhals-dev/narwhals/pull/3086/commits/c5ada4a1474532493700f8f991e63d95e23518e6)) I just moved them around 😂 

They're the public docs (I should mention this in the module docstring! 💡)

The part I'm trying to explain is *partly* how do we build them? But also what are all these other symbols doing in `_native.py`?

narwhals/_native.py
```diff
@@ -0,0 +1,384 @@
...
+If we want to describe a set of more specific types (e.g. in [`@overload`s]), then these protocols/aliases are the right tool.
+
+For common and easily-installed backends, aliases are composed of the native type(s):
+
+    NativePolars: TypeAlias = pl.DataFrame | pl.LazyFrame | pl.Series
+
+Otherwise, we need to define a [`Protocol`] which the native type(s) can match against *when* installed:
+
+    class NativeDask(NativeLazyFrame, Protocol):
+        _partition_type: type[pd.DataFrame]
```
dangotbanned: I might just link to some/all of `Namespace.from_native_object`?
https://github.com/narwhals-dev/narwhals/blob/b70de63cf4b207bbccdac76fec6817881428fa4f/narwhals/_namespace.py#L189-L281

In `_Implementation.__get__` - descriptors are added into the mix - which isn't very digestable 😄 

https://github.com/narwhals-dev/narwhals/blob/b70de63cf4b207bbccdac76fec6817881428fa4f/narwhals/_utils.py#L2108-L2150

dangotbanned: **Huh ...** I hadn't really thought about this before.


What's happening in both examples is kinda similar:

1. `Namespace.from_native_object`
    i. Matches a **Native** object to an `Implementation`
    ii. Then constructs and wraps the related `CompliantNamespace`
    iii. Returns `Namespace[CompliantNamespace]`
2. `_Implementation.__get__`
    i. Matches a **Narwhals** object to an `Implementation`, via the wrapped **Native** object
    ii. Returns `Implementation.<what?>`


They be even closer if **1.** didn't include `Namespace.from_backend`

narwhals/_native.py
ym-pett: "how do we build them" -> yes, that's getting through nicely for me. Have re-read this & thinking that the plugin Protocol that's currently in draft version in #2978 should also go into the lower parts of this file when it's merged? ('it' = `narwhals/_native.py`)

dangotbanned: > However, section (4) I found much more challenging to stay in the abstract.

@ym-pett I ended up finishing that example and using it for a comment upstream 😄 

[docs: Add a link to `ibis` issue walkthrough](https://github.com/narwhals-dev/narwhals/pull/3086/commits/16b0074b88f010645ea95036abf99d97e447c573)

It has the command-line output of `mypy` and `pyright`, links to loads of issues that built up to it and documents the current trade-off

dangotbanned:
@FBruzzesi just pinging you on the off-chance you can review before [I'm stranded](https://discord.com/channels/1235257048170762310/1272835922924273694/1416019771362775113) 😉

narwhals/_native.py
ym-pett: "ibis issue walkthrough" - found it! That's gonna take some time to digest for me! :)

FBruzzesi:
> @FBruzzesi just pinging you on the off-chance you can review before [I'm stranded](https://discord.com/channels/1235257048170762310/1272835922924273694/1416019771362775113) 😉

Hey @dangotbanned I will take a look before the end of the weekend/I leave for vacation :D

narwhals/_native.py
```diff
@@ -0,0 +1,387 @@
...
+
+    IntoThingT = TypeVar("IntoThingT", bound=IntoThing)
+
+Important:
+    In most situations, you'll want to use these as they **do preserve** the original native type.
+
+Putting it all together, we can now add a *narwhals-level* wrapper:
+
+    class Thing(Generic[IntoThingT]):
+        def to_native(self) -> IntoThingT: ...
```
FBruzzesi: Mostly nitpicking on the usage of `Thing`. I am trying to put myself into the shoes of a newcomer and I wonder if something along the following lines would be more helpful:

```suggestion
### (1) `Native(*Frame|Series)`
Minimal [`Protocol`]s for matching *almost any* supported native type of that group:

    class NativeObject(Protocol):
        def some_common_method(self, *args: Any, **kwargs: Any) -> Any: ...

Note:
    This group is primarily a building block for more useful types.

### (2) `Into(*Frame|Series)`
*Publicly* exported [`TypeAlias`]s of **(1)**:

    IntoNarwhalsObject: TypeAlias = NativeObject

**But**, occasionally, there'll be an edge-case which we can spell like:

    IntoNarwhalsObject: TypeAlias = Union[<type that does not fit the protocol>, NativeObject]

Tip:
    Reach for these when there **isn't a need to preserve** the original native type.

### (3) `Into(*Frame|Series)T`
*Publicly* exported [`TypeVar`]s, bound to **(2)**:

    IntoNarwhalsObjectT = TypeVar("IntoNarwhalsObjectT", bound=IntoNarwhalsObject)

Important:
    In most situations, you'll want to use these as they **do preserve** the original native type.

Putting it all together, we can now add a *narwhals-level* wrapper:

    class NarwhalsObject(Generic[IntoNarwhalsObjectT]):
        def to_native(self) -> IntoNarwhalsObjectT: ...
```

narwhals/_native.py
```diff
@@ -0,0 +1,387 @@
...
+
+### (4) `Native<Backend>`
+Everything so far has been focused on the idea of matching an *unknown* native object to
+a [`Protocol`] used by a [generic class]:
+
+    DataFrame[IntoDataFrameT]
+    LazyFrame[IntoLazyFrameT]
+    Series[IntoSeriesT]
```
FBruzzesi: Do we need bullet points or something? Not sure how it renders in VSCode or any other IDE

narwhals/_native.py
```diff
@@ -0,0 +1,387 @@
...
+
+If we want to describe a set of more specific types (e.g. in [`@overload`s]), then these protocols/aliases are the right tool.
+
+For common and easily-installed backends, aliases are composed of the native type(s):
+
+    NativePolars: TypeAlias = pl.DataFrame | pl.LazyFrame | pl.Series
+
+Otherwise, we need to define a [`Protocol`] which the native type(s) can match against *when* installed:
```
FBruzzesi: I have mixed feelings on this approach - should we mention that this is a hack to lower false positive and false negatives?

FBruzzesi (commented):
Thanks @dangotbanned - I didn't check line by line, I assumed that most of the changes were simply due to the different import source and somplifications. I double checked the new module and left a couple of comments. Let me know what your thoughts are

narwhals/_native.py
dangotbanned: Well, the goal certainly is this 😄 

> to lower false positive and false negatives

But this part:

> should we mention that this is a hack

Could you explain what you mean by *hack*?

narwhals/_native.py
dangotbanned: Here's as much as I could fit vertically in VSCode 😂 


image

narwhals/_native.py
dangotbanned: Thanks for the review @FBruzzesi!

> Mostly nitpicking on the usage of `Thing`

I am tempted to bikeshed on the name 😄 ...

But regardless of what *exactly* it is - would some kind of intro like this be helpful?

https://github.com/narwhals-dev/narwhals/blob/e92fcc989a5857d5b8a86929fac3b4786a913574/narwhals/_translate.py#L1-L11

narwhals/_native.py
FBruzzesi: What I mean is that what we imply by:

> For common and easily-installed backends

is completely arbitrary.

Correct me if I am wrong but it seems to me that:

- For some libraries, e.g. `polars`, we give for granted that that's installed. But if it's not then `NativePolars` should translate to `Any`
- Viceversa, other libraries we assume are not installed (e.g. Dask and PySpark), therefore we prefer to have _always_ a (suboptimal) protocol (so that it never translate to `Any`), rather than the full type/objects for those cases in which these libraries are installed.

I perfectly understand the tradeoff, I just wonder if it's worth mentioning the reason why we have two different approaches/cases

narwhals/_native.py
FBruzzesi: Oh nice, indentation is code block as in mkdocs

narwhals/_native.py
FBruzzesi: Yes I think it might help, but really I don't want to make a fuss out of it - I am not a fan of using a generic _thing_, and in this situation (i.e. a python context?!) `*Class` or `*Object` would be more appropriate 😂

dangotbanned: > I am tempted to bikeshed on the name 😄 ...
> But regardless of what *exactly* it is 

Sorry @FBruzzesi, I'm reading this back now and it isn't coming across the way that I intended.

My suggestion (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2365743977) was supposed to be *another* way to help with

> I am trying to put myself into the shoes of a newcomer

But not saying that should be *instead* of choosing a better name than `*Thing` 🙂

narwhals/_native.py
dangotbanned: > What I mean is that what we imply by:
> 
> > For common and easily-installed backends
> 
> is completely arbitrary.
> 
> Correct me if I am wrong but it seems to me that:
> 
> * For some libraries, e.g. `polars`, we give for granted that that's installed. But if it's not then `NativePolars` should translate to `Any`

Here's the longer version of *"For common and easily-installed backends"*:

- `polars`
  - I am (maybe optimistically) expecting downstream to have this installed
  - `narwhals` seems like an odd choice if you aren't testing `polars` 😂 
- `pandas`, `pyarrow` can be stubs-only installs
  - https://github.com/pandas-dev/pandas-stubs
  - https://github.com/zen-xu/pyarrow-stubs
- `duckdb`
  - I *thought* could be stubs-only as well
  - Maybe it could be in the future? 🙏 
    - https://github.com/duckdb/duckdb-python/issues/57#issuecomment-3315988281
- `sqlframe`
  - Bit of an outlier in how small of an install it is (*1.66MB* for me)

dangotbanned: > therefore we prefer to have *always* a (suboptimal) protocol

Would it suprise you if I said I'd *prefer* to use protocols for most/all of them?

FBruzzesi: > > therefore we prefer to have _always_ a (suboptimal) protocol
> 
> Would it suprise you if I said I'd _prefer_ to use protocols for most/all of them?

Isn't this counterintuitive? I would lean more towards asking anyone developing narwhals to have _all_ the backends installed. If they don't then they get an `Any` annotation and a possibly worse experience navigating the codebase.

Building and maintaining extremely specific protocols screams code debt to me

dangotbanned: > Building and maintaining extremely specific protocols screams code debt to me

Excluding `_BasePandasLike`, which is *also* solving issues elsewhere (#2714)

<details><summary>Show <code>_BasePandasLike</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/f1ac582fa55971244eb0fe8f77d6b6888a285da4/narwhals/_native.py#L178-L195

</p>
</details> 

I don't think that's a fair characterization of these protocols:

<details><summary><b>Show protocols</b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/f1ac582fa55971244eb0fe8f77d6b6888a285da4/narwhals/_native.py#L205-L233

</p>
</details> 

- Most have a single member
- Most are typed only using `Any`
- The cases *with* non-`Any` types are to act as [tagged unions](https://en.wikipedia.org/wiki/Tagged_union)
  - Where an `Any` would overlap with something we don't want

---

But I'll bite and discuss maintenance cost 😄 

Let's say - *worst-case* - all of these backends (`dask`, `cudf`, `ibis`, `modin`, `pyspark`) removed (*not deprecated*) the protocol member(s) I've specified on the same day.

We would find out as soon as we update the pins we have here:


<details><summary><b>Show dependency groups</b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/e92fcc989a5857d5b8a86929fac3b4786a913574/pyproject.toml#L65-L82

</p>
</details> 

As you can see, some are using older versions which helps us control for that kind of thing 🙂 

The actual *maintenance task* - when we choose to accept it - is:
1. Find some other unique attribute/method for the native object
2. Update the protocol to use that
3. Have the rest of our typing verify that what was picked *didn't* cause any overlaps

I feel like that isn't much work, even when taken to the extreme 😅

dangotbanned: > Isn't this counterintuitive? I would lean more towards asking anyone developing narwhals to have all the backends installed. 

I mean we certainly could ask ...

But I'm not sure how many downstream packages even document that they support everything we do.

## Fishing

Here's what I found from the first 4 in https://github.com/narwhals-dev/narwhals#used-by


### Altair
https://altair-viz.github.io/user_guide/data.html
*Outdated* but is only tested against eager + `pl.LazyFrame` + v1 interchange

### Bokeh
https://docs.bokeh.org/en/latest/docs/user_guide/basic/data.html#using-dataframes
> The data parameter can also be a dataframe (pandas, Polars, PyArrow, or any other **eager dataframe** supported by [Narwhals](https://narwhals-dev.github.io/narwhals/)) or a pandas GroupBy object:

### Darts
https://unit8co.github.io/darts/release_notes/RELEASE_NOTES.html#id115
> from_dataframe() and from_series() now support creating TimeSeries from additional DataFrame backends (Polars, PyArrow, …).
> We leverage narwhals as the compatibility layer between DataFrame libraries. See their [documentation](https://narwhals-dev.github.io/narwhals/) for all supported backends.

### Formulaic
- https://github.com/matthewwardrop/formulaic/pull/189


## Summary
For these packages, I would not recommend that they install lazy-only backends to resolve an upstream (`narwhals`) typing issue - since they don't support those backends.
I feel like that only makes sense if they say they support *everything* we do - at which point I'd expect they have the install for testing 😉

narwhals/_native.py
dangotbanned: So the naming scheme in (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2365639106) is *slightly* different from what I was trying to demonstrate

> and in this situation (i.e. a python context?!) `*Class` or `*Object` would be more appropriate 😂

Here's what each of those look like:

| Actual         | Current       | Proposed              | Alt 1          | Alt 2         |
| -------------- | ------------- | --------------------- | -------------- | ------------- |
| `NativeSeries` | `NativeThing` | `NativeObject`        | `NativeObject` | `NativeClass` |
| `IntoSeries`   | `IntoThing`   | `IntoNarwhalsObject`  | `IntoObject`   | `IntoClass`   |
| `IntoSeriesT`  | `IntoThingT`  | `IntoNarwhalsObjectT` | `IntoObjectT`  | `IntoClassT`  |
| `Series`       | `Thing`       | `NarwhalsObject`      | `Object`       | `Class`       |


> [!NOTE]
> FWIW, I don't like any of them now - including `Thing` 😭

FBruzzesi: Well it took me a bit to get back here but I finally did it!

> FWIW, I don't like any of them now - including Thing 😭

Whops! That's a bad side effect! I would opt for **Proposed**, optionally using `Nw` instead of `Narwhals`. My 2 cents

dangotbanned:
@FBruzzesi how do you feel about the rest of this *besides* (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2410520445)?

I really want to move forward on it to unblock #3125 and I also noticed (https://github.com/narwhals-dev/narwhals/compare/main...chore/refactor-pandas-like-pyarrow-branching) would be touching some of the changed bits as well

*If* everything else is okay - we could do one more round of brainstorming a better name than `Thing` to get it over the line ☺️

FBruzzesi:
> @FBruzzesi how do you feel about the rest of this _besides_ ([#3086 (comment)](https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2410520445))?
> 
> I really want to move forward on it to unblock #3125 and I also noticed ([main...chore/refactor-pandas-like-pyarrow-branching](https://github.com/narwhals-dev/narwhals/compare/main...chore/refactor-pandas-like-pyarrow-branching)) would be touching some of the changed bits as well
> 
> _If_ everything else is okay - we could do one more round of brainstorming a better name than `Thing` to get it over the line ☺️

Sorry @dangotbanned I went on vacation and somehow thought this PR was done. I will take another look in the weekend

narwhals/_native.py
FBruzzesi: Ok it was simpler than what I thought in https://github.com/narwhals-dev/narwhals/pull/3086#issuecomment-3414787959.

Regarding https://github.com/narwhals-dev/narwhals/pull/3086#issuecomment-3414627551:

> If everything else is okay - we could do one more round of brainstorming a better name than Thing to get it over the line ☺️

I would say so, but from your last comment in this thread I am assuming that **Proposed** is what you like the most, which I agree with (my last comment). So we can start with those instead of the `Thing`-ies, and iterate on that if needed?

dangotbanned: Thanks @FBruzzesi!

> but from your last comment in this thread I am assuming that **Proposed** is what you like the most, which I agree with (my last comment).

Oops, that's my bad 🤦‍♂️, lemme clear that up 

When I said (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2366343427)

> So the naming scheme in ([#3086 (comment)](https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2365639106)) is _slightly_ different from what I was trying to demonstrate
> 
> > <your suggestions>
> 
> Here's what each of those look like:

*Proposed*, *Alt 1*, *Alt 2* - were all just refering to what **you proposed**.

I was trying to highlight the fact that this **isn't documenting** the pattern I'm trying to document 😄 


| Actual Narwhals  | Current Docstring | Proposed by @FBruzzesi  |
| ---------------- | ----------------- | ----------------------- |
| Native`<Series>` | Native`<Thing>`   | Native`<Object>`        |
| Into`<Series>`   | Into`<Thing>`     | Into`<NarwhalsObject>`  |
| Into`<Series>`T  | Into`<Thing>`T    | Into`<NarwhalsObject>`T |
| `<Series>`       | `<Thing>`         | `<NarwhalsObject>`      |

I want to show how we're using **1** term and pre/suffixing three ways to describe types.
It's not even that I like it the least/most - but this uses `Object` *and* `NarwhalsObject`

dangotbanned: The `Class` and `Object` ideas (*Alt 1*, *Alt 2*), do fit the pattern - so that is a +1 for me vs (https://github.com/narwhals-dev/narwhals/pull/3086#discussion_r2439961811)

So far I've been fairly specific in the doc about where I use those terms.
**object** is used here to refer to an instance, and **class** is  there for a link to
- https://docs.python.org/3/library/typing.html#user-defined-generic-types

https://github.com/narwhals-dev/narwhals/blob/7e7be3ccf4f22af6909bfbf6826dde9c5d0856e4/narwhals/_native.py#L19

The weird thing *to me* about using them with things like:
- `Protocol`
- `TypeAlias`
- `TypeVar`

Is that those symbols are usually referred to as **types** or [**special forms**](https://typing.python.org/en/latest/spec/glossary.html#term-special-form)

- https://github.com/python/typeshed/blob/11c7821a79a8ab7e1982f3ab506db16f1c4a22a9/stdlib/typing.pyi#L238-L258

But maybe I need to touch grass 😂

FBruzzesi: At the end of the day it's a developer documentation, not even visible to a user so I won't be too picky nor a blocker.

> If everything else is okay - we could do one more round of brainstorming a better name than Thing to get it over the line ☺️

For iteration over it, I am quite open to anything which is not called `Thing` 😂

narwhals/_native.py
```diff
@@ -0,0 +1,413 @@
...
+
+The following examples use the placeholder type `Thing` which represents one of:
+- `DataFrame`: (Eager) 2D data structure representing data as a table with rows and columns.
+- `LazyFrame`: (Lazy) Computation graph/query against a DataFrame/database.
+- `Series`: 1D data structure representing a single column.
+
+Our goal is to **wrap** a *partially-unknown* native object **in** a [generic class]:
+
+    def wrapping_in_df(native: IntoDataFrameT) -> DataFrame[IntoDataFrameT]: ...
+    def wrapping_in_ldf(native: IntoLazyFrameT) -> LazyFrame[IntoLazyFrameT]: ...
```
FBruzzesi:
```suggestion
    def wrapping_in_lf(native: IntoLazyFrameT) -> LazyFrame[IntoLazyFrameT]: ...
```

FBruzzesi (approved):
Thanks @dangotbanned - I left a nitpick suggestion, everything else looks good to me. Sorry for the headache over naming _things_

narwhals/_native.py
dangotbanned: > For iteration over it, I am quite open to anything which is not called `Thing` 😂

Alright I'll try to churn out a list of non-`Thing`s to choose from shortly 😉

narwhals/stable/v1/typing.py
```diff
@@ -90,21 +66,6 @@ def __dataframe__(self, *args: Any, **kwargs: Any) -> Any: ...
     ...     return df.columns
 """
 
-IntoSeries: TypeAlias = "NativeSeries"
```
MarcoGorelli, line 93: so we can remove `IntoSeries` here because `NativeSeries `doesn't change between main and v1, right?

narwhals/_native.py
dangotbanned: Okay @FBruzzesi here we go:

- `Array`
- `Data`
- `Column`
- `Matrix`
- `Dog`
- `Scalar`
- `LinkedList`
- `Box`
- `Bear`

MarcoGorelli:
thanks for your PR!

if you're both happy with it feel free to take it forwards

---

it wasn't clear to me on reading why we're now returning `Into*` instead of `Native*` in the tests but perhaps it'll become clear if i expand all the convos

dangotbanned:
> thanks for your PR!
> 
> if you're both happy with it feel free to take it forwards

Thanks @MarcoGorelli 😍 

> it wasn't clear to me on reading why we're now returning `Into*` instead of `Native*` in the tests but perhaps it'll become clear if i expand all the convos

Hmm, I seem to have done a bad job on the docs then 😅

Here I was trying to explain ...

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/_native.py#L34-L44


... which means we can write like this

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/stable/v1/typing.py#L27

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/_native.py#L279

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/stable/v1/typing.py#L40

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/_native.py#L292

This wasn't the original purpose of the aliases, but I've kinda reclaimed them to work as a lil barrier between the protocols

narwhals/stable/v1/typing.py
dangotbanned: Yeah exactly!

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/stable/v1/typing.py#L5

https://github.com/narwhals-dev/narwhals/blob/3a004e836eee1ddd946f18c9ea8636bbd08aef1e/narwhals/stable/v1/typing.py#L10

Only the bits which are unique to `v1` are **defined in** `v1.typing`:

- The actual narwhals classes
- `DataFrameLike`
- `IntoLazyFrame` doesn't include `NativeIbis`, because `__dataframe__`

MarcoGorelli:
cool, thanks!

> Hmm, I seem to have done a bad job on the docs then 😅

oh no, you've done an excellent job, i just hadn't spent much time on it, happy to trust you two on this one

narwhals/_native.py
FBruzzesi: @dangotbanned I am sold! Let's ship it with `Bear`'s and `Dog`'s 😂

/sarcam

---

Really feel free to merge after resolving the conflict

dangotbanned: Thanks @FBruzzesi 😍 

If you decide on one you prefer (including `Bear` and `Dog` 😉), feel free to update the doc in a follow up 🙌

PR #3098, opened 2025-09-06
tests: Add backcompat for `time_unit`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Completes https://github.com/narwhals-dev/narwhals/pull/3007#discussion_r2327617848
- Unblocks #3007

## Checklist

- [x] Tests added
- [x] Documented the changes

Will avoid introducing [**22 new warnings**](https://github.com/narwhals-dev/narwhals/actions/runs/17515905732/job/49753350299?pr=3007) in the other PR

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -51,6 +51,7 @@ ConstructorPandasLike: TypeAlias = Callable[[Any], "pd.DataFrame"]
 ID_PANDAS_LIKE = frozenset(
     ("pandas", "pandas[nullable]", "pandas[pyarrow]", "modin", "modin[pyarrow]", "cudf")
 )
+ID_CUDF = frozenset(("cudf",))
 _CONSTRUCTOR_FIXTURE_NAMES = frozenset[str](
     ("constructor_eager", "constructor", "constructor_pandas_like")
 )
@@ -239,6 +240,7 @@ def time_unit_compat(time_unit: TimeUnit, request: pytest.FixtureRequest, /) -> 
     request_id = request.node.callspec.id
     if "duckdb" in request_id:
         return "us"
-    if PANDAS_VERSION < (2,) and any(pd_name in request_id for pd_name in ID_PANDAS_LIKE):
+    pandas_like = ID_PANDAS_LIKE - ID_CUDF
+    if PANDAS_VERSION < (2,) and any(name in request_id for name in pandas_like):
         return "ns"
     return time_unit


tests/utils.py
```diff
@@ -217,3 +225,20 @@ def maybe_collect(df: Frame) -> Frame:
...
+        msg = (
+            f"`time_unit_compat` requires the test function to use a `constructor*` fixture.\n"
+            f"Hint:\n\n"
+            f"Try adding one of these as a parameter:\n    {sorted(_CONSTRUCTOR_FIXTURE_NAMES)!r}"
+        )
+        raise NotImplementedError(msg)
+    request_id = request.node.callspec.id
+    if "duckdb" in request_id:
+        return "us"
+    if PANDAS_VERSION < (2,) and any(pd_name in request_id for pd_name in ID_PANDAS_LIKE):
```
FBruzzesi: Does cudf also depends on pandas version?

tests/utils.py
```diff
@@ -217,3 +225,20 @@ def maybe_collect(df: Frame) -> Frame:
...
+
+def time_unit_compat(time_unit: TimeUnit, request: pytest.FixtureRequest, /) -> TimeUnit:
+    """Replace `time_unit` with one that is supported by the requested backend."""
+    if _CONSTRUCTOR_FIXTURE_NAMES.isdisjoint(request.fixturenames):  # pragma: no cover
+        msg = (
+            f"`time_unit_compat` requires the test function to use a `constructor*` fixture.\n"
+            f"Hint:\n\n"
+            f"Try adding one of these as a parameter:\n    {sorted(_CONSTRUCTOR_FIXTURE_NAMES)!r}"
+        )
+        raise NotImplementedError(msg)
```
FBruzzesi, line 239: I am having a brain fart - what's happening here?
IIUC it's a way to avoid using this function outside a test that has a constructor. I must say that the "hint" was not very clear to me  🙈

tests/utils.py
dangotbanned: > Does cudf also depends on pandas version?

Possibly? 😄 

But the idea for this is to use it when the `time_unit` doesn't matter - but simply choosing one can causes issues if used for every constructor.

*However* for `cudf` - I don't believe it has issues with `time_unit` in the first place

we're only handling `modin` and `pandas` here:

https://github.com/narwhals-dev/narwhals/blob/9ca36029bdd240ac743e4385cd1e9279cd4610c6/narwhals/_pandas_like/utils.py#L495-L508

FBruzzesi: > we're only handling modin and pandas here:

Sure, but if `PANDAS_VERSION < (2,)` is installed and `cudf` is running, then the test changes also its default sience:


```py
ID_PANDAS_LIKE = frozenset(
    ("pandas", "pandas[nullable]", "pandas[pyarrow]", "modin", "modin[pyarrow]", "cudf")
)
```

I am not opposing to this, but wanted to make sure we are aware of it

dangotbanned: Just to be safe I've added this 🙂 
([test: Exclude `cudf` from compat replacement](https://github.com/narwhals-dev/narwhals/pull/3098/commits/3b3b4817f3f536276884696fc13406c4cf826fc0))

tests/utils.py
dangotbanned: > I am having a brain fart - what's happening here?

😂 

So [pretty recently](https://github.com/narwhals-dev/narwhals/pull/2957/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128) I added a `constructor_pandas_like` fixture and learned a bit more `pytest` 😄 

So the first check is to make sure the test function which `request: pytest.FixtureRequest` came from has got one of these 3 as a parameter

https://github.com/narwhals-dev/narwhals/blob/3b3b4817f3f536276884696fc13406c4cf826fc0/tests/utils.py#L55-L57

If it does, then `request.node.callspec.id` ...

https://github.com/narwhals-dev/narwhals/blob/3b3b4817f3f536276884696fc13406c4cf826fc0/tests/utils.py#L240-L241


... will be this string ...

```py
timestamp_test.py::test_timestamp_datetimes_tz_aware[duckdb-ms-ms-expected8]'
#                                                    ^^^^^^^^^^^^^^^^^^^^^^
```

... which gets the name from here

https://github.com/narwhals-dev/narwhals/blob/3b3b4817f3f536276884696fc13406c4cf826fc0/tests/conftest.py#L217-L234


### But why?
I don't know about you, but I've been bitten a few times by getting this pattern wrong

https://github.com/narwhals-dev/narwhals/blob/3b3b4817f3f536276884696fc13406c4cf826fc0/tests/expr_and_series/dt/timestamp_test.py#L158-L174

Mainly with 
- `"pyarrow"` unintentionally matching pandas-like
- always forgetting how to select every/a subset of pandas 😞

dangotbanned: Bringing this all back ...


`time_unit_compat` should just require that you've written a test like any of these:

```py
def test_concat_diagonal(
) -> None:

def test_array_dunder(
    request: pytest.FixtureRequest, constructor_eager: ConstructorEager
) -> None:
```


Then once you're in the test, you can do this without even knowing about `constructor` 🤯 

```py
def test_array_dunder(
    request: pytest.FixtureRequest, constructor_eager: ConstructorEager
) -> None:
    replaced = time_unit_compat("us", request)
```

dangotbanned: > Then once you're in the test, you can do this without even knowing about `constructor` 🤯

It was specifically that part I was thinking might lead to someone hitting that error message

FBruzzesi:
Thanks @dangotbanned, feel free to merge if you are happy with it

PR #3110, opened 2025-09-08
"chore: simplify parsing expressions in group-by""

MarcoGorelli:
Re-opening this PR so we can have more chance to discuss, my bad for having jumped the gun here

### DIFF
--- a/narwhals/_compliant/series.py
+++ b/narwhals/_compliant/series.py
@@ -37,6 +37,7 @@ if TYPE_CHECKING:
     from typing_extensions import NotRequired, Self, TypedDict
 
     from narwhals._compliant.dataframe import CompliantDataFrame
+    from narwhals._compliant.expr import CompliantExpr, EagerExpr
     from narwhals._compliant.namespace import EagerNamespace
     from narwhals._utils import Implementation, Version, _LimitedContext
     from narwhals.dtypes import DType
@@ -95,6 +96,8 @@ class CompliantSeries(
 
     def _with_native(self, series: Any) -> Self: ...
     def _with_version(self, version: Version) -> Self: ...
+
+    def _to_expr(self) -> CompliantExpr[Any, Self]: ...
 
     # NOTE: `polars`
     @property
@@ -242,6 +245,9 @@ class EagerSeries(CompliantSeries[NativeSeriesT], Protocol[NativeSeriesT]):
     def __narwhals_namespace__(
         self,
     ) -> EagerNamespace[Any, Self, Any, Any, NativeSeriesT]: ...
+
+    def _to_expr(self) -> EagerExpr[Any, Any]:
+        return self.__narwhals_namespace__()._expr._from_series(self)  # type: ignore[no-any-return]
 
     def _gather(self, rows: SizedMultiIndexSelector[NativeSeriesT]) -> Self: ...
     def _gather_slice(self, rows: _SliceIndex | range) -> Self: ...
--- a/narwhals/_expression_parsing.py
+++ b/narwhals/_expression_parsing.py
@@ -5,11 +5,10 @@
 from __future__ import annotations
 
 from enum import Enum, auto
-from itertools import chain
 from typing import TYPE_CHECKING, Any, Callable, Literal, TypeVar
 
 from narwhals._utils import is_compliant_expr, zip_strict
-from narwhals.dependencies import is_narwhals_series, is_numpy_array, is_numpy_array_1d
+from narwhals.dependencies import is_narwhals_series, is_numpy_array
 from narwhals.exceptions import InvalidOperationError, MultiOutputExpressionError
 
 if TYPE_CHECKING:
@@ -44,13 +43,6 @@ def is_series(obj: Any) -> TypeIs[Series[Any]]:
     from narwhals.series import Series
 
     return isinstance(obj, Series)
-
-
-def is_into_expr_eager(obj: Any) -> TypeIs[Expr | Series[Any] | str | _1DArray]:
-    from narwhals.expr import Expr
-    from narwhals.series import Series
-
-    return isinstance(obj, (Series, Expr, str)) or is_numpy_array_1d(obj)
 
 
 def combine_evaluate_output_names(
@@ -582,14 +574,6 @@ def check_expressions_preserve_length(*args: IntoExpr, function_name: str) -> No
     ):
         msg = f"Expressions which aggregate or change length cannot be passed to '{function_name}'."
         raise InvalidOperationError(msg)
-
-
-def all_exprs_are_scalar_like(*args: IntoExpr, **kwargs: IntoExpr) -> bool:
-    # Raise if any argument in `args` isn't an aggregation or literal.
-    # For Series input, we don't raise (yet), we let such checks happen later,
-    # as this function works lazily and so can't evaluate lengths.
-    exprs = chain(args, kwargs.values())
-    return all(is_expr(x) and x._metadata.is_scalar_like for x in exprs)
 
 
 def apply_n_ary_operation(
--- a/narwhals/_polars/series.py
+++ b/narwhals/_polars/series.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Any, ClassVar, cast, overload
 
 import polars as pl
 
+from narwhals._polars.expr import PolarsExpr
 from narwhals._polars.utils import (
     BACKEND_VERSION,
     SERIES_ACCEPTS_PD_INDEX,
@@ -149,6 +150,10 @@ class PolarsSeries:
     def __init__(self, series: pl.Series, *, version: Version) -> None:
         self._native_series = series
         self._version = version
+
+    def _to_expr(self) -> PolarsExpr:
+        # Polars can treat Series as Expr, so just pass down `self.native`.
+        return PolarsExpr(self.native, version=self._version)  # type: ignore[arg-type]
 
     @property
     def _backend_version(self) -> tuple[int, ...]:
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -19,8 +19,9 @@ from narwhals._exceptions import issue_warning
 from narwhals._expression_parsing import (
     ExprKind,
     check_expressions_preserve_length,
-    is_into_expr_eager,
+    is_expr,
     is_scalar_like,
+    is_series,
 )
 from narwhals._typing import Arrow, Pandas, _LazyAllowedImpl, _LazyFrameCollectImpl
 from narwhals._utils import (
@@ -43,14 +44,14 @@ from narwhals._utils import (
     supports_arrow_c_stream,
     zip_strict,
 )
-from narwhals.dependencies import is_numpy_array_2d, is_pyarrow_table
+from narwhals.dependencies import is_numpy_array_1d, is_numpy_array_2d, is_pyarrow_table
 from narwhals.exceptions import (
     ColumnNotFoundError,
     InvalidIntoExprError,
     InvalidOperationError,
     PerformanceWarning,
 )
-from narwhals.functions import _from_dict_no_backend, _is_into_schema
+from narwhals.functions import _from_dict_no_backend, _is_into_schema, col, new_series
 from narwhals.schema import Schema
 from narwhals.series import Series
 from narwhals.translate import to_native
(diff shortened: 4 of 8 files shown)


dangotbanned:
Okay I've taken more of a look...

Besides what was said in (https://github.com/narwhals-dev/narwhals/pull/3106#issue-3391870092) - am I understanding correctly that the main difference between these two:

```py
BaseFrame._extract_compliant
BaseFrame._parse_into_expr
```


is that the new version is populating `Expr._metadata` now?

Whereas - even prior to #3054 - `BaseFrame.__narwhals_namespace__` was being used and the route from there *did not* fill in the metadata.
Since that doesn't happen at the compliant-level?

MarcoGorelli:
thanks for taking a look!

yup that is indeed a consequence of doing `Expr.alias` before going to the compliant level 👍

dangotbanned:
> thanks for taking a look!
> 
> yup that is indeed a consequence of doing `Expr.alias` before going to the compliant level 👍

Great, okay at least I understand why now 😅

I have an idea I wanna try out tomorrow, which keeps the benefits of your PR - but also starts plugging the hole touched on in https://github.com/narwhals-dev/narwhals/pull/3106#issuecomment-3265695223

The two problems I see are:

1. Creating an `Expr` at the compliant-level doesn't produce metadata we need later
    i. This is *kinda* solved in this PR, at least for the use-case we have here
2. `CompliantSeries` is not fully defined, as we are actually relying on things in `EagerSeries`/`EagerNamespace`
    i. I'm not sure if this is immediately obvious how it's related
    ii. `CompliantSeries().__narwhals_namespace__._expr._from_series`? 


My hope is that this will get us closer to #2713 🤞

dangotbanned:
@MarcoGorelli this is the most minimal diff I could do towards the idea in (https://github.com/narwhals-dev/narwhals/pull/3110#issuecomment-3268074421)
- [`compare/revert-3109...3110/parse-expr-2`](https://github.com/narwhals-dev/narwhals/compare/revert-3109-revert-3106-better-extract-compliant...3110/parse-expr-2)

The main changes after that would be:
- Renaming
  - `Eager` (general)
    - Everything named `EagerMin*` -> `Eager*`
    - Everything (currently) named `Eager*` -> `EagerImpl*`
    - Which then accounts for the gap in `CompliantSeries.__narwhals_namespace__` having more requirements than `CompliantExpr.__narwhals_namespace__`
  - `EagerExpr._from_series` -> `EagerExpr.from_series`
    - Since this is required at the narwhals-level *as well*
- Moving up `EagerNamespace.{from_native,parse_into_expr}` (**low priority**)
  - Haven't done as it requires adding `NativeFrameT, NativeSeriesT` to the new protocol
  - Everything those methods depend on is available at that level though
- Some stray comments I've left 😄

MarcoGorelli:
thanks for taking a look

why do we need both `EagerExpr` and `EagerMinExpr`?

In my head all we need is:
- "public" compliant classes (which plugins / etc. are expected to respect): `CompliantNamespace`, `CompliantExpr`, `CompliantDataFrame`, `CompliantLazyFrame`, `CompliantSeries`
- anything else we consider an "impl" class. rename if necessary

I'm not sure about distinguishing `CompliantEager` and `CompliantLazy` for namespace / expr, not sure it should be necessary. if lazy-only implementations need to make a handful of methods `not_implemented` then that's ok

dangotbanned:
> why do we need both `EagerExpr` and `EagerMinExpr`?

What I've called `EagerMinExpr` is what I've found to be the **minimal** requirements for `Expr` -  when the `*Namespace` *also* has a `_series`.


`CompliantNamespace._series` does not exist and it shouldn't - because it isn't required if you only implement lazy support.

So this class reflects the requirements introduced by eager support:

<details><summary>New <code>EagerExpr</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/d46f8e4afd2ab5d8edc688955a2da888d754128e/narwhals/_compliant/expr.py#L219-L239

</p>
</details> 

Whereas this is our internal stuff we're using for code sharing (like mentioned in #2713)

<details><summary>New <code>EagerImplExpr</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/d46f8e4afd2ab5d8edc688955a2da888d754128e/narwhals/_compliant/expr.py#L242-L922

</p>
</details>

> [!NOTE]
> What I'm proposing is we document these additional requirements that `CompliantSeries` introduces


If a plugin implements `CompliantSeries` with `CompliantNamespace` - it won't work.

If a plugin implements `EagerSeries` with `EagerNamespace` - it works but they've got a huge amount of extra stuff that is specific to `pandas` & `pyarrow`


I want to define the most minimal thing 😄

dangotbanned:
> I'm not sure about distinguishing `CompliantEager` and `CompliantLazy` for namespace / expr, not sure it should be necessary.
> if lazy-only implementations need to make a handful of methods not_implemented then that's ok

Hmm, I understand using `not_implemented` in the cases where `polars` can do stuff that other lazy backends can't - on a granular level

But I see these as fundamentally different requirements:
- `LazyNamespace`
  - Introduces `_lazyframe`
  - Implements `from_native` which is restricted to just a native lazyframe
- `EagerNamespace`
  - Introduces `_dataframe`, `_series`, `from_numpy`
  - Implements `from_native` which is restricted to native dataframe, series
  - Extends what can be used in `parse_into_expr`
  - Extends what can be used in `concat`

It seems confusing to me to replace this with `not_implemented`, since it is handling type parameters, properties, and behavior inside methods

MarcoGorelli:
Sure, thanks, I thought about this again and I think we do indeed need to distinguish `EagerCompliantNamespace` and `LazyCompliantNamespace`

Still not sure about `Expr` though, nor do I see why `_from_series` needs defining at the compliant-expr level. It's just something I used in `EagerExpr` because it's available, but all that needs implementing that the compliant level is `series._to_expr`, which they're free to implement as they like

dangotbanned:
> Sure, thanks, I thought about this again and I think we do indeed need to distinguish `EagerCompliantNamespace` and `LazyCompliantNamespace`

Thanks for hearing me out 🙂  

> Still not sure about `Expr` though, nor do I see why `_from_series` needs defining at the compliant-expr level. It's just something I used in `EagerExpr` because it's available, but all that needs implementing that the compliant level is `series._to_expr`, which they're free to implement as they like

So I chose `*Expr.from_series` over `*Series._to_expr` here because I'm viewing this method (wherever it is defined 😉) as **a constructor for `*Expr`**. Well actually more than that - I want to promote it to a public constructor, to reflect that it is a hard requirement and something we will call in regular use.

It is true that someone *could* implement it another way - but currently - a lot more needs to be defined at *construction-time* for `*Expr` than for `*Series`. 
Keeping that in sync, *I think*, is easier alongside the other `*Expr` constructors.

Ultimately though - I'm not gonna lose any sleep over this one 😄

dangotbanned:
I took a trip down memory lane to find I added `_to_expr` in ([#2149-diff L283-L298](https://github.com/narwhals-dev/narwhals/pull/2149/files#diff-a7444043ebd0b49dde7b882e6912f8f6056c2ab1d05aec54c8692f4f4ee88ed2L283-L298))
Oh these were simpler times :joy: - but it looks like I landed exactly where you did in this PR

https://github.com/narwhals-dev/narwhals/blob/9b93ef82ccd76c393512d03652520101f75892b4/narwhals/_compliant/series.py#L38
https://github.com/narwhals-dev/narwhals/blob/9b93ef82ccd76c393512d03652520101f75892b4/narwhals/_compliant/series.py#L59-L62

Looking back, I think that I wrote ([#2149-diff L283-L298](https://github.com/narwhals-dev/narwhals/pull/2149/files#diff-a7444043ebd0b49dde7b882e6912f8f6056c2ab1d05aec54c8692f4f4ee88ed2L283-L298)) because the other functions were `_to_compliant_expr`, `col`, and `_create_*_series`.

`_to_compliant_expr` is actually very different from what's happening in `_to_expr` and I don't think I fully understood that at the time 🫣

dangotbanned:
Oh I'm not sure if I made it clear in (https://github.com/narwhals-dev/narwhals/pull/3110#issuecomment-3270188608), but what I'm proposing is only:

- `EagerExpr`
- `EagerNamespace`

Everything else is just using `CompliantDataFrame`, `CompliantSeries`

So there was this colossal rename (https://github.com/narwhals-dev/narwhals/commit/4392c33267ae918bbe51dd4ead6858c2480b9644) 😳 
*But* the new stuff is quite minor (https://github.com/narwhals-dev/narwhals/commit/5be916c2a13f6fdadc220008020abc3987a4f9e1) in comparison

MarcoGorelli:
I'm OK with EagerNamespace, it's EagerExpr I'm not sure about. I don't think there's anything in Expr which can't be supported for lazy-only libraries. `drop_nulls` and `unique` we should in principle be able to support (if followed by an aggregation) we just haven't done so yet

> So I chose *Expr.from_series over *Series._to_expr here because I'm viewing this method (wherever it is defined 😉) as a constructor for *Expr

My reasoning for having `Series._to_expr` is that whereas every `Series` can have a `_to_expr` method, not every `Expr` can have a `_from_series` method

dangotbanned:
Thanks for explaining @MarcoGorelli!

Okay so how about this compromise?

I've added some explanation in the commit description ([refactor: Try `to_expr`](https://github.com/narwhals-dev/narwhals/commit/d208cbcd4b50020fbb56b7d36f110e8c9b2914ce))

---

> [!NOTE]
> TL;DR: There are *some* downsides, but whether these are important *enough* I'll leave up to you 🙂 

> My reasoning for having `Series._to_expr` is that whereas every `Series` can have a `_to_expr` method, not every `Expr` can have a `_from_series` method

I 100% agree in isolation - but it's when we try to reason about what we can do with the *result* of `CompliantSeries.to_expr` that some things become visible

### Umm what? ...

The only wart noticeable *in the diff* is that it requires a `type: ignore[return-any]`

https://github.com/narwhals-dev/narwhals/blob/d208cbcd4b50020fbb56b7d36f110e8c9b2914ce/narwhals/_compliant/series.py#L207-L208

### ... who cares though, right?

**However**, that issue flares up if we move `EagerImplNamespace.parse_into_expr` into `EagerNamespace`.
(which I wanted to do 😞 https://github.com/narwhals-dev/narwhals/pull/3110#issuecomment-3270188608)

<p>

```diff
diff --git a/narwhals/_compliant/namespace.py b/narwhals/_compliant/namespace.py
index 38d423ab3..36db30133 100644
--- a/narwhals/_compliant/namespace.py
+++ b/narwhals/_compliant/namespace.py
@@ -31,7 +31,9 @@ if TYPE_CHECKING:
 
     from typing_extensions import TypeAlias
 
+    from narwhals._compliant.expr import CompliantExpr
     from narwhals._compliant.selectors import CompliantSelectorNamespace
+    from narwhals._compliant.series import CompliantSeries
     from narwhals._compliant.when_then import CompliantWhen, EagerImplWhen
     from narwhals._utils import Implementation, Version
     from narwhals.expr import Expr
@@ -183,6 +185,24 @@ class EagerNamespace(
             return self._dataframe.from_numpy(data, schema=schema, context=self)
         return self._series.from_numpy(data, context=self)
 
+    def parse_into_expr(
+        self,
+        data: Expr | Series[NativeSeriesT] | _1DArray | NonNestedLiteral,
+        /,
+        *,
+        str_as_lit: bool,
+    ) -> (
+        CompliantExprT
+        | NonNestedLiteral
+        | CompliantExpr[Any, CompliantSeries[NativeSeriesT]]
+        | CompliantExpr[Any, CompliantSeriesT_co]
+    ):
+        if not (is_series(data) or is_numpy_array(data)):
+            return super().parse_into_expr(data, str_as_lit=str_as_lit)
+        if is_series(data):
+            return data._compliant_series.to_expr()
+        return self._series.from_numpy(data, context=self).to_expr()
+
 
 class EagerImplNamespace(
     EagerNamespace[EagerImplDataFrameT, EagerImplSeriesT, EagerImplExprT],

```

</p>
</details> 

Our return type gets complicated from `EagerImplExprT | NonNestedLiteral`.
The 2 additions to the union are what `pyright` has inferred.

*But* even with that, the signature becomes an incompatible override 😭:

<details><summary><b>Show shouting</b></summary>
<p>




</p>
</details>

### Is that all?

Another case is trying to roundtrip `__narwhals_namespace__`.

With `to_expr` (*and without `EagerExpr`*) a type checker doesn't have the detail that we started in an `EagerNamespace`:


<details><summary><b>Show shouting x2</b></summary>
<p>


image

</p>
</details>

MarcoGorelli:
closing as #3152 covers this

Issue #3064, opened 2025-08-30
[Enh]: Allow `read_` and `scan_` functions to read from path-like or IO objects


29antonioac:
## Upstream
- [ ] https://github.com/pola-rs/polars/issues/17828
- [ ] https://github.com/pola-rs/polars/issues/24921

### Please describe the purpose of the new feature or describe the problem to solve.

Some backends allow having a path-like or IO object as their input files for read and scan functions.

Narwhals does not prevent this to be used: using Polars for example, Narwhals reads or scans from these objects without issues. However, typing doesn't reflect this so wypy has to be told to ignore the line, and not all backends allow this, like DuckDb.

### Suggest a solution if possible.

The read and scan functions could have the type hints updated, and in case of using these objects in a non compatible backend, we could create a temporary file, for example, so we make all of them compatible. 

### Additional information that may help us understand your needs.

I could work on this myself, I would update this issue when I can pick it up.

dangotbanned:
Thanks for opening this, I agree we should allow more than `str`

I was thinking about this recently in #3024 - since for the tests we just call `Path.__str__` beforehand

I'm fairly confident we can allow `str | Path` for a lot of cases and just add 

```py
source = str(source)
```

here

https://github.com/narwhals-dev/narwhals/blob/d318ad31ba52391384eb8b409a43892c60ac95ab/narwhals/functions.py#L597-L605

MarcoGorelli:
thanks for the request

`Path | str` should be easy, agree, it's the IO objects like
```
import io
import duckdb

duckdb.read_csv(io.StringIO('a,b,c\n1,2,3'))
```
that would be trickier

🤔 not sure about those. Could you start by raising a feature request in duckdb?

dangotbanned:
> I'm fairly confident we can allow `str | Path` for a lot of cases and just add
> 
> source = str(source)
> 


I need to correct myself here 😄 

What we should do instead is something like:

```py
from __future__ import annotations

from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:
    import os

    from narwhals._typing import EagerAllowed, IntoBackend
    from narwhals.dataframe import DataFrame


def _normalize_path(source: str | os.PathLike[str]) -> str:
    if isinstance(source, str):
        return source
    from pathlib import Path

    return str(Path(source))


def read_csv(
    source: str | os.PathLike[str], *, backend: IntoBackend[EagerAllowed], **kwargs: Any
[...]

PR #3112, opened 2025-09-08
feat: Allow `os.PathLike[str]` in `{read,scan}_*` functions

### LINKED ISSUE #3100: Allow `os.PathLike[str]` in `{read,scan}_*` functions
Possibly only needs this tweak

- https://github.com/narwhals-dev/narwhals/issues/3064#issuecomment-3240053997

We could add an alias like [`FileSource`](https://github.com/pola-rs/polars/blob/b8bfb07a4a37a8d449d6d1841e345817431142df/py-polars/polars/_typing.py#L332-L341) and add to it if/when #3101 lands

## Upstream
- [ ] https://github.com/pola-rs/polars/issues/17828

dangotbanned:
Closes #3100

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3100
- Related #3064

## Checklist

- [ ] Tests added
- [ ] Documented the changes

- WIP
- Tests are passing locally
- Need ci for pyspark

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -2148,3 +2148,11 @@ class _Implementation:
     def __get__(self, instance: LazyFrame[Any], owner: Any) -> _LazyAllowedImpl: ...
     def __get__(self, instance: Narwhals[Any] | None, owner: Any) -> Any:
         return self if instance is None else instance._compliant._implementation
+
+
+def normalize_path(source: str | os.PathLike[str], /) -> str:
+    if isinstance(source, str):
+        return source
+    from pathlib import Path
+
+    return str(Path(source))
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -765,7 +765,7 @@ def read_parquet(
     elif impl is Implementation.PYARROW:
         import pyarrow.parquet as pq  # ignore-banned-import
 
-        native_frame = pq.read_table(source, **kwargs)
+        native_frame = pq.read_table(source, **kwargs)  # type: ignore[arg-type]
     elif impl in {
         Implementation.PYSPARK,
         Implementation.DASK,
@@ -872,7 +872,7 @@ def scan_parquet(
     elif implementation is Implementation.PYARROW:
         import pyarrow.parquet as pq  # ignore-banned-import
 
-        native_frame = pq.read_table(source, **kwargs)
+        native_frame = pq.read_table(source, **kwargs)  # type: ignore[arg-type]
     elif implementation.is_spark_like():
         if (session := kwargs.pop("session", None)) is None:
             msg = "Spark like backends require a session object to be passed in `kwargs`."
--- a/narwhals/functions.py
+++ b/narwhals/functions.py
@@ -46,6 +46,7 @@ if TYPE_CHECKING:
     from narwhals.dataframe import DataFrame, LazyFrame
     from narwhals.typing import (
         ConcatMethod,
+        FileSource,
         FrameT,
         IntoDType,
         IntoExpr,
@@ -564,7 +565,7 @@ def show_versions() -> None:
 
 
 def read_csv(
-    source: str, *, backend: IntoBackend[EagerAllowed], **kwargs: Any
+    source: FileSource, *, backend: IntoBackend[EagerAllowed], **kwargs: Any
 ) -> DataFrame[Any]:
     """Read a CSV file into a DataFrame.
 
@@ -634,7 +635,7 @@ def read_csv(
 
 
 def scan_csv(
-    source: str, *, backend: IntoBackend[Backend], **kwargs: Any
+    source: FileSource, *, backend: IntoBackend[Backend], **kwargs: Any
 ) -> LazyFrame[Any]:
     """Lazily read from a CSV file.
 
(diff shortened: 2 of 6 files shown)


narwhals/functions.py
```diff
@@ -764,7 +766,7 @@ def read_parquet(
     elif impl is Implementation.PYARROW:
         import pyarrow.parquet as pq  # ignore-banned-import
 
-        native_frame = pq.read_table(source, **kwargs)
+        native_frame = pq.read_table(source, **kwargs)  # type: ignore[arg-type]
```
dangotbanned, line 770: > [fix(typing): Resolve most new warnings](https://github.com/narwhals-dev/narwhals/pull/3112/commits/aaa21f6ed04e998ef9660d10cd63e78edcc22a62)
> 
> `pyarrow.parquet.read_table` is the only one left (2x warns)
> - Seems to be a stub issue
> - Runtime *does* check `__fspath__`
> - https://github.com/apache/arrow/blob/982d31f35fd2cfe87494698dae9ef67d3333658c/python/pyarrow/parquet/core.py#L1381-L1391

narwhals/functions.py
```diff
@@ -857,6 +859,10 @@ def scan_parquet(
     implementation = Implementation.from_backend(backend)
     native_namespace = implementation.to_native_namespace()
     native_frame: NativeDataFrame | NativeLazyFrame
+
+    # NOTE: Add to this if/when other failures show up
+    if implementation is Implementation.DUCKDB:
+        source = normalize_path(source)
```
dangotbanned: The way that `sqlframe` is passing tests *currently* is pretty dodgy

https://github.com/eakmanrq/sqlframe/blob/032251aae8a02202e189b9a6562e5b501085f3fd/sqlframe/duckdb/readwriter.py#L24-L104

It relies on `source.__str__` which gets invoked by f-strings, e.g.

```py
            from_clause = f"'{path}'"
```

That is okay for `pathlib.Path`, since they're identical:

https://github.com/python/cpython/blob/d3b6bb1125576fe93e05ba2f4635804b2ed7ba6a/Lib/pathlib/_local.py#L166-L167

But I need to add a test for [`__fspath__`](https://peps.python.org/pep-0519/) which doesn't implement `__str__`.

### Update 1
- [test: Add failing `PathLike` tests](https://github.com/narwhals-dev/narwhals/pull/3112/commits/44eb382a6f245ab56a5c7dcf4551b7c15d20f090)

dangotbanned: Never would've guessed `pandas` and `pyarrow` would be the only ones supporting `__fspath__`! 😄 

We can add support for all of them with barely any effort - so it should be a nice QOL improvement

narwhals/_utils.py
```diff
@@ -2148,3 +2148,11 @@ def __get__(
...
     def __get__(self, instance: Narwhals[Any] | None, owner: Any) -> Any:
         return self if instance is None else instance._compliant._implementation
+
+
+def normalize_path(source: str | os.PathLike[str], /) -> str:
+    if isinstance(source, str):
+        return source
+    from pathlib import Path
+
+    return str(Path(source))
```
dangotbanned: When this is passed anything that *isn't* `str | os.PathLike[str]` - I'm currently piggybacking off the stdlib error - which is raised in `Path(source)`:

```py
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'BytesIO'
```

However, one quirk is that I haven't called `normalize_path` in *every* case that `pyarrow` is used.
It had its own branch in most cases already and supports all these inputs.

## Example
`buf_b` correctly warns:

> Argument of type "BytesIO" cannot be assigned to parameter "source" of type > "FileSource" in function "read_csv"
>   Type "BytesIO" is not assignable to type "FileSource"
>     "BytesIO" is not assignable to "str"
>     "BytesIO" is incompatible with protocol "PathLike[str]"
>       "__fspath__" is not present

```py

import io
from typing import Any

import polars as pl

import narwhals as nw
from narwhals._typing import IntoBackendEager

data = {"a": ["A", "B", "A"], "b": [1, 2, 3]}

def demo_weirdness(backend: IntoBackendEager) -> nw.DataFrame[Any]:
    with io.StringIO() as buf_s:
        pl.DataFrame(data).write_csv(buf_s)
        with io.BytesIO(buf_s.getvalue().encode()) as buf_b:
            return nw.read_csv(buf_b, backend=backend)
```

But runtime is different ...

👍 

```py
>>> demo_weirdness("polars")
TypeError: argument should be a str or an os.PathLike object where __fspath__ returns a str, not 'BytesIO'
```

👎 

```py
>>> demo_weirdness("pyarrow")
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|pyarrow.Table     |
|a: string         |
|b: int64          |
|----              |
|a: [["A","B","A"]]|
|b: [[1,2,3]]      |
└──────────────────┘
```

dangotbanned: ## Questions
1. Does this matter if we haven't documented support?
2. Should we try to avoid rejecting `IO` objects in backends that support it?
  i. `pandas`, `polars`, `pyarrow`

dangotbanned: I'm open to either direction - just thought it was worth seeing if anyone had a preference?

MarcoGorelli: > Should we try to avoid rejecting IO objects in backends that support it?

i don't think so

PR #3115, opened 2025-09-09
docs: Add python version support to backcompat

### LINKED ISSUE #2638: Add Python version support policy
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/issues/2084#issuecomment-2925189450_

> Personally, I'd like to drop `3.8` support **now** and at the same time introduce a Python version support policy (e.g. [python/typing_extensions#571](https://github.com/python/typing_extensions/issues/571)). 
> My suggestion for that is to keep it simple and stick to [EOL support](https://devguide.python.org/versions/) - which is what _most_ of the (upstream) backends do anyway. 
> We can always extend this to avoid breaking downstream but I think documenting it **5 months ahead** of `3.9` EOL will leave plenty of time to prepare CI workflows 🤞.
> 
> I'm open to any other ideas, just hoping to move this issue along if possible 🙂

It doesn't need to be anything huge and can **mostly** just point towards the [dev guide calendar](https://devguide.python.org/versions/).
We can also *borrow* some rationale from [SPEC 0 — Minimum Supported Dependencies](https://scientific-python.org/specs/spec-0000/), if needed.

The novel part will be explaining how this interacts with our [Perfect backwards compatibility policy](https://narwhals-dev.github.io/narwhals/backcompat/).
Currently, that page doesn't mention `python` at all which may be fine from a `narwhals`-perspective but could cause confusion if someone assumed we support all versions of `python` indefinitely 😅

dangotbanned:
Closes #2638

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2638

## Checklist

- [ ] Tests added
- [x] Documented the changes

### DIFF
--- a/docs/backcompat.md
+++ b/docs/backcompat.md
@@ -98,7 +98,7 @@ Here are exceptions to our backwards compatibility policy:
 - We may consider making some type hints more precise.
 - Anything labelled "unstable".
 - We may sometimes need to bump the minimum versions of supported backends.
-- We will drop support for old python versions, roughly in sync with [devguide](https://devguide.python.org/versions/)
+- We will drop support for old python versions, roughly in sync with [their end of life](https://devguide.python.org/versions/#supported-versions)
 
 In general, decision are driven by use-cases, and we conduct a search of public GitHub repositories
 before making any change.


docs/backcompat.md
```diff
@@ -98,6 +98,7 @@ Here are exceptions to our backwards compatibility policy:
 - We may consider making some type hints more precise.
 - Anything labelled "unstable".
 - We may sometimes need to bump the minimum versions of supported backends.
+- We will drop support for old python versions, roughly in sync with [devguide](https://devguide.python.org/versions/)
```
FBruzzesi: Tiny suggestion:

```suggestion
- We will drop support for old python versions, roughly in sync with [their end of life](https://devguide.python.org/versions/#supported-versions)
```
WDYT?

FBruzzesi (commented):
Thanks @dangotbanned - that makes sense to state explicitly. Should we think of a timeline? e.g. 1/2/3 months after end of life?

dangotbanned:
Thanks for the quick review @FBruzzesi 😍 

Re:
> Should we mention a rough timeline? e.g. 1/2/3 months after end of life?

I thought about this and even adding more context for *why* (e.g. #2638)

Ultimately though:
- #2638 
  - Had zero interest in the 3 months it was open
- #2084
  - Didn't get any external feedback in the 3 months it was open
  - https://github.com/plotly/plotly.py/issues/5054
    - Was the only *considered* blocker, which also didn't attract much feedback

So I would prefer to [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) this one 😉 - and make an extension to it if/when we hear back from downstream

Issue #3059, opened 2025-08-30
Transition to single-`Implementation` classes


dangotbanned:
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/3045#discussion_r2304912233_

I'm somewhat torn on these `CompliantExpr` constructors.

I **like** that all you need to do is implement these two:

https://github.com/narwhals-dev/narwhals/blob/29a3038adfa6de558fb1a240d70a722858d715f7/narwhals/_compliant/expr.py#L95-L106

<details><summary>And then you get these 4 <i>for "free"</i></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/29a3038adfa6de558fb1a240d70a722858d715f7/narwhals/_compliant/namespace.py#L53-L76

</p>
</details> 

What I **don't like** is that `_LimitedContext` specifies you need to pass an object with:

- `_implementation: Implementation`
- `_version: Version`

Recently, this PR removed the requirement of *also* having `_backend_version`:

- #2764

I'd like to continue that trend and work towards removing `_implementation` as well 😏 

I've been wanting to drop down to 1x `Implementation` per class for a while:
- https://github.com/narwhals-dev/narwhals/pull/2064#discussion_r1976635985
- https://github.com/narwhals-dev/narwhals/pull/2051#issuecomment-2682817033

Which would just mean defining it **on the class** like e.g. `Arrow*`

https://github.com/narwhals-dev/narwhals/blob/29a3038adfa6de558fb1a240d70a722858d715f7/narwhals/_arrow/expr.py#L28-L29

If we do that, then:
1. We don't need it included in any signatures
    i `ArrowExpr` still requires a sink in `__init__` even now 😢 

    https://github.com/narwhals-dev/narwhals/blob/29a3038adfa6de558fb1a240d70a722858d715f7/narwhals/_arrow/expr.py#L41-L42

2. `*Like` classes don't need to branch on `Implementation` inside methods
    i. We just override things in subclasses
3. We might be able to do something different for extensions (#3042)
    i. Where `Implementation.UNKNOWN` is just a constant to satisfy typing atm

--- 

> We could also then make `__native_namespace__` a `@classmethod` 🤯
> 
> That doesn't depend on `nw.Version`, so we don't need to initialize a `Compliant` object to access it

PR #3116, opened 2025-09-10
feat: Add support for `{Expr,Series}.str.to_titlecase`

### LINKED ISSUE #3060: enh: `str.to_titlecase`
This would be a pretty cool feature, i was looking for it today

It's non-trivial to implement for sql-like, but i think with a big of regexp it should be doable

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3060

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/expr_str.py
+++ b/narwhals/expr_str.py
@@ -465,15 +465,18 @@ class ExprStringNamespace(Generic[ExprT]):
         Warning:
             Different backends might follow different rules to determine what a "word" is:
 
-            - duckdb, polars and spark-like use **non-alphanumeric** characters to
+            - duckdb, polars and spark-like use non-**alphanumeric** characters to
                 define the word boundaries.
-            - pandas-like, pyarrow and dask use **non-alphabetic** characters to define
+            - pandas-like, pyarrow and dask use non-**alphabetic** characters to define
                 the word boundaries, matching the behavior of
                 [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title).
 
-            As an example of such difference, in the former case the string `"with123numbers"`
-            is mapped to `"With123numbers"` (notice lowercase **n** after the digits), while
-            in the latter to `"With123Numbers"` (notice uppercase **N** after the digits).
+            We can observe the difference with the string `"with123numbers"`:
+
+            - non-**alphanumeric** -> `"With123numbers"`
+                - notice lowercase **n** after the digits
+            - non-**alphabetic** -> `"With123Numbers"`
+                - notice uppercase **N** after the digits
 
         Examples:
             >>> import polars as pl


narwhals/_duckdb/expr_str.py
```diff
@@ -27,4 +29,28 @@ def to_date(self, format: str | None) -> DuckDBExpr:
         compliant_expr = self.compliant
         return compliant_expr.cast(compliant_expr._version.dtypes.Date())
 
+    def to_titlecase(self) -> DuckDBExpr:
+        if (version := self.compliant._backend_version) < (1, 3):  # pragma: no cover
+            msg = (
+                "`Expr.str.to_titlecase` is only available in 'duckdb>=1.3', "
+                f"found version {version!r}."
+            )
+            raise NotImplementedError(msg)
```
FBruzzesi: We can't use @requires.backend_version in namespaces

narwhals/_duckdb/expr_str.py
```diff
@@ -27,4 +29,28 @@ def to_date(self, format: str | None) -> DuckDBExpr:
...
+        def _to_titlecase(expr: Expression) -> Expression:
+            lower_expr = F("lower", expr)
+            extract_expr = F(
+                "regexp_extract_all", lower_expr, lit(r"[a-z0-9]+[^a-z0-9]*")
+            )
+            capitalized_expr = F(
+                "list_transform",
+                extract_expr,
+                SQLExpression("lambda s: upper(s[1]) || s[2:]"),
+            )
```
FBruzzesi: I couldn't figure out a way to express these lambda functions from python without wrapping them in a `SQLExpression`

narwhals/_polars/expr.py
```diff
@@ -400,6 +400,22 @@ class PolarsExprDateTimeNamespace(
...
 ):
+    def to_titlecase(self) -> PolarsExpr:
+        native_expr = self.native
+        backend_version = self.compliant._backend_version
+
+        if backend_version < (1, 5):  # pragma: no cover
+            native_result = (
+                native_expr.str.to_lowercase()
+                .str.extract_all(r"[a-z0-9]+[^A-Za-z0-9]*")
+                .list.eval(pl.element().str.to_titlecase())
```
FBruzzesi, line 422: At this point, this operation is "safe", since the non-alphanumeric chars are all at the end of the string

narwhals/_spark_like/expr_str.py
```diff
@@ -34,3 +34,22 @@ def to_date(self, format: str | None) -> SparkLikeExpr:
...
+            capitalized_expr = F.transform(extract_expr, f=F.initcap)
+            return F.array_join(capitalized_expr, delimiter="")
+
+        return self.compliant._with_elementwise(_to_titlecase)
+    ```
+    However:
+
+    * for pyspark: ColumnNotFoundError is raised
+    * for sqlframe: `regexp_extract_all` is not implemented
+    """
```
FBruzzesi: Send help 😭

narwhals/_duckdb/expr_str.py
dangotbanned: I could add support for that ahead of this PR if you want? 😄

`@requires` is typed with wider requirements than necessary - so I should probably narrow that anyway

https://github.com/narwhals-dev/narwhals/blob/ec5f4967bf7ea3513a9d0fa1cc837985b612e0f7/narwhals/_utils.py#L225-L231


https://github.com/narwhals-dev/narwhals/blob/ec5f4967bf7ea3513a9d0fa1cc837985b612e0f7/narwhals/_utils.py#L1926-L1933

narwhals/_duckdb/expr_str.py
dangotbanned: If we **need** to use `SQLExpression`, I'll open a PR for (https://github.com/narwhals-dev/narwhals/commit/c3456c3397ca3ed901f79fd8c7daedd95ebea998)

I'm curious though about why [`duckdb.LambdaExpression`](https://github.com/duckdb/duckdb-python/blob/636c3bce185ead2fecac9a896fd28003ddcd3b16/duckdb/__init__.pyi#L272) wouldn't work here?
I've never used it - but it looks like the right shoe for this foot

tests/expr_and_series/str/to_titlecase_test.py
```diff
@@ -21,8 +22,10 @@
         "They'Re Bill'S Friends From The Uk",
         "To Infinity,And Beyond!",
         "With123Numbers",
+        "__Dunder__Score_A1_.2B ?Three",
```
dangotbanned, line 24: FWIW, I don't understand why [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title) does this - but `polars` and `duckdb` *don't*

dangotbanned: I chose `"_"` because I was going to suggest using `r"[[:word:]]"` and [that includes](https://docs.rs/regex/latest/regex/#ascii-character-classes) `"_"`

```
[[:word:]]     word characters ([0-9A-Za-z_])
```

narwhals/_polars/expr.py
dangotbanned: Huge +1 on `list.eval` for this 😍

tests/expr_and_series/str/to_titlecase_test.py
FBruzzesi: Duckdb we can choose the behavior since we are defining the regex for how to define word boundaries.
polars and the regex I added does not consider a number as a boundary, but it does for an underscore hence word characters are [0-9A-Za-z], which means that `"__dunder__score_a1_.2b ?three" -> "__Dunder__Score_A1_.2b ?Three"` isn't it?

Edit: I now see that we should take care of leading non alphanumeric chars - update: https://github.com/narwhals-dev/narwhals/pull/3116/commits/599fb25bd3f96a965fd0a8bacdac7b8da0395e56

dangotbanned: > > FWIW, I don't understand why [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title) does this - but `polars` and `duckdb` _don't_

Sorry this was very unclear 😭 

I meant to say:

> I don't understand why `str.title` does this
> Regardless, the versions in `polars` and `duckdb` _don't_


Happy to see you've fixed it though 🥳

narwhals/_spark_like/expr_str.py
FBruzzesi: Fixed in https://github.com/narwhals-dev/narwhals/pull/3116/commits/aa40799ec9bdeb7bde753d63643dfe883e3f81c7 🥳

narwhals/_duckdb/expr_str.py
FBruzzesi: I didn't know about LambdaExpression. Trying it now

FBruzzesi: @dangotbanned how do you like this? https://github.com/narwhals-dev/narwhals/pull/3116/commits/0bba2d952d9f301bb736b4d84cf3baa71398f232

I have somewhat mixed feelings 😂 

FYI: `LambdaExpression` is available since v1.2

narwhals/expr_str.py
```diff
@@ -455,6 +455,54 @@ def to_lowercase(self) -> ExprT:
...
+            This is a form of case transform where the first letter of each word is
+            capitalized, with the rest of the word in lowercase.
+
+        Warning:
+            Different backends might follow different rules to determine what a "word" is:
+
+            - polars uses **non-alphanumeric** characters to define the word boundaries.
+            - pandas-like, pyarrow and dask use **non-alphabetic** characters to define
+                the word boundaries, matching the behavior of
+                [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title).
```
MarcoGorelli: if you've implemented this for duckdb and spark, do you want to include them here too?

tests/expr_and_series/str/to_titlecase_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+def test_str_to_titlecase_expr(
+    request: pytest.FixtureRequest, constructor: Constructor
+) -> None:
+    if "ibis" in str(constructor):
+        request.applymarker(pytest.mark.xfail)
+
+    if "duckdb" in str(constructor) and DUCKDB_VERSION < (1, 2):
+        reason = "version too old, duckdb 1.2 required for LambdaExpression."
+        request.applymarker(pytest.mark.xfail(reason=reason))
```
MarcoGorelli: can just skip

narwhals/_polars/expr.py
MarcoGorelli: sure, for backcompat code `list.eval` is fine, so long as for the most recent versions there's a specialised path

FBruzzesi: yes of course (line 415 😉)

narwhals/_duckdb/expr_str.py
dangotbanned: > FYI: `LambdaExpression` is available since v1.2

Well this is interesting ...

~~I have an idea for making this a backwards-compatible util 🤞~~

## Update
Ignore me - I didn't realise you needed `SQLExpression` (`duckdb>=1.3.0`) to write it the way you had before 🤦‍♂️

narwhals/_duckdb/expr_str.py
```diff
@@ -27,4 +29,30 @@ def to_date(self, format: str | None) -> DuckDBExpr:
...
+            capitalize = F(
+                "||",
+                F("upper", F("left", col("s"), lit(1))),
+                F("substr", col("s"), lit(2)),
+            )
+            capitalized_expr = F(
+                "list_transform", extract_expr, LambdaExpression("s", capitalize)
+            )
+            return F("list_aggregate", capitalized_expr, lit("string_agg"), lit(""))
```
dangotbanned: > @dangotbanned how do you like this? [0bba2d9](https://github.com/narwhals-dev/narwhals/commit/0bba2d952d9f301bb736b4d84cf3baa71398f232)

Could we use a better name than `"s"`?

I *do* get that it's meant to be like:

```py
lambda s: upper(s[1]) || s[2:]
```

But it looks a bit confusing here 🤔 

I'm gonna have a fiddle 😄

FBruzzesi: What about any of {"el", "elem", "element"}?

dangotbanned: > I'm gonna have a fiddle 😄

Spent way too long reading the docs to come up with this 😅 

[refactor(suggestion): Lightly sugar `LambdaExpression`](https://github.com/narwhals-dev/narwhals/pull/3116/commits/02edde4e28b47fecb86cd62b11968e92220fefe1)

> What about any of {"el", "elem", "element"}?

Yep, went with 

[refactor: Rename to `elem`](https://github.com/narwhals-dev/narwhals/pull/3116/commits/d93e9321428403c017d2981e92ab5d6a182a3d87)

narwhals/_duckdb/expr_str.py
```diff
@@ -27,4 +29,30 @@ def to_date(self, format: str | None) -> DuckDBExpr:
         compliant_expr = self.compliant
         return compliant_expr.cast(compliant_expr._version.dtypes.Date())
 
+    def to_titlecase(self) -> DuckDBExpr:
+        if (version := self.compliant._backend_version) < (1, 2):  # pragma: no cover
+            msg = (
+                "`Expr.str.to_titlecase` is only available in 'duckdb>=1.2', "
+                f"found version {version!r}."
+            )
+            raise NotImplementedError(msg)
```
FBruzzesi: @dangotbanned thanks to `lambda_expr` we can remove this block, can't we?

dangotbanned: Uhh, yeah sure

I do *prefer* the error message you have there now - but if we're gonna later support `@requires` (https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2339532354) - then either works for me until then 🙂

FBruzzesi: Then let's keep that around until we have another approach ;)

dangotbanned: > Then let's keep that around until we have another approach ;)

😉 @FBruzzesi 

- [x] #3127

narwhals/_polars/utils.py
```diff
@@ -312,6 +312,9 @@ class PolarsStringNamespace(PolarsAnyNamespace[CompliantT, NativeT_co]):
     _accessor: ClassVar[Accessor] = "str"
 
     # NOTE: Use `abstractmethod` if we have defs to implement, but also `Method` usage
+    @abc.abstractmethod
+    def to_titlecase(self) -> CompliantT: ...
+
```
dangotbanned, line 331: Nice, my comment worked :smile:

narwhals/_polars/expr.py
```diff
@@ -401,6 +401,22 @@ class PolarsExprDateTimeNamespace(
...
+            native_result = (
+                native_expr.str.to_lowercase()
+                .str.extract_all(r"[a-z0-9]*[^a-z0-9]*")
+                .list.eval(pl.element().str.to_titlecase())
+                .list.join("")
+            )
+        else:
+            native_result = native_expr.str.to_titlecase()
+
+        return self.compliant._with_native(native_result)
```
dangotbanned, line 428: The full diff has an import outside of this range:

<p>

```diff
diff --git a/narwhals/_polars/expr.py b/narwhals/_polars/expr.py
index eaea6018f..1bde93b97 100644
--- a/narwhals/_polars/expr.py
+++ b/narwhals/_polars/expr.py
@@ -5,6 +5,7 @@ from typing import TYPE_CHECKING, Any, Callable, ClassVar, Literal
 import polars as pl
 
 from narwhals._polars.utils import (
+    BACKEND_VERSION,
     PolarsAnyNamespace,
     PolarsCatNamespace,
     PolarsDateTimeNamespace,
@@ -402,20 +403,16 @@ class PolarsExprStringNamespace(
     PolarsExprNamespace, PolarsStringNamespace[PolarsExpr, pl.Expr]
 ):
     def to_titlecase(self) -> PolarsExpr:
-        native_expr = self.native
-        backend_version = self.compliant._backend_version
-
-        if backend_version < (1, 5):  # pragma: no cover
-            native_result = (
-                native_expr.str.to_lowercase()
+        return self.compliant._with_native(
+            self.native.str.to_titlecase()
+            if BACKEND_VERSION >= (1, 5)
+            else (
+                self.native.str.to_lowercase()
                 .str.extract_all(r"[a-z0-9]*[^a-z0-9]*")
                 .list.eval(pl.element().str.to_titlecase())
                 .list.join("")
             )
-        else:
-            native_result = native_expr.str.to_titlecase()
-
-        return self.compliant._with_native(native_result)
+        )
 
     @requires.backend_version((0, 20, 5))
     def zfill(self, width: int) -> PolarsExpr:

```

</p>
</details> 

```suggestion
    def to_titlecase(self) -> PolarsExpr:
        return self.compliant._with_native(
            self.native.str.to_titlecase()
            if BACKEND_VERSION >= (1, 5)
            else (
                self.native.str.to_lowercase()
                .str.extract_all(r"[a-z0-9]*[^a-z0-9]*")
                .list.eval(pl.element().str.to_titlecase())
                .list.join("")
            )
        )
```

narwhals/_spark_like/expr_str.py
```diff
@@ -33,4 +35,30 @@ def to_date(self, format: str | None) -> SparkLikeExpr:
...
+            impl.is_sqlframe()
+            and (version := impl._backend_version()) < sqlframe_required_version
+        ):  # pragma: no cover
+            required_str = requires._unparse_version(sqlframe_required_version)
+            found_str = requires._unparse_version(version)
+            msg = (
+                f"`str.to_titlecase` is only available in 'sqlframe>={required_str}', "
+                f"found version {found_str!r}."
+            )
+            raise NotImplementedError(msg)
```
dangotbanned, line 51: I guess this would be a case that #3059 would be nice for?

This'll do for now though 😄

narwhals/_ibis/expr_str.py
```diff
@@ -81,3 +81,4 @@ def fn(expr: ir.StringColumn) -> ir.DateValue:
         return self.compliant._with_callable(fn)
 
     replace = not_implemented()
+    to_titlecase = not_implemented()
```
dangotbanned, line 84: https://github.com/narwhals-dev/narwhals/issues/3060#issuecomment-3276649037

> ibis: does not implement a needed functionality

Do you have any more specifics on this?

Some *possibly* useful methods:
- [StringValue.capitalize](https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.capitalize)
- [StringValue.upper](https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.upper)
- [StringValue.lower](https://ibis-project.org/reference/expression-strings#ibis.expr.types.strings.StringValue.lower)
- [ArrayValue.map](https://ibis-project.org/reference/expression-collections#ibis.expr.types.arrays.ArrayValue.map)
- [ArrayValue.join](https://ibis-project.org/reference/expression-collections#ibis.expr.types.arrays.ArrayValue.join)

narwhals/expr_str.py
```diff
@@ -455,6 +455,55 @@ def to_lowercase(self) -> ExprT:
...
+        """Modify strings to their titlecase equivalent.
+
+        Notes:
+            This is a form of case transform where the first letter of each word is
+            capitalized, with the rest of the word in lowercase.
+
+        Warning:
+            Different backends might follow different rules to determine what a "word" is:
+
+            - duckdb, polars and spark-like uses **non-alphanumeric** characters to
```
dangotbanned:
```suggestion
            - duckdb, polars and spark-like use **non-alphanumeric** characters to
```

narwhals/expr_str.py
```diff
@@ -455,6 +455,55 @@ def to_lowercase(self) -> ExprT:
...
+                define the word boundaries.
+            - pandas-like, pyarrow and dask use **non-alphabetic** characters to define
+                the word boundaries, matching the behavior of
+                [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title).
```
dangotbanned, line 472: I think this information is super helpful 🎉

Just spit-balling two other ways to present it

### 1. Use a table?
Haven't managed to get VSCode to render *any* kind of table, so this may be a non-starter 😂

I was thinking something like this:

> | Word boundaries (characters)                                                                       | Backends                      |
> | -------------------------------------------------------------------------------------------------- | ----------------------------- |
> | **non-alphanumeric**                                                                               | duckdb, polars and spark-like |
> | **non-alphabetic** (like [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title)) | dask, pyarrow and pandas-like |


### 2. Use two examples
https://github.com/narwhals-dev/narwhals/blob/63c5022e347cef3f821f725350cd9d39e6e476c6/narwhals/this.py#L8-L9

Would probably need different data to begin with.
E.g. something that demonstrates the difference when passed to one of each could be nice 🙂

tests/expr_and_series/str/to_titlecase_test.py
```diff
@@ -0,0 +1,62 @@
...
+        "They'Re Bill'S Friends From The Uk",
+        "To Infinity,And Beyond!",
+        "With123Numbers",
+        "__Dunder__Score_A1_.2B ?Three",
+    ]
+}
+REPLACEMENTS = {
+    "With123Numbers": "With123numbers",
+    "__Dunder__Score_A1_.2B ?Three": "__Dunder__Score_A1_.2b ?Three",
+}
```
dangotbanned: 100% my bad for adding this terrible name 🤦‍♂️ 

Can you think of something more descriptive?
E.g. why are these replacements needed?

dangotbanned (commented):
Thanks @FBruzzesi - a few notes/suggestions from me - but looking good 🥳

narwhals/_ibis/expr_str.py
FBruzzesi: The only issue is that I couldn't manage to make [StringValue.re_extract](https://ibis-project.org/reference/expression-strings.html#ibis.expr.types.strings.StringValue.re_extract) return all the matches :(

narwhals/expr_str.py
FBruzzesi: That's a great idea, I am pushing changes that render as follows:

image

tests/expr_and_series/str/to_titlecase_test.py
FBruzzesi: I am pushing an extra long name: `non_alphabetic_to_non_alphanumeric_boundary_mapping`

narwhals/expr_str.py
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2366350519

I like how that looks on the API Reference 👍 

> Haven't managed to get VSCode to render any kind of table, so this may be a non-starter 

What I meant was that IDE hover would ...

<details><summary>... get mangled here</summary>
<p>

image

</p>
</details> 

But also the width of the doc *itself* is insane 😂 

I tried moving the link as well ... 

<details><summary>... but that came out worse </summary>
<p>

image


</p>
</details> 

---

Compared to something like this, which IMO looks relatively good in all cases:

<details><summary>MkDocs</summary>
<p>

image

</p>
</details> 





<details><summary>IDE Hover</summary>
<p>

*Ignoring the weird VSCode indent* which is unrelated

image

</p>
</details> 


<details><summary>Source code</summary>
<p>

image

</p>
</details> 

<p>

```diff
diff --git a/narwhals/expr_str.py b/narwhals/expr_str.py
index 0374f47ed..4c649ff08 100644
--- a/narwhals/expr_str.py
+++ b/narwhals/expr_str.py
@@ -462,42 +462,53 @@ class ExprStringNamespace(Generic[ExprT]):
             This is a form of case transform where the first letter of each word is
             capitalized, with the rest of the word in lowercase.
 
-        Warning:
-            Different backends might follow different rules to determine what a "word" is:
-
-            | Word boundaries (characters)                                                                        | Backends                      | Example                                  |
-            | --------------------------------------------------------------------------------------------------- | ----------------------------- | ---------------------------------------- |
-            | *non-alphanumeric*                                                                                  | duckdb, polars and spark-like | `"with123numbers"` -> `"With123numbers"` |
-            | *non-alphabetic* (same as [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title)) | dask, pyarrow and pandas-like | `"with123numbers"` -> `"With123Numbers"` |
-
         Examples:
-            >>> import polars as pl
+            Different backends follow different rules to determine what a "word" is:
+
             >>> import narwhals as nw
-            >>> df_native = pl.DataFrame(
-            ...     {
-            ...         "quotes": [
-            ...             "'e.t. phone home'",
-            ...             "you talkin' to me?",
-            ...             "to infinity,and BEYOND!",
-            ...         ]
-            ...     }
-            ... )
-            >>> df = nw.from_native(df_native)
-            >>> df.with_columns(quotes_title=nw.col("quotes").str.to_titlecase())
+            >>> data = {
+            ...     "a": [
+            ...         "'e.t. phone home'",
+            ...         "you talkin' to me?",
+            ...         "to infinity,and BEYOND!",
+            ...         "with123numbers",
+            ...     ]
+            ... }
+            >>> to_titlecase = nw.col("a").str.to_titlecase().alias("titlecase")
+
+            `duckdb`, `polars` and `spark-like` define word boundaries by
+            *non-alphanumeric* characters:
+
+            >>> nw.DataFrame.from_dict(data, backend="polars").with_columns(to_titlecase)
             ┌─────────────────────────────────────────────────────┐
             |                 Narwhals DataFrame                  |
             |-----------------------------------------------------|
-            |shape: (3, 2)                                        |
+            |shape: (4, 2)                                        |
             |┌─────────────────────────┬─────────────────────────┐|
-            |│ quotes                  ┆ quotes_title            │|
+            |│ a                       ┆ titlecase               │|
             |│ ---                     ┆ ---                     │|
             |│ str                     ┆ str                     │|
             |╞═════════════════════════╪═════════════════════════╡|
             |│ 'e.t. phone home'       ┆ 'E.T. Phone Home'       │|
             |│ you talkin' to me?      ┆ You Talkin' To Me?      │|
             |│ to infinity,and BEYOND! ┆ To Infinity,And Beyond! │|
+            |│ with123numbers          ┆ With123numbers          │|
             |└─────────────────────────┴─────────────────────────┘|
             └─────────────────────────────────────────────────────┘
+
+            Whereas `dask`, `pyarrow` and `pandas-like` define word boundaries by
+            *non-alphabetic* characters (same as [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title)):
+
+            >>> nw.DataFrame.from_dict(data, backend="pandas").with_columns(to_titlecase)
+            ┌───────────────────────────────────────────────────┐
+            |                Narwhals DataFrame                 |
+            |---------------------------------------------------|
+            |                         a                titlecase|
+            |0        'e.t. phone home'        'E.T. Phone Home'|
+            |1       you talkin' to me?       You Talkin' To Me?|
+            |2  to infinity,and BEYOND!  To Infinity,And Beyond!|
+            |3           with123numbers           With123Numbers|
+            └───────────────────────────────────────────────────┘
         """
         return self._expr._with_elementwise(
             lambda plx: self._expr._to_compliant_expr(plx).str.to_titlecase()

```


</p>
</details>

narwhals/_ibis/expr_str.py
dangotbanned: Ahhh I see

`ibis` maps to only `regexp_extract`

https://github.com/ibis-project/ibis/blob/88f493bc37524fc9e9e31d57b166869bfa7666c0/ibis/backends/sql/compilers/base.py#L364

Whereas upstream from them in `sqlglot` there is `RegexpExtract` *and* `RegexpExtractAll` 

https://github.com/tobymao/sqlglot/blob/66f9501d76d087798bad93e578273ab2a45e2575/sqlglot/expressions.py#L7140-L7159


>  :(

Indeed 😢

narwhals/expr_str.py
FBruzzesi: @dangotbanned any new thought on this?

dangotbanned: > @dangotbanned any new thought on this?

@FBruzzesi did you see all the hidden details in my comment here? 

https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2366363563

FBruzzesi: Yes! I just wanted to make sure if that's your favorite way forward. I have some litimited access, and will give github codespaces a go later today to apply the changes :)

dangotbanned: Ah I see 🙂

Yeah, it doesn't need to be *exactly* what I put in the diff.
But I would like something that VSCode won't trip up on 

So if you can convey the same info with shorter examples, that could be another option?

tests/expr_and_series/str/to_titlecase_test.py
dangotbanned: Sorry I (😄)-reacted without giving context

> `non_alphabetic_to_non_alphanumeric_boundary_mapping`

May have swung *too hard* into the descriptive direction 🫣

But also
1. I don't have an alternative suggestion atm
2. It's only in a test after all

dangotbanned: How about splitting those cases into another test where we either:

1. `xfail` those backends that need it
2. Only test the backends that need it, asserting that they return those edge cases?

narwhals/expr_str.py
FBruzzesi: @dangotbanned I would say that the suggestion in https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2366363563 is not too far from the original bullet point list. I am pushing a change that reverts to such state plus some more details. Let me know your thoughts.

(I am in codespaces for the first time, hope to not mess it up 😂)

here: https://github.com/narwhals-dev/narwhals/pull/3116/commits/6128b49729e778a3f004c9c89431027a0d711e58

tests/expr_and_series/str/to_titlecase_test.py
FBruzzesi: proposed change in https://github.com/narwhals-dev/narwhals/pull/3116/commits/6128b49729e778a3f004c9c89431027a0d711e58

narwhals/expr_str.py
dangotbanned: Thanks @FBruzzesi that does sound reasonable, I'll try to preview this version in all the places later 😄

dangotbanned: (https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2405175758)

Oooh this looks nicer everywhere 🥳 

<details><summary>MkDocs</summary>
<p>

image

</p>
</details> 

<details><summary>IDE Hover</summary>
<p>


image

</p>
</details>

I have one more tweak in mind now, but I'll start a new thread

narwhals/expr_str.py
```diff
@@ -455,6 +455,59 @@ def to_lowercase(self) -> ExprT:
...
+        Notes:
+            This is a form of case transform where the first letter of each word is
+            capitalized, with the rest of the word in lowercase.
+
+        Warning:
+            Different backends might follow different rules to determine what a "word" is:
+
+            - duckdb, polars and spark-like use **non-alphanumeric** characters to
+                define the word boundaries.
+            - pandas-like, pyarrow and dask use **non-alphabetic** characters to define
```
dangotbanned: This puts more of a focus on the different bit

```suggestion
            - duckdb, polars and spark-like use non-**alphanumeric** characters to
                define the word boundaries.
            - pandas-like, pyarrow and dask use non-**alphabetic** characters to define
```

narwhals/expr_str.py
```diff
@@ -455,6 +455,59 @@ def to_lowercase(self) -> ExprT:
...
+                the word boundaries, matching the behavior of
+                [`str.title`](https://docs.python.org/3/library/stdtypes.html#str.title).
+
+            As an example of such difference, in the former case the string `"with123numbers"`
+            is mapped to `"With123numbers"` (notice lowercase **n** after the digits), while
+            in the latter to `"With123Numbers"` (notice uppercase **N** after the digits).
```
dangotbanned: Since this keeps all the different parts *mostly* aligned, it was an easier read for my smooth brain :wink:
```suggestion
            We can observe the difference with the string `"with123numbers"`:

            - non-**alphanumeric** -> `"With123numbers"`
                - notice lowercase **n** after the digits
            - non-**alphabetic** -> `"With123Numbers"`
                - notice uppercase **N** after the digits
```

dangotbanned: Compare to (https://github.com/narwhals-dev/narwhals/pull/3116#discussion_r2405496712)

image

image

tests/expr_and_series/str/to_titlecase_test.py
dangotbanned: Yep nice one!

dangotbanned (approved):
Thanks @FBruzzesi, my suggestions aren't blocking - looking great

banarwhal

PR #3118, opened 2025-09-11
enh: Support `keep={'first', 'last'}` in `{DataFrame,LazyFrame}.unique`

### LINKED ISSUE #3085: enh: add `order_by` argument to `DataFrame.unique` / `LazyFrame.unique`
Currently `keep='first'` / `keep='last'` aren't allowed for `LazyFrame` because of row-order

Just like how in [with_row_index](https://narwhals-dev.github.io/narwhals/api-reference/lazyframe/#narwhals.dataframe.LazyFrame.with_row_index) we have `order_by`, we should add it to `unique` as well

Rules:
- for `DataFrame`, it's optional
- for `LazyFrame`:
  - if `keep` is `'any'` or `'none'`, it's optional
  - if `keep` is `'first'` or `'last`, it's required

MarcoGorelli:
closes #3085

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_ibis/dataframe.py
+++ b/narwhals/_ibis/dataframe.py
@@ -2,22 +2,24 @@ from __future__ import annotations
 
 import operator
 from io import BytesIO
-from typing import TYPE_CHECKING, Any, Literal, cast
+from typing import TYPE_CHECKING, Any, cast
 
 import ibis
 import ibis.expr.types as ir
 
-from narwhals._ibis.utils import evaluate_exprs, native_to_narwhals_dtype
+from narwhals._ibis.expr import IbisExpr
+from narwhals._ibis.utils import evaluate_exprs, lit, native_to_narwhals_dtype
 from narwhals._sql.dataframe import SQLLazyFrame
 from narwhals._utils import (
     Implementation,
     ValidateBackendVersion,
     Version,
+    generate_temporary_column_name,
     not_implemented,
     parse_columns_to_drop,
     zip_strict,
 )
-from narwhals.exceptions import ColumnNotFoundError, InvalidOperationError
+from narwhals.exceptions import InvalidOperationError
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping, Sequence
@@ -30,7 +32,6 @@ if TYPE_CHECKING:
     from typing_extensions import Self, TypeAlias, TypeIs
 
     from narwhals._compliant.typing import CompliantDataFrameAny
-    from narwhals._ibis.expr import IbisExpr
     from narwhals._ibis.group_by import IbisGroupBy
     from narwhals._ibis.namespace import IbisNamespace
     from narwhals._ibis.series import IbisInterchangeSeries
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
@jklaise in case this is useful to you, [you can use Narwhals + DuckDB to generate SQL](https://narwhals-dev.github.io/narwhals/generating_sql/), and then optionally SQLGlot too to transpile to Postgres

- `sort` followed by `select`:

```python
import narwhals as nw
import duckdb
import sqlglot

conn = duckdb.connect()
conn.sql("""CREATE TABLE df (a int64, b int64);""")

df = nw.from_native(conn.table('df'))
sql = df.sort('a', 'b').select('a').to_native().sql_query()
print(sqlglot.transpile(sql, pretty=True, read='duckdb', write='postgres')[0])
```
outputs:
```sql
SELECT
  a
FROM (
  SELECT
    *
  FROM main.df
  ORDER BY
    a NULLS FIRST,
    b NULLS FIRST
) AS df
```

- `unique` without using `first`:

```python
import narwhals as nw
import duckdb
import sqlglot

conn = duckdb.connect()
conn.sql("""CREATE TABLE df (a int64, b int64);""")

df = nw.from_native(conn.table('df'))
sql = df.unique('b', keep='any').select('a').to_native().sql_query()
print(sqlglot.transpile(sql, pretty=True, read='duckdb', write='postgres')[0])
```
outputs:
```sql
SELECT
  a
FROM (
  SELECT
    *
    EXCEPT (nwc4027c227c90b3)
  FROM (
    SELECT
      *
    FROM (
      SELECT
        *,
        ROW_NUMBER() OVER (PARTITION BY b) AS nwc4027c227c90b3
      FROM main.df
    ) AS df
    WHERE
      (
        nwc4027c227c90b3 = 1
      )
  ) AS df
) AS df
```

This PR also adds the possibility to use `keep='first'` and `keep='last'`, so long as (in the lazy case at least) `order_by` is also specified

jklaise:
@MarcoGorelli thanks for the write-up - good to know this feature works!

Somewhat unrelated, I was bit by the duckdb+sqlglot route before when my `SELECT` statements wrt to specific columns result in `EXCLUDE` statements in the [DuckDB SQL](https://duckdb.org/docs/stable/sql/query_syntax/select.html) and sqlglot keeps the `EXCLUDE` statements in the Postgres transpilation but this is a completely different operator with a different syntax in Postgres, so the transpiled query is invalid.

MarcoGorelli:
thanks, interesting!

we can rewrite to not use `exclude`, no big deal, then the generate sql should be portable? e.g.
```sql
SELECT
  a,
  b
FROM (
  SELECT
    *
  FROM (
    SELECT
      *,
      ROW_NUMBER() OVER (PARTITION BY b) AS nw5d572dacc43786
    FROM main.df
  ) AS df
  WHERE
    (
      nw5d572dacc43786 = 1
    )
) AS df
```
(we probably should make the temporary column names more readable, but that's a separate issue)

narwhals/typing.py
```diff
@@ -339,13 +339,6 @@ def Binary(self) -> type[dtypes.Binary]: ...
 - *"last"*: Keep last unique row.
 """
 
-LazyUniqueKeepStrategy: TypeAlias = Literal["any", "none"]
```
MarcoGorelli, line 342: doesn't look like removing this has any effect https://github.com/search?q=%22LazyUniqueKeepStrategy%22+-path%3Asite-packages%2F+-path%3Anarwhals%2F&type=code

Issue #2789, opened 2025-07-04
bug: `from_native` on symbol of type `Any` is revealed to be `LazyFrame`


MarcoGorelli:
```python
from typing import Any
import narwhals as nw


def func(df: Any) -> None:
    reveal_type(nw.from_native(df))
```

I think it should be revealed as `DataFrame | LazyFrame`

dangotbanned:
Could you link the `@overload` that it is hitting?

If anyone wants to tackle this, I really need to stress the current overloads are very fragile 😩

EdAbati:
My guess is that it is hitting this:

https://github.com/narwhals-dev/narwhals/blob/cc223788d5e33fe30221d4a1526c74fe15d254d0/narwhals/translate.py#L265-L274

dangotbanned:
https://github.com/narwhals-dev/narwhals/issues/2789#issuecomment-3055979352

@EdAbati IIRC to get that `@overload`, we'd need to pass in 4 keyword-only arguments.

But the example in the OP doesn't pass in any keywords.
Also the concern was the type being revealed as `LazyFrame` - so I think it would be one that has `LazyFrame` in the return type

My guess would be that it is one of the first few `@overload`s - where `LazyFrame` needs to be defined *first* as it would overlap with `DataFrame` if it came after 

Anyway, I'm just guessing here as I've lost too much time trying to fix them in the past 😅

MarcoGorelli:
😄 thanks for looking into it

not sure we should care too much, perhaps it just doesn't need fixing as this isn't something anyone would realistically write, hence the low-prio tag, it's just something i'd noticed

Issue #2239, opened 2025-03-18
[Bug]: `TypeVar`(s) used in `nw.(BaseFrame|Series)` are recursive


dangotbanned:
### Describe the bug

Discovered during (https://github.com/narwhals-dev/narwhals/pull/2236#discussion_r1999715946) that we probably need to narrow the `TypeVar`(s) used for the `narwhals` (not compliant) types.

<details><summary>Screenshot</summary>
<p>

![image](https://github.com/user-attachments/assets/d866ade8-2c71-4096-9607-5cc19fe90e0b)

</p>
</details> 

### Steps or code to reproduce the bug

```py
import polars as pl
from typing_extensions import reveal_type

import narwhals as nw

series = pl.Series([1, 2, 3])
nw_series = nw.from_native(series, series_only=True)
reveal_type(nw_series)
# Type of "nw_series" is "narwhals.series.Series[polars.series.series.Series]"

nw_series_nest_1 = nw.from_native(nw_series, series_only=True)
reveal_type(nw_series_nest_1)
# Type of "nw_series_nest_1" is "narwhals.series.Series[narwhals.series.Series[polars.series.series.Series]]"

nw_series_nest_2 = nw.from_native(nw_series_nest_1, series_only=True)
reveal_type(nw_series_nest_2)
# Type of "nw_series_nest_2" is "narwhals.series.Series[narwhals.series.Series[narwhals.series.Series[polars.series.series.Series]]]"
```

### Expected results

The nested types shouldn't be valid and are currently inaccurate.

At runtime we have:

```py
>>> type(nw_series_nest_1._compliant_series)
narwhals._polars.series.PolarsSeries

>>> type(nw_series_nest_2._compliant_series)
narwhals._polars.series.PolarsSeries
```

### Actual results

`nw.from_native` is masking what's going on, so a simple example is to try and create that type manually.

```py
import polars as pl

from narwhals._polars.series import PolarsSeries
from narwhals.utils import Version
from narwhals.utils import parse_version

compliant_series = PolarsSeries(
    series, backend_version=parse_version(pl), version=Version.MAIN
)
nw_series_manual = nw.Series(compliant_series, level="full")
reveal_type(nw_series_manual)
# Type of "nw_series_manual" is "Series[Unknown]"
```

So far, that's expected until this is resolved in #2236:

https://github.com/narwhals-dev/narwhals/blob/275c5b639c0171a4cc23451f01795742826537fa/narwhals/series.py#L80-L82

[...]

PR #3125, opened 2025-09-12
fix(typing): Overhaul `@overload`s in `nw.from_native`

### LINKED ISSUE #1928: missing overload for `nw.from_native(obj, eager_only=True, series_only=True)`

dangotbanned:
Closes #1928

- [x] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- An unexpected **Part 4** to #2944
- [x] ~~Staying in draft~~ Left in draft until #3086 was merged 

<details><summary><b>Show tasks</b></summary>
<p>

- [x] `Native{DataFrame,Series}`
  - [x] Move changes into `nw._native`
  - [x] Do another pass on `NativeSeries`
    - `filter` might not be needed anymore
- [x] Move new `TypedDict`s into `nw._translate`
- [x] Reuse and extend `TypedDict`s for `stable`
  - [x] `v1` ([thread](https://github.com/narwhals-dev/narwhals/pull/3125#discussion_r2445239929))
    - [refactor(DRAFT): Start shrinking `v1`](https://github.com/narwhals-dev/narwhals/pull/3125/commits/9a8c0253c6821d11ee878f27b5369b6a07c290d6)
    - [refactor(typing): Add `AllowAny` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/c61af0318b8b774a32529deb36c8117f1def3981)
    - [refactor(typing): Add `AllowSeries` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/252bcdb03f5abd31b8597a6c53cc1c26bf753a32)
    - [refactor(typing): Add `OnlySeries` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/c88761865ad9b58685848e19be9e09fb6159e102)
    - [refactor(typing): Add `ExcludeSeries` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/99527d5f38400838c3ff1746a461d11dc0f42ab1)
    - [refactor(typing): Add `AllowLazy` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/1dc56f021278f1b1152f42548f9d75474ebab03d)
    - [refactor(typing): Add `OnlyEagerOrInterchange` variants](https://github.com/narwhals-dev/narwhals/pull/3125/commits/10e7ddf2a9088976a28cbacf563b2dd430142da3)
  - [x] [refactor(typing): Update `v2`](https://github.com/narwhals-dev/narwhals/pull/3125/commits/fd7ab1694035c16f8f4396ff6a81eb7ae8ef433c)
    - Mostly copy/paste from `main`
- [x] Update `from_native()` `@overload`s 1-3 to use `TypedDict`s
- [x] Investigate how `mkdocs` presents the new typing
  - Is there a benefit to adding docstrings to the `TypedDict`s?
  - Does `Unpack` work out-of-the-box?
  - Are the current `@overload`s useful/noisy?

</p>
</details> 

## Previews
I've collected a bunch of things here that may be missed from the diff alone.
Be sure to check out the `<details>` 😉 

### [fix(typing): More gracefully handle narwhals in overloads](https://github.com/narwhals-dev/narwhals/pull/3125/commits/0605936c7a76c1e4ec448be9c668d1a48ef34e46)

`Any` was previously being used on the first few overloads to solve a [recursion issue](https://github.com/narwhals-dev/narwhals/pull/2347/files#r2029455047).

But that had the downside of swallowing completions *until* you matched an overload with `native_object`:

<details><summary><b>Show before</b></summary>
<p>

image

</p>
</details> 

**But now**, we have something more useful

<details><summary><b>Show after</b></summary>
<p>

image

</p>
</details> 

### Changes to `Native{Series,DataFrame}`

As mentioned in (https://github.com/narwhals-dev/narwhals/pull/3125#discussion_r2344954012), these protocols needed to be 
adapted to make them disjoint across `pandas`, `polars`, `pyarrow`:

| Class             | `drop` | `filter` | `unique` | `value_counts` |
| ----------------- | ------ | -------- | -------- | -------------- |
| `pd.Series`       | ✔️      | ✔️        | ✔️        | ✔️              |
| `pl.Series`       | ❌      | ✔️        | ✔️        | ✔️              |
| `pa.ChunkedArray` | ❌      | ✔️        | ✔️        | ✔️              |
|                   |        |          |          |                |
| `pd.DataFrame`    | ✔️      | ✔️        | ❌        | ✔️              |
| `pl.DataFrame`    | ✔️      | ✔️        | ✔️        | ❌              |
| `pa.Table`        | ✔️      | ✔️        | ❌        | ❌              |

### MkDocs Comparison
These changes shrink the API docs for `from_native` *quite* a lot.

However, if the specific arguments combinations are wanted here - then an options is adding the `TypedDict`s *somewhere* and enabling [`signature_crossrefs`](https://mkdocstrings.github.io/python/usage/configuration/signatures/#with-signature-cross-references).
IIRC, that feature used to be insiders-only but looks to be available since https://squidfunk.github.io/mkdocs-material/blog/2025/11/11/insiders-now-free-for-everyone/

<details><summary><b>Show <code>main</code></b></summary>
<p>

image

image

image

</p>
</details> 

<details><summary><b>Show (<code>636f787</code>)</b></summary>
<p>

image

image

</p>
</details> 

> [!NOTE]
> I'm fine with less detail here, but `signature_crossrefs` is an option if this is blocking 🙂

### DIFF
--- a/narwhals/typing.py
+++ b/narwhals/typing.py
@@ -101,6 +101,7 @@ if TYPE_CHECKING:
 _ShapeT = TypeVar("_ShapeT", bound="tuple[int, ...]")
 _NDArray: TypeAlias = "np.ndarray[_ShapeT, Any]"
 _1DArray: TypeAlias = "_NDArray[tuple[int]]"
+_1DArrayBool: TypeAlias = "np.ndarray[tuple[bool], np.dtype[np.bool_]]"
 _1DArrayInt: TypeAlias = "np.ndarray[tuple[int], np.dtype[np.integer[Any]]]"
 _2DArray: TypeAlias = "_NDArray[tuple[int, int]]"  # noqa: PYI047
 _AnyDArray: TypeAlias = "_NDArray[tuple[int, ...]]"  # noqa: PYI047
@@ -351,6 +352,8 @@ Either a string or an object that implements [`__fspath__`], such as [`pathlib.P
 _T = TypeVar("_T")
 _Slice: TypeAlias = "slice[_T, Any, Any] | slice[Any, _T, Any] | slice[None, None, _T]"
 _SliceNone: TypeAlias = "slice[None, None, None]"
+# Boolean positions
+SizedMultiBoolSelector: TypeAlias = "Sequence[bool] | _T | _1DArrayBool"
 # Index/column positions
 SingleIndexSelector: TypeAlias = int
 _SliceIndex: TypeAlias = "_Slice[int] | _SliceNone"
@@ -364,7 +367,9 @@ SizedMultiNameSelector: TypeAlias = "Sequence[str] | _T | _1DArray"
 MultiNameSelector: TypeAlias = "_SliceName | SizedMultiNameSelector[_T]"
 # Mixed selectors
 SingleColSelector: TypeAlias = "SingleIndexSelector | SingleNameSelector"
-MultiColSelector: TypeAlias = "MultiIndexSelector[_T] | MultiNameSelector[_T]"
+MultiColSelector: TypeAlias = (
+    "MultiIndexSelector[_T] | MultiNameSelector[_T] | SizedMultiBoolSelector[_T]"
+)
 
 
 __all__ = [
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -44,7 +44,7 @@ body:
     required: true
 - type: textarea
   attributes:
-    label: Please run narwhals.show_version() and enter the output below.
+    label: Please run narwhals.show_versions() and enter the output below.
     render: shell
     description: |
       You could do:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,13 @@
 
 <!--
-# Thanks for contributing a pull request! 
+# Thanks for contributing a pull request!
 ## Please make sure you see our contribution guidelines: https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md
+-->
+
+# Description
+
+<!--
+## If you have comments or want to explain your changes, please do so in this section
 -->
 
 ## What type of PR is this? (check all applicable)
@@ -24,5 +30,3 @@
 - [ ] Code follows style guide (ruff)
 - [ ] Tests added
 - [ ] Documented the changes
-
-## If you have comments or can explain your changes, please do so below
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Get started!
   - [Roadmap](#roadmap)
   - [Used by](#used-by)
   - [Sponsors and institutional partners](#sponsors-and-institutional-partners)
+  - [Support](#support)
   - [Appears on](#appears-on)
   - [Why "Narwhals"?](#why-narwhals)
 
@@ -213,9 +214,31 @@ provided some funding / development time:
 If you contribute to Narwhals on your organization's time, please let us know. We'd be happy to add your employer
 to this list!
 
+## Support
+
+If you'd like to say "thank you", please give us a ⭐ star ⭐.
+
+Please contact [hello_narwhals@proton.me](hello_narwhals@proton.me) if you would like to:
+
+- Receive professional support (e.g., if you're using or would like to use Narwhals at your company).
+- Have any Narwhals fixes / features prioritised.
+- Commission any Narwhals plugins for new backends.
+
 ## Appears on
 
 Narwhals has been featured in several talks, podcasts, and blog posts:
+
+- [Inspiring Computing Podcast](https://www.inspiringcomputing.com/2107763/episodes/16702460-the-rise-of-narwhals-in-open-source)
+  The Rise of Narwhals in Open-Source
+
+- [PyCon DE & PyData 2025](https://youtu.be/DJk782DWcss)
+  How Narwhals is silently bringing pandas, Polars, DuckDB, PyArrow, and more together
+
+- [The Python Exchange March 2025](https://youtu.be/TvFWFlK-2po)
+  What Can Narwhals Do for You?
+
+- [PyData London 2025](https://youtu.be/r2PxJlO7_QA)
+  How Narwhals brings Polars, DuckDB, PyArrow, & pandas together
 
 - [Talk Python to me Podcast](https://youtu.be/FSH7BZ0tuE0)
   Ahoy, Narwhals are bridging the data science APIs
@@ -232,10 +255,10 @@ Narwhals has been featured in several talks, podcasts, and blog posts:
 - [The Real Python Podcast](https://www.youtube.com/watch?v=w5DFZbFYzCM)
   Narwhals: Expanding DataFrame Compatibility Between Libraries
 
-- [Pycon Lithuania](https://www.youtube.com/watch?v=-mdx7Cn6_6E)  
+- [Pycon Lithuania 2024](https://www.youtube.com/watch?v=-mdx7Cn6_6E)  
   Marco Gorelli - DataFrame interoperatiblity - what's been achieved, and what comes next?
 
-- [Pycon Italy](https://www.youtube.com/watch?v=3IqUli9XsmQ)  
+- [Pycon Italy 2024](https://www.youtube.com/watch?v=3IqUli9XsmQ)  
   How you can write a dataframe-agnostic library - Marco Gorelli
 
 - [Polars Blog Post](https://pola.rs/posts/lightweight_plotting/)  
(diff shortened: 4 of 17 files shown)


narwhals/typing.py
```diff
@@ -35,13 +35,19 @@ def columns(self) -> Any: ...
...
     class NativeLazyFrame(NativeFrame, Protocol):
         def explain(self, *args: Any, **kwargs: Any) -> Any: ...
 
+    # Needs to have something `NativeDataFrame` doesn't?
     class NativeSeries(Sized, Iterable[Any], Protocol):
         def filter(self, *args: Any, **kwargs: Any) -> Any: ...
+        # `pd.DataFrame` has this - the others don't
+        def value_counts(self, *args: Any, **kwargs: Any) -> Any: ...
+        # `pl.DataFrame` has this - the others don't
+        def unique(self, *args: Any, **kwargs: Any) -> Any: ...
```
dangotbanned: ### Issue 1
`NativeDataFrame` and `NativeSeries` could overlap.

Although we don't define `NativeDataFrame.filter` - the classes that match it *do*:
- [`pandas.DataFrame.filter`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.filter.html)
- [`polars.DataFrame.filter`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.filter.html)
- [`pyarrow.Table.filter`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.filter)

That was leading to the strange `@overload` order and parameters changing the return type.

**But now** if we started with `NativeSeries`, then both `series_only` and `allow_series` will preserve that detail:

<details><summary><b>Prove it bro</b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/a0095b1d6c9cc4c5904edc7505c78700a87499be/tests/translate/from_native_test.py#L621-L723

</p>
</details>

narwhals/translate.py
```diff
@@ -99,150 +101,87 @@ def to_native(
...
-    pass_through: Literal[True],
-    eager_only: Literal[True],
-    series_only: Literal[False] = ...,
-    allow_series: Literal[True],
-) -> DataFrame[IntoDataFrameT] | Series[IntoSeriesT]: ...
+class OnlyEager(TypedDict, total=False):
+    pass_through: bool
+    eager_only: Required[Literal[True]]
+    series_only: bool
+    allow_series: bool | None
```
dangotbanned: ### Issue 2
There's too many combinations we need to account for in `@overload`s.

- 3x different, true `native_object`s
- 3x narwhals objects pretending to be native
- 4x flags
  - They all have a default, so needs `...`
  - `allow_series: bool | None` just adds to it

So instead of that:
- I've narrowed it down to 1-2 keywords that define unique behavior
- Given that a name
- Made a [`TypedDict`](https://typing.python.org/en/latest/spec/typeddict.html)
- Then relied on [`Unpack`](https://typing.python.org/en/latest/spec/typeddict.html#interaction-with-unpack) and [`Required`](https://typing.python.org/en/latest/spec/typeddict.html#required-and-notrequired) to merge a whole bunch of them

dangotbanned:
@MarcoGorelli this one isn't ready for a proper review yet, but I was curious about your thoughts on the idea generally?

I did some testing locally and I think depending on how these issues have progressed - docs might be one more area for experimenting 

- https://github.com/microsoft/pylance-release/issues/5006
- https://github.com/microsoft/pylance-release/issues/7602

MarcoGorelli:
> 3x different, true native_objects

tbh I'm still not too keen on matching on the native object, i kind of feel like `nw.from_native(obj: T)` should return `DataFrame[T] | LazyFrame[T]`, regardless of `T`

dangotbanned:
> > 3x different, true native_objects
> 
> tbh I'm still not too keen on matching on the native object, i kind of feel like `nw.from_native(obj: T)` should return `DataFrame[T] | LazyFrame[T]`, regardless of `T`

That isn't possible due to the issue from:

- #2239


Which is also why `Any` will match the first `@overload`, which means `(obj: Any) -> Any`

- #2789

Possibly these sections of the spec may be helpful?

- [Overloads - Implementation consistency](https://typing.python.org/en/latest/spec/overload.html#implementation-consistency)
- [Overloads - call evaluation](https://typing.python.org/en/latest/spec/overload.html#overload-call-evaluation)
- [Special types - Any](https://typing.python.org/en/latest/spec/special-types.html#any)
- [Type system concepts](https://typing.python.org/en/latest/spec/concepts.html#type-system-concepts)
  - Particularly chunky, but it explains `Any` in a **lot** of detail
---


These are the issues that come from having such a heavily polymorphic function I'm afraid 😢 
I'm hoping to make the best of a tricky situation and move the overloads into something that is (IMO) more maintainable

narwhals/_translate.py
```diff
@@ -184,3 +185,116 @@ class ToNarwhals(Protocol[ToNarwhalsT_co]):
...
+    #   - bool, bool
+    #   - True, bool
+    #   - bool, True
+    eager_only: bool  # False
+    # no lazyframe
+    eager_or_interchange_only: bool  # False
+
+    # same as others
+    series_only: bool  # False
+    allow_series: bool | None  # None
```
dangotbanned: I'm doing a pretty delicate dance here, `v1` is going to be a challenge but worth it 😅

dangotbanned: As of (https://github.com/narwhals-dev/narwhals/pull/3125/commits/c88761865ad9b58685848e19be9e09fb6159e102), I'm around 50% of the way through `v1`

Already at **-261** lines in `v1` alone 🥳 

### Updated
😄 

image

narwhals/stable/v2/__init__.py
```diff
@@ -191,7 +201,7 @@ def __getitem__(
     ) -> Series[Any] | Self | Any:
         return super().__getitem__(item)
 
-    def get_column(self, name: str) -> Series:
+    def get_column(self, name: str) -> Series[Any]:
```
dangotbanned, line 204: `v1.Series` only had a default `TypeVar` to avoid downstream noise from making it generic
- https://github.com/narwhals-dev/narwhals/pull/2110#discussion_r1973885324
- #1412

That shouldn't have carried over into `v2`

tests/translate/from_native_test.py
```diff
@@ -436,26 +490,288 @@ def test_pyspark_connect_deps_2517(constructor: Constructor) -> None:  # pragma:
...
+def test_from_native_lazyframe_exhaustive() -> None:  # noqa: PLR0914, PLR0915
+    pytest.importorskip("polars")
+    pytest.importorskip("typing_extensions")
+
+    import polars as pl
+    from typing_extensions import assert_type
+
+    pl_ldf = pl.LazyFrame(data)
+
+    pl_1 = nw.from_native(pl_ldf)
```
dangotbanned, line 520: Just to get ahead of the obvious question:

> Do we **need** all of this?

The answer is complicated.

IMO, the real source of all the typing issues with `from_native` has been that we test very little in terms of typing for such a complex function.
I think that the primary entry point for `narwhals` should require much more care to be sure we don't regress.

That's my opinion, but I wouldn't feel so confident in making such big changes to the typing without this in place.

I'm guessing Marco is familiar with this from `numpy`?

- https://github.com/numpy/numpy/blob/1aa31099403cbf5ba17dab857666986d57749834/numpy/typing/tests/data/reveal/array_constructors.pyi
- https://github.com/numpy/numpy/blob/1aa31099403cbf5ba17dab857666986d57749834/numpy/typing/tests/data/reveal/ma.pyi


Doing this in `.pyi` files rather than (mostly) in a `TYPE_CHECKING` block is another option 🙂

dangotbanned: This issue won't be resolved by this PR, but it does highlight the same gap

- [ ] #3319

We have no typing coverage for the example on the `README.md`?

```py
import narwhals as nw


def agnostic_function(
    df_native: IntoFrameT,
    date_column: str,
    price_column: str,
) -> IntoFrameT:
    return (
        nw.from_native(df_native)
        .sort(date_column)
        .to_native()
    )
```

`mypy` complains about this already 😳 

```py
narwhals\/.py:15: error: Value of type variable "IntoLazyFrameT" of "from_native" cannot be "IntoFrameT"  [type-var]
            nw.from_native(df_native)
            ^~~~~~~~~~~~~~~~~~~~~~~~~
```

MarcoGorelli (approved):
looks like the testing is exhaustive

thanks @dangotbanned !

dangotbanned:
Thanks @MarcoGorelli 

> looks like the testing is exhaustive

I'm exhausted just looking at them

MarcoGorelli:
aw i'm sorry for this api 😢

dangotbanned:
> aw i'm sorry for this api 😢

Don't be silly, I'm dunking on myself here 😄

PR #3130, opened 2025-09-13
chore: Add `CompliantNamespace.is_native`

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- To help with https://github.com/narwhals-dev/narwhals/pull/2978#discussion_r2346713884
- #2978
- https://github.com/MarcoGorelli/narwhals-daft/pull/15

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

PR #3134, opened 2025-09-15
test: fix `version_test` failing on old `venv`

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/3130#pullrequestreview-3223830780
- https://github.com/narwhals-dev/narwhals/pull/2978#discussion_r2282989454
- https://github.com/narwhals-dev/narwhals/pull/2978#issuecomment-3199901467
- https://github.com/narwhals-dev/narwhals/pull/2978#issuecomment-3199976748

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/version_test.py
+++ b/tests/version_test.py
@@ -1,14 +1,25 @@
 from __future__ import annotations
 
 import re
+from importlib.metadata import distribution
 from pathlib import Path
 
 import narwhals as nw
+from narwhals._utils import parse_version
 
 
-def test_version_matches_pyproject() -> None:
+def test_package_version() -> None:
+    version = nw.__version__
+    dist_version = distribution("narwhals").version
+
     with Path("pyproject.toml").open(encoding="utf-8") as file:
         content = file.read()
         pyproject_version = re.search(r'version = "(.*)"', content).group(1)  # type: ignore[union-attr]
 
-    assert nw.__version__ == pyproject_version
+    assert isinstance(version, str)
+    if version != pyproject_version:
+        # NOTE: metadata from venv is outdated (https://github.com/narwhals-dev/narwhals/pull/3130#issuecomment-3291578373)
+        version_comp = parse_version(version)
+        assert version_comp < parse_version(pyproject_version)
+        assert version_comp == parse_version(dist_version)
+    assert version == dist_version


tests/version_test.py
```diff
@@ -1,14 +1,26 @@
...
+    assert match is not None
+    pyproject_version = match.group(1)
+
+    assert isinstance(version, str)
+    if version != pyproject_version:  # pragma: no cover
+        # NOTE: metadata from venv is outdated (https://github.com/narwhals-dev/narwhals/pull/3130#issuecomment-3291578373)
+        version_comp = parse_version(version)
+        assert version_comp < parse_version(pyproject_version)
+        assert version_comp == parse_version(dist_version)
+    assert version == dist_version
```
FBruzzesi: While I understand that locally this can be problematic (it happened to me multiple times), I am not sure that working around the test itself is the best solution here.

- In CI this should never happen, unless we are _actually_ doing something wrong, and if that's the case, we could end up not noticing it.
- I would suggest to add a message in the current assert that explicitly suggestion why this happens and what to do, something along the lines of:

```diff
- assert nw.__version__ == pyproject_version
+ msg = "metadata is out of sync, either ignore this failure locally or re-install narwhals: `uv pip install -e .`"
+ assert nw.__version__ == pyproject_version, msg

dangotbanned: Thanks for the review @FBruzzesi 

> In CI this should never happen, unless we are actually doing something wrong, and if that's the case, we could end up not noticing it.

You make a good point! 😄

Would (https://github.com/narwhals-dev/narwhals/pull/3134/commits/2f22890a415500c507278f46769e783db825d773) and (https://github.com/narwhals-dev/narwhals/pull/3134/commits/0500e3c3844017e7afb52f6d32043d844ec6a7b7) help alleviate this concern?
(*provided I've RTFM correctly* 😂)

dangotbanned: https://github.com/narwhals-dev/narwhals/pull/3134#discussion_r2349707956
>  I am not sure that working around the test itself is the best solution here

I get that using `pyproject.toml` provides a proxy value with a different static source (+1).
But as we don't use a lockfile, I don't think it quite fits [this definition of assertion](https://en.wikipedia.org/wiki/Assertion_(software_development))
> ... that always should evaluate to true at that point in code execution

We've both *sadly* experienced this not being an *always* - because of how it interacts with [`-e`](https://docs.astral.sh/uv/pip/packages/#editable-packages) 😔

So the question for me was:
*Is there anything else we can confidently assert, where a failure still indicates a regression?*

Given that this is a very thin wrapper around [`importlib.metadata.version`](https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.version) our options are limited:

https://github.com/narwhals-dev/narwhals/blob/8bee263eaddde22d9b80177ff81eab816bcf9213/narwhals/__init__.py#L176-L183

**But** we can use that to our advantage! To me, that implies the likely points for breakage are:
1. A future refactor changes *how* we wrap `version`
2. We revert back to *not* using `version`

In either case - we'll know that something is wrong because `nw.__version__` no longer returns a string that is equal to `importlib.metadata.version("narwhals")`.

FBruzzesi (approved):
Thanks @dangotbanned 🙏🏼 nice trick for GHA 😉

PR #2731, opened 2025-06-25
chore(typing): More `isinstance_or_issubclass` overloads

dangotbanned:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- #2717
- #1997

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below

While I was looking for unreachable code, I noticed some `isinstance_or_issubclass` cases were marked *incorrectly* as unreachable.
They *are* reachable, but when I wrote (#1997) we only had up to a **3**-tuple.
The current max of **7** wasn't covered, so this PR fixes that

## Before 1
image

## After 1
image

## Before 2
image

## After 2
image

PR #3139, opened 2025-09-18
refactor(typing): Remove now-unused `isinstance_or_issubclass` overloads

dangotbanned:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Full revert of #2731
- Not needed since #2969

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Now, we only have 1 case which uses the 2x `@overload`(s).
I'm keeping the 3x as well for now, just in case that changes 🙂

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -146,10 +146,6 @@ if TYPE_CHECKING:
     _T1 = TypeVar("_T1")
     _T2 = TypeVar("_T2")
     _T3 = TypeVar("_T3")
-    _T4 = TypeVar("_T4")
-    _T5 = TypeVar("_T5")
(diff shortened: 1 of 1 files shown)


dangotbanned:
> As a general comment, the typing for this function seems a bit hard to maintain and not very scalable 

Couldn't agree more tbf 😂

(#2371) was really just trying to add typing to a complex situation we were *already in*

If we make some progress on (#2486), then we may be able to remove the need for it entirely 🙏

PR #3045, opened 2025-08-27
refactor: Organizing `Compliant*` APIs

dangotbanned:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2713
- [x] #3054

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

## If you have comments or can explain your changes, please do so below
- Everything listed in ([comment](https://github.com/narwhals-dev/narwhals/issues/2713#issuecomment-3225322537)) has now been moved/already had been
- Started on labelling ([comment](https://github.com/narwhals-dev/narwhals/issues/2713#issuecomment-3221755890)) what's `narwhals` vs `polars`
  - (https://github.com/narwhals-dev/narwhals/pull/3045/commits/a5f443232030705ce400559bddc01cabfc89e3e0) is entirely to help make decisions easier

> [!NOTE]
> All the threads are follow-up ideas

PR #3142, opened 2025-09-18
chore(typing): Ignore another `EagerDataFrame` intermittent [False Negative]

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- #3045

## Checklist

- [ ] Tests added
- [ ] Documented the changes

Been seeing this locally for a few days, but never in CI

image

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -79,7 +79,7 @@ jobs:
           uv venv -p 3.12
           git log
       - name: install-basics
-        run: uv pip install --upgrade tox virtualenv setuptools hatch --system
+        run: uv pip install --upgrade tox virtualenv setuptools hatch "click!=8.3.0" --system
       - name: install-marimo-dev
         run: |
           cd marimo
@@ -409,7 +409,7 @@ jobs:
       - name: create-hatch-env
         run: |
           cd formulaic
-          uv pip install hatch hatch-vcs --system
+          uv pip install hatch hatch-vcs "click!=8.3.0" --system
       - name: install-formulaic-dev
         run: |
           cd formulaic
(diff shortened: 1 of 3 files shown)


FBruzzesi:
Out of curiosity, do you know why this is the case?

dangotbanned:
@FBruzzesi 

> Out of curiosity, do you know why this is the case?

Maybe related to https://discuss.python.org/t/unsoundness-of-contravariant-self-type/86338

But I guess to zoom in on it a bit more ...

I think the issue is `EagerExprT` gets bound in `EagerDataFrame.aggregate` - but then *occasionally* fails to be re-used for `EagerDataFrame.select`.

My intention for the typing is to say:
> `EagerDataFrame` is generic over `EagerDataFrame._expr: type[EagerExprT]`
> The type of `EagerExprT` should bind to the class

This is problematic because here that type is **contravariant**:

https://github.com/narwhals-dev/narwhals/blob/76dbde29adbf41edce4d4eda09a0c6159a24415a/narwhals/_compliant/dataframe.py#L148

But here that type is **invariant**:

https://github.com/narwhals-dev/narwhals/blob/76dbde29adbf41edce4d4eda09a0c6159a24415a/narwhals/_compliant/dataframe.py#L324

... and then in `EagerExpr` *itself*, it depends on an **invariant** `EagerDataFrame` 😳:

https://github.com/narwhals-dev/narwhals/blob/76dbde29adbf41edce4d4eda09a0c6159a24415a/narwhals/_compliant/expr.py#L213-L226

MarcoGorelli (approved):
thanks!

PR #3147, opened 2025-09-20
feat: Add `prefix` argument to `generate_temporary_column_name`

### LINKED ISSUE #3121: enh: readable temporary column names
thought about this in https://github.com/narwhals-dev/narwhals/pull/3118#issuecomment-3280818136

i think `generate_temporary_column_name` could have a `prefix: str = 'nw'` argument, so we can specify the prefix and get a more readable (but still random) column name at at the other end

For example, instead of `nw5d572dacc43786`, we could get `row_idx_acc43786`

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3121

## Checklist

- [x] Tests added
- [x] Documented the changes

Added explicit prefix only on lazy backends operations since these are the ones that a user might be able to investigate in various way. Not really applicable in the eager case

### DIFF
--- a/narwhals/_utils.py
+++ b/narwhals/_utils.py
@@ -1248,17 +1248,19 @@ def is_ordered_categorical(series: Series[Any]) -> bool:
 
 
 def generate_unique_token(
-    n_bytes: int, columns: Container[str]
+    n_bytes: int, columns: Container[str], prefix: str = "nw"
 ) -> str:  # pragma: no cover
     msg = (
         "Use `generate_temporary_column_name` instead. `generate_unique_token` is "
         "deprecated and it will be removed in future versions"
     )
     issue_deprecation_warning(msg, _version="1.13.0")
-    return generate_temporary_column_name(n_bytes=n_bytes, columns=columns)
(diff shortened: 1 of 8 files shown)


narwhals/_utils.py
```diff
@@ -1270,12 +1273,15 @@ def generate_temporary_column_name(n_bytes: int, columns: Container[str]) -> str
...
+        ... )
+        >>> temp_name not in columns and temp_name.startswith("foo")
+        True
     """
     counter = 0
     while True:
-        # Prepend `'nw'` to ensure it always starts with a character
-        # https://github.com/narwhals-dev/narwhals/issues/2510
-        token = f"nw{token_hex(n_bytes - 1)}"
+        token = f"{prefix}{token_hex(n_bytes - 1)}"
```
FBruzzesi, line 1284: Should we ensure that prefix is not an empty string?

MarcoGorelli: naah

Issue #1943, opened 2025-02-05
Docstring guidelines


MarcoGorelli:
In #1942 I mentioned a desire to keep the package size down. One way to do this is by reducing overly-long docstrings.

Here are some general guidelines:
- The examples should be clear and to-the-point
- Most docstrings should just be:
  - import dataframe library (usually pandas, pyarrow, or duckdb)
  - import narwhals
  - construct native dataframe. Make it as small as possible: 2 rows and 2 columns should be plenty for most docstrings
  - call `nw.from_native` and apply operation
  - only show a single example
- Variety between docstrings is encouraged, they don't all need to follow the same exact template. They can use different libraries, some can end with `to_native`, some can use selectors or `nw.all`...this is all OK. Just keep it minimal
- Essential functionality (such as DataFrame.filter, DataFrame.with_columns, DataFrame.select), it's OK to have even 2-3 examples showing different ways to use it
- For examples which use null values, use a library with proper support for them (i.e. not pandas with classical data types). Such docstrings should also link to the [pandas null handling](https://narwhals-dev.github.io/narwhals/pandas_like_concepts/null_handling/) page

---

Also, the reason many of the current docstrings are so long is because that's how I'd originally encouraged them to be. I think that was a mistake on my part, I need to acknowledge this, and as the project has evolved, it's time to revisit

FBruzzesi:
Should we just add these guidelines to the contributing guide? If I am new to a repo/project, I would not be searching for issues to know how to write docs 🙈

dangotbanned:
@MarcoGorelli not sure how this works, but...

Found an example function with overloads and no returns section (via https://mkdocstrings.github.io/#used-by)

https://textual.textualize.io/api/logger/#textual.work

![Image](https://github.com/user-attachments/assets/94f83d47-53a3-4d7b-97ba-55bcaf3a84f1)

![Image](https://github.com/user-attachments/assets/12fdb565-96e5-4da1-aa8c-86d224522109)

MarcoGorelli:
thanks - so, if there's overloads, then the signature shows, but not if there's not (e.g. https://textual.textualize.io/api/logger/#textual.on)

looks like for `on` the return type isn't visible at all from the docs?

dangotbanned:
> thanks - so, if there's overloads, then the signature shows, but not if there's not (e.g. [textual.textualize.io/api/logger#textual.on](https://textual.textualize.io/api/logger/#textual.on))
> 
[...]

PR #3148, opened 2025-09-20
feat: Add `{nw,DataFrame}.from_dicts`

### LINKED ISSUE #3145: [Enh]: Add nw.from_records to convert a sequence of rows to a data frame

_No response_

[Pandas](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.from_records.html) and [Polars](https://docs.pola.rs/api/python/stable/reference/api/polars.from_records.html) both offer a `from_records` function that takes a sequence of items representing rows and returns a DataFrame.

Narwhals could include this function as a companion to `narwhals.from_dict`.

Implement `narwhals.from_records` along the same lines as `narwhals.from_dict`.

One alternative I have tried is to manually "transpose" a list of dicts to a dict of lists. Something like this:

```python
list_of_dicts = {}

columns = [...] # column names appearing in dicts
dict_of_lists = {colname: [] for colname in columns}

for row in list_of_dicts:
    for col in columns:
        dict_of_lists[col].append(row.get(col))

df = nw.from_dict(dict_of_lists, backend=...)

```

The above can work if you know the columns in advance and the dictionary values are of the correct types

I will look into raising a PR to add this. The contributing guidelines look very helpful. On the other hand, if there is a good reason not to add this function, that I have missed, please feel free to close the issue. Thank you!

felixgwilliams:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3145
- Closes #3145

## Checklist

- [x] Tests added
- [x] Documented the changes

This PR adds a `from_dicts` function and methods, that can be used to create a data frame from a sequence of dicts that represent rows. It is available for eager backends.

## Tracking
- [x] https://github.com/pola-rs/polars/pull/24584


narwhals/dataframe.py
```diff
@@ -610,6 +610,59 @@ def from_dict(
...
+        *,
+        backend: IntoBackend[EagerAllowed],
+    ) -> DataFrame[Any]:
+        """Instantiate DataFrame from a sequence of dictionaries representing rows.
+
+        Notes:
+            For pandas-like dataframes, conversion to schema is applied after dataframe
+            creation.
+
+        Args:
```
FBruzzesi:
```suggestion
        Arguments:
```

narwhals/dataframe.py
```diff
@@ -610,6 +610,59 @@ def from_dict(
...
+            data: Sequence of dictionaries to create DataFrame from.
```
FBruzzesi: Nitpicking a bit on the description, stealing from polars:

```suggestion
            data: Sequence with dictionaries mapping column name to value.
```

narwhals/dataframe.py
```diff
@@ -610,6 +610,59 @@ def from_dict(
...
+            schema: The DataFrame schema as Schema or dict of {name: type}. If not
+                specified, the schema will be inferred by the native library.
+            backend: specifies which eager backend instantiate to.
```
FBruzzesi: Keep it consistent:

```suggestion
            backend: Specifies which eager backend instantiate to.
```

narwhals/functions.py
```diff
@@ -331,6 +331,74 @@ def _from_dict_no_backend(
     return data, native_namespace
 
 
+@deprecate_native_namespace(warn_version="1.26.0")
```
FBruzzesi: @dangotbanned @MarcoGorelli should we keep this around also for new functionalities? Personally I would rather avoid  `native_namespace` completely. WDYT?

narwhals/functions.py
```diff
@@ -331,6 +331,74 @@ def _from_dict_no_backend(
...
+def from_dicts(
+    data: Sequence[dict[str, Any]],
+    schema: IntoSchema | None = None,
+    *,
+    backend: IntoBackend[EagerAllowed],
+    native_namespace: ModuleType | None = None,  # noqa: ARG001
+) -> DataFrame[Any]:
+    """Instantiate DataFrame from a sequence of dictionaries representing rows.
```
FBruzzesi, line 340: Same docstring comments as for the DataFrame classmethod

narwhals/dataframe.py
```diff
@@ -610,6 +610,59 @@ def from_dict(
         )
         raise ValueError(msg)
 
+    @classmethod
+    def from_dicts(
+        cls,
+        data: Sequence[dict[str, Any]],
```
FBruzzesi: Should we use 

```suggestion
        data: Sequence[dict[str, PythonLiteral]],
```

(here and everywhere else)

Summoning the typing master @dangotbanned

narwhals/functions.py
felixgwilliams: I found that the `test_stable_api_docstrings` test fails if I keep this. `from_dict` (which I was copying from) is skipped in this test. All the other functions don't have `native_namespace` so a new function shouldn't have it, so I removed it. Please let me know if you want it put back in.

FBruzzesi:
Thanks for all the adjustments @felixgwilliams - I would stall for a moment waiting for an opinion on https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2366077741, aside that the feature seems ready  👌🏼

narwhals/dataframe.py
dangotbanned: Thanks for the ping 😉 

> (here and everywhere else)

Yep, cannot be stressed enough - changing just one will make typing angery

---

So my *instinct* would be to use this:

```py
        data: Sequence[Mapping[str, Any]],
```

Which is closer to `nw.DataFrame.from_dict`

https://github.com/narwhals-dev/narwhals/blob/e92fcc989a5857d5b8a86929fac3b4786a913574/narwhals/dataframe.py#L554-L561

My *guess* is that `pyarrow` may actually require a `dict` - but the typing makes it look fine
- https://github.com/zen-xu/pyarrow-stubs/blob/dfe07a8415e516fbd90c0a46b0e1dddf2292a6f3/pyarrow-stubs/__lib_pxi/table.pyi#L2037-L2043

`pandas-stubs` seems to like that also 🎉 

- https://github.com/pandas-dev/pandas-stubs/blob/2f633544659a83d94cb7f527de21a5667d8a62c6/pandas-stubs/core/frame.pyi#L542-L557

Weirdly, `polars` is asking for a `dict` here 🤔 

- https://github.com/pola-rs/polars/blob/1dc7792dcbe5531700a501a5704f5a78a63187c4/py-polars/polars/convert/general.py#L117-L223

But the impl just calls `pl.DataFrame`, which uses a wider type:

- https://github.com/pola-rs/polars/blob/1dc7792dcbe5531700a501a5704f5a78a63187c4/py-polars/polars/dataframe/frame.py#L359-L369
- https://github.com/pola-rs/polars/blob/664656ea91291f8f4b4eb9d42af2a9da5394616e/py-polars/polars/_typing.py#L217-L227

AFAICT, the `polars` case will end up here - which uses [`@functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) and quite explicitly supports both `dict` and `Mapping` 🥳  
- https://github.com/pola-rs/polars/blob/664656ea91291f8f4b4eb9d42af2a9da5394616e/py-polars/polars/_utils/construction/dataframe.py#L696-L731

narwhals/functions.py
dangotbanned: >  Personally I would rather avoid native_namespace completely. WDYT?

@FBruzzesi I agree with this 👍 

We don't need the backcompat when nobody could call the function before this PR 🙂

narwhals/dataframe.py
felixgwilliams: Thanks for checking this. Is the verdict that the value type for the dict/mapping should stay as `Any` rather than `PythonLiteral`?

Also is it worth adding a test to ensure a `Mapping` other than a dictionary works or do we rely on the typing of pandas/polars/pyarrow?

narwhals/functions.py
dangotbanned: On a similar note - I'm gonna add some commits making the `functions.py` version just forward to the `dataframe.py` one(s).


@felixgwilliams absolutely no fault on your part here!
There's some longer term thing where the `pragma: no cover` path is planned to be removed as it isn't functional

(https://github.com/narwhals-dev/narwhals/issues/2786#issuecomment-3201413149)

narwhals/dataframe.py
dangotbanned: > Thanks for checking this. Is the verdict that the value type for the dict/mapping should stay as `Any` rather than `PythonLiteral`?

I would say `Any` makes more sense because upstream seems to be using that as well.

Even in a case like [`polars.DataFrame.to_dicts`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.to_dicts.html) which explictly mentions

> *"a dictionary of Python-native values."*

The value type still uses `Any`:

```py
    def to_dicts(self) -> list[dict[str, Any]]:
```

dangotbanned: > Also is it worth adding a test to ensure a `Mapping` other than a dictionary works or do we rely on the typing of pandas/polars/pyarrow?

Great question! 😄 

Yeah I think that would be a good idea and had a similar thought in [`Series.from_iterable`](https://github.com/narwhals-dev/narwhals/pull/2933#discussion_r2249983202)

> [!NOTE]
> No need to be so exhaustive here though 

A good candidate would be something that behaves like this:

```py
import types
from collections.abc import Mapping

mapping = types.MappingProxyType({})
>>> isinstance(mapping, dict), isinstance(mapping, Mapping)
(False, True)
```

Since that difference is what will throw the input into a different path if upstream doesn't *really* support `Mapping`

FBruzzesi: @felixgwilliams fancy using `types.MappingProxyType` as a non-dict mapping instead of creating a new class just for testing as in https://github.com/narwhals-dev/narwhals/pull/3148/commits/076a821f2f7d57e61f1dafea43d019331308bc13?
Sorry for the extra iteration here 😇

felixgwilliams: Ah OK I misunderstood the comment. I'll just use `types.MappingProxyType`. Pushing now

dangotbanned: Sorry my bad, not sure why I phrased (https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2366199504) that way 

`types.MappingProxyType` works, no need to create/find another 👍

dangotbanned: Ah actually, maybe there's some version-dependent ordering stuff going on here https://github.com/narwhals-dev/narwhals/actions/runs/17896855969/job/50884585965?pr=3148

dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/17896855968/job/50884585831?pr=3148

For this, I'd suggest adding a `# pyright: ignore[reportArgumentType]`.

```py
  /home/runner/work/narwhals/narwhals/narwhals/_polars/dataframe.py:340:36 - error: Argument of type "Sequence[Mapping[str, Any]]" cannot be assigned to parameter "data" of type "Iterable[dict[str, Any]]" in function "from_dicts"
    "Sequence[Mapping[str, Any]]" is not assignable to "Iterable[dict[str, Any]]"
      Type parameter "_T_co@Iterable" is covariant, but "Mapping[str, Any]" is not a subtype of "dict[str, Any]"
        "Mapping[str, Any]" is not assignable to "dict[str, Any]" (reportArgumentType)
```

We could also open an issue upstream, summarizing the polars parts of (https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2366187248) - but that isn't blocking for this PR 👍

felixgwilliams:
Looking into the "old versions" failures, it looks like `polars` only supported Mappings and not just dicts starting `1.30.0`. (See [here](https://github.com/pola-rs/polars/pull/22638)), which is pretty recent. Before that, it wouldn't split non-dict mappings into columns. Mappings do work for pandas and pyarrow.

Should I

- Switch back from `Mapping` to `dict` in the type hints and don't try to suport it
- Write some code to convert `Mappings` to `dict` for the `polars` backend, perhaps giving a performance warning?
- something else?

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3148#issuecomment-3317804179

> Looking into the "old versions" failures, it looks like `polars` only supported Mappings and not just dicts starting `1.30.0`. (See [here](https://github.com/pola-rs/polars/pull/22638)), which is pretty recent. Before that, it wouldn't split non-dict mappings into columns. Mappings do work for pandas and pyarrow.

Thanks for tracking all this down @felixgwilliams 

Agreed, 4 months ago is pretty recent 🤔

I think - *worst-case* - after checking we're below `1.30.0` we do something like this and raise early:

```py
if data:
    first = data[0]
    if not isinstance(first, dict) and isinstance(first, Mapping):
        msg = "..."
```

> - Write some code to convert Mappings to dict for the polars backend, perhaps giving a performance warning?

Another option might be using a generator, where we only need to convert the values?

```py

from collections.abc import Iterator, Mapping, Sequence  # noqa: TC003
from typing import Any  # noqa: F811

import polars as pl


def from_mappings(data: Sequence[Mapping[str, Any]]) -> pl.DataFrame:
    columns = tuple(data[0])

    def gen() -> Iterator[tuple[Any, ...]]:
        for mapping in data:
            yield tuple(mapping.values())

    return pl.DataFrame(gen(), schema=columns, orient="row")


mappings = [{"c": 1, "d": 5}, {"c": 2, "d": 6}]
>>> from_mappings(mappings)
shape: (2, 2)
┌─────┬─────┐
│ c   ┆ d   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 5   │
│ 2   ┆ 6   │
└─────┴─────┘
```

felixgwilliams:
I had a go at handling the `Mapping` case for polars<1.30. It will convert mappings to dicts if

- polars < 1.30
- the first entry in the list of mappings is not a dict

I think this should solve the problem and get the checks passing again.

narwhals/_polars/dataframe.py
```diff
@@ -322,6 +323,31 @@ def from_dict(
...
+        if not data:
+            native = pl.DataFrame(schema=pl_schema)
+        elif FROM_DICTS_ACCEPTS_MAPPINGS or isinstance(data[0], dict):
+            native = pl.from_dicts(data, pl_schema)  # type: ignore[arg-type]
+        else:
+
+            def _to_dict(row: Mapping[str, Any]) -> dict[str, Any]:
+                return {k: v for k, v in row.items()}  # noqa: C416
+
+            native = pl.from_dicts([_to_dict(row) for row in data], pl_schema)
```
dangotbanned: https://github.com/narwhals-dev/narwhals/pull/3148#issuecomment-3323494886

> I had a go at handling the `Mapping` case for polars<1.30. It will convert mappings to dicts if
> 
> * polars < 1.30
> * the first entry in the list of mappings is not a dict

Did you see my second suggestion in (https://github.com/narwhals-dev/narwhals/pull/3148#issuecomment-3320845039)?

A more complete version would be:

```suggestion
        else:  # pragma: no cover
            columns = pl_schema or tuple(data[0])
            native = pl.DataFrame(
                (tuple(row.values()) for row in data), schema=columns, orient="row"
            )
```

That would end up down [this wildly optimized path](https://github.com/pola-rs/polars/blob/c02e965f3c17311d06ada6896a7a07f0f3d14daf/py-polars/polars/_utils/construction/dataframe.py#L972-L1062).

Whereas the current version *I'd expect* to be more expensive.
Partly for the reasoning in [incorrect-dict-iterator `(PERF102)`](https://docs.astral.sh/ruff/rules/incorrect-dict-iterator/#incorrect-dict-iterator-perf102). But also we're materializing everything into a `list` at once, rather than letting `polars` chunk it 🙂

felixgwilliams: Thanks for the suggestion. I didn't use it at first because PyLance flagged it as an error because generators are not supported here according to the Polars type hint.

On later versions of Polars, there may not be an error if you don't have torch installed, because one of the types in the union, `torch.Tensor` resolves to `Unknown`.

dangotbanned: > On later versions of Polars, there may not be an error if you don't have torch installed, because one of the types in the union, `torch.Tensor` resolves to `Unknown`.

Ahhh yeah that would be it, well spotted @felixgwilliams 😄 


image

narwhals/stable/v1/__init__.py
```diff
@@ -1273,19 +1273,7 @@ def from_dict(
...
-    backend: IntoBackend[EagerAllowed],
-) -> DataFrame[Any]:
-    """Instantiate DataFrame from a sequence of dictionaries representing rows.
-
-    See `narwhals.from_dicts` for full docstring. Note that `native_namespace` is
-    an is the same as `backend` but only accepts module types - for new code, we
-    recommend using `backend`, as that's available beyond just `narwhals.stable.v1`.
-    """
-    return DataFrame.from_dicts(data, schema, backend=backend)
+from_dicts: Final = DataFrame.from_dicts
```
dangotbanned: @felixgwilliams I was wondering what you thought about this?

We can use a single doc in `DataFrame.from_dicts`, and that gets reused by:
- `v1.DataFrame.from_dicts`
- `v2.DataFrame.from_dicts`
- `v1.from_dicts`
- `v2.from_dicts`

### Preview 1

image

### Preview 2

image

narwhals/dataframe.py
```diff
@@ -610,6 +610,59 @@ def from_dict(
...
+            >>> import narwhals as nw
+            >>> data = [{"c": 5, "d": 1}, {"c": 2, "d": 4}]
+            >>> nw.DataFrame.from_dicts(data, backend="pandas")
+            ┌──────────────────┐
+            |Narwhals DataFrame|
+            |------------------|
+            |        c  d      |
+            |     0  5  1      |
+            |     1  2  4      |
+            └──────────────────┘
```
dangotbanned: I've just pinned this issue that's has a mention *very deep* into [CONTRIBUTING](https://github.com/narwhals-dev/narwhals?tab=contributing-ov-file#8-writing-the-docstrings)

- #1943

What I'd like to see here vs the example in [`DataFrame.from_dict`](https://narwhals-dev.github.io/narwhals/api-reference/dataframe/#narwhals.dataframe.DataFrame.from_dict) is:
- Different data
  - although the shape *is* different, it is nice to see some variety
- A different backend
  - `"polars"` might be the only *other* one that displays nicely though


Besides that - get as creative as you like 😄

felixgwilliams: Thanks. I've clearly been using `from_dict` as a model for the new function. Happy to give the docstring a little bit more individuality!

felixgwilliams:
I noticed that when the schemas of the dicts is not consistent and schema is not specified, the behavior is different when using the PyArrow backend, because PyArrow only looks at the first row to establish the  ([see here](https://github.com/apache/arrow/blob/d50e7e200ca9ebc4c6a6619c147af4427a15a26e/python/pyarrow/table.pxi#L6425)), whereas pandas uses all the rows and polars seems to use the first 100 by default.

Do you think it's worth highlighting the differences in the docstring or do you think "If not specified, the schema will be inferred by the native library." covers it? I think it's surprising enough to be worth mentioning, but I'm conscious of not wanting to be too verbose.

narwhals/dataframe.py
```diff
@@ -610,6 +610,68 @@ def from_dict(
...
+
+                `backend` can be specified in various ways
+
+                - As `Implementation.<BACKEND>` with `BACKEND` being `PANDAS`, `PYARROW`,
+                    `POLARS`, `MODIN` or `CUDF`.
+                - As a string: `"pandas"`, `"pyarrow"`, `"polars"`, `"modin"` or `"cudf"`.
+                - Directly as a module `pandas`, `pyarrow`, `polars`, `modin` or `cudf`.
+
+        Examples:
+            >>> import pandas as pd
```
dangotbanned:
```suggestion
            >>> import polars as pl
```

narwhals/functions.py
```diff
@@ -331,6 +331,56 @@ def _from_dict_no_backend(
...
+
+            `backend` can be specified in various ways
+
+            - As `Implementation.<BACKEND>` with `BACKEND` being `PANDAS`, `PYARROW`,
+                `POLARS`, `MODIN` or `CUDF`.
+            - As a string: `"pandas"`, `"pyarrow"`, `"polars"`, `"modin"` or `"cudf"`.
+            - Directly as a module `pandas`, `pyarrow`, `polars`, `modin` or `cudf`.
+
+    Examples:
+        >>> import pandas as pd
```
dangotbanned:
```suggestion
        >>> import polars as pl
```

narwhals/dataframe.py
```diff
@@ -610,6 +610,68 @@ def from_dict(
...
+            |shape: (2, 3)             |
+            |┌───────┬────────┬───────┐|
+            |│ item  ┆ weight ┆ price │|
+            |│ ---   ┆ ---    ┆ ---   │|
+            |│ str   ┆ i64    ┆ f64   │|
+            |╞═══════╪════════╪═══════╡|
+            |│ apple ┆ 80     ┆ 0.6   │|
+            |│ egg   ┆ 55     ┆ 0.4   │|
+            |└───────┴────────┴───────┘|
+            └──────────────────────────┘
```
dangotbanned, line 668: Resolves (https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2376517543)


This is great thanks @felixgwilliams 

Man, that's a heavy egg 😂 
(don't change it!)

felixgwilliams: Is it?? I looked it up on the Tesco website

dangotbanned:
> I noticed that when the schemas of the dicts is not consistent and schema is not specified, the behavior is different
> - `pyarrow` only looks at the first row to establish the schema ([see here](https://github.com/apache/arrow/blob/d50e7e200ca9ebc4c6a6619c147af4427a15a26e/python/pyarrow/table.pxi#L6425))
> - `pandas` uses all the rows
> - `polars` uses the first 100 by default.

Well spotted! 

> Do you think it's worth highlighting the differences in the docstring or do you think 
> > "If not specified, the schema will be inferred by the native library." 
> 
> covers it?
> I think it's surprising enough to be worth mentioning, but I'm conscious of not wanting to be too verbose.

Agreed, this does seem like it would be helpful to document, considering:
- the JSON context (https://github.com/narwhals-dev/narwhals/issues/3145#issuecomment-3315003923) I framed this in
- missing keys are common enough to become a [typing](https://peps.python.org/pep-0655/) [feature](https://typing.python.org/en/latest/spec/typeddict.html#required-and-notrequired)

Someone will get burned by this eventually 😳 


## What to do?
### Docs
Although we *could* specify these differences in the docstring of `from_dicts`, I suspect we might **also** see the same kind of differences in:
- `DataFrame.from_dict`
- `Series.from_iterable`
- ... maybe others in #2116

*If* that's the case, then we could benefit from some [narrative docs](https://narwhals-dev.github.io/narwhals/concepts/null_handling/) in the user guide.
That way we could do side-by-side comparisons *without* clogging up the docstrings 😅

> [!NOTE]
> Might be best to split that out into a follow-up issue?

dangotbanned:
### Tests
> I noticed that when the schemas of the dicts is not consistent and schema is not specified, the behavior is different

Add tests for this then 😄 


Each should have `data` with a schema change in row:
1. 2 (xfail `pyarrow`)
2. 99 (xfail `pyarrow`) 
3. 101 (xfail `pyarrow`, `polars`)

I'd recommend using this fixture instead of `eager_backend`, but in this case it's because the input data size is larger than normal

https://github.com/narwhals-dev/narwhals/blob/63c5022e347cef3f821f725350cd9d39e6e476c6/tests/conftest.py#L320-L323

felixgwilliams:
> 
> Add tests for this then 😄 
> 

I'll write the tests this evening. Thanks for the pointers 👍.


As for the docstrings, are we happy with leaving them as is and covering the issue separately? Otherwise I wrote some bullet points explaining the differences that I haven't committed yet. I don't really have a good feeling for what is too long for a narwhals docstring.

narwhals/dataframe.py
dangotbanned: I stand corrected 😄 

These are pretty accurate for small apples/eggs

dangotbanned:
> Otherwise I wrote some bullet points explaining the differences that I haven't committed yet. I don't really have a good feeling for what is too long for a narwhals docstring.

I'm not 100% sure if you need more permissions for it, but I'd normally use a [suggestion (step 7)](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-comments-to-a-pull-request) for that kinda thing

I'm happy to take a look

narwhals/dataframe.py
```diff
@@ -610,6 +610,68 @@ def from_dict(
...
+        schema: IntoSchema | None = None,
+        *,
+        backend: IntoBackend[EagerAllowed],
+    ) -> DataFrame[Any]:
+        """Instantiate DataFrame from a sequence of dictionaries representing rows.
+
+        Notes:
+            For pandas-like dataframes, conversion to schema is applied after dataframe
+            creation.
+
```
felixgwilliams, line 626: Hi @dangotbanned. This is what I had written. What do you think?

```suggestion
        When a schema is not specified, the native libraries infer the schema differently:
        - pandas uses all rows
        - polars uses the first 100 rows
        - pyarrow uses only the first row
        If the rows in `data` have different keys, consider passing a schema for more consistent
        results

```

dangotbanned:
@felixgwilliams sorry for the delay, I'm hoping to review this later today 🙏

narwhals/dataframe.py
```diff
@@ -610,6 +610,68 @@ def from_dict(
...
+                specified, the schema will be inferred by the native library.
+            backend: Specifies which eager backend instantiate to.
+
+                `backend` can be specified in various ways
+
+                - As `Implementation.<BACKEND>` with `BACKEND` being `PANDAS`, `PYARROW`,
+                    `POLARS`, `MODIN` or `CUDF`.
+                - As a string: `"pandas"`, `"pyarrow"`, `"polars"`, `"modin"` or `"cudf"`.
+                - Directly as a module `pandas`, `pyarrow`, `polars`, `modin` or `cudf`.
+
```
dangotbanned, line 639: https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2379837835

@felixgwilliams pretty good, here's a few tweaks 🙂 

The content is pretty much the same, but re-worded to fit the idea of a [tip](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#+type:tip)

```suggestion

        Tip:
            If you expect non-uniform keys in `data`, consider passing `schema` for
            more consistent results, as **inference varies between backends**:

            - pandas uses all rows
            - polars uses the first 100 rows
            - pyarrow uses only the first row

```

### Previews
WDYT?

<details><summary><b>Show API Reference</b></summary>
<p>

image

</p>
</details> 


<details><summary><b>Show IDE Hover</b></summary>
<p>


image
</p>
</details>

narwhals/functions.py
```diff
@@ -331,6 +331,56 @@ def _from_dict_no_backend(
...
+            specified, the schema will be inferred by the native library.
+        backend: Specifies which eager backend instantiate to.
+
+            `backend` can be specified in various ways
+
+            - As `Implementation.<BACKEND>` with `BACKEND` being `PANDAS`, `PYARROW`,
+                `POLARS`, `MODIN` or `CUDF`.
+            - As a string: `"pandas"`, `"pyarrow"`, `"polars"`, `"modin"` or `"cudf"`.
+            - Directly as a module `pandas`, `pyarrow`, `polars`, `modin` or `cudf`.
+
```
dangotbanned, line 358: Whatever we land on for (https://github.com/narwhals-dev/narwhals/pull/3148#discussion_r2379837835) will need to go here as well

tests/frame/from_dicts_test.py
```diff
@@ -0,0 +1,75 @@
...
+    # pyarrow only checks 1 row
+    if "pyarrow" in str(eager_implementation) and n_first_schema >= 1:
+        request.applymarker(pytest.mark.xfail)
+    # polars checks 100 rows
+    if "polars" in str(eager_implementation) and n_first_schema >= 100:
+        request.applymarker(pytest.mark.xfail)
+    # no xfail for pandas as it always scans all rows
+
+    # first n rows have columns ["a"], the next row has columns ["a", "b"]
+    data = [*({"a": i} for i in range(n_first_schema)), {"a": n_first_schema, "b": 0}]
```
dangotbanned: I prefer trying to make comments *part of* the code when possible

```suggestion
    incomplete = ({"a": i} for i in range(n_first_schema))
    complete = {"a": n_first_schema, "b": 0}
    data = (*incomplete, complete)
```

Probably doesn't matter much here, but generally I think it helps avoid having the comment become out-of-sync with the code it describes 🙂

tests/from_dicts_test.py
```diff
@@ -0,0 +1,75 @@
...
+    # pyarrow only checks 1 row
+    if "pyarrow" in str(eager_implementation) and n_first_schema >= 1:
+        request.applymarker(pytest.mark.xfail)
+    # polars checks 100 rows
+    if "polars" in str(eager_implementation) and n_first_schema >= 100:
+        request.applymarker(pytest.mark.xfail)
+    # no xfail for pandas as it always scans all rows
+
+    # first n rows have columns ["a"], the next row has columns ["a", "b"]
+    data = [*({"a": i} for i in range(n_first_schema)), {"a": n_first_schema, "b": 0}]
```
dangotbanned: Same as ^^^

```suggestion
    incomplete = ({"a": i} for i in range(n_first_schema))
    complete = {"a": n_first_schema, "b": 0}
    data = (*incomplete, complete)
```

dangotbanned (commented):
Thanks for the PR @felixgwilliams

I've only got a few *minor* suggestions - I think we're pretty much good to go

praise be narwhal

narwhals/functions.py
felixgwilliams:
```suggestion

        Tip:
            If you expect non-uniform keys in `data`, consider passing `schema` for
            more consistent results, as **inference varies between backends**:
            - pandas uses all rows
            - polars uses the first 100 rows
            - pyarrow uses only the first row
```
copied from comment above

felixgwilliams: Hmm looks I can't use the suggestion feature to make all the changes at once.
I'll just add this in a followup commit

felixgwilliams:
Thanks @dangotbanned for your helpful suggestions. I'll be sure to remember the tip about comments being part of the code.

PR #3152, opened 2025-09-25
feat: Support `over` expressions more freely, make expressions printable, rewrite internals (travelling pr 🌴 )

MarcoGorelli:
I've been on holiday I had a bit of time while travelling recently, so I tried rewriting the internals (yes, again)

The general idea is that `narwhals.Expr` stores a list of expressions in `._nodes`, rather than passing around a bunch of opaque lambdas

#### Expressions are pretty-printable

For example:
```python
In [2]: nw.col('a').abs().rank()
Out[2]: col(a).abs().rank(method=average, descending=False)
```
This is still quite basic, but we could make it more complex by introducing line breaks if it gets too long. Seems like the kind of thing @camriddell might be interested in?

#### We can do simple expression rewrites

Currently, `(nw.col('a').mean() + 1).over('b')` isn't supported:
- for pandas-like, it's a non-elementary operation in `over`
- for sql, `(mean(a) + 1) over (partition by b)` isn't valid syntax, it should be `mean(a) over (partition by b) + 1`

With this PR, however, it is!

What we do here is, when inserting an `over` node, we push it down before any elementwise operations (such as `+`, `.abs()`, `sum_horizontal`, ...) and apply it to all expressions. There's some more details in the expansion to "how it works"

So now, expressions like `(nw.col('a').mean() + 1).over('b')` can be supported for all backends

This rewrite is extremely simple and cheap, it's just a matter of inserting a node at some position `i` rather than at the end of a list. In general, query optimisation is out of scope for Narwhals. But, given that this enables more of Polars' flexibility for other backends, I think this can be in scope.

#### Per-group broadcasting

Previously, `nw.col('a') - nw.col('a').mean()` would be fine, but `(nw.col('a') - nw.col('a').mean()).over('b')` would raise for sql-like backends. Now, it works fine across all backends! Really useful for feature engineering

#### Simplified internals

- We can completely get rid of `depth`, `function_name`, `scalar_kwargs`
- Replace `CompliantWhen` / `CompliantThen` and their complicated interaction with just `CompliantNamespace.when_then`

In fact, this goes as far as _reducing_ package size by almost 1%. Not that that was the objective with this work, but it's nice to see that it doesn't make the package bigger

### What this may open the doors to

- serialisation / deserialisation of expressions. e.g. `nw.Expr.from_json(expr.to_json())`
- chained window functions, like `nw.col('a').shift(7).rolling_mean(7).over('store', order_by='date')`
- non-elementary group-by aggregations for pandas/pyarrow, like `df.group_by('a').agg((nw.col('b')-nw.col('c')).mean())`

### Benchmarks

- https://www.kaggle.com/code/marcogorelli/plotly-timings?scriptVersionId=268524779
- https://www.kaggle.com/code/marcogorelli/narwhals-vs-ibis-overhead?scriptVersionId=268533320

I'm not seeing any meaningful impact on performance

### DIFF
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -82,8 +82,8 @@ class Expr:
             return self.__class__(*new_nodes)
         if node.kwargs["order_by"] and any(node.is_orderable() for node in new_nodes[:i]):
             new_nodes.insert(i, node)
-        elif node.kwargs["partition_by"] and not all(
-            node.is_elementwise() for node in new_nodes[:i]
+        elif node.kwargs["partition_by"] and any(
+            not node.is_elementwise() for node in new_nodes[:i]
         ):
             new_nodes.insert(i, node_without_order_by)
         elif all(node.is_elementwise() for node in new_nodes):


narwhals/_expression_parsing.py
```diff
@@ -202,6 +206,111 @@ def __and__(self, other: ExpansionKind) -> Literal[ExpansionKind.MULTI_UNNAMED]:
...
+
+    def __init__(
+        self,
+        kind: ExprKind,
+        name: str,
+        /,
+        *exprs: IntoExpr | NonNestedLiteral,
+        str_as_lit: bool = False,
+        allow_multi_output: bool = False,
+        **kwargs: Any,
```
camriddell, line 224: Why split out kwargs and exprs? Does this presume that kwargs should not have expressions in it and all expressions will be passed as positional arguments?

narwhals/_expression_parsing.py
```diff
@@ -202,6 +206,111 @@ def __and__(self, other: ExpansionKind) -> Literal[ExpansionKind.MULTI_UNNAMED]:
...
+                expr_node.is_elementwise() for expr_node in expr._nodes
+            ):
+                exprs.append(expr._with_node(over_node_without_order_by))
+            else:
+                # If there's no `partition_by`, then `over_node_without_order_by` is a no-op.
+                exprs.append(expr)
+        self.exprs = exprs
+
+    def is_orderable(self) -> bool:
+        if self._is_orderable_cached is None:
```
camriddell, line 291: Could we use [functools.cached_property](https://docs.python.org/3/library/functools.html#functools.cached_property) instead?

narwhals/_expression_parsing.py
```diff
@@ -376,20 +580,20 @@ def with_ordered_over(self) -> ExprMetadata:
                 "  + `nw.col('price').diff().over(order_by='date') + 1`\n"
             )
             raise InvalidOperationError(msg)
-        if self.last_node.is_orderable_window:
+        if next(self.op_nodes_reversed()).kind.is_orderable and n_orderable_ops > 0:
```
camriddell, line 600: Is there a risk of a `StopIteration`? or is `self.op_nodes_reversed()` guaranteed to produce at least one element based on how this method is called?

narwhals/_expression_parsing.py
```diff
@@ -202,6 +206,111 @@ def __and__(self, other: ExpansionKind) -> Literal[ExpansionKind.MULTI_UNNAMED]:
...
+        str_as_lit: Whether to interpret strings as literals when they
+            are present in `exprs`.
+        allow_multi_output: Whether to allow any of `exprs` to be multi-output.
+        kwargs: Other (non-expressifiable) arguments to function.
+    """
+
+    def __init__(
+        self,
+        kind: ExprKind,
+        name: str,
```
camriddell, line 219: `name` can be any string, but at runtime it is any string that represents a valid dotted path on a given `Namespace` for root nodes and a valid dotted path on a given `Expr` for subsequent nodes.

I wonder if there is a way to more formally express this relationship aside from storing strings with apriori knowledge about how they will be parsed?

camriddell (commented):
Wow, this is incredibly impressive! I made my way through the high level implementation (nothing backend specific) and have a few thoughts. There are a lot of moving pieces here, so I'm going to continue thinking on this to provide more feedback.

Aside form some of the specific code comments, I do have one design concern over the use of strings to intermediate `ExprNode`s to the actual backend call stack. This is discussed in one of the sub-comments with no solution answer yet.

narwhals/_expression_parsing.py
MarcoGorelli: Yup that's generally the idea!

narwhals/_expression_parsing.py
MarcoGorelli: i think we tried that in the past and ran into issues on some python versions. it's simple enough to implement ourselves though

narwhals/_expression_parsing.py
MarcoGorelli: all expressions must have at least one node, so this isn't a risk

narwhals/_expression_parsing.py
MarcoGorelli: Yup, that's a very good point!

Every expression is covered by tests (and we enforce 100% coverage here), so if we had an invalid string here, it would surely get picked up by the test suite

docs/how_it_works.md
```diff
@@ -238,49 +241,98 @@ print(nw.to_native(df.select(nw.col("a") + 1)))
...
+```python exec="1" result="python" session="pandas_impl" source="above"
+print(expr._nodes[2].as_dict())
+```
+
+This tells us a few things:
+
+- We're performing an aggregation.
+- The name of the function is `'std'`. This will be looked up in the compliant object.
+- It takes keyword arguments `ddof=1`.
+- We'll look at the others later.
```
FBruzzesi: Do we need this point? Above we state: "Let's start with the third one: [...] This tells us a few things:"

docs/how_it_works.md
```diff
@@ -238,49 +241,98 @@ print(nw.to_native(df.select(nw.col("a") + 1)))
...
+print(expr._nodes[3].as_dict())
+```
+
+Note how now, the `exprs` attribute is populated. Indeed, we are summing another expression: `col('b')`.
+The `exprs` parameter holds arguments which are either expressions, or should be interpreted as expressions.
+The `str_as_lit` parameter tells us whether string literals should be interpreted as literals (e.g. `lit('foo')`)
+or columns (e.g. `col('foo')`). Finally `allow_multi_output` tells us whether multi-outuput expressions
+(more on this in the next section) are allowed to appear in `exprs`.
+
+Node that the expression in `exprs` also has its own nodes:
```
FBruzzesi: Too many nodes 😂

```suggestion
Note that the expression in `exprs` also has its own nodes:
```

docs/how_it_works.md
```diff
@@ -377,3 +428,67 @@ Narwhals triggers a broadcast in these situations:
...
+
+```python
+nw.col("a").sum().over("c") + nw.col("b").sum().over("c")
+```
+
+In general, query optimisation is out-of-scope for Narwhals. We consider this
+expression rewrite acceptable because:
+
+- It's simple.
+- It allows us to evaluate operations which otherwise wouldn't be allowed for certain backends.
```
FBruzzesi: Can we wrap this into a big fat mkdocs material admonition?

narwhals/expr.py
```diff
@@ -39,74 +40,62 @@
...
-                *args,
-                str_as_lit=False,
-                allow_multi_output=False,
-                to_single_output=False,
-            ),
-        )
+    def __init__(self, *nodes: ExprNode) -> None:
+        self._nodes = nodes
+
+    def __call__(self, ns: CompliantNamespace[Any, Any]) -> CompliantExpr[Any, Any]:
```
dangotbanned: I'm not so sure `Expr.__call__` is a good idea.

All of the other `__dunders__` (excluding `__bool__`) are used for public API and return `Expr`.

But this is very different, and leads to a confusing error

<details><summary>Show error</summary>
<p>

image

</p>
</details> 



I'm also not getting any results for 

<details><summary>Show <b>Find All References</b></summary>
<p>

image

</p>
</details> 



That could be worked around if I had a name to search for 😉

narwhals/expr.py
```diff
@@ -39,74 +40,62 @@
...
+        for node in nodes[1:]:
+            ce = evaluate_node(ce, node, ns)
+        return ce
+
+    def _with_node(self, node: ExprNode) -> Self:
+        if node.kind is ExprKind.OVER:
+            # insert `over` before any elementwise operations.
+            # check "how it works" page in docs for why we do this.
+            new_nodes = list(self._nodes)
+            kwargs_no_order_by = {
```
dangotbanned: I think this should be a separate method.

The caller of `_with_node` has `node: ExprNode` so presumably, they would either:
1. *Already* know that the kind is `ExprKind.OVER` (e.g. based on the method we're in)
2. Be able to *first* check `node.kind` and then call the right method?

tests/expression_parsing_test.py
```diff
@@ -4,96 +4,114 @@
...
-
-
-def test_rank_aggregation() -> None:
-    with pytest.raises(InvalidOperationError):
-        nw.col("a").mean().rank()
-    with pytest.raises(InvalidOperationError):
-        nw.col("a").mean().is_unique()
+def test_over_pushdown(
+    constructor: Constructor, expr: nw.Expr, expected: list[float]
+) -> None:
```
dangotbanned, line 56: These were the cases I wrote for #2571.

<details><summary>Show <code>tests.plan.expr_rewrites_test.py#L49-L114</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/tests/plan/expr_rewrites_test.py#L49-L114

</p>
</details> 

From looking at the changes in `nw.Expr`, I think adding an `.alias()` into the mix may cause trouble over here?

### `test_rewrite_elementwise_over_simple`
```py
# Before
col('a').sum().abs().over([col('b')])
# After
a=col('a').sum().over([col('b')]).abs()
```



### `test_rewrite_elementwise_over_multiple`
```py
# Before
col(['b', 'c']).last().replace_strict().over([col('d')])

# After
b=col('b').last().over([col('d')]).replace_strict()
c=col('c').last().over([col('d')]).replace_strict()
```

### `test_rewrite_elementwise_over_complex`
This one mixes in expressions that are skipped - but safe to call the rewrite on:
```py
# Before
## 0
col('a')
## 1
col('b').cast(String)
## 2
col('c').max().alias('x').fill_null([lit(50)]).over([col('a')]).alias('x2')
## 3
col('d').is_duplicated().alias('d*').alias('d**').over([col('b')]).not_()
## 4
cs.string().str.contains('some').name.suffix('_some')
## 5
nth((3, 4, 1)).null_count().sqrt().over([col('f'), col('g'), col('j')]).name.to_uppercase()

# After
## 0
a=col('a')
## 1
b=col('b').cast(String)
## 2
x2=col('c').max().over([col('a')]).fill_null([lit(50)])
## 3
d**=col('d').is_duplicated().over([col('b')]).not_()
## 4
f_some=col('f').str.contains('some')
g_some=col('g').str.contains('some')
h_some=col('h').str.contains('some')
## 5
D=col('d').null_count().over([col('f'), col('g'), col('j')]).sqrt()
E=col('e').null_count().over([col('f'), col('g'), col('j')]).sqrt()
B=col('b').null_count().over([col('f'), col('g'), col('j')]).sqrt()
```

MarcoGorelli: thank you!

could you elaborate on the alia please? i just tried this out and it looks ok
```python
In [1]: import narwhals as nw

In [2]: import duckdb

In [3]: rel = duckdb.sql("select * from values (1,4,0),(1,5,1),(2,6,2) df(a,b,i)")

In [4]: nw.from_native(rel).with_columns(nw.col('b').alias('c').sum().over('a')).sort('i')
Out[4]: 
┌──────────────────────────────────┐
|        Narwhals LazyFrame        |
|----------------------------------|
|┌───────┬───────┬───────┬────────┐|
|│   a   │   b   │   i   │   c    │|
|│ int32 │ int32 │ int32 │ int128 │|
|├───────┼───────┼───────┼────────┤|
|│     1 │     4 │     0 │      9 │|
|│     1 │     5 │     1 │      9 │|
|│     2 │     6 │     2 │      6 │|
|└───────┴───────┴───────┴────────┘|
└──────────────────────────────────┘

In [5]: nw.from_native(rel).with_columns(nw.col('b').sum().alias('c').over('a')).sort('i')
Out[5]: 
┌──────────────────────────────────┐
|        Narwhals LazyFrame        |
|----------------------------------|
|┌───────┬───────┬───────┬────────┐|
|│   a   │   b   │   i   │   c    │|
|│ int32 │ int32 │ int32 │ int128 │|
|├───────┼───────┼───────┼────────┤|
|│     1 │     4 │     0 │      9 │|
|│     1 │     5 │     1 │      9 │|
|│     2 │     6 │     2 │      6 │|
|└───────┴───────┴───────┴────────┘|
└──────────────────────────────────┘

In [6]: nw.from_native(rel).with_columns(nw.col('b').sum().over('a').alias('c')).sort('i')
Out[6]: 
┌──────────────────────────────────┐
|        Narwhals LazyFrame        |
|----------------------------------|
|┌───────┬───────┬───────┬────────┐|
|│   a   │   b   │   i   │   c    │|
|│ int32 │ int32 │ int32 │ int128 │|
|├───────┼───────┼───────┼────────┤|
|│     1 │     4 │     0 │      9 │|
|│     1 │     5 │     1 │      9 │|
|│     2 │     6 │     2 │      6 │|
|└───────┴───────┴───────┴────────┘|
└──────────────────────────────────┘
```
will try some other examples anyway

dangotbanned: I was thinking there would be a conflict between 

https://github.com/narwhals-dev/narwhals/blob/d0b3acde95438682c6d07086fd596737700d988e/narwhals/expr.py#L60-L61

And 

https://github.com/narwhals-dev/narwhals/blob/d0b3acde95438682c6d07086fd596737700d988e/narwhals/expr.py#L119


https://github.com/narwhals-dev/narwhals/blob/d0b3acde95438682c6d07086fd596737700d988e/narwhals/expr.py#L139


An important part in some of the examples for (https://github.com/narwhals-dev/narwhals/pull/3152#discussion_r2417997015), is that `alias` and `name` ops can be used freely - and never impact the ability to rewrite

MarcoGorelli: yup, agreed, given that they're elementwise

MarcoGorelli: yup, agreed, given that they're elementwise

docs/how_it_works.md
```diff
@@ -177,7 +178,7 @@ The way you access the Narwhals-compliant wrapper depends on the object:
 
 - `narwhals.DataFrame` and `narwhals.LazyFrame`: use the `._compliant_frame` attribute.
 - `narwhals.Series`: use the `._compliant_series` attribute.
-- `narwhals.Expr`: call the `._to_compliant_expr` method, and pass to it the Narwhals-compliant namespace associated with
+- `narwhals.Expr`: call the `.__call__` method, and pass to it the Narwhals-compliant namespace associated with
```
FBruzzesi: possible "leftover" from https://github.com/narwhals-dev/narwhals/pull/3152/commits/6e9eab463d51b765bc030e465501f0e1e7ecabc9?

narwhals/_expression_parsing.py
```diff
@@ -263,71 +376,158 @@ def __repr__(self) -> str:  # pragma: no cover
...
 
+    @classmethod
+    def from_node(  # noqa: PLR0911
+        cls, node: ExprNode, *ces: CompliantExprAny
+    ) -> ExprMetadata:
+        if node.kind is ExprKind.SERIES:
+            return cls.from_selector_single(node)
+        if node.kind is ExprKind.COL:
+            return (
+                ExprMetadata.from_selector_single(node)
```
FBruzzesi: Should we use `cls` in place of `ExprMetadata` (everywhere in this method)? In practice it probably doesn't matter too much (unless someone is crazy enough to subclass this 😂)

narwhals/_expression_parsing.py
```diff
@@ -401,110 +605,63 @@ def with_partitioned_over(self) -> ExprMetadata:
...
             is_elementwise=False,
             is_scalar_like=False,
             is_literal=False,
+            nodes=(*self.nodes, node),
         )
 
-    def with_orderable_filtration(self) -> ExprMetadata:
-        if self.is_scalar_like:
+    def with_orderable_filtration(self, node: ExprNode) -> ExprMetadata:
+        if self.is_scalar_like:  # pragma: no cover
```
FBruzzesi: `# pragma: no cover` 🤔 how's this not covered now?

narwhals/_expression_parsing.py
```diff
@@ -202,6 +200,122 @@ def __and__(self, other: ExpansionKind) -> Literal[ExpansionKind.MULTI_UNNAMED]:
...
+        # Note: please keep this as a for-loop (rather than a list-comprehension)
+        # so that pytest-cov highlights any uncovered branches.
+        for expr in self.exprs:
+            if not is_expr(expr):
+                exprs.append(expr)
+            elif over_node.kwargs["order_by"] and any(
+                expr_node.is_orderable() for expr_node in expr._nodes
+            ):
+                exprs.append(expr._with_node(over_node))
+            elif over_node_without_order_by.kwargs["partition_by"] and not all(
```
FBruzzesi: These two can be evaluated outside the loop as they are independent of `expr` itself:

- `over_node.kwargs["order_by"]`
- `over_node_without_order_by.kwargs["partition_by"]`

---

Also just by focusing in here, it gives for granted that these keys always exist (and can be None/empty sequences) - is that the right assumption?

narwhals/_expression_parsing.py
```diff
@@ -263,71 +376,158 @@ def __repr__(self) -> str:  # pragma: no cover
...
+            return self.with_orderable_aggregation(node)
+        if node.kind is ExprKind.WINDOW:
+            return self.with_window(node)
+        if node.kind is ExprKind.OVER:
+            if node.kwargs["order_by"]:
+                return self.with_ordered_over(node)
+            if not node.kwargs["partition_by"]:  # pragma: no cover
+                msg = "At least one of `partition_by` or `order_by` must be specified."
+                raise InvalidOperationError(msg)
+            return self.with_partitioned_over(node)
```
FBruzzesi: Getting OCD-ed here:

```suggestion
            if node.kwargs["order_by"]:
                return self.with_ordered_over(node)
            if node.kwargs["partition_by"]:
                return self.with_partitioned_over(node)
            msg = "At least one of `partition_by` or `order_by` must be specified."    # pragma: no cover
            raise InvalidOperationError(msg)    # pragma: no cover
            
```

narwhals/expr.py
```diff
@@ -39,74 +40,64 @@
...
+            return self.__class__(*new_nodes)
+        return self.__class__(*self._nodes, node)
 
     def __repr__(self) -> str:
-        return f"Narwhals Expr\nmetadata: {self._metadata}\n"
+        """Pretty-print the expression by combining all nodes in the metadata."""
+        result: str = repr(self._nodes[0])
+        for node in self._nodes[1:]:
+            result = f"{result}.{node!r}"
+        return result
```
FBruzzesi:
```suggestion
        return ".".join(repr(node) for node in self._nodes)
```

narwhals/_expression_parsing.py
```diff
@@ -263,71 +376,158 @@ def __repr__(self) -> str:  # pragma: no cover
...
+    @classmethod
+    def from_elementwise(cls, node: ExprNode, *ces: CompliantExprAny) -> ExprMetadata:
+        return combine_metadata(*ces, to_single_output=True, nodes=(node,))
+
     @property
     def is_filtration(self) -> bool:
         return not self.preserves_length and not self.is_scalar_like
 
-    def with_aggregation(self) -> ExprMetadata:
+    def with_aggregation(self, node: ExprNode) -> ExprMetadata:
```
FBruzzesi: Is it worth making a private `_update` method along the lines of:

```py
def _update(self, node: ExprNode, **overrides: Any) -> ExprMetadata:
    """Creates a new ExprMetadata instance with updated values."""
    # Start with the current metadata attributes
    current = {
        "expansion_kind": self.expansion_kind,
        "has_windows": self.has_windows,
        "n_orderable_ops": self.n_orderable_ops,
        "preserves_length": self.preserves_length,
        "is_elementwise": self.is_elementwise,
        "is_scalar_like": self.is_scalar_like,
        "is_literal": self.is_literal,
        "nodes": (*self.nodes, node),
    }
    # Apply the specific changes for this node
    return ExprMetadata(current | overrides)
```

in order to simplify other `with_<op>` which are then called in `with_node`?

narwhals/_expression_parsing.py
dangotbanned: > In practice it probably doesn't matter too much (unless someone is crazy enough to subclass this 😂)

They won't get very far if they try 😉 

https://github.com/narwhals-dev/narwhals/blob/ebb2a40bab2f18b485512707ffb8cd440bb17172/narwhals/_expression_parsing.py#L258-L260

> Should we use `cls` in place of `ExprMetadata` (everywhere in this method)?

We can do if we're in a `@classmethod` and want to shrink things.
Alternatively, we could make everything a `@staticmethod` and always use `ExprMetadata`.

I think using `cls` is probably more readable, and the other option is better if we want to be as strict as possible that you cannot extend it 🙂

tests/expression_parsing_test.py
MarcoGorelli: yup, `alias` and `name` are elementwise, so they don't impact the ability to rewrite

narwhals/_expression_parsing.py
MarcoGorelli: thanks but tbh i prefer the early return pattern

narwhals/_expression_parsing.py
dangotbanned: @FBruzzesi what a great idea ... 😉

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/_immutable.py#L84-L94

MarcoGorelli:
thanks for your review!

> I am assuming that the relevant changes are in _expression_parsing.py and expr.py, and everything else is a byproduct of such.

yup!

FBruzzesi:
@MarcoGorelli so I added a PR to target your branch https://github.com/MarcoGorelli/narwhals/pull/7. To keep the discussion here, quoting from the PR description there:

> I have an idea I want to try out - hopefully I can get there soon

_The main purpose is to avoid having to create a new tuple for each operation (nodes=(*self.nodes, node) which compounds back for each node) while we can simply use a reference_

narwhals/_expression_parsing.py
MarcoGorelli: I think this seems like a good suggestion, thanks!

However, it might be orthogonal to this PR? As in, it can already be done, independently of what's being proposed here, right? If so, fancy opening a PR to do so? (or I can try later)

narwhals/expr.py
MarcoGorelli: Not sure tbh

- the caller just wants to insert a node, and so calls `self._with_node(node)`
- it's then `_with_node`'s responsibility where and how `node` gets inserted

dangotbanned: Could you link some code with an example of the two caller cases?

> * it's then `_with_node`'s responsibility where and how `node` gets inserted

I agree with this on *responsibility*.
But I guess what I'm trying to question is if the caller already knows the answer to the condition - then wouldn't it be cheaper to not ask the question for every call?

E.g. `Expr.alias` will **always** evaluate to false here, so why do we check?

I *suspect* the case for `node.kind is ExprKind.OVER == True` is *less common* - which if that is true - I think it shouldn't be a part of every method call and warrants it's own path 🙂

narwhals/_compliant/expr.py
```diff
@@ -90,7 +89,14 @@ class CompliantExpr(
...
     _alias_output_names: AliasNames | None
-    _metadata: ExprMetadata | None
+    # This should be set with extreme care, only in `_expression_parsing.py`,
+    # and never from within any compliant class.
+    _opt_metadata: ExprMetadata | None = None
+
+    @property
+    def _metadata(self) -> ExprMetadata:
+        assert self._opt_metadata is not None  # noqa: S101
+        return self._opt_metadata
```
dangotbanned, line 107: Making this a property that always returns `ExprMetadata` is a step in the right direction 👍 

Have you thought about formalising this invariant *more* than just using `None` and `assert`?
For example, if we could write something like:

```py
ExprMetadata.UNSET  # patent-pending
```

Then this sentinel could raise an informative error whenever it is used.

IIRC, you could run into that assertion if you try to construct certain kinds of expressions at the compliant-level, since they don't have the same checking/propagation as the narwhals-level.

For the benefit of *typing* the attribute/property would always return `ExprMetadata` and never `None`.
But in the case where *we* have written something incorrectly, at runtime we would get an error saying:

```
Hey dude, clippy here, looks like you forgot to do the thing in this way?
Maybe try this instead? 😉 
```

MarcoGorelli: sure, i've added a develop-facing error message, thanks (I think I prefer `None` to `UNSET` though)

narwhals/expr.py
MarcoGorelli: sure, have split into `_append_node` and `_with_over_node`

MarcoGorelli:
Unless there's objections, I'll do a careful read-though of everything and then will plan to ship this by the end of the week

PR #3143, opened 2025-09-18
feat(expr-ir): Support `group_by`, utilize `pyarrow.acero`

dangotbanned:
## What type of PR is this? (check all applicable)
- [x] ✨ Feature

## Related issues
- Child of ##2572

## Description
This one spiralled into something pretty interesting IMO 😏 

[Acero](https://arrow.apache.org/docs/cpp/acero/user_guide.html) is used under-the-hood for `pyarrow` features like:
- `Table.group_by` (most important to this PR)
  - See https://github.com/narwhals-dev/narwhals/blob/e0a3684ba4f77a59fe45d2914c74b4cff25cf344/narwhals/_plan/arrow/group_by.py)
- `Table.filter`
- `Table.{join,join_asof}`

But it also has *limited support* for other operations like:
- [`project`](https://arrow.apache.org/docs/cpp/acero/user_guide.html#project)
- [`aggregate`](https://arrow.apache.org/docs/cpp/acero/user_guide.html#aggregate)
- [`order_by`](https://arrow.apache.org/docs/python/generated/pyarrow.acero.OrderByNodeOptions.html)

I've been experimenting with those a bit in (https://github.com/narwhals-dev/narwhals/blob/e0a3684ba4f77a59fe45d2914c74b4cff25cf344/narwhals/_plan/arrow/acero.py).
Anticipating that combining those nodes into one big soup will result in `over(...)` 😄

## Tasks
- [x] [test: Port over `tests/frame/group_by_test`](https://github.com/narwhals-dev/narwhals/pull/3143/commits/1aa2464d1178950d8792f9675d30dcc40e87cccc)
- [x] ([fix: Avoid shadowed output aggregation names](https://github.com/narwhals-dev/narwhals/pull/3143/commits/ce18f518ac21d5484d5014282047125690960fc9))
  - [x] `test_double_same_aggregation`
  - [x] `test_group_by_len_1_column`
- [x] [feat(expr-ir): Rewrite, fix ordered aggregations](https://github.com/narwhals-dev/narwhals/pull/3143/commits/16148b2cb8f1e7e98d3d7287cccbcc5c2debcff1)
  - Forced a failure [here](https://github.com/narwhals-dev/narwhals/actions/runs/17879971676/job/50846380750?pr=3143)
- [x] [feat(expr-ir): Support `group_by(drop_null_keys=True)`](https://github.com/narwhals-dev/narwhals/pull/3143/commits/4b77500764ca1a872d5852cf31521bfe9a302eb9)
[...]

PR #3173, opened 2025-10-02
feat(expr-ir): Acero `order_by`, `hashjoin` , `DataFrame.{filter,join}`, `Expr.is_{first,last}_distinct`

dangotbanned:
## Tracking
- [x] #3182

## Related issues

- Child of #2572
- Follow-up to #3143

Building out more `acero` parts to be able to support `.over(*partition_by)`

### DIFF
--- a/tests/plan/frame_filter_test.py
+++ b/tests/plan/frame_filter_test.py
@@ -8,7 +8,7 @@ import pytest
 pytest.importorskip("pyarrow")
 
 import narwhals._plan as nwp
-from narwhals.exceptions import ColumnNotFoundError, InvalidOperationError, ShapeError
+from narwhals.exceptions import ColumnNotFoundError, ShapeError
 from tests.plan.utils import assert_equal_data, dataframe, series
 
 if TYPE_CHECKING:
@@ -42,11 +42,27 @@ def test_filter_single(
     assert_equal_data(result, expected)
 
 
-@pytest.mark.xfail(reason=("Not sure why this isn't allowed on `main`"))
-def test_filter_raise_on_agg_predicate(data: Data) -> None:
+def test_filter_aggregated_predicate(data: Data) -> None:
+    # NOTE: Unclear why this isn't permitted on `main`
+    pytest.importorskip("polars")
+    import polars as pl
+
+    expected = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8.0, 9.0]}
+    expected_invert: Data = {"a": [], "b": [], "z": []}
     df = dataframe(data)
-    with pytest.raises(InvalidOperationError):
-        df.filter(nwp.col("a").max() > 2)
+    df_pl = pl.DataFrame(data)
+    predicate = nwp.col("a").max() > 2
+    predicate_pl = pl.col("a").max() > 2
+
+    result = df.filter(predicate)
+    result_pl = df_pl.filter(predicate_pl).to_dict(as_series=False)
+    assert_equal_data(result, expected)
+    assert_equal_data(result, result_pl)
+
+    result = df.filter(~predicate)
+    result_pl = df_pl.filter(~predicate_pl).to_dict(as_series=False)
+    assert_equal_data(result, expected_invert)
+    assert_equal_data(result, result_pl)
 
 
 def test_filter_raise_on_shape_mismatch(data: Data) -> None:


narwhals/_plan/common.py
```diff
@@ -109,9 +111,21 @@ def into_dtype(dtype: DTypeT | type[NonNestedDTypeT], /) -> DTypeT | NonNestedDT
     return dtype
 
 
-# TODO @dangotbanned: Review again and try to work around (https://github.com/microsoft/pyright/issues/10673#issuecomment-3033789021)
+# NOTE: See (https://github.com/microsoft/pyright/issues/10673#issuecomment-3033789021)
 # The issue is `T` possibly being `Iterable`
 # Ignoring here still leaks the issue to the caller, where you need to annotate the base case
+@overload
+def flatten_hash_safe(iterable: Iterable[OneOrIterable[str]], /) -> Iterator[str]: ...
```
dangotbanned, line 119: It's an improvement over the previous version, but far from ideal.

Still doesn't resolve this case, and I'm not entirely sure why *yet*

https://github.com/narwhals-dev/narwhals/blob/f77bb4c348b3bb38636a373e4906b1b801971b91/narwhals/_plan/compliant/column.py#L49-L60

narwhals/_plan/arrow/acero.py
```diff
@@ -189,10 +191,15 @@ def _order_by(
...
-    raise NotImplementedError(msg)
+def sort_by(
+    by: OneOrIterable[str],
+    *more_by: str,
+    descending: OneOrIterable[bool] = False,
+    nulls_last: bool = False,
+) -> Decl:
+    return SortMultipleOptions.parse(
+        descending=descending, nulls_last=nulls_last
+    ).to_arrow_acero(tuple(flatten_hash_safe((by, more_by))))
```
dangotbanned, line 269: As of [feat(expr-ir): Impl `acero.sort_by`](https://github.com/narwhals-dev/narwhals/pull/3173/commits/f77bb4c348b3bb38636a373e4906b1b801971b91), I still need to make use of this in a plan.

A good candidate might be in either/both of

## `over(order_by=...)` 

https://github.com/narwhals-dev/narwhals/blob/f77bb4c348b3bb38636a373e4906b1b801971b91/narwhals/_plan/arrow/expr.py#L328-L350

### `is_{first,last}_distinct`
- [test: Port over `is_first_distinct` tests](https://github.com/narwhals-dev/narwhals/pull/3173/commits/36ddce064e55cd7b4bc87eb27c538c29b6f33712)

https://github.com/narwhals-dev/narwhals/blob/715be2219b337e2a4868787b558e5b2f9df11a97/narwhals/_arrow/series.py#L719-L747

narwhals/_plan/arrow/acero.py
```diff
@@ -189,10 +199,81 @@ def _order_by(
...
+    left_on: OneOrIterable[str],
+    right_on: OneOrIterable[str],
+    suffix: str = "_right",
+    *,
+    coalesce_keys: bool = True,
+) -> Decl:
+    """Heavily based on [`pyarrow.acero._perform_join`].
+
+    [`pyarrow.acero._perform_join`]: https://github.com/apache/arrow/blob/f7320c9a40082639f9e0cf8b3075286e3fc6c0b9/python/pyarrow/acero.py#L82-L260
+    """
```
dangotbanned: ### TODO: Investigate using non-`table_source` nodes
- AFAICT, [`hashjoin`](https://arrow.apache.org/docs/cpp/acero/user_guide.html#hash-join) should be able to accept things like [`project`](https://arrow.apache.org/docs/cpp/acero/user_guide.html#project)
- Defining the handling for `cross` joins using `acero` *directly* seems very achievable
  https://github.com/narwhals-dev/narwhals/blob/f4787d3f9e027306cb1786db7b471f63b393b8d1/narwhals/_arrow/dataframe.py#L400-L422
- There does need to be a new layer for tracking `Schema` changes
  - Which is needed for `with_columns` also
  - Generally, the responsibility for a future `LogicalPlan`

tests/plan/join_test.py
```diff
@@ -0,0 +1,373 @@
...
+        (["b"], (nwp.col("b") < 5), {"a": [1, 3], "b": [4, 4], "zor ro": [7, 8]}),
+        (["a", "b"], (nwp.col("b") < 5), {"a": [1, 3], "b": [4, 4], "zor ro": [7, 8]}),
+    ],
+)
+def test_join_semi(
+    on: On, predicate: nwp.Expr, expected: Data
+) -> None:  # pragma: no cover
+    data = {"a": [1, 3, 2], "b": [4, 4, 6], "zor ro": [7.0, 8.0, 9.0]}
+    df = dataframe(data)
+    other = df.filter(predicate)  # type: ignore[attr-defined]
```
dangotbanned: It's interesting I got this far *without* `DataFrame.filter` 😂


I do have an `acero` version, which operates over `pc.Expression` though

https://github.com/narwhals-dev/narwhals/blob/b0c2a4db692cd267e660c73ef0bd1b462d904cf1/narwhals/_plan/arrow/acero.py#L172-L174


So the missing link between those two is approximately this:

https://github.com/apache/datafusion-python/blob/e75addfa64a91c0d91ef059917a451e04495b83a/src/pyarrow_filter_expression.rs#L96-L181


But with a fallback to an eager path like `main`:

https://github.com/narwhals-dev/narwhals/blob/8ac061c96ef942352505bb730d2bd5e014e37ef1/narwhals/_arrow/dataframe.py#L521-L529


### Edit
Moved to (#2572), since it wasn't *needed* yet

narwhals/_plan/dataframe.py
```diff
@@ -172,3 +213,89 @@ def group_by(
...
+                    dtype=self.version.dtypes.Boolean(),
+                    backend=self.implementation,
+                )
+            elif is_series(first):
+                series = first
+            else:
+                return super().filter(first)
+            return self._with_compliant(self._compliant.filter(series._compliant))
+        non_mask = cast("tuple[OneOrIterable[IntoExprColumn],...]", predicates)
+        return super().filter(*non_mask, **constraints)
```
dangotbanned: If anyone see's this - don't copy this logic to solve (#3182)

I haven't handled it here either

### Update
Should be clearer now with these failing tests ([test: Add `test_filter_mask_mixed`](https://github.com/narwhals-dev/narwhals/pull/3173/commits/8433b2d75438df98004a3c850ad23628e2376836))

narwhals/_plan/arrow/dataframe.py
```diff
@@ -163,14 +163,11 @@ def join_cross(self, other: Self, *, suffix: str = "_right") -> Self:
...
-                mask = resolved.broadcast(len(self)).native
-            else:
-                mask = acero.lit(resolved.native)
+        resolved = Expr.from_named_ir(predicate, self)
+        if isinstance(resolved, Expr):
+            mask = resolved.broadcast(len(self)).native
         else:
-            mask = predicate.native
+            mask = acero.lit(resolved.native)
         return self._with_native(self.native.filter(mask))
```
dangotbanned, line 173: @FBruzzesi https://github.com/narwhals-dev/narwhals/pull/3183#discussion_r2411356288

This is very nice now 😄

narwhals/_plan/arrow/acero.py
dangotbanned: The cross join case is resolved as of (https://github.com/narwhals-dev/narwhals/pull/3173/commits/0f0647960293272d4af5feb54145699a48795b2c)

Moved the rest to #2572

PR #3175, opened 2025-10-03
fix: add `--upgrade` flag to `uv sync --dev`

akmalsoliev:
The `uv sync` command was using pinned dependency versions from `uv.lock` instead of fetching the latest versions. This caused a hidden issue where tests were running against outdated dependencies - notably discovered with duckdb failing locally while CI tests passed.

Adding the `--upgrade` flag forces `uv sync` to fetch the latest versions of dependencies, ignoring the pinned versions in the lock file.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [X] ✅ Test
- [ ] 🐳 Other

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -79,7 +79,9 @@ jobs:
           uv venv -p 3.12
           git log
       - name: install-basics
-        run: uv pip install --upgrade tox virtualenv setuptools hatch "click!=8.3.0" --system
+        run:
+          uv pip install --upgrade tox virtualenv setuptools hatch
+          "click!=8.3.0" --system
       - name: install-marimo-dev
         run: |
           cd marimo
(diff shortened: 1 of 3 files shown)


.github/workflows/downstream_tests.yml
```diff
@@ -502,7 +505,7 @@ jobs:
       - name: install-validoopsie-dev
         run: |
           cd validoopsie
-          uv sync --dev
+          uv sync --dev --upgrade
```
FBruzzesi, line 508: Can we keep only these changes? The other ones seem not relevant 👀

akmalsoliev: > s the issue with duckdb something we can/should address in Narwhals or something that it's downstream? (Sorry I am reviewing with limited access and don't have an easy way to check the logs)

Already debugged with @MarcoGorelli, he should have more intel. 

> Can we keep only these changes? The other ones seem not relevant 👀

You don't like my formatter glorious formatter!? 🤣 yes sure, just keep that change, I have format on save, so it automatically does that when I close the file.

MarcoGorelli:
thanks @akmalsoliev !

this does make ci fail (nice!), but we'll need to address it to merge

for duckdb 1.4+, we can use `fetch_arrow_table`, and before that, just `arrow`, in `collect`

MarcoGorelli (approved):
thanks @akmalsoliev !

PR #3190, opened 2025-10-09
perf: Prefer `Iterator > tuple > list`, use native `pyarrow.repeat`, simplify `nw.concat_str` for DuckDB backend

FBruzzesi:
# Description (Updated)

Summarizing from conversations below: the PR spiraled a bit, but in general I am quite happy with it's state. There are a few performance improvements:

- A general preference to use iterators over tuples over lists
- Use native `pyarrow.repeat`
- Minimizing loops as well - this should always be something to keep in mind
- Simplify (and optimize `concat_str` for DuckDB backend

---

There was also a conversation in using `itertools.repeat` over `(value, ) * n`.

I shared a tiny benchmark here and in a private conversation. My current understanding is that `itertools.repeat` has some initial overhead in the object creation, hence for tiny since of `n`, which should be our use case here (*), we will prefer the pattern `(value, ) * n`.

(*) The use cases are to create a iterable that matches the size of sort/partition by columns. My argument is that generally the size of these cases is <10. For other order of magnitude it won't be a tuple creation to create a bottleneck (e.g. sorting by 100s of columns).

---

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

_It's not much but it's honest work._

Especially for small size `(x, ) * n` is much faster (10x) than `[x] * n`: it's basically a very low overhead vs a very very low overhead.

The `pyarrow.repeat` can be considered a nicer change 😂

---

~~Really feel free to close this, I won't get offended~~

### DIFF
--- a/narwhals/_duckdb/namespace.py
+++ b/narwhals/_duckdb/namespace.py
@@ -37,6 +37,8 @@ if TYPE_CHECKING:
     from narwhals._utils import Version
     from narwhals.typing import ConcatMethod, IntoDType, NonNestedLiteral
 
+VARCHAR = duckdb_dtypes.VARCHAR
+
 
 class DuckDBNamespace(
     SQLNamespace[DuckDBLazyFrame, DuckDBExpr, "DuckDBPyRelation", Expression]
@@ -102,21 +104,13 @@ class DuckDBNamespace(
         self, *exprs: DuckDBExpr, separator: str, ignore_nulls: bool
     ) -> DuckDBExpr:
         def func(df: DuckDBLazyFrame) -> list[Expression]:
-            cols = tuple(chain.from_iterable(expr(df) for expr in exprs))
-            if not ignore_nulls:
-                n_cols = len(cols)
-                sep_exp = lit(separator)
-                null_mask_result = reduce(operator.or_, (s.isnull() for s in cols))
-                # Alternate column values and the separator to then concatenate them all
-                # together. We use an empty string instead of the separator after the very
-                # last column (i.e. when i == n_cols) so that when concatenating there is
-                # not extra trailing separator in the result.
-                cols_separated = chain.from_iterable(
-                    (col.cast(duckdb_dtypes.VARCHAR), lit("") if i == n_cols else sep_exp)
-                    for i, col in enumerate(cols, start=1)
-                )
-                return [when(~null_mask_result, concat_str(*cols_separated))]
-            return [concat_str(*cols, separator=separator)]
+            cols: Iterable[Expression] = chain.from_iterable(e(df) for e in exprs)
+            if ignore_nulls:
+                return [concat_str(*cols, separator=separator)]
+            cols = tuple(cols)
+            null_mask = reduce(operator.or_, (s.isnull() for s in cols))
+            cols_str = (c.cast(VARCHAR) for c in cols)
+            return [when(~null_mask, concat_str(*cols_str, separator=separator))]
 
         return self._expr(
             call=func,


dangotbanned:
There's also [`itertools.repeat`](https://docs.python.org/3/library/itertools.html#itertools.repeat) you could try out?

I've got a couple examples related to `sort`/`over` in (#2572)

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/options.py#L169-L176

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/options.py#L198-L213

FBruzzesi:
Thanks for the quick review @dangotbanned 🙏🏼

I tried a dumb/naive benchmark, but realistically when sorting I would argue that most use cases are with less than 10 columns. Am I missing something obvious here?

image

narwhals/_arrow/dataframe.py
```diff
@@ -809,10 +809,7 @@ def unpivot(
                     pa.Table.from_arrays(
                         [
                             *(self.native.column(idx_col) for idx_col in index_),
-                            cast(
-                                "ChunkedArrayAny",
-                                pa.array([on_col] * n_rows, pa.string()),
-                            ),
+                            pa.repeat(on_col, n_rows).cast(pa.string()),
```
FBruzzesi: @dangotbanned I thought you would care about this: apparently `pa.repeat(pa.scalar(on_col, type=pa.string()), n_rows)` behaves pretty bad with the type checker/pyarrow-stubs. It might be worth reporting upstream?! I know you have been involved there as well 🔥

dangotbanned:
I only peeked a lil bit at the code, it just reminded me of something I wrote

> Am I missing something obvious here?

Apologies, yes kinda 😂 

Let's say you have a situation where you don't need to keep the repeated thing around ...
... that's where `repeat` is handy, because you can just iterate over and forget it.

Whereas `(x, ) * n` *first* needs to create the thing to size `n` **and then** you can iterate


`repeat` has the same cost, regardless of how many you ask for:

image

narwhals/_arrow/dataframe.py
dangotbanned: Are you able to pacify `mypy`/`pyright` by wrapping `on_col`?

```suggestion
                            pa.repeat(lit(on_col), n_rows).cast(pa.string()),
```

**Provided `lit` has been imported** 😄

dangotbanned: I think I misread (https://github.com/narwhals-dev/narwhals/pull/3190#discussion_r2417536423) 😭

But there is potential difference by including `type=...
`

dangotbanned: ooooh, ignore the above - this `Unknown` is probably the issue

<details><summary>Show screenshot</summary>
<p>

image

</p>
</details> 



I'm gonna take a stab at fixing locally 🤞

dangotbanned: @FBruzzesi this got a lil out of control but to explain

[refactor(suggestion): Simplify, fix typing in `ArrowDataFrame.unpivot`](https://github.com/narwhals-dev/narwhals/pull/3190/commits/5a434ba745300c6943db5a214bbdbd12c973e8c5)

- We now have `utils.repeat`
  - Handles the rough (typing) edges internally (https://github.com/narwhals-dev/narwhals/pull/3190#discussion_r2417536423)
  - Although I couldn't see why `pa.string` was needed at all?
- Version branching happens *once* (rather than every method call)
  - And we can use it for any other calls to `concat_tables`
- Removed the quadratic `*(self.native.column(idx_col) for idx_col in index_),` constant loop
  - That selection happens once (`index_cols`)
  - Then the loop is just two "appends" 😅
    - Note that all arrow data is immutable!

narwhals/_arrow/utils.py
```diff
@@ -423,8 +440,8 @@ def pad_series(
     offset_left = window_size // 2
     # subtract one if window_size is even
     offset_right = offset_left - (window_size % 2 == 0)
-    pad_left = pa.array([None] * offset_left, type=series._type)
-    pad_right = pa.array([None] * offset_right, type=series._type)
+    pad_left = pa.repeat(None, offset_left).cast(series._type)
+    pad_right = pa.repeat(None, offset_right).cast(series._type)
     concat = pa.concat_arrays([pad_left, *series.native.chunks, pad_right])
     return series._with_native(concat), offset_left + offset_right
```
dangotbanned:
```suggestion
    chunks = series.native.chunks
    arrays = nulls_like(offset_left, series), *chunks, nulls_like(offset_right, series)
    return series._with_native(pa.concat_arrays(arrays)), offset_left + offset_right
```

narwhals/_compliant/expr.py
```diff
@@ -387,7 +387,7 @@ def _reuse_series_inner(
             for series in self(df)
         ]
         aliases = self._evaluate_aliases(df)
-        if [s.name for s in out] != list(aliases):  # pragma: no cover
+        if tuple(s.name for s in out) != tuple(aliases):  # pragma: no cover
             msg = (
                 f"Safety assertion failed, please report a bug to https://github.com/narwhals-dev/narwhals/issues\n"
                 f"Expression aliases: {aliases}\n"
```
dangotbanned: I've never really understood why we have uncovered assertions like this, *but* ...

... here's a version that doesn't need `list`(s) or `tuple`(s) 😏 

```suggestion
        names = (s.name for s in out)
        if any(
            name != alias for name, alias in zip_strict(names, self._evaluate_aliases(df))
        ):  # pragma: no cover
            msg = (
                f"Safety assertion failed, please report a bug to https://github.com/narwhals-dev/narwhals/issues\n"
                f"Expression aliases: {self._evaluate_aliases(df)}\n"
```

narwhals/_ibis/namespace.py
```diff
@@ -80,7 +80,7 @@ def concat_str(
         self, *exprs: IbisExpr, separator: str, ignore_nulls: bool
     ) -> IbisExpr:
         def func(df: IbisLazyFrame) -> list[ir.Value]:
-            cols = list(chain.from_iterable(expr(df) for expr in exprs))
+            cols = tuple(chain.from_iterable(expr(df) for expr in exprs))
```
dangotbanned:
```suggestion
            cols = chain.from_iterable(expr(df) for expr in exprs)
```

narwhals/_duckdb/namespace.py
```diff
@@ -102,19 +102,20 @@ def concat_str(
...
+                        if i == n_cols
                         else (col.cast(duckdb_dtypes.VARCHAR), lit(separator))
-                        for i, col in enumerate(cols)
-                    ]
+                        for i, col in enumerate(cols, start=1)
+                    )
                     for y in x
-                ]
+                )
                 return [when(~null_mask_result, concat_str(*cols_separated))]
```
dangotbanned: I was *initially* trying to see if I could avoid collecting `cols`.

I can't seem to wrap my head arund this method, so take this instead :smile:

```suggestion
                null_mask_result = reduce(operator.and_, (s.isnotnull() for s in cols))
                cols_separated = (
                    y
                    for x in (
                        (col.cast(duckdb_dtypes.VARCHAR),)
                        if i == n_cols
                        else (col.cast(duckdb_dtypes.VARCHAR), lit(separator))
                        for i, col in enumerate(cols, start=1)
                    )
                    for y in x
                )
                return [when(null_mask_result, concat_str(*cols_separated))]
```

IMO, there are waaaay tpoo many loops here:
- 4x `for` statements here alone
- 1x `chain.from_iterable` just above

And all of that happens *inside* `func`

narwhals/_pandas_like/namespace.py
```diff
@@ -350,7 +350,7 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
                 ]
 
                 sep_array = init_value.from_iterable(
-                    data=[separator] * len(init_value),
+                    data=(separator,) * len(init_value),
```
dangotbanned: Was kinda surprised this worked locally :smile:

```suggestion
                xp = series[0]._array_funcs
                sep_array = init_value.from_iterable(
                    xp.repeat(separator, len(init_value)),
```

narwhals/_duckdb/namespace.py
FBruzzesi: I don't think there is a real way to reduce the 3 main for statement.
The nested one can be replaced by:

```py
cols_separated = tuple(
    chain.from_iterable(
        (col.cast(duckdb_dtypes.VARCHAR), lit(separator)) for col in cols
    )
)[:-1]
```

which is something I quite like

FBruzzesi: Settled on:

```py
sep_exp = lit(separator)
cols_separated = chain.from_iterable(
    (
        col.cast(duckdb_dtypes.VARCHAR),
        # Use empty string instead of separator for the trailing character
        lit("") if i == n_cols else sep_exp,
    )
    for i, col in enumerate(cols, start=1)
)
```

in https://github.com/narwhals-dev/narwhals/pull/3190/commits/28f5daa57ef4301b67cd4350a86555163fd45d17

dangotbanned: @FBruzzesi I really like this!

One last thing though, is it possible to rewrite it so your comment *is* the code?

```py
# Use empty string instead of separator for the trailing character
```

If I took that advice literally, we'd have:

```py
sep_exp = lit(separator)
empty_string = lit("")
cols_separated = chain.from_iterable(
    (
        col.cast(duckdb_dtypes.VARCHAR),
        empty_string if i == n_cols else sep_exp,
    )
    for i, col in enumerate(cols, start=1)
)
```

But if you were able to write, it could be even clearer at a glance 🙂 

```py
trailing_char = empty_string if <some condition> else sep_exp
```


> [!NOTE]
> This is more of a *wish* than anything, so don't worry if not

narwhals/_duckdb/dataframe.py
```diff
@@ -394,24 +394,22 @@ def unique(
         if error := self._check_columns_exist(subset_):
             raise error
         tmp_name = generate_temporary_column_name(8, self.columns, prefix="row_index_")
-        descending_and_nulls_last_value = (
-            (True,) * len(order_by) if order_by and keep == "last" else None
-        )
+        flags = extend_bool(True, len(order_by)) if order_by and keep == "last" else None
         if keep == "none":
```
dangotbanned, line 398: Aha you were braver than me 😂

dangotbanned: @FBruzzesi if you're able to come up a with a better short name than `flags` - I'm open to it

The only other one I thought of was `order`, but there's already `order_by` so it seemed overloaded 😔

narwhals/_duckdb/namespace.py
FBruzzesi: > But if you were able to write, it could be even clearer at a glance 🙂
> ```py
> trailing_char = empty_string if <some condition> else sep_exp
> ```

I don't think that's the right variable name to begin with. What I meant to represent was the trailing value _for the entire iterable_, not for the single tuple.

I am updating with the comment with:

```py
# Alternate column values and the separator to then concatenate them all
# together. We use an empty string instead of the separator after the very
# last column (i.e. when i == n_cols) so that when concatenating there is
# not extra trailing separator in the result.
cols_separated = chain.from_iterable( ...
```

dangotbanned: I wanted less comment not more 😭

dangotbanned: I meant if you could rewrite it so you *could* assign a name to what was a comment - that would really help readibility for me

dangotbanned: - https://blog.codinghorror.com/coding-without-comments/
- https://medium.com/@mklstudio80/why-you-should-write-fewer-comments-not-more-b8b2ffa73eaa

FBruzzesi: I will gladly take your updated comment to get away with this! 😂

> This is more of a wish than anything, so don't worry if not

---

> I meant if you could rewrite it so you could assign a name to what was a comment - that would really help readibility for me

The only way I can think of to assign it to a variable _just_ for the sake of naming is by having a function (that takes a boolean `is_trailing_char` and return the second value of the tuple) and calling that in the for loop (which I am not a big fan of). Also the _honest_ variable name would be something like `separator_or_trailing_char` which is not very descriptive in my opinion?!

Also it would take the same space as the 3-4 lines comment 🙈

dangotbanned: > I will gladly take your updated comment to get away with this! 😂

Absolutely fine by me ☺️

narwhals/_duckdb/dataframe.py
FBruzzesi: I think it's pretty much ok, to be verbose we could add `window_flags` but I don't think there is the need as these functions are pretty much scoped/centered to create a single window_expression to begin with anyway.
Yes I would avoid `order` for sure!

narwhals/_duckdb/namespace.py
dangotbanned: I woke up with another idea for this 😄 

It didn't pan out, but led to a better one 

@FBruzzesi I'll make the suggestion as a commit, but don't feel obligated to take it

### Updated
- [refactor(suggestion): Even simpler `concat_str`](https://github.com/narwhals-dev/narwhals/pull/3190/commits/4c37bf755d19cacfd8072c68a81a3298a586ace8)

FBruzzesi: Oh well, that's pretty neat! 🙏🏼

FBruzzesi:
Thanks @dangotbanned, I am happy to hear that :) 

Would something like the following be too much as a title/scope? 

"perf: Prefer `Iterator > tuple > list`, use native `pyarrow.repeat`, simplify `nw.concat_str` for DuckDB backend"

In my defense, it's easy to spiral in these kind of changes 😂

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3190#issuecomment-3393384469

LGTM!

A title change + a new issue for the other backends and I think we should merge 🙌

MarcoGorelli:
nice one, thanks both!

PR #1045, opened 2024-09-23
feat: add `DataFrame|LazyFrame.cast` method

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added 
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below.

Additionally extends `Expr|Series.cast` with `strict` argument

PR #3191, opened 2025-10-09
chore: Align `nw.concat(..., how="vertical")` behavior across backend for schema mismatch

FBruzzesi:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [x] Tests added
- [x] Documented the changes

We were a bit far from polars behavior for vertical concat. This PR aims to better align with it.

Polars (and some other backends) do not allow to concatenate with different dtypes. The casting has to happen explicitly rather than implicitly. Hence we now check for schema equality rather than column names only.

The error raised tries to match the (lazy) polars' one.

The schemas of the various frames are evaluated lazily for performance

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -47,6 +47,16 @@ repos:
       entry: python -m utils.check_api_reference
       language: python
       additional_dependencies: [polars]
+    - id: sort-api-reference
+      name: sort-api-reference
+      pass_filenames: false
+      entry: python -m utils.sort_api_reference
+      language: python
+    - id: check-slotted-classes
+      name: check-slotted-classes
+      pass_filenames: false
+      entry: python -m utils.check_slotted_classes
+      language: python
     - id: imports-are-banned
       name: import are banned (use `get_pandas` instead of `import pandas`)
       entry: python utils/import_check.py
--- a/docs/api-reference/dataframe.md
+++ b/docs/api-reference/dataframe.md
@@ -44,13 +44,13 @@
         - shape
         - sort
         - tail
-        - top_k
         - to_arrow
         - to_dict
         - to_native
         - to_numpy
         - to_pandas
         - to_polars
+        - top_k
         - unique
         - unpivot
         - with_columns
--- a/docs/api-reference/expr_dt.md
+++ b/docs/api-reference/expr_dt.md
@@ -18,12 +18,12 @@
         - replace_time_zone
         - second
         - timestamp
+        - to_string
         - total_microseconds
         - total_milliseconds
         - total_minutes
         - total_nanoseconds
         - total_seconds
-        - to_string
         - truncate
         - weekday
         - year
--- a/docs/api-reference/lazyframe.md
+++ b/docs/api-reference/lazyframe.md
@@ -25,8 +25,8 @@
         - sink_parquet
         - sort
         - tail
-        - top_k
         - to_native
+        - top_k
         - unique
         - unpivot
         - with_columns
--- a/docs/api-reference/narwhals.md
+++ b/docs/api-reference/narwhals.md
@@ -44,9 +44,9 @@ Here are the top-level functions available in Narwhals.
         - read_parquet
         - scan_csv
         - scan_parquet
+        - show_versions
         - sum
         - sum_horizontal
-        - show_versions
         - to_native
         - to_py_scalar
         - when
--- a/docs/api-reference/schema.md
+++ b/docs/api-reference/schema.md
@@ -4,13 +4,13 @@
     handler: python
     options:
       members:
-        - names
         - dtypes
         - from_arrow
         - from_native
         - from_pandas_like
         - from_polars
         - len
+        - names
         - to_arrow
         - to_pandas
         - to_polars
--- a/docs/api-reference/series.md
+++ b/docs/api-reference/series.md
@@ -59,8 +59,8 @@
         - median
         - min
         - mode
-        - name
         - n_unique
+        - name
         - null_count
         - pipe
         - quantile
@@ -76,8 +76,8 @@
         - scatter
         - shape
         - shift
-        - sort
         - skew
+        - sort
         - sqrt
         - std
         - sum
@@ -86,10 +86,10 @@
         - to_dummies
         - to_frame
         - to_list
+        - to_native
         - to_numpy
         - to_pandas
         - to_polars
-        - to_native
         - unique
         - value_counts
         - var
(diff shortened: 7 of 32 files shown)


narwhals/_utils.py
```diff
@@ -2109,3 +2109,22 @@ def normalize_path(source: FileSource, /) -> str:
     from pathlib import Path
 
     return str(Path(source))
+
+
+def validate_concat_vertical_schemas(schemas: Iterable[Mapping[str, DType]]) -> None:
```
FBruzzesi, line 2125: By passing an iterable, we can avoid to immediately evaluate all the schemas

narwhals/_utils.py
```diff
@@ -2109,3 +2109,22 @@ def normalize_path(source: FileSource, /) -> str:
...
+    schemas_iter = iter(schemas)
+    schema_0 = next(schemas_iter)
+    for idx, schema_current in enumerate(schemas_iter, start=1):
+        if schema_0 != schema_current:
+            msg = (
+                "'union'/'concat' inputs should all have the same schema,got\n"
+                f"{_pretty_format_schema(schema_0, index=0)}\n"
+                " and\n"
```
FBruzzesi, line 2133: The space here is not a typo, that's what polars has 😂

FBruzzesi:
QQ: Should we consider pushing the validation at the narwhals level to have it fully standardized? 🤔

dangotbanned:
> Polars (and some other backends) do not allow to concatenate with different dtypes. The casting has to happen explicitly rather than implicitly. Hence we now check for schema equality rather than column names only.

Could we implement the `_relaxed` variants *first*? (After finding which backends support it)

https://docs.pola.rs/api/python/stable/reference/api/polars.concat.html

I'm pretty sure that *at least* [`pyarrow.concat_tables`](https://arrow.apache.org/docs/python/generated/pyarrow.concat_tables.html) would work via `promote_options="permissive"`

IMO, it seems like this change could break things downstream - so having an alternative to suggest may make the transition smoother 🤞

FBruzzesi:
Currently (at latest dependencies versions) polars, pyarrow and ibis would raise natively with schema mismatch. Notice that for duckdb and ibis we also already have a schema check, which we don't have for pyspark and dask.

Table view:

| Backend | Narwhals behavior | Native behavior on schema mismatch |  
| -- | -- | -- |
| polars | Nothing | It raises two different exceptions (eager vs lazy) |
| pandas-like | Raises on **columns** mismatch | Relaxed |
| pyarrow | Raises on **columns** mismatch | Raises |
| dask | Raises on **columns** mismatch | Relaxed |
| duckdb | Raises on **schema** mismatch | Relaxed |
| ibis | Raises on **schema** mismatch | Raises |
| spark-like | Raises on **columns** mismatch | Relaxed |

> Could we implement the _relaxed variants first? (After finding which backends support it)

Oh boy, that's definitely a _way_ bigger task than anticipated, and I am not really interested in it to be fair 😂

> IMO, it seems like this change could break things downstream

- For the eager case: polars and pyarrow would already raise on schema mismatch, pandas is the only relaxed behavior (which is definitely _too_ relaxed in some cases).
- For the lazy case: the behavior is mixed and we do not even match the native one.

I understand this is a risk, but to me it seems uncrackable/unreliable from a user perspective. Arguably we could maintain the current behavior for stable.v1, _maybe_ for stable.v2, but for sure I would want the changes in the main namespace (which has a few less promises on stability).

> so having an alternative to suggest may make the transition smoother 🤞

Sure, one alternative is to suggest to cast the dtypes before calling `nw.concat(..., how="vertical")`. Now that I think about it, also for the `how="diagonal"` case we should check common columns dtypes

dangotbanned:
## `_relaxed` variants
> [!NOTE]
> I'm about 90% sure I want to split this into another issue
> just letting the idea flow out a bit 😉  

> > Could we implement the `_relaxed` variants first? (After finding which backends support it)
> 
> Oh boy, that's definitely a _way_ bigger task than anticipated, and I am not really interested in it to be fair 😂
> 

That's fair!

I was thinking if a backend only had a strict `UNION`, then we'd just need to implement a version of `polars`' supertyping rules on *our* `DTypes`:
- https://github.com/pola-rs/polars/blob/40e03c19b67dbda3b81becb5321ece3900fec222/crates/polars-core/src/utils/supertype.rs#L141-L504

Or I guess the [DuckDB rules](https://duckdb.org/docs/stable/sql/data_types/typecasting#casting-operations-matrix) have a nicer visual to digest 😄 

<details><summary><b>Show Casting Operations Matrix</b></summary>
<p>

typecasting-matrix

</p>
</details> 

A nice thing about that would be we can ensure consistency by performing the same casts for every backend.
So maybe the implementation would just live in `narwhals.functions`?
E.g. `"vertical_relaxed"`, just applies the supertyping rules at the `narwhals`-level, then calls `concat(..., how="vertical")`

dangotbanned:
> > so having an alternative to suggest may make the transition smoother 🤞
> 
> Sure, one alternative is to suggest to cast the dtypes before calling `nw.concat(..., how="vertical")`.

It would be quite natural to do this with [`DataFrame.cast`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.cast.html) and selectors.
I'd love to revive an old PR of yours for this 🙂
 
- #1045

I found myself doing something pretty similar the other day, but working directly on `pyarrow.{DataType,Schema}`

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/arrow/functions.py#L140-L158

FBruzzesi:
\<rant\>How many different error messages did polars change?\<\\rant\>

MarcoGorelli:
thanks @FBruzzesi !

i'm not totally sure, my original hesitation for being as strict as polars was that in other backends it's easier to accidentally change the datatype (e.g. if a missing value gets inserted), although it is encouraging that the downstream tests pass 🤔

FBruzzesi:
Hey @MarcoGorelli 👋🏼

> i'm not totally sure, my original hesitation for being as strict as polars was that in other backends it's easier to accidentally change the datatype (e.g. if a missing value gets inserted)

Doesn't this apply only to pandas non-nullable dtypes?

> although it is encouraging that the downstream tests pass 🤔

Not sure how used and tested this feature is to be honest.

---

I will move it back to draft state and make a proposal that is also consistent for `how="diagonal"`, and even see how easy it is to implement the "_relaxed" versions as suggested by @dangotbanned. I had an idea but no time to try it out yet :)

Issue #3171, opened 2025-10-02
perf: avoid full-broadcast in horizontal functions


MarcoGorelli:
here for example: https://github.com/narwhals-dev/narwhals/blob/c5f4058e6b88ab77b7c373a6323b2e2ed399380c/narwhals/_pandas_like/namespace.py#L335

it should be possible to avoid doing a full broadcast

PR #3193, opened 2025-10-10
fix: correctly preserve arrow dtypes for pandas-like, improve `concat_str` performance

### LINKED ISSUE #3192: bug: casting `int64[pyarrow]` to `nw.string` results in non-pyarrow type
I think i'd misunderstood `string[pyarrow]`:
- I thought it was a pyarrow nullable dtype, like `int64[pyarrow]`
- but it actually behaves like a numpy type, it just happens to use pyarrow for storage

demo:
```python
import pandas as pd
import pyarrow as pa

s = pd.DataFrame({'a': ['foo', None]}, dtype='string[pyarrow]')['a']
print(s.dtype)
print(s.str.len())

s = pd.DataFrame({'a': ['foo', None]}, dtype=pd.ArrowDtype(pa.string()))['a']
print(s.dtype)
print(s.str.len())import pandas as pd
import pyarrow as pa
```
```
string
0       3
1    <NA>
Name: a, dtype: Int64
string[pyarrow]
0       3
1    <NA>
Name: a, dtype: int32[pyarrow]
```

Note how in the first case, `str.len` actually returns `int64`!

Furthermore, the dtype returned when we do `convert_dtypes` is `pd.ArrowDtype(pa.string())`, not `string[pyarrow]`

so, when a user does `cast(nw.string)` on a pyarrow-backed pandas series, we should convert to `pd.ArrowDtype(pa.string())`, not `string[pyarrow]`

i'll make a pr

MarcoGorelli:
closes #3192 
closes part of #3171 

Bonus perf improvement:
```python
import pandas as pd
import pyarrow as pa
import narwhals as nw
import time

df = nw.from_native(pd.DataFrame({'a': ['foo', 'bar']*100_000}, dtype='string[pyarrow]'))
start = time.perf_counter_ns()
res = df.select(nw.format('bar {} bat', nw.col('a')))
print(time.perf_counter_ns() - start)
print(res.to_native().dtypes)

```

On `main`, best of 5:
- 138_794_095 ns

Here:
- 41_209_072 ns

### DIFF
--- a/tests/expr_and_series/cast_test.py
+++ b/tests/expr_and_series/cast_test.py
@@ -13,7 +13,7 @@ from tests.utils import (
     Constructor,
     ConstructorEager,
     assert_equal_data,
-    is_windows,
+    is_pyarrow_windows_no_tzdata,
     time_unit_compat,
 )
 
@@ -213,8 +213,9 @@ def test_cast_datetime_tz_aware(
         or "ibis" in str(constructor)
     ):
         request.applymarker(pytest.mark.xfail)
-    if "pyarrow" in str(constructor) and is_windows():
-        pytest.skip()
+    request.applymarker(
+        pytest.mark.xfail(is_pyarrow_windows_no_tzdata(constructor), reason="no tzdata")
+    )
 
     data = {
         "date": [
@@ -242,8 +243,9 @@ def test_cast_datetime_utc(
         or "sqlframe" in str(constructor)
     ):
         request.applymarker(pytest.mark.xfail)
-    if "pyarrow" in str(constructor) and is_windows():
-        pytest.skip()
+    request.applymarker(
+        pytest.mark.xfail(is_pyarrow_windows_no_tzdata(constructor), reason="no tzdata")
+    )
 
     data = {
         "date": [


tests/frame/schema_test.py
```diff
@@ -387,6 +387,11 @@ def test_all_nulls_pandas() -> None:
 def test_schema_to_pandas(
     dtype_backend: DTypeBackend | Sequence[DTypeBackend] | None, expected: dict[str, Any]
 ) -> None:
+    if (
+        dtype_backend == "pyarrow"
+        or (isinstance(dtype_backend, list) and "pyarrow" in dtype_backend)
+    ) and PANDAS_VERSION < (1, 5):
+        pytest.skip()
```
dangotbanned, line 394: What's going on here? 🧐

(feel free to ignore if this is temporary)

MarcoGorelli: if `dtype_backend='pyarrow'`, then it can't be used on old pandas versions

dangotbanned: Oh I see!

You might be able to do something funky with a fixture for version compat?

I think there are some in that file for `Schema.from_native` for other `pandas` quirks

MarcoGorelli: dunno about funky fixtures 😄 

any concerns on the rest of the changes?

narwhals/_pandas_like/namespace.py
```diff
@@ -332,9 +332,8 @@ def concat_str(
 
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             expr_results = [s for _expr in exprs for s in _expr(df)]
-            align = self._series._align_full_broadcast
-            series = align(*(s.cast(string) for s in expr_results))
-            null_mask = align(*(s.is_null() for s in expr_results))
+            series = [s.cast(string) for s in expr_results]
+            null_mask = [s.is_null() for s in expr_results]
```
FBruzzesi, line 336: Took me 5 mins to figure out that alignment is happening in the reduce `or_` and `__add__` ops since these are compliant series and not native 🫡

narwhals/_pandas_like/series.py
FBruzzesi, line 1: 👌🏼👌🏼 beautiful!

narwhals/_pandas_like/utils.py
```diff
@@ -494,6 +493,25 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
...
+            # Note: this is different from `string[pyarrow]`, even though the repr
+            # looks the same.
+            # >>> pd.DataFrame({'a':['foo']}, dtype='string[pyarrow]')['a'].str.len()
+            # 0    3
+            # Name: a, dtype: Int64
+            # >>> pd.DataFrame({'a':['foo']}, dtype=pd.ArrowDtype(pa.string()))['a'].str.len()
+            # 0    3
+            # Name: a, dtype: int32[pyarrow]
+            #
+            # `ArrowDType(pa.string())` is what `.convert_dtypes(dtype_backend='pyarrow')`, so
```
FBruzzesi: > `ArrowDType(pa.string())` is what `.convert_dtypes(dtype_backend='pyarrow')` 

is something missing? 😂

tests/frame/schema_test.py
dangotbanned: I'll do a quick once over 🙂

narwhals/_pandas_like/series.py
```diff
@@ -387,103 +393,87 @@ def first(self) -> PythonLiteral:
...
+    def _with_binary(self, op: Callable[..., PandasLikeSeries], other: Any) -> Self:
+        ser, other_native = align_and_extract_native(self, other)
+        preserve_broadcast = self._broadcast and getattr(other, "_broadcast", True)
+        return self._with_native(
+            op(ser, other_native), preserve_broadcast=preserve_broadcast
+        ).alias(self.name)
+
+    def _with_rbinary(self, op: Callable[..., PandasLikeSeries], other: Any) -> Self:
+        return self._with_binary(lambda x, y: op(y, x), other).alias(self.name)
+
```
dangotbanned: Personally would go for a suffixed version of `_with_binary` so they show up together in auto-complete

Although there aren't many `_with`'s here so who knows

image

dangotbanned: Like ...

https://github.com/narwhals-dev/narwhals/blob/e3fd9954422797f64e2fdaadede4513b8138d1bc/narwhals/_compliant/expr.py#L463-L470

narwhals/_pandas_like/utils.py
```diff
@@ -494,6 +493,25 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
         return pd_type
     if into_pd_type := NW_TO_PD_DTYPES_BACKEND.get(base_type):
         return into_pd_type[dtype_backend]
+    if isinstance_or_issubclass(dtype, dtypes.String):
```
dangotbanned: If we don't need the type checking on `dtype` (e.g. for an attribute), then we can use a quicker check on the *already resolved* `base_type` 😎

```suggestion
    if issubclass(base_type, dtypes.String):
```

narwhals/_pandas_like/utils.py
```diff
@@ -494,6 +493,25 @@ def narwhals_to_native_dtype(  # noqa: C901, PLR0912
...
+            #
+            # `ArrowDType(pa.string())` is what `.convert_dtypes(dtype_backend='pyarrow')` converts to,
+            # so we use that here.
+            return pd.ArrowDtype(pa.string())
+        if dtype_backend == "numpy_nullable":
+            return "string"
+        return str
     if isinstance_or_issubclass(dtype, dtypes.Datetime):
         if is_pandas_or_modin(implementation) and PANDAS_VERSION < (
             2,
```
dangotbanned, line 517: This is making we wonder if for the other `pd.ArrowDtype`s we could simplify some stuff like:

```py
import pandas as pd
import pyarrow as pa

pd_type = pd.ArrowDtype(pa.string())
>>> pd_type
string[pyarrow]

>>> pd_type.pyarrow_dtype
DataType(string)

import pyarrow.types

>>> pyarrow.types.is_string(pd_type.pyarrow_dtype)
True

pd_type = pd.ArrowDtype.construct_from_string("date32[day][pyarrow]")
>>> pyarrow.types.is_date(pd_type.pyarrow_dtype)
True
```

That might be nicer to work with for temporal dtypes vs regex parsing?

tests/expr_and_series/cast_test.py
```diff
@@ -209,11 +209,12 @@ def test_cast_datetime_tz_aware(
         "dask" in str(constructor)
         or "duckdb" in str(constructor)
         or "cudf" in str(constructor)  # https://github.com/rapidsai/cudf/issues/16973
-        or ("pyarrow_table" in str(constructor) and is_windows())
         or "pyspark" in str(constructor)
         or "ibis" in str(constructor)
     ):
         request.applymarker(pytest.mark.xfail)
+    if "pyarrow" in str(constructor) and is_windows():
+        pytest.skip()
```
dangotbanned: Tbf, we should probably have this in loads of places

```suggestion
    request.applymarker(
        pytest.mark.xfail(is_pyarrow_windows_no_tzdata(constructor), reason="no TZDATA")
    )
```

tests/frame/schema_test.py
dangotbanned: @MarcoGorelli I'm still not following why this test *didn't* require this condition before 

What would happen *prior* to your changes?

(I wrote the original test IIRC 😅)

MarcoGorelli: because now we use `pd.ArrowDType`

PR #3194, opened 2025-10-10
perf: Add `__slots__` to all `DType`s

### LINKED ISSUE #3185: Add `__slots__` to all `DType`s
### Related
- #500

### Description
I've been using [`__slots__`](https://docs.python.org/3/reference/datamodel.html#slots) *quite* heavily in #2572 and all of the `DType` classes seem like good candidates for using them as well:

- We create many instances of them
- They have very few (or 0) attributes
- We don't need the dynamic functionality of [`__dict__`](https://docs.python.org/3/reference/datamodel.html#object.__dict__)

### Context
Recently I tried experimenting with `Expr` de/serialization in (https://github.com/narwhals-dev/narwhals/commit/d5c00bc2c7ceda094e0b5cf8054689f4c8e5c6c4).
Any `DType`s which have attributes were an issue.

I'd need to be able to discern *which* attributes they have and also if they need passing to the constructor:
- `Datetime`
- `Duration`
- `Enum`
- `Struct`
- `List`
- `Array`

This was already solved for *most* of the new classes I'd added, since they use `__slots__` as a means for 
- *"immutability"*-enforcement
- [`__replace__`](https://docs.python.org/3/library/copy.html#object.__replace__)
- [`typing.dataclass_transform`](https://docs.python.org/3/library/typing.html#typing.dataclass_transform)

<details><summary><b>Show <code>Immutable</code></b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/_immutable.py#L40-L142

</p>
</details> 

It would be great to upstream some of these ideas, but with a limited scope to `DType` for now.
I imagine it would help with serialization following (#3152) as well

dangotbanned:
Closes #3185

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

- [x] Implementent on `nw.dtypes`
- [x] ~~Check if~~ `stable.*` needs empty `__slots__` (**It did**)

### DIFF
--- a/utils/check_slotted_classes.py
+++ b/utils/check_slotted_classes.py
@@ -43,7 +43,7 @@ if unslotted_classes:
     msg = "The following classes are expected to define `__slots__` but they don't:\n"
     cls_list = "\n".join(f"  * {name}" for name in unslotted_classes)
     url = "https://docs.python.org/3/reference/datamodel.html#slots"
-    hint = f"Hint: See for detail {url!r}"
+    hint = f"Hint: For more details see {url!r}"
     print(f"{msg}{cls_list}")
 
 sys.exit(ret)


tests/expr_and_series/cast_test.py
```diff
@@ -371,7 +371,7 @@ def test_cast_typing_invalid() -> None:
     # feel free to update the types used
     # See (https://github.com/narwhals-dev/narwhals/pull/2654#discussion_r2142263770)
 
-    with pytest.raises(AttributeError):
+    with pytest.raises(TypeError):
```
dangotbanned, line 378: See my comment above 😄 

This is a weird error though, it is a `TypeError` because it tries to call `__iter__` on a [descriptor](https://docs.python.org/3/howto/descriptor.html#member-objects-and-slots).

`dtype.fields.__iter__` is `nw.Struct.fields.__iter__`

https://github.com/narwhals-dev/narwhals/blob/2fe3d8c18c42ecdb92cb906fc71d97bf342eb25d/narwhals/_polars/utils.py#L227-L231

dangotbanned:
I'm thinking `Schema` may be a good candidate as well (provided it is allowed on `dict` subclasses)

narwhals/dtypes.py
FBruzzesi, line 1: Have you considered introducing a metaclass to do the job (and avoid forgetting about adding `__slots__ = ()`?

```py

from typing import Any


class SlottedMeta(type):
    def __new__(cls,
        name: str,
        bases: tuple[type, ...],
        namespace: dict[str, Any],
        **kwargs: Any
    ) -> type:
        # Only add empty slots if __slots__ isn't already defined
        if "__slots__" not in namespace:
            namespace["__slots__"] = ()
        return super().__new__(cls, name, bases, namespace, **kwargs)

class DType(metaclass=SlottedMeta):
    ...
```

<details><summary>Checks</summary>

**! Disclaimer**: The following checks are gently offered by an LLM

```py
class Base(metaclass=SlottedMeta):
    pass

class Derived(Base):
    pass

class WithSlots(metaclass=SlottedMeta):
    __slots__ = ("x", "y")


class DerivedWithSlots(Base):
    __slots__ = ("value",)


class Parent(metaclass=SlottedMeta):
    __slots__ = ("a",)

class Child(Parent):
    __slots__ = ("b",)

class GrandChild(Child):
    pass  # Gets empty slots


# Test 1: Base class with no explicit slots gets empty slots
assert Base.__slots__ == ()
print("✓ Test 1: Base class gets empty slots by default")

# Verify no __dict__
b = Base()
assert not hasattr(b, "__dict__")
print("✓ Test 1b: Base instance has no __dict__")

# Verify cannot add attributes
try:
    b.value = 42
    assert False, "Should not be able to add attributes"
except AttributeError:
    print("✓ Test 1c: Cannot add attributes to base instance")


# Test 2: Class with explicit slots keeps them
assert WithSlots.__slots__ == ("x", "y")
print("✓ Test 2: Explicit slots are preserved")

ws = WithSlots()
ws.x = 10
ws.y = 20
assert ws.x == 10 and ws.y == 20
print("✓ Test 2b: Can set slotted attributes")


# Test 3: Derived class without explicit slots gets empty slots
assert Derived.__slots__ == ()
print("✓ Test 3: Derived class gets empty slots by default")

d = Derived()
assert not hasattr(d, "__dict__")
try:
    d.value = 42
    assert False, "Should not be able to add attributes"
except AttributeError:
    print("✓ Test 3b: Cannot add attributes to derived instance")

# Test 4: Derived class with explicit slots works
assert DerivedWithSlots.__slots__ == ("value",)
print("✓ Test 4: Derived class can define its own slots")

dws = DerivedWithSlots()
dws.value = 42
assert dws.value == 42
print("✓ Test 4b: Can set attributes in derived slotted class")

# Test 5: Multi-level inheritance
assert Parent.__slots__ == ("a",)
assert Child.__slots__ == ("b",)
assert GrandChild.__slots__ == ()
print("✓ Test 5: Multi-level inheritance works correctly")

gc = GrandChild()
gc.a = 1  # From Parent
gc.b = 2  # From Child
assert gc.a == 1 and gc.b == 2
print("✓ Test 5b: GrandChild can use parent and grandparent slots")

try:
    gc.c = 3  # Not defined anywhere
    assert False, "Should not be able to add new attributes"
except AttributeError:
    print("✓ Test 5c: Cannot add attributes not in hierarchy")

print("\n✅ All tests passed!")
```

</details>

tests/expr_and_series/cast_test.py
FBruzzesi: > See my comment above 😄

🤔

> This is a weird error though, it is a TypeError because it tries to call __iter__ on a [descriptor](https://docs.python.org/3/howto/descriptor.html#member-objects-and-slots).

Am I getting correctly that this is expected behavior?

dangotbanned: This test was about making sure **static typing** errors would show up when you use things like `nw.Struct` incorrectly

The actual runtime error isn't important, just that our typing warns you when you're naughty 😄 

- https://github.com/narwhals-dev/narwhals/pull/2654#discussion_r2142263770

narwhals/dtypes.py
dangotbanned: I **do** want to add a metaclass to `DType`, but for this specific case you can use `__init_subclass__`

https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/narwhals/_plan/_immutable.py#L99-L104

I marked my own comment as *off-topic* 😂, but we can use a metaclass to get nicer repr on the bare class
- https://github.com/narwhals-dev/narwhals/pull/2969#discussion_r2266125855

FBruzzesi: > for this specific case you can use __init_subclass__

Oh wow that's new to me. I thought `__slots__` had to be specified at creation time, not at initialization time.

Then sure, I am more concerned about forgetting to add them somewhere than anything else to be honest 😂

> I do want to add a metaclass to DType

It's sad that combining metaclasses is not really a easy thing

dangotbanned: > It's sad that combining metaclasses is not really a easy thing

The issues with combining shouldn't be hard for these classes (most are empty 😂)

> Oh wow that's new to me. I thought `__slots__` had to be specified at creation time, not at initialization time.

Determining names that *should* be in there **would** require a metaclass (https://github.com/narwhals-dev/narwhals/pull/3194#discussion_r2421857715) or creating a new class entirely (like [`@dataclasses(slots=True)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass))

> *slots*: If true (the default is `False`), [`__slots__`](https://docs.python.org/3/reference/datamodel.html#object.__slots__) attribute will be generated and new class will be returned instead of the original one. 


**But** empty/inherited slots - that's a breeze 😎

FBruzzesi: > The issues with combining shouldn't be hard for these classes (most are empty 😂)

Not sure my message was clear. What I meant was suppose we have two metaclasses:

```py
class FancyReprMeta: ...
class SlottedMeta: ...
```

As far as I know there is not _one solution fits all_ to combine multiple metaclasses into a single one. But that's a problem for another day.

dangotbanned: @FBruzzesi I've been testing some things out, brain is going to melt ...

Do you want the good news or bad news first? 😭

FBruzzesi: > testing some things out, brain is going to melt ...

&

> dangotbanned marked this pull request as draft

made me partially scared 😂

> brain is going to melt ...

Also, no need to rush anything

dangotbanned: ### Good news (for you)
You were right about needing a metaclass to do this *correctly*.

In (https://github.com/narwhals-dev/narwhals/blob/2403f1b8760ba4df610193a2574e18537008b5d2/tests/plan/immutable_test.py) I was feeling confident with my tests, but it turns out `__dict__` **was still** being created the way I did it.

I missed that because:
1. I *also* added a manual override to `__setattr__`
    i. hiding the fact that it didn't raise in subclasses *without* an empty `__slots__`
2. Since `__slots__` *was* always there, it didn't cause any issues


> [!NOTE]
> But I guess this is good news for me too, since you've helped improve #2572 😍

dangotbanned: ### Good news (all round)
Combining mulitple metaclasses is fine (https://github.com/narwhals-dev/narwhals/pull/3194#discussion_r2422978253).

It's really just as simple as this 🙂 

- https://github.com/python/typeshed/blob/776508741d76b58f9dcb2aaf42f7d4596a48d580/stdlib/builtins.pyi#L184-L218

```diff
-class DType:
+class _DTypeMeta(type):
+    def __new__(
+        cls: type[_typeshed.Self],
+        name: str,
+        bases: tuple[type, ...],
+        namespace: dict[str, Any],
+        /,
+        **kwds: Any,
+    ) -> _typeshed.Self:
+        namespace.setdefault("__slots__", ())
+        return super().__new__(cls, name, bases, namespace, **kwds)
+
+
+class DType(metaclass=_DTypeMeta):

-    __slots__ = ()  # <---- lots of those being skipped
-
 
-class _DatetimeMeta(type):
+class _DatetimeMeta(_DTypeMeta):
     @property
     def time_unit(cls) -> TimeUnit:
         """Unit of time. Defaults to `'us'` (microseconds)."""
@@ -546,7 +512,7 @@ class Datetime(TemporalType, metaclass=_DatetimeMeta):
         return f"{class_name}(time_unit={self.time_unit!r}, time_zone={self.time_zone!r})"
 
 
-class _DurationMeta(type):
+class _DurationMeta(_DTypeMeta):
```

dangotbanned: ## Bad news
### 1
`mypy` seems to not be a fan of any kind of `__new__` that I've written a signature for.

```py
narwhals/dtypes.py:73: error: Returning Any from function declared to return "_typeshed.Self"  [no-any-return]
narwhals/dtypes.py:73: error: Argument 2 for "super" not an instance of argument 1  [misc]
            return super().__new__(cls, name, bases, namespace, **kwds)
```

### 2

`mypy` also seems to infer (and then require annotating) bare `DType`s in some cases

```py
tests/expr_and_series/cast_test.py:88: error: Argument 1 to "cast" of "Expr" has incompatible type "_DTypeMeta"; expected "DType | type[NumericType] | type[TemporalType] | type[String] | type[Boolean] | type[Binary] | type[Categorical] | type[Unknown] | type[Object]"  [arg-type]
            nw.col(col_).cast(dtype) for col_, dtype in schema.items()
```

<details><summary>Show diff that fixes it</summary>
<p>

```diff
diff --git a/tests/expr_and_series/cast_test.py b/tests/expr_and_series/cast_test.py
index 549fd76b3..bd9b2d40b 100644
--- a/tests/expr_and_series/cast_test.py
+++ b/tests/expr_and_series/cast_test.py
@@ -18,7 +18,9 @@ from tests.utils import (
 )
 
 if TYPE_CHECKING:
-    from narwhals.typing import NativeLazyFrame
+    from collections.abc import Mapping
+
+    from narwhals.typing import NativeLazyFrame, NonNestedDType
 
 DATA = {
     "a": [1],
@@ -38,7 +40,7 @@ DATA = {
     "o": ["a"],
     "p": [1],
 }
-SCHEMA = {
+SCHEMA: Mapping[str, type[NonNestedDType]] = {
     "a": nw.Int64,
     "b": nw.Int32,
     "c": nw.Int16,
@@ -86,7 +88,7 @@ def test_cast(constructor: Constructor) -> None:
         nw.col(col_).cast(dtype) for col_, dtype in schema.items()
     )
 
-    cast_map = {
+    cast_map: Mapping[str, type[NonNestedDType]] = {
         "a": nw.Int32,
         "b": nw.Int16,
         "c": nw.Int8,
@@ -134,7 +136,7 @@ def test_cast_series(
         .lazy()
         .collect()
     )
-    cast_map = {
+    cast_map: Mapping[str, type[NonNestedDType]] = {
         "a": nw.Int32,
         "b": nw.Int16,
         "c": nw.Int8,

```

</p>
</details> 

### 3
Simply adding a metaclass to `DType` (and making the other metaclasses inherit it) triggers:

```py
31 failed, 9325 passed, 354 skipped, 614 xfailed
```

I think they all boil down to correcting things like this:

```diff
- if type(other) is type:
+ if type(other) is _DTypeMeta
```

<details><summary>Show failures</summary>
<p>

```py
FAILED tests/frame/explode_test.py::test_explode_multiple_cols[pandas[pyarrow]-l3-more_columns1-expected1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[pandas-l3-expected_values1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[duckdb-l3-expected_values1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[sqlframe-l3-expected_values1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_multiple_cols[pandas[pyarrow]-l1-more_columns0-expected0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_shape_error[pandas] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[ibis-l2-expected_values0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_multiple_cols[pandas-l3-more_columns1-expected1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[pandas[pyarrow]-l3-expected_values1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[sqlframe-l2-expected_values0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_shape_error[pandas[pyarrow]] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_shape_error[duckdb] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[ibis-l3-expected_values1] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/v1_test.py::test_enum_v1_is_enum_unstable - AssertionError: assert Enum == <class 'narwhals.dtypes.Enum'>
FAILED tests/frame/explode_test.py::test_explode_shape_error[ibis] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/v1_test.py::test_v1_enum_polars - assert Enum == <class 'narwhals.stable.v1._dtypes.Enum'>
FAILED tests/frame/explode_test.py::test_explode_single_col[pandas[pyarrow]-l2-expected_values0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_shape_error[sqlframe] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[duckdb-l2-expected_values0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_single_col[pandas-l2-expected_values0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/frame/explode_test.py::test_explode_multiple_cols[pandas-l1-more_columns0-expected0] - narwhals.exceptions.InvalidOperationError: `explode` operation not supported for dtype `List(Int32)`, expected List type
FAILED tests/series_only/is_ordered_categorical_test.py::test_is_ordered_categorical_pandas - AssertionError: assert False
FAILED tests/series_only/is_ordered_categorical_test.py::test_is_ordered_categorical_polars - assert False
FAILED tests/frame/schema_test.py::test_dtypes - AssertionError: assert Schema({'a': ...'a': Int64})}) == {'a': <class ...s.Int8'>, ...}
FAILED tests/dtypes_test.py::test_struct_valid - AssertionError: assert Struct({'a': <class 'narwhals.dtypes.Int64'>}) == <class 'narwhals.dtypes.Struct'>
FAILED tests/dtypes_test.py::test_enum_valid[categories1] - AssertionError: assert Enum(categories=[np.str_('a'), np.str_('b')]) == <class 'narwhals.dtypes.Enum'>
FAILED tests/dtypes_test.py::test_enum_valid[categories0] - AssertionError: assert Enum(categories=['a', 'b']) == <class 'narwhals.dtypes.Enum'>
FAILED tests/dtypes_test.py::test_list_valid - AssertionError: assert List(<class 'narwhals.dtypes.Int64'>) == <class 'narwhals.dtypes.List'>
FAILED tests/dtypes_test.py::test_array_valid - AssertionError: assert Array(<class 'narwhals.dtypes.Int64'>, shape=(2,)) == <class 'narwhals.dtypes.Array'>
FAILED tests/dtypes_test.py::test_enum_valid[categories3] - AssertionError: assert Enum(categories=[1, 2, 3]) == <class 'narwhals.dtypes.Enum'>
FAILED tests/dtypes_test.py::test_enum_valid[Test] - AssertionError: assert Enum(categories=[1, 2]) == <class 'narwhals.dtypes.Enum'>

```

</p>
</details> 

> [!IMPORTANT]
> But I dunno man, maybe we save this for when adding a metaclass has a greater pay off? 😅

FBruzzesi: > But I dunno man, maybe we save this for when adding a metaclass has a greater pay off? 😅

Agreed!

I added a simple script to check manually in pre-commit. I tried to leave space to add other modules/classes easily (as _"I'm thinking Schema may be a good candidate as well (provided it is allowed on dict subclasses"_)

https://github.com/narwhals-dev/narwhals/pull/3194/commits/b2f6976a0da3868caa9d84fd1da266ec063b20e2

---

Edit: I appreciate the entire pros and cons analysis above that gives me/us a much better picture and I agree with the final conclusion that for the small gain it's not worth it for now.

I went ahead proposing a different safe guard mechanism with a simple custom pre-commit hook :)

dangotbanned: Oooh nice one, thanks @FBruzzesi 🙏 

Will take a proper look later

utils/check_slotted_classes.py
```diff
@@ -0,0 +1,57 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from types import ModuleType, UnionType
+
+    from typing_extensions import TypeAlias
+
+    if sys.version_info >= (3, 10):
+        _ClassInfo: TypeAlias = type | UnionType | tuple["_ClassInfo", ...]
+    else:
+        _ClassInfo: TypeAlias = type | tuple["_ClassInfo", ...]
```
dangotbanned: I guess we don't run this across versions so it didn't show the issues

- `UnionType` *the symbol* requires `(3, 10)`
  - https://github.com/python/typeshed/blob/776508741d76b58f9dcb2aaf42f7d4596a48d580/stdlib/types.pyi#L712-L741
- `type.__or__` *the operator* also requires `(3, 10)`
  - https://github.com/python/typeshed/blob/776508741d76b58f9dcb2aaf42f7d4596a48d580/stdlib/builtins.pyi#L228-L232

But anyway, using `UnionType` in this way was actually *removed* as a `ruff` lint 😄 

- https://docs.astral.sh/ruff/rules/non-pep604-isinstance/

dangotbanned: I wrote ^ and then went in a different direction after checking it out 🫣 

Based on (https://github.com/narwhals-dev/narwhals/blob/5838dde8e63dc368f7879f5bdd8e36a36d5bf066/tests/get_dtype_backend_test.py)

utils/check_slotted_classes.py
```diff
@@ -0,0 +1,39 @@
...
+        if children := base.__subclasses__():
+            yield from _iter_descendants(*children)
+        else:
+            yield base
+
+
+def iter_unslotted_classes(*bases: type[T_co]) -> Iterator[str]:
+    """Find classes in that inherit from `bases` but don't define `__slots__`."""
+    for tp in sorted(set(_iter_descendants(*bases)), key=repr):
+        if "__slots__" not in tp.__dict__:
```
dangotbanned, line 34: If you remove this filter you get:

```py
('narwhals.dtypes.Array',
 'narwhals.dtypes.Binary',
 'narwhals.dtypes.Boolean',
 'narwhals.dtypes.Categorical',
 'narwhals.dtypes.Date',
 'narwhals.dtypes.Decimal',
 'narwhals.dtypes.Field',
 'narwhals.dtypes.Float32',
 'narwhals.dtypes.Float64',
 'narwhals.dtypes.Int128',
 'narwhals.dtypes.Int16',
 'narwhals.dtypes.Int32',
 'narwhals.dtypes.Int64',
 'narwhals.dtypes.Int8',
 'narwhals.dtypes.List',
 'narwhals.dtypes.Object',
 'narwhals.dtypes.String',
 'narwhals.dtypes.Struct',
 'narwhals.dtypes.Time',
 'narwhals.dtypes.UInt128',
 'narwhals.dtypes.UInt16',
 'narwhals.dtypes.UInt32',
 'narwhals.dtypes.UInt64',
 'narwhals.dtypes.UInt8',
 'narwhals.dtypes.Unknown',
 'narwhals.stable.v1._dtypes.Datetime',
 'narwhals.stable.v1._dtypes.Duration',
 'narwhals.stable.v1._dtypes.Enum')
```

FBruzzesi: Parent classes are missing e.g. `DType` and (non-v1) `Datetime`. I think we should check slots for all of them.

Should it be:

```diff
- yield from _iter_descendants(*children)
+ yield from (base, *children, *_iter_descendants(*children))
```
?

Also, in both cases I am not managing to make v1 types show up 😭

utils/check_slotted_classes.py
FBruzzesi: It was stolen from the python `issubclass` signature itself :)

utils/check_slotted_classes.py
dangotbanned: I will take another look soon, IIRC I explicitly only wanted the most-derived `pandas` dtypes in the other case - but it is easier to *not* do that 😂

dangotbanned: @FBruzzesi an update to (https://github.com/narwhals-dev/narwhals/pull/3194#discussion_r2423665839) following (https://github.com/narwhals-dev/narwhals/pull/3194/commits/3795a4937454f0e24a372d4ebbdc8482e35dccfe)

```
The following classes are expected to define `__slots__` but they don't:
  * narwhals.dtypes.Array
  * narwhals.dtypes.Binary
  * narwhals.dtypes.Boolean
  * narwhals.dtypes.Categorical
  * narwhals.dtypes.DType
  * narwhals.dtypes.Date
  * narwhals.dtypes.Datetime
  * narwhals.dtypes.Decimal
  * narwhals.dtypes.Duration
  * narwhals.dtypes.Enum
  * narwhals.dtypes.Field
  * narwhals.dtypes.Float32
  * narwhals.dtypes.Float64
  * narwhals.dtypes.FloatType
  * narwhals.dtypes.Int128
  * narwhals.dtypes.Int16
  * narwhals.dtypes.Int32
  * narwhals.dtypes.Int64
  * narwhals.dtypes.Int8
  * narwhals.dtypes.IntegerType
  * narwhals.dtypes.List
  * narwhals.dtypes.NestedType
  * narwhals.dtypes.NumericType
  * narwhals.dtypes.Object
  * narwhals.dtypes.SignedIntegerType
  * narwhals.dtypes.String
  * narwhals.dtypes.Struct
  * narwhals.dtypes.TemporalType
  * narwhals.dtypes.Time
  * narwhals.dtypes.UInt128
  * narwhals.dtypes.UInt16
  * narwhals.dtypes.UInt32
  * narwhals.dtypes.UInt64
  * narwhals.dtypes.UInt8
  * narwhals.dtypes.Unknown
  * narwhals.dtypes.UnsignedIntegerType
  * narwhals.stable.v1._dtypes.Datetime
  * narwhals.stable.v1._dtypes.Duration
  * narwhals.stable.v1._dtypes.Enum
```

dangotbanned: The issue was that I tested it *before* removing the imports that were there (https://github.com/narwhals-dev/narwhals/pull/3194/commits/e96f4633c485084cc83d30ba7fe89e230ec41c44)

utils/check_slotted_classes.py
```diff
@@ -0,0 +1,49 @@
...
+
+ret = 0
+unslotted_classes = tuple(iter_unslotted_classes(DType, Field))
+
+if unslotted_classes:
+    ret = 1
+    msg = "The following classes are expected to define `__slots__` but they don't:\n"
+    cls_list = "\n".join(f"  * {name}" for name in unslotted_classes)
+    url = "https://docs.python.org/3/reference/datamodel.html#slots"
+    hint = f"Hint: See for detail {url!r}"
```
FBruzzesi: Kinda nitpicking:

```suggestion
    hint = f"Hint: For more details see {url!r}"
```

jonasdedden:
Sorry for asking here (I'm also not a maintainer of this library, but just sumbled over this issue):

Instead of manually having to explicitly set `__slots__` everywhere (which introduces maintainance load and introduces the risk of setting them improperly), why not use `typing.NamedTuple` (if immutability is not a problem) or `dataclass` with `slots=True` (if mutability is required)?
There are already tools in the stdlib that do these kinds of things automatically for you.

Sorry again, just a curious question from an outsider!

FBruzzesi:
> Sorry again, just a curious question from an outsider!

Hey @jonded94 - no need to apologize, actually thank you for questioning these kind of choices. We should definitely not dive into a solution without thinking about pros and cons of different approaches.

> which introduces maintainance load and introduces the risk of setting them improperly

I feel you, I tried to propose a metaclass that can lower such burden :) and #3201 is using something towards such direction, so let's see if it makes it in the main branch soon as well.

> why not use `typing.NamedTuple` (if immutability is not a problem) or `dataclass` with `slots=True` (if mutability is required)? There are already tools in the stdlib that do these kinds of things automatically for you.

We probably give it for granted in many places, but the main idea is that we want to have an API which is very close to the polars one. That's not always possible, but I would rather not have a discrepancy such as:

```py
from dataclasses import is_dataclass
from polars import DataType

is_dataclass(DataType)
# False

is_dataclass(DType)
# True
```

or similarly:

```py
from collections import namedtuple
import polars as pl

Int64 = namedtuple('Int64', field_names=())

isinstance(pl.Int64(), tuple), isinstance(Int64(), tuple)
# False, True
```

I hope it makes sense 😇

PR #3198, opened 2025-10-12
feat: Add support for `{Expr,Series}.{ceil,floor}`

### LINKED ISSUE #3163: enh: `ceil` and `floor`
https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.floor.html
https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.ceil.html

If anyone feels like contributing these, I'd be happy to review a PR

m-richards:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3163
## Checklist

- [x] Tests added
- [x] Documented the changes

I've been following narwhals for a while and wanted to better understand how it works / how it's implemented  - so fair warning this is my first time working my way through the narwhals codebase, I might not have gone about everything the right way. I hope this is mostly on the right track, but let me know if not 😄.

### DIFF
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -2073,7 +2073,7 @@ class Series(Generic[IntoSeriesT]):
             >>> import narwhals as nw
             >>> s_native = pa.chunked_array([[1.1, 4.3, -1.3]])
             >>> s = nw.from_native(s_native, series_only=True)
-            >>> s.floor().to_native()  # doctest:+ELLIPSIS
+            >>> s.ceil().to_native()  # doctest:+ELLIPSIS
             <pyarrow.lib.ChunkedArray object at ...>
             [
               [


narwhals/_pandas_like/series.py
```diff
@@ -798,6 +798,50 @@ def tail(self, n: int) -> Self:
     def round(self, decimals: int) -> Self:
         return self._with_native(self.native.round(decimals=decimals))
 
+    def floor(self) -> Self:
+        native = self.native
+        native_cls = type(native)
+        implementation = self._implementation
+        if get_dtype_backend(native.dtype, implementation=implementation) == "pyarrow":
```
m-richards, line 795: I noticed that using np.floor loses the pyarrow datatypes, so I've borrowed this approach from `log`. I supect it would be simpler to just cast the result of np.floor, but I'm inferring that the design philosophy is that it makes sense to use pyarrow compute for pyarrow dtypes?

Happy to collate some of this boilerplate into a shared helper if that makes sense?

I realise as writing this and checking back that the numpy case is going to have issues for cudf too I think.

tests/expr_and_series/floor_ceil_test.py
```diff
@@ -0,0 +1,52 @@
...
+    data = {"a": [1.12345, 2.56789, 3.901234, -0.5]}
+    df_raw = constructor(data)
+    df = nw.from_native(df_raw)
+
+    expected_data = {"a": [1.0, 2.0, 3.0, -1.0]}
+    result_frame = df.select(nw.col("a").floor())
+
+    if "pandas_pyarrow" in str(constructor):
+        pandas_input = df_raw["a"]  # type:ignore[index]
+        assert pandas_input.dtype == result_frame["a"].to_native().dtype
```
m-richards: I'm not sure if there's a better existing idiom for this (checking that the dtype stays a pyarrow double instead of a numpy float64), just that assert equal data won't pick it up.

docs/api-reference/expr.md
FBruzzesi, line 1: Can we keep the list sorted alphabetically?

docs/api-reference/series.md
FBruzzesi, line 1: Same as in `docs/api-reference/expr.md`, let's keep this alphabetically sorted

narwhals/_pandas_like/series.py
FBruzzesi: > I noticed that using np.floor loses the pyarrow datatypes, so I've borrowed this approach from log. I supect it would be simpler to just cast the result of np.floor, but I'm inferring that the design philosophy is that it makes sense to use pyarrow compute for pyarrow dtypes?

Yes that's totally the right assumption.

> Happy to collate some of this boilerplate into a shared helper if that makes sense?

Now that's the same pattern is used 4 times I think it would be nice to abstract into a boilerplate function. I am happy to keep this as a follow up so that implementation details don't drag our attention away from the main PR purpose

> I realise as writing this and checking back that the numpy case is going to have issues for cudf too I think.

Yes that's taken care of by using `array_funcs = self._array_funcs`, which for cudf will be the `cupy` module

narwhals/_polars/expr.py
FBruzzesi, line 1: Also prefer alphabetically sorted

narwhals/_polars/series.py
FBruzzesi, line 1: Also prefer alphabetically sorted

narwhals/series.py
```diff
@@ -2045,6 +2045,46 @@ def round(self, decimals: int = 0) -> Self:
         """
         return self._with_compliant(self._compliant_series.round(decimals))
 
+    def floor(self) -> Self:
+        r"""Compute the numerical floor.
+
+        Examples:
+            >>> import pandas as pd
+            >>> import narwhals as nw
```
FBruzzesi:
```suggestion
            >>> import pyarrow as pa
            >>> import narwhals as nw
```

narwhals/series.py
```diff
@@ -2045,6 +2045,46 @@ def round(self, decimals: int = 0) -> Self:
...
+            ]
+        """
+        return self._with_compliant(self._compliant_series.floor())
+
+    def ceil(self) -> Self:
+        r"""Compute the numerical ceiling.
+
+        Examples:
+            >>> import pandas as pd
+            >>> import narwhals as nw
```
FBruzzesi:
```suggestion
            >>> import pyarrow as pa
            >>> import narwhals as nw
```

tests/expr_and_series/floor_ceil_test.py
FBruzzesi, line 1: Can we split this into two tests? `floor_test.py` and `ceil_test.py`?

Additionally, can we test both for `expr` and `series` cases? Also to address your comment below on how to test pandas dtype, I think that [log_test.py](https://github.com/narwhals-dev/narwhals/blob/main/tests/expr_and_series/log_test.py) has all the boilerplate you need: one test for expr, one for series, and multiple tests for pandas dtypes

docs/api-reference/expr.md
dangotbanned: @FBruzzesi we should probably lint this in ci somehow 🤔

FBruzzesi: Here: #3200

tests/expr_and_series/floor_ceil_test.py
MarcoGorelli: tbh i think it's ok to have them in the same test, like we do for `first_last` and some others

agree on the rest of the comments 👍

MarcoGorelli:
> So we can either flag it as not_implemented or find a workaround (maybe via [dask.array.floor](https://docs.dask.org/en/latest/generated/dask.array.floor.html)?).

`not_implemented` + feature request in dask please 🙏

narwhals/_polars/expr.py
m-richards: makes sense, didn't spot the pattern as I was going through, I was generally trying to keep side by side with round, which I figured was the closest analogue.

tests/expr_and_series/floor_ceil_test.py
m-richards: @FBruzzesi I'm probably missing something here, but I thought I was already was testing both expr and series cases. I copied the test naming convention from test_round - but I see test_log has expr explicitly in the function name. I updated the naming but let me know if there's something else I need to add too.

m-richards:
> > So we can either flag it as not_implemented or find a workaround (maybe via [dask.array.floor](https://docs.dask.org/en/latest/generated/dask.array.floor.html)?).
> 
> `not_implemented` + feature request in dask please 🙏

Ah sorry I missed this, I went through the comments in the diff first and added dask in via dask.array.floor in what I've just pushed 😅. 

Would it be likely to be added to dask given that Series.floor isn't a method in pandas? - I thought dask targeted the pandas api, but I haven't used it directly in quite a while.

Happy to change over to not implemented if that makes more sense.

tests/expr_and_series/floor_ceil_test.py
FBruzzesi: > I'm probably missing something here

No you are not! I definitely was. Sorry for the false alarm 🙈

FBruzzesi:
> Would it be likely to be added to dask given that Series.floor isn't a method in pandas? - I thought dask targeted the pandas api, but I haven't used it directly in quite a while.

That's a very good point! I am honestly quite happy with `dask.array.<op>`

> Happy to change over to not implemented if that makes more sense.

but I will let @MarcoGorelli make the final call here

narwhals/series.py
```diff
@@ -2045,6 +2045,46 @@ def round(self, decimals: int = 0) -> Self:
...
+
+    def ceil(self) -> Self:
+        r"""Compute the numerical ceiling.
+
+        Examples:
+            >>> import pyarrow as pa
+            >>> import narwhals as nw
+            >>> s_native = pa.chunked_array([[1.1, 4.3, -1.3]])
+            >>> s = nw.from_native(s_native, series_only=True)
+            >>> s.floor().to_native()  # doctest:+ELLIPSIS
```
FBruzzesi: Some copy pasting gone wrong :)

```suggestion
            >>> s.ceil().to_native()  # doctest:+ELLIPSIS
```

(spotted via https://github.com/narwhals-dev/narwhals/actions/runs/18460845319/job/52592345769?pr=3198)

dangotbanned:
> > Would it be likely to be added to dask given that Series.floor isn't a method in pandas? - I thought dask targeted the pandas api, but I haven't used it directly in quite a while.
> 
> That's a very good point! I am honestly quite happy with `dask.array.<op>`
> 
> > Happy to change over to not implemented if that makes more sense.
> 
> but I will let @MarcoGorelli make the final call here

I agree with @FBruzzesi 

If we can improve `dask` support by using more of its own API - surely that's a good thing right?

MarcoGorelli:
thanks all! agree with the `dask.array` decision, sorry for having prematurely pushed back (i was expecting it to be more complicated, as things in dask often are 😩 )

but, this is really nice! thanks for having taken care of the details with the pandas dtypes. perhaps making a common utility to be reused between log/floor/ceil could make sense?

PR #3205, opened 2025-10-13
test: Add serde tests for `DType`s

dangotbanned:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Follow-up to #3194
- Ported tests from https://github.com/pola-rs/polars/blob/a143eb0d7077ee9da2ce209a19c21d7f82228081/py-polars/tests/unit/test_serde.py
- Also related (https://github.com/narwhals-dev/narwhals/blob/4980a04927b9879f556aa97b239af2235a86f0fe/narwhals/_plan/serde.py)

## Checklist

- [x] Tests added
- [x] Documented the changes

Starting off fairly small here, trying to work out if we'll need to implement any other dunders mentioned in [Pickling Class Instances](https://docs.python.org/3/library/pickle.html#pickling-class-instances)

I ~~have~~ had a feeling `Enum` with deferred categories might need some work.
*But*, turns out it worked without changes 🥳 (https://github.com/narwhals-dev/narwhals/pull/3205/commits/b2ba74919bd8754c4fbf7aee9c739c799915fe34)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -83,12 +83,12 @@ repos:
       name: don't import from narwhals.dtypes (use `Version.dtypes` instead)
       entry: |
           (?x)
-            import\ narwhals.dtypes|
-            from\ narwhals\ import\ dtypes|
-            from\ narwhals.dtypes\ import\ [^D_]+|
-            import\ narwhals.stable.v1.dtypes|
-            from\ narwhals.stable\.v.\ import\ dtypes|
-            from\ narwhals.stable\.v.\.dtypes\ import
+            import\ narwhals(\.stable\.v\d)?\.dtypes|
+            from\ narwhals(\.stable\.v\d)?\ import\ dtypes|
+            ^from\ narwhals(\.stable\.v\d)?\.dtypes\ import
+              \ (DType,\ )?
+              ((Datetime|Duration|Enum)(,\ )?)+
+              ((,\ )?DType)?
       language: pygrep
       files: ^narwhals/
       exclude: |
--- a/docs/api-reference/testing.md
+++ b/docs/api-reference/testing.md
@@ -1,0 +1,7 @@
+# `narwhals.testing`
+
+::: narwhals.testing
+    handler: python
+    options:
+      members:
+        - assert_series_equal
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -69,6 +69,7 @@ nav:
     - api-reference/dtypes.md
     - api-reference/exceptions.md
     - api-reference/selectors.md
+    - api-reference/testing.md
     - api-reference/typing.md
     - api-reference/utils.md
   - This: this.md
--- a/narwhals/testing/__init__.py
+++ b/narwhals/testing/__init__.py
@@ -1,0 +1,5 @@
+from __future__ import annotations
+
+from narwhals.testing.asserts.series import assert_series_equal
+
+__all__ = ("assert_series_equal",)
--- a/narwhals/testing/asserts/series.py
+++ b/narwhals/testing/asserts/series.py
@@ -1,0 +1,291 @@
+from __future__ import annotations
+
+from functools import partial
+from typing import TYPE_CHECKING, Any, Callable
+
+from narwhals._utils import qualified_type_name, zip_strict
+from narwhals.dependencies import is_narwhals_series
+from narwhals.dtypes import Array, Boolean, Categorical, List, String, Struct
+from narwhals.functions import new_series
+from narwhals.testing.asserts.utils import raise_series_assertion_error
+
+if TYPE_CHECKING:
+    from typing_extensions import TypeAlias
+
+    from narwhals.series import Series
+    from narwhals.typing import IntoSeriesT, SeriesT
+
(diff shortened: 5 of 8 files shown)


dangotbanned:
I'm pleasantly suprised this all worked without changing a single line of (non-test) code 😅 

I've even tried [protocol 3](https://docs.python.org/3/library/pickle.html#object.__reduce_ex__) locally and still no issues.

Can someone burst my bubble and break this? 🙏

tests/serde_test.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+@namespaces
+@time_units
+def test_serde_duration_dtype(
+    namespace: DTypes, time_unit: TimeUnit, roundtrip: Identity
+) -> None:
+    dtype = namespace.Duration(time_unit)
+    result = roundtrip(dtype)
+    assert result == namespace.Duration(time_unit)
```
FBruzzesi: Can't we simply 

```suggestion
    dtype = namespace.Duration(time_unit)
    result = roundtrip(dtype)
    assert result == dtype
```
(and everywhere else)?

They are not the same object (`result is dtype` -> `False`), but it makes it easier especially for nested types if we ever want to change something/add oher cases and parametrize

dangotbanned: Tbh yeah I do see what you mean.
There also seems to be a mix of both in `polars`:
- https://github.com/pola-rs/polars/blob/a143eb0d7077ee9da2ce209a19c21d7f82228081/py-polars/tests/unit/test_serde.py


I guess the way it is written *currently*, it reads to me like we're testing that what came out of the pickle is the same as starting all over again.

I don't feel too strongly towards keeping it as-is vs changing for a quicker check.

I'll give that whirl tomorrow 🙂

tests/serde_test.py
```diff
@@ -0,0 +1,154 @@
...
+    return fn
+
+
+@pytest.fixture(
+    params=[_roundtrip_pickle(), _roundtrip_pickle(4), _roundtrip_pickle(5)],
+    ids=["pickle-None", "pickle-4", "pickle-5"],
+)
+def roundtrip(request: pytest.FixtureRequest) -> Identity:
+    fn: Identity = request.param
+    return fn
```
dangotbanned, line 53: The idea I had here is we can pretty easily extend this by adding more functions that can roundtrip.

`pickle` is all that I've used, but the [protocols which are being called](https://docs.python.org/3/library/pickle.html#pickling-class-instances) can be used elsewhere

```py
import narwhals as nw

dtype = nw.Struct({"a": nw.List(nw.Array(nw.String, 5))})

>>> dtype.__reduce_ex__(5)
(<function copyreg.__newobj__(cls, *args)>,
 (narwhals.dtypes.Struct,),
 (None,
  {'fields': [Field('a', List(Array(<class 'narwhals.dtypes.String'>, shape=(5,))))]}),
 None,
 None)


>>> dtype.__getstate__()
(None,
 {'fields': [Field('a', List(Array(<class 'narwhals.dtypes.String'>, shape=(5,))))]})
```

tests/serde_test.py
FBruzzesi: Don't stress about it. Not a blocker from me. I think the advantage would be if we want to parametrize more tests (e.g. with inner types)

tests/serde_test.py
dangotbanned: Barely on topic, but noticable from the example I used

@FBruzzesi nested `DType`s are where the [metaclass repr](https://github.com/narwhals-dev/narwhals/pull/2969#discussion_r2266125855) from polars would be nice

```py
import narwhals as nw
>>> nw.Struct({"a": nw.List(nw.Array(nw.String, 5))})
Struct({'a': List(Array(<class 'narwhals.dtypes.String'>, shape=(5,)))})
```

```py
import polars as pl
>>> pl.Struct({"a": pl.List(pl.Array(pl.String, 5))})
Struct({'a': List(Array(String, shape=(5,)))})
```

FBruzzesi (approved):
@dangotbanned if you are happy with the state of this, I think we can merge these changes for now and iterate on top of them if needed.

Quoting my previous self https://github.com/narwhals-dev/narwhals/pull/3205#pullrequestreview-3333002233

> I know close to nothing about custom pickling

dangotbanned:
> @dangotbanned if you are happy with the state of this, I think we can merge these changes for now and iterate on top of them if needed.

Thanks @FBruzzesi!

> Quoting my previous self [#3205 (review)](https://github.com/narwhals-dev/narwhals/pull/3205#pullrequestreview-3333002233)
> 
> > I know close to nothing about custom pickling


That shouldn't be an issue, since we're utilizing the default picking ☺️

[`object.__getstate__`](https://docs.python.org/3/library/pickle.html#object.__getstate__)

> - For a class that has `__slots__` and no instance `__dict__`, the default state is a `tuple` whose first item is `None` and whose second item is a dictionary mapping slot names to slot values described in the previous bullet.

PR #3189, opened 2025-10-09
fix(DONT MERGE): `duckdb>=1.4.1` typing & warnings

dangotbanned:
Will close #3188

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues
- [ ] https://github.com/duckdb/duckdb-python/discussions/124
- [x] #3188
- [x] #3137
- [x] https://github.com/eakmanrq/sqlframe/issues/531
- [x] https://github.com/akmalsoliev/Validoopsie/issues/30

## If you have comments or can explain your changes, please do so below

> [!NOTE]
> Typing is *"fixed"*
> Still need to open some issues/PRs upstream in `duckdb` (https://github.com/narwhals-dev/narwhals/pull/3189#issuecomment-3393279293)

PR #3207, opened 2025-10-14
fix: Always accept `{Expr,Series}.is_in(other: Iterable)` 

### LINKED ISSUE #3195: `Expr.is_in(Iterable)` raises inconsistently
## Related
(TODO - move write-up here)
- https://github.com/narwhals-dev/narwhals/pull/3189#discussion_r2422717399
- https://github.com/narwhals-dev/narwhals/pull/3189#discussion_r2422721481

## Description
We document that `Expr.is_in` supports `Iterable`, and check it at runtime.

https://github.com/narwhals-dev/narwhals/blob/ebb2a40bab2f18b485512707ffb8cd440bb17172/narwhals/expr.py#L971-L975

But in #3189 I found that our tests only cover `list`. 

Kinda suprised by which backends do/don't work.
Fixing it for all of them is any easy task - just something like this really:

https://github.com/narwhals-dev/narwhals/blob/64f6b4f4017d0aca260386e33cab0175d1cd8a41/narwhals/_duckdb/expr.py#L260-L262

## Repro

```py
import narwhals as nw

data = {"a": [1, 4, 2, 5]}
df = nw.from_dict(data, backend="polars")
sequence = 4, 2
other = sequence
>>> df.select(nw.col("a").is_in(other))
```

```py
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|  shape: (4, 1)   |
|  ┌───────┐       |
|  │ a     │       |
|  │ ---   │       |
|  │ bool  │       |
|  ╞═══════╡       |
|  │ false │       |
|  │ true  │       |
|  │ true  │       |
|  │ false │       |
|  └───────┘       |
└──────────────────┘
```

### Eager

```py
other = iter(sequence)
>>> df.select(nw.col("a").is_in(other))
````

```py
TypeError: cannot create expression literal for value of type tuple_iterator.

```

```py
expr = nw.col("a").is_in(iter(sequence))
>>> nw.from_dict(data, backend="pandas").select(expr)
````

```py
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|            a     |
|     0  False     |
|     1   True     |
|     2   True     |
|     3  False     |
└──────────────────┘
```

```py
expr = nw.col("a").is_in(iter(sequence))
>>> nw.from_dict(data, backend="pyarrow").select(expr)
````

```py
┌────────────────────────────┐
|     Narwhals DataFrame     |
|----------------------------|
|pyarrow.Table               |
|a: bool                     |
|----                        |
|a: [[false,true,true,false]]|
```

### Lazy

```py
df = nw.from_dict(data, backend="polars")
expr = nw.col("a").is_in(iter(sequence))
>>>df.lazy("ibis").select(expr).collect("polars")
```

```py
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|  shape: (4, 1)   |
|  ┌───────┐       |
|  │ a     │       |
|  │ ---   │       |
|  │ bool  │       |
|  ╞═══════╡       |
|  │ false │       |
|  │ true  │       |
|  │ true  │       |
|  │ false │       |
|  └───────┘       |
└──────────────────┘
```

```py
df = nw.from_dict(data, backend="polars")
expr = nw.col("a").is_in(iter(sequence))
>>>df.lazy("duckdb").select(expr).collect("polars")
```

```py
NotImplementedException: Not implemented Error: Unable to transform python value of type '<class 'tuple_iterator'>' to DuckDB LogicalType
```

```py

df = nw.from_dict(data, backend="polars")
expr = nw.col("a").is_in(iter(sequence))
>>> df.lazy("sqlframe", session=DuckDBSession()).select(expr).collect("polars")
```

```py
ValueError: Cannot convert <tuple_iterator object at 0x000001DEC8260970>
```

```py
df = nw.from_dict(data, backend="polars")
expr = nw.col("a").is_in(iter(sequence))
>>> df.lazy("dask").select(expr).collect("polars")
```

```py
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|  shape: (4, 1)   |
|  ┌───────┐       |
|  │ a     │       |
|  │ ---   │       |
|  │ bool  │       |
|  ╞═══════╡       |
|  │ false │       |
|  │ true  │       |
|  │ true  │       |
|  │ false │       |
|  └───────┘       |
└──────────────────┘
```

dangotbanned:
Will close #3195

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/3189#discussion_r2422717399
- https://github.com/narwhals-dev/narwhals/pull/2933#issuecomment-3168376214

## Checklist

- [x] Tests added
- [ ] Documented the changes

I'm expecting most of the diff to be from this commit:
- [test: Repurpose `Series.from_iterable` constructors](https://github.com/narwhals-dev/narwhals/pull/3207/commits/234a09121ca4a89d914742664b389dd056c22428)

The fixing per-backend should only be a couple lines or handling it in `nw.Expr`

- [x] Get tests passing
- [x] Update annotations
- [ ] Do another pass on `iterable_to_sequence`
- [ ] Get a decision on [`Expr.is_in(other: Series)`](https://github.com/narwhals-dev/narwhals/pull/3207#discussion_r2430089632)


tests/expr_and_series/is_in_test.py
```diff
@@ -50,3 +56,40 @@ def test_filter_is_in_with_series(constructor_eager: ConstructorEager) -> None:
...
+) -> None:
+    test_name = request.node.name
+    # NOTE: This *could* be supported by using `ExtensionArray.tolist` (same path as numpy)
+    request.applymarker(
+        pytest.mark.xfail(
+            ("polars" in test_name and "pandas" in test_name and "array" in test_name),
+            raises=TypeError,
+            reason="Polars doesn't support `pd.array`.\nhttps://github.com/pola-rs/polars/issues/22757",
+        )
+    )
```
dangotbanned, line 121: Can confirm that works for the `Expr` (polars) path

However, there seems to be an old version of `duckdb` that doesn't like it 
([failure](https://github.com/narwhals-dev/narwhals/actions/runs/18500501268/job/52715724355?pr=3207))

```py
# pandas==1.1.5, duckdb==1.2.0
duckdb.duckdb.NotImplementedException: 
Not implemented Error: 
Unable to transform python value of type '<class 'numpy.int64'>' to DuckDB LogicalType
```

### Update
Resolved in (https://github.com/narwhals-dev/narwhals/pull/3207/commits/0c66432e30ebf35f456e54528e74a0f21d8eec85)

narwhals/_utils.py
```diff
@@ -2120,3 +2122,58 @@ def extend_bool(
...
     """
     return (value,) * n_match if isinstance(value, bool) else tuple(value)
+
+
+class _CanTo_List(Protocol):  # noqa: N801
+    def to_list(self, *args: Any, **kwds: Any) -> list[Any]: ...
+
+
+class _CanToList(Protocol):
+    def tolist(self, *args: Any, **kwds: Any) -> list[Any]: ...
```
dangotbanned, line 2132: FWIW I hate this as well 😂

tests/conftest.py
```diff
@@ -317,7 +329,109 @@ def eager_backend(request: pytest.FixtureRequest) -> EagerAllowed:
...
+"""`into_iter` fixtures use the suffix `_<n>` to denote the maximum number of constructors.
+
+Anything greater than **10** may return less depending on available dependencies.
+"""
+
+
+@pytest.fixture(params=_into_iter(16), scope="session", ids=_ids_into_iter)
+def into_iter_16(request: pytest.FixtureRequest) -> IntoIterable:
+    function: IntoIterable = request.param
+    return function
```
dangotbanned, line 492: Kinda outdated now since (https://github.com/narwhals-dev/narwhals/pull/3207/commits/633a06d67835e150a071a22dccb316a601492e85)

narwhals/_utils.py
```diff
@@ -2120,3 +2122,58 @@ def extend_bool(
...
+    return is_numpy_array_1d(obj) or _hasattr_static(obj, "tolist")
+
+
+def can_to_pylist(obj: Any) -> TypeIs[_CanTo_PyList]:
+    return (
+        (pa := get_pyarrow()) and isinstance(obj, (pa.Array, pa.ChunkedArray))
+    ) or _hasattr_static(obj, "to_pylist")
+
+
+# TODO @dangotbanned: Use in `{Expr,Series}.is_in`
```
dangotbanned: TO~~DO~~**DONE**
```suggestion
```

narwhals/expr.py
```diff
@@ -991,10 +996,9 @@ def is_in(self, other: Any) -> Self:
             └──────────────────┘
         """
         if isinstance(other, Iterable) and not isinstance(other, (str, bytes)):
+            other = other.to_native() if is_series(other) else iterable_to_sequence(other)
             return self._with_elementwise(
-                lambda plx: self._to_compliant_expr(plx).is_in(
-                    to_native(other, pass_through=True)
-                )
+                lambda plx: self._to_compliant_expr(plx).is_in(other)
```
dangotbanned: I've updated this but it is still wrong. 
Didn't notice the issue until I started trying to add typing to the compliant-level

https://github.com/narwhals-dev/narwhals/blob/849326f7682ac741b3bad4edca775cec79fe11f3/narwhals/_compliant/column.py#L173

I don't understand why we were allowing **any** kind of `Native*` to be passed to **every** backend? 🤔 

Gonna add another test for at least the more common case of a `Series` from the wrong backend

### Updated
[test: Add `test_expr_is_in_series_wrong_backend`](https://github.com/narwhals-dev/narwhals/pull/3207/commits/a6921ea6256879485eb566e78ddee9db6ec967db)

I'm not sure of a safe way to keep the same behavior (if it is desired).
Since we don't know the backend at this stage, the options I see are:
1. Unconditionally convert to `list | tuple`
2. Raise *elsewhere* when a `Series` is passed to a lazy backend
3. Disallow `Expr.is_in(other: Series)`
4. Do nothing

dangotbanned: @MarcoGorelli @FBruzzesi
do you guys have any preference on a path forward here?

This case is a bit different to some of the other places we'd reject `nw.Series` for lazy backends - since the length isn't an issue.
But the safer option of converting all `nw.Series` is gonna be less performant than the currently unsafe version (which only works on a matching eager implementation)

Everything seems like a tradeoff to me 😔

narwhals/_utils.py
dangotbanned: Maybe a better idea would be to ...

Rename `_CanTo_List` -> `ToList`, and move to `_translate.py` alongside:

https://github.com/narwhals-dev/narwhals/blob/f05011cf917c046dfed8664fd188715b1f4b6201/narwhals/_translate.py#L124-L125

Move these as well, but rename to reflect they their naming originates from `numpy` and `pyarrow` (respectively):

https://github.com/narwhals-dev/narwhals/blob/0c66432e30ebf35f456e54528e74a0f21d8eec85/narwhals/_utils.py#L2131-L2132

https://github.com/narwhals-dev/narwhals/blob/0c66432e30ebf35f456e54528e74a0f21d8eec85/narwhals/_utils.py#L2135-L2136

The names of the guards can still stay the same, since their implementations will (*after updating the protocol names*) the link between origin, protocol, method name:

https://github.com/narwhals-dev/narwhals/blob/0c66432e30ebf35f456e54528e74a0f21d8eec85/narwhals/_utils.py#L2139-L2144
https://github.com/narwhals-dev/narwhals/blob/0c66432e30ebf35f456e54528e74a0f21d8eec85/narwhals/_utils.py#L2147-L2148
https://github.com/narwhals-dev/narwhals/blob/0c66432e30ebf35f456e54528e74a0f21d8eec85/narwhals/_utils.py#L2151-L2154

narwhals/expr.py
dangotbanned: I've moved out of draft for visibility (https://github.com/narwhals-dev/narwhals/pull/3207#event-20359344046)

FBruzzesi: Hey @dangotbanned I will try to take a look during the weekend, but I am not sure I can manage.

My thoughts from the context in the messages here (I didn't open the code changes just yet):

- Realistically I don't think it's _too_ common to pass one series from a different backend but we should not exclude such possibility
- My opinion would be to:
  - Use the native series if `isinstance(other, Series) and expr._implementation == other._implementation)`
  - Otherwise convert it to a list (we can do that with native methods thankfully), but warn that such conversion is happening with a `UserWarning`. This case includes a series passed to a lazy backend

**Update**: Regarding _native_ series, then I would prefer to raise an exception in such case

narwhals/_utils.py
FBruzzesi: I updated my comment (https://github.com/narwhals-dev/narwhals/pull/3207#discussion_r2442275362) - I would argue that native series are _not_ ok, while numpy 1d arrays are.

My thought process for this is that if someone is doing something along the lines of:

```py
import narwhals as nw
import polars as pl

def agnostic_func(frame: IntoDataFrameT) -> IntoDataFrameT:
    other = pl.Series([1, 2, 3])  # <- notice how this is a native series!!!
    return nw.from_native(frame).filter(nw.col("x").is_in(other)).to_native()
```

then the function is clearly not agnostic and polars would be required in this case.

A different case would be if a narwhals series with a different backend is provided. This could mean that the function is agnostic but a user is "mixin" backends:

```py
def is_left_in_right(left_series: IntoSeriesT, right_series: IntoSeriesT) -> IntoSeriesT:
    left_nw = nw.from_native(left_series, series_only=True)
    right_nw = nw.from_native(right_series, series_only=True)
    return left_nw.is_in(right_nw).to_native()

# but now it a user to mix it up, not the library itself

is_left_in_right(pl.Series([1,2,3]), pd.Series([0, 1]))
```

This is the case I suggested to yell at the user with a warning.

---

From our side, I think it would greatly simplify (read as, _get rid of_) most of the protocols here, the type guards as well as `iterable_to_sequence` function.

PR #3213, opened 2025-10-15
chore: Add `DTypeClass` to improve `DType.__repr__` and simplify `__slots__`

### LINKED ISSUE #3212: Improve `DType.__repr__` and simplify `__slots__` maintenance via metaclasses
In the last few PRs focused on improving dtypes there were a couple of comments that stud out as follow-ups:

1. Improve dtypes `__repr__` (see https://github.com/narwhals-dev/narwhals/pull/3205#discussion_r2428723032)
2. Simplify `__slots__` maintenance (see #3201 and https://github.com/narwhals-dev/narwhals/pull/3194#issuecomment-3396669600)

Both these improvements can be achieved via a single metaclass.

I went ahead and started a branch ([chore/dtype-meta](https://github.com/narwhals-dev/narwhals/tree/chore/dtype-meta)) that mostly takes care of both, but it fails on typing (pyright is happy, mypy is not).

I will open a draft PR, but I am happy to keep this issue to discuss alternatives or reframe it if we need additional features

FBruzzesi:

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3212

## Checklist

- [x] Tests added
- [x] Documented the changes


tests/dtypes/dtypes_test.py
```diff
@@ -493,6 +493,17 @@ def test_enum_hash() -> None:
...
 
+@pytest.mark.parametrize("dtype_name", ["Datetime", "Duration", "Enum"])
+def test_dtype_repr_versioned(dtype_name: str) -> None:
+    from narwhals.stable import v1 as nw_v1
+
+    dtype_class_main = getattr(nw, dtype_name)
+    dtype_class_v1 = getattr(nw_v1, dtype_name)
+
+    assert dtype_class_main is not dtype_class_v1
+    assert repr(dtype_class_main) != repr(dtype_class_v1)
```
dangotbanned, line 533: I pushed this (failing) test as I've been thinking about *if* we want to distinguish here?

FBruzzesi: I think it could make sense to distinguish between the two...buuuut do we have a valid proposal? There is none I can think of that I _really_ like to be honest with you 🤔

dangotbanned: ### Detour

I experimented with this a little bit in #2571

https://github.com/narwhals-dev/narwhals/blob/1ab15991f6128cc693ceaa4058c6e88d262fe19a/narwhals/_plan/expr.py#L56-L61

Which looks like:

```py
import narwhals._plan as nwp

nwp.col("a").drop_nulls().alias("b")
```

```py
nw._plan.Expr(main):
col('a').drop_nulls().alias('b')
```

But I'd forgotten I did that, since `_repr_html_` overrides that:

image

dangotbanned: ## Proposal

But since we only have 3x `DType`s, maybe something like:

```py
import narwhals as nw


def repr_dtype_class(dtype: type[DType], version: Version) -> str:
    v = f"[{version.name.lower()}]" if version is not Version.MAIN else ""
    return f"{dtype.__name__}{v}"


repr_dtype_class(nw.Datetime, Version.MAIN), repr_dtype_class(nw.Datetime, Version.V1)
```

```py
('Datetime', 'Datetime[v1]')
```

> [!NOTE]
> I'm open to changes, this was just the first idea I thought of

FBruzzesi: > ('Datetime', 'Datetime[v1]')

I kinda like this - Should this differentiation apply only to _new_ types or all of them when calling from a non-main version?

@MarcoGorelli gentle summoning to get your thoughts as well here

dangotbanned: > Should this differentiation apply only to new types or all of them when calling from a non-main version?

Ah yeah that is a quirk I suppose

I used `Version` in (https://github.com/narwhals-dev/narwhals/pull/3213#discussion_r2437367651) because it was how I'd done it before, but I suppose we haven't defined it on `DType`s.

And everything besides these guys are just re-exports of main:

```py
from narwhals.stable import v1

v1.Datetime
Datetime[v1]

v1.Duration
Duration[v1]

v1.Enum
Enum[v1]
```

I think the most important part is being able to tell the difference between `Enum` and `v1.Enum`, since only one has a constructor.

For everything else, I can see the argument for consistency - but we would be adding noise for things like:

```py
import narwhals as nw
from narwhals.stable import v2

nw.List
List

v1.List
List[v1]

v2.List
List[v2]
```

Or if we extended the idea to `DType` instances (totally optional):

```py
nw.List(nw.String)
List(String)

v1.List(v1.String)
List[v1](String[v1])

v2.List(v2.String)
List[v2](String[v2])
```

narwhals/dtypes.py
```diff
@@ -59,10 +60,40 @@ def _validate_into_dtype(dtype: Any) -> None:
...
 
 
-class DType:
-    __slots__ = ()
+class DTypeClass(type):
+    """Metaclass for DType classes.
 
-    def __repr__(self) -> str:  # pragma: no cover
+    * Nicely print classes.
+    * Ensure [`__slots__`] are always defined to prevent `__dict__` creation (empty by default).
```
dangotbanned: Nit: 

This would match the syntax from https://github.com/narwhals-dev/narwhals/blob/b6c900edaee4dc9c4ebeb1a3f483062b6b5278d6/narwhals/typing.py
```suggestion
    - Nicely print classes.
    - Ensure [`__slots__`] are always defined to prevent `__dict__` creation (empty by default).
```

dangotbanned (approved):
Thanks @FBruzzesi, I've been wanting this for a while 🥳 

(https://github.com/narwhals-dev/narwhals/pull/3213#discussion_r2437271987) isn't blocking for me


cat-ptain on a narwhal

tests/dtypes/dtypes_test.py
FBruzzesi: @dangotbanned what should we do about this? I have been hesitant to merge because of it 😂

dangotbanned: Ffs I've been waiting for your response 😭

dangotbanned: But I also did say this wasn't blocking 😉

https://github.com/narwhals-dev/narwhals/pull/3213#pullrequestreview-3351154940

FBruzzesi: > Ffs I've been waiting for your response 😭

ahahah yes my bad! 

> But I also did say this wasn't blocking 😉

I know but it still feels a bit incomplete

dangotbanned: > I know but it still feels a bit incomplete

@FBruzzesi Just merge this already man 😭

FBruzzesi: Ok, I am pressing the button now

PR #3223, opened 2025-10-18
fix: Allow pyarrow extension dtypes (mapped to `nw.Unknown`)

### LINKED ISSUE #3222: [Bug]: pyarrow `native_to_narwhals_dtype` fails with extension dtype (since non-hashable)
### Describe the bug

I spotted this thank to working interactively in marimo.

I am working with a [`geopandas.GeoDataFrame`](https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoDataFrame.html). When converting to a pyarrow table, then the geometry dtype is converted into a binary and everything works as expected.

Now if I want to use the [`geoarrow.pyarrow`](https://geoarrow.org/geoarrow-python/main/index.html), then the same `.to_arrow()` operation fails since `geoarrow.pyarrow` makes geometry into an extension dtype, which is non-hashable and computing the schema fails due to the lru cache.

I have no idea why a [pyarrow.ExtensionType](https://arrow.apache.org/docs/python/generated/pyarrow.ExtensionType.html) is non-hashable by default.

_Extending_ from the example in the `pyarrow.ExtensionType` docs:

**repro.py**
```py
import narwhals as nw
import pyarrow as pa

class RationalType(pa.ExtensionType):
    def __init__(self, data_type: pa.DataType):
        super().__init__(
            pa.struct(
                [
                    ("numer", data_type),
                    ("denom", data_type),
                ],
            ),
            "my_package.rational",
        )
    def __arrow_ext_serialize__(self) -> bytes:
        return b""

    @classmethod
    def __arrow_ext_deserialize__(cls, storage_type, serialized):
        return RationalType(storage_type[0].type)
    
pa.register_extension_type(RationalType(pa.int64()))
rational_type = RationalType(pa.int32())

storage_array = pa.array(
    [
        {"numer": 10, "denom": 17},
        {"numer": 20, "denom": 13},
    ],
    type=rational_type.storage_type
)
rational_array = rational_type.wrap_array(storage_array)

# This is the new bit
table = pa.table({"rational": rational_array})
nw.from_native(table).schema
```

```terminal
$ python repro.py

python repro.py
  File "/Users/francesco.bruzzesi/Desktop/oss/narwhals-dev/repro.py", line 41, in <module>
    nw.from_native(table).schema
  File "/Users/francesco.bruzzesi/Desktop/oss/narwhals-dev/narwhals/dataframe.py", line 1327, in schema
    return super().schema
           ^^^^^^^^^^^^^^
  File "/Users/francesco.bruzzesi/Desktop/oss/narwhals-dev/narwhals/dataframe.py", line 177, in schema
    return Schema(self._compliant_frame.schema.items())
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/francesco.bruzzesi/Desktop/oss/narwhals-dev/narwhals/_arrow/dataframe.py", line 310, in schema
    field.name: native_to_narwhals_dtype(field.type, self._version)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'RationalType'
```

---

Now, if we were to add:

```py
    def __hash__(self):
        return hash((self.__class__.__name__, self.storage_type))
```

in the extension dtype, then everything works as expected. One follow up could be to flag it in the `geoarrow` project, but I think we _should_  deal with this.

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3222

## Checklist

- [x] Tests added
- [x] Documented the changes

Notes:

- We can't round trip back to the original dtype since in all cases we end up mapping to `nw.Unknown`
- For pyarrow we had to split the `native_to_narwhals_dtype` into two parts: first check if we receive an extension dtype, only then use the original mapping function that has a `lru_cache`
- I added a test for pandas, but as you can see there are no changes in the code as `__hash__` is taken care of from pandas side

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -52,13 +52,13 @@ jobs:
           cache-suffix: pytest-windows-${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: install-reqs
-        # we are not testing pyspark on Windows here because it is very slow
-        run: uv pip install -e ".[dask, modin, ibis]" --group core-tests --group extra --system
+        # we are not testing pyspark, modin, or dask on Windows here because nobody got time for that
+        run: uv pip install -e ".[ibis]" --group core-tests --group extra --system
       - name: show-deps
         run: uv pip freeze
       - name: Run pytest
         run: |
-          pytest tests --cov=narwhals --cov=tests --runslow --cov-fail-under=95 --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,modin[pyarrow],polars[eager],polars[lazy],dask,duckdb,sqlframe,ibis --durations=30
+          pytest tests --cov=narwhals --cov=tests --runslow --cov-fail-under=95 --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe,ibis --durations=30
 
   pytest-full-coverage:
     strategy:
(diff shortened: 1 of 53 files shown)


MarcoGorelli:
Thanks! I haven't checked the code yet but am on board with Unknown here

PR #3224, opened 2025-10-18
feat(expr-ir): Support `over(*partition_by)`

dangotbanned:
## Related issues

- Child of #2572
- Follows #3173

> [!IMPORTANT]
> The bulk of the diff is from this child PR (#3233), which has a more detailed summary

<details><summary>Show tasks</summary>
<p>

- [x] Get all original tests passing
- [x] Add [`DataFrame.partition_by`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.partition_by.html)
  - The new internals required (for `pyarrow`) to support more kinds of `over(*partition_by)` are a superset of what's needed for it
- [x] #3233
  - [x] Selectors work that started here
    - [x] [fix(expr-ir): Ensure nested nodes expand correctly](https://github.com/narwhals-dev/narwhals/pull/3224/commits/c4d4030fc6fea84885ac89355d8bbb51c5755400)
    - [x] Fix `by_dtype` naive hash
      - [feat(expr-ir): Add dedicated selectors for problem children](https://github.com/narwhals-dev/narwhals/pull/3224/commits/88dfdbc6e8f219c3cdfe59321daf18d6e0562000)
      - [fix(expr-ir): Ensure `by_dtype` handles bare parametric types](https://github.com/narwhals-dev/narwhals/pull/3224/commits/ecde261d799a711c2e0a7acf11b108bc45035dc9)
- [x] Adapting `DataFrame.partition_by`
  - [x] Reuse for `Group_by.__iter__`
  - [x] Reach a dead end for the other bits (see [thread](https://github.com/narwhals-dev/narwhals/pull/3224#discussion_r2500502941))

</p>
</details>

## Whats new?
- `Expr.over(*partition_by: OneOrIterable[IntoExprColumn])`
- `DataFrame.partition_by(*by: OneOrIterable[ColumnNameOrSelector])`
- `ArrowExpr.null_count().over(*partitions))` (and in `group_by.agg(...)`)
- `df.with_columns(nwp.all().sum().over("a").name.suffix("_sum"))`

## Whats implemented?
*Already in `main`, but now here too*
- `ArrowExpr`
  - `shift`
  - `diff`
  - `cum_count`
  - `cum_min`
  - `cum_max`
  - `cum_prod`
  - `cum_sum`
  - `null_count`

### DIFF
--- a/narwhals/_plan/arrow/functions.py
+++ b/narwhals/_plan/arrow/functions.py
@@ -92,19 +92,24 @@ def modulus(lhs: Any, rhs: Any) -> Any:
     return sub(lhs, multiply(floor_div, rhs))
 
 
+# TODO @dangotbanned: Somehow fix the typing on this
+# - `_ArrowDispatch` is relying on the gradual typing
 _DISPATCH_BINARY: Mapping[type[ops.Operator], BinOp] = {
+    # BinaryComp
     ops.Eq: eq,
     ops.NotEq: not_eq,
     ops.Lt: lt,
     ops.LtEq: lt_eq,
     ops.Gt: gt,
     ops.GtEq: gt_eq,
-    ops.Add: add,
-    ops.Sub: sub,
-    ops.Multiply: multiply,
-    ops.TrueDivide: truediv,
-    ops.FloorDivide: floordiv,
-    ops.Modulus: modulus,
+    # BinaryFunction (well it should be)
+    ops.Add: add,  # BinaryNumericTemporal
+    ops.Sub: sub,  # pyarrow-stubs
+    ops.Multiply: multiply,  # pyarrow-stubs
+    ops.TrueDivide: truediv,  # [[Any, Any], Any]
+    ops.FloorDivide: floordiv,  # [[ArrayOrScalar, ArrayOrScalar], Any]
+    ops.Modulus: modulus,  # [[Any, Any], Any]
+    # BinaryLogical
     ops.And: and_,
     ops.Or: or_,
     ops.ExclusiveOr: xor,


narwhals/_plan/arrow/group_by.py
```diff
@@ -178,3 +167,59 @@ def agg(self, irs: Seq[NamedIR]) -> Frame:
...
+) -> Iterator[pa.Table]:
+    """Optimized path for single-column partition."""
+    arr_dict: Incomplete = fn.array(native.column(by).dictionary_encode("encode"))
+    indices: pa.Int32Array = arr_dict.indices
+    if not include_key:
+        native = native.remove_column(native.schema.get_field_index(by))
+    for idx in range(len(arr_dict.dictionary)):
+        # NOTE: Acero filter doesn't support `null_selection_behavior="emit_null"`
+        # Is there any reasonable way to do this in Acero?
+        yield native.filter(pc.equal(pa.scalar(idx), indices))
```
MarcoGorelli, line 202: is this for use in `over(partition_by=...)`?

if so, just as a heads up, we won't be able to accept a solution which involves looping over partitions in python

dangotbanned: Oh hi Marco, fancy seeing you here 😄

> is this for use in `over(partition_by=...)`?

No this part is just for `DataFrame.partition_by("one_column")`.
The multi-column variant of that *is* run in the c++ engine.

For `over(partition_by=...)` I need the partitions put back together and only for a single column vs a full table here.
These are different enough that I'd more likely use `union` (https://github.com/narwhals-dev/narwhals/pull/3224/commits/41d8cc2c40b8732f91b3fd194ed6afa90a4e938c) and keeping things threaded will probably make more sense.
I'm also thinking ahead for `over(*partition_by, order_by=...)` which will be inserting one of [these nodes](https://github.com/narwhals-dev/narwhals/blob/41d8cc2c40b8732f91b3fd194ed6afa90a4e938c/narwhals/_plan/arrow/acero.py#L258-L267) into the plan

**Edit**: I've just updated the description to try and be clearer that these are *related* problems - but only in a conceptual sense


---

But to clarify

> we won't be able to accept a solution which involves looping over partitions in python

This isn't looping over partitions, it is looping over an index into the partition key.
The index is being used to create a mask from indices - which AFAICT isn't expensive.
At the very least, I'm expecting this to be cheaper than what `ArrowGroupBy.__iter__` **currently does**. Which involves casting + concatenating columns, before filtering on the resulting keys in a python loop


But the note ...

> Is there any reasonable way to do this in Acero?

... is me leaving myself a TODO to try and benchmark that later 😅

dangotbanned: @MarcoGorelli 

long-awaited (maybe?) follow-up to this rabbit hole...

I *thought* I saw a path to being able to *finally* allow:

<details><summary>Show <code>test_group_by_complex_polars</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/60c00eae656e49074bc414f1f9e305b02373ddc5/tests/frame/group_by_test.py#L54-L63


</p>
</details> 

```py
df.group_by("a").agg((nw.col("b") - nw.col("c").mean()).mean())
```

## Can this be translated to Acero?
Well, yeah kinda.

<details><summary>Detecting this kind of <code>Expr</code></summary>
<p>

This part is easy 🥳 

```py
def is_decomposable_aggregation(expr: ir.ExprIR) -> TypeIs[ir.AggExpr]:
    """Defined by using a subset of expressions, ending with an aggregation.

    Okay:

        (col("a") + col("b")).min()
        (col("a").max() + col("b")).mean()
        (col("a").count() + col("b").len()).sum()

    Not okay:

        col("a") + col("b")
        (col("a") + col("b").mean())
        (col("a").min() + col("b"))

    Maybe should be allowed, but not covered by this rule yet:

        col("a").min() + col("b").max()
    """
    return isinstance(expr, ir.AggExpr) and all(
        isinstance(e, (ir.Column, ir.AggExpr, ir.BinaryExpr)) for e in expr.iter_left()
    )
```

</p>
</details> 

This version includes the nested right-hand-side `mean()`:

<details><summary>Version 1</summary>
<p>

The tricky part is that [projections](https://arrow.apache.org/docs/cpp/acero/user_guide.html#project) and [aggregations](https://arrow.apache.org/docs/cpp/acero/user_guide.html#aggregate) simply don't mix.
Here I tried to work around that with a join - but the performance hit was roughly 3x **Version 2** 😳 

```py
<pyarrow.acero.Declaration>
ExecPlan with 18 nodes:
:ConsumingSinkNode{}
  :UnionNode{}
    :ScalarAggregateNode{aggregates=[
	min(a, {skip_nulls=true, min_count=1}),
	mean(b, {skip_nulls=true, min_count=1}),
]}
      :ProjectNode{projection=[a, "b": subtract(cast(b, {to_type=double, ...}), c)]}
        :HashJoinNode{implementation=SwissJoin}
          :ScalarAggregateNode{aggregates=[
	min(nw44e556231f0d34, {skip_nulls=true, min_count=1}),
	mean(c, {skip_nulls=true, min_count=1}),
]}
            :FilterNode{filter=(nw44e556231f0d34 == 1)}
              :TableSourceNode{}
          :FilterNode{filter=(nw44e556231f0d34 == 1)}
            :TableSourceNode{}
    :ScalarAggregateNode{aggregates=[
	min(a, {skip_nulls=true, min_count=1}),
	mean(b, {skip_nulls=true, min_count=1}),
]}
      :ProjectNode{projection=[a, "b": subtract(cast(b, {to_type=double, ...}), c)]}
        :HashJoinNode{implementation=SwissJoin}
          :ScalarAggregateNode{aggregates=[
	min(nw44e556231f0d34, {skip_nulls=true, min_count=1}),
	mean(c, {skip_nulls=true, min_count=1}),
]}
            :FilterNode{filter=(nw44e556231f0d34 == 0)}
              :TableSourceNode{}
          :FilterNode{filter=(nw44e556231f0d34 == 0)}
            :TableSourceNode{}


```

</p>
</details> 

And this one skips that part, since both have the same result:

<details><summary>Version 2</summary>
<p>

Although this specific case has an equivalent result - performance is still an order of magnitude slower than native grouped aggregates 😢 

```py
<pyarrow.acero.Declaration>
ExecPlan with 10 nodes:
:ConsumingSinkNode{}
  :UnionNode{}
    :ScalarAggregateNode{aggregates=[
	min(a, {skip_nulls=true, min_count=1}),
	mean(b, {skip_nulls=true, min_count=1}),
]}
      :ProjectNode{projection=[a, "b": subtract(cast(b, {to_type=double, ...}), c)]}
        :FilterNode{filter=(nwbead2bb446a61b == 1)}
          :TableSourceNode{}
    :ScalarAggregateNode{aggregates=[
	min(a, {skip_nulls=true, min_count=1}),
	mean(b, {skip_nulls=true, min_count=1}),
]}
      :ProjectNode{projection=[a, "b": subtract(cast(b, {to_type=double, ...}), c)]}
        :FilterNode{filter=(nwbead2bb446a61b == 0)}
          :TableSourceNode{}
```

</p>
</details> 

<details><summary>Show output</summary>
<p>

```py
pyarrow.Table
a: int64
b: double
----
a: [[3],[1]]
b: [[-3],[-3.5]]
```

</p>
</details> 

## Can this be *executed efficiently* Acero?
Absolutely not 😂

I'd thought that building up the plan in a loop, to then be executed and collected in parallel would have *some* benefit.

AFAICT, even when increasing one of (or both) the number of partitions or the total dataset size - **disabling threading consistently performed better**.

## Summary
Bit of a dead end *for now*, but happy to have seen it through at least 😅

dangotbanned: Oh also, my experimenting with this was **locally** and hasn't/won't be added to the PR (if that wasn't clear 🙏)

Issue #3097, opened 2025-09-06
fix: re-implement `cat.get_categories` to use `unique`, deviate from Polars


MarcoGorelli:
I must confess, I'm beyond confused by Polars' design here: https://github.com/pola-rs/polars/issues/24273#issuecomment-3237316163

> .cat.categories() is not per-column

> If you want just the unique categories that actually exist in a column, compute that with unique

I kinda feel like we should just use `.unique` here then. If that represents a deviation from Polars, then OK, so be it, they've made a breaking change here, and one that I find very surprising

dangotbanned:
> I'm beyond confused by Polars' design here: 

I haven't wrapped my head around it yet, but the motivation seems to be streaming:

- https://github.com/pola-rs/polars/pull/23016
- https://github.com/pola-rs/polars/issues/22568
- https://github.com/pola-rs/polars/issues/24385


> In a future PR we will expose the new capabilities of the new `Categories` system`, which lets you specify in the DataType which columns should share the same categorical mapping.

Is it worth waiting to see what this is *first*?

MarcoGorelli:
yeah let's wait to see what happens

camriddell:
Since this came back up, `polars.Categories` is definitely an approach.

I am +1 on using `unique` instead of `cat.get_categories` on a Polars backed object.

MarcoGorelli:
thanks!

not sure about deviating from Polars any more, but maybe a `nw.factorize` function would be useful

MarcoGorelli:
Ooh it's deprecated in polars https://github.com/pola-rs/polars/pull/28299

dangotbanned:
Maybe we could suggest/transition to using `Enum` instead?

The usage I've been seeing in these recent issues (ordered & easy access to members) would be served where polars suggests `Enum`
- https://docs.pola.rs/user-guide/expressions/categorical-data-and-enums/

FBruzzesi:
Should we consider this one closed by #3902 and #3924 ? @MarcoGorelli

MarcoGorelli:
yup thanks!

PR #3241, opened 2025-10-25
ci: Make it a bit greener

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

---

Current CI is giving me nightmares. An attempt to make it a bit better

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -187,6 +187,8 @@ jobs:
         run: |
           cd py-shiny
           . .venv/bin/activate
+          # temporary pin to get CI green
+          uv pip install "flake8-bugbear<25.10.21"
           make narwhals-install-shiny
       - name: install-narwhals-dev
         run: |
--- a/.github/workflows/pytest-pyspark.yml
+++ b/.github/workflows/pytest-pyspark.yml
@@ -74,7 +74,8 @@ jobs:
           java-version: 17
 
       - name: install-reqs
-        run: uv pip install -e . --group core-tests --group extra --system
+        # temporary pin to get CI green
+        run: uv pip install -e . --group core-tests --group extra "pyarrow<22.0" --system
       - name: install pyspark
         run: echo "setuptools<78" | uv pip install -e . "pyspark[connect]==${SPARK_VERSION}" --system
       - name: show-deps


FBruzzesi:
@MarcoGorelli @dangotbanned regarding this failure (https://github.com/narwhals-dev/narwhals/actions/runs/18802435540/job/53651753272?pr=3241), it seems there is something wrong when we get the categories.

I  noticed that `get_categories` is a elementwise ops, but I don't think that's correct. It should be a filtration:

https://github.com/narwhals-dev/narwhals/blob/4720a58f08ec8b1bdc74241fb58516925a3cc5bf/narwhals/expr_cat.py#L37-L38

Yet I don't see how's that could be the problem

dangotbanned:
Thanks for the ping


> Yet I don't see how's that could be the problem

Could this be related to #3097?

dangotbanned:
> I noticed that `get_categories` is a elementwise ops, but I don't think that's correct. It should be a filtration:

I think you're right that it shouldn't be elementwise, upstream it is classed as `groupwise`

https://github.com/pola-rs/polars/blob/1257df085fd70eecc1c840a316957da2a811caf1/crates/polars-plan/src/plans/aexpr/function_expr/cat.rs#L40

But AFAICT, that being wrong wouldn't lead to the result containing categories *unrelated to* the original column.
The issue I'd expect to see is a (native) `ShapeError` being raised when not used in isolation

FBruzzesi:
> But AFAICT, that being wrong wouldn't lead to the result containing categories unrelated to the original column.

Yes that's definitely something else going on!

> The issue I'd expect to see is a (native) ShapeError being raised when not used in isolation

👌🏼

FBruzzesi:
I assume these values come from the conftest in `tests/testing/conftest.py`, yet I can't tell if the issue is us, pytest or polars 😂 I have not been able to replicate it locally

tests/expr_and_series/cat/get_categories_test.py
```diff
@@ -23,6 +24,21 @@ def test_get_categories(constructor_eager: ConstructorEager) -> None:
     assert_equal_data({"a": result_series}, expected)
 
 
+def test_get_categories_lazy(constructor_eager: ConstructorEager) -> None:
```
FBruzzesi, line 27: This might be confusing:

```py
[...]_lazy(constructor_eager, [...]
```

For all the lazy backends we don't support `.cat` at all, so there is no point in testing them

dangotbanned: I haven't peeked at the code, but I'm assuming that would be *except* `polars` and (maybe) `dask`?

MarcoGorelli:
thanks! yeah i'd say the categories isssue is probably https://github.com/narwhals-dev/narwhals/issues/3097

MarcoGorelli (approved):
thanks @FBruzzesi !

so, the only remaining nightly failure is the one already fixed in polars, cool!

PR #3248, opened 2025-10-28
[Enh] added modifications to make tests accessible to plugins

### LINKED ISSUE #3232: [Enh]: have a constructor in Narwhals so that its testing suite can be used in plugins and other extensions

To facilitate further development in [narwhals-daft](https://github.com/MarcoGorelli/narwhals-daft), we'd like to have access to the Narwhals testing suite. At the moment, we don't have a way of running `nw` tests for the plugin code. We think if this was available, it would also help other plugin or extension developers. 

see above

From discussions with @MarcoGorelli, the initial suggestion was: 

- in narwhals-daft, clone the narwhals repository as a git submodule
- point pytest to the narwhals/tests directory
- get pytest to pick up some daft_constructor
- we'd need some way to specify which tests are expected to fail

Tbh I don't have a clear idea on how the connection via a constructor happens but `cudf` is using the Narwhals testing suite (https://github.com/rapidsai/cudf/issues/17662), so I'm trying to understand how it happens there & if we can adapt it to `nw`. 

24.10.25 update: 

we had another synch. I think the idea is to have a `run_tests.sh` script, a bit like `cudf` has here for running the narwhals test suite in their repo : [ci/test_narwhals.sh](https://github.com/rapidsai/cudf/blob/main/ci/test_narwhals.sh).

The proposal would be proceed as follows:

```
# git clone narwhals (pin a particular commit)
# make a list of failing tests
# cd narwhals
# pytest --constructors=daft -k "not FAILING_TESTS"
  # possibly with a `-p narwhals_daft.testing.narwhals_test_plugin`
  # which uses our constructor?
  # can we override `pytest_generate_tests`?
```

Chunks of code in `cudf` are likely to be of use:

in their shell script:

https://github.com/rapidsai/cudf/blob/51b109136f26638e0596a95cf693e3cbe0f2f531/ci/test_narwhals.sh#L31-L54

I think the syntax with the `-k` flag and the group of tests to skip is particularly relevant?

`cudf` actually has a plugin for using the nw testsuite, which gets called in the shell script. In that python script, they mark the tests that are expected to fail:

https://github.com/rapidsai/cudf/blob/51b109136f26638e0596a95cf693e3cbe0f2f531/python/cudf/cudf/testing/narwhals_test_plugin.py#L25-L28C54

_No response_

_No response_

ym-pett:

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #2959
- Closes #3232
- the corresponding work in `narwhals-daft` is at https://github.com/MarcoGorelli/narwhals-daft/pull/19

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -66,7 +66,7 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         "--use-external-constructor",
         action="store_true",
         default=False,
-        help="run tests with all cpu constructors",
+        help="run tests with external constructor",
     )
     parser.addoption(
         "--constructors",


tests/conftest.py
```diff
@@ -62,6 +62,12 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         default=False,
         help="run tests with all cpu constructors",
     )
+    parser.addoption(
+        "--use-external-constructor",
+        action="store_true",
+        default=False,
+        help="run tests with all cpu constructors",
```
MarcoGorelli: i may have copy-pastad the "help" but it needs updating

ym-pett: silly me! I've changed this

PR #3252, opened 2025-10-29
fix: allow for `None` dtype in `from_dict`

### LINKED ISSUE #3176: [Bug]: narhwals fails to create polars dataframe from_dict with None in schema
### Describe the bug

When creating a polars dataframe using nw.from_dict, if a schema is passed and one of the columns is assigned None (for polars to infer the type), it raises an error. In previous versions of narwhals no error was raised.

```python
import narwhals as nw

schema={'a':nw.String, 'b':nw.Categorical, 'c': None}
data_result = {'a': ['a','b', '3.0'], 'b':['c', 'c', None], 'c': [None, None, None]}
df = nw.from_dict(data_result,schema=schema, backend="polars")
```

Also, this still works as expected without raising an error

```python
import polars as pl

schema={'a':pl.String, 'b':pl.Categorical, 'c': None}
data_result = {'a': ['a','b', '3.0'], 'b':['c', 'c', None], 'c': [None, None, None]}
df_control = pl.from_dict(data_result,schema=schema)
```

### Expected results

No error should be raised and the dataframe should be created normally

### Actual results

the following error is raised:

```
  File "/home/otto/githubrepos/pyreadstat/test_narbug.py", line 11, in <module>
    df_control = nw.from_dict(data_result,schema=schema, backend="polars")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/stable/v2/__init__.py", line 1106, in from_dict
    return _stableify(nw_f.from_dict(data, schema, backend=backend))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/_utils.py", line 1427, in wrapper
    return fn(*args, **kwds)
           ^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/functions.py", line 299, in from_dict
    return ns._dataframe.from_dict(data, schema=schema, context=ns).to_narwhals()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/_polars/dataframe.py", line 322, in from_dict
    pl_schema = Schema(schema).to_polars() if schema is not None else schema
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/schema.py", line 330, in to_polars
    pl.Schema(schema)
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/polars/schema.py", line 102, in __init__
    for name, tp in input:
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/schema.py", line 326, in <genexpr>
    (name, narwhals_to_native_dtype(dtype, self._version))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/otto/miniconda3/envs/pyreadstat/lib/python3.12/site-packages/narwhals/_polars/utils.py", line 193, in narwhals_to_native_dtype
    base_type = dtype.base_type()
                ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'base_type'
```

```shell
System:
    python: 3.12.0 | packaged by Anaconda, Inc. | (main, Oct  2 2023, 17:29:18) [GCC 11.2.0]
executable: /home/otto/miniconda3/envs/pyreadstat/bin/python
   machine: Linux-5.15.0-157-generic-x86_64-with-glibc2.35

Python dependencies:
     narwhals: 2.6.0
        numpy: 2.3.1
       pandas: 2.3.1
        modin: 
         cudf: 
      pyarrow: 20.0.0
      pyspark: 
       polars: 1.31.0
         dask: 
       duckdb: 
         ibis: 
     sqlframe:
```

### Relevant log output

```shell

```

MarcoGorelli:
closes #3176

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/from_dict_test.py
+++ b/tests/from_dict_test.py
@@ -100,8 +100,12 @@ def test_from_dict_dtype_none(eager_backend: EagerAllowed) -> None:
     assert_equal_data(result, data)
 
 
-def test_from_dict_schema_too_long(eager_backend: EagerAllowed) -> None:
+def test_from_dict_schema_mismatched(eager_backend: EagerAllowed) -> None:
     schema = {"a": nw.String(), "b": None, "c": None}
+    data = {"a": ["a", "b"], "b": [1, 2]}
+    with pytest.raises(InvalidOperationError):
+        nw.from_dict(data, schema, backend=eager_backend)
+    schema = {"a": nw.String()}
     data = {"a": ["a", "b"], "b": [1, 2]}
     with pytest.raises(InvalidOperationError):
         nw.from_dict(data, schema, backend=eager_backend)


narwhals/_compliant/dataframe.py
```diff
@@ -189,8 +190,8 @@ def from_dict(
         data: Mapping[str, Any],
         /,
         *,
-        context: _LimitedContext,
-        schema: IntoSchema | None,
+        ns: CompliantNamespaceAny,
```
dangotbanned: Why can't you just use `context._implementation._backend_version`?

AFAICT, this was changed everywhere but the version was only needed for `pyarrow`.
If that's the case, then an even smaller change would be importing this 🙂

https://github.com/narwhals-dev/narwhals/blob/bc6a77fd0c7ad7b89a0124f4aec3fe0f13e80e90/narwhals/_arrow/utils.py#L62

tests/test_plugin/test_plugin/namespace.py
```diff
@@ -17,6 +17,10 @@ def __init__(self, *, version: Version) -> None:
     def from_native(self, native_object: DictFrame) -> DictLazyFrame:
         return DictLazyFrame(native_object, version=self._version)
 
+    @property
+    def _backend_version(self) -> tuple[int, ...]:  # pragma: no cover
```
dangotbanned: Surely needing to add a new property on a lazy plugin, to fix an issue on an eager-only constructor seems fishy, right?

MarcoGorelli: yeah it didn't seem great but also seemed innocuous enough, thanks for providing a better solution

tests/from_dict_test.py
```diff
@@ -89,6 +90,23 @@ def test_from_dict_empty_with_schema(eager_backend: EagerAllowed) -> None:
...
 
 
+def test_from_dict_dtype_none(eager_backend: EagerAllowed) -> None:
+    if str(eager_backend) == "pyarrow" and PYARROW_VERSION < (14,):
+        pytest.skip()
+    schema = {"a": nw.String(), "b": None}
+    data = {"a": ["a", "b"], "b": [1, 2]}
+    result = nw.from_dict(data, schema, backend=eager_backend)
+    assert result.schema == {"a": nw.String(), "b": nw.Int64}
+    assert_equal_data(result, data)
```
dangotbanned, line 100: Should we be testing both `None` and `nw.Unknown` behave the same?

`polars` should aready work (https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3461699749), but the others might need to test for `dtype is None or issubclass(dtype.base_type(), nw.Unknown)`

MarcoGorelli: I'm not sure about `nw.Unknown`, I don't know if it happens to work in Polars by accident, I haven't seen it documented

dangotbanned: You could try:

1. Investigating in a debugger, or
2. Asking on discord

If it isn't expected to work, maybe they'll want to close that edge case?

MarcoGorelli: sure, will do so, but we can keep it separate from this pr right? any other comments?

dangotbanned: > sure, will do so, but we can keep it separate from this pr right?

👍 as long as I can point to something (even if unresolved) in a reply to (https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3461539839), so @ofajardo isn't caught out

> any other comments?

Will take another look later.

I should probably open another issue for the `from_native` variant of (#2486) now 🤔

dangotbanned: Thanks for following up with https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3471998235 @MarcoGorelli ♥️

dangotbanned (approved):
Thanks @MarcoGorelli 

hungry narwhal

tests/from_dict_test.py
```diff
@@ -89,6 +90,23 @@ def test_from_dict_empty_with_schema(eager_backend: EagerAllowed) -> None:
...
+
+
+def test_from_dict_schema_too_long(eager_backend: EagerAllowed) -> None:
```
camdell: Is it also worth adding a test for when the supplied schema is too short (underspecified) as well?

camriddell: Other than the minor comment ^ this looks good to me! Thanks for the super fast turn around @MarcoGorelli

MarcoGorelli:
cool, thanks both!

tests/from_dict_test.py
```diff
@@ -89,6 +90,27 @@ def test_from_dict_empty_with_schema(eager_backend: EagerAllowed) -> None:
     assert result.schema == schema
 
 
+def test_from_dict_dtype_none(eager_backend: EagerAllowed) -> None:
+    if str(eager_backend) == "pyarrow" and PYARROW_VERSION < (14,):
+        pytest.skip()
+    schema = {"a": nw.String(), "b": None}
+    data = {"a": ["a", "b"], "b": [1, 2]}
+    result = nw.from_dict(data, schema, backend=eager_backend)
+    assert result.schema == {"a": nw.String(), "b": nw.Int64}
```
camriddell, line 99: What if we end up with some backends in the future don't cast the `"b"` column to `Int64` but to another `Int` subtype or even a `Float`?

Would it be worth instead to just assert that the resultant schema has `"b"` in it? Or shall we cross this bridge when we get to it?

MarcoGorelli: that's a good question! i'd say let's cross it when we need to

PR #3254, opened 2025-10-29
enh: Introduce `narwhals.sql`

MarcoGorelli:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/sql.py
```diff
@@ -0,0 +1,63 @@
...
+    column_mapping = {
+        col: narwhals_to_native_dtype(dtype, Version.MAIN, TZ)
+        for col, dtype in schema.items()
+    }
+    dtypes = ", ".join(f"{col} {dtype}" for col, dtype in column_mapping.items())
+    CONN.sql(f"""
+        CREATE TABLE "{name}"
+        ({dtypes});
+        """)
+    return from_native(CONN.table(name))
```
MarcoGorelli: 🤔 maybe this could return a subclass of `LazyFrame` which has a `to_sql` method, with an optional dependency on `sqlglot` if people want to pass `dialect != 'duckdb'` or `pretty=True`

narwhals/sql.py
```diff
@@ -0,0 +1,106 @@
...
+    msg = (
+        "`narwhals.sql` requires DuckDB to be installed.\n\n"
+        "Hint: run `pip install -U narwhals[sql]`"
+    )
+    raise ModuleNotFoundError(msg) from _exc
+
+CONN = duckdb.connect()
+TZ = DeferredTimeZone(
+    CONN.sql("select value from duckdb_settings() where name = 'TimeZone'")
+)
```
MarcoGorelli, line 26: I think (hope?) that in the future we could make a `narwhals-sqlglot` or `narwhals-substrait` plugin and use that here. But for now, I think using DuckDB for this is quite nice

MarcoGorelli:
There's not been objections, and this is left out of the stable API, so i think it's ok to ship it let people try it out

PR #3256, opened 2025-10-30
tests: Allow to run polars without extra deps

### LINKED ISSUE #1726: test: allow for tests to be run without Polars / PyArrow/ pandas installed

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #1726  (I would argue that running pandas, pyarrow and polars without additional deps closes the issue. Other libraries might still require some interdependencies)
- Related: https://github.com/pola-rs/polars/pull/25081

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/hypothesis/getitem_test.py
+++ b/tests/hypothesis/getitem_test.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 
     from narwhals.typing import IntoDataFrame
 
+pytest.importorskip("pandas")
 pytest.importorskip("polars")
 import polars as pl
 
--- a/tests/translate/from_native_test.py
+++ b/tests/translate/from_native_test.py
@@ -316,7 +316,7 @@ def test_from_native_strict_native_series() -> None:
 
 @pytest.mark.skipif(lf_pl is None, reason="polars not found")
 def test_from_native_strict_native_series_polars() -> None:
-    pytest.importorskip("mumpy")
+    pytest.importorskip("numpy")
     obj: list[int] = [1, 2, 3, 4]
     np_array = pl.Series(obj).to_numpy()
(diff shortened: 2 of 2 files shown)


.github/workflows/pytest.yml
```diff
@@ -118,17 +118,24 @@ jobs:
...
-        run: uv pip freeze
-      - name: Run pytest
+        run: |
+          uv pip install -U pyarrow --system
+          uv pip freeze
+      - name: Run pytest (pandas[pyarrow] and pyarrow)
         run: pytest tests --runslow --constructors=pandas[pyarrow],pyarrow
+      - name: install-polars
+        run: |
+          uv pip uninstall pandas pyarrow --system
```
FBruzzesi, line 134: Notice that numpy will still remain installed, but the changes allow to run polars even without numpy

MarcoGorelli: out of interest, why not uninstall numpy then?

FBruzzesi: I should have expanded more on this.

Some polars functionalities require numpy to be installed. It comes down to either:

- uninstall numpy and do not test them
- keep numpy and test them

In both cases it's no longer dependent on pandas and pyarrow installation, which was the initial goal.

What's your take on this?

MarcoGorelli: this is probably fine, thanks!

PR #3262, opened 2025-10-31
chore: `IntoSchema` overhaul

FBruzzesi:
## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3257

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

## If you have comments or can explain your changes, please do so below

Opening as draft - I will follow up with a few more tests

PR #3266, opened 2025-10-31
chore: Fix `IntoSchema` typing, allow to pass `Sequence[tuple[str, IntoDType]] | SupportsKeysAndGetItem[str, IntoDType]`

### LINKED ISSUE #3257: `typing.IntoSchema` too narrow
This runs fine, as expected:
```python
print(nw.from_dict({'a': [1,2,3]}, schema={'a': nw.Int64}, backend='pyarrow'))
```

but the currently typing wants us to write
```py
print(nw.from_dict({'a': [1,2,3]}, schema={'a': nw.Int64()}, backend='pyarrow'))
```

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #3257

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/typing.py
```diff
@@ -302,13 +302,12 @@ def Binary(self) -> type[dtypes.Binary]: ...
     └──────────────────┘
 """
 
-
-# TODO @dangotbanned: fix this?
-# Constructor allows tuples, but we don't support that *everywhere* yet
-IntoSchema: TypeAlias = "Mapping[str, dtypes.DType] | Schema"
+IntoSchema: TypeAlias = (
+    "Mapping[str, IntoDType] | Sequence[tuple[str, IntoDType]] | Schema"
+)
```
FBruzzesi, line 307: @dangotbanned I hope you are proud 😇

tests/frame/schema_test.py
FBruzzesi, line 1: Randomly removing some initialization to test the typechecking

narwhals/typing.py
```diff
@@ -331,8 +330,31 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+    |a: [[1,2,3]]            |
+    |b: [[null,"hi","howdy"]]|
+    |c: [[2.1,2,null]]       |
+    └────────────────────────┘
 """
 
+IntoNullableSchema: TypeAlias = (
+    "Mapping[str, IntoDType | None] | Sequence[tuple[str, IntoDType | None]]"
+)
+"""Schema specification with possible None values."""
```
FBruzzesi: TODO: Please let's come up with a better description

narwhals/_utils.py
```diff
@@ -2107,3 +2112,21 @@ def extend_bool(
     Stolen from https://github.com/pola-rs/polars/blob/b8bfb07a4a37a8d449d6d1841e345817431142df/py-polars/polars/_utils/various.py#L580-L594
     """
     return (value,) * n_match if isinstance(value, bool) else tuple(value)
+
+
+class NullableSchema(OrderedDict[str, "IntoDType | None"]):
```
FBruzzesi, line 2116: Reason for this class is _mostly_ two folded:

- Use as a utility to convert a `Sequence[tuple[str, DType {|None}]]` into a mapping. Hence make it easy to use the same API (`key, dtype in obj.items()`).
- Easily set a flag to know if any value passed is None

narwhals/dataframe.py
```diff
@@ -603,6 +604,17 @@ def from_dict(
...
             data, backend = _from_dict_no_backend(data)
+        if (
+            schema
+            and data
+            and (
+                diff := set(NullableSchema(schema).keys()).symmetric_difference(
+                    data.keys()
+                )
+            )
+        ):
```
FBruzzesi: Slightly pissed by this formatting 🤔

narwhals/typing.py
dangotbanned: 🥳 

But now to ruin your day slightly ...


We got our typing for `Schema.__init__` from `polars`, but even this is too strict.
Here's what `typeshed` has for `dict`, which we support (except `**kwargs`)

https://github.com/python/typeshed/blob/bf7214784877c52638844c065360d4814fae4c65/stdlib/builtins.pyi#L1158-L1187


**Luckily** we don't need to use any overloads, but we should have this in `IntoSchema` somewhere:

```py
Iterable[tuple[str, IntoDType]] | SupportsKeysAndGetItem[str, IntoDType]
```

dangotbanned: Interesting timing though, I just pushed a test where I passed a non-`Mapping` object to `nw.Schema` 😂 

https://github.com/narwhals-dev/narwhals/blob/1579b32d6d15f982d59cde40f7f884f6f3e42f2b/tests/plan/schema_test.py#L18-L20

FBruzzesi: Thanks @dangotbanned

> ```python
> Iterable[tuple[str, IntoDType]] | SupportsKeysAndGetItem[str, IntoDType]
> ```

- `SupportsKeysAndGetItem` what the hell is this 😂 how's Mapping not enough anymore?
- Regarding `Iterable`, I always have a hard time being comfortable with it, since it can be infinite (and lazy).
  - How does infinite converts to a dict?
  - Lazy makes it a bit more of a headache when we want to check it's content (e.g. first element is a string or a tuple), but that's alright
  - This is the reason why I went with `Sequence`, although polars `Schema.__init__` signature is with the `Iterable`

FBruzzesi:
Regarding CI failures: the only one related to this PR seem to be tubular 👀

narwhals/dataframe.py
FBruzzesi: Tricked ruff into a slightly better formatting 😉

narwhals/typing.py
MarcoGorelli: > We got our typing for Schema.__init__ from polars, but even this is too strict.

can we resolve that later and just get the `type[Dtype]` / `Dtype` part sorted out in this one?

dangotbanned: > `SupportsKeysAndGetItem` what the hell is this 😂 how's Mapping not enough anymore?

Lol, it is an object that supports those methods ofc 😉

just import it in a `TYPE_CHECKING` block from here

https://github.com/python/typeshed/blob/be34e9201db75891a67d4d3ce5e5705ee6636f6f/stdlib/_typeshed/__init__.pyi#L161-L164

```py
from _typeshed import SupportsKeysAndGetItem
```

dangotbanned: > Regarding Iterable, I always have a hard time being comfortable with it, since it can be infinite 

If someone passes an infinite stream to an initializer - they need to learn the lesson that the call will never complete 

>  (and lazy).

This is a desirable property for the caller and I think if *we* want to validate things, then that is our responsibility

### Reducing scope
How about aligning just `Schema.__init__` with `dict.__init__`?

We are only checking for `None` there, so the issues you have seem to not apply there

narwhals/_utils.py
dangotbanned: The name `Nullable*` is making me think this is related to https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3368292079 again 🫣

I'm *not* opposed to having classes to make our internal API cleaner btw 👍

narwhals/typing.py
FBruzzesi: @dangotbanned attempt in https://github.com/narwhals-dev/narwhals/pull/3266/commits/9ce61fe18d682d3f92a566c18f97139ad0e564e1, which now fails for mypy when the dict has mix of initialized and un-initialized dtypes, which was the initial purpose of the PR 🤔

Edit: By changing Iterable into Sequence there are no issue (see https://github.com/narwhals-dev/narwhals/pull/3266/commits/0e0d7a80def04e57258eb59d99e21f636d05af82). I am personally not interested in diving deeper than this.

Last to fix is the forward reference for tubular. Which I am unsure if it's our fault or them not importing something within a `if TYPE_CHEKING:` block

narwhals/_utils.py
FBruzzesi: > The name Nullable* is making me think this is related to https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3368292079 again 🫣

Not sure I see how it's related. Passing `None` seems more like to be a free card, "live and let live" kind of behavior.

> I'm not opposed to having classes to make our internal API cleaner btw 👍

Any preference in having it prefixed by `_`?

dangotbanned: > Any preference in having it prefixed by `_`?

It is already in `_utils`, so no need for a prefix 🙂

The current `_` names are an artifact from when we had everything in `utils.py`

dangotbanned: > Not sure I see how it's related. Passing `None` seems more like to be a free card, "live and let live" kind of behavior.

I'll try to give a more complete explanation later, but I was hinting at `Nullable`, `Null`, `None` being a bit overloaded.
Linking to (https://github.com/narwhals-dev/narwhals/issues/3176#issuecomment-3368292079) was supposed to show that *I made this mistake already* 😂

narwhals/typing.py
```diff
@@ -377,13 +398,16 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
     "IntoBackend",
+    "IntoDType",
     "IntoDataFrame",
     "IntoDataFrameT",
     "IntoExpr",
     "IntoFrame",
     "IntoFrameT",
     "IntoLazyFrame",
     "IntoLazyFrameT",
+    "IntoNullableSchema",
```
MarcoGorelli: Are we sure about making this a public type?

Is it possible to only focus on `Dtype` -> `intodtype` changes here without introducing new public types?

FBruzzesi: Ok I honestly think there might be something wrong in the PR to begin with. If I refocus on the basic changes to fix #3257, then passing a dictionary with mixed initialized and un-initialized dtypes does still complain (pyright is ok with it, mypy is not)

FBruzzesi:
I will keep this PR open for a bit longer - if anyone is interested to bringing to the finish line, I am happy to hand it over 🙈

narwhals/functions.py
```diff
@@ -273,33 +266,7 @@ def from_dict(
...
-            msg = "Unknown namespace is expected to implement `from_dict` function."
-            raise AttributeError(msg) from e
-        return from_native(native_frame, eager_only=True)
-    msg = (
-        f"{implementation} support in Narwhals is lazy-only, but `from_dict` is an eager-only function.\n\n"
-        "Hint: you may want to use an eager backend and then call `.lazy`, e.g.:\n\n"
-        f"    nw.from_dict({{'a': [1, 2]}}, backend='pyarrow').lazy('{implementation}')"
-    )
-    raise ValueError(msg)
+    return Version.MAIN.dataframe.from_dict(data, schema, backend=backend)
```
dangotbanned, line 270: I like this change and think we can split those like it into a PR that makes a dent into #2786

narwhals/functions.py
```diff
@@ -423,41 +390,10 @@ def from_numpy(
...
-            msg = "Unknown namespace is expected to implement `from_numpy` function."
-            raise AttributeError(msg) from e
-        return from_native(native_frame, eager_only=True)
-    msg = (
-        f"{implementation} support in Narwhals is lazy-only, but `from_numpy` is an eager-only function.\n\n"
-        "Hint: you may want to use an eager backend and then call `.lazy`, e.g.:\n\n"
-        f"    nw.from_numpy(arr, backend='pyarrow').lazy('{implementation}')"
-    )
-    raise ValueError(msg)
+    return Version.MAIN.dataframe.from_numpy(data, schema, backend=backend)
```
dangotbanned, line 393: (Same comment as `from_dict`)

narwhals/typing.py
dangotbanned: > Last to fix is the forward reference for tubular. Which I am unsure if it's our fault or them not importing something within a `if TYPE_CHEKING:` block

`tubular` uses a runtime type checker ([`beartype`](https://github.com/beartype/beartype)).
I ran into a similar problem the last time I changed the definition of `Schema`:

- https://github.com/azukds/tubular/issues/455
- #2894

narwhals/schema.py
```diff
@@ -74,17 +75,15 @@ class Schema(OrderedDict[str, "DType"]):
...
+    def __init__(self, schema: IntoSchema | None = None) -> None:
         schema = schema or {}
         super().__init__(schema)
 
     def names(self) -> list[str]:
         """Get the column names of the schema."""
         return list(self.keys())
 
-    def dtypes(self) -> list[DType]:
+    def dtypes(self) -> list[IntoDType]:
```
dangotbanned, line 86: I *think* this may be difficult to work with.

`list[DType]` is invariant *but* names a single concrete type.

OTOH, `list[IntoDType]` will expand to this:

```py
list[DType | type[NumericType | TemporalType | String | Boolean | Binary | Categorical | Unknown | Object]
```

And the problem there is that type would be rejected by another annotation which used `list[DType]` 😢

narwhals/typing.py
dangotbanned: Regarding (https://github.com/narwhals-dev/narwhals/pull/3266/files#r2483418646), (https://github.com/narwhals-dev/narwhals/pull/3266/files#r2483400506)

**This is my bad**, but we should keep that part and my original TODO as a *todo* 
https://github.com/narwhals-dev/narwhals/blob/d7bb050e91a6a6528eba5fcf8e81fa8bfeecaa8b/narwhals/typing.py#L307-L308

These things *are* related, but the timing of #3252 complicated both this PR *and* my TODO 😞

narwhals/_utils.py
dangotbanned: > > Not sure I see how it's related. Passing `None` seems more like to be a free card, "live and let live" kind of behavior.
> 
> I'll try to give a more complete explanation later, but I was hinting at `Nullable`, `Null`, `None` being a bit overloaded.


I guess *now* still counts as later 😂 

Somebody wrote docs about the term *nullable*, and this usage is not to do with that

- https://fbruzzesi.github.io/anyschema/user-guide/metadata/#nullable

dangotbanned (commented):
Thanks for the PR (and reminding me about it 🙏) @FBruzzesi 

I think there are parts of this that are ready to go!

But the combination of (#3262 and #3252) + **my** suggestion took the PR in a direction that you couldn't have anticipated when opening #3262

It seems like I made a suggestion on that PR that may have been lost (https://github.com/narwhals-dev/narwhals/pull/3262#issuecomment-3474049061):

> > [#3262 (comment)](https://github.com/narwhals-dev/narwhals/pull/3262#issuecomment-3473769641)
> 
> @FBruzzesi I think if you were looking to clean things up following #3252, then #2486 may be what we're missing?

This issue is still open and we've both mentioned it recently, maybe there's something to that?

- #2486

FBruzzesi:
I will close this as:

* Majority of features were introduces in #3756
* It's easier to start from scratch than to solve conflicts here if we want to allow `SupportsKeysAndGetItem[str, IntoDType]`

Issue #1223, opened 2024-10-18
[Enh]: Add `Series|Expr.replace`


FBruzzesi:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

This would enable plotly to do custom sorting without filtering + concatenating:

```diff
- nw.concat(
-     [df.filter(nw.col(names) == value) for value in order],
-     how="vertical"
- )
+ (df.with_columns(
+    __custom_sort_col = nw.col(names).replace({v:i for i,v in enumerate(order)}
+ )
+ .sort("__custom_sort_col")
+ .drop("__custom_sort_col")
+ )
```

([code snippet](https://github.com/plotly/plotly.py/pull/4790/files#diff-f4833c7cb58691cdcdffb16480b420516a074457d6f8ed44bcc12660a5391771R2042-R2044))

### Please describe the purpose of the new feature or describe the problem to solve.

Replicate polars [`Expr|Series.replace`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.replace.html)

MarcoGorelli:
thanks @FBruzzesi !

I think the Polars-native solution would be:
```python
df.sort(pl.col(names).replace({x: i for i, x in enumerate(order)}))
```
, without setting any temporary columns

that would require both `DataFrame.sort` taking expressions, and `Expr.replace` - i'll take a look

FBruzzesi:
Thanks for the feedback! Yes I would imagine that polars can get around that using expressions as sort key - yet we currently don't support expressions in these contexts, and I have the impression that it may not be trivial to allow that in the current settings?

MarcoGorelli:
yup, definitely not trivial...but I think you're right, `replace` (or rather, `replace_all`?) is what we need here - and definitely an improvement over `concat`

MarcoGorelli:
what would you think about doing this using a join? Polars does a join under the hood to do this anyway

https://github.com/pola-rs/polars/blob/dbbd93fae922b94cb1b5e700472a75b5a975fc44/crates/polars-ops/src/series/ops/replace.rs#L203-L233

Example:
```python
@nw.narwhalify(eager_only=True)
def func(df, key, order):
    order_key = generate_unique_token(8, df.columns)
    order_df = nw.from_dict({key: order, order_key: range(len(order))}, native_namespace=nw.get_native_namespace(df))
[...]

PR #3276, opened 2025-11-03
feat: Add `default` in `{Expr,Series}.replace_strict`

### LINKED ISSUE #3209: enh: replace_strict with default values
### Discussed in https://github.com/narwhals-dev/narwhals/discussions/3208

<div type='discussions-op-text'>

<sup>Originally posted by **edenian-prince** October 14, 2025</sup>
I'm trying to map values from a dict. I have a column of values in my dataframe, and want something like if `apples` is a value in my dataframe column and it is in the dict, then return the value mapped to apples, which is `red`. I need a default value where if a value in my dataframe column is not in the dict, then it would assign `default`.

it looks like `default` is not in narwhals currently. is there a simple way to use replace_strict and assign default values?

in polars it's pretty simple:

```python
# Create simple test DataFrame
df = pl.DataFrame({
    "fruit": ["apple", "banana", "cherry", "durian"]
})

# Mapping (note: "durian" is missing!)
mapping = {
    "apple": "red",
    "banana": "yellow",
    "cherry": "red"
}

df.with_columns(
   pl.col("fruit").replace_strict(mapping,default='default_string')
)
```

Here's an example with narwhals: 

```python
import polars as pl
import narwhals as nw

# Create simple test DataFrame
df = pl.DataFrame({
    "fruit": ["apple", "banana", "cherry", "durian"]
})

# Convert to Narwhals
data = nw.from_native(df)

# Mapping (note: "durian" is missing!)
mapping = {
    "apple": "red",
    "banana": "yellow",
    "cherry": "red"
}

# Attempt to apply replace_strict
# Expected: unmatched values fallback to default
# Actual: InvalidOperationError, and default= is NOT supported
keys = list(mapping.keys())
values = list(mapping.values())

data.with_columns(
    nw.col("fruit")
    .replace_strict(keys, values)  # ⚠️ No `default` param allowed
    .alias("color")
)
```

And I get an error like this, even though there is no default parameter:

```python
InvalidOperationError: incomplete mapping specified for `replace_strict`

Hint: Pass a `default` value to set unmapped values.
Cell In[22], line 7
      4 keys = list(mapping.keys())
      5 values = list(mapping.values())
----> 7 data.with_columns(
      8     nw.col("fruit")
      9     .replace_strict(keys, values)  # ⚠️ No `default` param allowed
     10     .alias("color")
     11 )
```

there's gotta be a simple way to do this? i tried .with/then statements but face other issues, like it's trying to assign the whole list of values to a specific row

```python
TypeError: cannot create expression literal for value of type Expr.

(
    data
    .with_columns(
        nw.when(nw.col("lab").is_in(keys))
    .then(nw.col("fruit").replace_strict(keys, values))  # not strict!
    .otherwise(pl.lit("default_string"))
    )
)

```</div>

### PR DESCRIPTION
**Pull request by FBruzzesi**

- [ ] 💾 Refactor
- [x] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3209
- Fixes an untracked bug in the case that in the mapping there is `{value: None, ...}`

## Checklist

- [x] Tests added
- [x] Documented the changes

Majority of the diff is in the test file

### DIFF
--- a/narwhals/_arrow/series.py
+++ b/narwhals/_arrow/series.py
@@ -803,7 +803,7 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
                     "replace_strict did not replace all non-null values.\n\n"
                     f"The following did not get replaced: {unmatched_values}"
                 )
-                raise ValueError(msg)
+                raise InvalidOperationError(msg)
         else:
             result_native, default = extract_native(result, default)
             # Only fill with default where the value wasn't matched (not where result is null due to mapping)
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -688,7 +688,7 @@ class PandasLikeSeries(EagerSeries[Any]):
                     "replace_strict did not replace all non-null values.\n\n"
                     f"The following did not get replaced: {unmatched_values}"
                 )
-                raise ValueError(msg)
+                raise InvalidOperationError(msg)
             # For unmatched values (nulls in original), set to null
             native_result = native_result.where(was_matched, None)
         else:
--- a/narwhals/expr.py
+++ b/narwhals/expr.py
@@ -849,8 +849,6 @@ class Expr:
     ) -> Self:
         """Replace all values by different values.
 
-        This function must replace all non-null input values (else it raises an error).
-
         Arguments:
             old: Sequence of values to replace. It also accepts a mapping of values to
                 their replacement as syntactic sugar for
@@ -862,6 +860,10 @@ class Expr:
             return_dtype: The data type of the resulting expression. If set to `None`
                 (default), the data type is determined automatically based on the other
                 inputs.
+
+        Raises:
+            InvalidOperationError: If any non-null values in the original column were not
+                replaced, and no default was specified.
 
         Examples:
             >>> import pandas as pd
@@ -887,27 +889,8 @@ class Expr:
 
             Replace values and set a default for values not in the mapping:
 
-            >>> df = nw.from_native(pd.DataFrame({"a": [1, 2, 3, 4]}))
-            >>> df.with_columns(
-            ...     b=nw.col("a").replace_strict(
-            ...         [1, 2], ["one", "two"], default="other", return_dtype=nw.String
-            ...     )
-            ... )
-            ┌──────────────────┐
-            |Narwhals DataFrame|
-            |------------------|
-            |      a      b    |
-            |   0  1    one    |
-            |   1  2    two    |
-            |   2  3  other    |
-            |   3  4  other    |
-            └──────────────────┘
-
-            >>> df = nw.from_native(
-            ...     pd.DataFrame(
-            ...         {"a": [1, 2, 3, 4], "b": ["beluga", "narwhal", "orca", "vaquita"]}
-            ...     )
-            ... )
+            >>> data = {"a": [1, 2, 3, 4], "b": ["beluga", "narwhal", "orca", "vaquita"]}
+            >>> df = nw.from_native(pd.DataFrame(data))
             >>> df.with_columns(
             ...     a_replaced=nw.col("a").replace_strict(
             ...         {1: "one", 2: "two"},
@@ -933,11 +916,11 @@ class Expr:
             new = list(old.values())
             old = list(old.keys())
 
-        if isinstance(default, Expr):
+        if default is no_default:
             node = ExprNode(
                 ExprKind.ELEMENTWISE,
                 "replace_strict",
-                default,
+                default=default,
                 old=old,
                 new=new,
                 return_dtype=return_dtype,
@@ -946,10 +929,11 @@ class Expr:
             node = ExprNode(
                 ExprKind.ELEMENTWISE,
                 "replace_strict",
-                default=default,
+                default,
                 old=old,
                 new=new,
                 return_dtype=return_dtype,
+                str_as_lit=True,
             )
         return self._append_node(node)
 
--- a/narwhals/series.py
+++ b/narwhals/series.py
@@ -1293,8 +1293,6 @@ class Series(Generic[IntoSeriesT]):
     ) -> Self:
         """Replace all values by different values.
 
-        This function must replace all non-null input values (else it raises an error).
-
         Arguments:
             old: Sequence of values to replace. It also accepts a mapping of values to
                 their replacement as syntactic sugar for
@@ -1306,6 +1304,10 @@ class Series(Generic[IntoSeriesT]):
             return_dtype: The data type of the resulting expression. If set to `None`
                 (default), the data type is determined automatically based on the other
                 inputs.
+
+        Raises:
+            InvalidOperationError: If any non-null values in the original column were not
+                replaced, and no default was specified.
 
         Examples:
             >>> import pandas as pd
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

File: narwhals/_arrow/series.py [hunk 1]
```diff
@@ -767,24 +769,42 @@ def unique(self, *, maintain_order: bool = True) -> Self:
 
     def replace_strict(
         self,
+        default: Any | NoDefault,
```
FBruzzesi, line 774: I must say that this is not super intuitive and it took me a good 5 minutes to recall why expressifiable arguments must go first. For a newcomer it would be a treasure hunt

narwhals/_arrow/series.py
```diff
@@ -767,24 +769,42 @@ def unique(self, *, maintain_order: bool = True) -> Self:
...
         old: Sequence[Any] | Mapping[Any, Any],
         new: Sequence[Any],
         *,
         return_dtype: IntoDType | None,
     ) -> Self:
         # https://stackoverflow.com/a/79111029/4451315
         idxs = pc.index_in(self.native, pa.array(old))
+        # Tracks which values were found in the mapping (even if mapped to None)
+        was_matched = pc.is_valid(idxs)
```
FBruzzesi, line 783: This fixes the existing issue - I tried to keep a few inline comments

narwhals/expr.py
```diff
@@ -888,15 +933,25 @@ def replace_strict(
...
-        )
+        else:
+            node = ExprNode(
+                ExprKind.ELEMENTWISE,
+                "replace_strict",
+                default=default,
+                old=old,
+                new=new,
+                return_dtype=return_dtype,
+            )
```
FBruzzesi: Hacking type hinting. The top case only would complain. The bottom one only would lead to wrong results.

FBruzzesi:
Oh boy! Outer join (required for modin) sorts the keys lexicographically 😢  that's why the test fails and the sort is `["one", "three", "two"]`. I will sleep on this

---

Update:

> I will sleep on this

was fruitful, I refactored in https://github.com/narwhals-dev/narwhals/pull/3276/commits/75c60e3bb8a14d73c056b70d136bda7b8f77a2ee without using a join

narwhals/_pandas_like/series.py
```diff
@@ -642,37 +643,58 @@ def shift(self, n: int) -> Self:
...
-            raise ValueError(msg)
-        return result
+
+        # Use pandas Index.get_indexer to find positions of values in old
+        idxs = namespace.Index(old).get_indexer(native)
+        was_matched = idxs >= 0
+
+        new_series = namespace.Series(new, dtype=dtype, name=self.name)
+        if dtype_backend:
+            new_series = new_series.convert_dtypes(dtype_backend=dtype_backend)
```
FBruzzesi: On a self review, this should be needed only if `return_dtype` is None

```suggestion
        if return_dtype is None and dtype_backend is not None:
            new_series = new_series.convert_dtypes(dtype_backend=dtype_backend)
```

narwhals/_arrow/series.py
MarcoGorelli: yup, that's a valid point, will think about how to clarify

narwhals/expr.py
MarcoGorelli: should this be updated now?

narwhals/expr.py
```diff
@@ -879,6 +884,46 @@ def replace_strict(
...
+            ... )
+            ┌──────────────────────────────┐
+            |      Narwhals DataFrame      |
+            |------------------------------|
+            |   a        b       a_replaced|
+            |0  1   beluga              one|
+            |1  2  narwhal              two|
+            |2  3     orca     default_orca|
+            |3  4  vaquita  default_vaquita|
+            └──────────────────────────────┘
```
MarcoGorelli: in the interest of not expanding docstrings too much, perhaps it's ok to leave this one out? I think if people see the example above with `default='other'` it's enough?

narwhals/expr.py
```diff
@@ -888,15 +933,25 @@ def replace_strict(
...
+        else:
+            node = ExprNode(
+                ExprKind.ELEMENTWISE,
+                "replace_strict",
+                default=default,
+                old=old,
+                new=new,
+                return_dtype=return_dtype,
+            )
+        return self._append_node(node)
```
MarcoGorelli: how about
```python
        if default is no_default:
            node = ExprNode(
                ExprKind.ELEMENTWISE,
                "replace_strict",
                default=default,
                old=old,
                new=new,
                return_dtype=return_dtype,
            )
        else:
            node = ExprNode(
                ExprKind.ELEMENTWISE,
                "replace_strict",
                default,
                old=old,
                new=new,
                return_dtype=return_dtype,
                str_as_lit=True
            )
```
?

I think the way it's currently written may not work for `Series` input (if it does, I think it may break if/when we become stricter about parsing inputs)

MarcoGorelli (commented):
thanks for working on this!

MarcoGorelli:
Just to plan for future work - if `default` is specified, then we'll be able to support this for sql-like too? e.g.
```py
In [33]: duckdb.sql("""
    ...: from rel
    ...: select *,
    ...:     case when a == 1 then 'one'
    ...:          when a == 2 then 'two'
    ...:          when a is null then null
    ...:          else 'unmatched'
    ...:     end as b
    ...: """)
Out[33]:
┌───────┬───────────┐
│   a   │     b     │
│ int64 │  varchar  │
├───────┼───────────┤
│     1 │ one       │
│     2 │ two       │
│     3 │ unmatched │
│     4 │ unmatched │
└───────┴───────────┘
```
, so long as `old` and `new` aren't expressions (which they currently aren't anyway. I think the most common use-case would be to replace with a dict anyway)

narwhals/expr.py
dangotbanned: I actually prefer this example to the other one, since the description here mentions a mapping but uses lists (and no other example uses a mapping):


https://github.com/narwhals-dev/narwhals/blob/ab33c2fd50e7992425ea31abc8cc21177a42d6b8/narwhals/expr.py#L888-L893

would this be a short enough version?

```py
df.with_columns(
    a_replaced=nw.col("a").replace_strict(
        {1: "one", 2: "two"}, default=nw.format("default_{}", "b"), return_dtype=nw.String
    )
)
```

MarcoGorelli: sure, we can use a dict in the example above it

FBruzzesi: Sure I can keep this one with mapping + default as an expression

narwhals/expr.py
FBruzzesi: I need to check what happens in the case of `default` being a python literal, both at runtime and for the type  checker.

> may not work for Series input

Not sure if this should be accepted/supported in the expression case. For series of course it is

FBruzzesi:
> Just to plan for future work - if `default` is specified, then we'll be able to support this for sql-like too? e.g.

Yes, this should be supported! I was planning two follow-ups:

- (non-polars) lazy backends with required default
- Maybe `replace` (https://github.com/narwhals-dev/narwhals/issues/1223) can get in almost for free

> so long as `old` and `new` aren't expressions (which they currently aren't anyway. I think the most common use-case would be to replace with a dict anyway)

Yep that's currently not supported anyway

narwhals/expr.py
FBruzzesi: Added in a `Raises` section as polars does. Here how it renders:

image

narwhals/expr.py
FBruzzesi: Nice one, it works, thanks @MarcoGorelli

FBruzzesi:
Did something changed for `is_finite` in compliant polars? (see [failure](https://github.com/narwhals-dev/narwhals/actions/runs/19084198449/job/54520535246?pr=3276))

narwhals/_arrow/series.py
```diff
@@ -769,24 +771,44 @@ def unique(self, *, maintain_order: bool = True) -> Self:
...
         result_native = pc.take(pa.array(new), idxs)
         if return_dtype is not None:
-            result_native.cast(narwhals_to_native_dtype(return_dtype, self._version))
```
MarcoGorelli, line 781: 😄 did we just cast without returning? 😳 🤦

MarcoGorelli (approved):
thanks @FBruzzesi !

narwhals/_arrow/series.py
FBruzzesi: I definitely looks like that 🙈

PR #3283, opened 2025-11-05
test: xfail polars `cat.get_categories` (flaky)

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3097
- Related many CI failures

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,7 @@ ci:
 repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: 'v0.14.2'
+  rev: 'v0.14.3'
   hooks:
     # Run the formatter.
     - id: ruff-format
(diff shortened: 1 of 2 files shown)


camriddell:
I did a touch of digging and found a reliable reprod based on the ordering of the tests run and how they independently update the cache. As expected, this is a *feature* of Polars so we will either need to implement #3097 or wait until Polars releases some local categorical storage [which is in the works](https://github.com/pola-rs/polars/issues/24673#issuecomment-3351237855)

For the reprod, I'll focus on the following example:
a. `tests/testing/assert_series_equal_test.py::test_categorical_as_str` (updates and uses the string cache)
b `tests/expr_and_series/cat/get_categories_test.py::test_get_categories_eager` (updates the string cache)

If the ordering of the test is...
- a. → b.  then a. adds some values into the string cache that b. then (mistakenly) accesses and returns those values raising the failure
- b. → a.  then b. sees only its categorical values that are added to the string cache. Then later a. will update the cache, but since no other tests inspect it, then it doesn't matter.

Issue #3216, opened 2025-10-17
api: move `filter` from `CompliantSeries` to `CompliantExpr`


MarcoGorelli:
Splitting this off from https://github.com/narwhals-dev/narwhals/pull/3152, where I noticed it

Although we don't currently implement `filter` for lazy expressions, that's just an implementation detail, and it should be possible to support `filter` if the expression results in a scalar-like value

For example:
- `nw.col('a').filter(nw.col('b')>1).sum()` should be valid lazy + eager
- `nw.col('a').filter(nw.col('b')>1)` should be eager-only

Issue #3174, opened 2025-10-03
api: remove most (compliant) Series methods, just re-use `to_frame/select`


MarcoGorelli:
I think we can remove most compliant Series methods, by just implementing them as something like
```python
self.to_frame().select(col(self.name).method(*args, **kwargs))
```

So then `CompliantSeries` would be left with Series-only methods, like `__getitem__`, `item`, ...

MarcoGorelli:
i tried this out and it simplifies a few things

in theory there should be no overhead as `to_frame` doesn't make a copy

but, i'd like to do some benchmarking to check anyway

dangotbanned:
I'm generally on board 👍
FYI, this is the approach I've been doing for `ExprIR`, which has a very short (but still incomplete) [`CompliantSeries`](https://github.com/narwhals-dev/narwhals/blob/8f7d2f3dd36ae3e98f0dc929a15830b6b5cb6c93/narwhals/_plan/compliant/series.py)

Since this would be a pretty big internal change - it would be super helpful to see more of a plan 🙏

PR #3285, opened 2025-11-06
docs: API Completeness overhaul

### LINKED ISSUE #2858: docs: API Completeness page

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2858

## Checklist

- [x] Tests added
- [x] Documented the changes

Big overhaul for the API Completeness generation. I realized we have quite a few exceptional cases to deal with 🙈 
Now everything should render as expected but please double check. The top level functionalities through namespaces parsing are also added in a new page.

The javascript allows to have the tables to be sortable and with a free text filter by method name. Disclaimer: JS part is partially AI generated, partially from the mkdocs-material docs

<details> <summary> screenshot </summary> 

image

</details>

 ---

I tried using mkdocs-marimo, which is indeed great as tables come full of features out of the box and it was super easy to add the dropdown filter to select which backends to keep in the table. Yet:

- docs build time went from 3 seconds to 45-50 seconds and I had no patience for that
- the table rendering was a bit off, and not as clean as in marimo notebooks themselves

### DIFF
--- a/docs/css/extra.css
+++ b/docs/css/extra.css
@@ -7,6 +7,28 @@
   margin-left: 5vw;
 }
 
+/* API Completeness Table Styles */
+.table-container {
+  overflow-x: auto;
+}
+
+article table:not([class]) {
+  width: auto !important;
+  max-width: 100%;
+  table-layout: auto;
+}
+
+article table:not([class]) th,
+article table:not([class]) td {
+  white-space: nowrap;
+  padding: 0.6rem 1rem;
+}
+
+article table:not([class]) th:first-child,
+article table:not([class]) td:first-child {
+  min-width: 150px;
+}
+
 /* Column Selector Styles */
 .column-selector-button:hover {
   background: var(--md-primary-fg-color--dark, #303f9f) !important;
--- a/docs/javascripts/table.js
+++ b/docs/javascripts/table.js
@@ -203,8 +203,14 @@ function addColumnSelector(table) {
   container.appendChild(button)
   container.appendChild(dropdown)
   
-  // Insert before the table
-  table.parentNode.insertBefore(container, table)
+  // Wrap table in a container for better layout control
+  var tableContainer = document.createElement('div')
+  tableContainer.className = 'table-container'
+  table.parentNode.insertBefore(tableContainer, table)
+  tableContainer.appendChild(table)
+  
+  // Insert button container before the table container
+  tableContainer.parentNode.insertBefore(container, tableContainer)
 }
 
 function toggleColumn(table, columnIndex, visible) {


FBruzzesi:
Thanks @MarcoGorelli - addressed the deprecated methods in https://github.com/narwhals-dev/narwhals/pull/3285/commits/2823ebee644098c9200f1658888c522803240825

While doing so, I noticed that we don't have a mention for `LazyFrame.tail` in the [main vs stable.v1 differences](https://narwhals-dev.github.io/narwhals/backcompat/#main-vs-stablev1-differences) section.

---

I will wait for @dangotbanned opinion before merging 🙏🏼

dangotbanned:
@FBruzzesi thanks for the summons 😉, screenshot looks great!

I need to take this for a spin locally so bear with me please!

For now though, I have some general comments from looking at the diff:

1. AFAICT, there isn't anything here to ensure the script stays in sync?
    i. Fixing that part of (#2858) like a one-off event and hoping we don't regress seems ... *hopeful?*
2. It looks like there's multiple reimplementations of things from (https://github.com/narwhals-dev/narwhals/blob/2823ebee644098c9200f1658888c522803240825/utils/check_api_reference.py)
    i. Can anything be shared/reused?
3. I know OOP gets a bad rep in data circles 😂, but since the topic is *what do these classes do?*, would it not be appropriate here?
    i. E.g. there are loads of functions that have a docstring and comments on every line (🙄✨?).
    Future *you* (and current me 😉) will thank you if the code explains itself 🙏

FBruzzesi:
Thanks for the perspective @dangotbanned 

Let me add a general comment that kind of addresses all the points below: I treated this as a one off effort, definitely I didn't invest the same time and effort as I would for the main codebase. I mostly focused on going from broke to working, even cutting corners when possible 🙈

> 1. AFAICT, there isn't anything here to ensure the script stays in sync?
>    i. Fixing that part of ([docs: API Completeness page #2858](https://github.com/narwhals-dev/narwhals/issues/2858)) like a one-off event and hoping we don't regress seems ... _hopeful?_

Do you have any idea on how to make it a _guarantee_? I see it hard to make it "stable forever" considering the underlying target might continuously change. I focus on avoiding false positive and false negative, which led to a lot of "exception" cases to handle. At least those are tracked now, but some depends on the current implementation in narwhals itself. I can add TODO comments based on potential/expected changes.

> 2. It looks like there's multiple reimplementations of things from (https://github.com/narwhals-dev/narwhals/blob/2823ebee644098c9200f1658888c522803240825/utils/check_api_reference.py)
>    i. Can anything be shared/reused?

I honestly didn't even try to check/compare it with other scripts

> 3. I know OOP gets a bad rep in data circles 😂, but since the topic is _what do these classes do?_, would it not be appropriate here?
>    i. E.g. there are loads of functions that have a docstring and comments on every line (🙄✨?).
>    Future _you_ (and current me 😉) will thank you if the code explains itself 🙏

That's fair, I will try to put some more engineering effort. Maybe Claude can come handy in refactoring a script like this

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/3285#issuecomment-3502565281) Thanks for explaining @FBruzzesi 

> Do you have any idea on how to make it a *guarantee*? 
> I see it hard to make it "stable forever" considering the underlying target might continuously change.

Well to make things easier, you'll be glad to hear that I'm **not** expecting a *guarantee* of *forever-stability* 😄 


> [!IMPORTANT]
> What I want to see is an effort to *minimize* the burden of keeping this in sync

---

Before I suggest anything, I just wanna highlight this comment as I think we're viewing the issue of (#2858) quite differently:

> I treated this as a one off effort, definitely I didn't invest the same time and effort as I would for the main codebase. I mostly focused on going from broke to working, even cutting corners when possible

In my mind, the **API Completeness** docs *should be* seen as one of the most important assets we have.
If a package is considering adopting `narwhals` - this might be the last thing they see before deciding they *can't* make the switch. 
So if what's documented isn't accurate - then all the effort you (obviously) put into other areas of `narwhals` may not get the attention it deserves ❤️

dangotbanned:
## Suggestions
The public API is already defined and checked for documentation in (https://github.com/narwhals-dev/narwhals/blob/2823ebee644098c9200f1658888c522803240825/utils/check_api_reference.py)

If we moved some of those utilities into somewhere we can import from, without an error like:

```py
from utils.check_api_reference import iter_api_reference_names
import narwhals as nw

frozenset(iter_api_reference_names(nw.DataFrame))
An exception has occurred, use %tb to see the full traceback.

SystemExit: 0
```

Then we'd have a starting point like this, which is known to be in sync:

```py
frozenset({'clone',
           'collect_schema',
           'columns',
           'drop',

           'write_csv',
           'write_parquet'})
```

Next, since we want to know more about `Compliant*` implementations, we can compare these names against what is defined in the `Compliant*` protocol:

```py
from narwhals._compliant import CompliantDataFrame
from typing_extensions import get_protocol_members

get_protocol_members(CompliantDataFrame)
```

<details><summary>Show biiiiiiig output</summary>
<p>

```py
frozenset({'__array__',
           '__getitem__',
           '__len__',
           '__narwhals_dataframe__',
           '__narwhals_namespace__',
           '__native_namespace__',
           '_implementation',
           '_is_native',
           '_native_frame',
           '_version',
           '_with_native',
           '_with_version',
           'clone',
           'collect_schema',
           'columns',
           'drop',
           'drop_nulls',
           'estimated_size',
           'explode',
           'filter',
           'from_arrow',
           'from_dict',
           'from_dicts',
           'from_native',
           'from_numpy',
           'gather_every',
           'get_column',
           'group_by',
           'head',
           'is_unique',
           'item',
           'iter_columns',
           'iter_rows',
           'join',
           'join_asof',
           'lazy',
           'native',
           'pivot',
           'rename',
           'row',
           'rows',
           'sample',
           'schema',
           'select',
           'shape',
           'simple_select',
           'sort',
           'to_arrow',
           'to_dict',
           'to_narwhals',
           'to_numpy',
           'to_pandas',
           'to_polars',
           'unique',
           'unpivot',
           'with_columns',
           'with_row_index',
           'write_csv',
           'write_parquet'})
```

</p>
</details> 

So now the problem is about comparing 2 sets of well-defined interfaces 🙂 

I'd say the next part would be looking at what you *currently* have defined as edge-cases and see if any patterns emerge.
Keeping in mind that:
1. The size of these APIs is expected to grow
2. There are sub-protocols and concrete classes implementing them
3. A certain BDFL is prone to large refactors of the internals 😉

FBruzzesi:
Thanks @dangotbanned, we agree on everything you said in the last two comments.

If I focus on:

> What I want to see is an effort to minimize the burden of keeping this in sync

I would say that the changes here makes it quite easy to:

- extend special cases
- keep almost everything else in sync

which in my mind makes the _what_ really aligned with the purpose we want to achieve.

> I treated this as a one off effort, definitely I didn't invest the same time and effort as I would for the main codebase. I mostly focused on going from broke to working, even cutting corners when possible

This comment of mine was most likely taken to an extreme, which I hope it's not what the final result and implementation are reflecting.

From your previous comment:

> I know OOP gets a bad rep in data circles 😂, but since the topic is what do these classes do?, would it not be appropriate here?

my understanding is that you are not convinced for the _how_ this is implemented.

I can try to propose a different approach, which is more OOP based, but for now I don't see a big gap in what we can achieve either way.

dangotbanned:
> my understanding is that you are not convinced for the *how* this is implemented.

Yeah sorry for not being clearer - I **do like the output** (*what*)! 😄

Forget that I mentioned OOP for now, here are two examples of structuring things differently, that I've worked on

## 1
[This module](https://github.com/vega/altair/blob/11e3146d837ea5130b304ef179d75a414a3047a8/tools/vega_expr.py) parses [markdown docs](https://vega.github.io/vega/docs/expressions/) and transforms the output into [this python module](https://github.com/vega/altair/blob/11e3146d837ea5130b304ef179d75a414a3047a8/altair/expr/__init__.py)

## 2
[This module](https://github.com/vega/altair/blob/11e3146d837ea5130b304ef179d75a414a3047a8/tools/datasets/datapackage.py) parses [json](https://github.com/vega/vega-datasets/blob/cde43707ec65d8755b9d5b8d97845bf6d0d688d4/datapackage.json) into an [intermediate representation](https://github.com/vega/altair/blob/11e3146d837ea5130b304ef179d75a414a3047a8/tools/datasets/models.py) to generate [typing](https://github.com/vega/altair/blob/11e3146d837ea5130b304ef179d75a414a3047a8/altair/datasets/_typing.py) and [a few different summary formats](https://github.com/vega/altair/tree/11e3146d837ea5130b304ef179d75a414a3047a8/altair/datasets/_metadata)

## Why is this relevant?
- They operate on input outside of their control
- The input is expected to change
- They need to handle lots of hairy edge cases

To anticipate that the input may change, the input is moved into an intermediate representation *first*, and then I transformed that *more stable* version into some output(s).

This means upstream changes that break things should only impact one side of the problem.
Even within the first translation - I tried to generalize edge cases as much as possible, such that you will rarely see conditions comparing a value to a string defined inline.

## Why would more structure benefit this script?
How much would break here following any of these changes?
Where would these changes need to be made?
Would all of the current edge casing still make sense when the underlying backends have changed?

- #3174
- #3216
- #2116
- #1561

From my perspective, I think we'd need to make changes all over the script 🤔

FBruzzesi:
> To anticipate that the input may change, the input is moved into an intermediate representation _first_, and then I transformed that _more stable_ version into some output(s).
> This means upstream changes that break things should only impact one side of the problem. Even within the first translation - I tried to generalize edge cases as much as possible, such that you will rarely see conditions comparing a value to a string defined inline.

I agree this is something likable to have, although it's so simple to do these parsing that the need for a intermediate layer for me is just _more_ complexity, almost for the sake of it.

The next point is where we most likely disagree on:

> From my perspective, I think we'd need to make changes all over the script 🤔

> How much would break here following any of these changes? Where would these changes need to be made? Would all of the current edge casing still make sense when the underlying backends have changed?
> 
> * [api: remove most (compliant) Series methods, just re-use `to_frame/select` #3174](https://github.com/narwhals-dev/narwhals/issues/3174)

Realistically switch one pointer. I am not sure how to answer this one, since it might depend on what the implementation entail. But it would be the same with whatever implementation we choose.

> * [api: move `filter` from `CompliantSeries` to `CompliantExpr` #3216](https://github.com/narwhals-dev/narwhals/issues/3216)

Nothing to change for new methods, they are automatically picked up

> * [API: io functions for v2 #2116](https://github.com/narwhals-dev/narwhals/issues/2116)

Simplify/remove one special case that's currently exists because the logic is entirely at the narwhals level

> * [[Enh]: Support for `narwhals.Expr.hist` #1561](https://github.com/narwhals-dev/narwhals/issues/1561)

Automatically picked up

---

One thing to stress from my side: I would rather keep/maintain a mapping of {backend: non-implemented methods}, since in all honesty they are so few at the moment, than having to maintain a overly complex codebase. For instance, in the API reference check, we have a manual list of `SERIES_ONLY` methods, I don't think we have many more "not implemented" methods to track. That said, the only way to know it to try it out

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/3285#issuecomment-3508337464)

No worries @FBruzzesi, I'm happy to agree to disagree on this and move past it 😄

---

I'll try to take another look at the **visual** part of the PR today.

Being able to filter the backend-axis the table (https://github.com/narwhals-dev/narwhals/issues/2858#issuecomment-3104851456) was something that felt missing to me IIRC

FBruzzesi:
> I'll try to take another look at the **visual** part of the PR today.
> 
> Being able to filter the backend-axis the table ([#2858 (comment)](https://github.com/narwhals-dev/narwhals/issues/2858#issuecomment-3104851456)) was something that felt missing to me IIRC

I hope that https://github.com/narwhals-dev/narwhals/pull/3285/commits/44601f6506af5ee6ba562270460568e58726ad79 solves the last bit (as you might guess from the commit message, it wasn't me)

<details><summary> screenshot </summary>

image

</details>

MarcoGorelli:
we ok to ship this?

FBruzzesi:
> we ok to ship this?

I was quite happy with the rework. My take again is: I rather merge it with a solution which is not perfect but renders the correct state of the project and come back to it, than keeping it broken indefinitely in the current documentation because we cannot agree on some details.

MarcoGorelli:
agree - thanks @FBruzzesi !

FBruzzesi:
Ok... one detail went as unplanned... The table filter applies to the **arguments** in the api reference as well, as that's also a table.

<details><summary>Screenshots</summary>
<p>


image

image

</p>
</details> 

---

Might consider revert or fix, sorry folks 🥲

MarcoGorelli:
thanks for spotting this!

 no worries

> Might consider revert or fix, sorry folks 🥲

up to you, how hard would a fix be?

FBruzzesi:
> up to you, how hard would a fix be?

I am checking now, give me _some time_ to get back to you

PR #3239, opened 2025-10-25
perf: Rewrite `with_row_index` so it's faster for pandas/pyarrow and doesn't use `rank`

MarcoGorelli:
With
```python
from __future__ import annotations

from typing import TypeAlias

import pandas as pd
import pyarrow as pa
import ibis
import narwhals as nw
import duckdb
import polars as pl

n = 100_000
data = ({'a': ['o', 'b', 's']*n, 'b': [1,2,3]*n})
dfpd = nw.from_native(pd.DataFrame(data))
dfpa = nw.from_native(pa.table(data))

# timeit:
dfpa.with_row_index(order_by='b')
dfpd.with_row_index(order_by='b')
````

I'm seeing

here:
```
In [1]: %timeit dfpd.with_row_index(order_by='b')
19.7 ms ± 504 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [2]: %timeit dfpa.with_row_index(order_by='b')
4.65 ms ± 147 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

main:
```
In [1]: %timeit dfpd.with_row_index(order_by='b')
37.4 ms ± 497 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [2]: %timeit dfpa.with_row_index(order_by='b')
25 ms ± 297 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #3292, opened 2025-11-07
fix: `with_row_index` was returning incorrect results when used with `order_by` for pandas, pyarrow, and polars

### LINKED ISSUE #3289: Incorrect results for `with_row_index(order_by=...)`
Noticed while looking at the `pyarrow` version for (#3224) that the index seems to be wrong when the column isn't already ordered.

Possibly a regression from #3239? (cc @MarcoGorelli)
Just guessing as that's the most recent change

### Repro

`pandas`, `pyarrow` and `polars` all produce the same result, which I think is wrong?

```py
import narwhals as nw

data = {"a": ["A", "B", "A"], "b": [1, 2, 3], "c": [9, 2, 4]}

df = nw.from_dict(data, backend="polars")
df.with_row_index(order_by="c").sort("index")
```
```
┌───────────────────────────┐
|    Narwhals DataFrame     |
|---------------------------|
|shape: (3, 4)              |
|┌───────┬─────┬─────┬─────┐|
|│ index ┆ a   ┆ b   ┆ c   │|
|│ ---   ┆ --- ┆ --- ┆ --- │|
|│ i64   ┆ str ┆ i64 ┆ i64 │|
|╞═══════╪═════╪═════╪═════╡|
|│ 0     ┆ A   ┆ 3   ┆ 4   │|
|│ 1     ┆ A   ┆ 1   ┆ 9   │|
|│ 2     ┆ B   ┆ 2   ┆ 2   │|
|└───────┴─────┴─────┴─────┘|
└───────────────────────────┘
```

### Expected
`duckdb`, `ibis`, `sqlframe`, `dask` all produce this - which is an index ordered by `"c"`

```py
df.lazy("duckdb").with_row_index(order_by="c").sort("index").collect("polars")
```

```
┌───────────────────────────┐
|    Narwhals DataFrame     |
|---------------------------|
|shape: (3, 4)              |
|┌───────┬─────┬─────┬─────┐|
|│ index ┆ a   ┆ b   ┆ c   │|
|│ ---   ┆ --- ┆ --- ┆ --- │|
|│ i64   ┆ str ┆ i64 ┆ i64 │|
|╞═══════╪═════╪═════╪═════╡|
|│ 0     ┆ B   ┆ 2   ┆ 2   │|
|│ 1     ┆ A   ┆ 3   ┆ 4   │|
|│ 2     ┆ A   ┆ 1   ┆ 9   │|
|└───────┴─────┴─────┴─────┘|
└───────────────────────────┘
```

MarcoGorelli:
closes #3289

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_arrow/dataframe.py
+++ b/narwhals/_arrow/dataframe.py
@@ -533,7 +533,7 @@ class ArrowDataFrame(
 
     def with_row_index(self, name: str, order_by: Sequence[str] | None) -> Self:
         plx = self.__narwhals_namespace__()
-        data = arange(0, len(self), 1, self._backend_version)
+        data = arange(0, len(self), 1)
         if order_by is None:
             row_index = plx._expr._from_series(
                 plx._series.from_iterable(data, context=self, name=name)


narwhals/_arrow/dataframe.py
```diff
@@ -531,18 +532,23 @@ def to_dict(
...
             row_index = plx._expr._from_series(
-                self.select(row_index, *(plx.col(x) for x in order_by))
-                .sort(*order_by, descending=False, nulls_last=False)
-                .get_column(name)
+                plx._series.from_iterable(data, context=self, name=name)
+            )
+        else:
+            rank = cast(
+                "ArrowExpr",
+                nw_col(order_by[0]).rank(method="ordinal")._to_compliant_expr(plx),
```
dangotbanned: Oops, I commented on the wrong branch the first time 😂 

Trying again ...


This is relying on some newer `pyarrow` features, but might perform differently to `rank`?

```py
def with_row_index(native: pa.Table, name: str, order_by: Sequence[str]) -> pa.Table:
    indices = pc.sort_indices(native, [(by, "ascending") for by in order_by])
    return native.add_column(
        0, name, pc.scatter(pa.arange(0, len(native)), indices.cast(pa.int64()))
    )
```

MarcoGorelli: thanks, that does indeed perform better (and is correct, as far as I can tell!)

narwhals/_arrow/dataframe.py
```diff
@@ -28,6 +29,7 @@
 )
 from narwhals.dependencies import is_numpy_array_1d
 from narwhals.exceptions import ShapeError
+from narwhals.functions import col as nw_col
```
dangotbanned: This seems a bit fishy to me, since it wasn't here before (#3239)

I can't find any other cases where a backend is using narwhals-level APIs like this.
What's going on 🤔?

MarcoGorelli: It's an easy way to produce a compliant expr with all the metadata it needs

narwhals/_arrow/dataframe.py
dangotbanned: @MarcoGorelli I could be wrong, but this seems like it should be faster than the `rank` branch:

[perf: Massively simplify `pyarrow<20` branch](https://github.com/narwhals-dev/narwhals/pull/3295/commits/6fe0af8a3070c616b61c389edc7822ad0b367ae0)

I thought I'd try adapting `ArrowSeries.scatter` until I got the same behavior as `pc.scatter`:


<details><summary>Show <code>ArrowSeries.scatter</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/d9ca87a584b34aa3cab7682b2f4e70e2c9c2259e/narwhals/_arrow/series.py#L453-L483

</p>
</details>

And I landed on this 😏 


https://github.com/narwhals-dev/narwhals/blob/6fe0af8a3070c616b61c389edc7822ad0b367ae0/narwhals/_plan/arrow/functions.py#L300-L302

MarcoGorelli: nice, thanks!

MarcoGorelli:
shipping then, thanks for your help and review! 🙏

PR #3306, opened 2025-11-14
chore: Add project classifiers in pyproject.toml

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

---

It's not much

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ requires-python = ">=3.9"
 authors = [
   { name = "Marco Gorelli", email = "hello_narwhals@proton.me" },
 ]
-license = {file = "LICENSE"}
+license = {file = "LICENSE.md"}
 description = "Extremely lightweight compatibility layer between dataframe libraries"
 readme = "README.md"
 keywords = ["dataframes", "interoperability", "pandas", "polars", "pyarrow", "dask", "modin", "cudf"]


pyproject.toml
```diff
@@ -11,13 +11,22 @@ requires-python = ">=3.9"
...
-  "Programming Language :: Python",
   "Operating System :: OS Independent",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: Free Threading :: 2 - Beta",
```
FBruzzesi, line 29: We are not really testing anything in a free threading way. We mostly rely on underlying libraries to be compliant, but we don't have test for anything narwhals introduces

MarcoGorelli:
thanks - ok with the license, but i don't think we need to list all the python versions, in all likelihood any version of narwhals will be forwards-compatible with new versions of python

FBruzzesi:
> thanks - ok with the license, but i don't think we need to list all the python versions, in all likelihood any version of narwhals will be forwards-compatible with new versions of python

I think it's nice to have, especially for versions that reached EOL. What should be assumed for a generic `"Programming Language :: Python :: 3"`? For instance, we still support 3.9 which is EOL, but in some time that will be no longer the case. It's very low effort to keep these in sync.

MarcoGorelli (approved):
thanks @FBruzzesi

PR #3315, opened 2025-11-18
feat: Add `{Expr,Series}.any_value`

### LINKED ISSUE #3298: [Enh]: Add `Expr.any_value`?

https://github.com/anam-org/metaxy

I would like to perform a group by on a DataFrame and take a random (any) row from each group. 

I could have used `df.group_by().agg(nw.all().first())`, because rows within groups are not sorted already, but this currently fails with: 

```
E           narwhals.exceptions.InvalidOperationError: Order-dependent expressions are not supported for use in LazyFrame.
E
E           Hint: To make the expression valid, use `.over` with `order_by` specified.
E
E           For example, if you wrote `nw.col('price').cum_sum()` and you have a column
E           `'date'` which orders your data, then replace:
E
E              nw.col('price').cum_sum()
E
E            with:
E
E              nw.col('price').cum_sum().over(order_by='date')
E                                       ^^^^^^^^^^^^^^^^^^^^^^
E
E           See https://narwhals-dev.github.io/narwhals/concepts/order_dependence/.
```

In my case performing a sort would introduce a performance penalty, so I would like to avoid it.

It seems like the best I could do now is with `.is_first_distinct()` over a dummy column: 

```py
def group_and_take_any(df: FrameT, cols: Sequence[str]) -> FrameT:
     return (
            df.with_columns(nw.lit(True).alias("_dummy"))
            .filter(
                nw.col("_dummy")
                .is_first_distinct()
                .over(*cols, order_by="_dummy")
            )
            .drop("_dummy")
        )
```

_No response_

_No response_

_No response_

FBruzzesi:

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3298

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,13 +11,22 @@ requires-python = ">=3.9"
 authors = [
   { name = "Marco Gorelli", email = "hello_narwhals@proton.me" },
 ]
+license = {file = "LICENSE.md"}
 description = "Extremely lightweight compatibility layer between dataframe libraries"
 readme = "README.md"
 keywords = ["dataframes", "interoperability", "pandas", "polars", "pyarrow", "dask", "modin", "cudf"]
 classifiers = [
+  "Development Status :: 5 - Production/Stable",
   "License :: OSI Approved :: MIT License",
-  "Programming Language :: Python",
   "Operating System :: OS Independent",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
+  "Programming Language :: Python :: Free Threading :: 2 - Beta",
   "Typing :: Typed",
 ]
 


MarcoGorelli:
thanks!

some initial thoughts:

- I think the function just needs to not make any guarantees about order, I don't think we need to intentionally add randomness, just taking `first` for eager backends should be enough (it's likely more performant anyway, right?)
- i'm slightly concerned that sql ignores nulls in this function, but we're suggesting to not to. if polars were to add this method, would they ignore nulls or not? perhaps we could have a mandatory `ignore_nulls` keyword argument, similar to how we have it in `any_horizontal`?

FBruzzesi:
Thanks @MarcoGorelli 

> * I think the function just needs to not make any guarantees about order, I don't think we need to intentionally add randomness, just taking `first` for eager backends should be enough (it's likely more performant anyway, right?)

Wouldn't using `first` actually make a lot of guarantees on order for eager backends? 

> * i'm slightly concerned that sql ignores nulls in this function, but we're suggesting to not to. if polars were to add this method, would they ignore nulls or not? perhaps we could have a mandatory `ignore_nulls` keyword argument, similar to how we have it in `any_horizontal`?

Sure I will add such option 🙏🏼

MarcoGorelli:
> Wouldn't using first actually make a lot of guarantees on order for eager backends?

Sure, but we don't guarantee it - an in, it happens to return the first value, but it's not guaranteed (and the name `any_value` strongly implies there's nothing guaranteed)

Polars does something similar in joins - in the eager case it may happen to return rows in the same order when it's more efficient to do so, but it's not guaranteed (and it's documented that it's not guaranteed)

narwhals/_ibis/expr.py
```diff
@@ -127,6 +127,12 @@ def _last(self, expr: ir.Value, *order_by: str) -> ir.Value:
             order_by=self._sort(*order_by), include_null=True
         )
 
+    def _any_value(self, expr: ir.Value, *, ignore_nulls: bool) -> ir.Value:
+        # !NOTE: ibis arbitrary returns a random non-null value
+        # See: https://ibis-project.org/reference/expression-generic.html#ibis.expr.types.generic.Column.arbitrary
+        expr = cast("ir.Column", expr)
+        return expr.arbitrary() if ignore_nulls else expr.first(include_null=True)
```
FBruzzesi, line 134: `expr.first(include_null=not ignore_nulls)` is not behaving as expected 🫠

narwhals/_pandas_like/expr.py
```diff
@@ -109,7 +109,13 @@ def window_kwargs_to_pandas_equivalent(  # noqa: C901
...
             "ignore_na": kwargs["ignore_nulls"],
         }
-    elif function_name in {"first", "last"}:
+    elif function_name in {"first", "last", "any_value"}:
+        if kwargs.get("ignore_nulls"):
+            msg = (
+                "`Expr.any_value(ignore_nulls=True)` is not supported in a `over` "
+                "context for pandas-like backend."
+            )
+            raise NotImplementedError(msg)
```
FBruzzesi, line 118: This is an issue if a group has _all_ null values

narwhals/_pandas_like/group_by.py
```diff
@@ -191,6 +195,12 @@ def native_agg(self) -> _NativeAgg:
         native_name = PandasLikeGroupBy._remap_expr_name(self.leaf_name)
         last_node = next(self.expr._metadata.op_nodes_reversed())
         if self.leaf_name in _REMAP_ORDERED_INDEX:
+            if last_node.kwargs.get("ignore_nulls"):
+                msg = (
+                    "`Expr.any_value(ignore_nulls=True)` is not supported in a `group_by` "
+                    "context for pandas-like backend"
+                )
+                raise NotImplementedError(msg)
```
FBruzzesi, line 203: This is an issue if a group has _all_ null values

MarcoGorelli:
thanks for updating, will try to review next week

FBruzzesi:
@MarcoGorelli just to make sure we align before going on a tangent. We briefly touched the possibility of _explicitly_ marking this feature as unstable as it's outside of polars API. Next steps would be:

- Simple warning in the main namespace _docstrings_
- Code/runtime warning in the stable namespaces OR even raise an error in the stable namespaces

Let me know what your thoughts are :) 

---

Update: Ok I pushed a proposal in https://github.com/narwhals-dev/narwhals/pull/3315/commits/e8d44f6033cdfcdab477e78ab41b804009d7d98f

PR #3325, opened 2025-11-25
feat(expr-ir): Finish* implementing `ArrowExpr`

dangotbanned:
## Related issues

- Child of #2572
- Parent of 
  - #3347
  - #3349

### Tracking
- [ ] https://github.com/apache/arrow/issues/48477

## Description

This PR is a bit of a mixed bag.

I've really tried to focus on getting *an* implementation for all of the methods in the `Expr` namespace. 
Man there a **lot** of them now 😅

I have a pretty long list, so things are grouped and highlighted to what was the most interesting *to me*. 

> [!TIP]
> ✨ - New **feature** (either for `pyarrow` as a backend or `narwhals` itself)
> 💾 - **Refactor** from `main` (particularly trying to avoid *unconditional-dependence* on `numpy`)

<details><summary><b>Show top-level functions</b></summary>
<p>

- `{all,any}_horizontal(ignore_nulls=...)`
- `coalesce`
- `format`
- `linear_space` (✨, see #3349)

</p>
</details>

<details><summary><b>Show <code>Expr</code> methods</b></summary>
<p>

- `ceil`
- `clip`, `clip_lower`, `clip_upper`
- `drop_nulls`
- `exp`
- `fill_nan`
- `fill_null`
- `fill_null_with_strategy` (💾)
  - Related #3327
- `floor`
- `gather_every` (deprecated)
- `hist` (✨)
  - *Almost all* `numpy` usage avoided
  - Returns as struct when needed
- `is_{duplicated,unique}` (💾, ✨)
  - Supports `over(*partition_by)`
- `is_not_null` (✨)
- `is_not_nan` (✨)
- `{kurtosis,skew}`, (💾, ✨)
  - Supports `over(*partition_by)` and `group_by`
- `rolling_expr` (💾, cc francesco)
  - `rolling_sum`
  - `rolling_mean`
  - `rolling_var`
  - `rolling_std`
- `log`
- `map_batches(is_elementwise=..., returns_scalar=...)` (✨)
- `mode(keep: ModeKeepStrategy)` (💾)
- `replace_strict`
- `replace_strict(default=...)`
- `round`
- `sample` (deprecated)
  - Avoids `numpy` when `with_replacement=True`
- `sqrt`
- `unique`

</p>
</details>

<details><summary><b>Show <code>Expr*Namespace</code> methods</b></summary>
<p>

- `cat.get_categories` (💾)
- `list.contains` (✨)
  - Also supports scalar `Expr`
- `list.get`
- `list.join` (✨)
[...]

Issue #3204, opened 2025-10-13
Drop support for Python `3.9`


dangotbanned:
## Description
`3.9` EOL is scheduled for [2025-10-31](https://endoflife.date/python).

There's some select features mentioned in (https://github.com/narwhals-dev/narwhals/issues/2084#issuecomment-2925189450), and the full [What’s New In Python 3.10](https://docs.python.org/3.10/whatsnew/3.10.html) for more info.

Probably will make sense to time this around/shortly after (#3184).

# Related
## Meta
- [x] `3.8`
  - [x] #2084
  - [x] #2638
- [ ] `3.14`
  - [ ] #3184
- [x] #3405

## [Critical] Bugs
Fixes have been available since [v2.13.0](https://github.com/narwhals-dev/narwhals/releases/tag/v2.13.0)
- [x] #3307
- [x] #3311 
- [x] #3300

## Backends
- [x] https://github.com/pandas-dev/pandas/pull/62066
- [x] https://github.com/pola-rs/polars/pull/25984
- [x] https://github.com/apache/arrow/pull/47478
- [x] https://github.com/rapidsai/cudf/pull/21225
- [x] https://github.com/apache/spark/pull/51631
- [x] https://github.com/apache/spark/pull/52911
- [x] https://github.com/dask/dask/pull/11245
- [x] https://github.com/ibis-project/ibis/commit/aa16d0885ff62ac92ec1027a245158896f74cc0d
- [x] `duckdb`
- [ ] `sqlframe`
- [ ] `modin` [^1]

[^1]: `modin` appears to have only been receiving bug-fixes by Snowflake for a while now [modin-project/modin/releases](https://github.com/modin-project/modin/releases).

## Ecosystem
As a project aimed at the maintainers of other projects - tooling matters too!
- [x] https://github.com/pytest-dev/pytest/pull/13724
- [x] https://github.com/python/mypy/pull/20156
- [ ] https://github.com/python/typeshed/issues/13782
- [x] https://github.com/pypa/hatch/releases/tag/hatch-v1.16.0

## Meta TODO
- [ ] Figure out new upstream minimums ([see `3.8` drop table](https://github.com/narwhals-dev/narwhals/issues/2084#issuecomment-2833356123))

[Critical]: https://github.com/narwhals-dev/narwhals/issues/3204#issuecomment-3541281889

FBruzzesi:
[...]

PR #3332, opened 2025-11-28
feat: add list aggregate methods

raisadz:
# Description

The following list methods are implemented:

        - list.max
        - list.mean
        - list.median
        - list.min
        - list.sum

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_arrow/utils.py
+++ b/narwhals/_arrow/utils.py
@@ -503,7 +503,7 @@ def list_agg(
 ) -> ChunkedArrayAny:
     lit_: Incomplete = lit
     aggregation = (
-        ("values", "sum", pc.ScalarAggregateOptions(min_count=0))
+        ("values", func, pc.ScalarAggregateOptions(min_count=0))
         if func == "sum"
         else ("values", func)
     )


narwhals/_spark_like/expr_list.py
```diff
@@ -33,3 +34,21 @@ def _get(expr: Column) -> Column:
...
+    def max(self) -> SparkLikeExpr:
+        def func(expr: Column) -> Column:
+            F = self.compliant._F
+            return F.array_max(F.array_compact(expr))
+
+        return self.compliant._with_elementwise(func)
+
+    mean = not_implemented()
+    median = not_implemented()
+    sum = not_implemented()
```
FBruzzesi: `mean` and `sum` can be implemented quite directly with [`F.aggregate`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.aggregate.html):

```py
def sum(self) -> SparkLikeExpr:
    def func(expr: Column) -> Column:
        F = self.compliant._F
        return F.aggregate(expr, F.lit(0.0), operator.add)
    
    return self.compliant._with_elementwise(func)

def mean(self) -> SparkLikeExpr:
    def func(expr: Column) -> Column:
        F = self.compliant._F
        return F.aggregate(expr, F.lit(0.0), operator.add) / F.array_size(expr)
    
    return self.compliant._with_elementwise(func)
```

Calculating the median row-wise is trickier because it requires sorting but it would still be doable (at least in a simplified way):

```py
def median(self) -> SparkLikeExpr:
    def func(expr: Column) -> Column:
        F = self.compliant._F
        sorted_expr = F.sort_array(expr)
        size = F.array_size(expr)
        mid_index = (size / 2).cast("int")
        odd_case = sorted_expr.getItem(mid_index)
        even_case = (sorted_expr.getItem(mid_index-1) + sorted_expr.getItem(mid_index))/2
        return F.when(size % 2 == 1, odd_case).otherwise(even_case)
    return self.compliant._with_elementwise(func)
```

> [!CAUTION]
>  Careful consideration for nulls are required

FBruzzesi (commented):
Thanks @raisadz - this looks amazing! Just left a comment to possibly support more in the spark-like case 😇

narwhals/_spark_like/expr_list.py
raisadz: Awesome, thanks a lot for the suggestion @FBruzzesi ! I added those functions with handling nulls with `array_compact`. Looks that there is some issue for median for pyspark[connect] but pyspark passes

narwhals/_arrow/utils.py
```diff
@@ -494,3 +495,18 @@ def arange(start: int, end: int, step: int) -> ArrayAny:
...
+) -> ChunkedArrayAny:
+    return (
+        pa.Table.from_arrays(
+            [pc.list_flatten(array), pc.list_parent_indices(array)],
+            names=["values", "offsets"],
+        )
+        .group_by("offsets")
+        .aggregate([("values", func)])
+        .column(f"values_{func}")
+    )
```
dangotbanned, line 519: @raisadz I'm pretty excited by this! 😄 

+1 from me on (https://github.com/narwhals-dev/narwhals/pull/3332#pullrequestreview-3519374373)

---

I've just tried this out with the test case for `list.unique`:

https://github.com/narwhals-dev/narwhals/blob/2c6c3c3245536d52ca255e14630a02b0e819244d/tests/expr_and_series/list/unique_test.py#L13

The result for that *should* be:

```
[[None, 2, 3], None, [], [None]]
```

But using `list_agg` seems to have dropped 2/4 lists and all nulls 🤔 

```py

import pyarrow as pa

data = {"a": [[2, 2, 3, None, None], None, [], [None]]}
ca = pa.chunked_array([pa.array(data["a"])])
result = list_agg(ca, "distinct").to_pylist()
print(result)
```
```
[[2, 3], []]
```

---

I managed to get *slightly* closer to what we want, by passing in `options` for the `group_by`:

<details><summary>Show <code>list_agg_opts</code></summary>
<p>

```py
from typing import Any

import pyarrow as pa
import pyarrow.compute as pc

def list_agg_opts(
    array: pa.ChunkedArray[Any], func: Any, options: Any = None
) -> pa.ChunkedArray[Any]:
    return (
        pa.Table.from_arrays(
            [pc.list_flatten(array), pc.list_parent_indices(array)],
            names=["values", "offsets"],
        )
        .group_by("offsets")
        .aggregate([("values", func, options)])  # <-------
        .column(f"values_{func}")
    )
```

</p>
</details> 

These are the correct results for 2/4 of the lists 🎉 

But where did the other 2 go? 😳 

```py
result = list_agg_opts(ca, "distinct", pc.CountOptions("all")).to_pylist()
print(result)
```
```
[[2, 3, None], [None]]
```


**Edit**: I missed it myself lol, fixed in (https://github.com/narwhals-dev/narwhals/pull/3325/commits/d8363e1f0dc89bf43e0d3adb9a6eb8b36ae067a0)

raisadz:
Thank you both! Yes in fact those tests with empty lists or all Nones lists fail for multiple backends. I pushed some changes but this is wip and I will continue to work on that

raisadz:
> Thanks @raisadz, I have a couple more comments, apologies for the fragmented review 🙈
> 
> 1. Could you add a couple of test cases:
>    a. All nulls in list
>    b. Empty list
>    c. `polars.Expr.list.sum` says: _If there are no non-null elements in a row, the output is 0._ For the other aggregations it's unclear what the output should be, and I wonder how consistent it is across all different backends.
> 2. Could you mix the docstring examples a bit other than polars?

Thank you @FBruzzesi, I adjusted my code for the edge cases now. Could you please review it when you have time?

Mixing docstrings was also very useful as I found that pyarrow `approximate_median` sometimes has a different behaviour from other backends (I added a test for that). For the non-list implementation of `median` for pyarrow the same function `approximate_median` is used. I tried replacing it with  `quantile` instead which is listed in the documentation https://arrow.apache.org/docs/python/generated/pyarrow.compute.quantile.html and should have q=0.5 by default, but I had an error that this function wasn't found

narwhals/_arrow/utils.py
```diff
@@ -501,12 +502,33 @@ def list_agg(
     array: ChunkedArrayAny,
     func: Literal["min", "max", "mean", "approximate_median", "sum"],
 ) -> ChunkedArrayAny:
-    return (
+    if func == "approximate_median" and sys.version_info < (3, 10):  # pragma: no cover
+        msg = f"The minimum supported Python version for {func}"
```
FBruzzesi: This message reads a bit odd to me, is something missing?

narwhals/_arrow/utils.py
```diff
@@ -501,12 +502,33 @@ def list_agg(
...
+    agg = pa.array(
         pa.Table.from_arrays(
             [pc.list_flatten(array), pc.list_parent_indices(array)],
             names=["values", "offsets"],
         )
         .group_by("offsets")
         .aggregate([("values", func)])
+        .sort_by("offsets")
         .column(f"values_{func}")
     )
```
FBruzzesi, line 519: This is a comment rather than anything else, but I want mention it - maybe someone has ideas.

In the case of multiple expressions on the same list column, we would end up running the same (expensive?) operations (flatten + group by + sort). I wonder if for the case of expressions there is a way to execute them together.

In code, I would the following:

```py
frame.select(
    a_sum = nw.col("a").list.sum(),
    a_mean = nw.col("a").list.mean()
)
```

to execute as:

```py
agg = pa.array(
    pa.Table.from_arrays(
        [pc.list_flatten(array), pc.list_parent_indices(array)],
        names=["values", "offsets"],
    )
    .group_by("offsets")
    .aggregate([("values", "sum"), ("values", "mean")])
    .sort_by("offsets")
)
... # Here get the 2 columns
```

narwhals/_arrow/utils.py
raisadz: Thanks @FBruzzesi, yes sorry missed the sentence ending, updated it now 😅

FBruzzesi:
Hey @raisadz - I will try to do a final pass today for this PR. Sorry for the slow iteration loop 🫠

narwhals/_arrow/utils.py
```diff
@@ -494,3 +496,43 @@ def arange(start: int, end: int, step: int) -> ArrayAny:
...
+    array: ChunkedArrayAny,
+    func: Literal["min", "max", "mean", "approximate_median", "sum"],
+) -> ChunkedArrayAny:
+    version = sys.version_info
+    if func == "approximate_median" and version < (3, 10):  # pragma: no cover
+        msg = (
+            f"The minimum supported Python version for {func} is 3.10."
+            f"\nGot: {version.major}.{version.minor}.{version.micro}."
+        )
+        raise NotImplementedError(msg)
```
FBruzzesi: I double checked what's happening here - I was a bit skeptical because there is no error for `{Series, Expr}.median()`.

For pyarrow itself this is not an issue, you can remove the exception and the tests will pass. Then also `pandas` (and `pandas[pyarrow]`) will pass (assuming pandas>=2.2).

So in summary, I am just a bit confused about this error and when it shows up. I think we should try to remove it and see what happens on CI

narwhals/_spark_like/expr_list.py
```diff
@@ -33,3 +34,51 @@ def _get(expr: Column) -> Column:
...
+
+    def median(self) -> SparkLikeExpr:
+        def func(expr: Column) -> Column:  # pragma: no cover
+            # sqlframe issue: https://github.com/eakmanrq/sqlframe/issues/548
+            F = self.compliant._F
+            sorted_expr = F.array_compact(F.sort_array(expr))
+            size = F.array_size(sorted_expr)
+            mid_index = (size / 2).cast("int")
+            odd_case = sorted_expr[mid_index]
+            even_case = (sorted_expr[mid_index] - 1 + sorted_expr[mid_index]) / 2
```
FBruzzesi:
```suggestion
            even_case = (sorted_expr[mid_index - 1] + sorted_expr[mid_index]) / 2
```

narwhals/_arrow/utils.py
raisadz: Thanks @FBruzzesi ! Yes it seems that there is no need for the minimum Python version and the tests pass. The current CI failures seem to be unrelated to this PR

narwhals/_arrow/utils.py
dangotbanned: > maybe someone has ideas.

100% agree on the idea, but probably too complex in current `narwhals` IMO.

This issue isn't unique to `list.*` expressions, we'd have a similar problem for these:

### Example

Efficient:

```py
frame.select(
    nw.col("a", "b").sum().over("c"),
)
```

Same intention, but worse performance for `pandas`, `pyarrow` vs the first spelling:

```py
frame.select(
    nw.col("a").sum().over("c"),
    nw.col("b").sum().over("c")
)
```

### Why are these too hard?
I think what you're describing corresponds to common subplan/expression elimination (see [`datafusion` optimizer rules](https://docs.rs/datafusion-optimizer/latest/datafusion_optimizer/#modules)).

The main challenges are:
- knowing that `list.sum` becomes `op_1, ..., op_n -> group_by -> sort` is an implementation detail (+ specific to `pyarrow`)
  - not something that an `Expr` can *"learn"*
- Expressions don't *"know"* anything about the context they're evaluated in
  - Are we in `select`?
  - What other expressions come before/after me?
  - Am I doing work that can be reused?

@FBruzzesi if this sounds similar to what I mentioned RE (#2572) and a `LogicalPlan` - you'd have the right hunch 😄

Basically, the problem would be much easier to generalize if we had:

```py
frame.select(
    a_sum = nw.col("a").list.sum(),
    a_mean = nw.col("a").list.mean()
)
```

Outputs:

```py
nw_plan = Select(
    inputs=[
        Alias(name="a_sum", expr=ListSum(Column(name="a"))),
        Alias(name="a_mean", expr=ListMean(Column(name="a"))),
    ]
)
```

But then we transform the `pyarrow` plan (expressions) into:

<details><summary>Warning - verbosity overload and invented data model</summary>
<p>

```py
pyarrow_list_sum = Sort(
    inputs=[
        GroupBy(
            inputs=[
                HConcat(
                    inputs=[ListExplode(Column(name="a")), ListOffsets(Column(name="a"))],
                    names=["values", "offsets"],
                )
            ],
            keys=["offsets"],
            aggs=[Alias(name="a_sum", expr=Sum(Column(name="value")))],
        )
    ],
    by=["offsets"],
)


pyarrow_list_mean = Sort(
    inputs=[
        GroupBy(
            inputs=[
                HConcat(
                    inputs=[ListExplode(Column(name="a")), ListOffsets(Column(name="a"))],
                    names=["values", "offsets"],
                )
            ],
            keys=["offsets"],
            aggs=[Alias(name="a_mean", expr=Mean(Column(name="value")))],
        )
    ],
    by=["offsets"],
)
```

</p>
</details> 

Then detecting what is shared between `list.sum` and `list.mean` is actually less complex than you might think.
And as a bonus, we'd have something that is generalized to expressions that rely on `group_by` 🙂

FBruzzesi:
@raisadz the action failing for python 3.9 seems to be on windows only (I cannot replicate that on mac nor ubuntu, but I don't have a windows machine to try it). I would say let's xfail such case and consider:

- reporting it upstream
- issue a warning from our side? I am 50-50 split on this

tests/expr_and_series/list/max_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+import pytest
+
+import narwhals as nw
+from tests.utils import PANDAS_VERSION, assert_equal_data
+
+if TYPE_CHECKING:
+    from tests.utils import Constructor, ConstructorEager
+
+data = {"a": [[3, None, 2, 2, 4, None], [-1], None, [None, None, None], []]}
```
dangotbanned, line 13: I've been a bit afraid to bring this up, but AFAICT all of the `list.*` methods have only been tested against 1 level of nesting.

I think 2 (or more) levels *might* be a problem for `pyarrow` and `pandas`, because lists aren't hashable yet.

I'd be really happy to wrong on this though 🙏

If it is an issue, I don't think it needs to be a blocker - just something to keep in mind 🙂

FBruzzesi: polars seems to properly support these ops only for `List(<numeric_type>)` (and maybe temporal types?) and either fail or return nulls otherwise:

```py
import polars as pl

data = [
    [[1], [2,3]],
    [[4,5,6], [7,8]]
]

series = pl.Series(data)
print(series.dtype)
print()
for op in ("min", "mean", "max", "median", "sum"):
    print(f"Executing {op}")
    try:
        print("result", getattr(series.list, op)())
    except Exception as exc:
        print("error", exc)
    print()
```

```terminal
List(List(Int64))

Executing min
error `min` operation not supported for dtype `list[i64]`

Executing mean
result shape: (2,)
Series: '' [f64]
[
        null
        null
]

Executing max
error `max` operation not supported for dtype `list[i64]`

Executing median
result shape: (2,)
Series: '' [f64]
[
        null
        null
]

Executing sum
error `sum` operation not supported for dtype `list[i64]`

Hint: you may mean to call `concat_list`
```

There are a few other cases for which we check the dtype before performing an operation. We might do the same here

raisadz:
> @raisadz the action failing for python 3.9 seems to be on windows only (I cannot replicate that on mac nor ubuntu, but I don't have a windows machine to try it). I would say let's xfail such case and consider:
> 
> * reporting it upstream
> * issue a warning from our side? I am 50-50 split on this

Thanks @FBruzzesi ! I missed this failure and I am skipping it now. I don't think we should worry about 3.9 version on windows as it is almost obsolete at this point

dangotbanned:
> I don't think we should worry about 3.9 version on windows as it is almost obsolete at this point

This might be good to mention in #3204?

tests/expr_and_series/list/max_test.py
FBruzzesi: @raisadz I think this is the last missing bit before merging. If we can standardize the error message for all backends it would be amazing. I would be in favor in raising for `list.{mean,median}` as well, and ask in polars if the current output is expected.

Update1: Asked in their [discord channel](https://discord.com/channels/908022250106667068/957930511999832064/1449010385285025824)

Update2: reply:

> it's an older leftover I think
> unsupported aggregates used to return None and we've been slowly transitioning over time to nice errors

narwhals/_arrow/utils.py
```diff
@@ -494,3 +495,36 @@ def arange(start: int, end: int, step: int) -> ArrayAny:
...
+        .column(f"values_{func}")
+    )
+    non_empty_mask = pa.array(pc.not_equal(pc.list_value_length(array), lit(0)))
+    if func == "sum":
+        agg = agg.fill_null(lit(0))  # pyright:ignore[reportArgumentType]
+        base_array = pc.if_else(non_empty_mask.is_null(), None, 0)
+    else:
+        base_array = pc.if_else(
+            non_empty_mask, 0, None
+        )  # zero is just a placeholder which is replaced below
```
MarcoGorelli: does it work to just do
```py
        base_array = pa.repeat(lit(None, type=agg.type), len(array))
```
?

the `0` placeholder feels a bit magical

MarcoGorelli: here's something i worked through in the live stream, could it work?

```py
    lit_: Incomplete = lit
    aggregation = ('values', 'sum', pc.ScalarAggregateOptions(min_count=0)) if func == 'sum' else ('values', func)
    agg = pa.array(
        pa.Table.from_arrays(
            [pc.list_flatten(array), pc.list_parent_indices(array)],
            names=["values", "offsets"],
        )
        .group_by("offsets")
        .aggregate([aggregation])
        .sort_by("offsets")
        .column(f"values_{func}")
    )
    non_empty_mask = pa.array(pc.not_equal(pc.list_value_length(array), lit(0)))
    if func == "sum":
        # Make sure sum of empty list is 0.
        base_array = pc.if_else(non_empty_mask.is_null(), None, 0)
    else:
        base_array = pa.repeat(lit_(None, type=agg.type), len(array))
    return pa.chunked_array(
        [
            pc.replace_with_mask(
                base_array,
                non_empty_mask.fill_null(False),  # type: ignore[arg-type]
                agg,
            )
        ]
    )
```

dangotbanned: I've got like 97 variations of this now 😂

Pretty much everything that involves `list.*` needs to do this dance around `[]` and `None`

<details><summary>Show rabbit hole</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/2d84d14e86b9931a0bf3d1a3270273cd000ed46a/narwhals/_plan/arrow/functions.py#L396-L795

</p>
</details>

tests/expr_and_series/list/max_test.py
MarcoGorelli: > I think 2 (or more) levels might be a problem for pyarrow and pandas, because lists aren't hashable yet.

tbh i wouldn't worry about it, just leave it to each backend

raisadz:
Thanks for the suggestion @MarcoGorelli ! I applied it now to the list_agg function

tests/expr_and_series/list/median_test.py
```diff
@@ -0,0 +1,80 @@
...
+        if PANDAS_VERSION < (2, 2):
+            pytest.skip()
+        pytest.importorskip("pyarrow")
+    if (
+        any(backend in str(constructor) for backend in ("pandas", "pyarrow"))
+        and sys.version_info < (3, 10)
+        and is_windows
+    ):  # pragma: no cover
+        reason = "The issue only affects old Python versions on Windows."
+        pytest.skip(reason=reason)
```
dangotbanned, line 39: Sorry to have caught this so late, I just noticed in (https://github.com/narwhals-dev/narwhals/compare/oh-nodes...expr-ir/list-agg) 

Since `is_windows` is a **function** ...

<details><summary>Show <code>is_windows</code></summary>
<p>

image

</p>
</details> 

... this condition is always `True`:

```py
from tests.utils import is_windows

bool(is_windows)
```
```
True
```

Meaning that all platforms skip on `sys.version_info < (3, 10)` 😱

I usually prefer `xfail` to `skip` since the former is the only one that'll tell you when something's amiss 😉

MarcoGorelli: good catch, thanks!

tests/expr_and_series/list/median_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+import narwhals as nw
+from tests.utils import PANDAS_VERSION, POLARS_VERSION, assert_equal_data, is_windows
+
+if TYPE_CHECKING:
+    from tests.utils import Constructor, ConstructorEager
+
+data = {"a": [[3, None, 2, 2, 4, None], [-1], None, [None, None, None], [], [3, 4, None]]}
+expected = [2.5, -1, None, None, None, 3.5]
+expected_pyarrow = [2.5, -1, None, None, None, 3]
```
dangotbanned, line 17: @raisadz @FBruzzesi (re https://github.com/narwhals-dev/narwhals/pull/3332#issuecomment-3596671034, https://github.com/narwhals-dev/narwhals/pull/3332#pullrequestreview-3526230489)

I've done some experimenting and think I've found what the `pyarrow` issue is:

```py
import pyarrow as pa
import pyarrow.compute as pc


def median(*values: float | None) -> pa.DoubleScalar:
    return pc.approximate_median(pa.array(values, pa.float64()))


def median_pretty(*values: float | None) -> None:
    print(f"median({list(values)!a:21}) = {median(*values)}")
```

I wonder if you can spot it too 😄:

```py
median_pretty()
median_pretty(3)
median_pretty(3, 4)
median_pretty(3, 4, None)
median_pretty(3, 4, None, None)
median_pretty(3, 4, None, None, 5)
median_pretty(3, 4, 5)
median_pretty(5, 3, 4)
median_pretty(5, 3)
median_pretty(5, 2)
median_pretty(5, 2, 50)
median_pretty(None, 2, 50)
median_pretty(None, 2, 50, 2)
median_pretty(50, 2, 50, 2)
```

```
median([]                   ) = None
median([3]                  ) = 3.0
median([3, 4]               ) = 3.0
median([3, 4, None]         ) = 3.0
median([3, 4, None, None]   ) = 3.0
median([3, 4, None, None, 5]) = 4.0
median([3, 4, 5]            ) = 4.0
median([5, 3, 4]            ) = 4.0
median([5, 3]               ) = 3.0
median([5, 2]               ) = 2.0
median([5, 2, 50]           ) = 5.0
median([None, 2, 50]        ) = 2.0
median([None, 2, 50, 2]     ) = 2.0
median([50, 2, 50, 2]       ) = 26.0
```

tests/expr_and_series/list/median_test.py
raisadz: Thanks @dangotbanned ! I added a follow-up PR to fix this https://github.com/narwhals-dev/narwhals/pull/3354

Issue #79, opened 2024-05-04
Support for categorical data types


cbourjau:
Apologies if I didn't find existing information about this! I am wondering if future support for categorical data types is in the envisioned scope of this very interesting project.

MarcoGorelli:
Definitely!   Thanks for your interest

MarcoGorelli:
@cbourjau just to know how to prioritise, do you have a use case where categorical data type is needed, or is this more of a general question?

cbourjau:
Our feature engineering pipelines regularly process categorical data. Currently, they are built around Pandas, but I'd love to migrate them to Narwahls. The lack of categorical data types is thus far the only blocker I can see in our case.

MarcoGorelli:
Cool - fancy sharing an example of what you do with categorical data? If we have an example of a dataframe-agnostic function you'd like to write, it'll be much easier

cbourjau:
It is not trivial to say exactly which part of the Pandas API we use throughout our codebases with respect to categorical data types. However, I just noticed that narwhals is not complaining when presented with a data frame that has a categorical column - it is quietly passed through (which makes sense):

```python
import pandas as pd
import narwhals as nw

pdf = pd.DataFrame({"foo": ["bar", "baz"]}).astype({"foo": "category"})
ndf = nw.from_native(pdf)
assert isinstance(ndf.to_pandas()["foo"].dtype, pd.CategoricalDtype)
```

The categorical column can even be used:
```python
ndf.with_columns((nw.col("foo") == "baz").alias("buz")).to_pandas().dtypes
>>> foo    category
buz        bool
dtype: object
```
Is this by accident, or do you think this will continue to work? If so, I think this should be enough to get an internal prototype off the ground already! Whenever we do something more involved with categorical data, we may drop down to the native framework (Pandas, Polars, and an ONNX-based library, in our case). After that exercise we should have a clearer idea what is need to make the code framework agnostic. I'm afraid I don't have a timeline, but it seems that this is not blocking us from trying out narwhals.

MarcoGorelli:
> Is this by accident, or do you think this will continue to work?

this'll continue to work, definitely. support in operations like `.cast` will come

MarcoGorelli:
we've now got cast as well

closing for now then, but please do let me know if there's anything else we should do here, or anything else you'd like!

PR #3346, opened 2025-12-09
refactor: making _compliant selectors public

ym-pett:
# Description

a small refactor to make the selectors in the `_compliant` layer public facing. It would help us over in narwhals-daft. For the moment we actually only need

    `CompliantSelector`,
    `CompliantSelectorNamespace`

but I've added the other two in case we need them in future. If that's wrong let me know, thanks!

I was able to run `pytest` locally without any failures. 

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue in  narwhals-daft: [#79](https://github.com/narwhals-dev/narwhals-daft/issues/79)
- similar to #3344 

## Checklist

- [ ] Tests added
- [ ] Documented the changes


narwhals/compliant.py
```diff
@@ -18,6 +18,12 @@
     StructNamespace,
 )
 from narwhals._compliant.expr import CompliantExprNameNamespace
+from narwhals._compliant.selectors import (
+    CompliantSelector,
+    CompliantSelectorNamespace,
+    EagerSelectorNamespace,
+    LazySelectorNamespace,
```
MarcoGorelli: we don't need these two, do we?)

ym-pett: correct, I erred on the side of including them for future use but have now taken them out

PR #3347, opened 2025-12-09
feat(expr-ir): Add `{DataFrame,Series}.explode(empty_as_null, keep_nulls)`

dangotbanned:
# Description

Discovered a way to do `ArrowDataFrame.explode`, while staying vectorized.

Additionally, (https://github.com/pola-rs/polars/releases/tag/py-1.36.0) just added some options which allow some optimizations - so I've implemented those as well 😏
This is particularly helpful for `pyarrow`, as `explode(empty_as_null=False, keep_nulls=False)` is what happens natively if we don't get in the way.

> [!NOTE]
> I've split this out from (#3325) to make it easier to upstream if wanted
> 50% of the LOC is in (https://github.com/narwhals-dev/narwhals/blob/e86c66dd5976fdbf24b68eff58bd4878b909d227/tests/plan/explode_test.py)

## Related issues

- Child of #3325
- https://github.com/narwhals-dev/narwhals/pull/1644#issuecomment-2751243106
- https://github.com/narwhals-dev/narwhals/pull/1644#issuecomment-3622051763
- https://github.com/pola-rs/polars/issues/17664
- https://github.com/pola-rs/polars/pull/25289
- https://github.com/pola-rs/polars/pull/25369

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

PR #3353, opened 2025-12-14
feat(expr-ir): Add `list.*` aggregate methods

dangotbanned:
# Description

By the time I merged (#3325), I was behind again - but luckily this was able to pickyback off other features 😅

## From (#3332)
- `list.max`
- `list.mean`
- `list.median`
- `list.min`
- `list.sum`

## New!
Before anyone asks, take a look at how small these commits were 😉 

(https://github.com/narwhals-dev/narwhals/pull/3353/commits/3fefcdb216ea9c44195027566282a3c21fde1b0a, https://github.com/narwhals-dev/narwhals/pull/3353/commits/96b66385c9e699b5ae2742f66da3e72283728cf0, https://github.com/narwhals-dev/narwhals/pull/3353/commits/5b310c622dd312f4dd555225240dd8b73580354c)
- `list.any`
- `list.all`
- `list.first`
- `list.last`
- `list.n_unique`

I can see a similar path for `list.std` and `list.var` - but I think they're less important than other *non-aggregating* `list.*` methods that I'd like to do first

## Related issues

- Child of #2572
- Trying to get ahead of #3332

### DIFF
--- a/narwhals/_plan/arrow/options.py
+++ b/narwhals/_plan/arrow/options.py
@@ -1,20 +1,25 @@
-"""Cached `pyarrow.compute` options classes, using `polars` defaults.
+"""Cached [`pyarrow.compute` options], using `polars` defaults and naming conventions.
 
-Important:
-    `AGG` and `FUNCTION` mappings are constructed on first `__getattr__` access.
+See `LazyOptions` for [`__getattr__`] usage.
+
+[`pyarrow.compute` options]: https://arrow.apache.org/docs/dev/python/api/compute.html#compute-options
+[`__getattr__`]: https://docs.python.org/3/reference/datamodel.html#module.__getattr__
 """
 
 from __future__ import annotations
 
 import functools
-from typing import TYPE_CHECKING, Any, Literal
+from collections.abc import Mapping
+from typing import TYPE_CHECKING, Any, Literal, TypeVar
 
 import pyarrow.compute as pc
 
 from narwhals._utils import zip_strict
 
 if TYPE_CHECKING:
-    from collections.abc import Mapping, Sequence
+    from collections.abc import Sequence
+
+    from typing_extensions import TypeAlias
 
     from narwhals._plan import expressions as ir
     from narwhals._plan.arrow import acero
@@ -26,20 +31,57 @@ if TYPE_CHECKING:
 __all__ = [
     "AGG",
     "FUNCTION",
+    "LIST_AGG",
     "array_sort",
     "count",
     "join",
     "join_replace_nulls",
+    "match_substring",
     "rank",
     "scalar_aggregate",
     "sort",
+    "split_pattern",
     "variance",
 ]
 
 
-AGG: Mapping[type[agg.AggExpr], acero.AggregateOptions]
-FUNCTION: Mapping[type[ir.Function], acero.AggregateOptions]
-LIST_AGG: Mapping[type[ir.lists.Aggregation], acero.AggregateOptions]
+_T = TypeVar("_T", bound="type[ir.ExprIR | ir.Function]")
+
+LazyOptions: TypeAlias = Mapping[_T, "acero.AggregateOptions"]
+"""Lazily constructed mapping to `pyarrow.compute.FunctionOptions` instances.
+
+Examples:
+    >>> from narwhals import _plan as nwp
+    >>> from narwhals._plan import expressions as ir
+    >>> from narwhals._plan.arrow import options
+    >>>
+    >>> expr = nwp.col("a").first()
+    >>> expr_ir = expr._ir
+    >>> expr_ir
+    col('a').first()
+    >>> if isinstance(expr_ir, ir.AggExpr):
+    >>>     print(options.AGG.get(type(expr_ir)))
+    ScalarAggregateOptions(skip_nulls=false, min_count=0)
+
+    The first access to `AGG` generated the mapping
+
+    >>> lazy = {"AGG", "FUNCTION", "LIST_AGG"}
+    >>> [key for key in options.__dict__ if key in lazy]
+    ['AGG']
+
+    We *didn't* generate `FUNCTION`, but it'll be there *when* we need it
+
+    >>> options.FUNCTION.get(ir.functions.NullCount)
+    CountOptions(mode=NULLS)
+
+    >>> [key for key in options.__dict__ if key in lazy]
+    ['AGG', 'FUNCTION']
+"""
+
+AGG: LazyOptions[type[agg.AggExpr]]
+FUNCTION: LazyOptions[type[ir.Function]]
+LIST_AGG: LazyOptions[type[ir.lists.Aggregation]]
+
 
 _NULLS_LAST = True
 _NULLS_FIRST = False
@@ -185,7 +227,6 @@ def _generate_function() -> Mapping[type[ir.Function], acero.AggregateOptions]:
 
 
 # ruff: noqa: PLW0603
-# NOTE: Using globals for lazy-loading cache
 if not TYPE_CHECKING:
 
     def __getattr__(name: str) -> Any:


narwhals/_plan/arrow/group_by.py
```diff
@@ -51,9 +52,19 @@
     agg.Last: "hash_last",
     fn.MinMax: "hash_min_max",
 }
+SUPPORTED_LIST_AGG: Mapping[type[ir.lists.Aggregation], type[agg.AggExpr]] = {
```
dangotbanned, line 55: Eventually, it'll make more sense to define this *somewhere* in `_plan/expressions/` - but this'll do for now

narwhals/_plan/arrow/group_by.py
```diff
@@ -65,16 +76,42 @@
...
+    "mean": fn.F64,
+    "skew": fn.F64,
+    "stddev": fn.F64,
+    "variance": fn.F64,
+}
+"""Scalar aggregates that have an output type **not** dependent on input types*.
+
+For use in list aggregates, where the input was null.
+
+*Except `"mean"` can be `Decimal`.
```
dangotbanned:
```suggestion
*Except `"mean"` will preserve `Decimal`, if that's where we started.
```

tests/plan/list_agg_test.py
```diff
@@ -0,0 +1,151 @@
...
+
+def cases_scalar(
+    expr: nwp.Expr,
+    rows: Iterable[Sequence[NonNestedLiteral] | None],
+    expected: Sequence[NonNestedLiteral],
+) -> Iterator[ParameterSet]:
+    for idx, row_expected in enumerate(zip_strict(rows, expected), start=1):
+        row, out = row_expected
+        name = get_dispatch_name(expr._ir).removeprefix("list.")
+        yield pytest.param(expr, row, out, id=f"{name}-R{idx}")
```
dangotbanned, line 120: Well that didn't need to be there 🤔

```suggestion
    name = get_dispatch_name(expr._ir).removeprefix("list.")
    for idx, row_expected in enumerate(zip_strict(rows, expected), start=1):
        row, out = row_expected
        yield pytest.param(expr, row, out, id=f"{name}-R{idx}")
```

narwhals/_plan/arrow/functions.py
```diff
@@ -442,6 +442,17 @@ def explode(
...
+        """Explode list elements, expanding one-level into a table indexing the origin.
+
+        Returns a 2-column table, with names `"idx"` and `"values"`:
+
+            >>> from narwhals._plan.arrow import functions as fn
+            >>>
+            >>> arr = fn.array([[1, 2, 3], None, [4, 5, 6], []])
+            >>> fn.ExplodeBuilder().explode_with_indices(arr).to_pydict()
+            {'idx': [0, 0, 0, 1, 2, 2, 2, 3], 'values': [1, 2, 3, None, 4, 5, 6, None]}
+            # ^ Which sublist we came from     ^ The exploded values themselves
```
FBruzzesi: Nitpicking:

```suggestion
            {'idx': [0, 0, 0, 1, 2, 2, 2, 3], 'values': [1, 2, 3, None, 4, 5, 6, None]}
            # ^ Which sublist values come from   ^ The exploded values themselves
```

narwhals/_plan/arrow/group_by.py
```diff
@@ -65,16 +76,42 @@
...
     **_version_dependent,
 }
 
 del _version_dependent
 
 
+SUPPORTED_LIST_FUNCTION: Mapping[type[ir.lists.Aggregation], type[ir.Function]] = {
+    ir.lists.Any: ir.boolean.Any,
+    ir.lists.All: ir.boolean.All,
+}
```
FBruzzesi, line 90: I would expect these to be aggregations as well (and go in `SUPPORTED_LIST_AGG`), but apparently aren't? That would simplify `_from_list_agg`

narwhals/_plan/arrow/options.py
FBruzzesi, line 1: First look: my brain hurts

Second look: should all these `_generate_*` functions be cached? What's the reason to rewrite global variable each time we try to access them? Is this a common pattern?

narwhals/_plan/arrow/group_by.py
```diff
@@ -155,6 +213,29 @@ def implode(cls, name: str) -> Self:
...
+                return fn.lit(None, SCALAR_OUTPUT_TYPE.get(func, scalar.type.value_type))
+            result = pc.call_function(func, [scalar.values], self._option)
+            return result
+        result = self.over_index(fn.ExplodeBuilder().explode_with_indices(native), "idx")
+        result = fn.when_then(native.is_valid(), result)
+        if self._is_n_unique():
+            # NOTE: Exploding `[]` becomes `[None]` - so we need to adjust the unique count *iff* we were unlucky
+            len_not_eq_0 = fn.not_eq(fn.list_len(native), fn.lit(0))
+            if not fn.all_(len_not_eq_0, ignore_nulls=False).as_py():
+                result = fn.when_then(fn.not_(len_not_eq_0), fn.lit(0), result)
```
FBruzzesi: The mental load on the double negations is felt! We could flip the conditions (suggestion assuming this functions exist)

```suggestion
            len_eq_0 = fn.eq(fn.list_len(native), fn.lit(0))
            if fn.any_(len_eq_0, ignore_nulls=False).as_py():
                result = fn.when_then(len_eq_0, fn.lit(0), result)
```

narwhals/_plan/arrow/group_by.py
dangotbanned: I too was confused by this initially 😄

But this is something I've inherited from `polars`

### `polars`

- https://github.com/pola-rs/polars/blob/99bfe6a92f5cac6bcc6973ef6e8da16b2aa2333d/crates/polars-plan/src/plans/aexpr/function_expr/boolean.rs
- https://github.com/pola-rs/polars/blob/99bfe6a92f5cac6bcc6973ef6e8da16b2aa2333d/crates/polars-plan/src/plans/aexpr/mod.rs#L39-L74

### Here
- https://github.com/narwhals-dev/narwhals/blob/1550febd99a8057ebb328333ddc01361a02a8a8b/narwhals/_plan/expressions/boolean.py
- https://github.com/narwhals-dev/narwhals/blob/1550febd99a8057ebb328333ddc01361a02a8a8b/narwhals/_plan/expressions/aggregation.py

## But why?
There might be stronger reasoning that I haven't found yet, but from what I understand:
- **All** `AggExpr`s aggregate to a single value
- **Some** `FunctionExpr`s aggregate (but not many), and they are marked with `FunctionOptions.aggregation`

If I had to guess, it may be that these *aggregating functions* place additional constraints on their inputs.
These two cases must also have `Boolean` inputs.
Some others like [`NullCount`](https://github.com/pola-rs/polars/blob/99bfe6a92f5cac6bcc6973ef6e8da16b2aa2333d/crates/polars-plan/src/plans/aexpr/function_expr/mod.rs#L1021) do not observe order (I haven't added this concept, it was new 😅)

That being said, I wouldn't be opposed to deviating from upstream here if it can simplify things 🙂

narwhals/_plan/arrow/options.py
dangotbanned: > First look: my brain hurts

😂 

> Second look: should all these `_generate_*` functions be cached?

I think this may be a documentation issue on my part.

<details><summary>I tried to explain this in the module doc</summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/5b310c622dd312f4dd555225240dd8b73580354c/narwhals/_plan/arrow/options.py#L1-L5

</p>
</details>


<details><summaryAnd then again on <code>arrow.options.__getattr__</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/5b310c622dd312f4dd555225240dd8b73580354c/narwhals/_plan/arrow/options.py#L187-L205

</p>
</details>

So what we have here is a two-tier cache.

## Part 1

Constructors for [`pyarrow.compute.FunctionOptions`](https://github.com/zen-xu/pyarrow-stubs/blob/f3c148a1169d0210c73bbba40fea9a1728cbf62f/pyarrow-stubs/_compute.pyi) **are** cached

<details><summary>Show <code>options.count</code></summary>
<p>


https://github.com/narwhals-dev/narwhals/blob/5b310c622dd312f4dd555225240dd8b73580354c/narwhals/_plan/arrow/options.py#L56-L60

</p>
</details> 


They *also* may have their defaults overriden and parameter names changed ...

<details><summary>... to align more closely with <code>polars</code></summary>
<p>

image

</p>
</details> 

So the idea is both for performance and to try and lessen the burden of remembering these differences for each backend 🙂

dangotbanned: ## Part 2
> What's the reason to rewrite global variable each time we try to access them?

[`ModuleType.__getattr__`](https://docs.python.org/3/reference/datamodel.html#module.__getattr__) is only called when module attribute access fails.
The first time this happens, for a given `name`, the `global` keyword is used to **define** that variable and populate it.

```py
from narwhals._plan.arrow import options

options.__dict__.get("LIST_AGG", "nothing here boss")
# 'nothing here boss'
```
But where is it?

image

```py
options.LIST_AGG
# {narwhals._plan.expressions.lists.Sum: ScalarAggregateOptions(skip_nulls=true, min_count=0),
#  ...
#  narwhals._plan.expressions.lists.NUnique: CountOptions(mode=ALL)}
```

And here we see `__getattr__` is definitely not called again 🙂 

```py
options.LIST_AGG is options.__dict__.get("LIST_AGG", "nothing here boss")
# True
```

## But why?
There's quite a lot of code being run here, that ideally:
- We don't want running at the time of importing `arrow.options` (mainly indirectly)
- Is only needed if you write certain kinds of expression(s) (with `pyarrow`)

So this way is intended to minimize the costs across 2 levels and try to be more *pay-for-what-you-use* 

> Is this a common pattern?

@FBruzzesi Do you remember the last time this came up? 😉 

- https://github.com/narwhals-dev/narwhals/pull/2514#discussion_r2083104410

dangotbanned: I definitely need to document this stuff better

narwhals/_plan/arrow/group_by.py
dangotbanned: > The mental load on the double negations is felt! 

Yeah you're not wrong 😉 

> suggestion assuming this functions exist

Indeed they do! 

I will definitely take this if it doesn't change the behavior.

Here's what I have but written with `polars`:

```py
import polars as pl

data = {
    "a": [
        [1.0, None, None, 3.0],
        [1.0, None, 4.0, 5.0, 1.1, 4.0, None, 1.0],
        [1.0, None, None, 1.0, 2.0, 2.0, 2.0, None, 3.0],
        [],
        [None, None, None],
        None,
    ]
}
expr = (pl.col("a").list.len() != 0).not_()
df = pl.DataFrame(data).with_columns(is_empty_list=expr)
print(df)
has_empty_list = df.get_column("is_empty_list").any(ignore_nulls=False)
print(f"{has_empty_list=}")
```

```
shape: (6, 2)
┌────────────────────┬───────────────┐
│ a                  ┆ is_empty_list │
│ ---                ┆ ---           │
│ list[f64]          ┆ bool          │
╞════════════════════╪═══════════════╡
│ [1.0, null, … 3.0] ┆ false         │
│ [1.0, null, … 1.0] ┆ false         │
│ [1.0, null, … 3.0] ┆ false         │
│ []                 ┆ true          │
│ [null, null, null] ┆ false         │
│ null               ┆ null          │
└────────────────────┴───────────────┘
has_empty_list=True
```

narwhals/_plan/arrow/functions.py
dangotbanned: Yep, much better!

narwhals/_plan/arrow/options.py
dangotbanned: @FBruzzesi Hopefully this should be more useful than relying on what is in my head 😄 

[docs: More clearly demo `arrow.options` lazy mappings](https://github.com/narwhals-dev/narwhals/pull/3353/commits/f74c4ddbb1327a5e1c75ced822fc2afd403ab291)

image

image

---

I really wish Pylance fixed the doctest wrapping, but its better than nothing 

image

dangotbanned:
Thanks for the review @FBruzzesi, your comments were very helpful!

> I am not sure if you are looking for something specifically in the review - I must admit I stopped following the progress in #2572 at a certain point.

> [!NOTE]
> **TL;DR:** Small diff gives interesting peek into big diff

Any feedback is better than none 😄 

I know #2572 is a bit daunting to dive into, but I thought this PR highlighted one of its strengths.

Everything here is either **directly using** or **building on** an existing (#2572) feature.
I'm really just adding a *bit* of plumbing between (#3347) and `GroupBy`
With some tweaking, this could also support (`list.`): `var`, `std`, `count`, `null_count`, `kurtosis`, `skew`.

I can see a path towards some other [informal `Expr` rewriting](https://github.com/narwhals-dev/narwhals/blob/f74c4ddbb1327a5e1c75ced822fc2afd403ab291/narwhals/_plan/arrow/expr.py#L573-L600) (which extends support for `over(*partition_by, **kwds)`) working with `list`s too. Especially once that's redefined as [transformations on a `LogicalPlan`](https://github.com/narwhals-dev/narwhals/pull/3332#discussion_r2602301476).

narwhals/_plan/arrow/group_by.py
dangotbanned: After some git archaeology on the parts of (https://github.com/pola-rs/polars/tree/05f6f2db49c6721f8208b3bbcca5ec54568e34c4/crates) that I'm *vaguely* familiar with - I still wasn't able to find anything explaining what defines something being an `AggExpr` vs a `FunctionExpr` which aggregates.

An interesting find though was that `IRAggExpr` have a corresponding [`GroupByMethod`](https://github.com/pola-rs/polars/blob/05f6f2db49c6721f8208b3bbcca5ec54568e34c4/crates/polars-plan/src/plans/aexpr/mod.rs#L129-L173) - but this is a [deprecated feature](https://github.com/pola-rs/polars/blob/05f6f2db49c6721f8208b3bbcca5ec54568e34c4/crates/polars-core/src/frame/group_by/mod.rs#L357) that is [still geting updated](https://github.com/pola-rs/polars/pull/24888/files#diff-b0b2c1ae26233e2f3350169347d500514f5dc6cdf9460e76e4c8508288209075)? 🤔 

---

I think until I understand *why* there's distinction between the two - it might be best to assume someone smarter than me made the right decision 😅

I'm definitely curious though and wanna revisit it later

PR #3356, opened 2025-12-15
feat: add `list.sort`

raisadz:
# Description

I tried to have some workarounds for the edge cases for pyarrow that we discussed [#3332 (comment)](https://github.com/narwhals-dev/narwhals/pull/3332#pullrequestreview-3519374373) (None, empty lists and lists with only None elements) with [`pc.sort_indices`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.sort_indices.html) and [`pc.replace_with_mask`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.replace_with_mask.html) but the latter doesn't seem to work for the list types. There is an open issue https://github.com/apache/arrow/issues/48060 that can make it work for pyarrow, pandas and modin when solved.

Also, I opened a couple of issues in sqlframe and ibis related to this PR:
	
https://github.com/eakmanrq/sqlframe/issues/559
https://github.com/eakmanrq/sqlframe/issues/560
https://github.com/ibis-project/ibis/issues/11735

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

PR #3359, opened 2025-12-17
feat(expr-ir): Add `list.sort`

dangotbanned:
# Description

I mentioned in (https://github.com/narwhals-dev/narwhals/pull/3356#issuecomment-3665124739) that I had a solution for this one.

## Related issues

- Child of #2572
- Third-cousin of #3356

### DIFF
--- a/narwhals/_plan/arrow/expr.py
+++ b/narwhals/_plan/arrow/expr.py
@@ -1008,7 +1008,6 @@ class ArrowListNamespace(
         if isinstance(previous, ArrowScalar):
             result = fn.list_sort_scalar(previous.native, node.function.options)
         else:
-            # TODO @dangotbanned: Maybe keep options together?
             descending = node.function.options.descending
             nulls_last = node.function.options.nulls_last
             result = fn.list_sort(


narwhals/_plan/arrow/functions.py
```diff
@@ -792,7 +798,46 @@ def list_contains(
...
+
+# TODO @dangotbanned: Needs more tidying
+def list_sort_messy(
+    native: ChunkedList, *, descending: bool = False, nulls_last: bool = False
+) -> ChunkedList:
+    """Pretty similar to `list_unique`, split things up and put back together."""
+    idx, v = "idx", "values"
+    names = idx, v
+    # null, 0-length, AND 1-length are **all** already sorted
+    len_gt_1 = gt(list_len(native), lit(1))
```
dangotbanned: Rename to `needs_sorting` or `is_not_sorted`

narwhals/_plan/arrow/functions.py
```diff
@@ -792,7 +798,46 @@ def list_contains(
...
+    indexed = concat_horizontal([int_range(len(native)), native], names)
+    valid = indexed.filter(len_gt_1)
+    invalid = indexed.filter(or_(native.is_null(), not_(len_gt_1)))
```
dangotbanned: ### 1 
These should have names reflecting that (currently) `valid` needs to be sorted and `invalid` doesn't.

### 2
I think this:

```py
indexed.filter(or_(native.is_null(), not_(len_gt_1)))
```

May be equivalent to:
```py
indexed.filter(not_(len_gt_1), null_selection_behavior="emit_null")
```

Even though it is longer, it saves a few calls

narwhals/_plan/arrow/expr.py
```diff
@@ -1002,6 +1002,20 @@ def aggregate(
         agg = AggSpec._from_list_agg(node.function, "values")
         return self.with_native(agg.agg_list(previous.native), name)
 
+    def sort(self, node: FExpr[lists.Sort], frame: Frame, name: str) -> Expr | Scalar:
+        previous = node.input[0].dispatch(self.compliant, frame, name)
+        result: ChunkedOrScalarAny
+        if isinstance(previous, ArrowScalar):
+            result = fn.list_sort_scalar(previous.native, node.function.options)
+        else:
+            # TODO @dangotbanned: Maybe keep options together?
```
dangotbanned: Too complicated, since this becomes `descending=[False, descending]`

```suggestion
```

narwhals/_plan/options.py
```diff
@@ -164,6 +164,9 @@ def to_arrow(self, by: Sequence[str]) -> pc.SortOptions:
 
         return sort(*by, descending=self.descending, nulls_last=self.nulls_last)
 
+    def _ensure_single_nulls_last(self, backend: Backend, /) -> bool:
+        return self.nulls_last
+
```
dangotbanned, line 169: This is a bit of a hack to avoid a runtime check between `SortOptions` and `SortMultipleOptions` in `sort_indices`

https://github.com/narwhals-dev/narwhals/blob/451606ad76c19b8b0eedb356e0196d0acc98788b/narwhals/_plan/options.py#L190-L193

PR #3371, opened 2025-12-24
feat(expr-ir): Add `concat`

dangotbanned:
# Description
Episode 5 is aiming for [`DslPlan::Union`](https://github.com/pola-rs/polars/blob/00d7f7e1c3b24a54a13f235e69584614959f8837/crates/polars-plan/src/dsl/plan.rs#L142-L146) and [`DslPlan::HConcat`](https://github.com/pola-rs/polars/blob/00d7f7e1c3b24a54a13f235e69584614959f8837/crates/polars-plan/src/dsl/plan.rs#L147-L151).

https://github.com/narwhals-dev/narwhals/blob/63dbc65bc68d6dea849f927711da46c287f1a029/narwhals/_plan/logical_plan/plan.py#L300-L308

https://github.com/narwhals-dev/narwhals/blob/63dbc65bc68d6dea849f927711da46c287f1a029/narwhals/_plan/logical_plan/plan.py#L311-L319

---

Back over here, the compliant-level was already there - just need to get the tests in place now.

## Related issues

- Child of #2572
- Follows #3370

### DIFF
--- a/narwhals/_plan/_guards.py
+++ b/narwhals/_plan/_guards.py
@@ -138,9 +138,7 @@ def is_literal(obj: Any) -> TypeIs[ir.Literal[Any]]:
     return isinstance(obj, _ir().Literal)
 
 
-# TODO @dangotbanned: Coverage
-# Used in `ArrowNamespace._vertical`, but only horizontal is covered
-def is_tuple_of(obj: Any, tp: type[T]) -> TypeIs[Seq[T]]:  # pragma: no cover
+def is_tuple_of(obj: Any, tp: type[T]) -> TypeIs[Seq[T]]:
     """Return True if the **first** element of the tuple `obj` is an instance of `tp`."""
     return bool(isinstance(obj, tuple) and obj and isinstance(obj[0], tp))
 


narwhals/_plan/dataframe.py
```diff
@@ -224,7 +222,17 @@ def _dataframe_from_dict(
...
+
+    def __narwhals_namespace__(
+        self,
+    ) -> EagerNamespace[
+        EagerDataFrame[Any, NativeDataFrameT_co, NativeSeriesT],
+        CompliantSeries[NativeSeriesT],
+        Any,
+        Any,
+    ]:
+        return self._compliant.__narwhals_namespace__()
```
dangotbanned, line 235: ## Follow-up
Really need to move more things from `EagerDataFrame` up into `CompliantDataFrame`.

The big one is this part, which is how we know the namespace has a `CompliantDataFrame`.


https://github.com/narwhals-dev/narwhals/blob/ff2cd902df838f4a6a068e7ee3a9a94cbc719534/narwhals/_plan/compliant/dataframe.py#L274-L278

The version I'm using doesn't define `concat` in `CompliantNamespace`.
I wanted to model that the eager version works for `DataFrame` and `Series` - but you can't invite both to the same party 😉:


https://github.com/narwhals-dev/narwhals/blob/ff2cd902df838f4a6a068e7ee3a9a94cbc719534/narwhals/_plan/compliant/namespace.py#L115-L121

Whereas, the lazy version only allows `LazyFrame` - while not including the the equivalent of `Series`.
Also I seem to have forgetten the part where `LazyFrame` doesn't support `"horizontal"` 🤦‍♂️ 

https://github.com/narwhals-dev/narwhals/blob/ff2cd902df838f4a6a068e7ee3a9a94cbc719534/narwhals/_plan/compliant/namespace.py#L181-L186

dangotbanned: Marking as *Resolved*, since it was added to the big TODO

Issue #2526, opened 2025-05-10
[Enh]: add `Expr.first` / `Expr.min_by` / `Expr.max_by`


MathiasHolmstrom:
I often find myself desiring a .first() expression. Is there a plan to support that expression?

Polars Example
```py
df = pl.DataFrame(
    {
        "group": ["A", "A", "A", "B", "B", "B"],
        "value": [1, 2, 3, 4, 5, 6],
    }
)
df = df.with_columns(
    pl.col("value").first().over("group").alias("mean_value")
)
```

## Related
- [x] #1793
- [x] #2534

MarcoGorelli:
Definitely in scope thank you for the suggestion

Anyone fancy contributing this?

dangotbanned:
> Definitely in scope thank you for the suggestion
> 
> Anyone fancy contributing this?

I'd be interested!

Couple questions:

1. Would we also want `Expr.last()` support?
2. Are there any rules you'd wanna introduce/reuse on how where it can appear?

Interestingly, we could use it in `group_by().agg(...) for `pyarrow`
https://github.com/zen-xu/pyarrow-stubs/blob/8a2c16444f2293f6df224148e26d8250a74ebd5d/pyarrow-stubs/__lib_pxi/table.pyi#L74-L76

MarcoGorelli:
For lazy backends this would require `order_by=...` in the `over` statement

rwhitten577:
I was going to ask for this too - both `Expr.first()` and `Expr.last()` are useful

dangotbanned:
# Update
#2528 stalled a while back, due to concerns with:
- [`pandas`](https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2085080744) when using `df.group_by(...).agg(nw.col("a").first())`
- [Lazy support](https://github.com/narwhals-dev/narwhals/pull/2528#discussion_r2083557149) requiring `nw.col("a").first().over(order_by="b")`

> [!IMPORTANT]
> I'm hoping we can unblock this and get *some* level of support into `narwhals` 🙏 

## `pandas`
Handling the version dispatch will be more realistic after (#2680), since there *appears* to be 2-3 different behaviors we need to normalize across our supported range.
I gave a [summary](https://github.com/narwhals-dev/narwhals/pull/2680#discussion_r2160134917) on how this'll be easier than **further** increasing the complexity (#2489) to include version branching for `(first|last)`

[...]

Issue #3363, opened 2025-12-19
enh: add (optional) `order_by` to `rolling_*`, `cum_*`, `first`, ...


MarcoGorelli:
Currently, we don't allow
```py
nw.col('price').cum_sum()
```
in the lazy case, requiring instead
```py
nw.col('price').cum_sum().over(order_by='date')
```

We could also allow
```py
nw.col('price').cum_sum(order_by='date')
```
with the following restriction:
- it's not allowed to do `nw.col('price').cum_sum(order_by='date').over('asset', order_by='date')`. specify `order_by` in either `over` or `cum_sum` but not both

What this enables:
- we can add `order_by` to `first`, so then we you can do `df.group_by('a').agg(nw.col('b').first(order_by='c'))`
- i think it's conceptually easier to teach `nw.col('price').cum_sum(order_by='date')` than it is to teach `nw.col('price').cum_sum().over(order_by='date')`

I'm just trying this out, but i think the impact on complexity is relatively small

Issues:
- slight divergence from Polars API, but I'm ok with it, i think it still feels in the polars spirit

dangotbanned:
Isn't this just adding `sort_by` with a different spelling? 😅

### More context
What I have in this test is *where* the `sort_by` would be for equivalence

https://github.com/narwhals-dev/narwhals/blob/9b9122b4ab38a6aebe2f09c29ad0f6191952a7a7/tests/plan/over_test.py#L459-L501

- The signatures of all the `Expr` methods can stay the same
- The implementations can just peek behind for the order
- It would naturally support `descending` and `nulls_last`
  - Which I imagine would be the next ask after adding `order_by: str | Sequence[str]`

MarcoGorelli:
it would be kind of like adding `sort_by` but imposing that it be followed by `cum_sum` / `first` / ...

Issue #2534, opened 2025-05-11
[Enh]: Add `Expr.sort_by`


dangotbanned:
## Related
- https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-13106199
- https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.sort_by.html

## Description
_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/discussions/1657#discussioncomment-13106199_

<details><summary>Proposal</summary>
<p>

I'm surprised this didn't get mentioned before, but wouldn't [`polars.Expr.sort_by`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.sort_by.html) be an alternative to use in some of these cases?

I was looking at [Order-dependence](https://narwhals-dev.github.io/narwhals/basics/order_dependence/) which has this example:

```py
import pandas as pd
from sqlframe.duckdb import DuckDBSession

import narwhals as nw

session = DuckDBSession()
data = {"a": [1, 3, 4], "i": [0, 1, 2]}

df_pd = pd.DataFrame({"a": [1, 3, 4], "i": [0, 1, 2]})
sqlframe_df = session.createDataFrame(df_pd)
ldf = nw.from_native(sqlframe_df)

>>> ldf.with_columns(a_cum_sum=nw.col("a").cum_sum().over(order_by="i")).collect("pandas")
┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|   a  i  a_cum_sum|
|0  1  0          1|
|1  3  1          4|
|2  4  2          8|
└──────────────────┘
```

Translating that `Expr` directly to `polars` isn't valid at runtime:

```py
import polars as pl

ldf = pl.LazyFrame(data)

>>> ldf.with_columns(a_cum_sum=pl.col("a").cum_sum().over(order_by="i")).collect()
TypeError: Expr.over() missing 1 required positional argument: 'partition_by'
```

But we can use the shorter `sort_by`:

```py
>>> ldf.with_columns(a_cum_sum=pl.col("a").cum_sum().sort_by("i")).collect()
shape: (3, 3)
┌─────┬─────┬───────────┐
│ a   ┆ i   ┆ a_cum_sum │
│ --- ┆ --- ┆ ---       │
│ i64 ┆ i64 ┆ i64       │
╞═════╪═════╪═══════════╡
│ 1   ┆ 0   ┆ 1         │
[...]

PR #3372, opened 2025-12-26
enh: Introduce (optional) `order_by` in `first` / `last`

MarcoGorelli:
Unfortunately I don't think there's any more efficient way to do it in pyarrow/pandas, as their top-k-like methods put nulls last

closes https://github.com/narwhals-dev/narwhals/issues/2526

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -90,6 +90,8 @@ jobs:
           . .venv/bin/activate
           # Temporary pin until it's addressed upstream.
           uv pip install "pandas<3"
+          # Temporary pin to get CI green
+          uv pip install "sqlglot<28.7.0"
           uv pip install -e ".[dev]"
           which python
       - name: install-narwhals-dev
@@ -241,6 +243,8 @@ jobs:
         run: |
           cd tea-tasting
           uv sync --group test
+          # Temporary pin to get CI green
+          uv pip install "sqlglot<28.7.0"
       - name: install-narwhals-dev
         run: |
           cd tea-tasting
@@ -462,6 +466,8 @@ jobs:
           uv venv -p ${{ matrix.python-version }}
           . .venv/bin/activate
           uv pip install . --group dev
+          # Temporary pin to get CI green
+          uv pip install "sqlglot<28.7.0"
           uv pip install pytest pytest-cov pytest-snapshot pandas polars "ibis-framework[duckdb,mysql,postgres,sqlite]>=9.5.0" chatlas shiny
       - name: install-narwhals-dev
         run: |


MarcoGorelli:
Merging as there's been no objections

FBruzzesi:
@MarcoGorelli speaking for myself here (but maybe Dan can relate), we recently kept ourselves busy with a bunch of other PRs, aaaand at least I completely forgot to review this. If you are ok with waiting for one more release, I can promise to take a deep look

MarcoGorelli:
ok sure, thanks!

narwhals/_polars/expr.py
```diff
@@ -185,6 +185,12 @@ def rolling_std(
         )
         return self._with_native(native)
 
+    def min_by(self, by: Sequence[str]) -> Self:
+        return self._with_native(self.native.sort_by(*by).first())
+
+    def max_by(self, by: Sequence[str]) -> Self:
+        return self._with_native(self.native.sort_by(*by).last())
```
FBruzzesi: Shouldn't be using the native `{min, max}_by` from the most recent version onwards?

narwhals/expr.py
```diff
@@ -1688,6 +1688,58 @@ def last(self) -> Self:
         """
         return self._append_node(ExprNode(ExprKind.ORDERABLE_AGGREGATION, "last"))
 
+    def min_by(self, by: str | Iterable[str]) -> Self:
+        """Get minimum, ordered by another expression.
+
```
FBruzzesi: Can we add the same disclaimers that polars has on no guarantees for breaking equality and flagged as unstable:

```py
        If the by expression has multiple values equal to the maximum it is not
        defined which value will be chosen.

        .. warning::
            This functionality is considered **unstable**. It may be changed
            at any point without it being considered a breaking change.
```

[source](https://github.com/pola-rs/polars/blob/9f9dee823241b9d304969fe92dcc73e04136925d/py-polars/src/polars/expr/expr.py#L3141-L3146)

narwhals/_arrow/group_by.py
```diff
@@ -91,50 +92,49 @@ def _configure_agg(
...
+            md = next(expr._metadata.op_nodes_reversed())
+            if md.name not in self._OPTION_ORDERED:
+                # [pyarrow-36709]: https://github.com/apache/arrow/issues/36709
+                continue
+            use_threads = False
+            if _by := md.kwargs.get("by", ()):
+                if by and _by != by:
+                    msg = f"Only one `by` can be specified in `group_by`. Found both {by} and {_by}."
+                    raise NotImplementedError(msg)
+                by = _by
```
FBruzzesi: I have a couple of suggestions, mostly for readability, on `_by` vs `by`:

- `by` -> `current_by`
- `_by` -> `candidate_by` / `new_by`

Remove one nesting level with:

```py
if not (_by := md.kwargs.get("by", ())):
    continue
if by and _by != by:
    msg = f"Only one `by` can be specified in `group_by`. Found both {by} and {_by}."
by = _by
```

narwhals/_arrow/group_by.py
```diff
@@ -91,50 +92,49 @@ def _configure_agg(
...
+        if not use_threads and self.compliant._backend_version < (14, 0):
             msg = (
-                f"Using `{name}()` in a `group_by().agg(...)` context is only available in 'pyarrow>=14.0.0', "
-                f"found version {requires._unparse_version(backend_version)!r}.\n\n"
+                f"Using `first/last/min_by/max_by` in a `group_by().agg(...)` context is only available in 'pyarrow>=14.0.0', "
+                f"found version {requires._unparse_version(self._df._backend_version)!r}.\n\n"
```
FBruzzesi: It's not much but walrus:

```py
if not use_threads and (pa_version := self.compliant._backend_version) < (14, 0):
    msg = (
        f"Using `first/last/min_by/max_by` in a `group_by().agg(...)` context is only available in 'pyarrow>=14.0.0', "
        f"found version {requires._unparse_version(pa_version)!r}.\n\n"
        f"See https://github.com/apache/arrow/issues/36709"
    )
```

narwhals/_arrow/group_by.py
```diff
@@ -91,50 +92,49 @@ def _configure_agg(
...
+        if by:
+            return pa.TableGroupBy(
+                self.compliant.sort(*by, descending=False, nulls_last=False).native,
+                self._keys,
+                use_threads=use_threads,
+            )
+        if not use_threads:
+            return pa.TableGroupBy(self.compliant.native, self._keys, use_threads=False)
+        # TODO(unassigned): combine with `return` above once PyArrow 15 is the minimum.
+        return pa.TableGroupBy(self.compliant.native, self._keys)
```
FBruzzesi, line 130: Am I missing something obvious?

```suggestion
        return pa.TableGroupBy(self.compliant.native, self._keys, use_threads=use_threads)
```

narwhals/_pandas_like/group_by.py
```diff
@@ -168,10 +170,10 @@ def is_len(self) -> bool:
         return self.leaf_name == "len"
 
     def is_last(self) -> bool:
-        return self.leaf_name == "last"
+        return self.leaf_name in {"last", "max_by"}
 
     def is_first(self) -> bool:
-        return self.leaf_name == "first"
+        return self.leaf_name in {"first", "min_by"}
```
FBruzzesi: Nitpick: proposing to rename to `is_{last, first}_like`

narwhals/_pandas_like/group_by.py
```diff
@@ -258,38 +262,51 @@ def __init__(
...
             if not self._is_simple(expr):
                 all_aggs_are_simple = False
+            md = next(expr._metadata.op_nodes_reversed())
+            if md.name not in _REMAP_ORDERED_INDEX:
+                continue
+            if _by := md.kwargs.get("by", ()):
+                if by and _by != by:
+                    msg = f"Only one `by` can be specified in `group_by`. Found both {by} and {_by}."
+                    raise NotImplementedError(msg)
+                by = _by
```
FBruzzesi: Same feedback as in the pyarrow case

narwhals/_pandas_like/group_by.py
```diff
@@ -109,15 +111,15 @@ def with_expand_names(self, group_by: PandasLikeGroupBy, /) -> AggExpr:
         return self
 
     def _getitem_aggs(
-        self, group_by: PandasLikeGroupBy, /
+        self, group_by: PandasLikeGroupBy, grouped: NativeGroupBy, /
     ) -> pd.DataFrame | pd.Series[Any]:
```
FBruzzesi: I personally find passing two quite similar arguments a bit confusing.

I would be in favour of maintaining the assignment to the `_grouped` attribute in:

```diff
if by:
    grouped: NativeGroupBy = self._native.sort_values(
        list(by), na_position="first"
    ).groupby(self._keys.copy(), **self._group_by_kwargs)
else:
    grouped = self._native.groupby(self._keys.copy(), **self._group_by_kwargs)
+ self._grouped = grouped
```

and then assign it inside this function once:

```py
def _getitem_aggs(
        self, group_by: PandasLikeGroupBy /
    ) -> pd.DataFrame | pd.Series[Any]:
    grouped = group_by._grouped
```

tests/expr_and_series/first_last_test.py
```diff
@@ -176,3 +178,188 @@ def test_first_expr_over_order_by_partition_by(
         "c_last": [None, None, 8.0],
     }
     assert_equal_data(result, expected)
+
+
+def test_first_expr_in_group_by(
```
dangotbanned, line 194: This applies to all of new tests:

- The tests for `{min,max}_by` should contain the method name(s) in the test ids
- The names of the columns should be suffix-ing `{min,max}_by`, if anything, too
- If you want to share a module w/ `first`, `last` (which isn't necessarily a bad thing)
  - The module should be renamed so that it is clear what's in there

*Currently*, if any of these tests fail, you really need to dig to find out that `first`, `last` are not the issue

narwhals/expr.py
```diff
@@ -1688,6 +1688,58 @@ def last(self) -> Self:
         """
         return self._append_node(ExprNode(ExprKind.ORDERABLE_AGGREGATION, "last"))
 
+    def min_by(self, by: str | Iterable[str]) -> Self:
+        """Get minimum, ordered by another expression.
```
dangotbanned: I think we shouldn't allow `Iterable[str]` here, since `polars` seems to broadcast multi-output:

```py
import polars as pl

data = {"a": [None, 4, 9, 3], "idx": [9, None, None, 7], "idx2": [9, 8, 7, 7]}
pl.DataFrame(data).select(pl.col("a").min_by(pl.col("idx", "idx2")))
```
```
DuplicateError: projections contained duplicate output name 'a'. It's possible that multiple expressions are returning the same default column name. 
If this is the case, try renaming the columns with `.alias("new_name")` to avoid duplicate column names.
```

And the [docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.min_by.html#polars.Expr.min_by) are using the singular *"Column"*

> Column used to determine the smallest element. Accepts expression input. Strings are parsed as column names.

dangotbanned: @FBruzzesi's suggestion would reveal that https://github.com/narwhals-dev/narwhals/pull/3372#discussion_r2716540820

narwhals/_sql/expr.py
```diff
@@ -765,6 +765,34 @@ def func(
...
+
+        def window_f(
+            df: SQLLazyFrameT, inputs: WindowInputs[NativeExprT]
+        ) -> Sequence[NativeExprT]:
+            return [
+                self._window_expression(self._first(expr, *by), inputs.partition_by)
+                for expr in self(df)
+            ]
+
+        return self._with_callable(f, window_f)
```
dangotbanned: Was there a reason for *not* using the native `{min,max}_by`s?

- [DuckDB](https://duckdb.org/docs/stable/sql/functions/aggregates#arg_minarg-val)
- [PySpark](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.min_by.html)
- [Ibis](https://ibis-project.org/reference/expression-generic#ibis.expr.types.generic.Column.argmin), [also](https://github.com/ibis-project/ibis/blob/7cc3686cb60db81ee63bafda48da2a9bde2fc855/ibis/backends/sql/compilers/base.py#L327-L328)

tests/expr_and_series/first_last_test.py
```diff
@@ -169,3 +171,188 @@ def test_first_expr_over_order_by_partition_by(
...
+        "b": [9, 7, 10, 8],
+        "c": [9, None, 10, 8],
+        "idx": [9, None, None, 7],
+        "idx2": [9, 8, 7, 7],
+    }
+    df = nw.from_native(constructor(data))
+    with pytest.raises(InvalidOperationError):
+        df.select("idx", "idx2", nw.col("a").min_by("idx").over(order_by="idx2"))
+    with pytest.raises(InvalidOperationError):
+        df.select("idx", "idx2", nw.col("a").max_by("idx").over(order_by="idx2"))
```
dangotbanned: I think it would be useful to check the error contains the change from:
- https://github.com/MarcoGorelli/narwhals/blob/341d11bed07dce58c020671f81db262ad5f5365b/narwhals/dataframe.py#L2384

narwhals/_arrow/group_by.py
```diff
@@ -91,50 +90,48 @@ def _configure_agg(
...
-        return grouped, pc.ScalarAggregateOptions(skip_nulls=ignore_nulls)
+        if order_by:
+            return pa.TableGroupBy(
+                self.compliant.sort(*order_by, descending=False, nulls_last=False).native,
+                self._keys,
+                use_threads=use_threads,
+            )
+        if not use_threads:
+            return pa.TableGroupBy(self.compliant.native, self._keys, use_threads=False)
+        # TODO(unassigned): combine with `return` above once PyArrow 15 is the minimum.
```
dangotbanned, line 129: I'm really struggling to follow this.

There's the same issue @FBruzzesi mentioned in (https://github.com/narwhals-dev/narwhals/pull/3372#discussion_r2716620202).

But if you *need* to move this check, then the first thing in this method should be checking if we have an aggregation that needs it, right?

https://github.com/narwhals-dev/narwhals/blob/e4c63f8f7c49a2325174ee78223970607d3c3d08/narwhals/_arrow/group_by.py#L92-L96

narwhals/_arrow/group_by.py
MarcoGorelli: the `use_theads` keyword isn't available before pyarrow 15

FBruzzesi: ah now I understand the comment, apologies

MarcoGorelli:
thanks both!

indeed, this needs revisiting 🙏  will get back to it

MarcoGorelli:
Aside from this, we may want to actually introduce `order_by` in `first`, as `min_by` / `max_by` in Polars ignore nulls (and nans 🤯 ), whereas I was hoping for a way to do this
```py
In [29]: df = pl.DataFrame({"a": [-1.0, None, 1.0], "b": ["x", "y", "z"]})

In [30]: duckdb.sql("""
    ...: from df
    ...: select first(b order by a nulls first)
    ...: """)
Out[30]:
┌───────────────────────────────────┐
│ "first"(b ORDER BY a NULLS FIRST) │
│              varchar              │
├───────────────────────────────────┤
│ y                                 │
└───────────────────────────────────┘

In [31]: df.select(pl.col('b').min_by('a'))
Out[31]:
shape: (1, 1)
┌─────┐
│ b   │
│ --- │
│ str │
╞═════╡
│ x   │
└─────┘
```
Both seem useful tbh. I think I'll repurpose this to do add `order_by` to `first` / `last`, as it originally was, and separately we can also add `min_by` / `max_by`

narwhals/_arrow/group_by.py
MarcoGorelli: sure, i've moved the check earlier:

https://github.com/narwhals-dev/narwhals/blob/bec1ca6a49b88212e5f176891dec98e4f1b28bba/narwhals/_arrow/group_by.py#L101-L102

MarcoGorelli:
Any objections to adding this, despite it being a superset of the Polars API?

dangotbanned:
> Any objections to adding this, despite it being a superset of the Polars API?

I'm against it *because of* that.
But I'm still in favor of the functionality you want to expose 🙂 

I really think the takeaway from the [discord comment you shared earlier](https://discord.com/channels/1235257048170762310/1276640597284884490/1466103161553944612) - should be that the Polars API is a selling-point for Narwhals.
The README says it best 😉 

> - ✅ **Just use** [a subset of **the Polars API**](https://narwhals-dev.github.io/narwhals/api-reference/), no need to learn anything new

From that perspective, *I think* deviating from it needs very strong motivation - and I haven't seen it here.

If you read #3363 and #2534 again - they're trying to solve the same problem - but only one deviates from Polars.
I do understand that having `order_by` provided in the same expression is *convenient* for the current Narwhals model. 
But I'd argue that adding an `order_by` parameter to many expressions isn't something that can be (easily) undone later

MarcoGorelli:
Thanks for sharing your perspective!

I appreciate your points, but I still can't accept `sort_by`

We may need to agree to disagree on this one. Perhaps just adding it but keeping it out of the stable namespaces is enough of a compromise?

MarcoGorelli:
This was discussed today on the community call and people were OK with adding an extra keyword argument that Polars doesn't have

MarcoGorelli:
thanks all for comments! shipping this then, happy to address any further comments in follow-ups

PR #3373, opened 2025-12-27
feat(expr-ir): Support `DataFrame.pivot`

### LINKED ISSUE #2179: [Enh]: Add `pivot` to Arrow

_No response_

Implement the pivot function for Arrow. Discovered as missing in #2177.

_No response_

_No response_

_No response_

dangotbanned:
# Description

Chapter 6 gets us closer to the *recently-introduced* [`DslPlan::Pivot`](https://github.com/pola-rs/polars/blob/072d83d89f35b916fabca2810ee1759179c8efbe/crates/polars-plan/src/dsl/plan.rs#L108-L118).

But you're probably reading this because the PR finally adds support for `pivot` on `pyarrow>=20` 😄

> [!IMPORTANT]
> This was a pretty complicated one

After I figured out how [`pyarrow.compute.pivot_wider`](https://arrow.apache.org/docs/dev/python/generated/pyarrow.compute.pivot_wider.html) worked, I discovered it didn't natively support:
1. Any kind of `aggregate_function`
2. Multiple `on` columns

So ... after quite a journey, I've got those working too 🥳

## Related issues

- Can be adapted to close #2179
- Child of #2572
- Follows #3371
- https://github.com/apache/arrow/pull/45562
- https://github.com/apache/arrow/issues/48679


narwhals/_plan/arrow/functions.py
```diff
@@ -1590,6 +1564,14 @@ def concat_str(
...
 ) -> ChunkedArray[StringScalar]: ...
 @t.overload
+def concat_str(
+    *arrays: Unpack[
+        tuple[str | ChunkedArrayAny, Unpack[tuple[ChunkedArrayAny | str, ...]]]
+    ],
+    separator: str = ...,
+    ignore_nulls: bool = ...,
+) -> ChunkedArray[StringScalar]: ...
+@t.overload
```
dangotbanned: Truly cursed typing, BUT it allows for this to type check

https://github.com/narwhals-dev/narwhals/blob/993c98b8e2867b1539b005c17aff7975737e780b/narwhals/_plan/arrow/dataframe.py#L493-L495

Using `str` instead of `pa.Scalar` was the only way I saw to avoid overlapping overloads.
This is also why `str` doesn't appear in the others, despite it working at runtime

narwhals/_plan/arrow/dataframe.py
```diff
@@ -343,3 +428,73 @@ def with_arrays(
...
+    """
+    result: Iterable[Any]
+    if on_columns.num_columns == 1:
+        on_column = on_columns.column(0)
+        if len(values) == 1:
+            result = on_column.to_pylist()
+        else:
+            t_left = fn.to_table(fn.array(values))
+            # NOTE: still don't know why pyarrow outputs the cross join in reverse
+            t_right = fn.to_table(on_column[::-1])
```
dangotbanned: I forgot I had a less-cryptic wrapper

```suggestion
            t_right = fn.to_table(fn.reverse(on_column))
```

narwhals/_plan/arrow/group_by.py
```diff
@@ -109,6 +110,23 @@
...
+    "mean": agg.Mean,
+    "median": agg.Median,
+    "len": agg.Len,
+}
+"""Needs to work a bit differently to the others.
+
+- Starting with a string
+- Have all of the `values` names resolved and will use the same agg
+- `{"first","last"}` disable threading
+"""
```
dangotbanned: Not very useful doc

narwhals/_plan/arrow/options.py
```diff
@@ -190,6 +196,32 @@ def split_pattern(by: str, n: int | None = None) -> pc.SplitPatternOptions:
...
 
 
+def pivot_wider(
+    on_columns: Sequence[Any] | ChunkedOrArrayAny,
+    /,
+    unexpected_key_behavior: Literal["ignore", "raise"] = "raise",
+) -> pc.FunctionOptions:
+    """Tries to wrap [`pc.PivotWiderOptions`], and raises if we're on an old `pyarrow`.
+
+    `key_names` appears to be the same as `on_columns`, but here it is required.
```
dangotbanned: - Was more of a note-to-self, but is kinda outdated
- Maybe just add an *Arguments* section instead?
- The casting before collecting to `list` is definitely worth mentioning

narwhals/_plan/arrow/options.py
```diff
@@ -190,6 +196,32 @@ def split_pattern(by: str, n: int | None = None) -> pc.SplitPatternOptions:
     return pc.SplitPatternOptions(by)
 
 
+def pivot_wider(
+    on_columns: Sequence[Any] | ChunkedOrArrayAny,
+    /,
+    unexpected_key_behavior: Literal["ignore", "raise"] = "raise",
+) -> pc.FunctionOptions:
```
dangotbanned: Remove `unexpected_key_behavior` from the signature and inline the default override instead

narwhals/_plan/compliant/dataframe.py
```diff
@@ -222,6 +222,25 @@ def join_cross(self, other: Self, *, suffix: str = "_right") -> Self: ...
...
+        separator: str = "_",
+    ) -> Self: ...
+    def pivot_agg(
+        self,
+        on: Sequence[str],
+        on_columns: Self,
+        *,
+        index: Sequence[str],
+        values: Sequence[str],
+        aggregate_function: PivotAgg,  # not sure if possible for pyarrow yet
```
dangotbanned: It was possible! 🎉 

```suggestion
        aggregate_function: PivotAgg,
```

narwhals/_plan/dataframe.py
```diff
@@ -452,6 +473,79 @@ def partition_by(
...
+            )
+            if sort_columns:
+                nw_on_cols = nw_on_cols.sort(on_)
+            on_cols = nw_on_cols._compliant
+        elif len(on_) != 1:  # pragma: no cover
+            msg = (
+                f"`on_columns: {qualified_type_name(on_columns)}` is not compatible with multiple `on` values, got:\n"
+                f"`on={on!r}`\n`on_columns={on_columns!r}`"
+            )
+            raise InvalidOperationError(msg)
```
dangotbanned: Raising for `on: list[str], on_columns: Sequence[str]` should stay, but:
1. The signature needs widening to `on_columns: Sequence[str] | Series | Self | None = None`, see [`pl.DataFrame.pivot`](https://github.com/pola-rs/polars/blob/473951bcf8c49fc23bee5ee7b8853b5dd063cb9d/py-polars/src/polars/dataframe/frame.py#L9280)
2. Most of these branches can be skipped if we already have `DataFrame`
3. The error message should suggest passing `DataFrame` for multiple columns or leaving as `None` and we'll do the work for you

narwhals/_plan/dataframe.py
```diff
@@ -261,6 +263,25 @@ def fn(values: Iterable[Any], /) -> Series[NativeSeriesT]:
...
 
+    # NOTE: Most of this is for the edge-case of a 0-column dataframe
+    # otherwise, we can reuse the method on an existing `Series` instance
+    def _parse_into_compliant_series(
+        self, other: Series[Any] | Iterable[Any], /, name: str = ""
+    ) -> CompliantSeries[NativeSeriesT]:  # pragma: no cover
+        if columns := self.columns:
+            compliant = self.get_column(columns[0])._parse_into_compliant(other)
+            return compliant if not name else compliant.alias(name)
+        tp_series = self.__narwhals_namespace__()._series
```
dangotbanned: Most likely gonna defer coverage on the *0-column* case, as it doesn't make sense for `pivot`.

Would rather have the defensive code in here now though

narwhals/_plan/arrow/dataframe.py
```diff
@@ -325,6 +337,79 @@ def partition_by(self, by: Sequence[str], *, include_key: bool = True) -> list[S
...
+    ) -> Self:
+        native = self.native
+        on_columns_ = on_columns.native
+        if len(on) == 1:
+            pivot = acero.pivot_table(native, on[0], on_columns_.column(0), index, values)
+        else:
+            temp_name = temp.column_name(native.column_names)
+            on_columns_w_idx = on_columns.with_row_index(temp_name)
+            on_columns_encoded = on_columns_w_idx.get_column(temp_name).native
+            single_on = self.join_inner(on_columns_w_idx, list(on)).drop(on).native
```
dangotbanned: Really bugging me that this is identical to:

https://github.com/narwhals-dev/narwhals/blob/993c98b8e2867b1539b005c17aff7975737e780b/narwhals/_plan/arrow/dataframe.py#L387-L390

I'd prefer it to be written once, but can't think of a meaningful name and there a slight difference that follows:


https://github.com/narwhals-dev/narwhals/blob/993c98b8e2867b1539b005c17aff7975737e780b/narwhals/_plan/arrow/dataframe.py#L359-L361

https://github.com/narwhals-dev/narwhals/blob/993c98b8e2867b1539b005c17aff7975737e780b/narwhals/_plan/arrow/dataframe.py#L392-L394

narwhals/_plan/arrow/acero.py
```diff
@@ -320,6 +334,22 @@ def group_by_table(
...
 
+def pivot_table(
+    native: pa.Table,
+    on: str,
+    on_columns: ChunkedOrArrayAny | Sequence[Any],
+    /,
+    index: Sequence[str],
+    values: Sequence[str],
+) -> pa.Table:
+    """Partial `pivot` implementation."""
```
dangotbanned: Should explain what *Partial* means!

The table returned will have a schema like this:

<details><summary>Show invented syntax</summary>
<p>

```py
output_schema = {
    index[0]: "<input_type>",
    index[-1]: "<input_type>",
    values[
        0
    ]: "struct<on_columns[0]: <values[0]_type>, on_columns[-1]: <values[0]_type>>",
    values[
        -1
    ]: "struct<on_columns[0]: <values[-1]_type>, on_columns[-1]: <values[-1]_type>>",
}
```

</p>
</details> 



## Examples
This simple pivot specification will have returned:
https://github.com/narwhals-dev/narwhals/blob/65f3f37f81207f487ed59982b44352444fb9e4f4/tests/plan/pivot_test.py#L378-L383

```py
{'name': String, 'test_1': Struct({'maths': Int64, 'physics': Int64})}
```

That is great for the simpler cases.
But when there are multiple columns defined in `on` - the column names will have been replaced with indices.

Each index refers to a row in the `on_columns` table, that provides the components of the final column name

https://github.com/narwhals-dev/narwhals/blob/65f3f37f81207f487ed59982b44352444fb9e4f4/tests/plan/pivot_test.py#L292-L298

```py
{'idx_1': Int64, 'foo': Struct({'0': Int64, '1': Int64, '2': Int64, '3': Int64})}
```

And when we have multiple `values` columns, those indicies are repeated in each struct:

https://github.com/narwhals-dev/narwhals/blob/65f3f37f81207f487ed59982b44352444fb9e4f4/tests/plan/pivot_test.py#L299-L312

```py
{
    "idx_1": Int64,
    "foo": Struct({"0": Int64, "1": Int64, "2": Int64, "3": Int64}),
    "bar": Struct({"0": Int64, "1": Int64, "2": Int64, "3": Int64}),
}
```

dangotbanned: ## TODO
- Explore concatenating the `on` columns **first** instead of using an index
- Thinking it would simplify everything if the struct always had field names like `'{"b","X"}'` already
  - No need to generate an index
  - Skip an initial join
  - The final rename code only needs to handle the multiple `values` prefix bits
    - Where the prefix is already the name of the struct column itself
    - No need for `product` or a cross join
  - Prefixing for single `on` but multiple `values` is then identical to multiple `on`

dangotbanned: >     * Explore concatenating the `on` columns **first** instead of using an index

Yeah that was a good idea and led to ([refactor: Fully merge `pivot`, `pivot_agg` into 1 method](https://github.com/narwhals-dev/narwhals/pull/3373/commits/8cec922d991eb495bb5ed632331a5e6fad787e19))


Next step is making use of [`pyarrow.Table.flatten`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.flatten) and piggybacking off the generated names

dangotbanned: Aaaaaaand that's resolved now with ([perf: Huge simplify unnest/rename](https://github.com/narwhals-dev/narwhals/pull/3373/commits/dabb2666585cd0f61d8f3321ec06612d766f279a))

narwhals/_plan/arrow/dataframe.py
dangotbanned: Resolved in https://github.com/narwhals-dev/narwhals/pull/3373#discussion_r2649827854

PR #1571, opened 2024-12-12
feat: add Decimal dtype

MarcoGorelli:
closes #1287 

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #3377, opened 2025-12-30
feat: Improve support for `Decimal` DType

### LINKED ISSUE #3375: [Enh]: Improve support for Decimal dtype
Polars [Decimal dtype](https://docs.pola.rs/api/python/dev/reference/api/polars.datatypes.Decimal.html) support is no longer unstable (as of polars 1.35.0, https://github.com/pola-rs/polars/pull/25020).

It might be time for us to have better support for it as well

FBruzzesi:
# Description

Adds `precision` and `scale` parameters to `Decimal` dtype.

**TODO**: Decide on stable API

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3375
- **Blocking** https://github.com/FBruzzesi/anyschema/issues/101

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_ibis/utils.py
+++ b/narwhals/_ibis/utils.py
@@ -174,9 +174,14 @@ def native_to_narwhals_dtype(ibis_dtype: IbisDataType, version: Version) -> DTyp
             ]
         )
     if is_decimal(ibis_dtype):
-        return dtypes.Decimal(
-            precision=ibis_dtype.precision, scale=(ibis_dtype.scale or 0)
-        )
+        # Same default as in ibis.Decimal.{to_polars, to_pyarrow}
+        # https://github.com/ibis-project/ibis/blob/5028d8e5e1921d5de48f59fad48b86c0de541b0d/ibis/formats/polars.py#L82-L86
+        # https://github.com/ibis-project/ibis/blob/5028d8e5e1921d5de48f59fad48b86c0de541b0d/ibis/formats/pyarrow.py#L165-L177
+        # According to their own comment:
+        # > set default precision and scale to something; unclear how to choose this
+        # source: (see https://github.com/ibis-project/ibis/blob/5028d8e5e1921d5de48f59fad48b86c0de541b0d/ibis/formats/pyarrow.py#L166)
+        scale = 9 if ibis_dtype.scale is None else ibis_dtype.scale
+        return dtypes.Decimal(precision=ibis_dtype.precision, scale=scale)
     if ibis_dtype.is_time():
         return dtypes.Time()
     if ibis_dtype.is_binary():


FBruzzesi:
Regarding the stable API, how do we feel for this feature? I would say that keeping it backward compatible for v2 might be a stretch, but for v1 it would be ok. cc @MarcoGorelli

MarcoGorelli:
thanks for your pr!

i was under the impression that `precision` was going to be removed, but it looks like that didn't happen, i'll study https://github.com/pola-rs/polars/issues/19784 and get back to this

dangotbanned:
> thanks for your pr!
> 
> i was under the impression that `precision` was going to be removed, but it looks like that didn't happen, i'll study [pola-rs/polars#19784](https://github.com/pola-rs/polars/issues/19784) and get back to this

To save you a long read, there was nothing conclusive in that issue.
*But* the PR that closed it does have a summary that's more digestible 

- https://github.com/pola-rs/polars/pull/24542


On the Python-side, all that's changed is `precision=38` gets assigned if you don't provide one to `Decimal`.

There are a lot more changes on the Rust-side (who would've guessed that?)
It appears that `precision` is here to stay, but there are a lot of (positive?) behavioral changes which we may have a tough time trying to align between backends.

I'm not pitching we do/don't *do that* - just trying to summarize the old-fashioned way 😉

narwhals/dtypes.py
```diff
@@ -205,16 +205,44 @@ class NestedType(DType):
...
+        Decimal(precision=2, scale=1)
     """
 
+    __slots__ = ("precision", "scale")
+
+    precision: int
+    scale: int
+
+    def __init__(self, precision: int | None = None, scale: int = 0) -> None:
+        self.precision = 38 if precision is None else precision
```
MarcoGorelli: not sure what i'm missing but why not
```
precision: int = 38
```
?

MarcoGorelli (approved):
thanks @FBruzzesi !

narwhals/dtypes.py
dangotbanned: I suppose if we got an old `pl.Decimal`, that could have a `None` already?

Otherwise - yeah if we can avoid introducing the contructor with a `None` - I'm on board 😄

FBruzzesi: It's solely to mirror polars signature: [polars.datatypes.Decimal](https://docs.pola.rs/api/python/dev/reference/api/polars.datatypes.Decimal.html) and [code](https://github.com/pola-rs/polars/blob/778dbb645ccbff8b1e5999a279037571a03c718b/py-polars/src/polars/datatypes/classes.py#L450-L459)

```py
    def __init__(
        self,
        precision: int | None = None,
        scale: int = 0,
    ) -> None:
        if precision is None:
            precision = 38


        self.precision = precision
        self.scale = scale
```

FBruzzesi:
@MarcoGorelli thanks for reviewing this. I have a couple of questions that might need to be discussed before merging:

1. Stable API - What's your take here?
2. Polars Decimal allows for any value of precision and scale at datatype initialization, however it raises an `InvalidOperationError` at runtime if scale is greater than precision. See example below. I would argue that to standardize the error message, we could raise such error during initialization.

Example:

```py
import polars as pl

dtype = pl.Decimal(2, 3)  # Ok so far
pl.Series([1.2]).cast(dtype)  
> InvalidOperationError: scale must be less than or equal to precision
```

dangotbanned:
@FBruzzesi 

RE: (https://github.com/narwhals-dev/narwhals/pull/3377#issuecomment-3699989491), (https://github.com/narwhals-dev/narwhals/pull/3377#issuecomment-3730539869)
I thought `Decimal` was marked as unstable in `narwhals`, but it looks like it wasn't? 🤔 

@MarcoGorelli did you mean to introduce it as stable in #1571?
`polars` didn't stablize until a month after https://github.com/pola-rs/polars/pull/24542 in https://github.com/pola-rs/polars/pull/25020

dangotbanned:
> 2. Polars Decimal allows for any value of precision and scale at datatype initialization, however it raises an `InvalidOperationError` at runtime if scale is greater than precision. 
> See example below. I would argue that to standardize the error message, we could raise such error during initialization.


@FBruzzesi I agree, let's do that

> [!NOTE]
> Sorry for blowing up CI with `pandas>=3` warnings 🫣 
> https://github.com/narwhals-dev/narwhals/pull/3377/checks?sha=075ceb86353c6a511c068ea2bbf5986d83e3ccd9

dangotbanned:
> Regarding the stable API, how do we feel for this feature? I would say that keeping it backward compatible for v2 might be a stretch, but for v1 it would be ok

@FBruzzesi Initially I was thinking we should do the same thing for `v1` and `v2`.
The only precedent we have is #2192, where the attributes were only added in `main`.

**But** the more I've read through your PR, I've started to think we might not need it?
AFAICT, for most backends the issue *would have* been that we change the native data types that *we produce*.
It looks like we didn't do that much: 
- Raised instead of producing a native data type
  - `arrow`, `duckdb`, `pandas_like`, `polars`, `spark_like`
- Continues to not have native support
  - `dask`

> [!NOTE]
> **Only `ibis`** could see a runtime change by using our new defaults

Have I understood all of this correctly?
If so, I would suggest adding (if needed) version-specific behavior for `ibis`-only, instead of an entire `DType`

MarcoGorelli:
> @MarcoGorelli thanks for reviewing this. I have a couple of questions that might need to be discussed before merging:
> 
> 1. Stable API - What's your take here?

seems fine, it should be stable in polars now

> 2. Polars Decimal allows for any value of precision and scale at datatype initialization, however it raises an `InvalidOperationError` at runtime if scale is greater than precision. See example below. I would argue that to standardize the error message, we could raise such error during initialization.
> 
> Example:
> 
> ```python
> import polars as pl
> 
> dtype = pl.Decimal(2, 3)  # Ok so far
> pl.Series([1.2]).cast(dtype)  
> > InvalidOperationError: scale must be less than or equal to precision
> ```

Sure sounds good (tbh i wouldn't have worried about standardising it, but no objections as you've already done it)

FBruzzesi:
Thanks both for reviewing 🙏🏼

@dangotbanned as mentioned in discord, I did some digging in the ibis code and settled on https://github.com/narwhals-dev/narwhals/pull/3377/commits/f98deccd76f8f3681527349371afcced9d2b9a7d

Let me know if you are happy with the state of this PR 😇

PR #3387, opened 2026-01-04
test: unxfail sqlframe for `list.median`

raisadz:
# Description

Based on the suggested solution in https://github.com/eakmanrq/sqlframe/issues/568, we can unxfail sqlframe for `list.median`.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/expr_list.py
+++ b/narwhals/_spark_like/expr_list.py
@@ -75,7 +75,9 @@ class SparkLikeExprListNamespace(
             if self.compliant._implementation.is_sqlframe():
                 # indexing is different in sqlframe
                 # https://github.com/eakmanrq/sqlframe/issues/568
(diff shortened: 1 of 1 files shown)


MarcoGorelli (approved):
thanks @raisadz !

> A workaround though would be to add 1 to your index column and then it will return the correct result. If you are wanting to maintain compatibility with pyspark you could do a check if running in SQLFrame and if so then add 1.

😄 sure, i guess this is what we have to do

Issue #3386, opened 2026-01-03
[Enh]: Add support for `concat(..., how={"vertical_relaxed", "diagonal_relaxed"})`


FBruzzesi:
Opening an issue to track the reason why we are not moving forward with #3191

To implement the relaxed variants, we need to implement the casting rules between dtypes (See https://github.com/narwhals-dev/narwhals/pull/3191#issuecomment-3389117044)

dangotbanned:
*You* probably know already that I'm quite keen to support `polars`'s supertyping rules 😄 

Just thought I'd add that it is on my list for #2572:

>  * [ ]  Review [`DataType` propagation](https://github.com/pola-rs/polars/blob/b5956c30ef578949f87189330c036f86da32c516/crates/polars-plan/src/plans/aexpr/schema.rs) and [supertyping](https://github.com/pola-rs/polars/blob/b5956c30ef578949f87189330c036f86da32c516/crates/polars-core/src/utils/supertype.rs) rules
>    * It's looking like some version of this will be required to correctly rewrite `LogicalPlan`s
>    * Otherwise, each sub-plan has to be eager and request a new schema from the backend after processing the previous step
>      * That's current `narwhals` behavior, but likely more of an issue when `Selector`s can be used in more places
>  * A [little write-up](https://github.com/narwhals-dev/narwhals/blob/63dbc65bc68d6dea849f927711da46c287f1a029/narwhals/_plan/logical_plan/plan.py#L29-L85) comparing `oh-nodes` to `polars` internals

I'm more than happy to help out/work on it if anyone else is interested

PR #3393, opened 2026-01-07
chore: Add `DType.__call__`

dangotbanned:
# Description
As part of #3386, @FBruzzesi and I identified an issue with simply reusing `IntoDType.__eq__` to compare `DType`s.

When we start **nesting** `DType`s, there are cases like this where a change from a default makes the `Schema` incompatible:

```py
import datetime as dt

import polars as pl

data = {
    "a": [1, 2, 3],
    "b": ["hello", None, "its me again"],
    "c": [
        dt.datetime(2000, 1, 1, 12, 0, 0),
        dt.datetime(2000, 1, 1, 13, 0, 0),
        dt.datetime(2000, 2, 1, 9, 0, 0),
    ],
}

base = pl.DataFrame(data, schema_overrides={"c": pl.Datetime("us")})  # the default, but for visibility
collect_struct = pl.struct(pl.all()).alias("struct")
base_struct = base.select(collect_struct)
c = pl.col("c")

diff_time_unit = base.with_columns(c.dt.cast_time_unit("ms")).select(collect_struct)     # ✔️ 
diff_time_zone = base.with_columns(c.dt.convert_time_zone("UTC")).select(collect_struct) # ❌
```

Here, different `time_unit`s are allowed and we should just pick the less-precise one:

```py
pl.concat([base_struct, diff_time_unit], how="vertical_relaxed").schema
```
```
Schema([('struct',
         Struct({'a': Int64, 'b': String, 'c': Datetime(time_unit='ms', time_zone=None)}))])
```

But if the difference is on `time_zone`s - we should raise:

```py
pl.concat([base_struct, diff_time_zone], how="vertical_relaxed")
```
```
SchemaError: failed to determine supertype of struct[3] and struct[3]
```

[This commit](https://github.com/narwhals-dev/narwhals/commit/bdbb1c5730583508e931af532b61fb20c31eaa4a) shows how many `isinstance` checks and `import`s that adding `DType.__call__` saves.
The cost might seem small, but it adds up once you consider `concat` need to do it for:
- a pairwise comparison of many `Schema`s
- each field may contain any number of nested `DType`s
- those `DType`s can also be nested 😳 

Caching **will help**, but since this problem can scale in multiple directions - it's probably best we minimize anything we can 😄 

- [x] 🔧 Optimization
- [x] 🐳 Other

## Related issues

- https://github.com/narwhals-dev/narwhals/commit/bdbb1c5730583508e931af532b61fb20c31eaa4a
- #3386

Needed to see if this broke downstream ci, but all unrelated:
- [`darts` - unrelated](https://github.com/narwhals-dev/narwhals/actions/runs/20788242327/job/59703638781?pr=3393)
- [`marimo` - unrelated](https://github.com/narwhals-dev/narwhals/actions/runs/20788242327/job/59703638831?pr=3393)
- [`scikit-lego` unrelated](https://github.com/narwhals-dev/narwhals/actions/runs/20788242327/job/59703638846?pr=3393)

### DIFF
--- a/narwhals/dtypes.py
+++ b/narwhals/dtypes.py
@@ -176,6 +176,31 @@ class DType(metaclass=DTypeClass):
         return hash(self.__class__)
 
     def __call__(self) -> Self:
+        # NOTE: (Internal doc)
+        # But, why?
+        #
+        # Let's say we have a function that looks like this:
+        #
+        #     >>> import pyarrow as pa
+        #     >>> from narwhals.dtypes import DType
+        #     >>> from narwhals.typing import NonNestedDType
+        #
+        #     >>> def convert(dtype: DType | type[NonNestedDType]) -> pa.DataType: ...
+        #
+        #
+        # If we need to resolve the union to a *class*, we can use:
+        #
+        #     >>> always_class = dtype.base_type()
+        #
+        # But what if instead, we need an *instance*?:
+        #
+        #     >>> always_instance_1 = dtype if isinstance(dtype, DType) else dtype()
+        #     >>> always_instance_2 = dtype() if isinstance(dtype, type) else dtype
+        #
+        # By defining `__call__`, we can save an instance check and keep things readable:
+        #
+        #    >>> always_instance = dtype()
+        #    >>> always_class = dtype.base_type()
         return self
 
 


narwhals/dtypes.py
```diff
@@ -175,6 +175,34 @@ def __eq__(self, other: DType | type[DType]) -> bool:  # type: ignore[override]
...
+        # But what if instead, we need an *instance*?:
+        #
+        #     >>> always_instance_1 = dtype if isinstance(dtype, DType) else dtype()
+        #     >>> always_instance_2 = dtype() if isinstance(dtype, type) else dtype
+        #
+        # By defining `__call__`, we can save an instance check and keep things readable:
+        #
+        #    >>> always_instance = dtype()
+        #    >>> always_class = dtype.base_type()
+        return self
```
dangotbanned, line 204: I'm intentionally not making this public API, but want there to be a reminder on how it helps 🙂

tests/dtypes/dtypes_test.py
```diff
@@ -585,6 +586,52 @@ def test_dtype_base_type_nested(nested_dtype: NestedOrEnumDType) -> None:
...
 
 
+def test_dtype___call___(non_nested_type: type[NonNestedDType]) -> None:
+    dtype = non_nested_type()
+    assert dtype == non_nested_type()()
+    assert dtype() is dtype
+
+
+def _resolve_base_type_dtype(into: IntoDType, /) -> tuple[type[DType], DType]:
+    return into.base_type(), into()
```
dangotbanned, line 596: I suppose this is a TL;DR

We can extract the class and instance from `IntoDType` - only paying the cost of initialization *if* we needed to anyway

narwhals/dtypes.py
FBruzzesi: 100% agree to avoid making this public 👌🏼

camriddell:
I'm not a huge fan of relying on `__call__` for this behavior because it becomes very hard to tell whether you are instantiating an object or produce a no-op by reading the code alone. Instead, we could use a descriptor attached to a semantically meaningful name that creates a bit more clarity as to what the intent of the code is at the call site.

```python
from typing import ClassVar

class Instance[T]:
    def __get__(self, instance: T | None, owner: type[T]) -> T:
        if instance is None:
            return owner()
        return instance

class Dtype:
    instance: ClassVar[Instance[Dtype]] = Instance()


print(Dtype().instance) # <__main__.Dtype object at 0x7512677f5940>
print(Dtype.instance)   # <__main__.Dtype object at 0x751267771450>
```

MarcoGorelli:
thanks @camriddell - yeah i think i prefer that too

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3393#issuecomment-3729332185

Thanks for the input @camriddell! 🙂 

I'm having trouble writing up something concise.
For now I just wanna zoom into what the two look like side-by-side, in the context that I'm using `DType.__call__`:

```py
def list_supertype_current(left: List, right: List, version: Version) -> List | None:
    if inner := get_supertype(left.inner(), right.inner(), version=version):
        return List(inner)
    return None


def list_supertype_proposed(left: List, right: List, version: Version) -> List | None:
    if inner := get_supertype(left.inner.instance, right.inner.instance, version=version):
        return List(inner)
    return None
```


in relation to ...

> I'm not a huge fan of relying on `__call__` for this behavior 
> because it becomes very hard to tell whether you are instantiating an object or produce a no-op by reading the code alone. 


I think I understand your concern, but this swings too far in the opposite direction IMO.

Let's pretend we don't know what the descriptor protocol is for a moment.
Which version *looks like* a call to constructor?

```py
left.inner()
left.inner.instance
```

From reading the code, the second looks like an attribute access and is hiding that it could be constructing a `DType`.

If we could only highlight *doing* work vs *not doing* work - surely a silent no-op is preferable to the inverse, right?

FBruzzesi:
Let me chime in the discussion as well as I _kind of_ started both:

- The check for `left_inner = left.inner if isinstance(left.inner, DType) else left.inner()`
- The discussion with @camriddell to find a different solution than implementing `__call__`

---

## Theory

In theory, nested types allow for the inner type to be any `IntoDType`

https://github.com/narwhals-dev/narwhals/blob/d7bb050e91a6a6528eba5fcf8e81fa8bfeecaa8b/narwhals/typing.py#L278-L279

which means that also classes rather than object are allowed.

This aligns with what polars does - I assume to improve ergonomics for users - for example in `Expr.cast` it's possible to pass `{pl, nw}.Int8` or `{pl, nw}.List(inner={pl, nw}.Int8)`.

## Practice

In practice, _if_ our only use case for `get_supertype` is to promote datatypes in `concat(..., how={"vertical_relaxed", "diagonal_relaxed"})` or more generally for internal use only, then it should _never_ be the case that the inner type of a nested type is not initialized. In fact all the `native_to_narwhals_dtype(dtype: NativeDType, ...) -> DType` return an instance and never the class, and that's what we use in `collect_schema` methods.

In other words, the only case in which we can get a class rather than an instance is when a user passes it.

## In conclusion

I mostly wanted to state the obvious,  trying to bring the attention to the fact that we are solution-ing/debating for a non-problem as we may not need to do the `inner()` call in the first place. The main reason to have it is to be type safe/make the type checker happy

PR #3397, opened 2026-01-10
fix(test): Change error message for polars, un-xfail sqlframe `list.mean`

FBruzzesi:
# Description

Drive by polars 1.37 changing the error message

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/expr_list.py
+++ b/narwhals/_spark_like/expr_list.py
@@ -72,12 +72,11 @@ class SparkLikeExprListNamespace(
             sorted_expr = F.array_compact(F.sort_array(expr))
             size = F.array_size(sorted_expr)
             mid_index = (size / 2).cast("int")
-            if self.compliant._implementation.is_sqlframe():
+            impl = self.compliant._implementation
+            if impl.is_sqlframe() and impl._backend_version() < (3, 44, 1):
                 # indexing is different in sqlframe
                 # https://github.com/eakmanrq/sqlframe/issues/568
-                mid_index = mid_index + 1
-            else:
-                pass  # pragma: no cover
+                mid_index = mid_index + 1  # pragma: no cover
             odd_case = sorted_expr[mid_index]
             even_case = (sorted_expr[mid_index - 1] + sorted_expr[mid_index]) / 2
             return (


FBruzzesi:
This seems a regression in SQLFrame/SQLGlot

narwhals/_spark_like/expr_list.py
```diff
@@ -72,12 +72,11 @@ def func(expr: Column) -> Column:
...
             mid_index = (size / 2).cast("int")
-            if self.compliant._implementation.is_sqlframe():
+            impl = self.compliant._implementation
+            if impl.is_sqlframe() and impl._backend_version() < (3, 44, 1):
                 # indexing is different in sqlframe
                 # https://github.com/eakmanrq/sqlframe/issues/568
-                mid_index = mid_index + 1
-            else:
-                pass  # pragma: no cover
+                mid_index = mid_index + 1  # pragma: no cover
```
FBruzzesi, line 79: Hey @eakmanrq 👋🏼 Sorry for the direct ping. I just wanted to make sure that whatever changed since your [comment](https://github.com/eakmanrq/sqlframe/issues/568#issuecomment-3707149841) is expected.

TL;DR as of yesterday's SQLFrame version, we don't need to do `mid_index = mid_index + 1` to find the right index for each row

eakmanrq: Yeah I changed the logic to use brackets by default for the lookup which means this workaround isn't needed anymore. Thanks for checking.

tests/frame/get_column_test.py
```diff
@@ -12,7 +12,8 @@ def test_get_column(constructor_eager: ConstructorEager) -> None:
     assert_equal_data({"a": result}, {"a": [1, 2]})
     assert result.name == "a"
     with pytest.raises(
-        (KeyError, TypeError), match=r"Expected str|'int' object cannot be converted|0"
+        (KeyError, TypeError),
+        match=r"Expected str|'int' object cannot be (converted|cast)|0",
```
dangotbanned, line 16: Should we split this one out and merge ahead of the `sqlframe` stuff until we know more on (https://github.com/narwhals-dev/narwhals/pull/3397/changes#r2678798597)?

narwhals/_spark_like/expr_list.py
FBruzzesi: Thanks for the quick feedback! We'll merge this then 🙏🏼

tests/frame/get_column_test.py
FBruzzesi: Merging now with all changes 😉

dangotbanned: On the pattern itself, which has this failure on `main`

```py
FAILED tests/frame/get_column_test.py::test_get_column[polars[eager]] - AssertionError: Regex pattern did not match.
  Expected regex: "Expected str|'int' object cannot be converted|0"
  Actual message: "argument 'name': 'int' object cannot be cast as 'str'"
```

Could we just do this?

```py
with pytest.raises((KeyError, TypeError), match="int")
```

dangotbanned: > Merging now with all changes 😉

oof 🫠

PR #3396, opened 2026-01-10
feat(DRAFT): Add `get_supertype`

dangotbanned:
# Description
> [!IMPORTANT]
> @FBruzzesi and I have been + are still iterating on this
> Core functionality is there, focusing on readability, performance + shrinking the test suite

This PR implements `polars`' concept of [supertyping](https://github.com/pola-rs/polars/blob/529f7ec642912a2f15656897d06f1532c2f5d4c4/crates/polars-core/src/utils/supertype.rs#L142-L543) - which more generally defines which types can be safely promoted/demoted/cast to other types.

I really like the DuckDB visualization of their version[^1] of these rules, so here's that for an example:

[^1]: DuckDB also mentions another set of rules called [Combination Casting](https://duckdb.org/docs/stable/sql/data_types/typecasting#combination-casting) - that is entirely implicit. 
The matrix doesn't relfect these and only one cast example is given, but it would apply to `nw.concat`:
*"This *combination casting* occurs for ..., set operations (`UNION` / `EXCEPT` / `INTERSECT`), and ..."*

<details><summary><b>Show Casting Operations Matrix</b></summary>
<p>

typecasting-matrix

</p>
</details> 

This is a **preliminary step** for implementing relaxed `concat` (#3386). 
The aim is **we own** a consistent set of rules that all/most backends can participate in.
We've already dropped some *supertypes* that are valid in `polars`, but *may* prove challenging in other backends such as #121.
Some others are directly mentioned in comments (e.g. `(Struct, DType) -> Struct`)

## Additional use-cases
*Supertyping* in `polars` is used for **much more** than just a subset of `concat`.
In (#2572), it is one of the larger concepts missing from the intermediate representation (see https://github.com/narwhals-dev/narwhals/issues/3386#issuecomment-3707241893).

[`polars-plan::plans::conversion::type_coercion`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/type_coercion) is full of examples of how deeply related the concept is with expressions.
**My aim is not to reproduce all of that** 😅 - but to be able to reason about `DType`s between `LazyFrame` operations *without* querying the backend for a `Schema` between every step 🤞 

## Related issues

- Prep for #3386
- #3393
- [x] #3377
- [ ] (https://github.com/pola-rs/polars/issues/22435)
- [ ] (https://github.com/pola-rs/polars/issues/27726)

## Tasks
- [x] Testing
[...]

PR #3398, opened 2026-01-11
feat: Add `concat(..., how="*_relaxed"})`

### LINKED ISSUE #3386: [Enh]: Add support for `concat(..., how={"vertical_relaxed", "diagonal_relaxed"})`

FBruzzesi:
# Description

As Marco often says, _in for a penny in for a pound_ 😂

This PR builds up on #3396 to showcase how that's useful to implement concat relaxed versions.
It's probably possible to lower some of the repetition, but considering how much time I spent on this (not _sooo_ much so far), I am quite happy with the shape of it.

On top of the new feature there are some minor adjustment for the concat implementations across backends

TODO:

- [x] Unknown types for narwhals should keep the original backend dtype and let such backend to deal with it. See comment https://github.com/narwhals-dev/narwhals/pull/3398#discussion_r2749499488
- [ ] Backward compatibility: this is more a concern for `how={"vertical", "diagonal"}`, and would be a followup to this PR

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3386 

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -268,7 +268,7 @@ def test_concat_diagonal_relaxed(
     assert_equal_data(result.collect(), expected_data)
 
 
-@pytest.mark.skipif(PYARROW_VERSION < (19, 0, 0))
+@pytest.mark.skipif(PYARROW_VERSION < (19, 0, 0), reason="Too old for pyarrow.uuid type")
 def test_pyarrow_concat_vertical_uuid() -> None:
     # Test that concat vertical and vertical_relaxed preserves unsupported types like UUID
     pa = pytest.importorskip("pyarrow")


narwhals/_pandas_like/utils.py
```diff
@@ -704,3 +707,36 @@ class PandasLikeSeriesNamespace(EagerSeriesNamespace["PandasLikeSeries", Any]):
 
 def make_group_by_kwargs(*, drop_null_keys: bool) -> dict[str, bool]:
     return {"sort": False, "as_index": True, "dropna": drop_null_keys, "observed": True}
+
+
+_DTYPE_BACKEND_PRIORITY: dict[DTypeBackend, Literal[0, 1, 2]] = {
+    "pyarrow": 2,
+    "numpy_nullable": 1,
+    None: 0,
+}
```
FBruzzesi, line 714: _Strong_ opinions here!

Disclaimer: pandas automatically does something similar already when concatenating dataframes vertically:

- `{"pyarrow", None} -> "pyarrow"`
- `{"numpy_nullable", None} -> "numpy_nullable"`
- `{"numpy_nullable", "pyarrow"} -> dtype('O')`  😭 WHY THOUGH?

narwhals/schema.py
```diff
@@ -363,3 +365,41 @@ def _from_pandas_like(
...
+        if promoted_dtype := get_supertype(left=ltype, right=rtype):
+            into_out_schema[lname] = promoted_dtype
+        else:
+            msg = f"failed to determine supertype of {ltype} and {rtype}"
+            raise SchemaMismatchError(msg)
+
+    return Schema(into_out_schema)
+
+
+def combine_schemas(left: Schema, right: Schema) -> tuple[Schema, Schema]:
```
FBruzzesi: Reading this over, `combine` is definitely the wrong verb here as we end up getting two schemas _and_ they preserve their original dtypes.

I think my first idea was to follow the meshing of the schemas with the `to_supertype` call, which indeed is always happening when we call this function in the codebase. I will think about a better name but open to suggestions

tests/frame/concat_test.py
```diff
@@ -1,13 +1,22 @@
...
 from narwhals.exceptions import InvalidOperationError
+from narwhals.schema import Schema
 from tests.utils import Constructor, ConstructorEager, assert_equal_data
 
+if TYPE_CHECKING:
+    from narwhals.typing import LazyFrameT
+
+
+def _cast(frame: LazyFrameT, schema: Schema) -> LazyFrameT:
+    return frame.select(nw.col(name).cast(dtype) for name, dtype in schema.items())
```
FBruzzesi: I still cry each time I try to reach out for `{DataFrame,LazyFrame}.cast` 😂

narwhals/schema.py
dangotbanned: Wasn't this called `merge_schemas` or something in polars?

narwhals/_duckdb/namespace.py
```diff
@@ -82,23 +83,62 @@ def _coalesce(self, *exprs: Expression) -> Expression:
...
         if how == "diagonal":
-            res = first.native
-            for _item in native_items[1:]:
+            res, *others = (item.native for item in items)
+            for _item in others:
+                # TODO(unassigned): use relational API when available https://github.com/duckdb/duckdb/discussions/16996
+                res = duckdb.sql("""
+                    from res select * union all by name from _item select *
+                """)
+            return first._with_native(res)
```
FBruzzesi, line 112: I was trying to build a one query instead of recursing on it. Something like:

```py
query = " union all by name ".join(...)
```

Now I was not able to figure out what to put in the "..." 😂 I will give it another try in isolation

FBruzzesi: Here https://github.com/narwhals-dev/narwhals/pull/3399

tests/frame/concat_test.py
dangotbanned: I want it too 😢

narwhals/schema.py
```diff
@@ -363,3 +365,41 @@ def _from_pandas_like(
...
+        if lname != rname:
+            msg = f"schema names differ: got {rname}, expected {lname}"
+            raise ComputeError(msg)
+        if promoted_dtype := get_supertype(left=ltype, right=rtype):
+            into_out_schema[lname] = promoted_dtype
+        else:
+            msg = f"failed to determine supertype of {ltype} and {rtype}"
+            raise SchemaMismatchError(msg)
+
+    return Schema(into_out_schema)
```
dangotbanned: @FBruzzesi WDYT of this version?

It avoids the intermediate `into_out_schema`, by passing a generator to `Schema(...)`

I've pretty much just moved things around, but the name checking only goes one-by-one if we already know we've failed

```suggestion
def _supertype(left: DType, right: DType) -> DType:
    if promoted_dtype := get_supertype(left, right):
        return promoted_dtype
    msg = f"failed to determine supertype of {left} and {right}"
    raise SchemaMismatchError(msg)


def _ensure_names_match(left: Schema, right: Schema) -> tuple[Schema, Schema]:
    if len(left) != len(right):
        msg = "schema lengths differ"
        raise ComputeError(msg)
    if left.names() != right.names():
        it = ((lname, rname) for (lname, rname) in zip(left, right) if lname != rname)
        lname, rname = next(it)
        msg = f"schema names differ: got {rname}, expected {lname}"
        raise ComputeError(msg)
    return left, right


def to_supertype(left: Schema, right: Schema) -> Schema:
    # Adapted from polars https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-core/src/schema/mod.rs#L83-L96
    left, right = _ensure_names_match(left, right)
    it = zip(left.keys(), left.values(), right.values())
    return Schema((name, _supertype(ltype, rtype)) for (name, ltype, rtype) in it)
```

narwhals/schema.py
dangotbanned: >  I will think about a better name but open to suggestions

@FBruzzesi I've got an idea that does a similar thing to ([perf: Avoid unnecessary `lambda`s](https://github.com/narwhals-dev/narwhals/pull/3398/commits/5bffccd79a9054764df765d31078d8b3ae0b4173))

Since every use of `combine_schema` *currently* looks like this:

```py
lambda x, y: to_supertype(*combine_schemas(x, y)),
```

We can just define that as the function that gets passed to `reduce` *instead*.

I used the name `to_supertype_diagonal` *for now*, but I'm not too attached it 😄 

<details><summary><b>Show diff</b> (sorry, very big)</summary>
<p>

```diff
diff --git a/narwhals/_arrow/namespace.py b/narwhals/_arrow/namespace.py
index bc450265a..aabb81ec2 100644
--- a/narwhals/_arrow/namespace.py
+++ b/narwhals/_arrow/namespace.py
@@ -19,7 +19,7 @@ from narwhals._expression_parsing import (
     combine_evaluate_output_names,
 )
 from narwhals._utils import Implementation
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
 
 if TYPE_CHECKING:
     from collections.abc import Iterator, Sequence
@@ -180,7 +180,7 @@ class ArrowNamespace(
     def _concat_diagonal_relaxed(self, dfs: Sequence[pa.Table], /) -> pa.Table:
         native_schemas = tuple(table.schema for table in dfs)
         out_schema = reduce(
-            lambda x, y: to_supertype(*combine_schemas(x, y)),
+            to_supertype_diagonal,
             (Schema.from_arrow(pa_schema) for pa_schema in native_schemas),
         ).to_arrow()
         to_schemas = (
diff --git a/narwhals/_dask/namespace.py b/narwhals/_dask/namespace.py
index 5254744ea..5114ee435 100644
--- a/narwhals/_dask/namespace.py
+++ b/narwhals/_dask/namespace.py
@@ -24,7 +24,7 @@ from narwhals._expression_parsing import (
 )
 from narwhals._pandas_like.utils import promote_dtype_backend
 from narwhals._utils import Implementation, zip_strict
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator
@@ -177,7 +177,7 @@ class DaskNamespace(
             dtypes = tuple(df.dtypes.to_dict() for df in dfs)
             dtype_backend = promote_dtype_backend(dfs, self._implementation)
             out_schema = reduce(
-                lambda x, y: to_supertype(*combine_schemas(x, y)),
+                to_supertype_diagonal,
                 (Schema.from_pandas_like(dtype) for dtype in dtypes),
             ).to_pandas(dtype_backend=dtype_backend.values())
 
diff --git a/narwhals/_duckdb/namespace.py b/narwhals/_duckdb/namespace.py
index 41c2e8d29..37f773de8 100644
--- a/narwhals/_duckdb/namespace.py
+++ b/narwhals/_duckdb/namespace.py
@@ -27,7 +27,7 @@ from narwhals._expression_parsing import (
 )
 from narwhals._sql.namespace import SQLNamespace
 from narwhals._utils import Implementation
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
@@ -116,9 +116,7 @@ class DuckDBNamespace(
 
         if how == "diagonal_relaxed":
             schemas = [Schema(df.collect_schema()) for df in items]
-            out_schema = reduce(
-                lambda x, y: to_supertype(*combine_schemas(x, y)), schemas
-            )
+            out_schema = reduce(to_supertype_diagonal, schemas)
             res, *others = (
                 item.select(
                     *(
diff --git a/narwhals/_pandas_like/namespace.py b/narwhals/_pandas_like/namespace.py
index 293589471..d43e42d5a 100644
--- a/narwhals/_pandas_like/namespace.py
+++ b/narwhals/_pandas_like/namespace.py
@@ -24,7 +24,7 @@ from narwhals._pandas_like.utils import (
     promote_dtype_backend,
 )
 from narwhals._utils import zip_strict
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
@@ -270,8 +270,7 @@ class PandasLikeNamespace(
         dtypes = tuple(native_schema(df) for df in dfs)
         dtype_backend = promote_dtype_backend(dfs, self._implementation)
         out_schema = reduce(
-            lambda x, y: to_supertype(*combine_schemas(x, y)),
-            (Schema.from_pandas_like(dtype) for dtype in dtypes),
+            to_supertype_diagonal, (Schema.from_pandas_like(dtype) for dtype in dtypes)
         ).to_pandas(dtype_backend=dtype_backend.values())
 
         native_res = (
diff --git a/narwhals/_spark_like/namespace.py b/narwhals/_spark_like/namespace.py
index 275f82589..e67a99e29 100644
--- a/narwhals/_spark_like/namespace.py
+++ b/narwhals/_spark_like/namespace.py
@@ -19,7 +19,7 @@ from narwhals._spark_like.utils import (
     true_divide,
 )
 from narwhals._sql.namespace import SQLNamespace
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
@@ -195,9 +195,7 @@ class SparkLikeNamespace(
 
         if how == "diagonal_relaxed":
             schemas = tuple(Schema(item.collect_schema()) for item in items)
-            out_schema = reduce(
-                lambda x, y: to_supertype(*combine_schemas(x, y)), schemas
-            )
+            out_schema = reduce(to_supertype_diagonal, schemas)
             native_items = (
                 item.select(
                     *(
diff --git a/narwhals/schema.py b/narwhals/schema.py
index 3f5ea8650..59a589034 100644
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -387,20 +387,21 @@ def _ensure_names_match(left: Schema, right: Schema) -> tuple[Schema, Schema]:
 
 
 def to_supertype(left: Schema, right: Schema) -> Schema:
+    """Take two schemas and try to find the supertypes between them."""
     # Adapted from polars https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-core/src/schema/mod.rs#L83-L96
     left, right = _ensure_names_match(left, right)
     it = zip(left.keys(), left.values(), right.values())
     return Schema((name, _supertype(ltype, rtype)) for (name, ltype, rtype) in it)
 
 
-def combine_schemas(left: Schema, right: Schema) -> tuple[Schema, Schema]:
-    """Extend both schemas with names and dtypes missing from the other.
+def to_supertype_diagonal(left: Schema, right: Schema) -> Schema:
+    """Align schemas for `concat(how="diagonal*")` and try to find the supertypes between them.
 
-    Returns a tuple of two schemas where each original schema is extended
-    with the columns that exist in the other schema but not in itself.
+    Both schemas are extended with the columns that exist in the other schema but not in itself.
 
-    The final order for both schemas is: left schema keys first (in order),
-    followed by keys missing from left (in the order they appear in right).
+    The final order for both schemas is:
+    - left schema keys first
+    - followed by keys missing from left (in the order they appear in right)
     """
     left_names = set(left.keys())
     missing_in_left = (kv for kv in right.items() if kv[0] not in left_names)
@@ -408,4 +409,4 @@ def combine_schemas(left: Schema, right: Schema) -> tuple[Schema, Schema]:
     extended_left = Schema((*left.items(), *missing_in_left))
     # Reorder right to match: left keys first, then right-only keys
     extended_right = Schema((kv[0], right.get(*kv)) for kv in extended_left.items())
-    return extended_left, extended_right
+    return to_supertype(extended_left, extended_right)

```

</p>
</details>

FBruzzesi: > I used the name `to_supertype_diagonal` for now, but I'm not too attached it 😄

I don't love it either 😂

What if we just call it `merge_schemas` as polars, which adds the `to_supertype` at the end of the current `combine_schemas` implementation?

dangotbanned: > What if we just call it `merge_schemas` as polars, which adds the `to_supertype` at the end of the current `combine_schemas` implementation?

I was wrong earlier in (https://github.com/narwhals-dev/narwhals/pull/3398#discussion_r2680060533).

AFAICT, [`merge_schemas`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/utils.rs#L317-L333) is only used for `"horizontal"`.


What I was *trying* to say yesterday on discord:

> (1) `Schema.to_supertype` operates on the simplest case
> 
> By this I mean, we only have to do this for `concat(how="vertical")`
> 

Was that `polars` uses `to_supertype` for **both paths**.
`"diagonal"` goes through [`convert_diagonal_concat`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L10-L68) *first* and then **both** go through the same [`convert_st_union`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L70-L115) bit

---

I'm not sure if that helps anything - just trying to correct myself 😅

tests/frame/concat_test.py
dangotbanned: @FBruzzesi let's revive #1045 as a new issue.

The example in that test **isn't** the motivating part.

---

What's here *could be* both more efficient *and* shorter

https://github.com/narwhals-dev/narwhals/blob/8fabb13e4f74be9236504b139334d6dbbb697684/narwhals/_arrow/namespace.py#L180-L194

```diff
+++ b/narwhals/_arrow/dataframe.py
+    def cast(self, schema: IntoSchema | Iterable[tuple[str, DType]]) -> Self:
+        return self._with_native(
+            self.native.cast(Schema(schema).to_arrow()), validate_column_names=False
+        )
+

+++ b/narwhals/_arrow/namespace.py
-from narwhals._arrow.utils import cast_to_comparable_string_types
+from narwhals._arrow.utils import cast_to_comparable_string_types, concat_tables
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
```

With that, we can avoid materializing as much all at once

```py
    def concat_diagonal_relaxed(self, dfs: Sequence[ArrowDataFrame], /) -> ArrowDataFrame:
        schemas = [Schema(df.schema) for df in dfs]
        supertypes = reduce(to_supertype_diagonal, schemas)
        to_concat = (
            df.cast((name, supertypes[name]) for name in schema).native
            for df, schema in zip(dfs, schemas)
        )
        return dfs[0]._with_native(concat_tables(to_concat))
```

narwhals/schema.py
dangotbanned: > Was that `polars` uses `to_supertype` for **both paths**. 
>
> `"diagonal"` goes through [`convert_diagonal_concat`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L10-L68) _first_ and then **both** go through the same [`convert_st_union`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L70-L115) bit

@FBruzzesi 

Okay I just pulled at this thread a bit locally and now have a **very rough** `narwhals` port of `convert_diagonal_concat`.

It's operating over `BaseFrame` atm, but probably belongs in `CompliantNamespace` (and using the compliant-level).

I've tested with `ibis` and I'm pretty sure we can support both `"diagonal*"` variants this way 🫰 

<p>

```diff
diff --git a/narwhals/schema.py b/narwhals/schema.py
index 59a589034..5fead8f4e 100644
--- a/narwhals/schema.py
+++ b/narwhals/schema.py
@@ -7,7 +7,7 @@ https://github.com/pola-rs/polars/blob/main/py-polars/polars/schema.py.
 from __future__ import annotations
 
 from collections import OrderedDict
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from functools import partial
 from typing import TYPE_CHECKING, cast
 
@@ -26,13 +26,15 @@ from narwhals.exceptions import ComputeError, SchemaMismatchError
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
-    from typing import Any, ClassVar
+    from typing import Any, ClassVar, TypeVar
 
     import polars as pl
     import pyarrow as pa
     from typing_extensions import Self
 
+    from narwhals.dataframe import BaseFrame
     from narwhals.dtypes import DType
+    from narwhals.expr import Expr
     from narwhals.typing import (
         DTypeBackend,
         IntoArrowSchema,
@@ -40,6 +42,8 @@ if TYPE_CHECKING:
         IntoPolarsSchema,
     )
 
+    FrameT = TypeVar("FrameT", bound="BaseFrame[Any]")
+
 
 __all__ = ["Schema"]
 
@@ -410,3 +414,44 @@ def to_supertype_diagonal(left: Schema, right: Schema) -> Schema:
     # Reorder right to match: left keys first, then right-only keys
     extended_right = Schema((kv[0], right.get(*kv)) for kv in extended_left.items())
     return to_supertype(extended_left, extended_right)
+
+
+def convert_concat_diagonal(frames: Sequence[FrameT]) -> Sequence[FrameT]:
+    """Adapted from [`convert_diagonal_concat`].
+
+    [`convert_diagonal_concat`]: https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L10-L68
+    """
+    import narwhals.functions as nw_f
+
+    schemas = [frame.collect_schema() for frame in frames]
+    it_schemas = iter(schemas)
+    total_schema = dict(next(it_schemas))
+    seen_names = set(total_schema)
+    to_add_fields: dict[str, DType] = {}
+    for sch in it_schemas:
+        to_add_fields.update(
+            {name: dtype for name, dtype in sch.items() if name not in seen_names}
+        )
+        seen_names.update(to_add_fields)
+    if not seen_names:
+        return frames
+    total_schema.update(to_add_fields)
+    total_names = tuple(total_schema)
+    added_exprs: dict[str, Expr] = {}
+    results: list[FrameT] = []
+    for frame, schema in zip(frames, schemas):
+        to_add_exprs: list[Expr] = []
+        for name, dtype in total_schema.items():
+            if name not in schema:
+                maybe_seen = added_exprs.get(name)
+                if maybe_seen is None:
+                    to_add_expr = nw_f.lit(None, dtype).alias(name)
+                    to_add_exprs.append(to_add_expr)
+                    added_exprs[name] = to_add_expr
+                else:
+                    to_add_exprs.append(maybe_seen)
+        result = frame
+        if to_add_exprs:
+            result = result.with_columns(to_add_exprs)
+        results.append(result.select(total_names))
+    return results
```

</p>
</details> 

We end up converting `"diagonal*"` -> `"vertical*"` (IIUC)


## Updated
I've got the `Compliant*` version working now over on (https://github.com/narwhals-dev/narwhals/compare/convert-concat-diagonal)

FBruzzesi: Thanks @dangotbanned - on a pragmatic side, I would say to keep this as a follow up separated from this PR. My understanding from now is that it would only benefit ibis. I would rather have correctness in relaxed and "strict" modes first, and then expand to ibis. Although `convert_concat_diagonal` might be re-used in duckdb and spark-like as well, I would argue that the current implementation is a bit easier to read and maintain 🙈

tests/frame/concat_test.py
FBruzzesi: Otherwise I am going to forget: https://github.com/narwhals-dev/narwhals/issues/3402

narwhals/schema.py
dangotbanned: > Thanks @dangotbanned - on a pragmatic side, I would say to keep this as a follow up separated from this PR. 

Absolutely!

> I would argue that the current implementation is a bit easier to read

Agreed on that part too @FBruzzesi 
I've just finished cleaning things up some and documenting in these few commits (https://github.com/narwhals-dev/narwhals/compare/f5c6407c7d75585039a2feb052f4d6a8e68e4581...convert-concat-diagonal)

---

> My understanding from now is that it would only benefit ibis.

For `how="diagonal"`? 
You might be right there, but I was thinking this replaces the need for [`combine_schemas`](https://github.com/narwhals-dev/narwhals/blob/8fabb13e4f74be9236504b139334d6dbbb697684/narwhals/schema.py#L396-L411) *entirely*.
That's used in *all* of the backends (besides `polars`).

Is `combine_schemas` shorter than `align_diagonal`?
**Definitely!** 😅

But it is also called `n_schemas - 1` times **per** `concat`.
*By my count*, each of those calls does:
- 1x `Schema.keys()`
- 3x `Schema.items()`
- 2x `Schema(...)`

I was quite alarmed by that when I noticed it 😳

---

So *in my head*, that ^^^


<details><summary>... and all of the extra stuff here ...</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/8fabb13e4f74be9236504b139334d6dbbb697684/narwhals/_arrow/namespace.py#L180-L193


</p>
</details> 

... seems like it could be avoided and we just do `align_diagonal -> _concat_vertical_relaxed`?


https://github.com/narwhals-dev/narwhals/blob/8fabb13e4f74be9236504b139334d6dbbb697684/narwhals/_arrow/namespace.py#L213-L218

> [!NOTE]
> I would rather fully disolve 🫠 than benchmark again, so I'm just asking to consider how many iterations each is doing 🙏

tests/frame/concat_test.py
```diff
@@ -138,3 +150,203 @@ def test_concat_diagonal_invalid(
...
+    expected_data = {
+        "idx": [0, 1, 2, 3, 4, 5],
+        "a": [1.0, 2.0, 3.5, 4.5, None, None],
+        "c": [10, 20, None, None, 50, 60],
+        "b": [None, None, 30.1, 40.2, 5.0, 6.0],
+    }
+    assert_equal_data(result.collect(), expected_data)
+
+
+def test_pyarrow_concat_vertical_uuid() -> None:
```
FBruzzesi, line 272: @MarcoGorelli I added 3 tests (pyarrow, duckdb and ibis) to check if the changes in https://github.com/narwhals-dev/narwhals/pull/3398/commits/4a1b946eeb3627812a0d44fd270f77837629da3c and https://github.com/narwhals-dev/narwhals/pull/3398/commits/b205ddb50c2d391efd7191bfaaf24389f8eeeb2e support what we mentioned yesterday during the community call. Namely that datatypes that are not supported by narwhals are preserved.

For polars, pandas and dask I am not applying these "workarounds" since we should support all their datatypes*

*Exception is made for poalrs Float16 and Null, but also for polars we just dispatch the method without touching the datatypes. Am I missing something for pandas and dask?

narwhals/_utils.py
```diff
@@ -2148,3 +2155,14 @@ def __repr__(self) -> str:  # pragma: no cover
 
 # Can be imported from types in Python 3.10
 NoneType = type(None)
+
+
+def safe_cast(
+    ns: CompliantNamespace[CompliantFrameT, CompliantExprT],
+    mapping: Mapping[str, dtypes.DType],
+) -> Iterable[CompliantExprT]:
+    Unknown = ns._version.dtypes.Unknown()  # noqa: N806
```
dangotbanned, line 2169: I would:
- Move this to https://github.com/narwhals-dev/narwhals/blob/4a1b946eeb3627812a0d44fd270f77837629da3c/narwhals/dtypes/_utils.py
- Use `from narwhals.dtypes._classes import Unknown` at the module-level
- Replace `==` with `dtype.base_type() is Unknown`

tests/frame/concat_test.py
dangotbanned: `pd.PeriodDtype`? 

But more generally, `pandas` supports registering extension types and so does `polars` (recently)

PR #3400, opened 2026-01-12
feat: Enable`list.{sort, sum}` for sqlframe

raisadz:
# Description

We can unxfail the sqlframe tests for `list.sum` and `list.sort` as the following issues were fixed:

- https://github.com/eakmanrq/sqlframe/issues/559
- https://github.com/eakmanrq/sqlframe/issues/567

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/narwhals/_spark_like/expr_list.py
+++ b/narwhals/_spark_like/expr_list.py
@@ -95,11 +95,10 @@ class SparkLikeExprListNamespace(
             F = self.compliant._F
             if not descending and nulls_last:
                 return F.array_sort(expr)
-            if descending and not nulls_last:  # pragma: no cover
+            if descending and not nulls_last:
                 impl = self.compliant._implementation
-                if impl.is_sqlframe():
-                    return F.array_reverse(F.array_sort(expr))
-                return F.reverse(F.array_sort(expr))
+                rev = F.array_reverse if impl.is_sqlframe() else F.reverse
+                return rev(F.array_sort(expr))
             return F.sort_array(expr, asc=not descending)
 
         return self.compliant._with_elementwise(func)


narwhals/_spark_like/expr_list.py
```diff
@@ -93,7 +96,9 @@ def func(expr: Column) -> Column:
             if not descending and nulls_last:
                 return F.array_sort(expr)
             if descending and not nulls_last:  # pragma: no cover
-                # https://github.com/eakmanrq/sqlframe/issues/559
+                impl = self.compliant._implementation
+                if impl.is_sqlframe():
+                    return F.array_reverse(F.array_sort(expr))
                 return F.reverse(F.array_sort(expr))
```
dangotbanned: I'm assuming we'd only want `# pragma: no cover` if `sqlframe` support wasn't there?

```suggestion
            if descending and not nulls_last:
                impl = self.compliant._implementation
                rev = F.array_reverse if impl.is_sqlframe() else F.reverse
                return rev(F.array_sort(expr))
```

narwhals/_spark_like/expr_list.py
```diff
@@ -52,7 +52,10 @@ def func(expr: Column) -> Column:
     def sum(self) -> SparkLikeExpr:
         def func(expr: Column) -> Column:
             F = self.compliant._F
-            return F.aggregate(F.array_compact(expr), F.lit(0.0), operator.add)
+            size = F.array_size(F.array_compact(expr))
+            return F.when((size.isNotNull()) & (size == 0), F.lit(0)).otherwise(
+                F.aggregate(F.array_compact(expr), F.lit(0.0), operator.add)
+            )
```
dangotbanned: (not blocking at all!)

I've found it a lot easier to work with all these APIs, when I don't need to remember what the names mean.
The first thing I did was searched for [`array_compact`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.array_compact.html), and had (quite wrongly!) guessed it was equivalent to `Expr.implode`.
What a weird need for `list.drop_nulls` 😄

So yeah, for the smooth-brained *like me* aliasing inline means I can read and understand each part quickly

```suggestion
            drop_nulls = F.array_compact(expr)
            len = F.array_size(drop_nulls)
            sum = F.aggregate(drop_nulls, F.lit(0.0), operator.add)
            return F.when((len.isNotNull()) & (len == 0), F.lit(0)).otherwise(sum)
```

dangotbanned (approved):
Thanks @raisadz for following this up!

My only other note was the PR title *"test: unxfail sqlframe for `list.sum` and `list.sort`"* is under-selling that you've added `sqlframe` support 😅 

banarwhal

narwhals/_spark_like/expr_list.py
raisadz: Thanks @dangotbanned ! Yes you are right, no cover was there because of sqlframe

PR #3403, opened 2026-01-14
Remove interchange from non v1

MarcoGorelli:
This is something @alexander.beedie [commented about on discord](https://discord.com/channels/1235257048170762310/1235257049626181656/1451687200814137587)

> Pretty much got the plugin working now (lazy only, as what I wrote at work doesn’t have an eager mode). Seems there’s a narwhals check for __dataframe__ that takes precedence over plugin support? I’ve worked around it for now while testing, and will probably just drop it, but just wondering if that’s expected - feels like the presence of an explicit plugin should supersede __dataframe__ 🤔 

Agree - the interchange protocol was a V1-only thing and is no longer supported in any way. Outside of V1 we shouldn't even be checking for it

### DIFF
--- a/tests/frame/interchange_select_test.py
+++ b/tests/frame/interchange_select_test.py
@@ -48,7 +48,7 @@ def test_interchange() -> None:
 def test_interchange_non_v1() -> None:
     df = CustomDataFrame({"a": [1, 2, 3], "b": [4, 5, 6], "z": [1, 4, 2]})
     with pytest.raises(TypeError, match="Unsupported dataframe type"):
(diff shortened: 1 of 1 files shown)


dangotbanned:
There were some other things that are still *lingering* from `v1`

- #2932
- https://github.com/narwhals-dev/narwhals/pull/2814#discussion_r2234109699

I feel like now is as good a time as any to finish that part 🙏

MarcoGorelli:
thanks both! 

gonna sneak this in as there's a release due, but will address the points in due course

PR #3404, opened 2026-01-15
feat: Support `concat(..., how="diagonal")` for `ibis`

dangotbanned:
# Description

While reviewing #3398, one option I raised would be to adopt how [`polars` handles `how="diagonal"`](https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/crates/polars-plan/src/plans/conversion/dsl_to_ir/concat.rs#L10-L68).

Whether we do/don't can be decided on later, but by adding support for `ibis` *now* ... 

> [!IMPORTANT]
> All backends support `concat={"diagonal", "vertical"}`

## Related issues

- Spin-off from https://github.com/narwhals-dev/narwhals/pull/3398#discussion_r2683760306
- https://github.com/rwhitten577/narwhals/blob/dd956a61b26f3d7afeef5dca59983a5e03fe1bc1/narwhals/_ibis/namespace.py#L53-L58

### DIFF
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -131,7 +131,8 @@ def test_concat_diagonal_invalid(
     context: Any
     if impl.is_polars() and POLARS_VERSION < (1,):  # pragma: no cover
         context = pytest.raises(
-            NarwhalsError, match=re.compile(r"int.+datetime", re.IGNORECASE)
+            NarwhalsError,
+            match=re.compile(r"(int.+datetime)|(datetime.+int)", re.IGNORECASE),
         )
     else:
         context = pytest.raises((InvalidOperationError, TypeError), match=r"same schema")


narwhals/_compliant/namespace.py
```diff
@@ -98,6 +100,61 @@ def is_native(self, obj: Any, /) -> TypeIs[Any]:
...
+class AlignDiagonal(Protocol[CompliantFrameT, CompliantExprT_co]):
+    """Mixin to help support `"diagonal*"` concatenation."""
+
+    def lit(
+        self, value: NonNestedLiteral, dtype: IntoDType | None
+    ) -> CompliantExprT_co: ...
+    def align_diagonal(
+        self, frames: Collection[CompliantFrameT], /
+    ) -> Sequence[CompliantFrameT]:
+        """Convert the inputs to `concat(..., how="diagonal")` into `concat(..., how="vertical")`.
```
dangotbanned: ## Why a new `Protocol`?
I first defined it as `CompliantNamespace._align_diagonal` but that had the effect of **requiring** it for duck-typing (https://github.com/narwhals-dev/narwhals/pull/3404/commits/f5c6407c7d75585039a2feb052f4d6a8e68e4581).

What's quite nice is that **only** `CompliantNamespace.lit` is needed, with the rest just operating over the inputs and result of `lit`.

If we don't end up using this in #3398, we can always just move it to `ibis` - but *for now* this is implementation should work for all backends 🙂

narwhals/_ibis/namespace.py
```diff
@@ -63,14 +67,22 @@ def _coalesce(self, *exprs: ir.Value) -> ir.Value:
...
+            natives = (lf.native for lf in frames)
+            try:
+                result = ibis.union(*natives)
+            except ibis.IbisError:
+                first = frames[0].schema
+                if not all(x.schema == first for x in frames):
+                    msg = "inputs should all have the same schema"
+                    raise TypeError(msg) from None
+                raise
+            return frames[0]._with_native(result)
```
dangotbanned: I feel like I can hear the questions about this already 😄 

per (https://github.com/narwhals-dev/narwhals/pull/3404/commits/9db77b2a0217b62c8cdfec7328c666d0aaba04dc)

> after aligning, all frames are guaranteed to have the same column names in the same order
> checking the dtypes *too*, when `ibis` will handle this already seems excessive

See https://github.com/ibis-project/ibis/blob/511d686cdb2f57aa8fc78ab6880d089c0a908db5/ibis/expr/operations/relations.py#L331-L353

tests/frame/concat_test.py
```diff
@@ -83,3 +82,32 @@ def test_concat_diagonal(
...
+    }
+    df_1 = nw.from_native(constructor(data_1)).lazy()
+    bad_schema = nw.from_native(constructor(data_2)).lazy()
+    impl = df_1.implementation
+    request.applymarker(
+        pytest.mark.xfail(
+            impl not in {Implementation.IBIS, Implementation.POLARS},
+            reason=f"{impl!r} does not validate schemas for `concat(how='diagonal')",
+        )
+    )
```
dangotbanned, line 130: I was pretty surprised we don't validate `"diagonal"` in the same way that we do for `"vertical"` (anywhere else)

FWIW, all backends *besides* `pandas` and `dask` fail, but the exceptions are not raised by us.

I'm assuming `pandas` behavior either results in `object` or downcasts `datetime` to a physical representation

narwhals/_compliant/namespace.py
```diff
@@ -98,6 +100,61 @@ def is_native(self, obj: Any, /) -> TypeIs[Any]:
...
+                    else:
+                        null_expr = self.lit(None, dtype).alias(name)
+                        missing_from_frame.append(null_expr)
+                        null_exprs[name] = null_expr
+            result = frame
+            if missing_from_frame:
+                result = result.with_columns(*missing_from_frame)
+                missing_from_frame.clear()
+            aligned.append(result.simple_select(*union_names))
+        return aligned
```
FBruzzesi: @dangotbanned we might exploit `dict_keys` objects (supporting `__sub__`) and the `dict.setdefault` trick _you_ taught me, to simplify this a bit:

```suggestion
        union_names = union_schema.keys()
        # Lazily populate null expressions as needed, shared across frames
        null_exprs: dict[str, CompliantExprT_co] = {}
        aligned = deque[CompliantFrameT]()
        for frame, schema in zip(frames, schemas):
            if (missing_names := union_names - schema.keys()):
                missing_exprs = [
                    null_exprs.setdefault(name, self.lit(None, union_schema[name]).alias(name))
                    for name in missing_names
                ]
                frame = frame.with_columns(*missing_exprs)
            aligned.append(frame.simple_select(*union_names))
        return aligned
```

tests/frame/concat_test.py
FBruzzesi: We are very lax in allowing each backend to do whatever they do natively. Once the relaxed versions are merged, then we should make sure that `how={"diagonal", "vertical"}` respect the dtypes of common/all columns (respectively)

narwhals/_compliant/namespace.py
dangotbanned: > @dangotbanned we might exploit `dict_keys` objects (supporting `__sub__`) and the `dict.setdefault` trick _you_ taught me, to simplify this a bit:

@FBruzzesi 
aha, nice uno reverse card there! 😂 

I will try this out locally, but IIRC, the reason I did it the *current* way was:
- I wasn't sure if using set ops here would guarantee order
- Using `.setdefault(name, self.lit(...` will evaluate the default, even if it goes unused
  - So we might have the expression *already*, but 
    - we always create a new one
    - which looks it up in `union_schema`
    - and also calls `.alias`

I can't tell *for sure* from the suggestion if this is *new work* - but I'll investigate!

FBruzzesi: > I wasn't sure if using set ops here would guarantee order

Does this really matter? We end up sorting the columns at the end with the `simple_select(*union_names)` call anyway

> Using .setdefault(name, self.lit(... will evaluate the default, even if it goes unused

That's... surprising, but I think I would still prefer having something like the following that explicitly loops only on missing columns:

```py
union_names = union_schema.keys()
# Lazily populate null expressions as needed, shared across frames
null_exprs: dict[str, CompliantExprT_co] = {}
aligned = deque[CompliantFrameT]()
for frame, schema in zip(frames, schemas):
    if (missing_names := union_names - schema.keys()):
        missing_exprs = []
        for name in missing_names:
            if (null_expr := null_exprs.get(name)) is None:
                null_expr = null_exprs[name] = self.lit(None, union_schema[name]).alias(name)
            missing_exprs.append(null_expr)
        frame = frame.with_columns(*missing_exprs)
    aligned.append(frame.simple_select(*union_names))
return aligned
```

dangotbanned: > That's... surprising, but I think I would still prefer having something like the following that explicitly loops only on missing columns:

I do think there's something here, I just wanna experiment a bit (and add more tests) ❤️

tests/frame/concat_test.py
```diff
@@ -61,11 +64,7 @@ def test_concat_vertical(constructor: Constructor) -> None:
         nw.concat([df_left, df_left.select("d")], how="vertical").collect()
 
 
-def test_concat_diagonal(
-    constructor: Constructor, request: pytest.FixtureRequest
-) -> None:
-    if "ibis" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
+def test_concat_diagonal(constructor: Constructor) -> None:
```
dangotbanned, line 70: (https://github.com/narwhals-dev/narwhals/pull/3404#discussion_r2694442810)

> I do think there's something here, I just wanna experiment a bit (and add more tests) ❤️

@FBruzzesi okay this led to discovering a bug (*but not in any of the new code or your suggestion*)

`ibis` is the only backend that *doesn't* guarantee the order of `union`.

I found that out by using the 3-tabled version of the test:
- (https://github.com/pola-rs/polars/blob/c2412600210a21143835c9dfcb0a9182f462b619/py-polars/tests/unit/functions/test_concat.py#L249-L260)

That fails for `ibis`, but it turns out `"vertical"` fails too:

<details><summary><b>Show test</b></summary>
<p>

```py
def test_concat_vertical_bigger(constructor: Constructor) -> None:
    data_1 = {"a": [1, 2], "b": [3, 4], "c": [0, None]}
    data_2 = {"a": [5, 6], "b": [0, None], "c": [7, 8]}
    data_3 = {"a": [0, None], "b": [9, 10], "c": [11, 12]}
    expected = {
        "a": [1, 2, 5, 6, 0, None],
        "b": [3, 4, 0, None, 9, 10],
        "c": [0, None, 7, 8, 11, 12],
    }
    df_1 = nw.from_native(constructor(data_1)).lazy()
    df_2 = nw.from_native(constructor(data_2)).lazy()
    df_3 = nw.from_native(constructor(data_3)).lazy()
    result = nw.concat([df_1, df_2, df_3], how="vertical")
```

</p>
</details> 


<details><summary><b>Show error</b></summary>
<p>

```py
E               AssertionError: Mismatch at index 0, key a: 0 != 1
E               Expected: {'a': [1, 2, 5, 6, 0, None], 'b': [3, 4, 0, None, 9, 10], 'c': [0, None, 7, 8, 11, 12]}
E               Got: {'a': [0, None, 5, 6, 1, 2], 'b': [9, 10, 0, None, 3, 4], 'c': [11, 12, 7, 8, 0, None]}
```

</p>
</details> 

~~I suppose I'm stuck with testing two tables for now then 😂~~
I've added (https://github.com/narwhals-dev/narwhals/pull/3404/commits/a8e8388ef0e6600fce9bad6d7d4fb71dfa98072a), but should probably follow this up with *another* issue.
(We (and `polars`) don't document that it is ordered, but we do test for it and [`polars.union`](https://docs.pola.rs/api/python/stable/reference/api/polars.union.html#polars.union) was recently introduced for *unordered*)

FBruzzesi: > ibis is the only backend that doesn't guarantee the order of union.

For a moment I thought order of columns, and I panicked sooo much 🤯

No guarantee in row order makes sense. You can add an index column and sort by such

narwhals/_compliant/namespace.py
dangotbanned: ### Notes
Can confirm (https://github.com/narwhals-dev/narwhals/pull/3404#discussion_r2694414151) works 🎉 

> I wasn't sure if using set ops here would guarantee order

It seems like using [views](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects) in this way is all good

---
 
I ran into ([`PLW2901`](https://docs.astral.sh/ruff/rules/redefined-loop-name/)) *as well* originally, which is why I had this extra assignment:

<details><summary><b>Show screenshot</b></summary>
<p>

image

</p>
</details> 

---

### Next
I'm assuming this switch wasn't intentional?:
- `missing_from_frame: deque[CompliantExprT_co]`
  - which is shared, grows, shrinks
- `missing_exprs: list`
  - lots of instances created

I chose [`deque`](https://docs.python.org/3/library/collections.html#collections.deque) quite delibrately here 😅 

**However**, what would be nicer than either is a generator function - because we're unpacking it anyway in the `frane.with_columns`.

I'm gonna take a look at that soon - based on your version with the `KeysView` anbd see how that goes ❤️

dangotbanned: Hopefully resolved in ([perf: Get super lazy, incorporate @FBruzzesi suggestion](https://github.com/narwhals-dev/narwhals/pull/3404/commits/759a1f6dace9d507fb92ca5de4cafd10304a2b48))

This is exactly the first time I've found a use for `nonlocal` 😂

narwhals/_compliant/namespace.py
```diff
@@ -98,6 +99,58 @@ def is_native(self, obj: Any, /) -> TypeIs[Any]:
...
+class AlignDiagonal(Protocol[CompliantFrameT, CompliantExprT_co]):
+    """Mixin to help support `"diagonal*"` concatenation."""
+
+    def lit(
+        self, value: NonNestedLiteral, dtype: IntoDType | None
+    ) -> CompliantExprT_co: ...
+    def align_diagonal(
+        self, frames: Collection[CompliantFrameT], /
+    ) -> Sequence[CompliantFrameT]:
+        """Convert the inputs to `concat(..., how="diagonal")` into `concat(..., how="vertical")`.
```
dangotbanned: Still not great phrasing, but *"inputs to ... into ..."* is a mouthful

```suggestion
        """Convert the inputs for `concat(..., how="diagonal")` into the equivalent `concat(..., how="vertical")`
```

narwhals/_ibis/namespace.py
```diff
@@ -63,14 +67,22 @@ def _coalesce(self, *exprs: ir.Value) -> ir.Value:
     def concat(
         self, items: Iterable[IbisLazyFrame], *, how: ConcatMethod
     ) -> IbisLazyFrame:
+        frames: Sequence[IbisLazyFrame] = list(items)
```
FBruzzesi: I couldn't manage to comment with a suggestion that can be applied from here, so here we go:

```py
def concat(
    self, items: Iterable[IbisLazyFrame], *, how: ConcatMethod
) -> IbisLazyFrame:
    frames: Sequence[IbisLazyFrame] = tuple(items)
    if how == "diagonal":
        frames = self.align_diagonal(frames)

    natives = (lf.native for lf in frames)
    try:
        result = ibis.union(*natives)
    except ibis.IbisError:
        first = frames[0].schema
        if not all(x.schema == first for x in frames[1:]):
            msg = "inputs should all have the same schema"
            raise TypeError(msg) from None
        raise
    return self._lazyframe.from_native(result, context=self)
```

narwhals/_compliant/namespace.py
FBruzzesi: Given the surrounding context, even something as simple as

```suggestion
        """Prepare frames with differing schemas for vertical concatenation.
```

would work in my opinion

narwhals/_ibis/namespace.py
dangotbanned: @FBruzzesi Nice one, there you go! (https://github.com/narwhals-dev/narwhals/pull/3404/commits/25a698bb35a0f98490498e5dcbf495d45ba3eabc)

PR #3405, opened 2026-01-16
fix(deps): `cudf` to `cudf-cu12` in pyproject.toml

camriddell:
cudf changed its naming scheme to include the underlying CUDA version. cudf on pypi only has version 0.6.1.post1 and has not been updated since 2020-01-01.

# Description

Attempting to run `uv run pytest`, `uv` fails to resolve the dependency for `cudf` with the below error.

```
❯ uv run python
  × No solution found when resolving dependencies for split (markers: python_full_version >= '3.10'):
  ╰─▶ Because only cudf==0.6.1.post1 is available and narwhals[cudf] depends on cudf>=24.10.0, we can
      conclude that narwhals[cudf]'s requirements are unsatisfiable.
      And because your project requires narwhals[cudf], we can conclude that your project's
      requirements are unsatisfiable.
```

Turns out, cudf changed its naming scheme to include the underlying CUDA version. cudf on pypi only has version `0.6.1.post1` and has not been updated since 2020-01-01. https://pypi.org/project/cudf/

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ Repository = "https://github.com/narwhals-dev/narwhals"
 # https://github.com/narwhals-dev/narwhals/issues/817 
 pandas = ["pandas>=1.1.3"]
 modin = ["modin"]
-cudf = ["cudf>=24.10.0"]
+cudf = ["cudf-cu12>=24.10.0"]
 pyarrow = ["pyarrow>=13.0.0"]
 pyspark = ["pyspark>=3.5.0"]
 pyspark-connect = ["pyspark[connect]>=3.5.0"]


dangotbanned:
I mentioned to @FBruzzesi about seeing that the other day

<details><summary>Show screenshot</summary>
<p>

image

</p>
</details> 


I'm pretty excited if you've fixed this @camriddell 🥳

dangotbanned:
> Attempting to run `uv run pytest`, `uv` fails to resolve the dependency for `cudf` with the below error.

You have indeed solved *that* error, but sadly there are more hiding beneath the surface 😢

### Command

```cmd
uv run pytest --no-cache --isolated
```

#### Output

```cmd
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.9.*'):
  ╰─▶ Because the requested Python version (>=3.9) does not satisfy Python>=3.10 and pandas-stubs==2.3.0.250703 depends on Python>=3.10, we can conclude that pandas-stubs==2.3.0.250703 cannot be used.
      And because narwhals:dev depends on pandas-stubs==2.3.0.250703 and your project requires narwhals:dev, we can conclude that your project's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.9) includes Python versions that are not supported by your dependencies (e.g., pandas-stubs==2.3.0.250703 only supports >=3.10). Consider using a more restrictive `requires-python` value (like >=3.10).

      hint: While the active Python version is 3.13, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
```

#### Extra
I figure that if you're not seeing this too, it would be because I have a `uv.toml` that looks like this:

```toml
# narwhals/uv.toml
[pip]

group = ["local-dev", "typing-ci"]
```

I'm still dreaming for #3204 which I imagine would solve this 🙏

camriddell:
@dangotbanned I unfortunately did encounter that second error and came to the same conclusion about 3.9 support :( (which I figured was here to stay, at least for now).

However, after temporarily changing the `requires-python` version to `>=3.10` this seems to have been the last layer of the onion. That said, I agree that #3204 needs to be accepted before we can finally use `uv run ...` with ease.

PR #3410, opened 2026-01-22
refactor: Replace ad-hoc dispatch with custom `@singledispatch`

dangotbanned:
# Description

Adds a slimmed-down version of [`@functools.singledispatch`](https://github.com/python/cpython/blob/cf71e34940e2314ee7ca00961d86a7172286eeea/Lib/functools.py#L752-L1012), with the initial use replacing [`_supertyping._same_supertype`](https://github.com/narwhals-dev/narwhals/pull/3396#discussion_r2679523423) in #3396.

I can see a few other places (particularly `DType`-related) that could *later* benefit.
So, I went ahead and added more tests + docs than I would normally do for an internal tool.

## Why not `@functools.singledispatch`?
Most of the stdlib code is dedicated to two features I don't want to use.

### [mro](https://docs.python.org/3/howto/mro.html)-based dispatch
That allows you to register `ABC`s (like `Mapping`, `Iterable`, etc) or any regular class and have it's subclasses match.

Definitely pretty clever stuff, but I'd prefer to have a simpler version that needs every class to be added explicitly

- https://github.com/python/cpython/blob/cf71e34940e2314ee7ca00961d86a7172286eeea/Lib/functools.py#L756-L894
- https://github.com/python/cpython/blob/cf71e34940e2314ee7ca00961d86a7172286eeea/Lib/functools.py#L921-L926

### Registration via annotation
Forward references are resolved eagerly[^1] and only classes and unions of classes are supported.

[^1]: `ibis` has [`@lazy_singledispatch`](https://github.com/ibis-project/ibis/blob/7cc3686cb60db81ee63bafda48da2a9bde2fc855/ibis/common/dispatch.py) that appears to solve this issue

IMO, that's just a complicated way of accepting `*types: type[Any]` in the decorator - so I did that instead 😅

- https://github.com/python/cpython/blob/cf71e34940e2314ee7ca00961d86a7172286eeea/Lib/functools.py#L937-L941
- https://github.com/python/cpython/blob/cf71e34940e2314ee7ca00961d86a7172286eeea/Lib/functools.py#L959-L991

## Related issues

- Child of #3396

### DIFF
--- a/narwhals/_dispatch.py
+++ b/narwhals/_dispatch.py
@@ -71,6 +71,9 @@ def just_dispatch(
         upper_bound: When there is no registered implementation for a specific type, it must
             be a subclass of `upper_bound` to use the default implementation.
 
+    Tip:
+        `@just_dispatch` should only be used to decorate **internal functions** as we lose the docstring.
+
     [`@functools.singledispatch`]: https://docs.python.org/3/library/functools.html#functools.singledispatch
     [MRO]: https://docs.python.org/3/howto/mro.html#python-2-3-mro
     [*just*]: https://github.com/jorenham/optype/blob/e7221ed1d3d02989d5d01873323bac9f88459f26/README.md#just


tests/dispatch_test.py
```diff
@@ -0,0 +1,168 @@
...
+    bool_ = nw.Boolean()
+    f64 = nw.Float64()
+    arr_bool_4 = nw.Array(bool_, 4)
+
+    def dtype_name(dtype: DType) -> str:
+        return type(dtype).__name__
+
+    # Default implementation serves as a fallback for subclasses of `upper_bound`
+    @just_dispatch(upper_bound=DType)
+    def dtype_repr_code(dtype: DType) -> str:
```
dangotbanned: All of these comments ~~are~~ **were** an attempt at a [*literate style*](https://github.com/astral-sh/ruff/tree/main/crates/ty_test#literate-style) 🙂 

Initially I wanted to have some examples in the docstrings (in `_dispatch.py`), but:
- Anything non-trivial requires a very large docstring
- I couldn't think of anything simple that captures the strengths of the concept

narwhals/_dispatch.py
```diff
@@ -0,0 +1,131 @@
...
+        raise TypeError(msg)
+
+    # TODO @dangotbanned: Turn all these notes into useful docs
+    def register(  # noqa: D417
+        self, tp: type[Any], *tps: type[Any]
+    ) -> Callable[[Passthrough], Passthrough]:
+        """Register types to dispatch via the decorated function.
+
+        Arguments:
+            *tps: One or more **concrete** types.
```
dangotbanned: I've left out validating the types *for now*.

- `3.9`: `@functools.singledispatch` supports `type[Any]`
- `3.10`: Same support, but that's the version which introduces `types.UnionType`
- `3.11`: `@functools.singledispatch` supports `typing.Union | types.UnionType | type[Any]`


Since #3204 is not resolved, I'd rather leave this runtime check until then

dangotbanned:
@FBruzzesi 
I'm not sure if (https://github.com/narwhals-dev/narwhals/pull/3396#discussion_r2725999904) got lost, but I thought I'd give convincing you *one last try* 🙂 

> [#3410](https://github.com/narwhals-dev/narwhals/pull/3410) is a beautiful solution, but, at the same time, quite intense to maintain. 
> It requires it's dedicated test suite 


I think this was a documentation issue on my part. 
I tried out ([literate-style tests](https://github.com/narwhals-dev/narwhals/pull/3410#discussion_r2716732809)), but as you can see from ([revert: Keep only what is required](https://github.com/narwhals-dev/narwhals/pull/3410/commits/14f81fc5afb85616ca75a837c8e8190e9e1f4788)) - removing all of that left only 2 lines uncovered.

Do you still have the opinion that (https://github.com/narwhals-dev/narwhals/blob/308f389f2e68079053e9c659394cdd0d4207410d/narwhals/_dispatch.py) is intense to maintain?

---

> and it serves a single purpose in the codebase, which makes me feel a bit uneasy on **it.**

I thought replacing lots of other *non-supertyping* stuff wouldn't be a good idea in this PR, since it was based on `dtypes/supertyping`.

Here's three unrelated use-cases (https://github.com/narwhals-dev/narwhals/compare/dtypes/supertyping-dispatch...%40just_dispatch-use-cases).
Only one was what I had planned in the PR description:

> I can see a few other places (particularly `DType`-related) that could *later* benefit.


I do feel the need to stress that these are not the only places that can use it! 😅
Just didn't wanna go replacing *too much* if it is a hard no

narwhals/_dispatch.py
```diff
@@ -0,0 +1,66 @@
...
+__all__ = ["just_dispatch"]
+
+R = TypeVar("R")
+R_co = TypeVar("R_co", covariant=True)
+Impl: TypeAlias = Callable[..., R]
+Passthrough = TypeVar("Passthrough", bound=Callable[..., Any])
+
+
+class JustDispatch(Generic[R_co]):
+    """Single-dispatch wrapper produced by decorating a function with `@just_dispatch`."""
```
FBruzzesi, line 22: I would explain some of the differences with singledispatch in here - some reduced version of the PR description would do it.

For me the most relevant difference is that there is no MRO resolution: either a type is registered or it falls back to the upper bound if applicable.

narwhals/_dispatch.py
```diff
@@ -0,0 +1,66 @@
...
+    class Deferred(Protocol):
+        def __call__(self, f: Impl[R], /) -> JustDispatch[R]: ...
+
+
+__all__ = ["just_dispatch"]
+
+R = TypeVar("R")
+R_co = TypeVar("R_co", covariant=True)
+Impl: TypeAlias = Callable[..., R]
+Passthrough = TypeVar("Passthrough", bound=Callable[..., Any])
```
FBruzzesi: Not a fan of these variable names.

- `Impl` is very reasonable, but in the narwhals context we often use it (not as a type alias though). I don't have an alternative to propose yet
- `Passthrough` is often called `Wrapped`?

narwhals/_dispatch.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def dispatch(self, tp: type[Any], /) -> Impl[R_co]:
+        """Get the implementation for a given type."""
+        if f := self._registry.get(tp):
+            return f
+        if issubclass(tp, self._upper_bound):
+            f = self._registry[tp] = self._registry[self._upper_bound]
+            return f
+        msg = f"{self._registry[self._upper_bound].__name__!r} does not support {tp.__name__!r}"  # pragma: no cover
+        raise TypeError(msg)  # pragma: no cover
```
FBruzzesi: Can we have a test for this case?

narwhals/_dispatch.py
dangotbanned: Sure will give that a whirl 🙂 

**Edit**: How's this for your eggs? (https://github.com/narwhals-dev/narwhals/pull/3410/commits/18d62737a19dbdefdc47663c412bf9ef05e5cdea)

narwhals/_dispatch.py
dangotbanned: > * `Passthrough` is often called `Wrapped`?

Ah, so this is actually a lil different to [`_Wrapped`](https://github.com/python/typeshed/blob/9c00c6e556cd864c044daffb07733f3a8c38b621/stdlib/functools.pyi#L104-L110)

Peeking at `typing.py`, I see two similar types:

- [`_IdentityCallable`](https://github.com/python/cpython/blob/837166f0999878ea83351f40726ef7a16cc692ba/Lib/typing.py#L3677-L3679) (from [`@dataclass_transform`](https://github.com/python/cpython/blob/837166f0999878ea83351f40726ef7a16cc692ba/Lib/typing.py#L3682-L3763))
- [`_Func`](https://github.com/python/cpython/blob/837166f0999878ea83351f40726ef7a16cc692ba/Lib/typing.py#L3766) (from [`@override`](https://github.com/python/cpython/blob/837166f0999878ea83351f40726ef7a16cc692ba/Lib/typing.py#L3769-L3801))

What I was trying to convey was the original function is **passed-through unchanged**.
Whereas [`@functools.wraps`](https://docs.python.org/3/library/functools.html#functools.update_wrapper) does some higher-order runtime magic to: 
> **Update** a wrapper function to **look like** the wrapped function. 


---

@FBruzzesi would something closer to `_IdentityCallable`, `_Func` (or [`_Fn`](https://github.com/narwhals-dev/narwhals/blob/ce3e730df9b5d653219924d6664a04413e02199a/narwhals/_utils.py#L1694-L1701)) work better for you?

dangotbanned: > * `Impl` is very reasonable, but in the narwhals context we often use it (not as a type alias though). I don't have an alternative to propose yet

I'm all ears for suggestions 😄 

It is annoying we already have an *implementation* concept, since that's what [`@functools.singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) uses too:

> To add overloaded **implementations** to the function, use the `register()` attribute of the generic function, 


#### Edit
I can also just remove `Impl`?

The module ends up being 2 lines shorter without it 😂 

<p>

```diff
diff --git a/narwhals/_dispatch.py b/narwhals/_dispatch.py
index 2514a5ad5..842f22899 100644
--- a/narwhals/_dispatch.py
+++ b/narwhals/_dispatch.py
@@ -5,17 +5,15 @@ from functools import partial
 from typing import TYPE_CHECKING, Any, Generic, Protocol, TypeVar, overload
 
 if TYPE_CHECKING:
-    from typing_extensions import TypeAlias
 
     class Deferred(Protocol):
-        def __call__(self, f: Impl[R], /) -> JustDispatch[R]: ...
+        def __call__(self, f: Callable[..., R], /) -> JustDispatch[R]: ...
 
 
 __all__ = ["just_dispatch"]
 
 R = TypeVar("R")
 R_co = TypeVar("R_co", covariant=True)
-Impl: TypeAlias = Callable[..., R]
 Passthrough = TypeVar("Passthrough", bound=Callable[..., Any])
 
 
@@ -24,11 +22,11 @@ class JustDispatch(Generic[R_co]):
 
     __slots__ = ("_registry", "_upper_bound")
 
-    def __init__(self, function: Impl[R_co], /, upper_bound: type[Any]) -> None:
+    def __init__(self, function: Callable[..., R_co], /, upper_bound: type[Any]) -> None:
         self._upper_bound: type[Any] = upper_bound
-        self._registry: dict[type[Any], Impl[R_co]] = {upper_bound: function}
+        self._registry: dict[type[Any], Callable[..., R_co]] = {upper_bound: function}
 
-    def dispatch(self, tp: type[Any], /) -> Impl[R_co]:
+    def dispatch(self, tp: type[Any], /) -> Callable[..., R_co]:
         """Get the implementation for a given type."""
         if f := self._registry.get(tp):
             return f
@@ -55,11 +53,11 @@ class JustDispatch(Generic[R_co]):
 
 
 @overload
-def just_dispatch(function: Impl[R], /) -> JustDispatch[R]: ...
+def just_dispatch(function: Callable[..., R], /) -> JustDispatch[R]: ...
 @overload
 def just_dispatch(*, upper_bound: type[Any] = object) -> Deferred: ...
 def just_dispatch(
-    function: Impl[R] | None = None, /, *, upper_bound: type[Any] = object
+    function: Callable[..., R] | None = None, /, *, upper_bound: type[Any] = object
 ) -> JustDispatch[R] | Deferred:
     """Transform a function into a single-dispatch generic function.
 

```

</p>
</details>

FBruzzesi: Thanks for bearing with me @dangotbanned 

> would something closer to _IdentityCallable, _Func (or [_Fn](https://github.com/narwhals-dev/narwhals/blob/ce3e730df9b5d653219924d6664a04413e02199a/narwhals/_utils.py#L1694-L1701)) work better for you?

All these options would work for me. `_IdentityCallable` is likely the most explicit. Even `PassthroughFn` could be an option?

> I can also just remove Impl?

Sure, if you don't feel particularly strong about it, it's ok to just remove it.

tests/dispatch_test.py
```diff
@@ -0,0 +1,106 @@
...
+    assert dispatch_no_bound(i64) == "int64"
+
+    @dispatch_no_bound.register(*SignedIntegerType.__subclasses__())
+    def repr_int(dtype: SignedIntegerType) -> str:
+        return f"i{type_name(dtype).strip(string.ascii_letters)}"
+
+    assert dispatch_no_bound(i64) == "i64"
+    assert repr_int(i64) == "i64"
+    assert dispatch_no_bound(nw.UInt8()) == "uint8"
+    assert dispatch_no_bound(stdlib_decimal) == "decimal"
```
FBruzzesi, line 54: Just to double check with you: there is no type checker error because regardless of the input argument annotation (`DType`), after being decorated, all we care about is the return type in the annotation (i.e. `JustDispatch[str]`).

tests/dispatch_test.py
```diff
@@ -0,0 +1,106 @@
...
+    assert dispatch_upper_bound(nw.Decimal()) == "decimal"
+
+    with pytest.raises(TypeError, match=r"'dtype_repr_code' does not support 'Decimal'"):
+        dispatch_upper_bound(stdlib_decimal)
+
+    dispatch_upper_bound.register(type(stdlib_decimal))(lambda _: "need to be explicit")
+    assert dispatch_upper_bound(stdlib_decimal) == "need to be explicit"
+
+
+def test_just_dispatch_no_auto_subclass(dispatch_no_bound: JustDispatch[str]) -> None:
```
FBruzzesi, line 79: Really nice test with `Datetime` distinction between main and v1 👌🏼

narwhals/_dispatch.py
FBruzzesi: Amazing, thank you!

tests/dispatch_test.py
dangotbanned: @FBruzzesi 
> Just to double check with you: there is no type checker error because regardless of the input argument annotation (`DType`), after being decorated, 

Yeah that is to be expected (*at least partially*)

So the `@just_dispatch`-decorated function gets this signature:

https://github.com/narwhals-dev/narwhals/blob/ddd5d98c8d8083fdc53bc22129a3ba58d2eb7b81/narwhals/_dispatch.py#L50-L52

I actually made it **stricter** than [`@functools.single_dispatch`], which doesn't require[^1] an argument:

```py
    def __call__(self, /, *args: Any, **kwargs: Any) -> _T: ...
```

Besides that tweak, I need to note this only applies to the *initally decorated* function. 
You can see in the [old tests how all others still work as expected](https://github.com/narwhals-dev/narwhals/blob/301f537aa84482c5735da754f3564e9b6bc6a8b5/tests/dispatch_test.py#L50-L67)

### Why?
I had initially planned to use `upper_bound` to derive the type, but removed it in (https://github.com/narwhals-dev/narwhals/pull/3410/commits/02b7811d1993db0e7bf428a644e32d74fa57384f).

*Partially* that was to get mypy off my back 😅 - however the more I thought about it - the only accurate type **is `object`**.

Here's a contrived example, but it shows that you can register **any type**:

<details><summary><b>Show code </b></summary>
<p>

```py
import polars as pl
from polars.datatypes import classes as pl_dtypes

import narwhals as nw
from narwhals._dispatch import just_dispatch
from narwhals._utils import qualified_type_name
from narwhals.dtypes._utils import validate_into_dtype
from narwhals.typing import IntoDType  # noqa: TC001


@just_dispatch
def into_dtype(obj: IntoDType) -> DType:
    if isinstance(obj, DType):
        return obj
    validate_into_dtype(obj)
    return obj()


@into_dtype.register(pl_dtypes.DataTypeClass)
def from_polars_meta(obj: type[pl.DataType] | pl_dtypes.DataTypeClass) -> DType:
    return into_dtype(getattr(nw, obj.__name__))


@into_dtype.register(
    *pl_dtypes.SignedIntegerType.__subclasses__(),
    *pl_dtypes.UnsignedIntegerType.__subclasses__(),
    pl.Float32,
    pl.Float64,
    pl.String,
    pl.Boolean,
    pl.Binary,
    pl.Date,
    pl.Time,
    pl.Categorical,
    pl.Unknown,
    pl.Object,
)
def from_polars(obj: pl_dtypes.DataType) -> DType:
    return from_polars_meta(obj.base_type())


ok_1 = into_dtype(nw.Date)
ok_2 = into_dtype(nw.Date())
ok_3 = into_dtype(pl.Date)
ok_4 = into_dtype(pl.Date())


print("\n".join(qualified_type_name(dtype) for dtype in [ok_1, ok_2, ok_3, ok_4]))

```

```
narwhals.dtypes._classes.Date
narwhals.dtypes._classes.Date
narwhals.dtypes._classes.Date
narwhals.dtypes._classes.Date
```

</p>
</details>

I didn't declare the function supported some types, but made it possible to do so later 😄 



[^1]: It does [runtime validation](https://github.com/python/cpython/blob/175ab31377d9e616efb95168099d8c2c9036504a/Lib/functools.py#L999-L1003) to enforce `>=1` argument, but that isn't helpful for typing 😢 

[`@functools.single_dispatch`]: https://github.com/python/typeshed/blob/a63dde194e5730932edc46cda038cbf5e4ccfa81/stdlib/functools.pyi#L213
[runtime validation]: https://github.com/python/cpython/blob/175ab31377d9e616efb95168099d8c2c9036504a/Lib/functools.py#L999-L1003

PR #3411, opened 2026-01-22
docs: Remove repeat inline supertype docs

dangotbanned:
# Description

I added the inline docs before we had the user guide.

It'll be less effort to maintain if we keep everything all in one place 🙂

## Related issues

- Child of #3396


dangotbanned:
> Thanks @dangotbanned I forgot to trim the rules 😇
> 
>  

😂

PR #3413, opened 2026-01-22
refactor: Simplify `pd.ArrowDtype` -> `nw.DType`

dangotbanned:
# Description
We can skip a lot of string manipulation by just using [`pandas.ArrowDtype.pyarrow_dtype`](https://pandas.pydata.org/pandas-docs/version/1.5/reference/api/pandas.ArrowDtype.html)

Pretty sure there's more of these to be found.
Was half expecting this to break in CI *somehow*, since it seemed too obvious

## Related issues
- Spotted while reviewing #3377

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -189,6 +189,20 @@ def rename(
 
 
 @functools.lru_cache(maxsize=16)
+def is_dtype_non_pyarrow_string(native_dtype: Any) -> bool:
+    """*There is no problem which can't be solved by adding an extra string type* pandas."""
+    # TODO @dangotbanned: Investigate how we could handle `cudf` without `str(native_dtype)`
+    # https://github.com/rapidsai/cudf/blob/a32b8cf62c9b086b645b0825b78b99f065b1887f/python/cudf/cudf/utils/dtypes.py#L646-L670
+    return isinstance(native_dtype, pd.StringDtype) or str(native_dtype) in {
+        "string",
+        "string[python]",
+        "string[pyarrow_numpy]",
+        "<StringDtype(na_value=nan)>",  # why? why? why?
+        "str",
+    }
+
+
+@functools.lru_cache(maxsize=16)
 def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> DType:  # noqa: C901, PLR0912
     dtype = str(native_dtype)
 
@@ -213,14 +227,7 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) -> 
         return dtypes.Float64()
     if dtype in {"float32", "Float32"}:
         return dtypes.Float32()
-    if dtype in {
-        # "there is no problem which can't be solved by adding an extra string type" pandas
-        "string",
-        "string[python]",
-        "string[pyarrow_numpy]",
-        "<StringDtype(na_value=nan)>",  # why? why? why?
-        "str",
-    }:
+    if is_dtype_non_pyarrow_string(native_dtype):
         return dtypes.String()
     if dtype in {"bool", "boolean"}:
         return dtypes.Boolean()


dangotbanned:
@FBruzzesi AFAICT this is all good so far in (https://github.com/narwhals-dev/narwhals/pull/3413/commits/89efe945f92b03819232df68ee4f5accd891ea7e) 

I'm gonna start pulling at the rest and see how much can be removed before collapsing 😂

narwhals/_pandas_like/utils.py
```diff
@@ -216,37 +194,25 @@ def non_object_native_to_narwhals_dtype(native_dtype: Any, version: Version) ->
...
-        "float32",
-        "Float32",
-        "Float32[pyarrow]",
-        "float32[pyarrow]",
-        "float[pyarrow]",
-    }:
+    if dtype in {"float32", "Float32"}:
         return dtypes.Float32()
     if dtype in {
         # "there is no problem which can't be solved by adding an extra string type" pandas
```
dangotbanned: I'm staying far away from this nuke

FBruzzesi: That's fair! I tested it out a bit and `"large_string[pyarrow]"` and `"string[pyarrow]"` will hit the other path (as `is_dtype_pyarrow(native_dtype)` is True).

This one `"string[pyarrow_numpy]"` I have no idea what the hell even is 😂

narwhals/_pandas_like/utils.py
```diff
@@ -355,21 +298,18 @@ def native_to_narwhals_dtype(
...
-            native_dtype, version, Implementation.CUDF
-        )
+            pa_dtype: pa.DataType = native_dtype.to_arrow()  # pyright: ignore[reportAttributeAccessIssue]
+        else:
+            pa_dtype = native_dtype.pyarrow_dtype
+        return arrow_native_to_narwhals_dtype(pa_dtype, version)
+    if str_dtype == "category":
+        return native_categorical_to_narwhals_dtype(native_dtype, version, implementation)
     if str_dtype != "object":
         return non_object_native_to_narwhals_dtype(native_dtype, version)
```
FBruzzesi, line 320: Should this function receive the `str_dtype` in the first place?

narwhals/_pandas_like/utils.py
dangotbanned: > This one `"string[pyarrow_numpy]"` I have no idea what the hell even is 😂

@FBruzzesi This is absolutely insane 😭 

```py
import pandas as pd

pd.StringDtype("pyarrow_numpy")
```
```
FutureWarning: The 'pyarrow_numpy' storage option name is deprecated and will be removed in pandas 3.0. Use 'pd.StringDtype(storage="pyarrow", na_value-np.nan)' to construct the same dtype.
Or enable the 'pd.options.future.infer_string = True' option globally and use the "str" alias as a shorthand notation to specify a dtype (instead of "string[pyarrow_numpy]").
  pd.StringDtype("pyarrow_numpy")
```
```
<StringDtype(na_value=nan)>
```


So they've deprecated it **and** changed the repr, which means we can't see this type:

```py
pd.StringDtype("pyarrow", np.nan)
```
```
<StringDtype(na_value=nan)>
```

And yes, the deprecation doesn't show up when you use parameters like any reasonable API would 😔

dangotbanned: See (https://github.com/narwhals-dev/narwhals/pull/3413/commits/2b55159e1b59f61e4245116409615096bc995802), (https://github.com/narwhals-dev/narwhals/pull/3413/commits/d56eed2ce1c0d8f3e8aa343ff6f0cbdb419dd379)

narwhals/_pandas_like/utils.py
dangotbanned: > Should this function receive the `str_dtype` in the first place?

I can see the motivation, since all the branches now use the strings exclusively.

But - and I was already leaning towards this before (https://github.com/narwhals-dev/narwhals/pull/3413#discussion_r2725506166) - I don't think we should rely on the repr **anywhere**.
They are not reliable, this one isn't even the same problem 😂 


https://github.com/narwhals-dev/narwhals/blob/d56eed2ce1c0d8f3e8aa343ff6f0cbdb419dd379/narwhals/_pandas_like/utils.py#L444-L459

PR #3364, opened 2025-12-21
feat(expr-ir): Impl `DataFrame.unique`

dangotbanned:
# Description
This'll be the first in a series of PRs that close gaps in non-`Expr` features.

Will be focusing on what is defined in [`DslPlan`](https://github.com/pola-rs/polars/blob/00d7f7e1c3b24a54a13f235e69584614959f8837/crates/polars-plan/src/dsl/plan.rs#L28-L177), and this one is for [`DslPlan::Distinct`](https://github.com/pola-rs/polars/blob/00d7f7e1c3b24a54a13f235e69584614959f8837/crates/polars-plan/src/dsl/plan.rs#L119-L123).

## Related issues

- Child of #2572
- https://github.com/narwhals-dev/narwhals/compare/oh-nodes...expr-ir/logical-plan

PR #3414, opened 2026-01-23
feat(expr-ir): Add `BaseFrame.unnest`

dangotbanned:
# Description
Back from a little detour w/ #3396 to **finally** conclude the *missing features for `LogicalPlan` series* (start from #3364).

This one is [`DslFunction::Unnest`](https://github.com/pola-rs/polars/blob/3e39d238b28da3e28501311c247bed8b5568b9fa/crates/polars-plan/src/plans/functions/dsl.rs#L48-L51), which we *currrently* don't have an equivalent of on `main`.
It'll come up eventually 😅 

## Related issues

- Child of #2572
- Follows  #3378

### DIFF
--- a/narwhals/_plan/arrow/dataframe.py
+++ b/narwhals/_plan/arrow/dataframe.py
@@ -247,7 +247,9 @@ class ArrowDataFrame(
             if len(names) == 1:
                 result = native.set_column(index, names[0], arrays[0])
             else:
-                result = insert_arrays(native.remove_column(index), index, names, arrays)
+                result = insert_arrays_at(
+                    native.remove_column(index), index, names, arrays
+                )
             return self._with_native(result)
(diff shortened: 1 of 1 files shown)


narwhals/_plan/arrow/dataframe.py
```diff
@@ -394,3 +422,20 @@ def with_arrays(
...
+# - The shape isn't great and doesn't fit that nicely w/ `with_array(s)`
+# TODO @dangotbanned: Fast-path for index = last
+# TODO @dangotbanned: Fast-path for index = first
+def insert_arrays(
+    table: pa.Table,
+    index: int,
+    names: Collection[str],
+    columns: Iterable[ChunkedOrArrayAny],
+    /,
+) -> pa.Table:
```
dangotbanned: Follow up these TODOs

PR #3419, opened 2026-01-24
refactor: Avoid subprocess to test TPCH queries, and fix q8

FBruzzesi:
# Description

While trying to fix the failing TPCH query I took the opportunity to refactor the module and avoid using subprocesses. I find debugging with such extremely annoying.

The diff for the fix is in `q8.py`

```diff
- from datetime import date
+ from datetime import datetime

-    date1 = date(1995, 1, 1)
+    date1 = datetime(1995, 1, 1)
-    date2 = date(1996, 12, 31)
+    date2 = datetime(1996, 12, 31)
```

---

Proof of success: https://github.com/narwhals-dev/narwhals/actions/runs/21320941203/job/61370704822?pr=3419

## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3416

PR #3401, opened 2026-01-12
feat: support window functions in filter

raisadz:
# Description

This is @MarcoGorelli 's implementation from his stream and I added a test.

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes https://github.com/narwhals-dev/narwhals/issues/2226

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [ ] Documented the changes

Issue #2226, opened 2025-03-16
api: support window functions in `filter`


MarcoGorelli:
I'd like to support
```python
df.filter(nw.col('b') == nw.col('b').max().over('a')
```

In SQL this isn't allowed because window functions get logically evaluated after filters

But, this isn't SQL, so we can allow it 👿 

One way to do this for sqlframe would be:
```python
     def filter(self: Self, predicate: SparkLikeExpr) -> Self:
-        # `[0]` is safe as the predicate's expression only returns a single column
-        condition = predicate._call(self)[0]
-        spark_df = self._native_frame.where(condition)
-        return self._from_native_frame(spark_df)
+        cols = self.columns
+        token = generate_temporary_column_name(8, self.columns)
+        df = self.with_columns(predicate.alias(token))
+        return df._from_native_frame(df._native_frame.filter(token).select(cols))
```

This is potentially more expensive than just calling `filter` directly, so we could do the following:
- if `predicate` is free from window functions, use current logic
- if `predicate` contains any window functions, use logic above

We could track whether expressions contain window functions in the Expr metadata

MarcoGorelli:
> In SQL this isn't allowed because window functions get logically evaluated after filters

Clarification: in SQL, it's allowed to use `qualify`, but that's neither part of the PySpark API, nor of DuckDB's relational one

```
In [43]: df
Out[43]:
shape: (3, 2)
┌─────┬─────┐
│ a   ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 4   │
│ 1   ┆ 5   │
│ 2   ┆ 6   │
└─────┴─────┘

In [44]: duckdb.sql("""
    ...: from df
    ...: select *
    ...: qualify b = max(b) over (partition by a)
    ...: """)
Out[44]:
┌───────┬───────┐
│   a   │   b   │
│ int64 │ int64 │
├───────┼───────┤
│     1 │     5 │
│     2 │     6 │
└───────┴───────┘
```

MarcoGorelli:
i started this in the last live stream, bit busy this week but if anyone wants to take it over it should be doable 

https://www.youtube.com/live/kHs2xBE-XdI?si=5nrOyP2ReY2hBXxj&t=4805

PR #3420, opened 2026-01-25
chore(typing): Improve `tpch` typing

dangotbanned:
# Description

Spun out from an unfinished review of #3419.

## Related issues

- #3419

### DIFF
--- a/narwhals/_pandas_like/utils.py
+++ b/narwhals/_pandas_like/utils.py
@@ -31,6 +31,7 @@ if TYPE_CHECKING:
     from collections.abc import Iterable, Iterator, Mapping
     from types import ModuleType
 
+    import pyarrow as pa
     from pandas._typing import Dtype as PandasDtype
     from pandas.core.dtypes.dtypes import BaseMaskedDtype
     from typing_extensions import TypeAlias, TypeIs
@@ -73,44 +74,12 @@ PD_DATETIME_RGX = r"""^
     \]                                            # Closing bracket for datetime64
 $"""
 PATTERN_PD_DATETIME = re.compile(PD_DATETIME_RGX, re.VERBOSE)
-PA_DATETIME_RGX = r"""^
-    timestamp\[
-        (?P<time_unit>s|ms|us|ns)                 # Match time unit: s, ms, us, or ns
-        (?:,                                      # Begin non-capturing group for optional timezone
-            \s?tz=                                # Match "tz=" prefix
-            (?P<time_zone>                        # Start named group for timezone
-                [a-zA-Z\/]*                       # Match timezone name (e.g., UTC, America/New_York)
-                (?:                               # Begin optional non-capturing group for offset
-                    [+-]\d{2}:\d{2}               # Match offset in format +HH:MM or -HH:MM
-                )?                                # End optional offset group
-            )                                     # End time_zone group
-        )?                                        # End optional timezone group
-    \]                                            # Closing bracket for timestamp
-    \[pyarrow\]                                   # Literal string "[pyarrow]"
-$"""
-PATTERN_PA_DATETIME = re.compile(PA_DATETIME_RGX, re.VERBOSE)
 PD_DURATION_RGX = r"""^
     timedelta64\[
         (?P<time_unit>s|ms|us|ns)                 # Match time unit: s, ms, us, or ns
     \]                                            # Closing bracket for timedelta64
 $"""
 PATTERN_PD_DURATION = re.compile(PD_DURATION_RGX, re.VERBOSE)
-PA_DURATION_RGX = r"""^
-    duration\[
-        (?P<time_unit>s|ms|us|ns)                 # Match time unit: s, ms, us, or ns
-    \]                                            # Closing bracket for duration
-    \[pyarrow\]                                   # Literal string "[pyarrow]"
-$"""
-PATTERN_PA_DURATION = re.compile(PA_DURATION_RGX, re.VERBOSE)
-PA_DECIMAL_RGX = r"""^
-    decimal128\(
-        (?P<precision>\d+)                        # Precision value
-        ,\s                                       # Literal string ", "
-        (?P<scale>\d+)                            # Scale value
-    \)
-    \[pyarrow\]                                   # Literal string "[pyarrow]"
-$"""
-PATTERN_PA_DECIMAL = re.compile(PA_DECIMAL_RGX, re.VERBOSE)
 
 NativeIntervalUnit: TypeAlias = Literal[
     "year",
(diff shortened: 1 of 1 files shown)


tpch/tests/conftest.py
```diff
@@ -29,17 +31,35 @@
...
+def pytest_addoption(parser: pytest.Parser) -> None:
+    from tests.conftest import DEFAULT_CONSTRUCTORS
+
+    parser.addoption(
+        "--constructors",
+        action="store",
+        default=DEFAULT_CONSTRUCTORS,
+        type=str,
+        help="<sink for defaults in VSC getting injected>",
+    )
```
dangotbanned, line 52: This is a workaround for this `conftest.py` conflicting with *my* [`python.testing.pytestArgs`](https://code.visualstudio.com/docs/python/testing#_test-discovery) setting in VSCode.

I needed this to be able to run locally, but I left it in since it gave me an idea.

**Wouldn't it be helpful if `--constructors` controlled which backends run for `tpch`?**

It could just use the intersection of those requested and `TPCHBackend`

https://github.com/narwhals-dev/narwhals/blob/f96dd4340e01bdbd90d3a6411a2ddf436efb2f51/tpch/typing_.py#L12-L14

tpch/tests/conftest.py
```diff
@@ -149,41 +157,42 @@ def skip_if_unsupported(query_id: str, backend_name: str) -> None:
...
-    native_namespace, kwargs = BACKEND_NAMESPACE_KWARGS_MAP[backend_name]
+    kwargs = BACKEND_KWARGS_MAP[backend_name]
+    backend = TPCH_TO_BACKEND_NAME[backend_name]
 
-    def _load_data(query_id: str) -> tuple[nw.LazyFrame[Any], ...]:
+    def _load_data(query_id: QueryID) -> tuple[nw.LazyFrame[Any], ...]:
         data_paths = QUERY_DATA_PATH_MAP[query_id]
         return tuple(
-            nw.scan_parquet(str(path), backend=native_namespace, **kwargs)
+            nw.scan_parquet(path.as_posix(), backend=backend, **kwargs)
```
dangotbanned, line 181: `str(path)` was causing me issues on Windows with `sqlframe`, so this is the **smallest change** I could do to fix it

```py
duckdb.duckdb.IOException: IO Error: No files found that match the pattern "C:\Users\danie\Documents\GitHub
arwhals pch\data\lineitem.parquet"
```

```py
sql = 'CREATE OR REPLACE TEMPORARY VIEW "r369bb122ae974dcab9b45fa1066d9fa0" AS WITH "t15681434" AS (SELECT * FROM READ_PARQUET([\'C:\\Users\\danie\\Documents\\GitHub\narwhals\tpch\\data\\lineitem.parquet\'])) SELECT * FROM "t15681434"'
```

---

The **correct change** would be updating the `str(Path(source))` here to `Path(source).as_posix()`

https://github.com/narwhals-dev/narwhals/blob/cad2b93e55c9e11a1d00f5dea1b4b2c9f5fd7f94/narwhals/_utils.py#L2117-L2122

tpch/tests/conftest.py
```diff
@@ -9,12 +9,14 @@
 import pytest
 
 import narwhals as nw
+from tpch.typing_ import QueryID
```
dangotbanned: Using `tpch.typing` was confusing the import system, think I wanted (stdlib) `typing`.

IIRC, that issue is related to how the `tpch` package is structured, but I stayed out of that rabbit hole for now

FBruzzesi: May I suggest `_typing.py` instead? `typing_.py` feels so odd 😂

tpch/tests/conftest.py
```diff
@@ -51,56 +71,44 @@ def _build_backend_namespace_kwargs_map() -> dict[str, tuple[ModuleType, dict[st
...
+BACKEND_KWARGS_MAP = _build_backend_kwargs_map()
 
 # Queries that need to be skipped for certain backends
-DUCKDB_SKIPS = frozenset(
-    [
-        "q15"  # needs `filter` which works with window expressions
-    ]
-)
+NEEDS_FILTER_FOR_WINDOW_EXPRESSIONS = "q15"
+DUCKDB_SKIPS = frozenset[QueryID]((NEEDS_FILTER_FOR_WINDOW_EXPRESSIONS,))
```
FBruzzesi: Can we keep the comment? Or even better we can link #2226 and #3401

dangotbanned: > Can we keep the comment? Or even better we can link #2226 and #3401

Link to the issue(s) sounds great to me 😄

If we put it in a trailing docstring, I think we can get it to show on hover here

image

tpch/tests/conftest.py
dangotbanned: Moved to #3422

tpch/tests/conftest.py
dangotbanned: > May I suggest `_typing.py` instead? `typing_.py` feels so odd 😂

I guess I've gotten the [PEP 8 convention](https://peps.python.org/pep-0008/#descriptive-naming-styles) *slightly wrong* 😅

> `single_trailing_underscore_`: used by convention to avoid conflicts with Python keyword, e.g. :
> `tkinter.Toplevel(master, class_='ClassName')`

But I usually use this for any kinda of name conflict, not *just* keywords.
Here's some external examples, because after reading PEP 8 I gaslit myself into thinking this *wasn't* a thing 😭:
- [`numpy`](https://github.com/numpy/numpy/blob/c3ca1c17ab01817402dc078a35ebf72fbe72daee/numpy/__init__.pyi#L4360)
- [`pyarrow`](https://github.com/apache/arrow/blob/7bc47c06509601e09c83177bda061c4faca84809/python/pyarrow/__init__.py#L153-L167)
- [`pandas`](https://github.com/pandas-dev/pandas/blob/54caaaa40ceb677e1f521888da5715a88768bba5/pandas/core/arrays/numpy_.py), [more `pandas`](https://github.com/pandas-dev/pandas/blob/54caaaa40ceb677e1f521888da5715a88768bba5/pandas/core/arrays/string_.py) and [oh dear, being common in `pandas` is usually a bad sign](https://github.com/pandas-dev/pandas/blob/54caaaa40ceb677e1f521888da5715a88768bba5/pandas/core/util/numba_.py)


Also a little closer to home 😉 

https://github.com/narwhals-dev/narwhals/blob/cad2b93e55c9e11a1d00f5dea1b4b2c9f5fd7f94/narwhals/__init__.py#L49

https://github.com/narwhals-dev/narwhals/blob/cad2b93e55c9e11a1d00f5dea1b4b2c9f5fd7f94/narwhals/__init__.py#L62

tpch/tests/conftest.py
dangotbanned: I've done you one better w/ (https://github.com/narwhals-dev/narwhals/pull/3420/commits/8bf7cfb64d5252a2847196c43fdfa42131a08251) and (https://github.com/narwhals-dev/narwhals/pull/3420/commits/f22141a01f6df6dca6b7c91b2951d2c9c1e3edc9)

PR #3427, opened 2026-01-29
chore(tpch): Create one folder for each `scale_factor`

FBruzzesi:
# Description

Related to https://github.com/narwhals-dev/narwhals/pull/3421#issuecomment-3813980421

This PR adds the possibility to store multiple scale factors in the same data folder.

As a consequence, it adds a parameter to pytest (`--scale-factor`, default 0.1) that is possible to pass. If the scale factor data don't exist, they will be created before running the tests.

All other capabilities are left the same.

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

https://github.com/narwhals-dev/narwhals/pull/3421#issuecomment-3813980421

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

Issue #3423, opened 2026-01-26
Support running `tpch` benchmarks at smaller scales


dangotbanned:
## Description

#3419 reminded me that we hadn't followed up on #805 yet.

We last left it with (https://github.com/narwhals-dev/narwhals/issues/805#issuecomment-2849311036), but #2516 turned out to take too long in CI

FBruzzesi:
I think one of the issue with codspeed could have been the subprocess part, as well as the fact that we could not isolate the backends. Now I am hopeful!

PR #3421, opened 2026-01-25
chore(tpch): Add a basic CLI for `tpch/generate_data.py`

dangotbanned:
# Description

A pretty modest start, but adds the ability to control the [scale factor](https://duckdb.org/docs/stable/core_extensions/tpch#data-generator-parameters) and prevents data generation happening on module import.

I've wanted to be able to control the (previously global) parameter since (https://github.com/narwhals-dev/narwhals/issues/805#issuecomment-2799042816) 😅

```console
$ uv run tpch/generate_data.py --help
Generate the data required to run TPCH queries.

Usage: generate_data.py [OPTIONS]

Options:
  -h, --help            show this help message and exit
  -sf, --scale-factor   Scale the database by this factor (default: 0.1)
                        ┌───────┬────────────┬─────────────┐
                        │ sf    ┆ Disk       ┆ Memory (db) │
                        ╞═══════╪════════════╪═════════════╡
                        │ 0.014 ┆    3.25 mb ┆    8.79 mb  │
                        │ 0.052 ┆   12.01 mb ┆   32.49 mb  │
                        │ 0.1   ┆   23.15 mb ┆   62.62 mb  │
                        │ 0.25  ┆   58.90 mb ┆  159.32 mb  │
                        │ 0.51  ┆  124.40 mb ┆  336.50 mb  │
                        │ 1.0   ┆  247.66 mb ┆  669.92 mb  │
                        │ 10.0  ┆    2.59 gb ┆    7.00 gb  │
                        │ 30.0  ┆    7.76 gb ┆   21.00 gb  │
                        └───────┴────────────┴─────────────┘
  --debug               Enable more detailed logging (default: False)
  --refresh             Re-run data generation, regardless of whether `--scale-factor` is already on disk (default: False)
```

Data generation is now very transparent, where previously the script was silent:
- All meaningful operations are timestamped
- Visualizations of the schemas at each stage can be seen with `--debug`

This is more important for generating larger datasets, but here's the smallest for reference:

<details><summary><b>Show <code>--debug</code> logs</b></summary>
<p>

```console
$ uv run tpch/generate_data.py --scale-factor=0.014 --refresh --debug 
21:44:15.813 [INFO] Refreshing data for scale_factor=0.014
21:44:15.869 [INFO] Connecting to DuckDB database: :memory:
21:44:15.883 [INFO] Installing DuckDB TPC-H Extension
21:44:15.949 [INFO] Generating data for scale_factor=0.014
[...]

PR #3429, opened 2026-01-31
chore(tpch): Restrict `scale_factor` to a set of choices

dangotbanned:
# Description

This makes the testing setup simpler, by only allowing a subset of what is known to work.
(*see (https://github.com/narwhals-dev/narwhals/pull/3421#discussion_r2743356336) for benchmarking that informed these choices*).

Previously, any `float` was permitted but required lots of special-casing to prevent test failures.

While narrowing down these options, an issue (https://github.com/narwhals-dev/narwhals/pull/3429#discussion_r2749527226) with the larger (*official*) scales came up.
These are now fixed, with the data generation utilizing `polars`' streaming engine **and** following a strategy dependent on the resulting database size:

| `scale_factor` | Database                 | Batches | Disk (GB)    | Memory (GB)  |
| -------------- | ------------------------ | ------- | ------------ | ------------ |
| <10  [^1]      | [In-memory]              | 1       | 0.003 - 0.25 | 0.009 - 0.67 |
| 10   [^2]      | [In-memory]              | 12      | 2.59         | 7.00         |
| 30             | [File-based] (temporary) | 4       | 7.76         | Negligible   |

[^1]: Everything prior to this PR worked this way, but it doesn't scale-up
[^2]: See [Resource Usage of the Data Generator]

[In-memory]: https://duckdb.org/docs/stable/clients/python/dbapi#in-memory-connection
[File-based]: https://duckdb.org/docs/stable/clients/python/dbapi#file-based-connection
[Resource Usage of the Data Generator]: https://duckdb.org/docs/stable/core_extensions/tpch#resource-usage-of-the-data-generator

## Related issues

- Child of #3421
- Follows #3427
- Towards #3423
- #3433

### DIFF
--- a/tpch/generate_data.py
+++ b/tpch/generate_data.py
@@ -115,12 +115,6 @@ class TPCHGen:
         return _scale_factor_dir(self.scale_factor)
 
     @property
-    def _batches(self) -> int:
-        if self.scale_factor in {"10.0", "30.0"}:
-            return 12 if self.scale_factor == "10.0" else 4
-        return 0
-
-    @property
     def _database(self) -> Path | Literal[":memory:"]:
         return DB_PATH if self.scale_factor == "30.0" else ":memory:"
 
@@ -181,11 +175,12 @@ class TPCHGen:
         return self
 
     def generate_database(self) -> TPCHGen:
-        logger.info("Generating data for scale_factor=%s", self.scale_factor)
-        if batches := self._batches:
-            self._generate_database_batched(batches)
+        sf = self.scale_factor
+        logger.info("Generating data for scale_factor=%s", sf)
+        if sf in {"10.0", "30.0"}:
+            self._generate_database_batched(12 if sf == "10.0" else 4)
         else:
-            self.sql(SQL_DBGEN, sf=self.scale_factor)
+            self.sql(SQL_DBGEN, sf=sf)
         logger.info("Finished generating data.")
         if logger.isEnabledFor(logging.DEBUG):
             msg = str(self.sql(SQL_SHOW_DB))[:-1]


tpch/typing_.py
```diff
@@ -39,27 +35,32 @@
     "q21",
     "q22",
 ]
-XFailRaises: TypeAlias = type[BaseException] | tuple[type[BaseException], ...]
-Artifact: TypeAlias = Literal["database", "answers"]
+ScaleFactor: TypeAlias = Literal[
+    "0.014", "0.052", "0.1", "0.25", "0.51", "1.0", "10.0", "30.0"
+]
```
dangotbanned, line 40: I'm using string literals, since float literals [aren't allowed](
https://typing.python.org/en/latest/spec/literal.html#illegal-parameters-for-literal-at-type-check-time)

The annoying part is I would actually want to have `"1", "10", "30"` work as well.
But including them would mean most of the `Literal[...]` contains values that are less likely to be used

tpch/generate_data.py
```diff
@@ -127,8 +129,7 @@ def connect(self) -> TPCHGen:
         self._con = duckdb.connect(database=":memory:")
         return self
 
-    # TODO @dangotbanned: Change to `LiteralString` after restricting `--scale-factor`
-    def sql(self, query: str) -> Rel:
+    def sql(self, query: LiteralString) -> Rel:
```
dangotbanned: First time I've managed to use `LiteralString` for the original use case 😄 

- https://peps.python.org/pep-0675/#motivation

tpch/typing_.py
```diff
@@ -39,27 +35,32 @@
...
+"""Values for `scale_factor` that are known to produce correct results.
 
+These three are blessed by [TPC-H v3.0.1 (Page 79)]:
 
-class QueryModule(Protocol):
-    def query(
-        self, *args: nw.LazyFrame[Any], **kwds: nw.LazyFrame[Any]
-    ) -> nw.LazyFrame[Any]: ...
+    "1.0", "10.0", "30.0"
```
dangotbanned, line 45: I'm gonna mark as draft while I investigate (https://duckdb.org/docs/stable/core_extensions/tpch#resource-usage-of-the-data-generator)

For anyone that isn't Francesco, I'm running into memory issues when generating `10` and `30` on windows

dangotbanned: Following (https://github.com/narwhals-dev/narwhals/pull/3429/commits/93fbf49273a76a31df775f8a5a38a731e72e702a), at least `10` is now working for me.

Just need to implement [this part](https://duckdb.org/docs/stable/core_extensions/tpch#resource-usage-of-the-data-generator), and then larger sizes should be pretty reliable:

> If you have a limited amount of memory available, you can run the dbgen function in steps. For example, you may generate SF300 in 10 steps:
> ```sql
> 
> CALL dbgen(sf = 300, children = 10, step = 0);
> CALL dbgen(sf = 300, children = 10, step = 1);
> ...
> CALL dbgen(sf = 300, children = 10, step = 9);
> ```

Without that, I was already at 80-90% memory usage *before* `polars` even had the data.
No idea how it even worked 😂

tpch/generate_data.py
```diff
@@ -11,49 +11,68 @@
...
-    from tpch.typing_ import Artifact, QueryID
+    from tpch.typing_ import Artifact, QueryID, ScaleFactor
 
 
 logger = logging.getLogger(LOGGER_NAME)
 
+
+# `mem_usage_scale = 2.705`
+# `pl.Config(tbl_hide_column_data_types=True, tbl_hide_dataframe_shape=True)`
+# https://duckdb.org/docs/stable/core_extensions/tpch#resource-usage-of-the-data-generator
```
dangotbanned:
```suggestion
```

tpch/generate_data.py
```diff
@@ -11,49 +11,68 @@
...
+SQL_DBGEN_STEPPED = "CALL dbgen(sf=$sf, children=$children, step=$step)"
 SQL_TPCH_ANSWER = "PRAGMA tpch({0})"
 SQL_FROM = "FROM {0}"
+SQL_SHOW_DB = """
+SELECT
+    "table": name,
+    "schema": MAP(column_names, column_types)
+FROM
+    (SHOW ALL TABLES)
+"""
```
dangotbanned, line 69: Should probably convert these to all use this [Named Parameters syntax](https://duckdb.org/docs/stable/clients/python/dbapi#named-parameters)

dangotbanned: Nope, can't use that everywhere - oh well

PR #3433, opened 2026-01-31
ci: Bump `duckdb==1.4.4` in `--group typing`

dangotbanned:
# Description
A more minimal version of the last attempt (#3189).

Biggest motivation is that `1.30.0` doesn't include (https://github.com/duckdb/duckdb-python/commit/62f17f4ca0966566d4fa38c9e72948a9a418b25a).
I'm hoping to resolve (https://github.com/narwhals-dev/narwhals/pull/3429#discussion_r2749527226) by being able to use:

```py
DuckDBPyRelation.pl(lazy=True)
```

## Related issues

- Supersedes #3189
- https://github.com/duckdb/duckdb-python/issues/131

### DIFF
--- a/narwhals/_duckdb/utils.py
+++ b/narwhals/_duckdb/utils.py
@@ -196,8 +196,9 @@ def native_to_narwhals_dtype(
         return dtypes.Datetime(time_zone=deferred_time_zone.time_zone)
 
     if duckdb_dtype_id == "decimal":
-        precision: Incomplete = duckdb_dtype.children[0][1]
-        scale: Incomplete = duckdb_dtype.children[1][1]
+        precision: Incomplete
+        scale: Incomplete
+        (_, precision), (_, scale) = duckdb_dtype.children
         return dtypes.Decimal(precision, scale)
 
     return _non_nested_native_to_narwhals_dtype(duckdb_dtype_id, version)


narwhals/_duckdb/utils.py
```diff
@@ -1,31 +1,44 @@
...
     from narwhals._compliant.typing import CompliantLazyFrameAny
     from narwhals._duckdb.dataframe import DuckDBLazyFrame
     from narwhals._duckdb.expr import DuckDBExpr
     from narwhals.dtypes import DType
     from narwhals.typing import IntoDType, TimeUnit
 
+Incomplete131: TypeAlias = Any
+"""Review if an progress is made towards [#131].
+
+[#131]: https://github.com/duckdb/duckdb-python/issues/131"""
```
FBruzzesi: Can we just call it `Incomplete`? This got me a laugh but when navigating the code I don't think it would help 😂

```suggestion
Incomplete: TypeAlias = Any
"""Review if a progress is made towards [#131].

[#131]: https://github.com/duckdb/duckdb-python/issues/131
"""
```

narwhals/_duckdb/utils.py
```diff
@@ -174,15 +189,16 @@ def native_to_narwhals_dtype(
...
 
     if duckdb_dtype_id == "timestamp with time zone":
         return dtypes.Datetime(time_zone=deferred_time_zone.time_zone)
 
     if duckdb_dtype_id == "decimal":
-        (_, precision), (_, scale) = duckdb_dtype.children
-        return dtypes.Decimal(precision=precision, scale=scale)
+        precision: Incomplete131 = duckdb_dtype.children[0][1]
+        scale: Incomplete131 = duckdb_dtype.children[1][1]
+        return dtypes.Decimal(precision, scale)
```
FBruzzesi, line 202: I presume there is no way to keep the original unpacking, isn't it?

It's probably just me but unpacking vs indexing, I prefer the former - it makes it much clearer to me how the structure is

dangotbanned: > I presume there is no way to keep the original unpacking, isn't it?

Both `precision` and `scale` need annotations

> It's probably just me but unpacking vs indexing, I prefer the former - it makes it much clearer to me how the structure is

It actually took me a while to figure out that this:

```py
(_, precision), (_, scale) = duckdb_dtype.children
```

Is implicitly saying:

```py
assert len(duckdb_dtype.children) == 2
assert len(duckdb_dtype.children[0]) == 2
assert len(duckdb_dtype.children[1]) == 2

precision = duckdb_dtype.children[0][1]
scale = duckdb_dtype.children[1][1]
```

I feel like extended unpacking *can* be a cool trick.
But I'm very against it when `duckdb` doesn't have the typing to fill in the gaps.

I'd be more than happy to have this when my IDE can help me 🙂

narwhals/_duckdb/utils.py
dangotbanned: Ah that's fair, my idea was that I wanted a distinct symbol for this issue.
So that I can do a *Find All References*, and only things related will show up.

A more general `Incomplete` (if used anywhere else) could muddy that.
But since we *don't* have an `Incomplete` here - I suppose we can worry about that later 😄

narwhals/_duckdb/utils.py
FBruzzesi: > Is implicitly saying:

I disagree with the "implicitly" part. I understand it _seems_ like a lot, but it also shows the full structure. On the other hand:

```py
precision = duckdb_dtype.children[0][1]
scale = duckdb_dtype.children[1][1]
```

tells me nothing about `duckdb_dtype.children` nor about its nested elements.

<details><summary>Different implications</summary>
<p>

While one implies:

```py
assert len(duckdb_dtype.children) == 2
assert len(duckdb_dtype.children[0]) == 2
assert len(duckdb_dtype.children[1]) == 2
```

for the other its much less precise statement:

```py
assert len(duckdb_dtype.children) >= 2
assert len(duckdb_dtype.children[0]) >= 2
assert len(duckdb_dtype.children[1]) >= 2
```

</p>
</details> 

That kind of unpacking is the closest you can get to structural pattern matching without the `match` keyword. 

---

> when duckdb doesn't have the typing to fill in the gaps

Yeah I feel that, but also adding "Any/Incomplete" is not giving us much more information than the original 🫠 
I might feel different if we use a `cast("int", duckdb_dtype.children[i][1])`.

---

Anyway I don't want to be a blocker on this

dangotbanned: > for the other its much less precise statement:

> Yeah I feel that, but also adding "Any/Incomplete" is not giving us much more information than the original 

I think we're looking at this from different angles.

I'm saying that because what we get from `duckdb` is:

```py
    @property
    def children(self) -> list[tuple[str, object]]: ...
```

It is unsafe to assume we have a length-2 list, because nothing communicates that is what we have.
If the stubs were like this, I'd be 100% fine with it:

```py
    @property
    def children(self) -> tuple[tuple[str, object], tuple[str, object]]: ...
```

In an ideal world, they would use actual classes instead of this nested tuple business 😭

dangotbanned: Hoping we're all good following this?

[ruffle less feathers on this duck](https://github.com/narwhals-dev/narwhals/pull/3433/commits/54de7354cda85faaef41357bb5756a62b5d72670)

FBruzzesi: Aawww 🤗 that's too kind! Thanks

PR #3435, opened 2026-02-01
feat: Allow scalars in horizontal expressions

### LINKED ISSUE #1868: enh: allow for scalars in `sum_horizontal`

FBruzzesi:
# Description

Some tiny extra gymnastic to allow this for pandas, pyarrow and dask. All other backends where solved by some refactor 

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes  #1868

## Checklist

- [x] Tests added
- [x] Documented the changes


narwhals/_arrow/namespace.py
```diff
@@ -139,13 +139,11 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
 
     def min_horizontal(self, *exprs: ArrowExpr) -> ArrowExpr:
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
-            init_series, *series = tuple(chain.from_iterable(expr(df) for expr in exprs))
-            native_series = reduce(
-                pc.min_element_wise, [s.native for s in series], init_series.native
+            series = tuple(chain.from_iterable(expr(df) for expr in exprs))
+            result = reduce(
+                lambda s1, s2: s1._with_binary(pc.min_element_wise, s2), series
```
FBruzzesi: Let `_with_binary` take care of the broadcasting for scalars

narwhals/_dask/namespace.py
```diff
@@ -172,11 +172,11 @@ def concat(
...
         def func(df: DaskLazyFrame) -> list[dx.Series]:
-            expr_results = [s for _expr in exprs for s in _expr(df)]
-            series = align_series_full_broadcast(df, *(s.fillna(0) for s in expr_results))
-            non_na = align_series_full_broadcast(
-                df, *(1 - s.isna() for s in expr_results)
+            expr_results = align_series_full_broadcast(
+                df, *[s for _expr in exprs for s in _expr(df)]
             )
+            series = (s.fillna(0) for s in expr_results)
+            non_na = (1 - s.isna() for s in expr_results)
```
FBruzzesi, line 179: First broadcast (once) so that we ensure to have series (i.e. no more `dask_expr.Scalar`'s), then it's possible to perform `fillna()` and `isna()` safely

narwhals/_pandas_like/namespace.py
```diff
@@ -235,7 +235,9 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
 
     def min_horizontal(self, *exprs: PandasLikeExpr) -> PandasLikeExpr:
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
-            series = list(chain.from_iterable(expr(df) for expr in exprs))
+            series = self._series._align_full_broadcast(
+                *chain.from_iterable(expr(df) for expr in exprs)
+            )
```
FBruzzesi, line 240: Need to broadcast scalars first

narwhals/_arrow/namespace.py
dangotbanned: `pc.min_element_wise` doesn't need broadcasting btw

FBruzzesi: Given a scalar input, by the time we get here, it's a length one array, not a scalar anymore, and we get a shape mismatch

narwhals/functions.py
```diff
@@ -1217,7 +1217,9 @@ def _expr_with_horizontal_op(name: str, *exprs: IntoExpr, **kwargs: Any) -> Expr
     )
 
 
-def sum_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
+def sum_horizontal(
+    *exprs: PythonLiteral | IntoExpr | Iterable[PythonLiteral | IntoExpr],
```
FBruzzesi, line 1270: I will comment here to exemplify the behavior. Polars definition of `IntoExpr` is already including python literal. They distinguish between:

```py
# Inputs that can convert into a `col` expression
IntoExprColumn: TypeAlias = Union["Expr", "Series", str]
# Inputs that can convert into an expression
IntoExpr: TypeAlias = PythonLiteral | IntoExprColumn | None
```

I am not sure if it's reasonable for us to eventually align with those definition and distinction

[code](https://github.com/pola-rs/polars/blob/119b8bf0c985fc5ae506b2093c7cd48ec16c7be9/py-polars/src/polars/_typing.py#L88C1-L91C60)

narwhals/_arrow/namespace.py
dangotbanned: Oh I see 🤦

I was thinking of this behavior:

```py
(Scalar, Scalar) - > Scalar
(Scalar, Array) -> Array
(Array, Array) -> Array
```

Which is what broadcasting is, but I forgot the `Scalar` preservation is only in #2572

narwhals/functions.py
dangotbanned: I like this in polars too.

I think an issue in `narwhals` *could* be the special-casing for pandas non-`str` column *"names"*.
But I'm not sure how far that support extends, e.g. do we actually support it everywhere?

narwhals/_arrow/namespace.py
FBruzzesi: https://github.com/narwhals-dev/narwhals/pull/3435/commits/0cad6c10072e96f34e720c94a345ec82202c00c4 should reduce the overhead of both "align"-ing first, and then reduce-ing

dangotbanned: @FBruzzesi 😍😍😍

narwhals/_pandas_like/namespace.py
FBruzzesi: @dangotbanned unlike the arrow case, here we need to align the series for how we implement the operation after

MarcoGorelli, line 240: i'd like to check if we need to do this align-full-broadcast, eye-balling `coalesce` again it looks like we didn't need to do it there, can't remember why right now

FBruzzesi: I think it's because we don't allow scalars in coalesce?

(Disclaimer: from mobile, I only checked signature)

PR #3437, opened 2026-02-01
RFC, feat: Add `SQLExpr.filter` method

### LINKED ISSUE #2203: enh: `Expr.filter` for lazyframes
For LazyFrames, we can support `Expr.filter` so long as it's followed by an aggregation

For spark-like, `Column.filter` doesn't exist, but we can use `F.expr` to accomplish the same:
```python
import sqlframe.duckdb.functions as F

df = DuckDBSession().createDataFrame(pd.DataFrame({'a': [1,1,2], 'b': [4,5,6]}))
df = df.select(
    F.expr('sum(b) filter (where a==1)').alias('c'),
    F.expr('sum(b) filter (where a!=1)').alias('d'),
)
df.show()
```
```
+---+---+
| c | d |
+---+---+
| 9 | 6 |
+---+---+
```

FBruzzesi:
# Description

Opening as RFC because support is not for all methods - I am specifically excluding `len`, but I might be missing some other cases.

As hinted in a comment in the test, for `len` we can get the right result only if there are no nulls, which of course we cannot guarantee. I am checking if there is another way, but for now I am explicitly raising a NotImplementedError

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2203

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/.github/workflows/check_tpch_queries.yml
+++ b/.github/workflows/check_tpch_queries.yml
@@ -5,7 +5,7 @@ on:
 
 env:
   PY_COLORS: 1
-
+  UV_SYSTEM_PYTHON: 1
 jobs:
   validate-queries:
     strategy:
@@ -26,10 +26,8 @@ jobs:
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: local-install
-        run: |
-          uv pip install -U -e ".[dask]" --group core-tests --system
-          uv pip install -U duckdb --system
+        run: uv pip install -U -e ".[dask]" --group core-tests
       - name: generate-data
-        run: cd tpch && python generate_data.py
+        run: cd tpch && python generate_data.py --debug
       - name: tpch-tests 
         run: cd tpch && pytest tests
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -8,6 +8,8 @@ env:
 
 jobs:
   altair:
+    env:
+      UV_SYSTEM_PYTHON: 1
     strategy:
       matrix:
         python-version: ["3.12"]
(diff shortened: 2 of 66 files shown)


tests/expr_and_series/filter_test.py
```diff
@@ -69,3 +69,35 @@ def test_filter_windows_over(
...
+        nw.col("a").filter(nw.col("i") < 3).mean(),
+        nw.col("b").filter(*mask_multi).sum(),
+        nw.col("b", "c").filter(*mask_multi).min().name.suffix("_min"),
+        count_c=nw.col("c").filter(*mask_multi).count(),
+        count_d=nw.col("d").filter(*mask_multi).count(),
+        # len=nw.col("d").filter(*mask_multi).len()  # noqa: ERA001
+        # !NOTE: Result should be {"len": [3]}, but we can get:
+        # 5: Without changing the current implementation for SQL backends
+        # 1: By accounting for the expression, (e.g. for spark F("count", expr)) but ignoring nulls
+    )
```
dangotbanned, line 94: @FBruzzesi are we using [Scalar Subqueries](https://duckdb.org/docs/stable/sql/expressions/subqueries#scalar-subquery) much in the sql backends?

This would be like:

```py
import duckdb
import polars as pl

df = pl.DataFrame(data).with_columns(d=pl.when(pl.col("i") > 2).then("i"))


def _(**named_agg: str) -> str:
    alias, expr = next(iter(named_agg.items()))
    return f"SELECT {expr} AS {alias}"


from_df = "FROM df"
where = "WHERE i > 0 AND a < 4"

query = (
    f"FROM ({_(a='mean(a)')} {from_df} WHERE i < 3),"
    f"({_(b='sum(b)')} {from_df} {where}),"
    f"({_(b_min='min(b)')} {from_df} {where}),"
    f"({_(count_c='count(c)')} {from_df} {where}),"
    f"({_(count_d='count(d)')} {from_df} {where}),"
    f"({_(len='count()')} {from_df} {where})"
)
duckdb.sql(query)
```

```
┌────────┬────────┬───────┬─────────┬─────────┬───────┐
│   a    │   b    │ b_min │ count_c │ count_d │  len  │
│ double │ int128 │ int64 │  int64  │  int64  │ int64 │
├────────┼────────┼───────┼─────────┼─────────┼───────┤
│    1.0 │     10 │     2 │       3 │       1 │     3 │
└────────┴────────┴───────┴─────────┴─────────┴───────┘
```

> [!NOTE]
> If we're already using them - at least the `len` issue can be resolved 😄

PR #3444, opened 2026-02-02
feat: Add support for indices and values `Series` in `Series.scatter`

### LINKED ISSUE #2155: enh: `scatter` should accept `narwhals.Series` for indices
Currently, `Series.scatter` only accepts `int | Sequence[int]` for the positions

I think we should also allow a `narwhals.Series`

FBruzzesi:
# Description

@MarcoGorelli this might pop up in sklearn/skrub 😂

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2155
- Arrow refactor drive by #3305

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/narwhals/_duckdb/expr_dt.py
+++ b/narwhals/_duckdb/expr_dt.py
@@ -10,10 +10,10 @@ from narwhals._constants import (
     US_PER_MINUTE,
     US_PER_SECOND,
 )
-from narwhals._duckdb.utils import UNITS_DICT, F, fetch_rel_time_zone, lit
+from narwhals._duckdb.utils import UNITS_DICT, F, fetch_rel_time_zone, lit, sql_expression
 from narwhals._duration import Interval
 from narwhals._sql.expr_dt import SQLExprDateTimeNamesSpace
-from narwhals._utils import not_implemented
+from narwhals._utils import not_implemented, requires
 
 if TYPE_CHECKING:
     from collections.abc import Sequence
@@ -97,14 +97,15 @@ class DuckDBExprDateTimeNamespace(SQLExprDateTimeNamesSpace["DuckDBExpr"]):
 
         return self.compliant._with_elementwise(_truncate)
 
+    @requires.backend_version((1, 3))
     def offset_by(self, by: str) -> DuckDBExpr:
         interval = Interval.parse_no_constraints(by)
         format = lit(f"{interval.multiple!s} {UNITS_DICT[interval.unit]}")
 
         def _offset_by(expr: Expression) -> Expression:
-            return F("date_add", format, expr)
+            return expr + sql_expression(f"interval {format}")
 
-        return self.compliant._with_callable(_offset_by)
+        return self.compliant._with_elementwise(_offset_by)
 
     def _no_op_time_zone(self, time_zone: str) -> DuckDBExpr:
         def func(df: DuckDBLazyFrame) -> Sequence[Expression]:
--- a/narwhals/_duckdb/expr_list.py
+++ b/narwhals/_duckdb/expr_list.py
@@ -29,7 +29,7 @@ class DuckDBExprListNamespace(
                 F("list_append", expr_distinct, lit(None)),
             ).otherwise(expr_distinct)
 
-        return self.compliant._with_callable(func)
+        return self.compliant._with_elementwise(func)
 
     def contains(self, item: NonNestedLiteral) -> DuckDBExpr:
         return self.compliant._with_elementwise(
@@ -63,7 +63,7 @@ class DuckDBExprListNamespace(
                 expr_sum
             )
 
-        return self.compliant._with_callable(func)
+        return self.compliant._with_elementwise(func)
 
     def sort(self, *, descending: bool, nulls_last: bool) -> DuckDBExpr:
         sort_direction = "DESC" if descending else "ASC"
--- a/narwhals/_spark_like/expr_dt.py
+++ b/narwhals/_spark_like/expr_dt.py
@@ -119,7 +119,7 @@ class SparkLikeExprDateTimeNamespace(SQLExprDateTimeNamesSpace["SparkLikeExpr"])
                 UNITS_DICT[unit], F.lit(multiple), expr
             )
 
-        return self.compliant._with_callable(_offset_by)
+        return self.compliant._with_elementwise(_offset_by)
 
     def _no_op_time_zone(self, time_zone: str) -> SparkLikeExpr:  # pragma: no cover
         def func(df: SparkLikeLazyFrame) -> Sequence[Column]:
--- a/tests/expr_and_series/dt/offset_by_test.py
+++ b/tests/expr_and_series/dt/offset_by_test.py
@@ -6,6 +6,7 @@ import pytest
 
 import narwhals as nw
 from tests.utils import (
+    DUCKDB_VERSION,
     PANDAS_VERSION,
     Constructor,
     ConstructorEager,
@@ -138,7 +139,8 @@ def test_offset_by(
     by: str,
     expected: list[datetime],
 ) -> None:
-    df = nw.from_native(constructor(data))
+    if "duckdb" in str(constructor) and DUCKDB_VERSION < (1, 3):
+        pytest.skip()
     if any(x in by for x in ("y", "q", "mo")) and any(
         x in str(constructor) for x in ("dask", "pyarrow", "ibis")
     ):
@@ -149,6 +151,8 @@ def test_offset_by(
         request.applymarker(pytest.mark.xfail())
     if by.endswith("d") and any(x in str(constructor) for x in ("dask", "ibis")):
         request.applymarker(pytest.mark.xfail())
+
+    df = nw.from_native(constructor(data))
     result = df.select(nw.col("a").dt.offset_by(by))
     assert_equal_data(result, {"a": expected})
 
@@ -238,6 +242,8 @@ def test_offset_by_series(constructor_eager: ConstructorEager) -> None:
 
 
 def test_offset_by_invalid_interval(constructor: Constructor) -> None:
+    if "duckdb" in str(constructor) and DUCKDB_VERSION < (1, 3):
+        pytest.skip()
     df = nw.from_native(constructor(data))
     msg = "Invalid `every` string"
     with pytest.raises(ValueError, match=msg):
@@ -257,4 +263,21 @@ def test_offset_by_date_pandas() -> None:
     df = nw.from_native(pd.DataFrame({"a": [date(2020, 1, 1)]}))
     result = df.select(nw.col("a").dt.offset_by("1d"))
     expected = {"a": [date(2020, 1, 2)]}
+    assert_equal_data(result, expected)
+
+
+def test_offset_by_3471(constructor: Constructor, request: pytest.FixtureRequest) -> None:
+    if any(x in str(constructor) for x in ("dask", "ibis")):
+        request.applymarker(pytest.mark.xfail())
+    if "duckdb" in str(constructor) and DUCKDB_VERSION < (1, 3):
+        pytest.skip()
+    date_nw = nw.from_native(constructor({"date": [date(2026, 1, 31)]}))
+
+    existent_col_offset = nw.col("date").dt.offset_by("-1d")
+    result = date_nw.with_columns(existent_col_offset)
+    expected = {"date": [datetime(2026, 1, 30)]}
+    assert_equal_data(result, expected)
+
+    new_lit_offset = nw.lit(datetime(2026, 1, 31)).dt.offset_by("-1d")
+    result = date_nw.with_columns(date=new_lit_offset)
     assert_equal_data(result, expected)


narwhals/_arrow/series.py
```diff
@@ -452,45 +451,11 @@ def _gather_slice(self, rows: _SliceIndex | range) -> Self:
...
-        mask: _1DArray = np.zeros(self.len(), dtype=bool)
-        mask[indices_native] = True
-        # NOTE: Multiple issues
-        # - Missing `values` type
-        # - `mask` accepts a `np.ndarray`, but not mentioned in stubs
-        # - Missing `replacements` type
-        # - Missing return type
-        pc_replace_with_mask: Incomplete = pc.replace_with_mask
-        return self._with_native(pc_replace_with_mask(self.native, mask, values_native))
+    def scatter(self, indices: Self, values: Self) -> Self:
```
dangotbanned, line 454: This is very satisfying for (#3305) to finally pay off 🙏 

> Thought it might be worth splitting into a separate PR, since it would be easy to upstream to main and (possibly) avoid a `numpy` import

dangotbanned:
@FBruzzesi would you be open to me porting over the [test suite](https://github.com/narwhals-dev/narwhals/blob/0d81b8b8cd1d24a68d360e6f8cf742b45cc2bdec/tests/plan/series_scatter_test.py) as well?

I'd need to make *some* tweaks, but I think I could get the diff negative 😄 

(no worries if not)

FBruzzesi:
Sure @dangotbanned - I will first try to get the CI green again 🥦 🥦 

---

Fixing CI is really the broccoli's to all the currently baking cakes  🍰

tests/series_only/scatter_test.py
```diff
@@ -1,23 +1,59 @@
...
+    from collections.abc import Collection
+
+
+def series(frame: ConstructorEager, name: str, values: Collection[Any]) -> nw.Series[Any]:
+    return nw.from_native(frame({name: values})).get_column(name)
+
+
+@pytest.mark.filterwarnings(
+    "ignore:.*all arguments of to_dict except for the argument:FutureWarning"
+)
```
dangotbanned, line 21: `modin` triggers this warning because it passes `orient=list` as positional.

So I copied over the filter from `test_to_dict`:

 https://github.com/narwhals-dev/narwhals/blob/1a1d3235ac8f46e6a1fda3828ff3d958f4978641/tests/frame/to_dict_test.py#L9-L12


But I think we should either:
1. Handle this in `narwhals`
2. Add it to `tool.pytest.ini_options.filterwarnings`

tests/series_only/scatter_test.py
```diff
@@ -1,23 +1,59 @@
...
+    unchanged_indexed = df.to_dict(as_series=False)
+    assert_equal_series(s.scatter(indices, values), expected, "s")
+    if not isinstance(indices, int):
+        assert_equal_series(s.scatter(constructor("i", indices), values), expected, "s")
+    if not isinstance(values, int):
+        assert_equal_series(s.scatter(indices, constructor("v", values)), expected, "s")
+    assert_equal_data(df, unchanged_indexed)
+
+
+def test_scatter_pandas_index() -> None:
```
dangotbanned, line 56: I renamed this to make it clearer, since `indices` is a parameter name for `scatter`.

Although, I'd prefer if we could use a less direct way of testing index preservation.
E.g. if this is important, why only `pandas` and not all the derivatives?

narwhals/series.py
```diff
@@ -394,12 +394,16 @@ def to_native(self) -> IntoSeriesT:
         """
         return self._compliant_series.native
 
-    def scatter(self, indices: int | Sequence[int], values: Any) -> Self:
-        """Set value(s) at given position(s).
+    def scatter(
+        self,
+        indices: Self | Iterable[int] | int,
+        values: Self | Iterable[PythonLiteral] | PythonLiteral | None,
```
dangotbanned: `PythonLiteral` includes `None` 🙂 

```suggestion
        values: Self | Iterable[PythonLiteral] | PythonLiteral,
```

narwhals/_polars/series.py
```diff
@@ -504,8 +504,8 @@ def sort(self, *, descending: bool, nulls_last: bool) -> Self:
 
         return self._with_native(result)
 
-    def scatter(self, indices: int | Sequence[int], values: Any) -> Self:
-        s = self.native.clone().scatter(indices, extract_native(values))
+    def scatter(self, indices: Self, values: Self) -> Self:
+        s = self.native.clone().scatter(extract_native(indices), extract_native(values))
```
dangotbanned: Since we know both will be `PolarsSeries` now

```suggestion
        s = self.native.clone().scatter(indices.native, values.native)
```

narwhals/series.py
```diff
@@ -436,8 +440,24 @@ def scatter(self, indices: int | Sequence[int], values: Any) -> Self:
...
+        if indices.is_empty():
+            return self
+
+        if isinstance(values, str) or not isinstance(values, Iterable):
+            values = Series.from_iterable(name="", values=[values], backend=impl)
+        elif not isinstance(values, Series):
+            values = Series.from_iterable(name="", values=values, backend=impl)
+
+        compliant_indices = self._extract_native(indices)
+        compliant_values = self._extract_native(values)
```
FBruzzesi: PR unrelated: Should we rename `_extract_native` to `_extract_compliant`?

dangotbanned: Lol, yeah I don't know when that slipped through 😂

FBruzzesi:
@dangotbanned let me know if you have time and manage to replicate the environment for which the tests are failing.

I am on an Apple Silicon (macOS ARM64), and the old versions of pandas (1.1.3) and numpy (1.19.3) don't have pre-built wheels for this architecture. When pip/uv tries to build them from source, it fails 😭

---

Update: solved in https://github.com/narwhals-dev/narwhals/pull/3444/commits/6b23fdce867dbfae5c46b227180c029306d83284

Might be nice to factor out `scatter` and `_scatter_in_place` for pandas before merging.

Update 2: https://github.com/narwhals-dev/narwhals/pull/3444/commits/b714afba92859780b53983285441cc691aacdaf9

narwhals/series.py
dangotbanned: @FBruzzesi I thought I remembered discussing this!

https://github.com/narwhals-dev/narwhals/pull/2886#discussion_r2245792714

> Weirdly, the same thing has the wrong name on `nw.Series` 🤔
> 
> https://github.com/narwhals-dev/narwhals/blob/0e806561cd76f052fe3038a46022a9443c581568/narwhals/series.py#L376-L381

FBruzzesi: @dangotbanned which algorithm do you use to index all past comments and conversations in your brain? I am jealous 😂

dangotbanned: > @dangotbanned which algorithm do you use to index all past comments and conversations in your brain? I am jealous 😂

Deju vu + obsessive thoughts 😭

narwhals/_pandas_like/series.py
```diff
@@ -280,34 +280,40 @@ def ewm_mean(
...
+    @overload
+    def _scatter(
+        self, indices: Self, values: Self, *, in_place: Literal[True]
+    ) -> None: ...
+    @overload
+    def _scatter(
+        self, indices: Self, values: Self, *, in_place: Literal[False]
+    ) -> Self: ...
+
+    def _scatter(self, indices: Self, values: Self, *, in_place: bool) -> Self | None:
```
dangotbanned: How would you feel about this?

The idea being that the default behavior is the safer option

```suggestion
    @overload
    def _scatter(self, indices: Self, values: Self) -> Self: ...
    def _scatter(
        self, indices: Self, values: Self, *, in_place: bool = False
    ) -> Self | None:
```

(Requires one other small change)

narwhals/_pandas_like/series.py
```diff
@@ -280,34 +280,40 @@ def ewm_mean(
...
+            series.iloc[indices_native] = values_native
+
+        if in_place:
+            return None
+
+        series.name = self.name
+        return self._with_native(series)
+
+    def scatter(self, indices: Self, values: Self) -> Self:
+        return self._scatter(indices=indices, values=values, in_place=False)
```
dangotbanned: (Related to the `@overload` change)
```suggestion
        return self._scatter(indices=indices, values=values)
```

narwhals/_pandas_like/series.py
```diff
@@ -280,34 +280,40 @@ def ewm_mean(
...
+        series = native_series if in_place else native_series.copy(deep=True)
+
+        if (is_pandas := impl.is_pandas()) and in_place and parse_version(np) < (2,):
             values_native = values_native.copy()  # pragma: no cover
+
         min_pd_version = (1, 2)
-        if implementation is Implementation.PANDAS and backend_version < min_pd_version:
-            self.native.iloc[indices.native.values] = values_native  # noqa: PD011
+        if is_pandas and backend_version < min_pd_version:
+            series.iloc[indices_native.values] = values_native  # noqa: PD011
```
dangotbanned: Not gonna install ancient `pandas` for this, but why are we using a deprecation from [`0.24.0`](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.24.0.html#accessing-the-values-in-a-series-or-index)? 😂

```suggestion
            series.iloc[indices_native.array] = values_native
```

narwhals/_pandas_like/series.py
```diff
@@ -280,34 +280,40 @@ def ewm_mean(
...
         else:
-            self.native.iloc[indices.native] = values_native
+            series.iloc[indices_native] = values_native
+
+        if in_place:
+            return None
+
+        series.name = self.name
+        return self._with_native(series)
```
dangotbanned: Isn't the `self.name` assignment already handled in `__init__`?

```suggestion
        return None if in_place else self._with_native(series)
```

narwhals/_pandas_like/series.py
```diff
@@ -280,34 +280,40 @@ def ewm_mean(
...
+
         min_pd_version = (1, 2)
-        if implementation is Implementation.PANDAS and backend_version < min_pd_version:
-            self.native.iloc[indices.native.values] = values_native  # noqa: PD011
+        if is_pandas and backend_version < min_pd_version:
+            series.iloc[indices_native.values] = values_native  # noqa: PD011
         else:
-            self.native.iloc[indices.native] = values_native
+            series.iloc[indices_native] = values_native
+
```
dangotbanned: This could also be combined with the `.values` -> `.array` comment, but I'm more confident this is a pure refactor

```suggestion
        if impl.is_pandas():
            if in_place and parse_version(np) < (2,):
                values_native = values_native.copy()  # pragma: no cover
            if backend_version < (1, 2):
                indices_native = indices_native.values  # noqa: PD011
        series.iloc[indices_native] = values_native
```

I know the `parse_version(np)` was already there - so don't worry too much.
But since `over` calls this in a loop; it could make sense to define a module-level `NUMPY_VERSION`

narwhals/series.py
```diff
@@ -436,8 +440,29 @@ def scatter(self, indices: int | Sequence[int], values: Any) -> Self:
...
         return self._with_compliant(
-            self._compliant_series.scatter(indices, self._extract_native(values))
+            self._compliant_series.scatter(compliant_indices, compliant_values)
         )
```
dangotbanned: Have combined a few suggestions from comparing to [narwhals/_plan/series.py#L178-L193](https://github.com/narwhals-dev/narwhals/blob/0d8d81b2e44b8e5de729c0b0669b3d53e8526ce4/narwhals/_plan/series.py#L178-L193).
This isn't identical to it though (I promise!)


---

- `Series.from_iterable` would pass a `Version.MAIN` in stable
  - Also 4x of those can be collapsed to 2x
- We don't need `Int64` or `impl` in all cases, so they're more local this way
- When we know we have a `Series`, we can just use `self._compliant`
  - And avoid the confusing `_extract_native` 😄

```suggestion
        into_series = type(self).from_iterable
        if not isinstance(indices, Iterable):
            indices = [indices]
        if not isinstance(indices, Series):
            dtypes = self._version.dtypes
            indices = into_series("", indices, dtypes.Int64, backend=self.implementation)
        if indices.is_empty():
            return self
        if not isinstance(values, Series):
            if not isinstance(values, Iterable) or isinstance(values, str):
                values = [values]
            values = into_series("", values, backend=self.implementation)
        result = self._compliant.scatter(indices._compliant, values._compliant)
        return self._with_compliant(result)
```

narwhals/_arrow/series.py
```diff
@@ -451,37 +451,11 @@ def _gather_slice(self, rows: _SliceIndex | range) -> Self:
...
+        mask = pc.is_in(arange(start=0, end=len(self), step=1), indices.native)
```
dangotbanned, line 455: I discovered this recently
https://github.com/narwhals-dev/narwhals/blob/0d8d81b2e44b8e5de729c0b0669b3d53e8526ce4/pyproject.toml#L88

> Type-checkers will no longer accept calls to `numpy.arange` with `start` as a keyword argument. 
This was done for compatibility with the Array API standard. At runtime it is still possible to use `numpy.arange` with `start` as a keyword argument.


I know we're not *actually using* numpy here, just thought it was interesting 😄

dangotbanned (approved):
Thanks @FBruzzesi!

I have a few suggestions, but overall just very happy to slowly creep #2572 into main 😉

praise be narwhal

narwhals/_pandas_like/series.py
FBruzzesi: I would rather keep it as mandatory for now. It's used only in two places - it won't save us a lot of typing

narwhals/_pandas_like/series.py
FBruzzesi: Fair enough, will address

narwhals/series.py
FBruzzesi: Thanks for the suggestion! I am adding a couple more tweaks and committing

FBruzzesi: Here you go https://github.com/narwhals-dev/narwhals/pull/3444/commits/2d19bff25a0ccff4a89ae5e5c1ee467fbae57ac4

narwhals/_pandas_like/series.py
FBruzzesi: Added in https://github.com/narwhals-dev/narwhals/pull/3444/commits/2d19bff25a0ccff4a89ae5e5c1ee467fbae57ac4

narwhals/_pandas_like/series.py
dangotbanned: @FBruzzesi 

> It's used only in two places - it won't save us a lot of typing

That wasn't what I was going for 😉 

> > The idea being that the **default** behavior is the **safer** option

I just looked at this again, and we can remove `_scatter` entirely while still respecting the signature of `CompliantSeries.scatter`.

<p>

```diff
diff --git a/narwhals/_pandas_like/expr.py b/narwhals/_pandas_like/expr.py
index f2e162227..944a136b1 100644
--- a/narwhals/_pandas_like/expr.py
+++ b/narwhals/_pandas_like/expr.py
@@ -238,7 +238,7 @@ class PandasLikeExpr(EagerExpr["PandasLikeDataFrame", PandasLikeSeries]):
 
             sorting_indices = df.get_column(token)
             for s in results:
-                s._scatter(sorting_indices, s, in_place=True)
+                s.scatter(sorting_indices, s, in_place=True)
             return results
 
         return self.__class__(
@@ -384,7 +384,7 @@ class PandasLikeExpr(EagerExpr["PandasLikeDataFrame", PandasLikeSeries]):
                     # Ignore settingwithcopy warnings/errors, they're false-positives here.
                     warnings.filterwarnings("ignore", message="\n.*copy of a slice")
                     for s in results:
-                        s._scatter(sorting_indices, s, in_place=True)
+                        s.scatter(sorting_indices, s, in_place=True)
                     return results
             if reverse:
                 return [s._gather_slice(slice(None, None, -1)) for s in results]
diff --git a/narwhals/_pandas_like/series.py b/narwhals/_pandas_like/series.py
index fe7defa90..8f90b7211 100644
--- a/narwhals/_pandas_like/series.py
+++ b/narwhals/_pandas_like/series.py
@@ -280,16 +280,15 @@ class PandasLikeSeries(EagerSeries[Any]):
         return self._with_native(result)
 
     @overload
-    def _scatter(
+    def scatter(self, indices: Self, values: Self) -> Self: ...
+    @overload
+    def scatter(
         self, indices: Self, values: Self, *, in_place: Literal[True]
     ) -> None: ...
-    @overload
-    def _scatter(
-        self, indices: Self, values: Self, *, in_place: Literal[False]
-    ) -> Self: ...
-
-    def _scatter(self, indices: Self, values: Self, *, in_place: bool) -> Self | None:
-        impl, backend_version = self._implementation, self._backend_version
+    def scatter(
+        self, indices: Self, values: Self, *, in_place: bool = False
+    ) -> Self | None:
+        impl = self._implementation
         native_series, indices_native = self.native, indices.native
         values_native = set_index(
             values.native, native_series.index[indices_native], implementation=impl
@@ -299,16 +298,13 @@ class PandasLikeSeries(EagerSeries[Any]):
         if impl.is_pandas():
             if in_place and NUMPY_VERSION < (2,):  # pragma: no cover
                 values_native = values_native.copy()
-            if backend_version < (1, 2):
+            if self._backend_version < (1, 2):
                 indices_native = indices_native.to_numpy()
 
         series.iloc[indices_native] = values_native
 
         return None if in_place else self._with_native(series)
 
-    def scatter(self, indices: Self, values: Self) -> Self:
-        return self._scatter(indices=indices, values=values, in_place=False)
-
     def cast(self, dtype: IntoDType) -> Self:
         if self.dtype == dtype and self.native.dtype != "object":
             # Avoid dealing with pandas' type-system if we can. Note that it's only

```


</p>
</details> 

But we can't do that if `in_place` doesn't have a default 😢

FBruzzesi: Applied in https://github.com/narwhals-dev/narwhals/pull/3444/commits/4c6dcd377b44aca3aa08073a3081a6b9194c4f78

FBruzzesi:
@dangotbanned if you are happy (or I would say happier since the last approval in https://github.com/narwhals-dev/narwhals/pull/3444#pullrequestreview-3766934057) with this, I will go ahead and merge

dangotbanned (approved):
Double approve

MarcoGorelli:
thanks all, solid!

narwhals/series.py
dangotbanned: > PR unrelated: Should we rename `_extract_native` to `_extract_compliant`?

@MarcoGorelli FYI, I will forget 😂

PR #3449, opened 2026-02-04
ci: bump `ruff==0.15.0`

dangotbanned:
# Description
See fancy blog post for changes https://astral.sh/blog/ruff-v0.15.0

How does that impact us? 
See ([chore(ruff): Ignore new issues](https://github.com/narwhals-dev/narwhals/pull/3449/commits/a02492c08530022c61b665f35259e48013627115)) and ([style(ruff): Run new style guide](https://github.com/narwhals-dev/narwhals/pull/3449/commits/3c663f00354ba179af32bf3ab76825398ff837de))

## Related issues

- https://github.com/narwhals-dev/narwhals/pull/3443#issuecomment-3846881466

### DIFF
--- a/.github/workflows/typing.yml
+++ b/.github/workflows/typing.yml
@@ -26,7 +26,7 @@ jobs:
           cache-suffix: typing-${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: Create venv
-        run: uv venv .venv
+        run: uv venv .venv --clear
       - name: install-reqs
         # TODO: add more dependencies/backends incrementally
         run: uv pip install -e ".[pyspark]" --group core --group typing-ci
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -129,7 +129,7 @@ def test_concat_diagonal_invalid(
         )
     )
     context: Any
-    if impl.is_polars() and POLARS_VERSION < (1, 1):  # pragma: no cover
+    if impl.is_polars() and POLARS_VERSION < (1, 18):  # pragma: no cover
         context = pytest.raises(
             NarwhalsError,
             match=re.compile(r"(int.+datetime)|(datetime.+int)", re.IGNORECASE),


dangotbanned:
> [!NOTE]
> Unrelated TPCH failure, see #3450

tests/dtypes/dtypes_test.py
```diff
@@ -68,7 +68,7 @@ def test_list_valid() -> None:
     assert dtype == nw.List(nw.List(nw.Int64))
     assert dtype == nw.List
     assert dtype != nw.List(nw.List(nw.Float32))
-    assert dtype in {nw.List(nw.List(nw.Int64))}
```
MarcoGorelli, line 71: are we sure about changing this? wouldn't we still want to test that `assert dtype in {nw.List(nw.List(nw.Int64))}` passes?

like this it feels like we're testing implementation details, it may be better to just noqa it?

dangotbanned: How about a function (with whatever in it), that has a name more fitting to the test?

E.g. 

```diff
-    assert dtype in {nw.List(nw.List(nw.Int64))}
+    assert_hash_equals(dtype, nw.List(nw.List(nw.Int64)))
```

I actually applied the [auto-fix](https://docs.astral.sh/ruff/rules/single-item-membership-test/) first, before realizing it changed the thing we test.

That seems like a pretty easy mistake to make and I might not have caught it without knowing the implementation detail 😅 


**Edit:**
How easy it is to make that mistake?

Well I managed to do it in the same commit by only getting the first one right 🤦‍♂️ 

https://github.com/narwhals-dev/narwhals/blob/b144f1e4bc775b2bcb317379a3157fe349b6ac35/tests/v1_test.py#L534-L537

MarcoGorelli: sure!

dangotbanned: Hopefully this'll do 🙂 (https://github.com/narwhals-dev/narwhals/pull/3449/commits/6d5e3d5e290a1c9da3fd3f985079343f7170f698) 

```py
>       assert_equal_hash(dtype, nw.Enum(["a", "b", "c"]))
E       AssertionError: inputs do not compare equal by `__hash__`
E       [left]: Enum(categories=['a', 'b'])
E       [right]: Enum(categories=['a', 'b', 'c'])

tests/dtypes/dtypes_test.py:548: AssertionError
```

Based on what's in `narwhals.tests`

tests/utils.py
```diff
@@ -166,6 +166,14 @@ def assert_equal_series(
     assert_equal_data(result.to_frame(), {name: expected})
 
 
+def assert_equal_hash(left: Any, right: Any) -> None:
+    """Assert that the left and right produce identical hash values."""
```
dangotbanned:
```suggestion
    """Assert that left and right produce identical hash values."""
```

narwhals/_dask/expr.py
```diff
@@ -235,7 +235,7 @@ def _with_alias_output_names(self, func: AliasNames | None, /) -> Self:
     def _with_binary(
         self, call: Callable[[dx.Series, Any], dx.Series], other: Any
     ) -> Self:
-        return self._with_callable(lambda expr, other: call(expr, other), other=other)
```
MarcoGorelli, line 238: 😄 nice

PR #3434, opened 2026-01-31
ci: Unpin `polars==1.34.0` in `--group typing`

dangotbanned:
# Description
I'm happy to go back to pinning if we have issues, but this one seems arbitrary.

I've been having to do lot's of manual edits when switching between (#2572).
#3352 needed a new pin since it tests against newer behavior

## Related
- Drive-by fix #3407

Issue #3407, opened 2026-01-19
Relax/fix `test_concat_diagonal_invalid` error for `polars>=1.0.0`


dangotbanned:
https://github.com/narwhals-dev/narwhals/actions/runs/21139459260/job/60789500820

```cmd
FAILED tests/frame/concat_test.py::test_concat_diagonal_invalid[polars[lazy]] - narwhals.exceptions.NarwhalsError: type Datetime(Microseconds, None) is incompatible with expected type Int64
FAILED tests/frame/concat_test.py::test_concat_diagonal_invalid[polars[eager]] - narwhals.exceptions.NarwhalsError: type Datetime(Microseconds, None) is incompatible with expected type Int64

```

I thought I'd caught all of these in (https://github.com/narwhals-dev/narwhals/pull/3404/commits/ceceb5b8ae13b9277513de7211a8151a86d92030), (https://github.com/narwhals-dev/narwhals/pull/3404/commits/75d5eac025719f53a020c44040a0cf912761c11d) and (https://github.com/narwhals-dev/narwhals/pull/3404/commits/f5a6fdc79f4f17d39c66d181f2e3e91c51c44eb3).
Seems like the error changed some time after `1.0.0`, not sure when though *yet*

PR #3451, opened 2026-02-04
test: fix polars `test_concat_diagonal_invalid` again

dangotbanned:
# Description

Was so close in (#3434) to getting (#3407)

[But fail](https://github.com/narwhals-dev/narwhals/actions/runs/21671044595/job/62478712486?pr=3449) ... maybe 5th's time the charm?

### DIFF
--- a/.github/workflows/check_tpch_queries.yml
+++ b/.github/workflows/check_tpch_queries.yml
@@ -26,10 +26,11 @@ jobs:
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: local-install
+        # temporary polars pin https://github.com/narwhals-dev/narwhals/issues/3450
         run: |
           uv pip install -U -e ".[dask]" --group core-tests --system
-          uv pip install -U duckdb --system
+          uv pip install -U duckdb "polars==1.37.1" --system
       - name: generate-data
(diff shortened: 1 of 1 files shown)


FBruzzesi:
Thanks @dangotbanned - I will run a few polars version to make sure we don't need to come bck to this again

tests/frame/concat_test.py
```diff
@@ -129,7 +129,7 @@ def test_concat_diagonal_invalid(
         )
     )
     context: Any
-    if impl.is_polars() and POLARS_VERSION < (1, 1):  # pragma: no cover
+    if impl.is_polars() and POLARS_VERSION < (1, 18):  # pragma: no cover
```
FBruzzesi, line 132: @dangotbanned it would have taken... 17 more attempts 😂

FBruzzesi (approved):

tests/frame/concat_test.py
dangotbanned: ahh, I guess we'd have returned to this rabbit hole the next time these were updated.
Thanks @FBruzzesi!

https://github.com/narwhals-dev/narwhals/blob/9a15d1c6c0193f1e5669160666e7d1cf309d6f55/utils/generate_random_versions.py#L16-L40

Issue #3450, opened 2026-02-04
bug: `tpch` queries suddenly using `string_view`


dangotbanned:
## Related
- [x] https://github.com/pola-rs/polars/issues/26435

## Description
Starting from (https://github.com/pola-rs/polars/releases/tag/py-1.38.0) the queries are [failing](https://github.com/narwhals-dev/narwhals/actions/runs/21670750471/job/62477688240?pr=3449).

<details><summary><b>Show very long logs</b></summary>
<p>

```py

=========================== short test summary info ============================
FAILED tests/queries_test.py::test_execute_query[q11-dask] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, int64)
FAILED tests/queries_test.py::test_execute_query[q15-pandas[pyarrow]] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, int64)
FAILED tests/queries_test.py::test_execute_query[q22-pyarrow] - pyarrow.lib.ArrowNotImplementedError: Function 'utf8_slice_codeunits' has no kernel matching input types (string_view)
FAILED tests/queries_test.py::test_execute_query[q20-pyarrow] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, uint32)
FAILED tests/queries_test.py::test_execute_query[q17-pandas[pyarrow]] - NotImplementedError: string_view
FAILED tests/queries_test.py::test_execute_query[q11-pandas[pyarrow]] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, int64)
FAILED tests/queries_test.py::test_execute_query[q13-pandas[pyarrow]] - NotImplementedError: string_view
FAILED tests/queries_test.py::test_execute_query[q7-dask] - NotImplementedError: string_view
FAILED tests/queries_test.py::test_execute_query[q19-pandas[pyarrow]] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, int64)
FAILED tests/queries_test.py::test_execute_query[q6-pandas[pyarrow]] - pyarrow.lib.ArrowNotImplementedError: Function 'array_take' has no kernel matching input types (string_view, int64)
[...]

PR #3446, opened 2026-02-03
ci: run ibis and modin tests in separate workflows

MarcoGorelli:
closes #3445

Nightly failure seems unrelated

Modin pins pandas <3, and ibis often requires sqlglot pinning

# Description

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #3453, opened 2026-02-04
ci(tpch): Pin `"polars==1.37.1"` for tpch queries

dangotbanned:
# Description
Lost a lot of time today trying to figure out a good solution to this.

AFAICT, there is no way to write strings anymore 😩

If anyone has any better ideas let me know.
I think `polars` needs to support this, or document how to opt-out of the change:

```py

pl.LazyFrame().sink_parquet(compat_level=pl.CompatLevel.oldest())
```

## Related issues
- [x] https://github.com/pola-rs/polars/pull/26436
- Temp bandage for #3450
- Upstream
  - https://github.com/pola-rs/polars/pull/26236
  - https://github.com/pola-rs/polars/pull/26323

### DIFF
--- a/.github/workflows/check_tpch_queries.yml
+++ b/.github/workflows/check_tpch_queries.yml
@@ -26,10 +26,11 @@ jobs:
           cache-suffix: ${{ matrix.python-version }}
           cache-dependency-glob: "pyproject.toml"
       - name: local-install
+        # temporary polars pin https://github.com/narwhals-dev/narwhals/issues/3450
         run: |
           uv pip install -U -e ".[dask]" --group core-tests --system
-          uv pip install -U duckdb --system
+          uv pip install -U duckdb "polars==1.37.1" --system
       - name: generate-data
-        run: cd tpch && python generate_data.py
+        run: cd tpch && python generate_data.py --debug
       - name: tpch-tests 
         run: cd tpch && pytest tests


.github/workflows/check_tpch_queries.yml
```diff
@@ -26,10 +26,11 @@ jobs:
...
           cache-dependency-glob: "pyproject.toml"
       - name: local-install
+        # temporary polars pin https://github.com/narwhals-dev/narwhals/issues/3450
         run: |
           uv pip install -U -e ".[dask]" --group core-tests --system
-          uv pip install -U duckdb --system
+          uv pip install -U duckdb "polars==1.37.1" --system
       - name: generate-data
-        run: cd tpch && python generate_data.py
+        run: cd tpch && python generate_data.py --debug
```
dangotbanned, line 34: Sadly this doesn't help the particular issue, but should be helpful in the future

- https://github.com/narwhals-dev/narwhals/actions/runs/21683680316/job/62524789316?pr=3453#step:6:19

dangotbanned:
Okay it seems this has gotten marked as a high priority regression.

Moving into draft

MarcoGorelli:
should we also pin to avoid https://github.com/pola-rs/polars/issues/26423

dangotbanned:
> should we also pin to avoid [pola-rs/polars#26423](https://github.com/pola-rs/polars/issues/26423)

Probably, but maybe in a similar vein to (#3446) - what do you think about running polars streaming in it's own workflow?

I wouldn't have guessed that was the issue because [it looks to be disabled?](https://github.com/narwhals-dev/narwhals/actions/runs/21683680356/job/62524789295?pr=3453#step:8:7):

```log
  NARWHALS_POLARS_NEW_STREAMING: false
```

I do understand this could end up being a lot of effort if there's big gaps in support (so a no is totally fine!).
Just trying to think how we can make our CI tell us more on a fail 😅 

---

Either way, should go in a different PR since the pin won't touch (https://github.com/narwhals-dev/narwhals/pull/3453/files) 🙂

dangotbanned:
Thanks @MarcoGorelli

PR #3455, opened 2026-02-05
ci: Replace `--system` with `$UV_SYSTEM_PYTHON`

dangotbanned:
# Description

Reduces the amount of noise in each command
https://docs.astral.sh/uv/reference/environment/#uv_system_python


.github/workflows/downstream_tests.yml
```diff
@@ -8,6 +8,8 @@ env:
 
 jobs:
   altair:
+    env:
+      UV_SYSTEM_PYTHON: 1
```
dangotbanned, line 12: I went per-job here because they weren't consistently using `--system`.

Not sure if that was intentional, but I've preserved that anyway

PR #3460, opened 2026-02-05
feat: Add `nw.corr`

### LINKED ISSUE #2191: enh: nw.corr
Very little self-contained feature I was looking for today, I think we should probably have it?

benrutter:
# Description

Implementation of `nw.corr` as a top level function.

A few things to highlight:
- There's (still) a crazy hack in `_expression_parsing.py`, I'll try and dig around a bit to see why it's needed, but have left this PR marked as 'draft' whilst it's still there (aside from that hack, I think everything else is ready)
- I've supported passing in a method of "spearman" and "pearson" for the coefficient method, which are supported by pandas/polars. A lot of libraries just support spearman, so for those, I've ignored the argument and issued a UserWarning if you, eg, try to call "pearson" on a duckdb dataframe (but the function will still work, it'll just evaluate to the spearman implementation)
- I *haven't* added support for `nw.corr(nw.col("a", "b"))` - would that be wanted? (I couldn't tell if that makes sense, or is a little bit non-sensical. Might need a little bit of "magic" to make implementable, since you need to pass in exactly two arguments.
- Polars API gives a `eager` argument - I *haven't* supported this after a steer from @FBruzzesi (thanks @FBruzzesi!). Instead, the Narwhals implementation will keep things either lazy of eager based on their initial state.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2191
- Probably needs integration with #3452 (i.e. depending which merges first, the new expr keyword needs to be integrated in `functions.py`)

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/tests/expr_and_series/corr_test.py
+++ b/tests/expr_and_series/corr_test.py
@@ -73,8 +73,8 @@ def test_corr_series(
     constructor_eager: ConstructorEager,
     request: pytest.FixtureRequest,
     output_name: str,
-    a: str | nw.Expr,
-    b: str | nw.Expr,
+    a: str,
+    b: str,
     expected_corr: float,
 ) -> None:
     if "pyspark" in str(constructor_eager) and expected_corr is None:
@@ -95,8 +95,8 @@ def test_corr_series_spearman(
     constructor_eager: ConstructorEager,
     request: pytest.FixtureRequest,
     output_name: str,
-    a: str | nw.Expr,
-    b: str | nw.Expr,
+    a: str,
+    b: str,
     expected_corr: float,
 ) -> None:
     if "pyspark" in str(constructor_eager) and expected_corr is None:


MarcoGorelli:
wow, thanks @benrutter !

> There's a absolutely mad hack in _expression_parsing.py (it's the only committed line in that file), it clearly should be merged, but I don't know why I had to add it or what's really going on to need that there? (taking it out throws an error because the class method it calls doesn't accept additional args and one for each expression for the a/b is given)

i'll take a look but don't worry about it for now, keep what you've got while you work out the other details

> I feel like my test doesn't cover enough scenarios, it only does nw.corr(nw.col("a"), nw.col("b")) what other contexts might nw.corr get used in?

there's also `nw.col('a', 'b')`, and `nw.col(df['a'], df['b'])`. but, fine to leave that til later if you with and just start with only accepting expressions

> I'm not particularly experienced with a fair few of the dataframe libraries (like duckdb or ibis), and might have accidentally implemented things in a silly/non-conventional/non-performant way.

as long as you're using native apis it's probably fine

narwhals/_duckdb/namespace.py
```diff
@@ -165,3 +162,23 @@ def func(_df: DuckDBLazyFrame) -> list[Expression]:
...
+        if method != "pearson":
+            msg = (
+                f"You have requested {method} correlation but duckdb only provides "
+                "native support for pearson correlation coefficients. Correlation will "
+                "be calculated using pearson."
+            )
+            issue_warning(msg)
+
+        def func(df: DuckDBLazyFrame) -> list[Expression]:
+            [a_], [b_] = [e(df) for e in [a, b]]
```
MarcoGorelli:
```py
            [a_], [b_] = [a(df), b(df)]
```
let's keep it simple?

or, even better, see if you can use the `evaluate_single_output_expr` utility (or something like that) here

benrutter: Ah nice! I've used evaluate here (and in other places where applicable too)

benrutter:
(Also I don't think I have permissions to rename the PR title so it might need a title of `feat: Add support for top level nw.corr function`)

benrutter:
Ok, think this is ready for review! I *think* I've fixed the workaround, although I had to introduce `ExprKind.MULTI_COLUMN_AGGREGATION` - from what I can figure, but I'm only like 30% sure) the issue is that all other scalar style things expect only one related column (`nw.col("a").sum()`) whereas `corr` is a but unusual because it's an aggregation but needs to be in a different form.

There's a very good chance that adding a new `ExprKind` is a bit overkill, the other option that I can figure would be to just allow passing in ignorable expressions into the `from_aggregation` argument:
```python
    @classmethod
    def from_aggregation(cls, node: ExprNode, *compliant_exprs: ComplaintExprAny) -> ExprMetadata:
        del compliant_exprs
        return cls(
            ExpansionKind.SINGLE,
            is_elementwise=False,
            preserves_length=False,
            is_scalar_like=True,
            current_node=node,
            prev=None,
        )
```

narwhals/_dask/namespace.py
```diff
@@ -320,3 +321,21 @@ def func(df: DaskLazyFrame) -> list[dx.Series]:
             alias_output_names=getattr(then, "_alias_output_names", None),
             version=self._version,
         )
+
+    def corr(
+        self, a: DaskExpr, b: DaskExpr, method: Literal["pearson", "spearman"] = "pearson"
+    ) -> DaskExpr:
+        def func(df: DaskLazyFrame) -> list[dx.Series]:
+            a_name, b_name = [s.name for _expr in [a, b] for s in _expr(df)]
```
MarcoGorelli: let's also use `evaluate` here?

narwhals/_duckdb/namespace.py
```diff
@@ -165,3 +168,28 @@ def func(_df: DuckDBLazyFrame) -> list[Expression]:
...
+        if method != "pearson":
+            msg = (
+                f"You have requested {method} correlation but duckdb only provides "
+                "native support for pearson correlation coefficients. Correlation will "
+                "be calculated using pearson."
+            )
+            issue_warning(msg, UserWarning)
+
+        def func(df: DuckDBLazyFrame) -> list[Expression]:
+            (_, a_), (_, b_) = evaluate_exprs_and_aliases(df, a, b)
```
MarcoGorelli: i'm confused about why we're using `evaluate_exprs_and_aliases` here and `evaluate` for ibis?

narwhals/_pandas_like/namespace.py
```diff
@@ -373,6 +373,29 @@ def _if_then_else(
...
         return where(when) if otherwise is None else where(when, otherwise)
 
+    def corr(
+        self,
+        a: NativeSeriesT,
+        b: NativeSeriesT,
+        method: Literal["pearson", "spearman"] = "pearson",
+    ) -> None:
+        def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
+            a_name, b_name = [s.name for _expr in [a, b] for s in _expr(df)]
```
MarcoGorelli: same?

narwhals/_arrow/namespace.py
```diff
@@ -254,3 +254,7 @@ def _if_then_else(
     ) -> ChunkedArrayAny:
         otherwise = pa.nulls(len(when), then.type) if otherwise is None else otherwise
         return pc.if_else(when, then, otherwise)
+
+    def corr(self, a: ArrowExpr, b: ArrowExpr, method: str = "pearson") -> ArrowExpr:
```
MarcoGorelli: `Literal["pearson", "spearman"]`?

narwhals/_polars/namespace.py
```diff
@@ -225,6 +225,14 @@ def selectors(self) -> CompliantSelectorNamespace[PolarsDataFrame, PolarsSeries]
             PolarsSelectorNamespace(self),
         )
 
+    def corr(
+        self,
+        a: PolarsExpr | PolarsSeries,
+        b: PolarsExpr | PolarsSeries,
```
MarcoGorelli: i think that by the time we get here, it must be `PolarsExpr`?

narwhals/_sql/namespace.py
```diff
@@ -86,3 +87,9 @@ def func_with_otherwise(cols: list[NativeExprT]) -> NativeExprT:
         return self._expr._from_elementwise_horizontal_op(
             func_with_otherwise, then, predicate, otherwise
         )
+
+    def corr(
+        self, a: SQLExprT, b: SQLExprT, method: Literal["pearson", "spearman"] = "pearson"
+    ) -> SQLExprT:
+        msg = "Correlation not implemented yet for SQL like dataframes"
+        raise NotImplementedError(msg)
```
MarcoGorelli: I think we should be able to share some code for the ibis/pyspark/duckdb implementations? they look quite similar

their namespaces should all inherit from SQLNamespace

narwhals/stable/v2/__init__.py
```diff
@@ -813,6 +813,38 @@ def max_horizontal(*exprs: IntoExpr | Iterable[IntoExpr]) -> Expr:
...
+    """Compute the Pearson's or Spearman rank correlation between two columns.
+
+    Arguments:
+        a: Column name or Expression
+        b: Column name or Expression
+        method: Correlation method ('pearson' or 'spearman')
+
+    Examples:
+        >>> import polars as pl
+        >>> import narwhals as nw
```
MarcoGorelli: we usually leave examples out from the stable api docstrings

pyproject.toml
```diff
@@ -287,6 +287,9 @@ filterwarnings = [
 
   "ignore:.*invalid value encountered in cast:RuntimeWarning:pandas",
   "ignore:.*The behavior of DataFrame concatenation with empty or all-NA entries is deprecated:FutureWarning",
+
+  # Warning raised due to correlation method support
+  "ignore:.*does not pass down coefficient methods",
```
MarcoGorelli: i'm not sure about ignoring this globally. is it only some particular backend that raises this warning, and is it unnecessary? if so, can we silence the warning in the place where it's called, if it's something that users wouldn't be able to do anything about anyway?

narwhals/_expression_parsing.py
```diff
@@ -99,6 +99,9 @@ class ExprKind(Enum):
     AGGREGATION = auto()
     """Reduces to a single value, not affected by row order, e.g. `nw.col('a').mean()`"""
 
+    MULTI_COLUMN_AGGREGATION = auto()
+    """Reduces to a single value, taking multiple columns as an argument, e.g. `nw.corr('a', 'b')`"""
```
MarcoGorelli: not sure what i'm missing but i don't see why we need this, why can't it just be `AGGREGATION`?

narwhals/_duckdb/namespace.py
benrutter: I'd imported them both from `_ibis.utils` and `_duckdb.utils`, I don't think `_duckdb.utils` has a `evaluate`, but from what I can figure, those functions (`evalute_exprs_and_alias` and `evaluate` are equivalents with slightly confusing names (they both return tuples of strings and evaluations)?

MarcoGorelli: I think you can do something like
```py
a = df._evaluate_single_output_expr(a)
```

narwhals/_expression_parsing.py
benrutter: Think it *could* be, the issue I was finding was that because multiple arguments get passed in, when `ExprMetadata.from_aggregation` gets called, along with just the `node`, the compliant expressions also get passed in (causing a fail because from_aggregation only takes 2 arguments including self).

I think `AGGREGATION` would work if it's ok to change this:

```python
    @classmethod
    def from_aggregation(cls, node: ExprNode) -> ExprMetadata:
        return cls(
            ExpansionKind.SINGLE,
            is_elementwise=False,
            preserves_length=False,
            is_scalar_like=True,
            current_node=node,
            prev=None,
        )
```

to this:

```python
    @classmethod
    def from_aggregation(cls, node: ExprNode, *compliant_exprs: CompliantExprAny) -> ExprMetadata:
        del compliant_exprs
        return cls(
            ExpansionKind.SINGLE,
            is_elementwise=False,
            preserves_length=False,
            is_scalar_like=True,
            current_node=node,
            prev=None,
        )
```

narwhals/stable/v2/__init__.py
benrutter: Ah nice, taken out!

narwhals/_arrow/namespace.py
benrutter: Doh! (thought I'd caught them all)

narwhals/_duckdb/namespace.py
benrutter: Oh, that's sooo much nicer!

pyproject.toml
benrutter: Switched to filter warnings just in the tests, it's actually one I deliberately added in and raised 🤦

narwhals/_expression_parsing.py
MarcoGorelli: thanks! yup that looks like a reasonable change

narwhals/_sql/namespace.py
benrutter: I wondered about this too, I wasn't sure if I was over complicating things. From what I could figure they all have slightly different behaviour (although some of this might be me not understanding the libraries well engough):
- Ibis calls `corr` as a method from the columns, and it needs "pop" as a "how" argument (doesn't support pearson/spearman)
- DuckDB calls it as a neat function, although it needs a specific warning because only pearson is supported
- And then spark works quite a lot like DuckDB, it only supports pearson too (although I changed the warning *a bit* so that they mention duckdb / pearson.

So I think I could get DuckDB and Spark to share something? Maybe not Ibis? (although I might be missing the kind of sharing of code you meant!)

narwhals/_expression_parsing.py
benrutter: (silly question, but do you mean the reasonable change is adding in the `*compliant_exprs` or adding in a new `MULTI_COLUMN_AGGREGATION` enum value? 😅)

dangotbanned: > to this:
> 
> ```python
>     @classmethod
>     def from_aggregation(cls, node: ExprNode, *compliant_exprs: CompliantExprAny) -> ExprMetadata:
>         del compliant_exprs
>         return cls(
>             ExpansionKind.SINGLE,
>             is_elementwise=False,
>             preserves_length=False,
>             is_scalar_like=True,
>             current_node=node,
>             prev=None,
>         )
> ```

Why is `*compliant_exprs` being passed in and deleted?
If we don't need to use it, then the problem to fix is on the caller-side, no?

benrutter: I think you're probably right, I'm a little lost in this part of the codebase so there's a decent chance I'm confused. Looking at it, I think:

- We get a bunch of CES passed down, like we would if we where doing `nw.max_horizontal`
- In `max_horizontal` we combine the metadata
- It's possible we should do that here? It seemed to me it wouldn't get used, but maybe it would be in more complex examples?
- The caller logic is dynamic based on the expression type, it seemed more complex to me to have a conditional deciding not to pass it down in this case, vs passing down and ignoring (but again, you probably have a lot more knowledge/context on whether that's a good call!)

Sorry, I know some of that's a little vague! 😅

benrutter: Ok, I think I've implemented something a little neater and more logical now with a fresh Monday morning brain!

I've retained the `MULTI_COLUMN_AGGREGATION` type, but instead of just removing the compliant expressions, I've used the `combine_metadata` function. Since this is an aggregation, I had to create arguments to optionally force `scalar` and `preserves_length` setting.

I think this makes a lot more sense that deleting the compliant expressions, whether through a new method or the old method, and should make edge case bugs around things like literals less likely.

narwhals/_duckdb/namespace.py
```diff
@@ -165,3 +167,29 @@ def func(_df: DuckDBLazyFrame) -> list[Expression]:
...
+        b: DuckDBExpr,
+        method: Literal["pearson", "spearman"] = "pearson",
+    ) -> DuckDBExpr:
+        if method != "pearson":
+            msg = (
+                f"You have requested {method} correlation but duckdb only provides "
+                "native support for pearson correlation coefficients. Correlation will "
+                "be calculated using pearson."
+            )
+            issue_warning(msg, UserWarning)
```
MarcoGorelli: this should raise an error, a warning isn't enough

narwhals/_ibis/namespace.py
```diff
@@ -141,3 +143,24 @@ def func(_df: IbisLazyFrame) -> list[ir.Value]:
...
         )
+
+    def corr(
+        self, a: IbisExpr, b: IbisExpr, method: Literal["pearson", "spearman"] = "pearson"
+    ) -> IbisExpr:
+        msg = (
+            "For wider dataframe support, Narwhals currently does not pass down coefficient "
+            'methods into Ibis .corr calls. ("pop" correlation used)'
+        )
+        issue_warning(msg, UserWarning)
```
MarcoGorelli: same

narwhals/_spark_like/namespace.py
```diff
@@ -218,3 +220,31 @@ def func(df: SparkLikeLazyFrame) -> list[Column]:
...
+        b: SparkLikeExpr,
+        method: Literal["pearson", "spearman"] = "pearson",
+    ) -> SparkLikeExpr:
+        if method != "pearson":
+            msg = (
+                f"You have requested {method} correlation but spark only provides "
+                "native support for pearson correlation coefficients. Correlation will "
+                "be calculated using pearson."
+            )
+            issue_warning(msg, UserWarning)
```
MarcoGorelli: same

narwhals/_sql/namespace.py
MarcoGorelli: yeah it's fine to share between duckdb and pyspark

not a blocker though, could be a follow-up, don't spend too long on this

tests/expr_and_series/corr_test.py
```diff
@@ -0,0 +1,44 @@
...
+data = {"a": [1, 3, 2], "b": [1, 2, 3]}
+
+
+@pytest.mark.filterwarnings("ignore::UserWarning")
+def test_expr_mean_expr(constructor: Constructor) -> None:
+    df = nw.from_native(constructor(data))
+    if any(x in str(constructor) for x in ("pyarrow_table_constructor",)):
+        with pytest.raises(NotImplementedError, match="Correlation"):
+            df.select(corr=nw.corr(nw.col("a"), nw.col("b")))
+        return
```
MarcoGorelli: we'd usually xfail such tests, see other tests for how this is done

tests/expr_and_series/corr_test.py
```diff
@@ -0,0 +1,44 @@
...
+
+import pytest
+
+import narwhals as nw
+from tests.utils import Constructor, ConstructorEager, assert_equal_data
+
+data = {"a": [1, 3, 2], "b": [1, 2, 3]}
+
+
+@pytest.mark.filterwarnings("ignore::UserWarning")
```
MarcoGorelli: we should be able to remove this

MarcoGorelli (commented):
awesome thanks!

narwhals/_duckdb/namespace.py
dangotbanned: @benrutter I know this might sound strange considering #3007

That warning was trying to fix an existing issue, whereas for `corr` we can raise an error from the start.
If the backend ever does support that option, we can always enable it in the future 🙂

narwhals/_duckdb/namespace.py
```diff
@@ -165,3 +167,29 @@ def func(_df: DuckDBLazyFrame) -> list[Expression]:
...
+
+        def func(df: DuckDBLazyFrame) -> list[Expression]:
+            a_ = df._evaluate_single_output_expr(a)
+            b_ = df._evaluate_single_output_expr(b)
+            return [F("corr", a_, b_)]
+
+        return self._expr(
+            call=func,
+            evaluate_output_names=lambda _df: ["corr"],
+            alias_output_names=None,
```
MarcoGorelli: these should also be combined

benrutter: (just coming to this now, see that you've already fixed this! Thanks 🙏)

narwhals/_duckdb/namespace.py
benrutter: That makes total sense! I was away last week so just working through these now, but I think @MarcoGorelli you've already gone ahead and changed those (I appreciate it, hope you didn't feel forced to though! I promise I was planning on coming back to this 😅)

benrutter:
Ok, I think this is all ready for final review and potentially smooshing in? (@MarcoGorelli / @dangotbanned) I haven't actually pushed any new changes, running tests and looking at the codebase - @MarcoGorelli I think you've solved the issues you raised!

narwhals/_arrow/namespace.py
```diff
@@ -253,3 +253,39 @@ def _if_then_else(
     ) -> ChunkedArrayAny:
         otherwise = pa.nulls(len(when), then.type) if otherwise is None else otherwise
         return pc.if_else(when, then, otherwise)
+
+    def corr(
+        self, a: ArrowExpr, b: ArrowExpr, *, method: Literal["pearson", "spearman"]
+    ) -> ArrowExpr:
+        if method != "pearson":
+            msg = "Only 'pearson' correlation is supported for Spark."
```
FBruzzesi: There are a few instances in which the copy-paste of this message went wrong for the backend name

tests/expr_and_series/corr_test.py
FBruzzesi, line 1: Can we add test cases with nulls and nan's? polars has a flag `propagate_nans`: (False by default), I want to make sure that the same behavior is matched

narwhals/_spark_like/namespace.py
```diff
@@ -218,3 +219,28 @@ def func(df: SparkLikeLazyFrame) -> list[Column]:
...
+        method: Literal["pearson", "spearman"],
+    ) -> SparkLikeExpr:
+        if method != "pearson":
+            msg = "Only 'pearson' correlation is supported for Spark."
+            raise NotImplementedError(msg)
+
+        def func(df: SparkLikeLazyFrame) -> list[Column]:
+            F = self._F
+            a_ = df._evaluate_single_output_expr(a)
+            b_ = df._evaluate_single_output_expr(b)
```
FBruzzesi, line 232: I am not sure about other backends, but I have a hunch that this might trigger the plan for `df` twice in pyspark, which is not ideal. 

Summoning @EdAbati here as you have some more experience with PySpark

narwhals/functions.py
```diff
@@ -1266,6 +1266,38 @@ def _expr_with_horizontal_op(name: str, *exprs: IntoExpr, **kwargs: Any) -> Expr
     )
 
 
+def corr(
+    a: IntoExpr, b: IntoExpr, method: Literal["pearson", "spearman"] = "pearson"
```
FBruzzesi: Would it be worth it to factor out `Literal["pearson", "spearman"]` into a `CorrMethod` type alias?

narwhals/functions.py
```diff
@@ -1266,6 +1266,38 @@ def _expr_with_horizontal_op(name: str, *exprs: IntoExpr, **kwargs: Any) -> Expr
...
+) -> Expr:
+    """Compute the Pearson's or Spearman rank correlation between two columns.
+
```
FBruzzesi, line 1272: Can we mention that the ddof is implicitly 1?

narwhals/_spark_like/namespace.py
MarcoGorelli: `_evaluate_single_output_expr` doesn't collect anything, there's no compute call, so i don't believe it should trigger the plan twice. it's just transforming a narwhals expression into a spark one

narwhals/functions.py
benrutter: That sounds like a nice idea - would be pretty easy to mispell one as "spreaman" or something and makes changing this argument later a lot neater.

tests/expr_and_series/corr_test.py
benrutter: Great shout! (added one in)

narwhals/_arrow/namespace.py
benrutter: Nice catch - gone through and (I think!) caught them all 🕵️

narwhals/typing.py
```diff
@@ -244,6 +244,8 @@ def Binary(self) -> type[dtypes.Binary]: ...
 ]
 """Interpolation method."""
 
+CorrMethod: TypeAlias = Literal["pearson", "spearman"]
```
dangotbanned: Fully agree with (https://github.com/narwhals-dev/narwhals/pull/3460#discussion_r2852442939), just thought I'd mention:


## (1) ~~`CorrMethod`~~ `CorrelationMethod`
We can steal the name `CorrelationMethod` from upstream 😄 
https://github.com/pola-rs/polars/blob/451f1234e6f1bb79132a039f825b80daf765263a/py-polars/src/polars/_typing.py#L203


## (2) Alias docstring?
The pattern the other aliases follow is *usually*, duplicating (😔) the parameter docstring.

**But**, for [`polars.corr`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.corr.html), that seems pretty unhelpful:

>     method : {'pearson', 'spearman'}
>        Correlation method.

On the other hand, the `polars` examples show pretty clearly how it changes behavior 
https://github.com/pola-rs/polars/blob/50a3bfbb4f663939a0868907ef1cc51c4288ce05/py-polars/src/polars/functions/lazy.py#L900-L909


### Question
Might there be a benefit to doing something like `IntoDType`?

It has an example of the 2 parts of a union, which here would be like an example for each `Literal` member:


<details><summary><b>Show <code>IntoDType</code></b></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/ada7648b44d1b80ca5e26089ca1ad11ee1a36781/narwhals/typing.py#L279-L305

</p>
</details>

benrutter: Those all sound like really good calls! 1 is pretty unambiguous, but I just want to make sure I've followed on 2, is the idea:
- Give the alias `CorrelationalMethod` (currently `CorrMethod`) a docstring like IntoDType has
- But include in it examples, like with IntoDType, that show a `nw.corr` call with both arguments?

(Asking because I think the other alternative would be to add in two separate examples within the `nw.corr` docstring, so wanted to make sure I'd understood - and yes, please do feel free to say "actually we should have both" 😁)

MarcoGorelli: agree with 1

disagree with 2, it's fine to repeat in the docstring, and we don't need an example for every value

MarcoGorelli:
the downstream failures look unrelated, but the pyspark one

```
FAILED tests/expr_and_series/corr_test.py::test_corr_expr[pyspark-a-a-c-None] - pyspark.errors.exceptions.captured.ArithmeticException: [DIVIDE_BY_ZERO] Division by zero. Use `try_divide` to tolerate divisor being 0 and return NULL instead. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22012
```
is probably solvable?

narwhals/typing.py
benrutter: Pushed a quick update for 1, since that's easy and seems agreed.

@MarcoGorelli and @dangotbanned, I'll let you both fight it out over the docstring changes, but I'm happy to push up a change if needed 😅

benrutter:
> the downstream failures look unrelated, but the pyspark one
> 
> ```
> FAILED tests/expr_and_series/corr_test.py::test_corr_expr[pyspark-a-a-c-None] - pyspark.errors.exceptions.captured.ArithmeticException: [DIVIDE_BY_ZERO] Division by zero. Use `try_divide` to tolerate divisor being 0 and return NULL instead. If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22012
> ```
> 
> is probably solvable?

Before I go off and implement something unhinged, my thinking is:
- I don't think there's a nice way to temporarily for instance set `spark.sql.ansi.enabled` to false and then switch it back
- The `try_divide` call is happening under the hood in pyspark's `corr` definition ([corr docs don't allow any additional arguments](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.corr.html?highlight=corr#pyspark.sql.functions.corr))
- The error, for consistency *should* be None, if any of the values in either column are Null
- Therefore, the best way to handle this error (short of hand rolling the `corr` function) would be to add some "if either columns has a null, return a literal of None"?

Does that sound sensible? In my head it's the most logical fix, but obviously has a performance implementation, so I just wanted to check there wasn't some neato spark trick I could do?

MarcoGorelli:
thanks!

> The try_divide call is happening under the hood in pyspark's corr definition ([corr docs don't allow any additional arguments](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.corr.html?highlight=corr#pyspark.sql.functions.corr))

ah ok - let's leave it like this then, you may just need an `xfail` for the test then

if that's resolved then maybe we can ship it (just do a final check over everything)

benrutter:
Ok, new fun! Looks like after adding in an xfail in that pyspark *works* on None values for older pspark version (https://github.com/narwhals-dev/narwhals/actions/runs/22725009080/job/65897490365?pr=3460).

I've changed that xfail to a skip - wanted to check that was the most idiomatic solution? (other thing I can think of would be to be some kind of version check based xfail)

That aside, I've has a last look through, and I'm at least happy with where everythings at (I did find one more "Spark" typo lurking in DuckDB that I've fixed)

narwhals/typing.py
dangotbanned: > Asking because I think the other alternative would be to add in two separate examples within the nw.corr docstring, so wanted to make sure I'd understood 

You could just add this tweak on the `nw.corr` example, and forget about adding examples to the `TypeAlias`?

```diff
-       >>> nw.from_native(df_native).select(correlation=nw.corr("a", "b"))
+       >>> df = nw.from_native(df_native)
+       >>> df.select(pearson=nw.corr("a", "b"), spearman=nw.corr("a", "b", "spearman"))
```

Would be a pretty brief way to communicate that the two produce different results 🙂

MarcoGorelli:
yup, thanks!

i'll just check about nans. i don't think we need to do anything as we don't give them any special treatment, and the polars default is `propagate_nans = False`

i'm also not sure about [the ddof comment](https://github.com/narwhals-dev/narwhals/pull/3460#discussion_r2852445182), i don't see other libraries mention it, and i'm not sure it's correct? personally i'd just remove it, sorry for conflicting requests here 😳

narwhals/typing.py
benrutter: Sounds sensible enough - I've updated the example to include both.

benrutter:
Ok, updated the docstring to include a pearson and spearman example. @MarcoGorelli I've taken out the ddof mention in the docstring too.

narwhals/_pandas_like/namespace.py
```diff
@@ -373,6 +373,27 @@ def _if_then_else(
...
         return where(when) if otherwise is None else where(when, otherwise)
 
+    def corr(
+        self, a: PandasLikeExpr, b: PandasLikeExpr, *, method: CorrelationMethod
+    ) -> PandasLikeExpr:
+        def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
+            a_series = df._evaluate_single_output_expr(a)
+            b_series = df._evaluate_single_output_expr(b)
+            _df = self._concat_horizontal([a_series.native, b_series.native])
+            corr = _df.corr(method=method).iloc[0, [1]]  # type: ignore[union-attr]
```
camriddell, line 383: Can one just do `a_series.corr(b_series)` with pandas? [Series.corr](https://pandas.pydata.org/docs/reference/api/pandas.Series.corr.html)

Would avoid the temporary creation of a DataFrame

tests/expr_and_series/corr_test.py
```diff
@@ -0,0 +1,61 @@
...
+        else pytest.raises(NotImplementedError)
+    )
+    df = nw.from_native(constructor(data))
+    with context:
+        result = df.select(result=nw.corr("a", "b", method="spearman"))
+        expected = {"result": [0.29617443887954614]}
+        assert_equal_data(result, expected)
+
+
+def test_corr_series(constructor_eager: ConstructorEager) -> None:
```
camriddell: Can this test be the same level of rigor as `test_corr_expr`? Also we should test "spearman" for the series implementation as well.

pyproject.toml
```diff
@@ -284,6 +284,7 @@ filterwarnings = [
 
   "ignore:.*invalid value encountered in cast:RuntimeWarning:pandas",
   "ignore:.*The behavior of DataFrame concatenation with empty or all-NA entries is deprecated:FutureWarning",
+
```
camriddell: delete newline

narwhals/_pandas_like/namespace.py
MarcoGorelli: ah nice!

camriddell:
> yup, thanks!
> 
> i'll just check about nans. i don't think we need to do anything as we don't give them any special treatment, and the polars default is `propagate_nans = False`
> 
> i'm also not sure about [the ddof comment](https://github.com/narwhals-dev/narwhals/pull/3460#discussion_r2852445182), i don't see other libraries mention it, and i'm not sure it's correct? personally i'd just remove it, sorry for conflicting requests here 😳

If I remember my maths correctly, since the correlation coefficient is a division `cov(x, y) / std(x) * std(y)`, the n vs n-1 adjustment is made equally to both the numerator and denominator and should cancel out anyways. This is likely why [numpy.corrcoef](https://numpy.org/devdocs/reference/generated/numpy.corrcoef.html) and [pandas.Series.corr](https://pandas.pydata.org/docs/reference/api/pandas.Series.corr.html) don't expose a `ddof` parameter.

narwhals/_pandas_like/namespace.py
benrutter: Need to figure out some things on this, seems like it *should* be as simple as this:
```python
            import pandas as pd

            a_series = df._evaluate_single_output_expr(a)
            b_series = df._evaluate_single_output_expr(b)
            corr = pd.Series(a_series.native.corr(b_series.native, method=method))
```
but I'm hitting runtime errors in one test now for `invalid value encoutered in divide`.

Will do some digging and try and find what the cause is.

camriddell: The runtime warnings are likely due to your test set up: column `c` has the values `[1, None, 1]`  and the variance of that is 0 which puts 0 on your denominator.

I presume we can ignore this warning in the test, or perhaps more robust is to capture this as an edge-case in addition to, but separately from, the Null value test that column `c` currently is.

camriddell: Actually this is wild, `pandas` has 2 separate paths for correlation coefficient calculation for `DataFrame.corr()` and `Series.corr(…)`. This interestingly explains why `Series.corr` has a "Divide by Zero" runtime warning, and `DataFrame.corr()` did not.

`DataFrame.corr()` calls into `_libs/algos.pyx` which contains pandas [cython(ized) correlation functions](https://github.com/pandas-dev/pandas/blob/e04b26f375035e5106cb913e47b6db612f4ebb11/pandas/core/frame.py#L13119-L13127)  to perform the calculation.

`Series.corr()` calls to into pandas [`nanops` module](https://github.com/pandas-dev/pandas/blob/e04b26f375035e5106cb913e47b6db612f4ebb11/pandas/core/series.py#L2774-L2776) which then delegates person correlation to `numpy.corrcoef` and spearman to `scipy.spearmanr`

---

I would say, let's keep what you had. Do the `concat -> DataFrame.corr().iloc[…]`. Does @MarcoGorelli agree?

benrutter: That is wild! I'd thought it *looked* like that initially, but assumed that couldn't be the case.

benrutter: (interesting implication of this is also that using the series would require pandas optional scipy dependency for spearman)

tests/expr_and_series/corr_test.py
benrutter: Yeah, 100% - I've also introduced rounding into the testing, I think this is a good call because otherwise tests become a bit messy due to very minor floating point differences between implementation, but just wanted to flag in case there was a "approx equals" utility I should have used instead (it didn't seem like this from looking)

camriddell: Can you parameterize both of the `test_corr_series_*` functions to have the same/similar test cases as the above `expr` versions?

narwhals/functions.py
```diff
@@ -1266,6 +1267,39 @@ def _expr_with_horizontal_op(name: str, *exprs: IntoExpr, **kwargs: Any) -> Expr
...
+    Examples:
+        >>> import polars as pl
+        >>> import narwhals as nw
+        >>>
+        >>> df_native = pl.DataFrame({"a": [1, 2, 3], "b": [0, 3, 2]})
+        >>> nw.from_native(df_native).select(
+        >>>     pearson=nw.corr("a", "b", method="pearson),
+        >>>     spearman=pl.corr("a", "b", method="spearman"),
+        >>> )
```
camriddell: A few small issues here:
1. line continuations are prefixed with `...` instead of `>>>`
2. accidental use of polars expression: `spearman=nw.corr(...)` instead of `spearman=pl.corr` 
3. missing a closing quote around `method="pearson"`
4. additionally the repr of the DataFrame is a touch off. This docstring should read:

```python
    Examples:
        >>> import polars as pl
        >>> import narwhals as nw
        >>>
        >>> df_native = pl.DataFrame({"a": [1, 2, 3], "b": [0, 3, 2]})
        >>> nw.from_native(df_native).select(
        ...     pearson=nw.corr("a", "b", method="pearson"),
        ...     spearman=nw.corr("a", "b", method="spearman"),
        ... )
        ┌───────────────────────┐
        |  Narwhals DataFrame   |
        |-----------------------|
        |shape: (1, 2)          |
        |┌──────────┬──────────┐|
        |│ pearson  ┆ spearman │|
        |│ ---      ┆ ---      │|
        |│ f64      ┆ f64      │|
        |╞══════════╪══════════╡|
        |│ 0.654654 ┆ 0.5      │|
        |└──────────┴──────────┘|
        └───────────────────────┘

```

camriddell:
@benrutter thanks so much for the updates! It seems like we're just about wrapped up on this one, I've commented above on a few minor things to help push CI into the green.

- [ ] add more test cases for the `test_*_series` functions
- [ ] fix the `corr` docstring

narwhals/functions.py
benrutter: Nice spot! Updated

benrutter:
> @benrutter thanks so much for the updates! It seems like we're just about wrapped up on this one, I've commented above on a few minor things to help push CI into the green.
> 
>     * [x]  add more test cases for the `test_*_series` functions
> 
>     * [x]  fix the `corr` docstring

No problem at all! Thanks for taking the time to look over everything.

I've added additional tests cases for the test series functions - essentially the additional ones from expressions (minus the expression specific tests that wouldn't make sense). Now there's parametrize in place, should be easy enough to add more as needed, let me know if there's anything specific you think should be being tested that isn't.

Plus I've corrected the docstring. I'd initially written it out by hand which was probably not a very good call 😅

tests/expr_and_series/corr_test.py
```diff
@@ -0,0 +1,118 @@
...
+@pytest.mark.parametrize(
+    ("output_name", "a", "b", "expected_corr"),
+    [("a", "a", "b", 0.87), ("a", "a", "c", None)],
+)
+def test_corr_series(
+    constructor_eager: ConstructorEager,
+    request: pytest.FixtureRequest,
+    output_name: str,
+    a: str | nw.Expr,
+    b: str | nw.Expr,
```
camriddell: I believe the typing errors we see are due to the annotations of `a` and `b` here. If you change it from

```python
    a: str | nw.Expr,
    b: str | nw.Expr,
```

to

```python
    a: str,
    b: str,
```

we should be on a good path to pass the typing test.

tests/expr_and_series/corr_test.py
```diff
@@ -0,0 +1,118 @@
...
+@pytest.mark.parametrize(
+    ("output_name", "a", "b", "expected_corr"),
+    [("a", "a", "b", 0.87), ("a", "a", "c", None)],
+)
+def test_corr_series_spearman(
+    constructor_eager: ConstructorEager,
+    request: pytest.FixtureRequest,
+    output_name: str,
+    a: str | nw.Expr,
+    b: str | nw.Expr,
```
camriddell: same comment here as above (note that this change is only relevant for the `*_series` tests).

tests/expr_and_series/corr_test.py
benrutter: Woops! Sorry by bad 🙈 pushed a fix now

camriddell:
Approved as well, thank you so much @benrutter!

PR #3462, opened 2026-02-06
ci: Use `--clear` in `uv venv` for 0.10.0

dangotbanned:
# Description

Since (https://github.com/astral-sh/uv/releases/tag/0.10.0), `uv` now has this change:

> ### Breaking changes
> 
> - Require `--clear` to remove existing virtual environments in `uv venv` ([#17757](https://github.com/astral-sh/uv/pull/17757))
> 
>     Previously, `uv venv` would prompt for confirmation before removing an existing virtual environment in interactive contexts, and remove it without confirmation in non-interactive contexts. 
>     Now, `uv venv` requires the `--clear` flag to remove an existing virtual environment. A warning for this change was added in [uv 0.8](https://github.com/astral-sh/uv/blob/main/changelogs/0.8.x.md#breaking-changes).
> 
>     You can opt out of this behavior by passing the `--clear` flag or setting `UV_VENV_CLEAR=1`.

Apparently only broke the `typing.yml` workflow, so good news overall I guess 

## Related issues

- https://github.com/narwhals-dev/narwhals/actions/runs/21746636942/job/62734017631?pr=3454
- https://github.com/astral-sh/uv/pull/17757

### DIFF
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -129,7 +129,7 @@ def test_concat_diagonal_invalid(
         )
     )
     context: Any
-    if impl.is_polars() and POLARS_VERSION < (1, 1):  # pragma: no cover
+    if impl.is_polars() and POLARS_VERSION < (1, 18):  # pragma: no cover
         context = pytest.raises(
             NarwhalsError,
             match=re.compile(r"(int.+datetime)|(datetime.+int)", re.IGNORECASE),


MarcoGorelli:
@dangotbanned for the future, can we use the `fix` label for user-facing fixes? else it ends up in the "bug fixes" part of the release notes, even though users don't need to know about it. i'd suggest the "internal" label for this one

dangotbanned:
> can we use the `fix` label for user-facing fixes? else it ends up in the "bug fixes" part of the release notes

@MarcoGorelli yeah sure thing! 

Would it be worth adding something to the label description for those that have a special meaning?

**Edit**: jfc I didn't realise how big that screenshot was 😭 (shrunk it 4x now)

Screenshot_20260223-104211 Firefox

PR #3475, opened 2026-02-11
chore: move tests to a narwhals-testing package

tokoko:
# Description
PR moves tests to a new package `narwhals-testing`. The goal is to make it easier for plugins that live outside the main repo to easily hook into the tests defined here.

I've yet to test that the change will indeed work on the plugin side, but wanted to first get the feedback here whether this seems like a right direction to you.


MarcoGorelli:
thanks for the pr - sure, seems like a good idea

perhaps it's worth testing out in https://github.com/narwhals-dev/narwhals-daft to see if/how it would work there?

PR #3479, opened 2026-02-19
chore(python): avoid using `type(self)` in protocols

MarcoGorelli:
based on https://github.com/facebook/pyrefly/issues/2420#issuecomment-3922431892

@dangotbanned is this ok as a soln?

### DIFF
--- a/narwhals/_compliant/expr.py
+++ b/narwhals/_compliant/expr.py
@@ -431,7 +431,7 @@ class EagerExpr(
                 results.append(result)
             return results
 
-        return type(self)(
+        return self.__class__(
             func,
             evaluate_output_names=self._evaluate_output_names,
             alias_output_names=self._alias_output_names,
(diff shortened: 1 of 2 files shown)


dangotbanned:
I could be wrong, but I'm not sure `self.__class__` is defined in the [typing spec](https://typing.python.org/en/latest/spec/).


If that's the case, then I would avoid using it since there are no guarantees each type checker will implement it the same way - and whatever they have now is fair game to break.

I guess that's probably a general thing **we** need to keep in mind now that `pyrefly` and `ty` are joining the party.

---

Compared to what I suggested in (https://github.com/facebook/pyrefly/issues/2420#issuecomment-3920684279), there is a [**lot** of specification](https://typing.python.org/en/latest/spec/constructors.html) which covers `__new__`.
The rest is just attribute assignability

MarcoGorelli:
yeah but then we need to keep this in sync with what happens in `__init__`, which seems messier / harder to maintain?

PR #3487, opened 2026-02-28
Update marimo downstream CI to use uv instead of hatch

manzt:
Pending on https://github.com/marimo-team/marimo/pull/8510, which replaces hatch with native uv commands in marimo. Uses `uv add ../ --editable` and a new `scripts/test-narwhals.sh` in marimo instead of the old hatch-based test invocation.

### DIFF
--- a/.github/workflows/downstream_tests.yml
+++ b/.github/workflows/downstream_tests.yml
@@ -104,7 +104,7 @@ jobs:
         if: ${{ matrix.dependencies == 'core,optional' }}
         run: |
           cd marimo
-          uv run --python ${{ matrix.python-version }} -c "import narwhals; print(narwhals.__file__)"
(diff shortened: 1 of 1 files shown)


.github/workflows/downstream_tests.yml
```diff
@@ -78,35 +78,22 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: clone-marimo
         run: |
-          git clone https://github.com/marimo-team/marimo.git --depth=1
+          # TODO: remove branch pin once https://github.com/marimo-team/marimo/pull/8510 is merged
```
manzt, line 81: Just flagging so this doesn't get merged before

MarcoGorelli (approved):
thanks so much @manzt !

manzt:
I had a typo in the CI. Sorry about that!

PR #3495, opened 2026-03-03
RFC fix (typing): Remove lazyframe-specific matching

### LINKED ISSUE #3493: [Doc]: Linting the example from the README gives an error when using `ty`.

Correction

Recall the example from the README:

```python
import narwhals as nw


def agnostic_function(
    df_native: IntoFrameT,
    date_column: str,
    price_column: str,
) -> IntoFrameT:
    return (
        nw.from_native(df_native)
        .sort(date_column)
        .to_native()
    )
```

When I check this with `ty` via `uvx -w narwhals ty check tmp1.py`, I get:

```
uvx -w narwhals ty check tmp1.py
error[unresolved-attribute]: Attribute `group_by` is not defined on `Series[Unknown]` in union `DataFrame[Unknown] | Series[Unknown]`
  --> tmp1.py:11:9
   |
 9 |   ) -> IntoFrameT:
10 |       return (
11 | /         nw.from_native(df_native)
12 | |         .group_by(nw.col(date_column).dt.truncate("1mo"))
   | |_________________^
13 |           .agg(nw.col(price_column).mean())
14 |           .sort(date_column)
   |
info: rule `unresolved-attribute` is enabled by default

Found 1 diagnostic
```

Whereas if I rewrite this by replacing `IntoFrameT ` with `IntoDataFrameT`, all checks pass.

```python
import narwhals as nw


def agnostic_function(
    df_native: IntoDataFrameT,
    date_column: str,
    price_column: str,
) -> IntoDataFrameT:
    return (
        nw.from_native(df_native)
        .sort(date_column)
        .to_native()
    )
```

No issues here:
```
uvx -w narwhals ty check tmp2.py
All checks passed!
```

I'd expect the example in the README to pass a check from `ty`. I am not sure if this is as simple as just swapping `IntoFrameT` with `IntoDataFrameT` in the README, of if this is deeper, underlying problem.

MarcoGorelli:
`nw.LazyFrame` can be backed by any native frame, not just the ones that are truly lazy. For example:
```py
In [3]: nw.from_native(pd.DataFrame()).lazy()
Out[3]:
┌──────────────────┐
|Narwhals LazyFrame|
|------------------|
| Empty DataFrame  |
| Columns: []      |
| Index: []        |
└──────────────────┘
```

So I think the current `LazyFrame` is too restrictive

What's important is that
```py
nw.from_native(duckdb_rel)
```
is that detected to be `nw.LazyFrame[DuckDBPyRelation]`, and that indeed is still the case - the reason is that the overload for `DataFrame` (the eager one) is placed before the more general `Frame` one

this also closes #3493 (both mypy and ty pass now)

There are some other issues which do come up, will investigate

### DIFF
--- a/narwhals/_native.py
+++ b/narwhals/_native.py
@@ -105,7 +105,7 @@ They differ by checking **all** native types/protocols in a single-call and usin
 from __future__ import annotations
 
 from collections.abc import Callable, Collection, Iterable, Sized
-from typing import TYPE_CHECKING, Any, Protocol, TypeVar, Union, cast
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar, cast
 
 from narwhals.dependencies import (
     get_cudf,
@@ -282,7 +282,7 @@ NativePySpark: TypeAlias = _PySparkDataFrame
 NativePySparkConnect: TypeAlias = _PySparkDataFrame
 NativeSparkLike: TypeAlias = "NativeSQLFrame | NativePySpark | NativePySparkConnect"
 NativeKnown: TypeAlias = "NativePolars | NativeArrow | NativePandasLike | NativeSparkLike | NativeDuckDB | NativeDask | NativeIbis"
-NativeUnknown: TypeAlias = "NativeDataFrame | NativeSeries | NativeFrame"
+NativeUnknown: TypeAlias = "NativeSeries | NativeFrame"
 NativeAny: TypeAlias = "NativeKnown | NativeUnknown"
 
 IntoDataFrame: TypeAlias = NativeDataFrame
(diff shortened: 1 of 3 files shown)


tests/frame/unique_test.py
```diff
@@ -56,8 +56,8 @@ def test_unique_first_last(
     assert_equal_data(result, expected)
 
     if isinstance(df, nw.DataFrame):
-        result = df.unique("b", keep=keep, order_by="i", maintain_order=True)
-        assert_equal_data(result, expected)
+        result = df.unique("b", keep=keep, order_by="i", maintain_order=True)  # pyright: ignore[reportCallIssue]
+        assert_equal_data(result, expected)  # pyright: ignore[reportCallIssue]
```
MarcoGorelli, line 60: this is an issue, because now pyright thinks `df` can only be lazyframe

PR #3496, opened 2026-03-03
tie `IntoFrameT` to `NativeFrame`

### LINKED ISSUE #3319: [Doc]: Linting is unhappy with IntoFrameT

Correction

My linter is not happy when I use the combined typeVar IntoFrameT for both DataFrame and LazyFrame with from_native()

It seems it's only happy with me using the IntoDataFrameT or IntoLazyFrameT and _not the_ IntoFrameT notation.

Maybe I'm using it wrong. But from the docs it seems like functions where I handle both Lazy and Eager cases, should be IntoFrameT.

Using narwhals 2.12.0.
Happy:

Image

Un-happy:

Image

_No response_

### LINKED ISSUE #3493: [Doc]: Linting the example from the README gives an error when using `ty`.

MarcoGorelli:
closes #3493 
closes #3319

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes


src/narwhals/translate.py
```diff
@@ -145,6 +144,10 @@ def from_native(
     native_object: IntoLazyFrameT, **kwds: Unpack[AllowLazy]
 ) -> LazyFrame[IntoLazyFrameT]: ...
 @overload
+def from_native(  # type: ignore[overload-overlap]
```
dangotbanned, line 153: ?

narwhals/translate.py
```diff
@@ -164,7 +167,7 @@ def from_native(
     series_only: bool,
     allow_series: bool | None,
 ) -> Any: ...
-def from_native(
+def from_native(  # pyright: ignore[reportInconsistentOverload]
```
dangotbanned: ?

dangotbanned (changes requested):
Thanks for the ping @MarcoGorelli

## General
> 🤔 might this be a solution to #3493 ?

I'd like to see more typing tests added to ensure this is behaving as we want.
I raised this before (https://github.com/narwhals-dev/narwhals/pull/3125#discussion_r2549531521)


## New Ignores
I was concerned by the new ignores (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r2904934046, https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r2904934602)

They made me think to check if this was possible:


```py
import polars as pl
from typing_extensions import assert_type, reveal_type

import narwhals as nw

lf = pl.LazyFrame()
df = pl.DataFrame()
either = lf if df.height else df

r_lf = nw.from_native(lf)
r_df = nw.from_native(df)
r_either = nw.from_native(either)

r2_lf = nw.from_native(r_lf)
r2_df = nw.from_native(r_df)
r2_either = nw.from_native(r_either)
```

These are all good 👍
```py
assert_type(r_lf, nw.LazyFrame[pl.LazyFrame])
assert_type(r_df, nw.DataFrame[pl.DataFrame])
assert_type(r2_lf, nw.LazyFrame[pl.LazyFrame])
assert_type(r2_df, nw.DataFrame[pl.DataFrame])
```

This one demonstrates the overlap issue 👎
```py
reveal_type(r_either)
```

> [!NOTE]
> Had to format the output with ruff as it got silly 😂 

```py
Type of "r_either" is

(
    narwhals.dataframe.DataFrame[
        polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
    ]
    | narwhals.dataframe.LazyFrame[
        polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
    ]
)
```

And this shows how that leads to an explosion 👎
```py
reveal_type(r2_either)
```
```py
Type of "r2_either" is

(
    narwhals.dataframe.DataFrame[
        narwhals.dataframe.DataFrame[
            polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
        ]
        | narwhals.dataframe.LazyFrame[
            polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
        ]
    ]
    | narwhals.dataframe.LazyFrame[
        narwhals.dataframe.DataFrame[
            polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
        ]
        | narwhals.dataframe.LazyFrame[
            polars.lazyframe.frame.LazyFrame | polars.dataframe.frame.DataFrame
        ]
    ]
)
```

## Possible test ideas
(Beyond just trying to get 0 errors on `mypy` for `README.md` 😅)

What happens when we use incompatible parts of the `*Frame` APIs?:

- `collect(backend)`
- `lazy(backend)`
- `with_row_index()`
- `join(other=<same|wrong-type|the-result-of-a-union>)`
- `unique(maintain_order=True)`
- `sink_parquet(...)`

How does this interact with `concat`?

## Related
- #2239
- #2356
- #2944

src/narwhals/translate.py
MarcoGorelli: 🤷 pyright's fine with it but mypy outputs
```
narwhals/translate.py:147: error: Overloaded function signatures 9 and 10 overlap with incompatible return types  [overload-overlap]
    def from_native(
    ^
narwhals/translate.py:147: error: Overloaded function signatures 9 and 11 overlap with incompatible return types  [overload-overlap]
    def from_native(
```

narwhals/translate.py
MarcoGorelli: pyright gives
```
  /home/marcogorelli/polars-api-compat-dev/narwhals/translate.py:170:5 - error: Overloaded implementation is not consistent with signature of overload 9
    Function return type "DataFrame[IntoFrameT@from_native] | LazyFrame[IntoFrameT@from_native]" is incompatible with type "LazyFrame[IntoLazyFrameT@from_native] | DataFrame[IntoDataFrameT@from_native] | Series[IntoSeriesT@from_native] | IntoFrameT@from_native"
      Type "DataFrame[IntoFrameT@from_native] | LazyFrame[IntoFrameT@from_native]" is not assignable to type "LazyFrame[IntoLazyFrameT@from_native] | DataFrame[IntoDataFrameT@from_native] | Series[IntoSeriesT@from_native] | IntoFrameT@from_native"
        Type "DataFrame[IntoFrameT@from_native]" is not assignable to type "LazyFrame[IntoLazyFrameT@from_native] | DataFrame[IntoDataFrameT@from_native] | Series[IntoSeriesT@from_native] | IntoFrameT@from_native"
          Type "DataFrame[IntoFrameT@from_native]" is not assignable to type "IntoFrameT@from_native"
          "DataFrame[IntoFrameT@from_native]" is not assignable to "DataFrame[IntoDataFrameT@from_native]"
            Type parameter "FrameT@DataFrame" is invariant, but "IntoFrameT@from_native" is not the same as "IntoDataFrameT@from_native"
          "DataFrame[IntoFrameT@from_native]" is not assignable to "LazyFrame[IntoLazyFrameT@from_native]"
          "DataFrame[IntoFrameT@from_native]" is not assignable to "Series[IntoSeriesT@from_native]" (reportInconsistentOverload)
1 error, 0 warnings, 0 informations
make: *** [Makefile:25: typing] Error 1
```

given that this part isn't user-facing (the overloads are used) i didn't it was too big of a deal

MarcoGorelli:
> This one demonstrates the overlap issue 👎

yup, thanks for taking a look!

MarcoGorelli:
> This one demonstrates the overlap issue 👎

Thanks! I think this is addressible with some extra specialised overloads, have pushed

> What happens when we use incompatible parts of the *Frame APIs?:

Aren't these generally tested already? Any particular tests you'd like to see?

akmalsoliev:
Hey folks, any update on this one? Drove me mad for 30 minutes till I realized it is pyrefly specific error

MarcoGorelli:
I don't think it's pyrefly-specific, pyright flags it too

to respond to Dan's comment point by point:

> - `collect(backend)`

If you do collect, then you can't hope to have a `IntoFrameT -> IntoFrameT` function anyway

> - lazy(backend)

same

> - with_row_index()

This isn't an incompatible API, it returns `Self`, it's the same as using `select` or `filter`

> - join(other=<same|wrong-type|the-result-of-a-union>)

Yup, this one's problematic (as it currently is on `main`). It's not made worse as far as I can tell

Something I wouldn't be opposed to would be to change the `LazyFrame.lazy` API to be
```py
def lazy(self) -> LazyFrame[Any]:
```
as, realistically, if you're calling .`lazy` on a lazyframe, it's because you don't know if it's lazy or eager

> - unique(maintain_order=True)

Also not an incompatible API, it returns `Self`

> - sink_parquet(...)

This returns `None`, so I'm not sure what there is to test here

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4183152973

Hey @MarcoGorelli just to clarify these bits:

> This isn't an incompatible API, it returns `Self`, it's the same as using `select` or `filter`

By *incompatible API*, I mean that the signature (specifically the parameters) are not compatible.

For the `LazyFrame` cases, they have required arguments (IIRC).
So the problem I was anticipating is that if we matched `IntoFrameT` - would we still warn if you didn't always provide the required arguments?

I expect we currently would warn - but I wasn't sure if `IntoFrameT` may let everything pass through, if that makes sense?

`sink_parquet` would be a method that shouldn't be accessible on `DataFrame` - so that's a slightly different version of the problem.
But the general idea is, are we introducing false positives?

Sorry that I haven't got back to this PR yet!

dangotbanned:
Just wanted to categorise the *potential* for issues from (https://github.com/narwhals-dev/narwhals/pull/3496#pullrequestreview-3914859960).

### Method exists for `LazyFrame`-only
-  `collect(backend)`
- `sink_parquet(...)`


### Argument exists for `DataFrame`-only
- `lazy(backend)` vs `lazy()`
- `unique(maintain_order=...)` vs `unique()`

### Required argument for `LazyFrame`-only
- `with_row_index(order_by=...)`

### `Self`
If both `LazyFrame` and `DataFrame` have the same `TypeVar` bound, this would be the most visible place to see the overlap issue 
- `join(other=<same|wrong-type|the-result-of-a-union>)`

MarcoGorelli:
sure, thanks! have added some tests for these functions, looks expected to me

src/narwhals/translate.py
```diff
@@ -129,6 +128,10 @@ def from_native(
 @overload
 def from_native(native_object: LazyFrameT, **kwds: Unpack[AllowLazy]) -> LazyFrameT: ...
 @overload
+def from_native(
+    native_object: LazyFrameT | DataFrameT, **kwds: Unpack[AllowLazy]
+) -> LazyFrameT | DataFrameT: ...
+@overload
```
dangotbanned, line 136: If this is now needed, does it mean that this is now a valid annotation?

`nw.DataFrame[nw.DataFrame[Any] | nw.LazyFrame[Any]]`

```suggestion
```

src/narwhals/translate.py
dangotbanned: I'm not feeling convinced by this or (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r2913206305)
AFAICT, both are the same issue reported by two different type checkers.

If you remove all the ignore comments, does `pyrefly` not report these as an issue as well?
For `ty`, we'd need (https://github.com/astral-sh/ty/issues/103) to get it reported - so probably nothing there yet

tests/translate/from_native_test.py
```diff
@@ -775,3 +777,59 @@ def test_from_native_series_exhaustive() -> None:  # noqa: PLR0914, PLR0915
...
+            .agg(nw.col(price_column).mean())
+            .sort(date_column)
+            .to_native()
+        )
+
+
+def test_from_eager_or_lazy_polars() -> None:
+    pytest.importorskip("polars")
+    import polars as pl
+
```
dangotbanned: **I wrote this so it is my bad btw**
There are *potentially* two problems with this test.


### 1 - `assert_type`
The original issue was with (at least) `mypy` (https://github.com/narwhals-dev/narwhals/issues/3493#issuecomment-3983681398), but `assert_type` is disabled for `mypy` in this file.

*Why, is explained in the module doc*

https://github.com/narwhals-dev/narwhals/blob/7e3fccaf9c27e0d85dac6d30bbf77a79f5f3d6a4/tests/translate/from_native_test.py#L22-L23

Would we want to move these new tests to another module, where `mypy` is fully enabled?

### 2 - Is `mypy` checking at all?
I would certainly hope so 😅 
I've been caught out a few times before by [no-errors-reported-for-obviously-wrong-code](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code).

The `-> None` *should* be enough, but if not then enabling [`--check-untyped-defs`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs) *may* reveal more

tests/translate/from_native_test.py
```diff
@@ -775,3 +777,59 @@ def test_from_native_series_exhaustive() -> None:  # noqa: PLR0914, PLR0915
...
+    def _agnostic_function(  # pragma: no cover
+        df_native: IntoFrameT, date_column: str, price_column: str
+    ) -> IntoFrameT:
+        return (
+            nw.from_native(df_native)
+            .group_by(nw.col(date_column).dt.truncate("1mo"))
+            .agg(nw.col(price_column).mean())
+            .sort(date_column)
+            .to_native()
+        )
```
dangotbanned: > check that readme example (as of March 2026) passes

While this is one in the positive column 🎉, I'd like to know what is being inferred at some of the intermediate steps.
Or I guess, does it match our expectations?

`test_from_eager_or_lazy_polars` covers a known type `pl.DataFrame | pl.LazyFrame`, whereas this is for `IntoFrameT`.

Totally fine to keep this chained version, but could you add another with an `assert_type` after each call?

tests/translate/from_native_test.py
```diff
@@ -775,3 +777,59 @@ def test_from_native_series_exhaustive() -> None:  # noqa: PLR0914, PLR0915
...
+
+
+def test_into_frame_t_incompatible_apis() -> None:
+    def _agnostic_function(  # pragma: no cover
+        df_native: IntoFrameT,
+    ) -> IntoFrameT:
+        nw.from_native(df_native).sink_parquet("...")  # type: ignore[union-attr]
+        _ = (
+            nw.from_native(df_native)
+            .with_row_index()  # type: ignore[call-arg]
```
dangotbanned: Thanks for adding these - very good to see 😄

narwhals/translate.py
dangotbanned: > given that this part isn't user-facing (the overloads are used) i didn't it was too big of a deal

How each type checker deals will overlapping overloads is a bit hairy.
The typing spec says this about solving overloads [at the callsite](https://typing.python.org/en/latest/spec/overload.html#step-5)

> If the return types are not equivalent, overload matching is ambiguous. 
> In this case, assume a return type of `Any` and stop.

AFAICT, there isn't a spec for *overlapping definitions* 😢 
So the conformance tests focus on [just the usage part](https://github.com/python/typing/blob/ca2a7dbd437f80c6ee4fbcf9899f6bd209cf5447/conformance/tests/overloads_evaluation.py).

### This error

Both `mypy` and `pyright` are trying to tell you that:

> As a parameter, `IntoFrameT` matches:
> > `IntoDataFrameT | IntoLazyFrameT`
> > `IntoFrameT`
> 
> But the return types do not match:
> > `DataFrame[IntoDataFrameT] | LazyFrame[IntoLazyFrameT]`
> > `DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]`


- I think `mypy` does the `Any` thing?
- I'm more sure that `pyright` would synthesize a signature with a `*` in it for inference
- (I'd assume) `ty` uses intersections


> [!IMPORTANT]
> The user facing problems I've experienced is that:
> 1. it makes type checkers run much slower, since they need to track 2 or more overloads that matched
> 2. this can make a language server produce flaky errors, since each run might take a different route
> 
> I left a screen capture in (https://github.com/zen-xu/pyarrow-stubs/pull/208#issuecomment-2820845169) with an extreme example or how slow it can get 😓

narwhals/dataframe.py
```diff
@@ -3254,7 +3254,7 @@ def join_asof(
             suffix=suffix,
         )
 
-    def lazy(self) -> Self:
+    def lazy(self) -> LazyFrame[Any]:
```
dangotbanned:
```suggestion
```

Ummmm, why would this *not* be `Self`?

```py
    return self
```

Introducing an `Any` is pretty suspicious for a PR that solves a typing issue 🤔

dangotbanned (changes requested):
Thanks Marco, I'm still feeling a bit skeptical about this.
I'm happier than I was following (https://github.com/narwhals-dev/narwhals/pull/3496#pullrequestreview-3914859960), and (https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4183208230) - so I wanted to highlight that ❤️ 


I would feel more confident if the PR didn't introduce any of:
- `Any`
- new type ignores

It's okay to disagree with a type checker (I often do), but I think documenting the issue - and why it is okay to ignore - is very helpful for future readers that may solve it


count narwhal

narwhals/dataframe.py
MarcoGorelli: this was for the join issue in https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4183152973 , but i didn't intend to commit it and we can keep it to a separate pr, have reverted

src/narwhals/translate.py
MarcoGorelli: without it, we'd get that `r2_either` in https://github.com/narwhals-dev/narwhals/pull/3496#pullrequestreview-3914859960 is
```
  /home/marcogorelli/polars-api-compat-dev/tests/translate/from_native_test.py:818:21 - error: "assert_type" mismatch: expected "narwhals.dataframe.DataFrame[polars.dataframe.frame.DataFrame] | narwhals.dataframe.LazyFrame[polars.lazyframe.frame.LazyFrame]" but received "narwhals.dataframe.DataFrame[narwhals.dataframe.DataFrame[polars.dataframe.frame.DataFrame] | narwhals.dataframe.LazyFrame[polars.lazyframe.frame.LazyFrame]] | narwhals.dataframe.LazyFrame[narwhals.dataframe.DataFrame[polars.dataframe.frame.DataFrame] | narwhals.dataframe.LazyFrame[polars.lazyframe.frame.LazyFrame]]" (reportAssertTypeFailure)
```

tests/translate/from_native_test.py
MarcoGorelli: sure, i've moved these to a new test file (`tests/translate/into_frame_t_test.py`)

tests/translate/from_native_test.py
MarcoGorelli: sure, added

MarcoGorelli:
thanks @dangotbanned , have addressed most comments

Regarding the overlap, don't type checkers usually match the first valid overload?

At least, for
```py
from typing import overload, Sequence, reveal_type

@overload
def foo(a: str) -> int: ...  # type: ignore[overload-overlap]

@overload
def foo(a: Sequence[str]) -> str: ...

def foo(a: str | Sequence[str]) -> int | str:
    if isinstance(a, str):
        return 0
    return 'zero'

reveal_type(foo('abc'))
reveal_type(foo(['abc']))
```

All of mypy, pyright, pyrefly, ty, and zuban, do exactly that. pandas and numpy both rely on this in their stubs, so I think this is unlikely to change?

```console
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ mypy t.py
t.py:14: note: Revealed type is "int"
t.py:15: note: Revealed type is "str"
Success: no issues found in 1 source file
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ pyright t.py
/home/marcogorelli/scratch/t.py
  /home/marcogorelli/scratch/t.py:14:13 - information: Type of "foo('abc')" is "int"
  /home/marcogorelli/scratch/t.py:15:13 - information: Type of "foo(['abc'])" is "str"
0 errors, 0 warnings, 2 informations
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ pyrefly check t.py
 INFO revealed type: int [reveal-type]
  --> t.py:14:12
   |
14 | reveal_type(foo('abc'))
   |            ------------
   |
 INFO revealed type: str [reveal-type]
  --> t.py:15:12
   |
15 | reveal_type(foo(['abc']))
   |            --------------
   |
 INFO 0 errors
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ ty check t.py
info[revealed-type]: Revealed type
  --> t.py:14:13
   |
12 |     return 'zero'
13 |
14 | reveal_type(foo('abc'))
   |             ^^^^^^^^^^ `int`
15 | reveal_type(foo(['abc']))
   |

info[revealed-type]: Revealed type
  --> t.py:15:13
   |
14 | reveal_type(foo('abc'))
15 | reveal_type(foo(['abc']))
   |             ^^^^^^^^^^^^ `str`
   |

Found 2 diagnostics
(scratch) marcogorelli@DESKTOP-U8OKFP3:~/scratch$ zuban check t.py
t.py:14: note: Revealed type is "builtins.int"
t.py:15: note: Revealed type is "builtins.str"
Success: no issues found in 1 source file
```

MarcoGorelli:
just checking if you still disagree with the approach of whether there can be a way forwards here?

dangotbanned:
Hey Marco, I haven't got back to this yet - so this isn't a comment on the most recent changes - but appreciate you taking things on board 😊

> just checking if you still disagree with the approach 

In my last review I didn't mean to give off that impression, sorry for that 😔
It was more that I felt there were some unknowns and worried that not having the full picture now would mean we'd here again soon, fixing more `from_native` edge cases again.

That isn't to say my worries are correct, and I would much rather your changes turn out to be fine 🤞

I'll do my best to go through your changes this weekend, if that's okay?

---

One thing for now, re overloads (https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4192222503)

I was thinking about this PR a couple of days ago when I saw this potential spec change:

- [ ] https://github.com/python/typing/pull/2250

*I think* the 2nd spec change there is related - might be worth reading to see if (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r3035465450) would be handled differently?

MarcoGorelli:
thanks! yeah no hurry at all, just checking if there was a potential way forwards or if you were firmly against it

btw, i think we're already relying on inconsistent overlapping overloads here

https://github.com/narwhals-dev/narwhals/blob/25bdb8f236bf844d80e44cba1782c2ea398eb5b9/narwhals/dataframe.py#L1049-L1057

dangotbanned:
> Regarding the overlap, don't type checkers usually match the **first valid overload**?
> ...
> All of mypy, pyright, pyrefly, ty, and zuban, do exactly that. pandas and numpy both rely on this in their stubs, so **I think this is unlikely to change?**

I read through (https://github.com/python/typing/pull/2250) and it looks like that would be non-conformant behavior:

- [`pyright`](https://github.com/rchen152/typing/blob/137e3fcbcf112f1514e1d2eb8c236d67b78125b1/conformance/results/pyright/overloads_evaluation.toml#L1-L5)
- [`ty`](https://github.com/rchen152/typing/blob/137e3fcbcf112f1514e1d2eb8c236d67b78125b1/conformance/results/ty/overloads_evaluation.toml#L1-L4)
- [`zuban`](https://github.com/rchen152/typing/blob/137e3fcbcf112f1514e1d2eb8c236d67b78125b1/conformance/results/zuban/overloads_evaluation.toml#L4)

It's worth noting that this is being proposed by a `pyrefly` maintainer.
But there looks to have been support from others in (https://github.com/python/typing/issues/566#issuecomment-4202338146) and [Spec change: ambiguous arguments in overload call evaluation](https://discuss.python.org/t/spec-change-ambiguous-arguments-in-overload-call-evaluation/106891)

narwhals/translate.py
```diff
@@ -145,10 +148,18 @@ def from_native(
...
     native_object: IntoDataFrameT | IntoSeriesT, **kwds: Unpack[AllowSeries]
 ) -> DataFrame[IntoDataFrameT] | Series[IntoSeriesT]: ...
 @overload
+def from_native(
+    native_object: IntoDataFrameT | IntoLazyFrameT, **kwds: Unpack[AllowLazy]
+) -> DataFrame[IntoDataFrameT] | LazyFrame[IntoLazyFrameT]: ...
+@overload
+def from_native(  # type: ignore[overload-overlap]
+    native_object: IntoFrameT, **kwds: Unpack[AllowLazy]
+) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
```
dangotbanned, line 161: I don't know for sure if it'll work, but one way to make the overload defs non-overlapping *might* be:

```py
@overload
def from_native(native_object: IntoFrameT) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
```

Since `AllowLazy` is using `total=False`, that might not be enough though 🤔

MarcoGorelli:
> I read through (https://github.com/python/typing/pull/2250) and it looks like that would be non-conformant behavior:

I'll ask, but my understanding is that it wouldn't be non-conformant because if you have
```py
@overload
def foo(a: str) -> int: ...  # type: ignore[overload-overlap]

@overload
def foo(a: Sequence[str]) -> str: ...

def foo(a: str | Sequence[str]) -> int | str:
    ...

foo('foo')
```
then all materialisations of the `a` argument match the first overload, and so the remaining overloads can be eliminated

MarcoGorelli:
To tie this back specifically to the spec, in step 5, it says

> For all arguments, determine whether all possible [materializations](https://typing.python.org/en/latest/spec/glossary.html#term-materialize) of the argument’s type are assignable to the corresponding parameter type for each of the remaining overloads. If so, eliminate all of the subsequent remaining overloads.

In this case, `a='foo'` is assignable to `a: str` and so the remaining overloads can be ignored

My reading of the linked issues (https://github.com/python/typing/issues/566#issuecomment-4202338146) and [Spec change: ambiguous arguments in overload call evaluation](https://discuss.python.org/t/spec-change-ambiguous-arguments-in-overload-call-evaluation/106891) is that they're about different issues

dangotbanned:
> My reading of the linked issues ([python/typing#566 (comment)](https://github.com/python/typing/issues/566#issuecomment-4202338146)) and [Spec change: ambiguous arguments in overload call evaluation](https://discuss.python.org/t/spec-change-ambiguous-arguments-in-overload-call-evaluation/106891) is that they're about different issues

I think this is the same issue
- https://github.com/rchen152/typing/blob/137e3fcbcf112f1514e1d2eb8c236d67b78125b1/conformance/tests/overloads_evaluation.py#L442-L462

Which `pyright` currently infers to the first overload, and this is what I'm trying to say is something we shouldn't rely on because it is unspecified and soon to be specified as non-conformant behavior for a type checker to do


```py
> # Step 5 eliminates the first overload because there exists a
> # materialization of `A[Any]` that is not assignable to `A[None]`. Step 6
> # picks the second overload.
```

<details><summary>Show pre-3.12 version</summary>
<p>

```py

from typing import TypeVar, Generic, Any, assert_type, overload

T = TypeVar("T")


class A(Generic[T]):
    x: T

    def f(self) -> T:
        return self.x


@overload
def example12(x: A[None]) -> A[None]: ...


@overload
def example12(x: A[Any]) -> A[Any]: ...


def example12(x: A[Any]) -> A[Any]:
    return x


def check_example12(x: Any):
    # Step 5 eliminates the first overload because there exists a
    # materialization of `A[Any]` that is not assignable to `A[None]`. Step 6
    # picks the second overload.
    ret = example12(x)
    assert_type(ret, A[Any])  # "assert_type" mismatch: expected "A[Any]" but received "A[None] "Pylance(reportAssertTypeFailure)"
```

</p>
</details>

MarcoGorelli:
That's dealing with a case where an argument of type `Any` is passed, which has a materialisation which doesn't match the first overload

In the example I gave, we're passing a string argument, in which every materialisation matches the first overload
```
@overload
def foo(a: str)
```

If you believe that the example I gave (https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4192222503) is non-conformant, may I please ask that you open a discussion on the typing discourse or on their github?

dangotbanned:
> If you believe that the example I gave https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4192222503 is non-conformant

This PR is not about that example, although both are related to overlapping overloads.
I'm interested in specifically (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r3035465450), which is where the change to `IntoFrameT` has caused all the new issues.

By my count, `IntoDataFrameT` and `IntoLazyFrameT` (with `Unpack[AllowLazy]`) each can match 2 of these 3 overloads:
- https://github.com/MarcoGorelli/narwhals/blob/a32fb68cfabf99a0de3f8314a1d9bd26f1629eb2/narwhals/translate.py#L146-L149
- https://github.com/MarcoGorelli/narwhals/blob/a32fb68cfabf99a0de3f8314a1d9bd26f1629eb2/narwhals/translate.py#L154-L157
- https://github.com/MarcoGorelli/narwhals/blob/a32fb68cfabf99a0de3f8314a1d9bd26f1629eb2/narwhals/translate.py#L158-L161

I believe this is why you see `# type: ignore[overload-overlap]` following the second two.
It's possible (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r3070080109) may be a path to avoiding the overlap.

At worst, it couldn't hurt to try?

---

I'm not sure what issue is being surpressed on this line:
- https://github.com/MarcoGorelli/narwhals/blob/a32fb68cfabf99a0de3f8314a1d9bd26f1629eb2/narwhals/translate.py#L178

But I do know that it would also [prevent other type checkers](https://discuss.python.org/t/allow-type-ignore-mypy-style-error-codes-in-conformance-tests/105433) from showing a diagnostic there if they had one 


---

So bringing it back to (https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4225524953)
I appreciate that you've addressed most of my points and I've marked lots of things as resolved 🥳 

But I currently feel:
1. That the warnings from type checkers are correct.
   If we can find a solution that focuses on avoiding them, it would be best for us in the long-term
2. If this is working right now (which I feel is reliant on shaky grounds), it's reasonable to think it may soon not be

MarcoGorelli:
> This PR is not about that example

Sure, but just to check that we're on the same page - do you agree that that example is conformant or not?

> By my count, IntoDataFrameT and IntoLazyFrameT (with Unpack[AllowLazy]) each can match 2 of these 3 overloads:

Yes, and all materialisations of `IntoDataFrameT` will match the second one ([L154-L157](https://github.com/MarcoGorelli/narwhals/blob/a32fb68cfabf99a0de3f8314a1d9bd26f1629eb2/narwhals/translate.py#L154-L157)), so the third one will be ignored (at least, that's my understanding)

> > It's possible (https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r3070080109) may be a path to avoiding the overlap.
> At worst, it couldn't hurt to try?

Yup, thanks 🙏 . Have tried
```diff
--git a/narwhals/translate.py b/narwhals/translate.py
index f9ef32461..5ed0bde06 100644
--- a/narwhals/translate.py
+++ b/narwhals/translate.py
@@ -148,17 +148,19 @@ def from_native(
     native_object: IntoLazyFrameT, **kwds: Unpack[AllowLazy]
 ) -> LazyFrame[IntoLazyFrameT]: ...
 @overload
-def from_native(  # type: ignore[overload-overlap]
+def from_native(
     native_object: IntoDataFrameT | IntoSeriesT, **kwds: Unpack[AllowSeries]
 ) -> DataFrame[IntoDataFrameT] | Series[IntoSeriesT]: ...
 @overload
 def from_native(
     native_object: IntoDataFrameT | IntoLazyFrameT, **kwds: Unpack[AllowLazy]
 ) -> DataFrame[IntoDataFrameT] | LazyFrame[IntoLazyFrameT]: ...
+# @overload
+# def from_native(  # type: ignore[overload-overlap]
+#     native_object: IntoFrameT, **kwds: Unpack[AllowLazy]
+# ) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
 @overload
-def from_native(  # type: ignore[overload-overlap]
-    native_object: IntoFrameT, **kwds: Unpack[AllowLazy]
-) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
+def from_native(native_object: IntoFrameT) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
 @overload
 def from_native(
     native_object: IntoDataFrameT | IntoLazyFrameT | IntoSeriesT, **kwds: Unpack[AllowAny]
@@ -175,7 +177,7 @@ def from_native(
     series_only: bool,
     allow_series: bool | None,
 ) -> Any: ...
-def from_native(  # type: ignore[misc]
+def from_native(
     native_object: IntoFrameT
     | IntoLazyFrameT
     | IntoDataFrameT
```
, and mypy reports
```
narwhals/translate.py:151: error: Overloaded function signatures 10 and 11 overlap with incompatible return types  [overload-overlap]
    def from_native(
    ^
narwhals/translate.py:180: error: Overloaded function implementation cannot produce return type of signature 12  [misc]
    def from_native(
    ^
```

MarcoGorelli:
For reference i asked about that on the pyrefly discord and got the response

> That seems correct to me! I'd expect this behavior (when the argument types are fully known, pick the first matching overload) to be pretty stable, if for no other reason than that the fallout would be pretty big if we tried to change it.

https://discord.com/channels/1319086885024567347/1494693618441519205/1494806090368679997

dangotbanned:
> For reference i asked about that on the pyrefly discord and got the response

Thanks for asking!

I think I'd need to join the discord to find out, but what was the question?

> > That seems correct to me! I'd expect this behavior (**when the argument types are fully known,** ...

I can't tell if this means they're talking about a case like `Sequence[str]` or including overlapping type variables too.

**Edit**:
*Well I just looked and indeed this was an example for `Sequence[str]` again*

```py
from typing import overload, Sequence, reveal_type

@overload
def foo(a: str) -> int: ...  # type: ignore[overload-overlap]

@overload
def foo(a: Sequence[str]) -> str: ...

def foo(a: str | Sequence[str]) -> int | str:
    if isinstance(a, str):
        return 0
    return 'zero'

reveal_type(foo('abc'))  # Revealed type is "int"
reveal_type(foo(['abc']))  # Revealed type is "str"
```

In my head I wouldn't think a `TypeVar` is fully known, but IIRC there are some terms like that defined in the typing spec 

Or maybe this is just saying the type doesn't contain `Any` or `Unknown`?

dangotbanned:
https://github.com/narwhals-dev/narwhals/pull/3496#issuecomment-4237964002

> > > It's possible ([#3496 (comment)](https://github.com/narwhals-dev/narwhals/pull/3496#discussion_r3070080109)) may be a path to avoiding the overlap.
> > > At worst, it couldn't hurt to try?
> 
> Yup, thanks 🙏 . Have tried
> <details><summary>Details</summary>
> <p>
> 
> ```diff
> --git a/narwhals/translate.py b/narwhals/translate.py
> index f9ef32461..5ed0bde06 100644
> --- a/narwhals/translate.py
> +++ b/narwhals/translate.py
> @@ -148,17 +148,19 @@ def from_native(
>      native_object: IntoLazyFrameT, **kwds: Unpack[AllowLazy]
>  ) -> LazyFrame[IntoLazyFrameT]: ...
>  @overload
> -def from_native(  # type: ignore[overload-overlap]
> +def from_native(
>      native_object: IntoDataFrameT | IntoSeriesT, **kwds: Unpack[AllowSeries]
>  ) -> DataFrame[IntoDataFrameT] | Series[IntoSeriesT]: ...
>  @overload
>  def from_native(
>      native_object: IntoDataFrameT | IntoLazyFrameT, **kwds: Unpack[AllowLazy]
>  ) -> DataFrame[IntoDataFrameT] | LazyFrame[IntoLazyFrameT]: ...
> +# @overload
> +# def from_native(  # type: ignore[overload-overlap]
> +#     native_object: IntoFrameT, **kwds: Unpack[AllowLazy]
> +# ) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
>  @overload
> -def from_native(  # type: ignore[overload-overlap]
> -    native_object: IntoFrameT, **kwds: Unpack[AllowLazy]
> -) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
> +def from_native(native_object: IntoFrameT) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]: ...
>  @overload
>  def from_native(
>      native_object: IntoDataFrameT | IntoLazyFrameT | IntoSeriesT, **kwds: Unpack[AllowAny]
> @@ -175,7 +177,7 @@ def from_native(
>      series_only: bool,
>      allow_series: bool | None,
>  ) -> Any: ...
> -def from_native(  # type: ignore[misc]
> +def from_native(
>      native_object: IntoFrameT
>      | IntoLazyFrameT
>      | IntoDataFrameT
> ```
> 
> </p>
> </details> 
> 
> 
> , and mypy reports
> 
> ```
> narwhals/translate.py:151: error: Overloaded function signatures 10 and 11 overlap with incompatible return types  [overload-overlap]
>     def from_native(
>     ^
> ```



Thanks, really appreciate you trying that out.




> ```py
> narwhals/translate.py:180: error: Overloaded function implementation cannot produce return type of signature 12  [misc]
>     def from_native(
>     ^
> ```


Just so I understand, this error was already present - but ignored - and my suggestion didn't change that right?

MarcoGorelli:
Yup, that's right!

We also have pyrefly running on the test suite now, and it's encouraging that that reports no issues with this pr

```
./.venv/bin/uv run --no-sync pyrefly check
 INFO Checking project configured at `/home/runner/work/narwhals/narwhals/pyproject.toml`
 INFO 0 errors (263 suppressed, 12 warnings not shown)
```

dangotbanned:
What would this PR look like if we approached this as widening **only** `LazyFrame[...]`?

- I can live with there being only one side that goes all *loosey-goosey*
- TBH this would reflect the `pyarrow` and `pandas` situation better
- The overloads would likely need to be rethought (maybe `DataFrame[IntoDataFrame]` gets priority?)
- Might be easier for plugins, where I expect lazy-only is gonna be more common


## Edit
The goal of this *would not* be to support `IntoFrameT` everywhere, but to make `from_native` to work for more cases than it does currently.
I think this goal is what the aim really is and IMO any changes to `from_native` are just gonna be balancing trade-offs

MarcoGorelli:
> What would this PR look like if we approached this as widening **only** `LazyFrame[...]`?

could you show how this would look in a separate PR please? i'm not sure how you're expecting the overloads to be written

MarcoGorelli:
thanks for bearing with me

sure, i think i'm convinced, let's soft-deprecate (i.e. just discourge, don't actually deprecate) `IntoFrameT` and just use `IntoDataFrameT` / `IntoLazyFrameT` in its place

MarcoGorelli:
closing in favour of https://github.com/narwhals-dev/narwhals/issues/3678

Issue #3226, opened 2025-10-20
`from_native(nw.LazyFrame, eager_only=True)` returns `nw.LazyFrame`


dangotbanned:
Noticed while working on #3125, but this is a runtime issue

```py
import polars as pl

import narwhals as nw

native = pl.LazyFrame()
narwhals = nw.from_native(native)


narwhals
nw.from_native(narwhals, eager_only=False)
nw.from_native(narwhals, eager_only=True)
```

All three of these have identical outputs

```py
┌────────────────────────────┐
|     Narwhals LazyFrame     |
|----------------------------|
|<LazyFrame at 0x249D2EF96D0>|
└────────────────────────────┘
```

## Expected
We have this for native, which seems correct

```py
nw.from_native(native, eager_only=True)
```

```py
TypeError: Cannot only use `eager_only` or `eager_or_interchange_only` with polars.LazyFrame
```

I was going to suggest doing what we have for `series_only`, which *looked* correct to me:

https://github.com/narwhals-dev/narwhals/blob/556c6a6672af7b2de01a6c65296301e3cb912dfc/narwhals/translate.py#L337-L341

But the error and warning are kinda weird:

```py
nw.from_native(narwhals, series_only=True)
```

```py
UserWarning: You passed a `<class 'narwhals.dataframe.LazyFrame'>` to `is_pandas_dataframe`.

Hint: Instead of e.g. `is_pandas_dataframe(df)`, did you mean `is_pandas_dataframe(df.to_native())`?
  nw.from_native(narwhals, series_only=True)

TypeError: Expected pandas-like dataframe, Polars dataframe, or Polars lazyframe, got: <class 'narwhals.dataframe.LazyFrame'>
```

MarcoGorelli:
thanks for spotting this - sure, sounds good

PR #3515, opened 2026-03-18
fix: Update version when passing narwhals objects in `from_native`

camriddell:
# Description
Calling `from_native(nw.DataFrame(…), version=…)` would always return the passed DataFrame/Series object.
This ignores the version that is specified in the version argument.

Instead, for a given narwhals DataFrame/Series `from_native` now returns the same object **if the versions match** and creates a new object if the versions do not match.

Note, 2 clean up items are also added in this PR.

Before: 

```python
import narwhals as nw, narwhals.stable.v1 as nw_v1
import polars as pl

native = pl.DataFrame({"a": [1, 2, 3]})

unstable = nw.from_native(native)
stablified = nw_v1.from_native(unstable)

print(
    isinstance(stablified, nw_v1.DataFrame), # False
    isinstance(stablified, nw.DataFrame),    # True
    sep='\n'
)

stable = nw_v1.from_native(native)
unstablified = nw.from_native(unstable)

print(
    isinstance(unstablified, nw_v1.DataFrame), # False
    isinstance(unstablified, nw.DataFrame),    # True
    sep='\n'
)
```

After:

```python
import narwhals as nw, narwhals.stable.v1 as nw_v1
import polars as pl

native = pl.DataFrame({"a": [1, 2, 3]})

unstable = nw.from_native(native)
stablified = nw_v1.from_native(unstable)

print(
    isinstance(stablified, nw_v1.DataFrame), # True
    isinstance(stablified, nw.DataFrame),    # True (since nw.v1.DataFrame is subclass of nw.DataFrame)
    sep='\n'
)

stable = nw_v1.from_native(native)
unstablified = nw.from_native(unstable)

print(
    isinstance(unstablified, nw_v1.DataFrame), # False
    isinstance(unstablified, nw.DataFrame),    # True
    sep='\n'
)
```

reported by hoxbro (Holoviews) on Discord #help channel

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/tests/translate/from_native_test.py
+++ b/tests/translate/from_native_test.py
@@ -239,7 +239,7 @@ def test_init_already_narwhals_unstable_to_stable() -> None:
 
     s = native["a"]
     unstable_s = nw.from_native(s, allow_series=True)
-    stablified_s = nw_v1.from_native(unstable_s, allow_series=True)
+    stablified_s = nw_v1.from_native(unstable_s, allow_series=True)  # type: ignore[var-annotated]
     assert isinstance(stablified_s, nw_v1.Series)
 
 
@@ -254,8 +254,8 @@ def test_init_already_narwhals_stable_to_unstable() -> None:
     assert isinstance(unstablified_df, nw.DataFrame)
 
     s = native["a"]
-    stable_s = nw_v1.from_native(s, allow_series=True)
-    unstablified_s = nw.from_native(stable_s, allow_series=True)
+    stable_s = nw_v1.from_native(s, allow_series=True)  # type: ignore[var-annotated]
+    unstablified_s = nw.from_native(stable_s, allow_series=True)  # type: ignore[var-annotated]
     assert isinstance(unstablified_s, nw.Series)
 
 


narwhals/dataframe.py
```diff
@@ -2361,6 +2361,8 @@ class LazyFrame(BaseFrame[LazyFrameT]):
         ```
     """
 
+    _version: ClassVar[Version] = Version.MAIN
```
camriddell, line 2364: Hoping this has been an oversight and wasn't excluded for a strong reason. This change makes the DataFrame/LazyFrame api a bit more consistent internally.

camriddell:
@MarcoGorelli I believe the nightly failure isn't due to this change but perhaps something that needs to be accounted for in pandas. Shall I open up an Issue with this? Or do you believe it is relevant to this PR?

narwhals/dataframe.py
dangotbanned: Yeah nice one, thanks @camriddell!

If you dig through the blame (I think) I added these on `Series` & `DataFrame` at the same time as the `from_*` constructors.

`LazyFrame` didn't get any new methods, so it slipped through 😅

tests/v1_test.py
```diff
@@ -621,7 +621,7 @@ def test_lazyframe_recursive_v1() -> None:
 
     pl_frame = pl.DataFrame({"a": [1, 2, 3]}).lazy()
     nw_frame = nw_v1.from_native(pl_frame)
-    with pytest.raises(AttributeError):
+    with pytest.raises(AssertionError):
```
dangotbanned, line 624: IIRC, there may be a comment somewhere mentioning the error?

(Ignore me if this is too vague 😂)

dangotbanned:
@camriddell I should probably check what was mentioned on discord - but from reading the PR I'm a bit confused on the problem that's being solved.

(https://github.com/narwhals-dev/narwhals/pull/3515#discussion_r2955644370) is all good - but something seems strange to me that calling `from_native` changes the version 🤔

FWIW, I don't think `from_native` should accept narwhals-level objects, but we are where we are 😅

MarcoGorelli:
thanks @camriddell !

on board with the suggestions, i think this just needs addressing

```
tests/translate/from_native_test.py:242: error: Need type annotation for
"stablified_s"  [var-annotated]
        stablified_s = nw_v1.from_native(unstable_s, allow_series=True)
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/translate/from_native_test.py:257: error: Need type annotation for
"stable_s"  [var-annotated]
        stable_s = nw_v1.from_native(s, allow_series=True)
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/translate/from_native_test.py:258: error: Need type annotation for
"unstablified_s"  [var-annotated]
        unstablified_s = nw.from_native(stable_s, allow_series=True)
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 3 errors in 1 file (checked 469 source files)
make: *** [Makefile:26: typing] Error 1
```

---

the nightly failure is just due to https://github.com/pandas-dev/pandas/issues/64749 and is unrelated to your pr

camriddell:
> @camriddell I should probably check what was mentioned on discord - but from reading the PR I'm a bit confused on the problem that's being solved.
> ([#3515 (comment)](https://github.com/narwhals-dev/narwhals/pull/3515#discussion_r2955644370)) is all good - but something seems strange to me that calling `from_native` changes the version 🤔

The primary concern here is surprise that the `v1.from_native` can give you a non-`v1` object back. The current fast path implementation means that if you call:  `v2.from_native(v1.DataFrame)` you get back the same `v1.DataFrame`, which is odd because every other input type will give you back a `v2.DataFrame`. Meaning the signature of `v2.from_native` return type is "usually a v2 object, but sometimes whatever you passed into it".

> FWIW, I don't think `from_native` should accept narwhals-level objects, but we are where we are 😅

I agree with this from a structurally-correct perspective. However, conveniences are nice for DX especially since our primary audience is dataframe-agnostic libraries. This code let's them treat **any** incoming frame the same way without having to manually check if it is already a narwhals DataFrame. Likely their end users won't be passing in narwhals DataFrames, but they could have multiple entry points may need to translate to a narwhals object and our "recursive" type support means they don't have to write explicit `isinstance` checks before every `from_native` call.

tests/v1_test.py
camriddell: I found this comment in the [comparable tests for DataFrames ](https://github.com/camriddell/narwhals/blob/835c683de608ba9f072cc674599005345f2969d7/tests/v1_test.py#L603-L605)

```
    # NOTE: (#2629) combined with passing in `nw_v1.DataFrame` (w/ a `_version`) into itself changes the error
    with pytest.raises(AssertionError):
        nw_v1.DataFrame(nw_frame, level="full")
```

The LazyFrame had an `AttributeError` because the `LazyFrame.__init__` would try to check the `._version` attribute which didn't exist. This change more closely aligns this LazyFrame with its DataFrame and Series counterparts since they all successfully fail with the same error now.

dangotbanned: > successfully fail 

😂

dangotbanned:
(https://github.com/narwhals-dev/narwhals/pull/3515#issuecomment-4107500896)

Thanks for the context @camriddell!

I have a couple more questions, are you okay if we sit on this until later today?

camriddell:
> ([#3515 (comment)](https://github.com/narwhals-dev/narwhals/pull/3515#issuecomment-4107500896))
> 
> Thanks for the context @camriddell!
> 
> I have a couple more questions, are you okay if we sit on this until later today?

Absolutely! I'm not in a rush to merge.

dangotbanned:
@camriddell thanks for your patience 😄


### 1. Similar issue
Did you know about this one?

- [ ] #3226

### 2. What's the worst that could happen?
I'm curious if we can reveal cases downstream that are (*unknowingly?*) relying on this behavior.

As an example, if I change [`tests/v1_test.py::test_with_row_index`](https://github.com/narwhals-dev/narwhals/blob/b8b7b102d81f1c5d173c42bd3fb2d109ddc62b3d/tests/v1_test.py#L445-L463) like this on `main` - it passes through no problemo.
But wouldn't that be a `v1` break following this PR?

<p>

```diff
diff --git a/tests/v1_test.py b/tests/v1_test.py
index 330211162..59ccdae61 100644
--- a/tests/v1_test.py
+++ b/tests/v1_test.py
@@ -447,7 +447,7 @@ def test_with_row_index(constructor: Constructor) -> None:
         pytest.skip()
     data = {"abc": ["foo", "bars"], "xyz": [100, 200], "const": [42, 42]}
 
-    frame = nw_v1.from_native(constructor(data))
+    frame = nw.from_native(nw_v1.from_native(constructor(data)))
 
```

</p>
</details> 


What happens if you try running the CI *but* change the [let's see who this really is](https://github.com/camriddell/narwhals/blob/835c683de608ba9f072cc674599005345f2969d7/narwhals/translate.py#L288) to **raising an exception**?


### 3. Is this discoverable?
Part of what I was getting at in (https://github.com/narwhals-dev/narwhals/pull/3515#issuecomment-4085418360) is that this looks like 2 user errors *to me*:

```py
# (1) Mixed versions
# (2) Asking for something you already have
nw.from_native(nw_v1.DataFrame(...))
```

*If* you think the feature to switch versions is something we need, wouldn't this be a better fit?:

```py
nw_v1.DataFrame(...).to_version("main")
nw_v1.DataFrame(...).to_version("v1")
nw_v1.DataFrame(...).to_version("v2")
```

There's less ambiguity with something along those lines IMO - and it avoids defining new behavior for `from_native`

hoxbro:
As it was me who reported the problem on discord. I just want to give some background the original problem was based on a this issue https://github.com/holoviz/holoviews/issues/6828. Where a user passed in a `MAIN` version of a narwhals DataFrame, and we internally in HoloViews uses `nw2.from_native`.

dangotbanned:
Thanks @hoxbro! It's definitely easier to think about when we can see the problem 🙂 

So my suggestion in (https://github.com/narwhals-dev/narwhals/pull/3515#issuecomment-4114192950) was similar to what [`marimo` have here](https://github.com/marimo-team/marimo/blob/2c42af823fef4621bf5b4d0264edcb6451b03def/marimo/_utils/narwhals_utils.py#L200-L238)
After looking through (https://github.com/holoviz/holoviews/issues/6828) and ([1], [2]) - I guess that might look like:

```diff
        if cls.narwhals_backend and isinstance(data, dict):
            data = nw.from_dict(data, backend=cls.narwhals_backend)
        elif cls.narwhals_backend and isinstance(data, tuple):
            dims = map(str, (*(kdims or ()), *(vdims or ())))
            data = nw.from_dict(dict(zip(dims, data, strict=False)), backend=cls.narwhals_backend)
        elif cls.narwhals_backend and isinstance(data, np.ndarray):
            dims = list(map(str, (*(kdims or ()), *(vdims or ()))))
            data = nw.from_numpy(data, schema=dims, backend=cls.narwhals_backend)
        else:
            data = nw.from_native(data, allow_series=True)

+       data = data.to_version("v2")
```

Although (https://github.com/holoviz/holoviews/issues/6828#issuecomment-4074674999) makes me *think* passing narwhals objects *directly* to holoviews wasn't expected?

---

My main concern with changing `from_native` is for `v1`.
There are [a few API breaks](https://narwhals-dev.github.io/narwhals/backcompat/#main-vs-stablev1-differences) - but so far [`v2`, not so much](https://narwhals-dev.github.io/narwhals/backcompat/#main-vs-stablev2-differences)


[1]: https://github.com/holoviz/holoviews/blob/2d748481804b77668c9780cf25464b7ef7bc9fd0/holoviews/core/data/narwhals.py#L77-L85
[2]: https://github.com/holoviz/holoviews/blob/2d748481804b77668c9780cf25464b7ef7bc9fd0/holoviews/core/data/narwhals.py#L91-L148

hoxbro:
I expected that everything coming out of the version 2 `nw.from_native` would be a version 2 Frame/Series. 

The `applies` method in HoloViews should likely accept every version of a Narwhals Frame/Series.

MarcoGorelli:
> I expected that everything coming out of the version 2 nw.from_native would be a version 2 Frame/Series.

Agree, and I'd consider any place where that's not the case to be a bug, so I'm OK with changing this

dangotbanned:
> looks fine to me, any further objections?

I'd have thought **2.** in (https://github.com/narwhals-dev/narwhals/pull/3515#issuecomment-4114192950) might be worth investigating first?
Alternatively, could we mention in the release notes that this *may* be breaking?


I'm okay with considering it a bug fix, but if we can help at least one person discover *where* that bug was - then I think we should 🙂

MarcoGorelli:
sure, @camriddell fancy trying out

> What happens if you try running the CI but change the [let's see who this really is](https://github.com/camriddell/narwhals/blob/835c683de608ba9f072cc674599005345f2969d7/narwhals/translate.py#L288) to raising an exception?

to see if anything happens?

camriddell:
@dangotbanned & @MarcoGorelli I tossed in an exception and am running tests in a different PR: https://github.com/narwhals-dev/narwhals/pull/3532

We'll see if any downstream codebases rely on this `{bug,feature}`?
Note: My opinion is that this addition is a clearer way forward and the previous behavior should be considered a bug.

MarcoGorelli:
thanks! i'm inclined to agree, and the linked PR doesn't seem to be catching any regressions (the Altair failure is unrelated typing, and the Marimo one also seems unrelated and is present in other PRs)

MarcoGorelli:
awesome, thanks for your attention to detail here!

PR #3549, opened 2026-04-14
docs: Clarify behavior of `Expr.over(order_by=...)`

### LINKED ISSUE #3542: [Doc]: Clarity on behaviour when setting `partition_by=None` (default) with `nw.Expr.over`

Improvement

For lazy operations where we can't rely on order dependence, something like `.over(order_by="date")` is useful!

From the [docs](https://narwhals-dev.github.io/narwhals/api-reference/expr/#narwhals.Expr.over) for `over`, it may not be immediately obvious to the reader what happens when `partition_by=None`, as there are only examples where this is specified to be a different column (similarly in the [`polars` docs](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.over.html)!), e.g., `nw.col("a").cum_sum().over("b")`

This is a suggestion then to include an explicit note on default behaviour here, i.e., that upon collection the expression is performed on a sorted frame and results returned with original row order (with original dimensions).

Cheers!

_No response_

FBruzzesi:
# Description

@danielarifmurphy let me know if this is helpful

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3542


narwhals/expr.py
```diff
@@ -1310,6 +1318,22 @@ def over(
             |1  2  x                    3|
             |2  4  y                    4|
             └────────────────────────────┘
+
+            When `partition_by` is omitted, the expression is computed over the
```
MarcoGorelli, line 1309: shall we remove the previous (now outdated) example now?

narwhals/expr.py
```diff
@@ -1279,9 +1279,17 @@ def over(
...
+                (i.e., no grouping is applied).
             order_by: Column(s) to order window functions by.
                 For lazy backends, this argument is required when `over` is applied
                 to order-dependent functions, see [order-dependence](../concepts/order_dependence.md).
+                When `order_by` is specified, the expression is evaluated on the frame
+                sorted by the given column(s), and, if applicable, the results are
+                returned with the original row order preserved.
+
+        Warning:
+            At least one of `partition_by` or `order_by` must be provided.
```
MarcoGorelli: maybe "note" rather than "warning"?

MarcoGorelli (commented):
thanks!

dangotbanned:
Just a note on the naming:

1. `partition_by=None` wouldn't be valid for our typing (https://github.com/narwhals-dev/narwhals/pull/3549#event-25170650538)
   The runtime equivalent is this mess `Expr.over(*(), order_by=...)` 😉
2. `"behaviour"` is the correct spelling where I'm from, 
   but I usually aim for American English in or about code (https://github.com/narwhals-dev/narwhals/pull/3549#event-25170656843)

dangotbanned (approved):
Thanks @FBruzzesi 

this narwhal loves you

MarcoGorelli:
we use "behaviour" in several places in the repo, no need to default to the american spelling imo

PR #3552, opened 2026-04-16
feat: Introduce `testing.constructors` module and a pytest plugin

### LINKED ISSUE #3653: Add  `testing.constructors` module and a pytest plugin
## Related
- https://github.com/narwhals-dev/narwhals/issues/739#issuecomment-3225435754
- https://github.com/narwhals-dev/narwhals/pull/3552#issue-4276020073
- #3569
- #3556

FBruzzesi:
# Description

## TL;DR

* This PR extracts the backend constructor machinery that used to live in `tests/conftest.py` into a first class, documented module (`narwhals.testing.constructors`) and ships it as an installable [**pytest plugin**](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#making-your-plugin-installable-by-others) (registered via the `pytest11` entry point). Downstream libraries that build on Narwhals can  get parametrised dataframe/lazyframe fixtures, without copy pasting our conftest.
* It also reworks how coverage is collected in CI (dropping `pytest-cov` in favour of a `coverage run -> combine -> report` flow), because registering an entry point plugin changes import timing in a way that breaks coverage measurement.

## Motivation

* The constructor fixtures (`constructor`, `constructor_eager`, the `--constructors` flag) were Narwhals-internal, undocumented, and trapped inside `tests/conftest.py`. Downstream projects that wanted to test against the same backend matrix had to vendor our conftest.
* Related to #739 * not marking it as closing, since the goal is to fully dogfood the plugin in our own suite first.
* The previous design used bare callables with `__name__` sniffing (`"pandas" in str(constructor)`) for backend detection. Moving to a typed `frame_constructor` object with explicit metadata (implementation, requirements, eager/lazy, nullability, GPU) makes that introspection first class.

### 1. New module: `narwhals/testing/constructors.py`

The core abstraction is `frame_constructor`, a callable wrapper around a backend builder function.

* Each constructor wraps exactly one backend builder and turns a column oriented `dict` into a native frame.
* Metadata lives on the instance: `implementation`, `requirements`, `is_eager`, `is_nullable`, `needs_gpu`.
* Registration is explicit and performed via the `@frame_constructor.register(...)` decorator, which instantiates the wrapper and stores it in a shared class level `_registry`.
* `is_*` properties delegate to `Implementation` (`is_pandas`, `is_polars`, `is_pandas_like`, `is_spark_like`, `is_duckdb`, `is_lazy`, `needs_pyarrow`, `is_available`, etc.), replacing the old string sniffing.
* `__call__(obj, namespace, **kwds)` builds the native frame and then wraps it via `namespace.from_native(...)`, so the fixture returns a **Narwhals** frame directly.

Module level helpers (all part of `__all__`):

* `available_backends()` / `available_cpu_backends()`: names whose libraries are importable.
* `get_backend_constructor(name)`: typed lookup, raises `ValueError` with the valid set on a bad name.
* `prepare_backends(include=..., exclude=...)`: filtered, sorted selection (`exclude` takes precedence).
* `is_backend_available(*packages)`: thin `importlib.util.find_spec` check.
* `pyspark_session()` / `sqlframe_session()`: session factories (moved out of `tests/utils.py`).
* `DEFAULT_BACKENDS`: the historical default subset (`pandas`, `pandas[pyarrow]`, `polars[eager]`, `pyarrow`, `duckdb`, `sqlframe`, `ibis`).

### 2. New module: `narwhals/testing/pytest_plugin.py`

The installable pytest plugin, wired up in `pyproject.toml`:

```toml
[project.entry-points.pytest11]
narwhals_testing = "narwhals.testing.pytest_plugin"
```

exposes

**Fixtures** (auto-parametrised against the selected, installed backends):

| Fixture | Parametrised over |
|---|---|
| `nw_frame` | every selected backend (eager + lazy) |
| `nw_dataframe` | eager backends only |
| `nw_lazyframe` | lazy backends only |
| `nw_pandas_like_frame` | pandas-like eager backends |

**CLI options** (under a `narwhals` option group):

* `--nw-backends=pandas,polars[lazy],duckdb`: comma separated selection. Defaults to `DEFAULT_BACKENDS` intersected with what is installed.
* `--all-nw-backends`: every installed CPU backend (overrides `--nw-backends`; excludes `modin` and `pyspark[connect]`).
* `--use-external-nw-backend`: escape hatch so a downstream suite can register its own parametrisation; the plugin still adds the options but stops parametrising the fixtures.
* Environment overrides: `NARWHALS_DEFAULT_BACKENDS` overrides the default list (useful e.g. under `cudf.pandas`).

### 3. Supporting type aliases: `narwhals/testing/typing.py`

Public aliases for downstream type hints: `Data` (the column oriented input mapping), `FrameConstructor`, `DataFrameConstructor`, `LazyFrameConstructor`, and the `NarwhalsNamespace` protocol (anything exposing `from_native`).

### 4. Test suite migration

The test suite partially moved off the old pattern to keep the changes more contained.

* Compatibility shims: `constructor`, `constructor_eager`, and `constructor_pandas_like` are kept as aliases over the new `nw_*` fixtures, wrapped in `_PatchedFrameConstructor` / `_PatchedDataFrameConstructor` proxies that default `namespace` to `narwhals`. This lets the legacy call sites keep working during the transition. A `TODO` marks their eventual removal.
* Because fixtures now return a Narwhals frame directly (`constructor(data)` instead of `nw.from_native(constructor(data))`), call sites simplified. Example from `join_test.py`: the `from_native_lazy(constructor(df))` helper collapsed to `constructor(df).lazy()`.
* `tests/utils.py`: the `Constructor` / `ConstructorEager` / `ConstructorPandasLike` type aliases are now re-exported from the conftest proxies; the `pyspark_session` / `sqlframe_session` helpers moved into `narwhals.testing.constructors`. `uses_pyarrow_backend` now keys off `str(constructor)` instead of `__name__`.
* New dedicated tests: `tests/testing/constructors_test.py` (registry, properties, `prepare_backends`, equality/hash) and `tests/testing/plugin_test.py` (uses pytest's `pytester` to assert the plugin parametrises eager/lazy fixtures and that `--use-external-nw-backend` disables parametrisation).

### 5. Coverage and CI rework

Registering the entry point plugin forces `narwhals/__init__.py` to import early (before `pytest-cov` installs its tracer), which silently dropped coverage. Fix:

1. Drop the `pytest-cov` dependency.
2. Drive coverage manually: `coverage run -m pytest ... && coverage combine && coverage report`, so the tracer starts at interpreter startup before any import.
3. Configure subprocess/`execv`/`fork` patches in `[tool.coverage.run]`. `execv` and `fork` are unsupported on Windows (coverage raises), so Windows CI jobs collapse those to `subprocess` via the `COVERAGE_PATCH_EXECV` / `COVERAGE_PATCH_FORK` env vars; coverage dedupes the final list. `COVERAGE_PROCESS_START=pyproject.toml` is set across the workflows.

Tooling changes:

* New `run-ci-coverage` Make target (`run -> combine -> report`, with optional `COV_SOURCE` to scope coverage, e.g. `src/narwhals/_spark_like`). `run-ci` stays for non coverage runs (doctests, narrow-dep, tpch, ibis, modin).
* All workflows migrated `--constructors` to `--nw-backends` and `--all-cpu-constructors` to `--all-nw-backends`.
* `utils/import_check.py`: a `testing` allowlist permits the deliberate lazy imports of every backend inside the constructors.
* `utils/sort_api_reference.py`: `testing` added to the skip set (its API doc is hand ordered).
* Docs: new `docs/api-reference/testing.md` documenting fixtures, options, and a quick start.
* `CONTRIBUTING.md` updated for the renamed flags.

## Public API surface

Newly importable and documented:

* `narwhals.testing.frame_constructor` (and the `constructors` module helpers above).
* `narwhals.testing.typing` aliases.
* The pytest fixtures `nw_frame`, `nw_dataframe`, `nw_lazyframe`, `nw_pandas_like_frame` and the `--nw-*` options.

## Quick start (downstream usage)

```python
import narwhals as nw
import narwhals.stable.v2 as nw_v2
from narwhals.testing.typing import Data, DataFrameConstructor, LazyFrameConstructor


def test_shape(nw_dataframe: DataFrameConstructor) -> None:
    data: Data = {"x": [1, 2, 3]}
    df = nw_dataframe(data, namespace=nw)
    assert df.shape == (3, 1)


def test_laziness(nw_lazyframe: LazyFrameConstructor) -> None:
    lf = nw_lazyframe({"x": [1, 2, 3]}, namespace=nw_v2)
    assert isinstance(lf, nw_v2.LazyFrame)
```

```bash
pytest --nw-backends="pandas,polars[lazy]"
pytest --all-nw-backends
```

## Out of scope

* A `nw_series_constructor` fixture. Ergonomically nice, but a separate feature request.
* Dropping the `constructor` / `constructor_eager` / `constructor_pandas_like` compatibility shims. Tracked by a `TODO`; they stay until every test calls the `nw_*` fixtures directly.

<details><summary>Old description</summary>
<p>

This PR introduces two new features:

1. A rework of `tests/conftest.py` into `narwhals/testing/constructors`
	- [ ] TBD what should be publicly exposed in here
3. A [pytest plugin](https://docs.pytest.org/en/stable/how-to/writing_plugins.html#making-your-plugin-installable-by-others) that exposes the fixtures for dataframe/lazyframe constructors.
    - [x] I wonder if we should already return them as narwhals DataFrame/LazyFrame. If we find a way to achieve the following points, then I **personally** think we should do so, as we almost always run `nw.from_native` right after.
    	- Distinguish between native frame and `nw.from_native` kwargs -> For this it's enough to provide it as different proper arguments in `__call__`, e.g. `nw_kwargs`, `backend_kwargs`.
    	- How do we know which version a user would use? Should we require to pass either the `Version.X` or the namespace `{nw, nw_v1, nw_v2}` ? If we do so, I would argue that we must enforce it and never assume for the user. In my opinion `nw_frame_constructor(data, version=nw)` would still look better than `nw.from_native(nw_frame_constructor(data))`.
    	**Update**: https://github.com/narwhals-dev/narwhals/pull/3569
    - [x] In light of the CI failures for downstream projects, I also realized that having fixtures called `constructor` and `constructor_eager` is not really intuitive. I am happy to brainstorm better names for downstream users. And similarly the flags `--all-cpu-constructors`, `--constructors`. (*) 
    **Update**: #3556
	- [x] We could provide a `nw_lazy_constructor` for "free"
	- [ ] I believe it might be useful to expose a series constructor (under the name `nw_series_constructor`). I know it's possible to make a dataframe and then get a column. But it would definitely improve users ergonomics. However, that's feature request/improvement and **out of scope of this PR**

## (*) Renamed in #3556 and #3569

* fixtures: `nw_frame`, `nw_dataframe`, `nw_lazyframe`
* pytest options: `--nw-all-backends` (misleading that has no CPU in it?), `--nw-backends=...`. I think the term constructor might be be too vague, and there is no other case in the codebase. While we have a bunch of cases in which we allow to pass a **backend** 

## pyest-cov changes

The `[project.entry-points.pytest11]` registers the pytest plugin which we dogfood in the testsuite here. When `pytest` starts, it loads all entry-point plugins before `pytest-cov` installs its coverage tracer.

Importing `narwhals.testing.pytest_plugin` forces Python to load `narwhals/__init__.py` (to resolve the package), which eagerly imports the entire public API.

On main, no entry point existed, so no narwhals code loaded before coverage started.

The fix:

1. Remove `pytest-cov` dependency
2. Add coverage options in `pyproject.toml` to track subprocess, execv and fork (the latter two are not available on windows, hence multiple hacks are performed in GithubAction's to dynamically populate the fields in `pyproject.toml`)
3. Manually run `coverage run -m pytest ...`, `coverage combine` and `coverage report`, so that coverage run starts the tracer at Python startup before any imports

---

Thanks to @EdAbati and @dangotbanned - I mostly adjusted from your work, not much more than that!

- [x] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #739 (not marking as closing, since we should fully dogfood in the test suite)

## Checklist

- [x] Tests added
- [x] Documented the changes

</p>
</details>


dangotbanned:
Just a quick one from me RE:

> Don't panic due to the number of file changes - I will explain
> 
> We have 257 file changes because all tests change the imports as: 
> 
> ```diff 
> - from tests.utils import Constructor, ConstructorEager 
> + from narwhals.testing.typing import Constructor, ConstructorEager
> ```

Is it possible to add something like this, and keep the imports in place (temporarily)?:

```py
# tests.utils.py
from narwhals.testing.typing import Constructor as Constructor , ConstructorEager as ConstructorEager
```

I really **REALLY** want this PR, but the diff 😳 (yes yes, I know I'm guilty of this too 😉)

FBruzzesi:
@dangotbanned reverted such commit 😉

---

Coverage is not picked up in CI though 🥹

narwhals/testing/constructors/_name.py
camriddell, line 1: The metadata for the constructor is a bit scattered:

- `ConstructorName` carries the string name as enum values
- Whether a given constructor is static information that is only available from within a method (not externally visible; e.g. see property `is_eager`
- The implementation name is held in a dictionary inside of `_name_to_impl`

Would it be possible to store all of the static metadata in one place? I hate to keep bringing up the idea, but a registration pattern fits nicely here:

<details>
<summary>registration mechanism for constructor metadata and behavior</summary>

```python
from enum import Flag, auto
from typing import NamedTuple
from dataclasses import dataclass, field

ALL_CONSTRUCTORS = {}
def register(constructor):
    if constructor.name in ALL_CONSTRUCTORS:
        raise ValueError("Already registered...")
    ALL_CONSTRUCTORS[constructor.name] = constructor
    return constructor


class Tag(Flag):
    IS_EAGER = auto()
    IS_LAZY = auto()
    IS_NON_NULLABLE = auto()
    NEEDS_GPU = auto()
    NEEDS_PYARROW = auto()


@dataclass(frozen=True)
class Constructor:
    name: str
    impl: Implementation
    tags: Tag
    func: Callable[[...], ...] = field(repr=False)

    def __post_init__(self):
        # individual validation can be done here
        assert Tag.IS_EAGER in self.tags or Tag.IS_LAZY in self.tags, "must be either eager or lazy"
        assert (Tag.IS_EAGER | Tag.IS_LAZY) not in self.tags, "cant be both eager and lazy"

    @classmethod
    def decorate(cls, name, impl, tags):
        def decorator(func):
            return cls(name, impl, tags, func)
        return decorator

    @property
    def is_nullable(self):
        return not (Tag.IS_NON_NULLABLE in self.tags)

@register
@Constructor.decorate("pandas", Implementation.PANDAS, Tag.IS_EAGER | Tag.IS_NON_NULLABLE)
def pandas_constructor(data):
    import pandas as pd
    return pd.DataFrame(data)

@register
@Constructor.decorate("pandas[nullable]'", Implementation.PANDAS, Tag.IS_EAGER)
def pandas_nullable_constructor(data):
    import pandas as pd
    return pd.DataFrame(data).converty_dtype('numpy_nullable')

@register
@Constructor.decorate("polars[lazy]'", Implementation.PANDAS, Tag.IS_LAZY)
def pandas_nullable_constructor(data):
    import polars as pl
    return pl.LazyFrame(data)

# Constructors can be dynamically defined
LAZY_CONSTRUCTORS, EAGER_CONSTRUCTORS = set(), set()
for constructor in ALL_CONSTRUCTORS.values():
    if Tag.IS_LAZY in constructor.tags:
        LAZY_CONSTRUCTORS.add(constructor)
    elif Tag.IS_EAGER in constructor.tags:
        EAGER_CONSTRUCTORS.add(constructor)

from pprint import pprint
pprint(EAGER_CONSTRUCTORS)
print()
pprint(LAZY_CONSTRUCTORS)
```

outputs

```python
{Constructor(name="pandas[nullable]'",
             impl=<Implementation.PANDAS: 'pandas'>,
             tags=<Tag.IS_EAGER: 1>),
 Constructor(name='pandas',
             impl=<Implementation.PANDAS: 'pandas'>,
             tags=<Tag.IS_EAGER|IS_NON_NULLABLE: 5>)}

{Constructor(name="polars[lazy]'",
             impl=<Implementation.PANDAS: 'pandas'>,
             tags=<Tag.IS_LAZY: 2>)}
```
</details>

If you made it through that code block, then you're awesome. The mechanics are free to change, like the idea of tags are nice but perhaps a bit too general. Considering that a backend must either be lazy or eager, we might consider a boolean value instead of an optional Tag.

camriddell (commented):
I am definitely a fan of this effort to make testing more accessible for our downstream and to better organize ourselves.

My main critique of the PR in its current form is that important metadata is smeared across these files which makes the code trickier to reason about and extend.

I'd also like to discuss some abstractions that can help us avoid the massive `if/elif` chains at the top of many tests that skip/xfail those tests. If we can refactor that logic into a decorator (or similar style) I think it would make reading the tests easier.

dangotbanned:
> I'd also like to discuss some abstractions that can help us avoid the massive if/elif chains at the top of many tests that skip/xfail those tests. 

@camriddell I've felt the same way about our test suite too.

This is an example of my most recent solution, which let's you `xfail` directly on the the fixture (`dataframe`) - without needing to use the constructor first:

<details><summary>Show <code>xfail</code> usage</summary>
<p>





https://github.com/narwhals-dev/narwhals/blob/a09be54ec5d97225110418a79843c1fbbc933a73/tests/plan/pivot_test.py#L322-L355

</p>
</details> 

Where the fixtures + xfail stuff are defined by:

<details><summary>Show <code>TestBackend</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/a09be54ec5d97225110418a79843c1fbbc933a73/tests/plan/utils.py#L292-L647

</p>
</details> 


And populated here [`conftest.py`](https://github.com/narwhals-dev/narwhals/blob/a09be54ec5d97225110418a79843c1fbbc933a73/tests/plan/conftest.py)


---

Personally though, I'm fine with dealing with shrinking our xfails/skips in a follow-up.
That PR will be massive on its own 😂

narwhals/testing/constructors/_name.py
FBruzzesi: Thanks for the review @camriddell - I fully agree with all the considerations in https://github.com/narwhals-dev/narwhals/pull/3552#pullrequestreview-4130607414 and I was expecting nothing less  😅

I honestly think that there is no need for a ConstructorName class as well and everything can be carried in the constructor.

Also thanks for the code suggestion. I will iterate and eventually try to figure out how the tags would play out in the test suite

FBruzzesi:
@dangotbanned I am happy to keep that as a follow up for the implementation (and crowdsource the migration?).
Here is the start: [feat/testing-constructors-xfail compare](https://github.com/narwhals-dev/narwhals/compare/feat/testing-constructors...feat/testing-constructors-xfail?expand=1).

---

I will get back to figure out the coverage issue for now

FBruzzesi:
@MarcoGorelli @dangotbanned @camriddell I would say this PR is ready to review as is - yet **not ready to be merged** because all the open points + conflicts with plotly & fairlearn pytest fixtures (#3556 addresses this last issue)

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+        if implementation is not None:
+            cls.implementation = implementation
+
+        cls.requirements = requirements
+        cls.legacy_name = legacy_name
+        cls.is_non_nullable = is_non_nullable
+        cls.needs_gpu = needs_gpu
+
+        if "name" not in cls.__dict__:
+            return
```
camriddler, line 137: Does this if check this only exist to catch classes that don't have a name to them? Thus meaning "if you don't have a name specified, you don't get inserted in to the registration mechanism? Seems surprising that if I forget to add a `name=` to my class that I get neither the registration behavior nor an error telling me I forgot something.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+    is_non_nullable: ClassVar[bool] = False
+    needs_gpu: ClassVar[bool] = False
+
+    def __init_subclass__(
+        cls,
+        *,
+        implementation: Implementation | None = None,
+        requirements: tuple[str, ...] = (),
+        legacy_name: str = "",
+        is_non_nullable: bool = False,
```
camriddler, line 111: I'd personally prefer `is_nullable` instead of `is_non_nullable` (if we don't have a backwards compat to worry about)? I find the boolean logic to be easier to understand when making assertions about the positive case as something like `not is_non_nullable` is a touch trickier to reason about than `not is_nullable`.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+            implementation: The [`Implementation`][] this constructor belongs to.
+            requirements: Package names that must be importable for this constructor
+                to be available (checked via `importlib.util.find_spec`).
+            legacy_name: Value returned by `str(constructor)` for backward compatibility
+                with test assertions that match on the old naming scheme.
+            is_non_nullable: Whether the backend lacks native null support.
+            needs_gpu: Whether the backend requires GPU hardware.
+            **kwargs: Forwarded to `super().__init_subclass__`.
+        """
+        super().__init_subclass__(**kwargs)
```
camriddler, line 127: Is there a suspected case for this `super`? I can't imagine nesting classes that override this logic.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+class EagerFrameConstructor(FrameConstructor):
+    """A constructor that returns an *eager* native dataframe."""
+
+    is_eager: ClassVar[bool] = True
+
+    def __call__(self, obj: Data, /, **kwds: Any) -> IntoDataFrame:
+        raise NotImplementedError
+
+
+class LazyFrameConstructor(FrameConstructor):
```
camriddler, line 259: Do these `Lazy|Eager...` sub-classes help anything? They set a default class variable for their children, but it feels like boilerplate outside of that.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+class PandasConstructor(
+    EagerFrameConstructor,
+    implementation=Implementation.PANDAS,
+    requirements=("pandas",),
+    legacy_name="pandas_constructor",
+    is_non_nullable=True,
+):
+    """Constructor backed by `pandas.DataFrame` with default NumPy dtypes."""
+
+    name = "pandas"
```
camriddler, line 280: Feels odd that this argument needs to be attached to the class body instead of passing into the class construction mechanism.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+        "ibis",
+    }
+)
+"""Subset of constructors enabled by default for parametrised tests when the
+user does not pass `--constructors` (mirrors the historical Narwhals defaults).
+"""
+
+ALL_CPU_CONSTRUCTORS: frozenset[str] = frozenset(
+    name for name, c in FrameConstructor._registry.items() if not c.needs_gpu
+)
```
camriddler, line 552: Can `ALL_CPU_CONSTRUCTORS` be a module property or function so that it is kept it in sync in-case the registry is updated after this variable is created?

Exactly as what was done with `available_constructors` below.

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+    except KeyError as exc:
+        valid = sorted(ALL_CONSTRUCTORS)
+        msg = f"Unknown constructor {name!r}. Expected one of: {valid}."
+        raise ValueError(msg) from exc
+
+
+def prepare_constructors(
+    *, include: Iterable[str] | None = None, exclude: Iterable[str] | None = None
+) -> list[FrameConstructor]:
+    """Return available constructors, optionally filtered.
```
camriddler, line 593: 1. We should work more with the underlying types when possible instead of strings. A tricky bug will probably be `prepare_constructors(include=["pndas"])` and you don't get anything back (not even an error). Whereas `prepare_constructors(include=testing.PndasConstructor`) provides an easy to understand error about what was mistyped.
2. Needs a comment that the `exclude` is given precedence here since it is the last conditions that are checked (e.g. it is currently allowed to do `prepare_constructors(include=["pandas"], exclude=["pandas"])` and becomes ambiguous as to what one should expect to return when perhaps this should be an errant case?

narwhals/testing/constructors.py
```diff
@@ -0,0 +1,676 @@
...
+    "available_constructors",
+    "get_constructor",
+    "is_backend_available",
+    "prepare_constructors",
+    "pyspark_session",
+    "sqlframe_session",
+)
+
+
+class FrameConstructor(Protocol):
```
camriddler, line 82: What is gained from making this a `typing.Protocol`?

camriddler (commented):
As a first pass I focused on the `testing.constructors.py` since that is where the implementation details are.

The current registration mechanism has implicit conditions (e.g. `name in cls.__dict__`) that can cause subclasses to fail to register without error (I believe this is used to implicitly differentiate leaf classes). Combined with metadata split between class-body attributes and subclass keyword arguments, this increases contributor error surface and makes registration behavior harder to reason about.

The above observations increase the chance that downstream users will need to read the source code implementation if they are cut by either of above details. (e.g. "why is my class not registering? Oh I forgot to add a `name=...`)

---
**Open for discussion** on this, but I'm still a fan of the decorator + function based registration for a few reasons:
1. Constructing instances is easier to validate/invalidate than initializing subclasses which may have some documentation-only fields (e.g. anything you can put into the class body is fair game)
2. The registration mechanism is an explicit "opt-in" (either the decorator is there or it is not) instead of being tucked behind a class definition + specific field `name` availablility
3. The functions can keep their same name as they originally had and we don't need concern with the `legacy_name` field at all while we transition away from the `str(constructor)` stuff.

```python
from dataclasses import dataclass, field

@dataclass(frozen=True)
class Constructor:
    name: str
    is_eager: bool
    # omitting fields for brevity

    func: Callable[[...], ...]

    _registry = {} # could be a global or class var

    def __call__(self, *args, **kwargs):
        return self.func(*args, **kwargs)

    @classmethod
    def from_func(cls, name, is_eager):
        def decorator(func):
            inst = cls(name=name, is_eager=is_eager, func=func)
            cls._registry[name] = inst
            return inst
        return decorator

@Constructor.from_func("pandas", is_eager=True)
def pandas_constructor(data): # can keep the original function calls, str(...) behavior does not change
    import pandas as pd
    return pd.DataFrame(data)

print(f'{Constructor._registry = }')

# constructors are the same as before, but have metadata attached directly to them
print(f'{pandas_constructor = }')
print(f'{pandas_constructor.is_eager = }')
```

outputs

```python
Constructor._registry = {'pandas': Constructor(name='pandas', is_eager=True)}
pandas_constructor = Constructor(name='pandas', is_eager=True)
pandas_constructor.is_eager = True
```

The caveat of course being that the function `pandas_constructor` is no longer a function, it is an instance of `Constructor` that has an identical `__call__` behavior to the function it decorated.

narwhals/testing/constructors.py
FBruzzesi: I think I am not even using it anymore now - if we are going to expose something like this, then I am definitely going to add better type checking. Sure will add better docstrings if it stays

narwhals/testing/constructors.py
FBruzzesi: Makes sense, thanks for suggesting it

narwhals/testing/constructors.py
FBruzzesi: It was an hack to allow to create the two intermediate `EagerFrameConstructor` and `LazyFrameConstructor` without passing a name (hence without registering the two)

narwhals/testing/constructors.py
FBruzzesi: Useful for type checking (especially `constructor_eager`) + setting one class variable.

narwhals/testing/constructors.py
FBruzzesi: Fair enough - I kept it as "you never know what users want to do"

narwhals/testing/constructors.py
FBruzzesi: Absolutely I fully agree

narwhals/testing/constructors.py
FBruzzesi: I _think_ it has to do with the two `EagerFrameConstructor` and `LazyFrameConstructor` but it was a bunch of commits ago... I swear I already forgot 😭

FBruzzesi (commented):
Thanks for the review @camriddell - there are a lot of great points made there! I replied to single comments

RE: https://github.com/narwhals-dev/narwhals/pull/3552#pullrequestreview-4143419742

My TL;DR is that I don't have a strong preference.

> The current registration mechanism has implicit conditions (e.g. `name in cls.__dict__`) that can cause subclasses to fail to register without error (I believe this is used to implicitly differentiate leaf classes). Combined with metadata split between class-body attributes and subclass keyword arguments, this increases contributor error surface and makes registration behavior harder to reason about.

That's totally valid, and yes we should pick one way of declaring what we need, and avoid having to declare both a class variable and metadata, it should be either one or the other. Also we should not fail silently.

---
> 3. The functions can keep their same name as they originally had and we don't need concern with the `legacy_name` field at all while we transition away from the `str(constructor)` stuff.

Oh man, that's a killer feature!
I will sleep on how to do it a bit longer, thanks for providing example code as well!

> The caveat of course being that the function `pandas_constructor` is no longer a function, it is an instance of `Constructor` that has an identical `__call__` behavior to the function it decorated.

I would not worry too much about this to be honest

camriddell:
@FBruzzesi want to correct an earlier statement, (yeah this one woke me up in the middle of the night)

> The functions can keep their same name as they originally had and we don't need concern with the legacy_name field at all while we transition away from the str(constructor) stuff.

This wasn't actually true, we would need to override the class `__str__` method as you had done previously. The only benefit would be that the name isn't a floating string, it can just be captured from the function name.

```python
print(f'{str(pandas_constructor) = }') # Constructor(name='pandas', is_eager=True, ...)
print(f'{str(pandas_constructor.func) = }') # '<function pandas_constructor at 0x7f6f96213b60>'
```

The modifications:

```python
class Constructor:
    ....
    def __str__(self):
        return self.func.__name__ # or str(self.func) for a more literal backwards compat

```

narwhals/testing/constructors.py
camriddell: If it isn't needed can we remove it? At least in my mind (and someone please update my way of thinking), but a `Protocol` shouldn't have any logic embedded into its methods just scaffolding for attribute & method names/signatures and types. But that's more of a conceptual framework rather than something enforced strongly at runtime.

narwhals/testing/constructors.py
camriddell: That's fair, and this won't hurt us so I'm +1 to keep it in. Was curious if you had something in mind :)

FBruzzesi:
@camriddell I moved to the registry pattern in https://github.com/narwhals-dev/narwhals/pull/3552/commits/e1d09e5436c8cbf759da97a6ca9fc2ef4d905bbb, looking forward to your feedback!

src/narwhals/testing/constructors.py
```diff
@@ -0,0 +1,656 @@
...
+    "prepare_constructors",
+    "pyspark_session",
+    "sqlframe_session",
+)
+
+T_co = TypeVar("T_co", covariant=True, bound="IntoFrame")
+R = TypeVar("R", bound="IntoFrame")
+
+
+class frame_constructor(Generic[T_co]):  # noqa: N801
```
camriddell, line 80: minor: any reason for the snake case class name?

narwhals/testing/constructors.py
dangotbanned: @camriddell I'm not sure exactly what the version that was here looked like - but there were quite a few explicit errors raised here at class-construction time:

https://github.com/narwhals-dev/narwhals/pull/3552#issuecomment-4270517139

The reason I didn't use decorators was because there's concept of a `TestBackend`, which has multiple `Constructor`s.

A downside of using decorators is that here there is a lot of duplicated metadata & operations:

https://github.com/narwhals-dev/narwhals/blob/9712559aadec357185bd27bbd38dad3906d88e92/narwhals/testing/constructors.py#L266-L300

https://github.com/narwhals-dev/narwhals/blob/9712559aadec357185bd27bbd38dad3906d88e92/narwhals/testing/constructors.py#L315-L355

https://github.com/narwhals-dev/narwhals/blob/9712559aadec357185bd27bbd38dad3906d88e92/narwhals/testing/constructors.py#L382-L393

Now, are classes the *only* way to make this DRYer?
Of course not, you can always  stack more decorators 😉

---

Anyway, do whatever floats your goat(s) *this isn't a request for changes* - I'm more than happy for either as an improvement on our current test suite 😘

src/narwhals/testing/constructors.py
```diff
@@ -0,0 +1,656 @@
...
+            inst: frame_constructor[R] = frame_constructor(
+                func,
+                name=name,
+                implementation=implementation,
+                requirements=requirements,
+                is_eager=is_eager,
+                is_nullable=is_nullable,
+                needs_gpu=needs_gpu,
+            )
+            cls._registry[name] = inst
```
camriddell, line 164: Any preference to do something if the key is already set?
1. Raise an error
2. emit a warning
3. let the user replace whatever was there before (current)

I think the current is fine, just wanted to see if you had thoughts here.

narwhals/testing/constructors.py
camriddell: Ah, the implementation I saw was different from what you had linked to! There are some differences, but to your point we're on the path to an improvement :)

Happy to have a chat about the design on another channel (Discord?) or perhaps the next iteration of the test suite! :laughing:

src/narwhals/testing/constructors.py
```diff
@@ -0,0 +1,656 @@
...
+
+    return pd.DataFrame(obj, **kwds).convert_dtypes(dtype_backend="numpy_nullable")
+
+
+@frame_constructor.register(
+    name="pandas[pyarrow]",
+    implementation=Implementation.PANDAS,
+    requirements=("pandas", "pyarrow"),
+    is_eager=True,
+)
```
camriddell, line 338: If duplication is a concern, we can have a secondary mechanism `register_with`. I'm okay with this volume of duplication so will leave this decision to you :)

```python
class frame_constructor:
    ...
    def register_with(self, name, ...): 
        def dec(func): 
            ... # register and instantiate 
            from copy import replace # I don't think this is supported by all of our python versions :(
            return replace(self, name=name, func=func, ...) 
       return dec
```

src/narwhals/testing/constructors.py
```diff
@@ -0,0 +1,656 @@
...
+    "pandas[pyarrow]",
+    "modin",
+    "modin[pyarrow]",
+    "cudf",
+    "polars[eager]",
+    "pyarrow",
+]
+LazyName: TypeAlias = Literal[
+    "polars[lazy]", "dask", "duckdb", "pyspark", "pyspark[connect]", "sqlframe", "ibis"
+]
```
camriddell, line 590: There's a risk of adding an Eager/Lazy backend and these Literals not being aware of it.

This isn't blocking, just wanted to raise a concern.

camriddell (commented):
Overall, I'm much happier with this version as you eliminated the reliance of string checking for class attributes that may/may not exist and am implicit registration mechanism.

A few minor comments/food-for-thoughts

camriddell:
@FBruzzesi I still need to look at the rest of the suite (so far only dug into `constructors.py`), I had finished my thoughts on that file so figured I would share them with you as it'll take me a bit more time to review the rest of the changes.

src/narwhals/testing/constructors.py
FBruzzesi: The camel case decorator gives me javascript vibes nightmares 😅

src/narwhals/testing/constructors.py
FBruzzesi: I think emitting a warning would be ok

src/narwhals/testing/constructors.py
dangotbanned: You can union `LiteralString` with the `Literal` to support extension

camriddell: TIL: `LiteralString` Thanks!

FBruzzesi:
Hey everyone @dangotbanned @camriddell @MarcoGorelli @EdAbati 👋🏼
I know we have a lot of issues and PRs ongoing and open, yet this feature is something I am looking for quite some time, and especially useful in downstream project integration.

I know the diff is quite large - I hope it's manageable, but please let me know how I can make it easier to review (I am going to update the current description soon). Unlike other past cases, I don't see how this could be split down into multiple PRs.

---

For context, there are a few followups to fully dogfooding on this PR, e.g. [feat/testing-constructors-xfail](https://github.com/narwhals-dev/narwhals/compare/feat/testing-constructors...feat/testing-constructors-xfail?expand=1), yet I am not planning to work on it before this one is at least somehow validated

EdAbati:
I'd be very happy to take a look and thank you again for working on this!

Today I'm busy but have some time tomorrow :)

dangotbanned:
> but please let me know how I can make it easier to review (I am going to update the current description soon).

One thing that could make it easier is if the description was interspersed with examples.

E.g.
- how we were doing things before?
- why was that bad?
- how do we improve?
- what are we doing instead?
- why is the new way is better? (and I'm sure it is btw 😘)

src/narwhals/stable/v1/typing.py
```diff
@@ -25,7 +30,9 @@ def __dataframe__(self, *args: Any, **kwargs: Any) -> Any: ...
 `nw.Expr`, e.g. `df.select('a')`.
 """
 
-IntoDataFrame: TypeAlias = Union["NativeDataFrame", "DataFrameLike", "NativeDuckDB"]
+IntoDataFrame: TypeAlias = Union[
+    "NativeDataFrame", "DataFrameLike", "NativeDuckDB", "NativeIbis"
+]
```
EdAbati: (general question, I may miss some context) why do we need to update this in this PR?

src/narwhals/testing/asserts/frame.py
```diff
@@ -165,8 +164,8 @@ def _check_correct_input_type(  # noqa: RET503
 
 
 def _assert_dataframe_equal(
-    left: DataFrameT,
-    right: DataFrameT,
+    left: DataFrame[Any],
+    right: DataFrame[Any],
     impl: Implementation,
```
EdAbati: same as before, curious on why we need to change the typing here?

src/narwhals/testing/pytest_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _default_backend_ids() -> list[str]:
+    """Resolve the default `--nw-backends` value for the current environment.
+
+    Honours `NARWHALS_DEFAULT_BACKENDS` if set, otherwise restricts
+    [`DEFAULT_BACKENDS`][] to backends whose libraries are importable.
+    """
+    if env := os.environ.get("NARWHALS_DEFAULT_BACKENDS"):  # pragma: no cover
+        return env.split(",")
```
EdAbati: will it be possible to drop this `pragma: no cover` if we test:

```py
def test_default_backends_env_var() -> None:
    with patch.dict("os.environ", {"NARWHALS_DEFAULT_BACKENDS": "pandas,polars[eager]"}):
        assert _default_backend_ids() == ["pandas", "polars[eager]"]
```

src/narwhals/testing/pytest_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+    # Escape hatch for downstream test suites that ship their own backend plugin.
+    # When set, this plugin still adds the CLI options but stops parametrising the fixtures.
+    group.addoption(
+        "--use-external-nw-backend",
+        action="store_true",
+        default=False,
+        help=(
+            "Skip narwhals-testing's parametrisation and let another plugin "
+            "provide the `nw_*frame_constructor` fixtures."
+        ),
```
EdAbati, line 86: Do we have a guide with an example on how to do this for plugin providers?
am I missing it?

src/narwhals/testing/pytest_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+    from narwhals.testing.constructors import available_cpu_backends, prepare_backends
+
+    _available_cpu_backends = available_cpu_backends()
+
+    # Intersection is needed otherwise prepare_backends raises an exception if
+    # modin or pyspark are not installed
+    _all_cpu_exclusions = frozenset({"modin", "pyspark[connect]"}).intersection(
+        _available_cpu_backends
+    )
+
```
EdAbati: this interection feels a bit of a hidden workaround

what if we add a `default_opt_in: bool = True` to the `frame_constructor` which will be `False` for modin and pyspark connect?

also `available_cpu_backends` could just be a `available_default_cpu_backends` and skip modin and pyspark connect by default?

what do you think?
I could make a PR to your branch to show you what I mean

src/narwhals/testing/pytest_plugin.py
```diff
@@ -0,0 +1,144 @@
...
+        ids = [c.name for c in params]
+        metafunc.parametrize("nw_dataframe", params, ids=ids)
+    elif "nw_frame" in fixturenames:
+        metafunc.parametrize("nw_frame", selected, ids=[c.name for c in selected])
+    elif "nw_pandas_like_frame" in fixturenames:
+        params = [c for c in selected if c.is_eager and c.is_pandas_like]
+        ids = [c.name for c in params]
+        metafunc.parametrize("nw_pandas_like_frame", params, ids=ids)
+    else:  # pragma: no cover
+        ...
```
EdAbati: (reposting because Github forgot about this)

1. Small bug in `nw_lazyframe` which parametrises **`nw_dataframe`** 😮 . should we add a test in our codebase that uses this fixutre? I don't think we even use it :D I couldn't find it
2. If I read this correctly essentially here we are saying that a user cannot set 2 `nw_*` fixtures in the same function, right? Which it makes sense but I think we could have a more specific error

What do you think about something like:


```suggestion
def pytest_generate_tests(metafunc: pytest.Metafunc) -> None:
    if metafunc.config.getoption("use_external_nw_backend"):  # pragma: no cover
        return

    fixture_filters: dict[str, Callable[[FrameConstructor], bool]] = {
        "nw_dataframe": lambda c: c.is_eager,
        "nw_lazyframe": lambda c: not c.is_eager,
        "nw_frame": lambda _: True,
        "nw_pandas_like_frame": lambda c: c.is_eager and c.is_pandas_like,
    }

    fixturenames = set(metafunc.fixturenames)
    matched_fixtures = fixturenames & fixture_filters.keys()
    if not matched_fixtures:
        return
    if len(matched_fixtures) > 1:
        msg = f"A test may only request one nw frame fixture, got: {matched_fixtures}"
        raise ValueError(msg)

    selected = _select_backends(metafunc.config)

    fixture_name = next(iter(matched_fixtures))
    filter_fn = fixture_filters[fixture_name]
    params = [c for c in selected if filter_fn(c)]
    metafunc.parametrize(fixture_name, params, ids=[c.name for c in params])
```

src/narwhals/stable/v1/typing.py
EdAbati: (in case we need to add it, it may make sense to ship these typing related changes separately so they get merged sooner)

src/narwhals/testing/pytest_plugin.py
EdAbati: see here https://github.com/narwhals-dev/narwhals/pull/3673

src/narwhals/testing/pytest_plugin.py
FBruzzesi: I guess we should document this in https://narwhals-dev.github.io/narwhals/extending/ with proper examples from the daft plugin: https://github.com/narwhals-dev/narwhals-daft/blob/main/pytest_constructor_override.py.

I would keep the documentation improvement unrelated/out of scope of these changes

src/narwhals/testing/pytest_plugin.py
FBruzzesi: Nice catch!

FBruzzesi: Applied in https://github.com/narwhals-dev/narwhals/pull/3552/commits/62ba0074b6bb8f6413e8e7c40440fb3350d6a1c9

src/narwhals/testing/pytest_plugin.py
FBruzzesi: Thanks, added in https://github.com/narwhals-dev/narwhals/pull/3552/commits/1db05c9f956b25e1374057f26b95c08b87d93a9c

src/narwhals/stable/v1/typing.py
FBruzzesi: reverted

.github/workflows/pytest.yml
```diff
@@ -32,7 +33,12 @@ jobs:
           cache-dependency-glob: "pyproject.toml"
       - name: Run pytest
         shell: bash
-        run: make run-ci DEPS="--group tests --group plugins --extra pandas --extra polars --extra pyarrow" CMD="pytest tests --cov=src --cov=tests --cov-fail-under=75 --constructors=pandas,pyarrow,polars[eager],polars[lazy]"
+        env:
+          # coverage's execv/fork patches raise on Windows; collapse to `subprocess`
+          # there (coverage dedupes) and keep the default values on Linux.
+          COVERAGE_PATCH_EXECV: ${{ matrix.os == 'windows-latest' && 'subprocess' || 'execv' }}
+          COVERAGE_PATCH_FORK: ${{ matrix.os == 'windows-latest' && 'subprocess' || 'fork' }}
```
dangotbanned: Since I'm unlucky enough to be a Windows user - is this something I'd need to be concerned about locally?

If the answer is yes, could this be handled ([example]) in our `Makefile` instead please? 🙏


[example]: https://github.com/pola-rs/polars/blob/fe2040d1dbb6ea0d8c0338e28f99089487e29b36/Makefile#L11-L15

src/narwhals/stable/v1/typing.py
```diff
@@ -8,7 +8,7 @@
 if TYPE_CHECKING:
     from typing import TypeAlias
 
-    from narwhals._native import NativeDataFrame, NativeLazyFrame
+    from narwhals._native import NativeDataFrame, NativeDuckDB, NativeLazyFrame
```
dangotbanned: There's already a runtime import a few lines up 😄 

```suggestion
    from narwhals._native import NativeDataFrame, NativeLazyFrame

```

tests/expr_and_series/fill_nan_test.py
```diff
@@ -36,7 +23,7 @@ def test_fill_nan(request: pytest.FixtureRequest, constructor: Constructor) -> N
     assert_equal_data(result, expected)
     assert result.lazy().collect()["float_na"].null_count() == 2
     result = df.select(nw.all().fill_nan(3.0))
-    if any(constructor is c for c in NON_NULLABLE_CONSTRUCTORS):
+    if not constructor.is_nullable:
```
dangotbanned: This is a HUGE readability improvement! 🥳

## Suggestion

One tweak I'd make is related to, this right below:

```py
        # no nan vs null distinction
```

And most usage being for the negative case, like:

```py
if not constructor.is_nullable: ...
if not constructor_eager.is_nullable: ...
```

What if we stole an idea from [`pyarrow.compute.is_null`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.is_null.html#pyarrow.compute.is_null)?

> Whether floating-point NaN values are considered null.

```py
if constructor.nan_is_null: ...
if constructor_eager.nan_is_null: ...
```

That would remove the need for the comments 😏

tests/frame/sample_test.py
```diff
@@ -30,11 +28,11 @@ def test_sample_fraction(constructor_eager: ConstructorEager) -> None:
...
+    df = constructor_eager({"a": range(size)})
 
     r1 = nw.to_native(df.sample(n=n, seed=123))
     r2 = nw.to_native(df.sample(n=n, seed=123))
     r3 = nw.to_native(df.sample(n=n, seed=42))
 
-    assert r1.equals(r2)  # type: ignore[attr-defined]
-    assert not r1.equals(r3)  # type: ignore[attr-defined]
+    assert r1.equals(r2)
+    assert not r1.equals(r3)
```
dangotbanned, line 75: This seems like a regression to me?

The `[attr-defined]` was because because the test used a method not defined here:

https://github.com/narwhals-dev/narwhals/blob/200baa7fbbbc33ecbffef4fa1cf759d0ac050477/src/narwhals/_native.py#L191-L198

I haven't checked locally, but is there an `Any` now somewhere? 🤔

tests/frame/schema_test.py
```diff
@@ -581,7 +581,7 @@ def origin_pandas_like(
         "d": [5.3, 4.99],
         "e": [datetime(2006, 1, 1), datetime(2001, 9, 3)],
     }
-    return constructor_pandas_like(data).dtypes.to_dict()
+    return constructor_pandas_like(data).to_native().dtypes.to_dict()  # type: ignore[no-any-return]
```
dangotbanned, line 584: `constructor_pandas_like(data).to_native()` should not be returning `Any`.

IIUC, that happens because of this guy

https://github.com/narwhals-dev/narwhals/blob/91a0fd498fa5aad413aae6d4e0ef26bf4915291f/tests/conftest.py#L163-L167

src/narwhals/testing/constructors.py
```diff
@@ -0,0 +1,723 @@
...
+        self: frame_constructor[IntoFrame],
+        obj: Data,
+        /,
+        namespace: NarwhalsNamespace,
+        **kwds: Any,
+    ) -> DataFrame[Any] | LazyFrame[Any]: ...
+
+    def __call__(
+        self, obj: Data, /, namespace: NarwhalsNamespace, **kwds: Any
+    ) -> DataFrame[Any] | LazyFrame[Any]:
```
dangotbanned, line 196: I'm not sure `namespace: NarwhalsNamespace` is gonna work for typing.

### Problem 1
The return type is always using the `narwhals.<...>` variants, so we get false-positives like this:

https://github.com/narwhals-dev/narwhals/blob/91a0fd498fa5aad413aae6d4e0ef26bf4915291f/tests/v1_test.py#L472

I'm kinda confused that I couldn't find any other cases like that, I thought we remove more api?

### Problem 2
We can't add `@overload`s  to fix that with the current definition of `NarwhalsNamespace`.
There isn't anything to match here

https://github.com/narwhals-dev/narwhals/blob/91a0fd498fa5aad413aae6d4e0ef26bf4915291f/src/narwhals/testing/typing.py#L28-L31

### Problem 3
If we did fix that [^1], we also need to grow another set of `@overload`s for `v1` and `v2`


[^1]: Either in the `NarwhalsNamespace` protocol, or by using the `Version` enum

tests/conftest.py
```diff
@@ -392,3 +121,64 @@ def non_nested_type(request: pytest.FixtureRequest) -> type[NonNestedDType]:
...
     return dtype
+
+
+# The following fixtures are aliases of those registered in `narwhals/testing/pytest_plugin.py`,
+# wrapped so that calling them without an explicit `namespace` defaults to the main
+# `narwhals` namespace. Tests can still pass `nw_v1` / `nw_v2` explicitly to opt in
+# to a stable namespace; the legacy pattern `nw.from_native(constructor(data))` keeps
+# working because `nw.from_native` is idempotent on narwhals objects.
+# TODO(FBruzzesi): Drop these aliases once every test calls `nw_frame` / `nw_dataframe`
+# directly with an explicit namespace.
```
dangotbanned, line 133: I can see why having an *"explicit namespace"* might be desirable for our internal tests.

However, *unless* you can show me that it is common for them to **test** multiple narwhals versions - I'm against making it a requirement for downstream to start worrying about it.

It should be simpler than this to ask for a `DataFrame`:

```py
import narwhals as nw

from narwhals.tests.typing import DataFrameConstructor

def test_something(nw_dataframe: DataFrameConstructor) -> None:
    df = nw_dataframe({"a": [1.0, 2.0, 4.0]}, nw)
```

dangotbanned: I'd also like to add that we'got 14 cases that use `namespace` in all of `v1_test` and `v2_test`.

dangotbanned: ### Alternatives
- `namespace` is optional (easiest)
- use different fixtures for different namespaces 
  - e.g use the version as a pre/suffix
- make the version(s) an option on the CLI
  - probably the most work, but would be explicit and avoid adding noise to each test

tests/translate/from_native_test.py
```diff
@@ -539,7 +539,7 @@ def test_eager_only_pass_through_main(constructor: Constructor) -> None:
     assert not isinstance(r3, nw.LazyFrame)
 
     with pytest.raises(TypeError, match=r"Cannot.+use.+eager_only"):
-        nw.from_native(df, eager_only=True, pass_through=False)  # type: ignore[type-var]
+        nw.from_native(df, eager_only=True, pass_through=False)
```
dangotbanned, line 542: Typing regression, we should still be warning if we raise

tests/series_only/is_sorted_test.py
```diff
@@ -16,7 +16,7 @@
 )
 def test_is_sorted(
     constructor_eager: ConstructorEager,
-    input_data: str,
+    input_data: list[int],
```
dangotbanned, line 19: 😂 nice catch

tests/preserve_pandas_like_columns_name_attr_test.py
```diff
@@ -24,7 +22,7 @@ def test_ops_preserve_column_index_name(
         request.applymarker(pytest.mark.xfail)
 
     data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8.0, 9.0]}
-    df_native = constructor(data)
+    df_native = constructor(data).to_native()
     df_native.columns.name = "foo"
```
dangotbanned, line 26: This should be warning now that `Callable[..., pd.DataFrame]` was removed 🤔

src/narwhals/testing/typing.py
```diff
@@ -0,0 +1,31 @@
...
+FrameConstructor: TypeAlias = "frame_constructor[IntoFrame]"
+"""Type alias for a constructor that returns a native eager or lazy frame."""
+
+DataFrameConstructor: TypeAlias = "frame_constructor[IntoDataFrame]"
+"""Type alias for a constructor that returns an eager native dataframe."""
+
+LazyFrameConstructor: TypeAlias = "frame_constructor[IntoLazyFrame]"
+"""Type alias for a constructor that returns a lazy native frame."""
+
+Data: TypeAlias = dict[str, Any]  # TODO(Unassined): This should have a better annotation
```
dangotbanned, line 24: Here you go 😉

```suggestion
Data: TypeAlias = dict[str, list[Any]]
```

https://github.com/narwhals-dev/narwhals/blob/200baa7fbbbc33ecbffef4fa1cf759d0ac050477/tests/conftest.py#L36

tests/dependencies/is_into_dataframe_test.py
```diff
@@ -42,7 +42,7 @@ def join(self, *args: Any, **kwargs: Any) -> Any: ...
 
 
 def test_is_into_dataframe(constructor: Constructor) -> None:
-    native_frame = constructor(data)
+    native_frame = constructor(data).to_native()
```
dangotbanned, line 45: These tests now all have `Any` btw

image

tests/expr_and_series/list/get_test.py
```diff
@@ -45,9 +45,8 @@ def test_get_series(
             pytest.skip()
         pytest.importorskip("pyarrow")
 
-    if (
-        constructor_eager.__name__.startswith("pandas")
-        and "pyarrow" not in constructor_eager.__name__
+    if str(constructor_eager).startswith("pandas") and "pyarrow" not in str(
+        constructor_eager
```
dangotbanned, line 49: I'm leaving this comment as something to *refer to* in the PR description. [^1]

[^1]: you don't need to change anything

This is an example of the *partial migration* that is being done to **keep the diff smaller** than it would've been.

It does mean that we don't get the full benefits *immediately* - but we started at **257** files (https://github.com/narwhals-dev/narwhals/pull/3552#issuecomment-4260737566) and ATOW the PR sits at a more reasonable **73**

tests/expr_and_series/corr_test.py
```diff
@@ -25,7 +25,7 @@ def test_corr_expr(
     output_name: str,
     a: str | nw.Expr,
     b: str | nw.Expr,
-    expected_corr: float,
+    expected_corr: float | None,
```
dangotbanned, line 28: This is correct, but the only changes in the file (now) are these annotation fixes

tests/expr_and_series/is_close_test.py
```diff
@@ -12,12 +12,6 @@
 
 import narwhals as nw
 from narwhals.exceptions import ComputeError, InvalidOperationError
-from tests.conftest import (
-    dask_lazy_p1_constructor,
-    dask_lazy_p2_constructor,
-    modin_constructor,
-    pandas_constructor,
-)
 from tests.utils import (
```
dangotbanned, line 15: 😍

tests/expr_and_series/is_finite_test.py
```diff
@@ -77,7 +64,7 @@ def test_is_finite_column_with_null(constructor: Constructor, data: list[float])
     result = df.select(nw.col("a").is_finite())
 
     expected: dict[str, list[Any]]
-    if any(constructor is c for c in NON_NULLABLE_CONSTRUCTORS):
+    if not constructor.is_nullable:
```
dangotbanned: this file is great

tests/expr_and_series/over_test.py
```diff
@@ -472,10 +472,10 @@ def test_over_quantile(constructor: Constructor, request: pytest.FixtureRequest)
     data = {"a": [1, 2, 3, 4, 5, 6], "b": ["x", "x", "x", "y", "y", "y"]}
 
     quantile_expr = nw.col("a").quantile(quantile=0.5, interpolation="linear")
-    native_frame = constructor(data)
+    native_frame = constructor(data).to_native()
 
     if "dask" in str(constructor):
-        native_frame = native_frame.repartition(npartitions=1)  # type: ignore[union-attr]
+        native_frame = native_frame.repartition(npartitions=1)
```
dangotbanned, line 478: typing regression

tests/frame/interchange_native_namespace_test.py
```diff
@@ -9,10 +9,7 @@
 pytest.importorskip("polars")
 import polars as pl
 
-if TYPE_CHECKING:
-    from collections.abc import Mapping
-
-data: Mapping[str, Any] = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}
+data: dict[str, list[Any]] = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}
```
dangotbanned: You can avoid the `typing` import entirely with this:

```suggestion
data = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}
```

Or use a bare `Final` (`data: Final = ...`) to get the `list` inferred

tests/frame/interchange_select_test.py
```diff
@@ -1,16 +1,13 @@
...
 import pytest
 
 import narwhals as nw
 import narwhals.stable.v1 as nw_v1
 
-if TYPE_CHECKING:
-    from collections.abc import Mapping
-
-data: Mapping[str, Any] = {"a": [1, 2, 3], "b": [4.0, 5.0, 6.1], "z": ["x", "y", "z"]}
+data: dict[str, list[Any]] = {"a": [1, 2, 3], "b": [4.0, 5.0, 6.1], "z": ["x", "y", "z"]}
```
dangotbanned:
```suggestion
data = {"a": [1, 2, 3], "b": [4.0, 5.0, 6.1], "z": ["x", "y", "z"]}
```

tests/frame/schema_test.py
```diff
@@ -606,7 +606,7 @@ def origin_pandas_like_pyarrow(
     df_nw = nw.from_native(df_pd).with_columns(
         nw.col("f").cast(nw.Date()), nw.col("g").cast(nw.Time())
     )
-    return df_nw.to_native().dtypes.to_dict()
+    return df_nw.to_native().dtypes.to_dict()  # type: ignore[no-any-return]
```
dangotbanned, line 609: Same as https://github.com/narwhals-dev/narwhals/pull/3552#discussion_r3373247728

tests/expr_and_series/arithmetic_test.py
```diff
@@ -258,7 +254,7 @@ def test_std_broadcating(constructor: Constructor) -> None:
     if "duckdb" in str(constructor) and DUCKDB_VERSION < (1, 3):
         # `std(ddof=2)` fails for duckdb here
         pytest.skip()
-    df = nw.from_native(constructor({"a": [1, 2, 3]}))
+    df = constructor({"a": [1, 2, 3]}, nw)
```
dangotbanned, line 257: *Not changed by this PR* but the test name made me chuckle

https://github.com/narwhals-dev/narwhals/blob/91a0fd498fa5aad413aae6d4e0ef26bf4915291f/tests/expr_and_series/arithmetic_test.py#L253

tests/testing/assert_frame_equal_test.py
```diff
@@ -24,12 +24,12 @@ def _assertion_error(detail: str) -> pytest.RaisesExc:
...
 def test_check_narwhals_objects(constructor: Constructor) -> None:
     """Test that a type error is raised if the input is not a Narwhals object."""
-    frame = constructor({"a": [1, 2, 3]})
+    frame = constructor({"a": [1, 2, 3]}).to_native()
     msg = re.escape(
         "Expected `narwhals.DataFrame` or `narwhals.LazyFrame` instance, found"
     )
     with pytest.raises(TypeError, match=msg):
-        assert_frame_equal(frame, frame)  # type: ignore[arg-type]
+        assert_frame_equal(frame, frame)
```
dangotbanned: We lost a `# type: ignore[arg-type]`

tests/testing/assert_frame_equal_test.py
```diff
@@ -41,9 +41,8 @@ def test_implementation_mismatch() -> None:
     import pyarrow as pa
 
     with _assertion_error("implementation mismatch"):
-        assert_frame_equal(
-            nw.from_native(pd.DataFrame({"a": [1]})),
-            nw.from_native(pa.table({"a": [1]})),  # type: ignore[type-var] # pyright: ignore[reportArgumentType]
+        assert_frame_equal(  # type: ignore[type-var]
+            nw.from_native(pd.DataFrame({"a": [1]})), nw.from_native(pa.table({"a": [1]}))
         )
```
dangotbanned: Could you split up the constructor call(s) and the `assert_frame_equal` please?

The `# *ignore`s have changed and it's easier to get them precise with less on the same line 🙂

tests/translate/from_native_test.py
```diff
@@ -294,10 +294,10 @@ def test_eager_only_lazy_dask(eager_only: Any, context: Any) -> None:
 
 def test_series_only_sqlframe() -> None:  # pragma: no cover
     pytest.importorskip("sqlframe")
-    df = sqlframe_pyspark_lazy_constructor(data)
+    df = get_backend_constructor("sqlframe")(data, nw).to_native()
 
     with pytest.raises(TypeError, match="Cannot only use `series_only`"):
-        nw.from_native(df, series_only=True)  # pyright: ignore[reportArgumentType, reportCallIssue]  # pyrefly: ignore[no-matching-overload]
+        nw.from_native(df, series_only=True)  # type: ignore[call-overload]  # pyrefly: ignore[no-matching-overload]
```
dangotbanned, line 300: I think this ignore change is expected, but the test is named `test_series_only_sqlframe` and `df` is now inferred less precisely (`IntoLazyFrame`)

Maybe a follow-up would be that `get_backend_constructor("sqlframe")` is `@overload`-ed in a *similar* way to [`Namespace`](https://github.com/narwhals-dev/narwhals/blob/200baa7fbbbc33ecbffef4fa1cf759d0ac050477/tests/namespace_test.py)?

tests/ibis_test.py
```diff
@@ -1,30 +1,14 @@
...
-    return func
+from narwhals.testing.constructors import get_backend_constructor
 
 
-def test_from_native(ibis_constructor: Constructor) -> None:
-    df = nw.from_native(ibis_constructor({"a": [1, 2, 3], "b": [4, 5, 6]}))
+def test_from_native() -> None:
+    ibis_constructor = get_backend_constructor("ibis")
+    if not ibis_constructor.is_available:
+        pytest.skip()
```
dangotbanned: 1. Does this hook the test into the include/exclude filters?
2. Could we inline an `pytest.importorskip("ibis")` to trigger inside `get_backend_constructor("ibis")`?

dangotbanned (commented):
Thanks @FBruzzesi for yet another wild PR 😘



> [!NOTE]
> As mentioned privately, this is a partial review to keep things moving.
> I've reviewed **61/73** files and the remaining one's are below.
> **I'm planning to approve this PR**, but I will need to finish reviewing the rest

<details><summary><b>Show unreviewed files</b></summary>
<p>

Ah you looked 🫣 

Okay so, this is most of the new stuff.
I will get to that - but addressing my comments will likely change *something* here:

image


</p>
</details>

## TL;DR
1. Consider deferring the handling of `namespace`/`Version` to another PR
2. Please try to keep existing `# *ignore` comments *as-is* [^1] or start a (comment) thread explaining why the change is positive [^2].

Request changes


[^1]: I think filling in some `Any`(s) in the new typing will resolve most/all of that

[^2]: Not all `# type: ignore`s are equal, and a lot that were removed were in tests that explicitly check invalid operations.

.github/workflows/pytest.yml
FBruzzesi: Addressed in https://github.com/narwhals-dev/narwhals/pull/3552/commits/ef3a00a21b7453f8f26f4c95ea6d50f5b1d8b008

src/narwhals/stable/v1/typing.py
FBruzzesi: Applied in https://github.com/narwhals-dev/narwhals/pull/3552/commits/81f0c7362439061997ba7864f09f87d911722c8d

src/narwhals/testing/typing.py
FBruzzesi: Well that actually lead to quite a few typing errors, so I didn't dare to touch it

tests/frame/schema_test.py
FBruzzesi: Yes that should indeed be the case for now! That's also the reason for some other comments you had below: having to move `.to_native()` leads to have `Any` object to deal with and therefore some typing regression. I will add a comment

src/narwhals/testing/constructors.py
FBruzzesi: I will seek for help for how to do this 🧐

tests/conftest.py
FBruzzesi: The latter two options are not ideal in my opinion because they limit what you can end up using _within_ the same test. So option 1 is that's left?!

tests/preserve_pandas_like_columns_name_attr_test.py
FBruzzesi: `df_native` is `Any`

tests/conftest.py
FBruzzesi: The other alternative would be to rollback the changes to return a narwhals object 🥹

PR #3556, opened 2026-04-18
refactor: Rename exposed fixtures and pytest options

FBruzzesi:
# Description

This PR addresses the section from #3552:

>  In light of the CI failures for downstream projects, I also realized that having fixtures called constructor and constructor_eager is not really intuitive. I am happy to brainstorm better names for downstream users. And similarly the flags --all-cpu-constructors, --constructors. (*)

with the following renaming:

* fixtures: `nw_frame_constructor` (eager + lazy), `nw_eager_constructor` (eager only), `nw_pandas_like_constructor` (pandas like only).
* pytest options: `--nw-backends`, `--all-nw-backends`, `--use-external-nw-backend` 

---

I know that `nw.from_native(nw_frame_constructor(data))` is extremely verbose, but bear with me - I am also proposing to directly return `nw.LazyFrame` or `nw.DataFrame` from the `nw_frame_constructor` fixture.

From #3552:

> If we find a way to achieve the following points, then I personally think we should do so, as we almost always run `nw.from_native` right after.
>
>  * Distinguish between native frame and `nw.from_native` kwargs -> For this it's enough to provide it as different proper arguments in `__call__`, e.g. `nw_kwargs`, `backend_kwargs`.
>  * How do we know which version a user would use? Should we require to pass either the `Version.X` or the namespace `{nw, nw_v1, nw_v2}` ? If we do so, I would argue that we must enforce it and never assume for the user. In my opinion `nw_frame_constructor(data, version=nw)` would still look better than `nw.from_native(nw_frame_constructor(data))`
 
---

TODO:

- [ ] Update Docs

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


dangotbanned:
> I know that `nw.from_native(nw_frame_constructor(data))` is extremely verbose, but bear with me

Agreed, IMO this was *already* too much ceremony.
Like this [blows up](https://github.com/narwhals-dev/narwhals/blob/e0892a23db428c11dc556dac6a54c5dde761d6f0/tests/frame/select_test.py#L31) to almost double the size 😰 

```py
result = pl.DataFrame({"a": [1, 2, 3]}).select()
result = nw.from_native(constructor_eager({"a": [1, 2, 3]}), eager_only=True).select() 
```

> I am also proposing to directly return `nw.LazyFrame` or `nw.DataFrame` from the `nw_frame_constructor` fixture.

👍 from me on skipping `from_native`.

> [!NOTE]
> you probably already knew I was on board, I just wanted to make it more visible


---

To bikeshed on the naming some more:

1. Is the `nw_` prefix to avoid clobbering the fixture namespace? 
   If so, `pytest` is pretty flexible in [*Overriding fixtures on various levels*](https://docs.pytest.org/en/stable/how-to/fixtures.html#overriding-fixtures-on-various-levels).
   Just seems like a problem that `pytest` would have a solution for hidden *somewhere* 😄
2. If all of our fixtures are suffixed by `_constructor`, is that part not redundant?

Combining those two points, you'd get something like this, where the concepts are embedded in the types, rather than the names:

<details><summary>Mandatory plug for <code>tests.plan</code></summary>
<p>

Made you click 😉 

I'm still missing a `frame` fixture, but that would come eventually

https://github.com/narwhals-dev/narwhals/blob/a09be54ec5d97225110418a79843c1fbbc933a73/tests/plan/conftest.py#L22-L27

https://github.com/narwhals-dev/narwhals/blob/a09be54ec5d97225110418a79843c1fbbc933a73/tests/plan/utils.py#L565-L572

</p>
</details> 



You can still see more detail in an IDE if you want it:

<details><summary>Show IDE screenshot</summary>
<p>

image



</p>
</details> 


Although another option is using a suffix for the *type of the fixture*, e.g.:

```py
lazyframe: LazyFrame
lazyframe: LazyFrameFixture
```

Then you can alias the import if you want something shorter 😎

FBruzzesi:
Thanks for your feedback @dangotbanned

> If all of our fixtures are suffixed by _constructor, is that part not redundant?

That's a valid point! Renamed to `nw_frame`, `nw_dataframe` and `nw_lazyframe`, although as of now they don't return narwhals objects. I have yet another branch to tackle that

> Is the nw_ prefix to avoid clobbering the fixture namespace?

Yes indeed that's the reason

> If so, pytest is pretty flexible in [Overriding fixtures on various levels](https://docs.pytest.org/en/stable/how-to/fixtures.html#overriding-fixtures-on-various-levels).
> Just seems like a problem that pytest would have a solution for hidden somewhere 😄

I am not sure that it would work with external fixtures fairlearn and plotly raise  `duplicate parametrization of 'constructor'`

PR #3562, opened 2026-04-21
fix(docs): Include dask and ibis in docs dependencies

FBruzzesi:
# Description

Required to properly build api-completeness pages

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -32,7 +32,7 @@ jobs:
         uses: astral-sh/setup-uv@v7
       - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
       - name: Install dependencies
-        run: uv pip install -e . --group docs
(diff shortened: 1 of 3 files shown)


MarcoGorelli:
thanks @FBruzzesi !

perhaps should we just add them to

https://github.com/narwhals-dev/narwhals/blob/e0892a23db428c11dc556dac6a54c5dde761d6f0/.github/workflows/deploy-docs.yml#L34-L35

if they're only needed for a single page (and it still renders fine without them)?

then if i need to build the docs locally, `uv pip install -e . --group docs` can still be relatively lightweight

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #3569, opened 2026-04-24
feat: Let constructor fixtures return narwhals object directly

FBruzzesi:
# Description

Alright, at this point I briefly mention this branch to all of you @MarcoGorelli @dangotbanned @camriddell so I might as well open the PR.

I _think_ it is promising, but I might need some help for the typing, here is a repro for the issue I am facing:

```py

from typing import TYPE_CHECKING, reveal_type

import narwhals as nw

if TYPE_CHECKING:
    from narwhals.testing.typing import FrameConstructor


def test_typing(nw_frame: FrameConstructor) -> None:
    data = {"a": [-1, 2, 3], "b": [1, 1, 2], "i": [0, 1, 2]}
    df = nw_frame(data, nw).lazy()
    reveal_type(df)  # -> LazyFrame[Any]

    native = df.to_native()
    reveal_type(native)  # -> Any
    
    frame = nw.from_native(native)
    reveal_type(frame)  # ->  Any
```

which seems far from ideal! I didn't manage to get the `__call__` overload quite right, but it should not be far from the correct solution. I tried a few other variations without success.

---

Regarding the patches in `conftest.py`:

1. I wouldn't worry too much for their typing, they are just temporary workaround to avoid touching every single test
2. The hack is needed for the same reason, namely make `narwhals` (main namespace) the default namespace to the `__call__` method and avoid having to pass it everywhere. This works since `nw.from_native` can also take a narwhals object


FBruzzesi:
> ```python
> from __future__ import annotations
> 
> from typing import TYPE_CHECKING, reveal_type
> 
> import narwhals as nw
> 
> if TYPE_CHECKING:
>     from narwhals.testing.typing import FrameConstructor
> 
> 
> def test_typing(nw_frame: FrameConstructor) -> None:
>     data = {"a": [-1, 2, 3], "b": [1, 1, 2], "i": [0, 1, 2]}
>     df = nw_frame(data, nw).lazy()
>     reveal_type(df)  # -> LazyFrame[Any]
> 
>     native = df.to_native()
>     reveal_type(native)  # -> Any
>     
>     frame = nw.from_native(native)
>     reveal_type(frame)  # ->  Any
> ```

I realized that's actually a false alarm, as such behavior is driven by `.lazy()`, which returns `LazyFrame[Any]`

---

I might give it a shot in another PR

PR #3575, opened 2026-04-25
docs: Add note on return type of columns property

### LINKED ISSUE #3571: [Bug]: narwhals.DataFrame.columns does not always return list of strings
### Describe the bug

Per API reference, the `.columns` property of a narwhals dataframe returns a list of strings. This is not true - if on purpose or not is a different question.

```py
import narwhals as nw
import pandas as pd

df = pd.DataFrame([[1, 2], [4, 5], [5, 6]], columns=[1, 2])
c = nw.from_native(df).columns
assert isinstance(c[0], str) 
```

It even allows mixed column names:
```py
df = pd.DataFrame([[1, 2], [4, 5], [5, 6]], columns=[1, "a"])
cols = nw.from_native(df).columns
[print(f"{c} has type {type(c)}") for c in cols]
```

### Expected results

The assertion passes, i.e. the API contract and type hints hold.

### Actual results

`AssertionError`

```shell
narwhals: 2.20.0
```

### Relevant log output

```shell

```

FBruzzesi:
# Description

1. Do you like the phrasing?
2. Should we add the same note for the `Schema` spec? For the same reason we cannot guarantee it to be a Mapping from str to DType.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3571

### DIFF
--- a/narwhals/dataframe.py
+++ b/narwhals/dataframe.py
@@ -1353,6 +1353,13 @@ class DataFrame(BaseFrame[DataFrameT]):
     def columns(self) -> list[str]:
         """Get column names.
 
+        Note:
+            The pandas-like and dask backends allow non-string column names
+            (e.g. integers or booleans). While discouraged, this is supported,
+            so the return type is not guaranteed to be `list[str]`.
+
+            See [concepts - column names](../concepts/column_names.md) for details.
+
         Examples:
             >>> import pyarrow as pa
             >>> import narwhals as nw
@@ -2637,6 +2644,13 @@ class LazyFrame(BaseFrame[LazyFrameT]):
     @property
     def columns(self) -> list[str]:
         r"""Get column names.
+
+        Note:
+            The pandas-like and dask backends allow non-string column names
+            (e.g. integers or booleans). While discouraged, this is supported,
+            so the return type is not guaranteed to be `list[str]`.
+
+            See [concepts - column names](../concepts/column_names.md) for details.
 
         Examples:
             >>> import duckdb
--- a/docs/concepts/column_names.md
+++ b/docs/concepts/column_names.md
@@ -1,6 +1,7 @@
 # Column names
 
-Polars and PyArrow only allow for string column names. What about pandas?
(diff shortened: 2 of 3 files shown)


narwhals/dataframe.py
```diff
@@ -1353,6 +1353,13 @@ def collect_schema(self) -> Schema:
     def columns(self) -> list[str]:
         """Get column names.
 
+        Note:
+            The pandas-like and dask backends allow non-string column names
+            (e.g. integers or booleans). While discouraged, this is supported,
+            so the return type is not guaranteed to be `list[str]`.
```
EdAbati, line 1359: I think the phrasing looks good!

I tried to find alternatives.

For the 2nd sentence we can use something like:
`This is discouraged but supported, so the return type is not guaranteed to be `list[str]`.`

Or have it all one sentence:

`Non-string column names (e.g. integers or booleans) are discouraged but supported by pandas-like and Dask backends, so the return type is not guaranteed to be `list[str]`.`

But again, I think your version looks good :)

EdAbati:
> Should we add the same note for the Schema spec? For the same reason we cannot guarantee it to be a Mapping from str to DType.

Do you mean in `schema`/`collect_schema`? I think think it'd be helpful  👍

We could also update the `column_names.md` to mention Dask too

FBruzzesi:
Thanks for the feedback @EdAbati 

> Do you mean in `schema`/`collect_schema`? I think think it'd be helpful 👍

Both `schema` and `collect_schema` return a `Schema` object. I added similar notes in such class in https://github.com/narwhals-dev/narwhals/pull/3575/commits/beb053a510e49bada9c9a0adadf644d5c5c988e7

> We could also update the `column_names.md` to mention Dask too

Updated the doc page in https://github.com/narwhals-dev/narwhals/pull/3575/commits/934d15db0134021b2b00fafacc85c73021834a52

EdAbati (approved):
Nice

PR #3587, opened 2026-05-01
chore: uppercase `no_default` sentinel

MarcoGorelli:
this avoids https://github.com/facebook/pyrefly/issues/3042


MarcoGorelli:
no way

sigh..

```py
tubular/nominal.py:11: in <module>
    from narwhals._utils import no_default  # noqa: PLC2701, need private import
```

PR #3609, opened 2026-05-09
ci: Add `scikit-learn` to downstream tests

### LINKED ISSUE #3602: CI: test scikit-learn downstream

FBruzzesi:
# Description

Opening as draft and in downstram-tests (not slow version) to see how long it does take. From a local attempt it's still below 5 mins

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Closes #3602
- #3680


MarcoGorelli:
10 mins in ci, that's fine with me!

FBruzzesi:
Thanks Marco, I will add a comment in scikit-learn PR asking if there is a better way to do this, and/or if we can use a pytest marker

dangotbanned:
Might it make sense to reintroduce [^1] this workflow for `scikit-learn`?

- https://github.com/narwhals-dev/narwhals/pull/3629

To me it looks like [`pointblank`] (7m 21s) and [`scikit-learn`] (11m 16s) are pretty big outliers [here](https://github.com/narwhals-dev/narwhals/actions/runs/27129955214/usage).
*Conceptually*, it is odd to group them with *Test Downstream Libraries - Fast* 😄

I wonder if we could speed up *some* runs by doing things like (https://github.com/narwhals-dev/narwhals/pull/2910)?
For example:
- we don't need to run the workflow if a PR only touches `docs/**`
- if they test a specific set of backends, could we look for changes in those packages to trigger?

[`pointblank`]: https://github.com/narwhals-dev/narwhals/actions/runs/27129955214/job/80068092033
[`scikit-learn`]: https://github.com/narwhals-dev/narwhals/actions/runs/27129955214/job/80068092139

[^1]: I'm not sure why `vegafusion` was removed, but assume there was a reason?

FBruzzesi:
Thanks for taking a look @dangotbanned - According to https://github.com/scikit-learn/scikit-learn/pull/31127#issuecomment-4425039398 we might run a subset of the tests.

In general I am a bit concerned/skeptical in running hardcoded tests, I would prefer having pytest markers. Happy to hear your thoughts

dangotbanned:
> In general I am a bit concerned/skeptical in running hardcoded tests, I would prefer having pytest markers. Happy to hear your thoughts

I'm not sure we're talking about the same thing.
I was interested in using filters in the workflow(s).

- https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events


Is a set of globs considered *hardcoding*?
Polars documents it here for the same motivation https://docs.pola.rs/development/contributing/ci/#design

> Only run checks when they need to be run. A change to the Rust code does not warrant a linting check of the Python code, for example.

FBruzzesi:
> I'm not sure we're talking about the same thing.

We are not 🤣

> I was interested in using filters in the workflow(s).

We can have both approaches! Filters in workflows will allow to skip workflows given specific paths. What I mentioned will enable much faster runtime _when_ it runs

dangotbanned:
> We can have both approaches!

Perfect!
Yeah both works for me

FBruzzesi:
I know it's a partial solution, but it might be worth merging to start running (fast) tests against scikit-learn.

.github/workflows/downstream_tests.yml
```diff
@@ -703,3 +703,50 @@ jobs:
         run: |
           cd LightGBM
           pytest --config-file python-package/pyproject.toml tests/python_package_test
+
+  scikit-learn:
+    if: github.head_ref != 'bump-version'
```
dangotbanned, line 708: Do we want to use anything like #3680 here?

dangotbanned (approved):
Thanks @FBruzzesi, just one comment from me 


squeeze cheeks narwhal

.github/workflows/downstream_tests.yml
FBruzzesi: The settings introduces there are all at workflow level, nothing job specific. We should be fine for now

PR #3613, opened 2026-05-11
feat: Add `is_into_lazyframe`, export missing functionalities, fix SQLFrame bug

### LINKED ISSUE #3612: feat: Add `is_into_lazyframe`
Add lazy equivalent of `narwhals.dependencies.is_into_dataframe`
 
Relevant for dlt integration, see https://github.com/dlt-hub/dlt/pull/3912#discussion_r3221444279

FBruzzesi:
# Description

Small late night rant: everything always turn out bigger than what it looks like. End of rant

---

As per title

* Introduces `is_into_lazyframe`
* Exports missing functionalities from `dependencies` modules
* Fixes SQLFrame bug (currently `is_into_lazyframe(sqlframe_object) -> True`
* Improves `_warn_if_narwhals_*` by getting the calling function name via `caller = sys._getframe(1).f_code.co_name` and replace `is_pandas_dataframe` hardcoded value

---

@dangotbanned I read your comment in #3612 when I was already down this rabbit hole  😭

- [ ] 💾 Refactor
- [x] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3612
- Related #3530


src/narwhals/stable/v1/dependencies.py
```diff
@@ -123,21 +143,115 @@ def is_pandas_like_series(ser: Any) -> bool:
...
+    from narwhals.stable.v1 import DataFrame
+
+    return (
+        isinstance(native_dataframe, DataFrame)
+        or hasattr(native_dataframe, "__narwhals_dataframe__")
+        or is_polars_dataframe(native_dataframe)
+        or is_pyarrow_table(native_dataframe)
+        or is_pandas_like_dataframe(native_dataframe)
+        or is_ibis_table(native_dataframe)
+        or is_duckdb_relation(native_dataframe)
```
FBruzzesi: 👀 that's a sad truth

narwhals/dependencies.py
```diff
@@ -547,6 +553,44 @@ def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoData
         or is_polars_dataframe(native_dataframe)
         or is_pyarrow_table(native_dataframe)
         or is_pandas_like_dataframe(native_dataframe)
+    ) and not is_sqlframe_dataframe(native_dataframe)
```
FBruzzesi: Currently in main:

```py
from narwhals.dependencies import is_into_dataframe

data = {"a": ["x", "y"], "b": [1, 2]}
session = DuckDBSession()

frame = session.createDataFrame([*zip(*data.values(), strict=False)], schema=list(data))

is_into_dataframe(frame)
```

> True

src/narwhals/stable/v1/dependencies.py
MarcoGorelli: 😄 true

BUT

we could just avoid adding this for v1? it's a new function so there's no old behaviour to preserve

FBruzzesi: Wouldn't that be straight up wrong behavior?

I don't think the following would be what a user expects:

```py

nw.dependencies.is_into_dataframe(duckdb_obj)  # False
isinstance(nw.from_native(duckdb_obj), nw.DataFrame)  # True
``

Comment (MarcoGorelli) [reply src/narwhals/stable/v1/dependencies.py]: ah right `is_into_dataframe` is already exported here in v1 😭 

ok sure looks good then 👍

**Comment by FBruzzesi** [comment]
downstream marimo is affected. I will investigate later today

File: src/narwhals/stable/v1/dependencies.py [hunk 3]
```diff
@@ -123,21 +143,115 @@ def is_pandas_like_series(ser: Any) -> bool:
...
+    if get_sqlframe() is not None:
+        from sqlframe.base.dataframe import BaseDataFrame
+
+        return isinstance(df, BaseDataFrame)
+    return False  # pragma: no cover
+
+
+def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoDataFrameT]:
+    """Check whether `native_dataframe` can be converted to a narwhals.stable.v1.DataFrame."""
+    from narwhals.stable.v1 import DataFrame
```
FBruzzesi, line 66: Thoughts on pulling `DataFrame` from v1 here? Should we do the same for v2?

src/narwhals/stable/v1/dependencies.py
FBruzzesi: So investigation I did https://github.com/narwhals-dev/narwhals/pull/3613#issuecomment-4433008556, I am in a big rabbit hole for the marimo CI failure, I ended up in [altair/vegafusion](https://github.com/vega/altair/blob/44097bdac03dedfb91416a01f12c2c65e442798b/altair/utils/_vegafusion_data.py#L52-L62):

```py
from narwhals.stable.v1.dependencies import is_into_dataframe

...

if VEGAFUSION_VERSION and Version("2.0.0a0") <= VEGAFUSION_VERSION:

    def is_supported_by_vf(data: Any) -> TypeIs[DataFrameLike]:
        # Test whether VegaFusion supports the data type
        # VegaFusion v2 support narwhals-compatible DataFrames
        return isinstance(data, DataFrameLike) or is_into_dataframe(data)

else:

    def is_supported_by_vf(data: Any) -> TypeIs[DataFrameLike]:
        return isinstance(data, DataFrameLike)
```

which would now return True for duckdb (and ibis), which is what marimo is testing for

---

Are we at a crossroad between correctness and stability?

* For marimo it would be enough to coordinate with them to regenerate test snapshots once we merge/release 
* For altair, we can suggest to move to stable.v2 if that's not an issue for them

cc @dangotbanned for the ongoing conversation in Discord regarding how altair/vega are interacting with narwhals

dangotbanned: > For altair, we can suggest to move to stable.v2 if that's not an issue for them

That should be fine 🙂 

- https://github.com/vega/altair/pull/3970#issuecomment-3999779433

narwhals/dependencies.py
```diff
@@ -547,6 +553,44 @@ def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoData
...
+        >>> is_into_lazyframe(lf_pl)
+        True
+        >>> is_into_lazyframe(np_arr)
+        False
+    """
+    from narwhals.dataframe import LazyFrame
+
+    return (
+        isinstance(native_lazyframe, LazyFrame)
+        or hasattr(native_lazyframe, "__narwhals_lazyframe__")
```
dangotbanned: We should be using `_hasattr_static` for all of these.

Here's a contrived false-positive 😉:

```py
>>> import pandas as pd
>>>
>>> df = pd.DataFrame()
>>> hasattr(df, "__narwhals_lazyframe__")
False

>>> df = pd.DataFrame({"__narwhals_lazyframe__":[]})
>>> hasattr(df, "__narwhals_lazyframe__")
True

>>> from narwhals._utils import _hasattr_static
>>> _hasattr_static(df, "__narwhals_lazyframe__")
False
```

narwhals/dependencies.py
```diff
@@ -547,6 +553,44 @@ def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoData
...
+        or is_polars_lazyframe(native_lazyframe)
+        or is_dask_dataframe(native_lazyframe)
+        or is_duckdb_relation(native_lazyframe)
+        or is_ibis_table(native_lazyframe)
+        or is_pyspark_dataframe(native_lazyframe)
+        or is_pyspark_connect_dataframe(native_lazyframe)
+        or is_sqlframe_dataframe(native_lazyframe)
```
dangotbanned: I think reusing these guards here is a potential performance footgun.

Imagine you were downstream and had a function like this:

```py
def concat(*native_frames: IntoLazyFrame) -> IntoLazyFrame:
    narwhals_frames = []
    for frame in native_frames:
        if is_into_lazyframe(frame):
            narwhals_frames.append(nw.from_native(frame))
        else:
            raise TypeError("super important message!")
    return nw.concat(narwhals_frames).to_native()
```

- We have 7 functions which call to `_warn_if_narwhals_df_or_lf`
  - that calls both `is_narwhals_dataframe` and `is_narwhals_lazyframe`
  - the former is unrelated, and the latter was imported in this function already
- We have 7 functions which check the contents of `sys.modules`
  - maybe that could be done more cheaply?
- For this `concat` example, the above is `cost * len(native_frames)` 😰

I guess my question is, would someone calling `is_into_lazyframe` expect this much is happening behind the scenes?

src/narwhals/dependencies.py
```diff
@@ -128,22 +128,28 @@ def get_sqlframe() -> Any:
     return sys.modules.get("sqlframe", None)
 
 
-def _warn_if_narwhals_df_or_lf(df: Any) -> None:
-    if is_narwhals_dataframe(df) or is_narwhals_lazyframe(df):
+def _warn_if_narwhals_df_or_lf(frame: Any, /) -> None:
+    if is_narwhals_dataframe(frame) or is_narwhals_lazyframe(frame):
+        from narwhals._utils import qualified_type_name
+
+        caller = sys._getframe(1).f_code.co_name
```
dangotbanned, line 135: Well aren't you a fancy boi 😄

dangotbanned (commented):
Thanks @FBruzzesi

I've not got much, but it's honest work 🙂 

spinny hands narwhal

narwhals/dependencies.py
dangotbanned: inb4 I fixed a bug with that (https://github.com/narwhals-dev/narwhals/pull/3613/commits/e1aeff51684c2a6dccd48903c3c40e70014f53c8)

narwhals/dependencies.py
FBruzzesi: Thanks @dangotbanned

I agree that it's _a lot_ of checkings and function calls. Yet I wonder if the scenario is realistic here: `is_into_lazyframe ` does not guarantee the same (lazy) backend, therefore `nw.concat(narwhals_frames)` risks to result in an error anyway. For performances, also keep in mind that `nw.from_native` calls the `is_*` over again.

---

That said, I don't want to focus on this specific example but on the larger picture

FBruzzesi: @dangotbanned https://github.com/narwhals-dev/narwhals/pull/3613/commits/35d0b88f2658a091970238ab936663c26b743ed0 reduces the nested `isinstance` calls for the warnings

src/narwhals/stable/v1/dependencies.py
dangotbanned: I'm not sure actually.

Usually I'd say yes, but `from_native` converts between versions now. So maybe there's not a benefit to the version distinction?

narwhals/dependencies.py
FBruzzesi: Thanks @dangotbanned - Added to other `is_into_*` as well

src/narwhals/stable/v1/dependencies.py
FBruzzesi: Hopefully I will find the time to make a PR into altair to transition to stable v2 then ✨

dangotbanned: > Here are exceptions to our backwards compatibility policy:

> - Unambiguous bugs. If a function contains what is unambiguously a bug, then we'll fix it, without considering that to be a breaking change.

I would consider `is_into_dataframe` on `main` being re-used for `v1` as a bug (our bad).

I think we've got more bugs with it though that should be addressed:


## 1. `DataFrameLike`
This is part of the deviation for `v1.typing.IntoDataFrame`, so it needs to be part of`v1.dependencies.is_into_dataframe`

https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/stable/v1/typing.py#L28-L29

## 2. `supports_dataframe_interchange`
Was going to suggest using this, but realising now that **I goofed** and used `hasattr` instead of `_hasattr_static` 🤦‍♂️ 
https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/_interchange/dataframe.py#L158-L159

We *could* leave fixing that for another PR, but in this one *at least* we shouldn't reuse that function

tests/dependencies/is_into_dataframe_test.py
```diff
@@ -26,32 +35,45 @@ def __narwhals_dataframe__(self) -> Self:  # pragma: no cover
...
 
 
-def test_is_into_dataframe_pyarrow() -> None:
-    pytest.importorskip("pyarrow")
-    import pyarrow as pa
-
-    assert is_into_dataframe(pa.table(DATA))
-
+@pytest.mark.filterwarnings("ignore:.*You passed a.*:UserWarning")
+def test_is_into_dataframe(constructor: Constructor) -> None:
```
dangotbanned, line 44: This comment kinda applies to all the guards.

In light of the `hasattr` issues, what do you think about adding a `conftest.py` fixture for detecting false positives?

Something like this could be used in a lot of places

```py
class Imposter:
    """Reveals where we needed the help of [`inspect.getattr_static`].

    `_hasattr_static` should always be preferred when working with unknown objects:

        hasattr(imposter, "i_dont_exist")          # True
        _hasattr_static(imposter, "i_dont_exist")  # False

    [`inspect.getattr_static`]: https://docs.python.org/3/library/inspect.html#inspect.getattr_static
    """

    def __getattr__(self, name: str) -> Any:
        return self


@pytest.fixture(scope="session")
def imposter() -> Imposter:
    """A fixture to help detect `hasattr` false positives."""
    return Imposter()
```

tests/dependencies/is_into_lazyframe_test.py
```diff
@@ -0,0 +1,78 @@
...
+
+class DictDataFrame:
+    def __init__(self, data: Mapping[str, Any]) -> None:
+        self._data = data
+
+    def __len__(self) -> int:  # pragma: no cover
+        return len(next(iter(self._data.values())))
+
+    def __narwhals_lazyframe__(self) -> Self:  # pragma: no cover
+        return self
```
dangotbanned: `DictDataFrame` 🤨

tests/dependencies/is_into_dataframe_test.py
```diff
@@ -26,32 +35,45 @@ def __narwhals_dataframe__(self) -> Self:  # pragma: no cover
...
+    assert not v1_is_into_dataframe(nw_frame)
+    assert not v2_is_into_dataframe(nw_frame)
 
 
 def test_is_into_dataframe_other() -> None:
     pytest.importorskip("numpy")
     import numpy as np
 
-    assert is_into_dataframe(DictDataFrame(DATA))  # pyrefly: ignore[bad-specialization]
+    assert is_into_dataframe(DictDataFrame(data))  # pyrefly: ignore[bad-specialization]
```
dangotbanned: The full output is:

```md
ERROR `DictDataFrame` is not assignable to upper bound `NativeDataFrame` of type variable `IntoDataFrameT` [bad-specialization]
```

Which we can fix with aligning `DictDataFrame` to `NativeDataFrame` 😄:

```diff
@@ -34,6 +34,16 @@ class DictDataFrame:
     def __narwhals_dataframe__(self) -> Self:  # pragma: no cover
         return self
 
+    @property
+    def columns(self) -> Any:
+        return list(self._data)
+
+    def join(self, *args: Any, **kwargs: Any) -> Any:
+        return self
+
+    def drop(self, *args: Any, **kwargs: Any) -> Any:
+        return self
+
 
 @pytest.mark.filterwarnings("ignore:.*You passed a.*:UserWarning")
 def test_is_into_dataframe(constructor: Constructor) -> None:
@@ -66,14 +76,14 @@ def test_is_into_dataframe_other() -> None:
     pytest.importorskip("numpy")
     import numpy as np
 
-    assert is_into_dataframe(DictDataFrame(data))  # pyrefly: ignore[bad-specialization]
+    assert is_into_dataframe(DictDataFrame(data))
     assert not is_into_dataframe(np.array([[1, 4], [2, 5], [3, 6]]))
     assert not is_into_dataframe(data)
 
-    assert v1_is_into_dataframe(DictDataFrame(data))  # pyrefly: ignore[bad-specialization]
+    assert v1_is_into_dataframe(DictDataFrame(data))
     assert not v1_is_into_dataframe(np.array([[1, 4], [2, 5], [3, 6]]))
     assert not v1_is_into_dataframe(data)
 
-    assert v2_is_into_dataframe(DictDataFrame(data))  # pyrefly: ignore[bad-specialization]
+    assert v2_is_into_dataframe(DictDataFrame(data))
     assert not v2_is_into_dataframe(np.array([[1, 4], [2, 5], [3, 6]]))
     assert not v2_is_into_dataframe(data)

```

tests/dependencies/is_into_lazyframe_test.py
dangotbanned: But more important than the name, is that this should be trying to satisfy this def:

https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/_native.py#L199-L200

(See https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3247362597 for example)

https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/_native.py#L189-L192

And `__len__` isn't needed at all

dangotbanned (commented):
Thanks again for taking things on board, here's round 2

NarrNarrTheNarwhalGIF

src/narwhals/stable/v1/dependencies.py
FBruzzesi: Thanks @dangotbanned, regarding `DataFrameLike`, the issue is that Dask would pass `supports_dataframe_interchange(frame)` since it implements it 😭 however unlike ibis and duckdb, we always had dask in the lazy category (rightfully so)

tests/dependencies/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+        obj = DynamicAttrOnly()
+        hasattr(obj, "i_dont_exist")          # True   (false positive)
+        _hasattr_static(obj, "i_dont_exist")  # False  (correct)
+
+    Use it in tests to prove that code under test relies on `_hasattr_static`
+    rather than `hasattr` when probing unknown objects: any check that accepts
+    a `DynamicAttrOnly` instance is, by construction, using the unsafe path.
+
+    [`inspect.getattr_static`]: https://docs.python.org/3/library/inspect.html#inspect.getattr_static
+    """
```
FBruzzesi: @dangotbanned I hope you like the name. For me it was better than `Imposter` 😇

dangotbanned: @FBruzzesi although I was only half-serious with the name `Imposter`, what I was trying to convey was *"this isn't what you think it is"*.

If we were going for describing the runtime behavior instead, how would you feel about this?

```diff
- DynamicAttrOnly
+ AlwaysHasAttr
```

FBruzzesi: Naming is hard 🤣 I guess `AlwaysHasAttr` also work. Will adjust (after dinner)

FBruzzesi: https://github.com/narwhals-dev/narwhals/pull/3613/commits/efaa06b3759ab251c26ee1be1a9f8709a237a313

src/narwhals/stable/v1/dependencies.py
dangotbanned: @MarcoGorelli this made me spiral a bit, so was wondering if you had an opinion on it?

## Example
I'm testing this on `main`, and I think this is a bug (https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3247177409).

- I was quite surprised about the `dd.DataFrame` part, I had no idea we explicitly rejected it's `__dataframe__` support
- `DataFrameLike` returning `False` for `v1.dependencies.is_into_dataframe` seems wrong? 

```py

from typing import TYPE_CHECKING, Any, assert_never, assert_type

from narwhals.stable import v1 as nw_v1

if TYPE_CHECKING:
    import dask.dataframe as dd


class UserDF:
    def __dataframe__(self, *args: Any, **kwargs: Any) -> Any:
        return self


def dask_1(native: dd.DataFrame) -> None:
    # this raises at runtime, but type checks
    boom = nw_v1.from_native(native, eager_or_interchange_only=True)

    assert_never(boom)
    # Argument of type "narwhals.stable.v1.DataFrame[dask.dataframe.dask_expr._collection.DataFrame]" cannot be assigned to parameter "arg" of type "Never" in function "assert_never"
    #     Type "narwhals.stable.v1.DataFrame[dask.dataframe.dask_expr._collection.DataFrame]" is not assignable to type "Never"


def dask_2(native: dd.DataFrame) -> None:
    if nw_v1.dependencies.is_into_dataframe(native):
        # this should be unreachable since we reject `dask` in `from_native`
        assert_never(native)
        # Argument of type "DataFrame" cannot be assigned to parameter "arg" of type "Never" in function "assert_never"
        #     Type "DataFrame" is not assignable to type "Never"


def from_native_unknown(native: UserDF) -> None:
    # typing matches runtime
    okay = nw_v1.from_native(native, eager_or_interchange_only=True)

    assert_type(okay, nw_v1.DataFrame[UserDF])

    if not nw_v1.dependencies.is_into_dataframe(native):
        # this should be unreachable since we accept `__dataframe__` in `from_native`
        assert_never(native)
        # Argument of type "UserDF" cannot be assigned to parameter "arg" of type "Never" in function "assert_never"
        #     Type "UserDF" is not assignable to type "Never"

```

MarcoGorelli: i'd be inclined to just leave v1 as-is tbh, unless it needs changing for some reason, it is what it is

dangotbanned: @MarcoGorelli I would generally agree with this sentiment.

The issue is that we're introducing a guard for `v1` and the one on `main` currently rejects valid input.

(https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3229454400) is taken directly from `altair` - which to me looks like working around this bug?

MarcoGorelli: does this need changing in altair? if not, i'd suggest just leaving it alone, else we risk introducing issues

FBruzzesi: @dangotbanned just to be sure I understand correctly, are you suggesting that `is_into_dataframe` for v1 namespace adds a check for the interchange protocol?
Regarding Dask, we would end up both `is_into_dataframe` and `is_into_lazyframe` to return True. I am not sure this is ideal. For ibis and duckdb we made the choice to keep them as interchange dataframe's for v1, but not v2. For Dask would be quite different and arguably interchange support has only a subset of eager/lazy operations so I don't see a reason for using that rather than the full support

dangotbanned: > @FBruzzesi 
> just to be sure I understand correctly, are you suggesting that `is_into_dataframe` for v1 namespace adds a check for the interchange protocol?

Yeah that's what I was suggesting here

<details><summary>Show comment</summary>
<p>

> ## 1. `DataFrameLike`
> 
> This is part of the deviation for `v1.typing.IntoDataFrame`, so it needs to be part of`v1.dependencies.is_into_dataframe`
> 
> https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/stable/v1/typing.py#L28-L29


</p>
</details> 

---

> @MarcoGorelli 
> does this need changing in altair? if not, i'd suggest just leaving it alone, else we risk introducing issues

The point I was trying to make with `altair` is that this **should** work when the target is `v1`'s `from_native`:

<p>

```diff
from narwhals.stable.v1.dependencies import is_into_dataframe
...
if VEGAFUSION_VERSION and Version("2.0.0a0") <= VEGAFUSION_VERSION:

-    def is_supported_by_vf(data: Any) -> TypeIs[DataFrameLike]:
-        # Test whether VegaFusion supports the data type
-        # VegaFusion v2 support narwhals-compatible DataFrames
-        return isinstance(data, DataFrameLike) or is_into_dataframe(data)
+    is_supported_by_vf = is_into_dataframe
else:

    def is_supported_by_vf(data: Any) -> TypeIs[DataFrameLike]:
        return isinstance(data, DataFrameLike)
```

</p>
</details> 



But whether we change `is_into_dataframe` or not, the **current code already checks for what we missed**.

---

> For Dask would be quite different and arguably interchange support

Just to be super clear, I am not trying to introduce interchange-level support for `dask`.
I don't have a use-case for it and `dask` is probably the backend with the lowest level of support to begin with 😂 



[This](https://github.com/narwhals-dev/narwhals/blob/d1a256cc84ba032d230c8f9a5a5b9ff9728cebdb/src/narwhals/stable/v1/dependencies.py#L63-L68) is an improvement over `main` 🥳 - but it is missing what I see as the defining feature of `v1`.

What I want is for our typing to match our runtime behavior *as-close-as-realistically-possible*.
When these things diverge, the effectiveness of the typing and the utility of the runtime suffer.

I was trying to demonstrate that in (https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3273548035), but we had a semi-related case in https://github.com/narwhals-dev/narwhals/pull/3639#discussion_r3261032002) too - where `IntoDataFrameT` has another meaning depending on a keyword-arg

> [!NOTE]
> I don't know what the right answer to this is, but I see issues with the typing and felt the need to bring them up 🙂

MarcoGorelli: > the current code already checks for what we missed

yeah sorry that's what i was getting to - if altair is already working around it, then maybe it doesn't need to change? just leave as-is, we're not really encouraging v1 usage nowadays anyway

dangotbanned: The more abstract problem is a function not behaving the way I expect it would.
An alternative to making any runtime or typing changes is to leave this to the docs?

MarcoGorelli: yup sure

dangotbanned: @FBruzzesi please vomit some words into the docstring 😘

FBruzzesi: WDYT: https://github.com/narwhals-dev/narwhals/pull/3613/commits/cc294dd477b6137389cd94da4c45cf187a9de566

dangotbanned: Perfect, thank you for bearing with me @FBruzzesi 😄

src/narwhals/stable/v1/dependencies.py
```diff
@@ -1,143 +1,157 @@
...
-def is_pandas_dataframe(df: Any) -> TypeIs[pd.DataFrame]:
-    """Check whether `df` is a pandas DataFrame without importing pandas."""
-    return ((pd := get_pandas()) is not None and isinstance(df, pd.DataFrame)) or any(
-        (mod := sys.modules.get(module_name, None)) is not None
-        and isinstance(df, mod.pandas.DataFrame)
-        for module_name in IMPORT_HOOKS
-    )
-
+def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoDataFrameT]:
+    """Check whether `native_dataframe` can be converted to a [narwhals.stable.v1.DataFrame][].
```
dangotbanned: The content is great!

But since this won't appear on the [API reference](https://narwhals-dev.github.io/narwhals/api-reference/dependencies/#narwhals.dependencies.is_into_dataframe) - could we use markdown that looks good in an IDE?

The main issue is the links

image

src/narwhals/dependencies.py
```diff
@@ -541,13 +679,41 @@ def is_into_dataframe(native_dataframe: Any | IntoDataFrameT) -> TypeIs[IntoData
...
-        isinstance(native_dataframe, DataFrame)
-        or hasattr(native_dataframe, "__narwhals_dataframe__")
-        or is_polars_dataframe(native_dataframe)
-        or is_pyarrow_table(native_dataframe)
-        or is_pandas_like_dataframe(native_dataframe)
-    )
+    if isinstance(native_dataframe, DataFrame):
+        return True
+    _warn_if_narwhals_df_or_lf(native_dataframe)
+    return _is_into_native_dataframe(native_dataframe)
```
dangotbanned: Is this warning still supposed to be here?

I tried removing this:

https://github.com/narwhals-dev/narwhals/blob/cc294dd477b6137389cd94da4c45cf187a9de566/tests/dependencies/is_into_dataframe_test.py#L44

And got this output:

```log
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[pandas[pyarrow]] - UserWarning: You passed a `narwhals.stable.v1.DataFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[pyarrow] - UserWarning: You passed a `narwhals.stable.v1.DataFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[duckdb] - UserWarning: You passed a `narwhals.dataframe.LazyFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[polars[eager]] - UserWarning: You passed a `narwhals.stable.v1.DataFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[pandas] - UserWarning: You passed a `narwhals.stable.v1.DataFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[sqlframe] - UserWarning: You passed a `narwhals.dataframe.LazyFrame` to `is_into_dataframe`.
FAILED tests/dependencies/is_into_dataframe_test.py::test_is_into_dataframe[ibis] - UserWarning: You passed a `narwhals.dataframe.LazyFrame` to `is_into_dataframe`.
```

So we're allowed to pass `DataFrame`, but not `LazyFrame` or `v1.DataFrame`?

FBruzzesi: I am preserving the past behavior. I would be ok to keep the warning only on the native checks e.g. `is_pandas_dataframe`

dangotbanned: Yeah that makes more sense to me.

The idea behind the warning was `nw.DataFrame[pd.DataFrame]` returning `False` for `is_pandas_dataframe` could be confusing.

But both the native & narwhals version have the same result here 🙂

src/narwhals/stable/v1/dependencies.py
FBruzzesi: Adjusted in https://github.com/narwhals-dev/narwhals/pull/3613/commits/a7de8688b19813eadb10d4432b8318d73859e040

dangotbanned (approved):
Thanks @FBruzzesi for powering through this

bullseye

src/narwhals/stable/v1/dependencies.py
```diff
@@ -1,143 +1,157 @@
...
-    """Check whether `ser` is a pandas Series without importing pandas."""
-    return ((pd := get_pandas()) is not None and isinstance(ser, pd.Series)) or any(
-        (mod := sys.modules.get(module_name, None)) is not None
-        and isinstance(ser, mod.pandas.Series)
-        for module_name in IMPORT_HOOKS
+    return (
+        isinstance(native_dataframe, DataFrame)
+        or _is_into_native_dataframe(native_dataframe)
+        or is_ibis_table(native_dataframe)
+        or is_duckdb_relation(native_dataframe)
```
MarcoGorelli: i'm confused, doesn't the docstring say that these aren't accepted here?

MarcoGorelli (changes requested):

src/narwhals/stable/v1/dependencies.py
dangotbanned: 🤨 

https://github.com/narwhals-dev/narwhals/blob/aa049802e6f19f6ad14270073686114928e468b5/src/narwhals/stable/v1/dependencies.py#L68

MarcoGorelli: well it says

> arbitrary objects implementing the `__dataframe__` interchange
        protocol `narwhals.stable.v1.typing.DataFrameLike` are accepted
        by `v1.from_native(..., eager_or_interchange_only=True)` but are **not**
        recognised by this function

so why are they being special-cased?

based on the previous thread, i thought we were preseving v1 behaviour for `is_into_dataframe`

FBruzzesi: I thought we were on the same page to fix the example in https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3225309408

```py

nw.dependencies.is_into_dataframe(duckdb_obj)  # False
isinstance(nw.from_native(duckdb_obj), nw.DataFrame)  # True
```

this requires to change `is_into_dataframe` runtime behavior for ibis and duckdb

dangotbanned: Yeah I was thinking the same.

The issue was with `dask` and `__dataframe__`.

I'm not sure what the purpose of a `v1` `is_into_dataframe` is if it can't test any of the unique types it describes?

https://github.com/narwhals-dev/narwhals/blob/98b4c8ccb21953a695272e8af69de90b39e06d0e/narwhals/stable/v1/typing.py#L28-L29

MarcoGorelli: sorry i wasn't clear

if there's no need to change it, i'd prefer to keep existing behaviour than to modify anything in v1

(and if it did need changing, it should be a separate pr labelled as bug fix)

FBruzzesi: I understand the feeling of 

> if there's no need to change it, i'd prefer to keep existing behaviour than to modify anything in v1

yet to me this seems like a clear bug. Since we are touching every `dependency` file anyway, I thought to address it.

---

> (and if it did need changing, it should be a separate pr labelled as bug fix)

In my defence the PR always had both `enhancement` and `fix` labels, and my first self comment was exactly on this very change to bring the focus on it https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3222734722

MarcoGorelli: true, thanks for looking into it. if it's ok, i'd still prefer to leave as-is, i'm concerned about unforeseen knock-on effects here not captured by ci or downstream tests

also true 🙌 but from a "do we need to revert this?" perspective, it's much easier if it's separate

src/narwhals/stable/v1/dependencies.py
```diff
@@ -1,143 +1,117 @@
...
-        (mod := sys.modules.get(module_name, None)) is not None
-        and isinstance(df, mod.pandas.DataFrame)
-        for module_name in IMPORT_HOOKS
-    )
-
+# TODO(Unassigned): For duckdb and ibis backends:
+#   * is_into_dataframe(native_frame) returns False
+#   * nw_v1.from_native(native_frame) returns a narwhals.stable.v1.DataFrame
+#   * Therefore is_into_dataframe(nw_v1.from_native(native_frame)) returns True
+# See discussion https://github.com/narwhals-dev/narwhals/pull/3613#discussion_r3288440039
```
FBruzzesi, line 63: @MarcoGorelli I added this comment to track the behavior after rolling back the changes

src/narwhals/dependencies.py
```diff
@@ -482,6 +589,43 @@ def is_cudf_dtype(
...
+    return (
+        _hasattr_static(obj, "__narwhals_lazyframe__")
+        or _is_polars_lazyframe(obj)
+        or _is_dask_dataframe(obj)
+        or _is_duckdb_relation(obj)
+        or _is_ibis_table(obj)
+        or _is_pyspark_dataframe(obj)
+        or _is_pyspark_connect_dataframe(obj)
+        or _is_sqlframe_dataframe(obj)
+    )
```
MarcoGorelli: not sure i get the name of this function

`is_into_lazyframe` means it can be converted to a narwhals LazyFrame

what does this mean, that it can be converted to a native lazyframe?

MarcoGorelli (commented):
generally looks good, just got a question on a name

src/narwhals/dependencies.py
FBruzzesi: The idea was: can a native frame be converted into a lazyframe? This function excludes `narwhals.LazyFrame` itself, which is added via instance checks in the `is_into_lazyframe` function

MarcoGorelli: `is_native_into_lazyframe` maybe?

MarcoGorelli (approved):
i'm removing my blocking "request changes" but haven't checked the rest in detail

if you're all happy with the rest, feel free to ship it

it mostly just keeps things as are, refactors a bit, and adds is_into_lazyframe right?

src/narwhals/dependencies.py
dangotbanned: `is_native_lazyframe`, perhaps?

FBruzzesi:
> i'm removing my blocking "request changes" but haven't checked the rest in detail

Thanks Marco, appreciate it!

> if you're all happy with the rest, feel free to ship it
> 
> it mostly just keeps things as are, refactors a bit, and adds is_into_lazyframe right?

Yes, after your previous comment, this PR is:

* Adding `is_into_lazyframe`
* Aligning which functions are exposed in each `dependencies` module

PR #3621, opened 2026-05-13
feat: add `Selector.__xor__` and `Expr.__xor__` for symmetric difference

### LINKED ISSUE #2589: Add `Selector.__xor__` (symmetric difference)
## Related
- #224
- #247
- https://github.com/pola-rs/polars/blob/b9dd8cdbd6e6ec8373110536955ed5940b9460ec/crates/polars-plan/src/dsl/selector.rs#L10-L16
- https://docs.pola.rs/api/python/stable/reference/selectors.html#set-operations
- https://en.wikipedia.org/wiki/Symmetric_difference

## Description
I can't seem to find this ever being discussed - but while working on modeling selectors in (#2572) it looks like we're missing 1 set operator (`^`).

Example from the `polars` docs:

```py
import polars.selectors as cs
import polars as pl

# set up an empty dataframe with plenty of columns of various dtypes
df = pl.DataFrame(
    schema={
        "abc": pl.UInt16,
        "bbb": pl.UInt32,
        "cde": pl.Float64,
        "def": pl.Float32,
        "eee": pl.Boolean,
        "fgg": pl.Boolean,
        "ghi": pl.Time,
        "JJK": pl.Date,
        "Lmn": pl.Duration,
        "opp": pl.Datetime("ms"),
        "qqR": pl.String,
    },
)



# Select the SYMMETRIC DIFFERENCE of numeric columns and columns that contain an "e"
assert df.select(cs.contains("e") ^ cs.numeric()).schema == {
    "abc": UInt16,
    "bbb": UInt32,
    "eee": Boolean,
}
```

Considering that we do support the other 4 (`&`, `|`, `-`, `~`) operators, it would round out the *set* nicely 😉

SAY-5:
# Description

Adds `Selector.__xor__` for symmetric difference between selectors, rounding out the set operator set (`&`, `|`, `-`, `~`, `^`) as discussed in #2589. `selA ^ selB` returns the columns in either selector but not both. Implementation mirrors the existing `__or__`/`__sub__` pattern; for Polars, falls back to `(a | b) - (a & b)` on versions before 1.0.0 where native `Selector.__xor__` was added.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2589

## Checklist

- [x] Tests added
- [x] Documented the changes


src/narwhals/selectors.py
```diff
@@ -56,6 +56,22 @@ def __and__(self, other: Any) -> Expr:  # type: ignore[override]
...
+                    "__xor__",
+                    exprs=(other,),
+                    str_as_lit=True,
+                    allow_multi_output=True,
+                )
+            )
+        msg = (
+            f"unsupported operand type(s) for op: ('Selector' ^ '{type(other).__name__}')"
+        )
+        raise TypeError(msg)
```
MarcoGorelli: in `__or__` we have
```
        return self._to_expr()._append_node(
            ExprNode(ExprKind.ELEMENTWISE, "__or__", exprs=(other,), str_as_lit=True)
        )
```
why are we raising an error here?

MarcoGorelli: ah thanks - i kinda feel like, if we're adding `__xor__` for selectors, we should add it for `Expr` too.

i'd suggest
- either do it here
- or, open an issue to do that as a follow-up

tests/expr_and_series/operators_test.py
```diff
@@ -60,6 +64,18 @@ def test_logic_operators_expr(
     assert_equal_data(result, {"a": expected})
 
 
+def test_xor_operator_expr(constructor: Constructor) -> None:
```
MarcoGorelli, line 67: can we also have a `test_xor_operator_expr_nulls` test, to check that null values propagate?

src/narwhals/_compliant/selectors.py
```diff
@@ -310,6 +310,45 @@ def names(df: FrameT) -> Sequence[str]:
...
+                return [
+                    *(x for x in lhs_names if x not in rhs_names),
+                    *(x for x in rhs_names if x not in lhs_names),
+                ]
+
+            return self.selectors._selector.from_callables(series, names, context=self)
+        # The narwhals-level Selector.__xor__ rejects non-selectors before reaching
+        # here, so this branch is a defensive guard only.
+        msg = f"unsupported operand type(s) for op: ('Selector' ^ '{type(other).__name__}')"  # pragma: no cover
+        raise TypeError(msg)  # pragma: no cover
```
MarcoGorelli: is this still needed if you've implemented it for `Expr`?

src/narwhals/_polars/expr.py
```diff
@@ -272,6 +272,27 @@ def __and__(self, other: PolarsExpr) -> Self:
...
         return self._with_native(self.native.__or__(extract_native(other)))
 
+    def __xor__(self, other: PolarsExpr) -> Self:
+        other_native = extract_native(other)
+        if pl.selectors.is_selector(self.native) and not pl.selectors.is_selector(
+            other_native
+        ):
+            # Polars coerces the right operand to a selector and takes the
+            # set difference; degrade to expressions for an elementwise xor.
+            return self._with_native(self.native.as_expr().__xor__(other_native))
```
MarcoGorelli, line 282: i'm confused by this, sorry - do you have an example please?

is it a bug which needs fixing in polars upstream?

PR #3608, opened 2026-05-09
chore: Adopt uv as package manager

FBruzzesi:
# Description

Related [discord thread](https://discord.com/channels/1235257048170762310/1272835922924273694/1489735866548097248)

Closes #3626

---

4.7k over 5.1k lines are due to committing the uv.lock file - The main reason for that is that these days security vulnerabilities are a nightmare and we could get some automatic dependabot alerts if we have a lock file. This is the same reason for which I started to pin some non-core dependencies (e.g. pytest has vulnerabilities fixed in v9.0.3).

`uv audit` is still experimental but also a good start to run periodically locally to check for known vulnerabilities

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

PR #3631, opened 2026-05-16
chore: Move to `src/` layout

### LINKED ISSUE #881: [Enh]: consider switching to `src` layout

_No response_

AFAIK the "`src`-layout" should be preferred to the "flat-layout"

Some opinions:

> - The src layout helps prevent accidental usage of the in-development copy of the code.
> - The src layout helps enforce that an [editable installation](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) is only able to import files that were meant to be importable.

[source](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/)

> If you use the [ad hoc layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) without an src directory, your tests do not run against the package as it will be installed by its users.

[source](https://hynek.me/articles/testing-packaging/)

I think it will also simplify the build and make sure we don't include files/directories we don't need. (see #879)

Of course, if we merge this change, all the open PRs will have to fix conflicts. It will be better to schedule well a change like this one (before a sprint for example :) )

_No response_

_No response_

Happy to hear what you all think! 👀

FBruzzesi:
# Description

Hopefully the minimal changes to address https://github.com/narwhals-dev/narwhals/pull/3608#pullrequestreview-4303367656

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #881


FBruzzesi:
CI failure is unrelated

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #3638, opened 2026-05-17
fix: `.over` was mutating expressions in some cases

MarcoGorelli:
thanks @dangotbanned for spotting this on discord https://discord.com/channels/1235257048170762310/1272835922924273694/1492563536696578248

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/src/narwhals/expr.py
+++ b/src/narwhals/expr.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import math
 from collections.abc import Callable, Iterable, Mapping, Sequence
+from copy import deepcopy
 from typing import TYPE_CHECKING, Any
 
 from narwhals._expression_parsing import ExprKind, ExprNode, evaluate_nodes
@@ -63,7 +64,7 @@ class Expr:
     def _with_over_node(self, node: ExprNode) -> Self:
         # insert `over` before any elementwise operations.
         # check "how it works" page in docs for why we do this.
-        new_nodes = list(self._nodes)
+        new_nodes = deepcopy(list(self._nodes))
         kwargs_no_order_by = {
             key: value if key != "order_by" else []
             for (key, value) in node.kwargs.items()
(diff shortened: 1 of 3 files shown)


src/narwhals/expr.py
```diff
@@ -63,7 +64,7 @@ def _append_node(self, node: ExprNode) -> Self:
     def _with_over_node(self, node: ExprNode) -> Self:
         # insert `over` before any elementwise operations.
         # check "how it works" page in docs for why we do this.
-        new_nodes = list(self._nodes)
+        new_nodes = deepcopy(list(self._nodes))
```
MarcoGorelli, line 67: based on the name, it looks like i was under the impression that `list` would copy the tuple `self._nodes`

dangotbanned: That is usually enough when trying to avoid the *dictionary changed size during iteration* issue - which is probably more common to run into

dangotbanned:
Thanks for following up on this @MarcoGorelli

Since I kinda fell into this one by chance, do you think there might be any other places since introducing `ExprNode` that may have a similar bug?

`when` could be another one to try out?

MarcoGorelli:
sure, added a test for when/then

src/narwhals/expr.py
MarcoGorelli: not sure what you mean here sorry

dangotbanned: I meant that if `self._nodes` were a `dict`, then the code you wrote would have avoided an error related to mutation.

I was trying to say that it wasn't unreasonable to expect it would work for `tuple` too 🙂

dangotbanned (approved):
Thanks @MarcoGorelli 

squeeze cheeks narwhal

Issue #3497, opened 2026-03-03
Properly documenting `ExprIR`


dangotbanned:
## Description
(#2572) and more recently ([expr-ir/logical-plan]) add a **lot** of code, but only a *little* of docs.

My thinking was to document the deviations from `main`, since very big chunks are the same and *already* well-documented.
I was also concerned about sinking much time into things while they were changing.

Well *now*, enough of the core is stable (enough) to start blabbing about - so its time!

# Documenting
## Classes/docstrings
There are a lot, but many are tiny and only *really* need an understanding of a base class:

<details><summary>Look how smol</summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/51cebab8c55b47220ac0599cf69ea4973900a06b/narwhals/_plan/expressions/aggregation.py#L46-L53

</p>
</details> 

So these guys are my picks for what could be most beneficial for better docstrings in `narwhals._plan`.

*Reading this back, okay this still looks like a lot 🤦‍♂️*

### Expressions
- [ ] `_expr_ir.ExprIR` (**priority**)
- [ ] `expressions.aggregation.AggExpr`
- [ ] `expressions.expr.Literal`
- [ ] `expressions.name.KeepName`
- [ ] `expressions.name.RenameAlias`
- [ ] `options.ExprIROptions`

### Selectors
- [ ] `_expr_ir.SelectorIR`
- [ ] `expressions.expr.RootSelector`
- [ ] `expressions.expr.BinarySelector`
- [ ] `expressions.expr.InvertSelector`
- [ ] `expressions.selectors.Selector` (**priority**)
- [ ] `expressions.selectors.DTypeSelector` (**priority**)

### Functions
- [ ] `_function.Function` (**priority**)
- [ ] `expressions.expr.FunctionExpr` (**priority**)
- [ ] `options.FunctionFlags`
- [ ] `options.FunctionOptions` (**priority**)
- [ ] `options.FunctionExprOptions`
- [ ] `_dispatch.Dispatcher` (**priority**)
- [ ] `_dtype.ResolveDType`  (new in `logical-plan`)

### Expansion
- [ ] `_expr_ir.NamedIR`
- [ ] `_expansion.Expander`
- [ ] `meta.MetaNamespace`
- [ ] `schema.FrozenSchema` (+ `freeze_schema` )

### Misc
- [ ] `_immutable.Immutable` (**priority**)

### Compliant
- [ ] `compliant.expr.CompliantExpr`
[...]

PR #3641, opened 2026-05-18
feat(expr-ir): Improve `lit` repr

dangotbanned:
# Description
Brief interlude before getting back to (#3497)

> [!NOTE]
> Just trying to keep myself sane with something unimportant 😅

## Before
Writing the docs (and doc tests) has made me very unhappy with whatever this **was**.

https://github.com/narwhals-dev/narwhals/blob/10857f998d27b3d8dc61d39eb2cf1b1f3d265508/narwhals/_plan/expressions/literal.py#L47

https://github.com/narwhals-dev/narwhals/blob/10857f998d27b3d8dc61d39eb2cf1b1f3d265508/narwhals/_plan/when_then.py#L116

## After
A little less noisy
https://github.com/narwhals-dev/narwhals/blob/df41a4444927efab9faa4499a7d787b04b78a8a1/narwhals/_plan/expressions/literal.py#L47

https://github.com/narwhals-dev/narwhals/blob/df41a4444927efab9faa4499a7d787b04b78a8a1/narwhals/_plan/when_then.py#L116

Those were smallest examples I could find.

<details><summary>These are the exhaustive tests</summary>
<p>

Hah, made you look 😉 

https://github.com/narwhals-dev/narwhals/blob/df41a4444927efab9faa4499a7d787b04b78a8a1/tests/plan/repr_test.py#L82-L263

</p>
</details> 

## What's changed?
- Worked backwards from what `pl.lit(...)` does
  - Mainly using the same *value* repr, but wrapping it instead of things like `"im a string"`
- The [*"short dtype repr code"*](https://github.com/pola-rs/polars/blob/4082dc3e849935c04297100ebc454f63c3d0a48d/crates/polars-core/src/datatypes/dtype.rs#L1207-L1264) is displayed in siuations that are ambiguous
  - If the value repr is a string, but the type is not `String`
    - E.g. `Enum`, `Categorical`, `Decimal`, `Temporal`
  - If the type was provided explicitly and cannot be inferred
    - `Numeric` (excluding `Int64`, `Float64`)
    - Anything else that isn't `Boolean`, `String` or `Binary`
  - The value is `None`, but there is an explicit dtype

## Related issues

- Child of #2572


narwhals/_plan/expressions/boolean.py
```diff
@@ -42,8 +42,8 @@ class _HorizontalBoolean(HorizontalFunction, BooleanFunction, dtype=dtm.BOOL):
     ignore_nulls: bool
 class All(_BooleanUnary, flags=AGGREGATION): ...
 class Any(_BooleanUnary, flags=AGGREGATION): ...
-class AllHorizontal(_HorizontalBoolean):...
-class AnyHorizontal(_HorizontalBoolean):...
+class AllHorizontal(_HorizontalBoolean): ...
+class AnyHorizontal(_HorizontalBoolean): ...
```
dangotbanned, line 46: Somehow this was causing `coverage` to think these lines were uncovered? 😂

PR #3650, opened 2026-05-25
chore(typing): Enable `allow-redefinition` for `mypy`

### LINKED ISSUE #3646: Enable `allow-redefinition` for `mypy`
## Description
I've had this enabled on (#2572) for a little bit now and it's been stable since [`mypy==2.0.0`](https://mypy.readthedocs.io/en/stable/changelog.html#new-behavior-for-allow-redefinition)

Bringing it up because I'm dealing with (#3631) locally and have an example of how it changes `main`.

<details><summary>Found 11 errors in 7 files</summary>
<p>

```py
src/narwhals/dtypes.py:975: error: Unused "type: ignore" comment  [unused-ignore]
                dtype_ = dtype_.inner  # type: ignore[assignment]
                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/narwhals/dtypes.py:975: error: Item "DType" of "DType | type[NumericType] | type[TemporalType] | type[String] | type[Boolean] | type[Binary] | type[Categorical] | type[Unknown] | type[Object]" has no attribute "inner"  [union-attr]
                dtype_ = dtype_.inner  # type: ignore[assignment]
                         ^~~~~~~~~~~~
src/narwhals/dtypes.py:975: note: Error code "union-attr" not covered by "type: ignore[assignment]" comment
src/narwhals/dtypes.py:975: error: Item "DTypeClass" of "DType | type[NumericType] | type[TemporalType] | type[String] | type[Boolean] | type[Binary] | type[Categorical] | type[Unknown] | type[Object]" has no attribute "inner"  [union-attr]
                dtype_ = dtype_.inner  # type: ignore[assignment]
                         ^~~~~~~~~~~~
src/narwhals/dtypes.py:975: error: Item "DType" of "DType | type[NumericType] | type[TemporalType] | type[String] | type[Boolean] | type[Binary] | type[Categorical] | type[Unknown] | type[Object] | Any" has no attribute "inner"  [union-attr]
                dtype_ = dtype_.inner  # type: ignore[assignment]
                         ^~~~~~~~~~~~
src/narwhals/dtypes.py:975: error: Item "DTypeClass" of "DType | type[NumericType] | type[TemporalType] | type[String] | type[Boolean] | type[Binary] | type[Categorical] | type[Unknown] | type[Object] | Any" has no attribute "inner"  [union-attr]
                dtype_ = dtype_.inner  # type: ignore[assignment]
                         ^~~~~~~~~~~~
src/narwhals/_polars/namespace.py:189: error: Unused "type: ignore" comment  [unused-ignore]
                    result = pl.fold(  # type: ignore[assignment]
                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/narwhals/_pandas_like/utils.py:198: error: Unused "type: ignore" comment  [unused-ignore]
        return cast("NativeNDFrameT", result)  # type: ignore[redundant-cast]
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/narwhals/_arrow/series.py:610: error: Argument 1 to "from_arrays" of "Table" has incompatible type "list[Any] | list[_PandasConvertible[Series[Any]]]"; expected "Collection[Array[Any] | ChunkedArray[Any]]"  [arg-type]
            val_count = pa.Table.from_arrays(arrays, names=[index_name_, value_name_])
                                             ^~~~~~
tests/series_only/value_counts_test.py:32: error: Unused "type: ignore" comment  [unused-ignore]
            expected_count = [v / len(data) for v in expected_count]  # type: ignore[misc]
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/frame/reindex_test.py:30: error: Unused "type: ignore" comment  [unused-ignore]
        expected = {"a": [1, 2, 3], "b": [4, 4, 6], "z": [7.0, 8.0, 9.0]}  # type: ignore[list-item]
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tests/frame/group_by_test.py:87: error: Unused "type: ignore" comment  [unused-ignore]
        expected_keys = [(1, 4), (3, 6)]  # type: ignore[list-item]
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 11 errors in 7 files (checked 684 source files)
```

</p>
</details> 

Half are `[unused-ignore]` 🥳 

@MarcoGorelli I could fix the rest if you're interested?
The newer behavior is closer to all the other type checkers

dangotbanned:
Closes #3646

# Description
Enables the *opt-in* behavior `mypy` has had since `1.20.0`.

This brings `mypy` closer to all the other type checkers, which **wouldn't describe this as redefining in the first place** [^1].

[^1]: See [pyright/mypy-comparison](https://microsoft.github.io/pyright/#/mypy-comparison?id=variable-type-declarations)

## Related

- Closes  #3646
- https://mypy.readthedocs.io/en/stable/changelog.html#new-behavior-for-allow-redefinition
- https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -338,6 +338,7 @@ exclude_also = [
 files = ["src/narwhals", "tests", "test-plugin"]
 pretty = true
 strict = true
+allow_redefinition = true  # https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition
 
 [[tool.mypy.overrides]]
 module = [
--- a/src/narwhals/dtypes.py
+++ b/src/narwhals/dtypes.py
@@ -972,6 +972,9 @@ class Array(NestedType):
         # Get leaf type
         dtype: Self | IntoDType = self
         tp_self = type(self)
+        # NOTE: This might look odd, but if we used a `while` loop
+        # it won't trigger a `RecursionError` like `for` does.
+        # https://github.com/narwhals-dev/narwhals/pull/3651
         for _ in self.shape:
             if isinstance(dtype, tp_self):
                 dtype = dtype.inner
--- a/src/narwhals/dtypes.py
+++ b/src/narwhals/dtypes.py
@@ -975,6 +975,9 @@ class Array(NestedType):
         for _ in self.shape:
             if isinstance(dtype, tp_self):
                 dtype = dtype.inner
+            else:  # pragma: no cover
+                msg = "Either something has gone horribly wrong in narwhals, or you've fiddled with the internals!"
+                raise TypeError(msg)
         return f"{tp_self.__name__}({dtype!r}, shape={self.shape})"
 
 
--- a/src/narwhals/dtypes.py
+++ b/src/narwhals/dtypes.py
@@ -972,7 +972,7 @@ class Array(NestedType):
         # Get leaf type
         dtype_ = self
         for _ in self.shape:
-            dtype_ = dtype_.inner  # type: ignore[assignment]
+            dtype_ = dtype_.inner  # type: ignore[union-attr]
 
         class_name = self.__class__.__name__
         return f"{class_name}({dtype_!r}, shape={self.shape})"
--- a/src/narwhals/dtypes.py
+++ b/src/narwhals/dtypes.py
@@ -970,12 +970,11 @@ class Array(NestedType):
 
     def __repr__(self) -> str:
         # Get leaf type
-        dtype_ = self
-        for _ in self.shape:
-            dtype_ = dtype_.inner  # type: ignore[union-attr]
-
-        class_name = self.__class__.__name__
-        return f"{class_name}({dtype_!r}, shape={self.shape})"
+        dtype: Self | IntoDType = self
+        tp_self = type(self)
+        while isinstance(dtype, tp_self):
+            dtype = dtype.inner
+        return f"{tp_self.__name__}({dtype!r}, shape={self.shape})"
 
 
 class Date(TemporalType):
--- a/src/narwhals/dtypes.py
+++ b/src/narwhals/dtypes.py
@@ -972,8 +972,9 @@ class Array(NestedType):
         # Get leaf type
         dtype: Self | IntoDType = self
         tp_self = type(self)
-        while isinstance(dtype, tp_self):
-            dtype = dtype.inner
+        for _ in self.shape:
+            if isinstance(dtype, tp_self):
+                dtype = dtype.inner
         return f"{tp_self.__name__}({dtype!r}, shape={self.shape})"
 
 
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -600,14 +600,11 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
 
         val_counts = pc.value_counts(self.native)
         values = val_counts.field("values")
-        counts = cast("ChunkedArrayAny", val_counts.field("counts"))
-
+        counts = val_counts.field("counts")
         if normalize:
-            arrays = [values, pc.divide(*cast_for_truediv(counts, pc.sum(counts)))]
-        else:
-            arrays = [values, counts]
-
-        val_count = pa.Table.from_arrays(arrays, names=[index_name_, value_name_])  # type: ignore[arg-type]
+            counts = pc.divide(*cast_for_truediv(counts, pc.sum(counts)))
+        arrays = values, counts
+        val_count = pa.Table.from_arrays(arrays, names=[index_name_, value_name_])
 
         if sort:
             val_count = val_count.sort_by([(value_name_, "descending")])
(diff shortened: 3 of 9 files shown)


pyproject.toml
```diff
@@ -338,6 +338,7 @@ exclude_also = [
 files = ["src/narwhals", "tests", "test-plugin"]
 pretty = true
 strict = true
+allow_redefinition = true  # https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-allow-redefinition
```
dangotbanned, line 341: Felt the need to add the link because the name makes it sound like a bad thing

src/narwhals/dtypes.py
```diff
@@ -970,12 +970,11 @@ def __hash__(self) -> int:
...
-        for _ in self.shape:
-            dtype_ = dtype_.inner  # type: ignore[assignment]
-
-        class_name = self.__class__.__name__
-        return f"{class_name}({dtype_!r}, shape={self.shape})"
+        dtype: Self | IntoDType = self
+        tp_self = type(self)
+        while isinstance(dtype, tp_self):
+            dtype = dtype.inner
+        return f"{tp_self.__name__}({dtype!r}, shape={self.shape})"
```
FBruzzesi, line 984: I remember @MarcoGorelli being against using a while loop in here even if that matches polars code: https://github.com/pola-rs/polars/blob/210b34bee60db71ece147db060f570a41ba63c5d/py-polars/src/polars/datatypes/classes.py#L1201-L1208

pyproject.toml
FBruzzesi: I think it's pretty useful to have a reference link for this feature

FBruzzesi (commented):
Thanks @dangotbanned - left a comment to summon Marco 🤣

src/narwhals/dtypes.py
dangotbanned: I'd be curious why 🤔

Besides the typing fix, does this actually make sense in isolation?

```py
dtype_ = self
for _ in self.shape:
    dtype_ = dtype_.inner
```

The first time I saw it I had no idea what it meant

dangotbanned: Okay so based on (https://github.com/narwhals-dev/narwhals/pull/1865#discussion_r1948228859)

I was originally going to ask, why do we need to handle something the interpreter does already?
- https://docs.python.org/3/library/exceptions.html#RecursionError
- https://docs.python.org/3/library/sys.html#sys.getrecursionlimit

But was surprised to find neither of these trigger that error? 😳

```py

from typing import Any, Literal


class Other: ...


class Thing:
    def __init__(self, inner: Thing | Other) -> None:
        self.inner = inner


def inner(obj: Any) -> Any:
    return obj.inner


def try_recurse(mode: Literal["attr", "function"]) -> None:
    other = Other()
    thing_1 = Thing(other)
    thing_2 = Thing(thing_1)
    thing_1.inner = thing_1

    thing = thing_2

    if mode == "attr":
        while isinstance(thing, Thing):
            thing = thing.inner
    else:
        while isinstance(thing, Thing):
            thing = inner(thing)


try_recurse(mode="attr")
```

I thought *at least* the `"function"` variant would do it.

But maybe that only applies when the outer function recurses in on itself? 🤔

dangotbanned: Another thing is that (https://github.com/pola-rs/polars/pull/16567) *did* add tests that guard against recursion.

We don't have those - so I think that's a better route than making this a `__repr__` concern

dangotbanned: I'm thinking of opening another PR to:

1. Add [`test_array_inner_recursive`](https://github.com/pola-rs/polars/blob/210b34bee60db71ece147db060f570a41ba63c5d/py-polars/tests/unit/datatypes/test_array.py#L360-L365)
2. Make any changes we need to pass it
3. Then update the repr

How does that sound?

FBruzzesi: Seems reasonable to me Dan :)

dangotbanned: > [!IMPORTANT]
> #3651 caught that one!
> I had to manually cancel every workflow in (https://github.com/narwhals-dev/narwhals/pull/3650/commits/b20cf47d3e4b27be696974d82bb511b88a0f65fd)


~~Will fix this tomorrow, moving back to draft~~ (https://github.com/narwhals-dev/narwhals/pull/3650#event-25940380727)


Thanks for spotting @FBruzzesi ❤️ 

### Edit
Fixed

MarcoGorelli (approved):
thanks both!

PR #3651, opened 2026-05-25
test: Port `Array` recursion tests from polars

dangotbanned:
# Description
Essentially, if we are concerned about recursion - we should test for it!

## Related issues

- Follow-up to https://github.com/narwhals-dev/narwhals/pull/3650#discussion_r3298398081
- https://github.com/narwhals-dev/narwhals/pull/1865#discussion_r1948228859
- https://github.com/pola-rs/polars/pull/16567

### DIFF
--- a/tests/dtypes/dtypes_test.py
+++ b/tests/dtypes/dtypes_test.py
@@ -20,6 +20,7 @@ from tests.utils import (
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
+    from narwhals._typing import _EagerAllowedImpl
     from narwhals.typing import IntoFrame, IntoSeries, NonNestedDType
     from tests.utils import Constructor, ConstructorPandasLike, NestedOrEnumDType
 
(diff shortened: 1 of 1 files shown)


dangotbanned:
@FBruzzesi everything `polars` is testing already passes.

Happy to add anything else, but I think this was enough to prevent *accidental* cases

https://github.com/narwhals-dev/narwhals/blob/bd9035f1dff394c827b2aa4aab50dffb3a471f63/src/narwhals/dtypes.py#L931-L933

## Edit
I added ([test: Add `test_array_inner_recursive_repr`](https://github.com/narwhals-dev/narwhals/pull/3651/commits/67c9fefd81667723757215aefb73f3c3d9909c91)), which does raise the error I expected to see in (https://github.com/narwhals-dev/narwhals/pull/3650#discussion_r3298332124)

> why do we need to handle something the interpreter does already?

I'm more confused now 😭

tests/dtypes/dtypes_test.py
```diff
@@ -180,6 +182,56 @@ def test_2d_array(constructor: Constructor, request: pytest.FixtureRequest) -> N
...
+            "Requires pandas 2.2+ and pyarrow for 2D array support",
+        ),
+        (
+            impl.is_pyarrow() and PYARROW_VERSION < (14,),
+            "PyArrow 14+ required for 2D array support",
+        ),
+    ]
+    for condition, reason in version_conditions:
+        if condition:
+            pytest.skip(reason)
```
FBruzzesi: Can we move skipping conditions at the top of the test?

tests/dtypes/dtypes_test.py
```diff
@@ -180,6 +182,56 @@ def test_2d_array(constructor: Constructor, request: pytest.FixtureRequest) -> N
...
+    assert s.name == ""
+    assert s.dtype == nw.Array(nw.List(nw.Array(nw.Int8, (2, 2))), 2)
+
+
+def test_array_inner_recursive() -> None:
+    # https://github.com/pola-rs/polars/blob/210b34bee60db71ece147db060f570a41ba63c5d/py-polars/tests/unit/datatypes/test_array.py#L360-L365
+    shape = (2, 3, 4, 5)
+    dtype = nw.Array(nw.Int64, shape=shape)
+    for dim in shape:
+        assert dtype.size == dim  # pyright: ignore[reportAttributeAccessIssue]
```
FBruzzesi:
```suggestion
        assert isinstance(dtype, nw.Array)
        assert dtype.size == dim
```

tests/dtypes/dtypes_test.py
```diff
@@ -180,6 +182,56 @@ def test_2d_array(constructor: Constructor, request: pytest.FixtureRequest) -> N
...
+        dtype = dtype.inner  # type: ignore[assignment]
+
+
+def test_array_inner_recursive_repr() -> None:
```
FBruzzesi, line 193: Nice one

tests/dtypes/dtypes_test.py
```diff
@@ -180,6 +182,56 @@ def test_2d_array(constructor: Constructor, request: pytest.FixtureRequest) -> N
     assert df.collect_schema()["a"] == nw.Array(nw.Array(nw.Int64(), 2), 3)
 
 
+def test_recursive_array_dtype(eager_implementation: _EagerAllowedImpl) -> None:
```
FBruzzesi: I am... not sure which recursion this is testing 🧐 If that's for a shape with more than 1 dimension, I think the test above (test_2d_array) should be enough? Sorry it has been a long day 🙈

FBruzzesi (commented):
Thanks Dan! Left some comments and some doubts 😅

tests/dtypes/dtypes_test.py
dangotbanned: Ah yes, now you see why I added my own test after copying polars 😂

dangotbanned: I can remove if you don't get it either?

FBruzzesi: Yes please 😅

FBruzzesi (approved):
Thanks for taking care of it @dangotbanned

PR #3654, opened 2026-05-25
fix(expr-ir): Allow `struct.field` to select from it's own context

dangotbanned:
# Description

The equivalent of this was previously raising:

```py
>>> import polars as pl
>>> pl.select(pl.lit({"a": 1, "b": 2}).struct.field("b"), eager=False).collect()
shape: (1, 1)
┌─────┐
│ b   │
│ --- │
│ i64 │
╞═════╡
│ 2   │
└─────┘
```

## Related issues

- Follow-up to a bug found in #3647
- Child of #2572

### DIFF
--- a/src/narwhals/_plan/expressions/struct.py
+++ b/src/narwhals/_plan/expressions/struct.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 
     from narwhals._plan.expressions import FunctionExpr, StructExpr
     from narwhals._plan.schema import FrozenSchema
-    from narwhals.dtypes import DType, Field, Struct
+    from narwhals.dtypes import DType
 
 STRUCT = Version.MAIN.dtypes.Struct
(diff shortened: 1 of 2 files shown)


src/narwhals/_plan/expressions/struct.py
```diff
@@ -57,15 +57,18 @@ def _field(self, dtype: Struct) -> Field:
...
-            and (struct := schema.get(struct_name))
+            (struct_name := arg.meta.output_name(raise_if_undetermined=False))
+            and (
+                (struct := schema.get(struct_name))
+                or (not schema and (struct := arg.resolve_dtype(schema)))
+            )
             and isinstance(struct, STRUCT)
         ):
             return into_dtype(self._field(struct).dtype)
+
```
dangotbanned: DO NOT MERGE in this state

It works now, but it's unreadable

PR #3659, opened 2026-05-29
feat(expr-ir): Almost finish `PolarsExpr`

dangotbanned:
# Description
This PR is a mad dash towards getting a 2nd backend finished.

#3611 introduced some `polars` bits to allow testing the lazy support.
However, there's still too much that's uncovered - so it felt urgent to get closer to addressing that.

So this is a big diff, but the changes fall into one or more of:
- Filling out gaps in `PolarsExpr` and `PolarsDataFrame`
- Adding version branching
- Plugging polars into most of the test suite
  - Discovered some things that aren't covered on `main`

## Follow-ups
- [ ] Shrinking `PolarsExpr` (see [thread](https://github.com/narwhals-dev/narwhals/pull/3659#discussion_r3323109374))
- [ ] Fill more gaps in 
  - [ ] `PolarsExpr`
    - [ ] `hist`
    - [ ] `cat`
    - [ ] `dt`
    - [ ] `list`
    - [ ] `str`
  - [ ] `PolarsDataFrame`
    - [ ] `group_by`
  - [ ] Rinse/repeat for lazy
    - *This is where the biggest coverage gap is*
- [ ] Fix
  - [ ] `ArrowExpr.cum_*.over(order_by=...)` has incorrect results
    - *Possibly* related to join
    - *Possibly* related to indices being taken too early/late
  - [ ] `pl.Expr.over` broadcasting in select
    - [ ] Figure out the current broadcast rules and base `ArrowExpr.over*` on that (see [thread](https://github.com/narwhals-dev/narwhals/pull/3659#discussion_r3326727146))
  - [ ] `{Expr,Series}.is_in`
    - [ ] `ArrowExpr.is_in` is equivalent to `pl.Expr.is_in(nulls_equal=True)` (see [thread](https://github.com/narwhals-dev/narwhals/pull/3659#discussion_r3328767650))
    - [ ] `list_as_series` deprecation  (see [thread](https://github.com/narwhals-dev/narwhals/pull/3659#discussion_r3328771091)
- [ ] Backports for `PolarsExpr.over`
    - I already have a fix for [`pl.int_range(pl.len()).over("order", "by", nulls_last=True).alias("row_index")`](https://discord.com/channels/1235257048170762310/1272836416048730122/1504584128438468701)
    - Would be great to generalize further though
    - The test suite has so much version gating, which I'd be super happy to avoid

## Related issues

- Child of #2572
- #3611


src/narwhals/_plan/polars/expr.py
```diff
@@ -1,127 +1,68 @@
...
-    def lit(value: Any, dtype: pl.DataType | type[pl.DataType] | None = None) -> pl.Expr:
-        return pl.struct(**value) if isinstance(value, dict) else pl.lit(value, dtype)
-
 
 ExprT_co = TypeVar("ExprT_co", bound="PolarsExpr", covariant=True)
 
 
+# TODO @dangotbanned: Do the long way first,
+# then circle back on what pattern reduces the code the most
+# (e.g maybe `pl.Expr.<method>` wrapping)
```
dangotbanned, line 49: > [!IMPORTANT]
> The current impl is more verbose than what I'm eventually aiming for

`ArrowExpr` has some tricks like this, which I'm intentionally saving for later:

https://github.com/narwhals-dev/narwhals/blob/32ebac475d33ce443e6e40f66f9f3e367d76438b/src/narwhals/_plan/arrow/expr.py#L843-L851

For that though, I'd like to see all the version compat *first*, so I can get a better idea of how that should fit in

tests/plan/over_test.py
```diff
@@ -560,13 +603,35 @@ def test_over_order_by_sort_by_asc_desc_nulls_first_last(
...
+        is_desc = "desc" in request.node.callspec.id
+        is_nulls_last = "nulls_last" in request.node.callspec.id
+        # TODO @dangotbanned: Fix polars broadcasting for `over(order_by=...)`, but aggregating for `sort_by`?
+        if POLARS_SUPPORTS_OVER_FULL or not (is_desc or is_nulls_last):
+            dataframe.xfail(
+                request,
+                True,
+                reason="BUG: polars is broadcasting for `over(order_by=...)`, but aggregating for `sort_by`",
+                raises=(ValueError, AssertionError),
+            )
```
dangotbanned, line 620: Quite a strange one here.

I modelled this test on whatever `polars` version I was using at the time - to get `pyarrow` to match.

Now it seems `polars` doesn't do this anymore 🤔 


## Related?
- https://github.com/pola-rs/polars/pull/26827

dangotbanned:
IT'S FINALLY GREEN 🥲

src/narwhals/_plan/compliant/typing.py
```diff
@@ -276,3 +280,39 @@ def __call__(
...
+_MapBatchesCallable: TypeAlias = Callable[
+    [SeriesAny], "SeriesAny | Iterable[Any] | _NumpyScalar"
+]
+
+# NOTE: This is a tradeoff
+# - `_MapBatchesOverloaded` works perfectly from a type checking perspective
+#   - But it means losing bi-directional inference
+#     (IDE completions when you write `map_batches(lambda s: s.<TAB>)`)
+# - Using a `Callable` is special-cased for that
+#   - But means `PolarsSeries`  cannot match the native signature
```
dangotbanned:
```suggestion
#   - But means `PolarsSeries` cannot match the native signature
```

src/narwhals/_plan/expressions/expr.py
```diff
@@ -155,8 +157,21 @@ def __repr__(self) -> str:
...
+    def descending(self) -> bool:
+        return self.options.descending
+
+    @property
+    def nulls_last(self) -> bool:
+        return self.options.nulls_last
+
 
+# TODO @dangotbanned: This should lower into `SortByNames` (TBD),
+# if expansion yields `is_seq_column(self.by)`
```
dangotbanned:
```suggestion
# TODO @dangotbanned: This should lower into `SortByNames` (TBD), when we've finished expanding 
and find `all(isinstance(e, ir.Col) for e in self.by)` returns True
```

src/narwhals/_plan/expressions/expr.py
```diff
@@ -155,8 +157,21 @@ def __repr__(self) -> str:
...
+# `ArrowExpr.sort_by` has this fastpath and it'll be useful in other backends
 class SortBy(ExprIR, dtype=same_dtype()):
+    """Sort an expression by the ordering of other expressions."""
```
dangotbanned:
```suggestion
class SortBy(ExprIR, dtype=same_dtype()):
    """Sort an expression by the ordering of other expressions.

    For SQL-like backends, there will be limitations on where `sort_by` is valid.

    The simplest case is to represent the [`ORDER BY` Clause in Aggregate Functions]:

        # defaults to `descending=False, nulls_last=False`
        nw.col("a").sort_by("b").first()

    Which can be translated to:

        first("a", ORDER BY "b" ASC NULLS FIRST)

    [`ORDER BY` Clause in Aggregate Functions]: https://duckdb.org/docs/current/sql/functions/aggregates#order-by-clause-in-aggregate-functions
    """
```

tests/plan/is_in_test.py
```diff
@@ -121,3 +120,74 @@ def test_expr_is_in_series(data: Data) -> None:
...
+    expr = nwp.col("c").is_in(("howdy", None))
+    expected = {"c": [True, False, False, True]}
+    dataframe.xfail(
+        request,
+        dataframe.is_polars(),
+        reason=(
+            "TODO @dangotbanned: Add `nulls_equal` parameter to `is_in`.\n"
+            "https://github.com/pola-rs/polars/pull/21426"
+        ),
+        raises=AssertionError,
```
dangotbanned, line 139: Follow-up: add `is_in(nulls_equal: bool = False)`

tests/plan/is_in_test.py
```diff
@@ -25,25 +25,28 @@ def data() -> Data:
     return {"a": [1, 4, 2, 5], "b": [1, 0, 2, 0], "c": [None, "hi", "hello", "howdy"]}
 
 
+# TODO @dangotbanned: Address polars deprecation compat (applies to all tests)
+# https://github.com/pola-rs/polars/pull/22178
```
dangotbanned, line 29: Follow-up: `is_in` using `list_as_series` was deprecated

- https://github.com/pola-rs/polars/pull/22178


https://github.com/narwhals-dev/narwhals/blob/531eaa2c84447ca5056e36f028e5ef4cc1c9798c/src/narwhals/_plan/polars/compat.py#L147

PR #3672, opened 2026-06-04
fix: preserve fireducks functionality

### LINKED ISSUE #3671: FireDucks DataFrames Accepted by is_into_dataframe() but Rejected by from_native()
Hi @MarcoGorelli, 

Hope you are doing well.

First of all, congratulations on the excellent work with Narwhals. 
It's impressive to see Narwhals already serving as a dependency for projects such as scikit-learn. 👍 

That brings me to the reason for reaching out! :slightly_smiling_face:
Recently, some FireDucks users reported an issue when passing FireDucks instances to scikit-learn algorithms.

For example, the following sample program used to work correctly up to scikit-learn 1.8. 
```python
import numpy as np
import fireducks.pandas as pd
from sklearn.linear_model import LogisticRegression

X = pd.DataFrame([[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]])
y = pd.Series([0, 0, 0, 0, 1, 1, 1, 1, 1, 1]) 
model = LogisticRegression()
model.fit(X, y) # works fine till scikit-learn 1.8
print(model.coef_)
```

However, starting with scikit-learn 1.9, it fails with a `TypeError` originating from the Narwhals translation path, specifically around nw.from_native().
```
  File "venv/lib/python3.13/site-packages/narwhals/translate.py", line 485, in _from_native_impl
    raise TypeError(msg)
TypeError: Unsupported dataframe type, got: <class 'fireducks.pandas.frame.DataFrame'>
```

After a preliminary investigation on our side, the issue appears to stem from an inconsistency between the following methods:
- is_pandas_like_dataframe() — recognizes a FireDucks DataFrame as a pandas-like DataFrame.
- is_native_pandas_like() — only accepts pandas, Modin, or cuDF instances.

On the scikit-learn side, the [validator calls nw.dependencies.is_into_dataframe()](https://github.com/scikit-learn/scikit-learn/blob/dd220898e4ce26ea809693379f1d625039828927/sklearn/utils/validation.py#L2360), 
which recognizes a FireDucks DataFrame because `nw.dependencies.is_pandas_dataframe()` returns True.
(※ As, `fireducks` is already registered in the [IMPORT_HOOKS frozenlist](https://github.com/narwhals-dev/narwhals/blob/8da0d890a3153182136427844aa4c542b2c86c60/src/narwhals/dependencies.py#L46), might be related to our [previous discussion](https://github.com/fireducks-dev/fireducks/issues/29#issuecomment-2493200885))

However, execution later reaches Narwhals [translation logic (_from_native_impl())](https://github.com/narwhals-dev/narwhals/blob/8da0d890a3153182136427844aa4c542b2c86c60/src/narwhals/translate.py#L356), 
where `nw._native.is_native_pandas_like()` returns False, ultimately resulting in the TypeError.

A simplified flow of the current behavior is shown below:
```
sklearn:  sklearn.utils.validation._get_feature_names(X):
sklearn:    if nw.dependencies.is_into_dataframe(X):  # True
narwhals:       nw.translate._from_native_impl(X):
narwhals:           \_ if is_native_pandas_like(X):   # False ---> TypeError
narwhals:               \_ if is_pandas_like_dataframe(X):  # True
```

Would it be possible to consider one of the following approaches?
1.   Remove "fireducks" from the `IMPORT_HOOKS` frozen list, so that methods such as is_into_dataframe() and is_pandas_like_dataframe() no longer recognize FireDucks instances.
2.   Alternatively, introduce support for FireDucks in the native checks (for example, an is_native_fireducks() implementation), allowing is_native_pandas_like() to recognize FireDucks instances consistently.

We already verified that (1) worked to resolve the current issue with scikit-learn 1.9
We would appreciate your kind thoughts on this and an appropriate corrective action for a future Narwhals release.

Looking forward to hearing your opinion.

Thanks and Best Regards,
Sourav Saha

MarcoGorelli:
closes #3671 

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## Checklist

- [ ] Tests added
- [ ] Documented the changes

### DIFF
--- a/.github/workflows/pytest-fireducks.yml
+++ b/.github/workflows/pytest-fireducks.yml
@@ -21,9 +21,9 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
         with:
-          python-version: "3.11"
+          python-version: "3.13"
           enable-cache: "true"
-          cache-suffix: fireducks-3.11
+          cache-suffix: fireducks-3.13
           cache-dependency-glob: "pyproject.toml"
       - name: Run pytest
         run: make run-ci DEPS="-U --group tests --extra pandas" RUN_ONLY="--with fireducks" CMD="python -m fireducks.pandas --hook-import-module -m pytest tests/fireducks_test.py"
--- a/.github/workflows/pytest-fireducks.yml
+++ b/.github/workflows/pytest-fireducks.yml
@@ -26,4 +26,4 @@ jobs:
           cache-suffix: fireducks-3.13
           cache-dependency-glob: "pyproject.toml"
       - name: Run pytest
-        run: make run-ci DEPS="-U --group tests --extra pandas" RUN_ONLY="--with fireducks" CMD="python -m fireducks.pandas --hook-import-module -m pytest tests/fireducks_test.py"
+        run: make run-ci DEPS="--group core-tests --extra pandas" RUN_ONLY="--with fireducks --with 'pandas>2.2.0'" CMD="python -m fireducks.pandas --hook-import-module -m pytest tests/fireducks_test.py"
--- a/.github/workflows/pytest-fireducks.yml
+++ b/.github/workflows/pytest-fireducks.yml
@@ -1,0 +1,29 @@
+name: PyTest FireDucks
+
+on:
+  pull_request:
+  workflow_call:
+
+env:
+  PY_COLORS: 1
+  UV_LOCKED: 0
+  UV_NO_DEV: 1
+
+permissions:
+  contents: read
+
+jobs:
+  pytest-fireducks:
+    if: github.head_ref != 'bump-version'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          python-version: "3.11"
+          enable-cache: "true"
+          cache-suffix: fireducks-3.11
+          cache-dependency-glob: "pyproject.toml"
+      - name: Run pytest
+        run: make run-ci DEPS="-U --group tests --extra pandas" RUN_ONLY="--with fireducks" CMD="python -m fireducks.pandas --hook-import-module -m pytest tests/fireducks_test.py"
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -49,6 +49,12 @@ jobs:
     permissions:
       contents: read
 
+  pytest-fireducks:
+    name: PyTest Fireducks
+    uses: ./.github/workflows/pytest-fireducks.yml
+    permissions:
+      contents: read
+
   typing:
     name: Type checking
     uses: ./.github/workflows/typing.yml
--- a/src/narwhals/_native.py
+++ b/src/narwhals/_native.py
@@ -104,10 +104,12 @@ They differ by checking **all** native types/protocols in a single-call and usin
 
 from __future__ import annotations
 
+import sys
 from collections.abc import Callable, Collection, Iterable, Sized
 from typing import TYPE_CHECKING, Any, Protocol, TypeVar, cast
 
 from narwhals.dependencies import (
+    IMPORT_HOOKS,
     get_cudf,
     get_modin,
     get_pandas,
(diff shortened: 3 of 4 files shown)


.github/workflows/pytest-fireducks.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Install uv
+        uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
+        with:
+          python-version: "3.13"
+          enable-cache: "true"
+          cache-suffix: fireducks-3.13
+          cache-dependency-glob: "pyproject.toml"
+      - name: Run pytest
+        run: make run-ci DEPS="--group core-tests --extra pandas" RUN_ONLY="--with fireducks --with 'pandas>2.2.0'" CMD="python -m fireducks.pandas --hook-import-module -m pytest tests/fireducks_test.py"
```
MarcoGorelli, line 29: without the pandas minimum bound here, I get

```
  × Failed to build `pandas==2.0.3`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta:__legacy__.build_wheel` failed (exit status: 1)

      [stderr]
        File "<string>", line 14, in <module>
          requires = get_requires_for_build({})
        File "/home/mgorelli/.cache/uv/builds-v0/.tmp3BFhUw/lib/python3.13/site-packages/setuptools/build_meta.py", line 333, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/mgorelli/.cache/uv/builds-v0/.tmp3BFhUw/lib/python3.13/site-packages/setuptools/build_meta.py", line 301, in _get_build_requires
          self.run_setup()
          ~~~~~~~~~~~~~~^^
        File "/home/mgorelli/.cache/uv/builds-v0/.tmp3BFhUw/lib/python3.13/site-packages/setuptools/build_meta.py", line 520, in run_setup
          super().run_setup(setup_script=setup_script)
          ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/home/mgorelli/.cache/uv/builds-v0/.tmp3BFhUw/lib/python3.13/site-packages/setuptools/build_meta.py", line 317, in run_setup
          exec(code, locals())
          ~~~~^^^^^^^^^^^^^^^^
        File "<string>", line 19, in <module>
          json.dump(requires, fp)
          ~~~~~~~~~^^^^^^^^^^^^^^
      ModuleNotFoundError: No module named 'pkg_resources'

hint: `pandas` (v2.0.3) was included because `fireducks` (v1.4.4) depends on `pandas`
hint: This error likely indicates that `pandas@2.0.3` depends on `pkg_resources`, but doesn't declare it as a build dependency. If `pandas` is a first-party package, consider adding `pkg_resources` to its `build-system.requires`. Otherwise, either add it to your `pyproject.toml` under:

[tool.uv.extra-build-dependencies]
pandas = ["pkg_resources"]

or `uv pip install pkg_resources` into the environment and re-run with `--no-build-isolation`.make: *** [Makefile:36: run-ci] Error 1
```

Without `--with fireducks`, there's no such issue

cc @qsourav just fyi in case this is something you can address in fireducks

qsourav: Hi @MarcoGorelli,

I believe the issue is caused by the combination of pandas==2.0.3 and Python 3.13. pandas 2.0.3 only supports up to Python 3.11, so it may not build correctly on Python 3.13. Could you try using pandas==2.3.3 instead and see if that resolves the problem?

MarcoGorelli: i'm not specifying pandas 2.0.3 anywhere, is the bound coming from fireducks?

qsourav: It should not be coming from FireDucks (as the lower bound is 1.5x). "uv lock" seems to be disabled as well in your case, so it should not be a locked uv issue. I am not sure the exact cause, but "uv lock --upgrade-package pandas" or selecting python3.11 might help?

MarcoGorelli: 🤷 don't know then - @FBruzzesi happy to know on the uv side?

anyway i don't think this is a blocker, putting `pandas>2.2` here works, we can make a patch release with this if it solves the issue

FBruzzesi (approved):
Thanks @MarcoGorelli 

My best guess is that `extreme-not-so-old-versions` group is used as feasible for resolution. Honestly it's hard to keep track of all the groups with possible conflicts considering that some boundaries are not only self imposed by us for testing, but also from upstream pinning.

MarcoGorelli:
thanks! ok let's do this

PR #3673, opened 2026-06-04
add available_default_cpu_backends

EdAbati:
[Note this PR is not for `main` but for @FBruzzesi PR :)]

As discussed in https://github.com/narwhals-dev/narwhals/pull/3552/changes#r3355590221

### DIFF
--- a/src/narwhals/testing/constructors.py
+++ b/src/narwhals/testing/constructors.py
@@ -64,6 +64,7 @@ if TYPE_CHECKING:
 __all__ = (
     "available_backends",
     "available_cpu_backends",
+    "available_default_cpu_backends",
     "frame_constructor",
     "get_backend_constructor",
     "is_backend_available",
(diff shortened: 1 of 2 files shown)


src/narwhals/testing/constructors.py
```diff
@@ -365,6 +374,7 @@ def modin_constructor(obj: Data, /, **kwds: Any) -> IntoDataFrame:  # pragma: no
     implementation=Implementation.MODIN,
     requirements=("modin", "pyarrow"),
     is_eager=True,
+    default_include=False,
 )
```
EdAbati, line 378: I guess also `modin[pyarrow]` shouldn't be included by default?

FBruzzesi: Correct!

FBruzzesi (commented):
Thanks @EdAbati ! Feel free to merge if you are happy with it

PR #3680, opened 2026-06-11
ci: Run workflows only if changes involves them

FBruzzesi:
# Description

Followup on https://github.com/narwhals-dev/narwhals/pull/3609#issuecomment-4649760945

Trying to follow the same polars philosophy:

> Only run checks when they need to be run

---

Difference between `src/**` and `**.py` case is to distinguish between:

* Something changed in the source code that we ship, then we should run all downstream deps
* Any python file changes (e.g. tpch, test-plugin, etc), then our test suite should run

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other


.github/workflows/extremes.yml
```diff
@@ -2,6 +2,8 @@ name: Min, old, and nightly versions
 
 on:
   pull_request:
+    paths:
+      - "**.py"
```
dangotbanned: Just want to double check, are these patterns considered equal in GitHub Actions?

- `**.py`
- `*.py`
- `**/*.py`

If it uses normal globbing rules, I think `**/*.py` does the job?

FBruzzesi: The example from the github docs is:

> '**.js': Matches all .js files in the repository.

Source: [patterns-to-match-file-paths](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#patterns-to-match-file-paths).

dangotbanned: oof 🤦‍♂️

dangotbanned:
@FBruzzesi If you wanna move this out of draft I'll review it.

It seems ready to me, but I don't wanna prioritize it over others that you've marked as ready 🙂

FBruzzesi:
@dangotbanned my only concern is about the repo ruleset for merging. Since some jobs do not run, they don't show up as succeeded, therefore merging is not available. I will try to take a look at the github docs this week

dangotbanned:
Ah I see, yeah I think changing those rules is above my pay grade sadly 😅

FBruzzesi:
@dangotbanned according to https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-jobs-with-conditions 

>  A job that is skipped will report its status as "Success". It will not prevent a pull request from merging, even if it is a required check.

---

I think that if you/someone approves it, then it should act as any other PR

.github/workflows/check_tpch_queries.yml
```diff
@@ -2,6 +2,8 @@ name: Tests for TPCH Queries
 
 on:
   pull_request:
+    paths:
+      - "**.py"
```
dangotbanned:
```suggestion
    paths:
      - "**.py"
      - .github/workflows/check_tpch_queries.yml
```

.github/workflows/downstream_tests.yml
```diff
@@ -2,6 +2,8 @@ name: Test Downstream Libraries - Fast
 
 on:
   pull_request:
+    paths:
+      - src/**
```
dangotbanned:
```suggestion
    paths:
      - src/**
      - .github/workflows/downstream_tests.yml
```

.github/workflows/extremes.yml
dangotbanned:
```suggestion
      - "**.py"
      - .github/workflows/extremes.yml
```

.github/workflows/pytest-fireducks.yml
```diff
@@ -2,6 +2,9 @@ name: PyTest FireDucks
 
 on:
   pull_request:
+    paths:
+      - src/**
+      - tests/**
```
dangotbanned:
```suggestion
      - src/**
      - tests/**
      - .github/workflows/pytest-fireducks.yml
```
Also we might be able to refine more with changes to pandas/dependencies, perhaps?

.github/workflows/pytest-ibis.yml
```diff
@@ -2,6 +2,11 @@ name: PyTest Ibis constructor
 
 on:
   pull_request:
+    paths:
+      - src/narwhals/_expression_parsing.py
+      - src/narwhals/_ibis/**
+      - src/narwhals/_sql/**      
+      - tests/**
```
dangotbanned:
```suggestion
      - tests/**
      - .github/workflows/pytest-ibis.yml
```

> [!NOTE]
> I keep forgetting to mention, but do we still need this job?

.github/workflows/pytest-modin.yml
```diff
@@ -2,6 +2,10 @@ name: PyTest Modin constructor
 
 on:
   pull_request:
+    paths:
+      - src/narwhals/_expression_parsing.py
+      - src/narwhals/_pandas_like/**
+      - tests/**
```
dangotbanned:
```suggestion
      - tests/**
      - .github/workflows/pytest-modin.yml
```

.github/workflows/random_ci_pytest.yml
```diff
@@ -2,6 +2,8 @@ name: Random Versions
 
 on:
   pull_request:
+    paths:
+      - "**.py"
```
dangotbanned:
```suggestion
      - "**.py"
      - .github/workflows/random_ci_pytest.yml
```

.github/workflows/pytest.yml
```diff
@@ -2,6 +2,8 @@ name: PyTest
 
 on:
   pull_request:
+    paths:
+      - "**.py"
```
dangotbanned:
```suggestion
      - "**.py"
      - .github/workflows/pytest.yml
```

.github/workflows/typing.yml
```diff
@@ -2,6 +2,8 @@ name: Type checking
 
 on:
   pull_request:
+    paths:
+      - "**.py"
```
dangotbanned: Finally one that is more interesting!

It is pretty common for typing issues to arise on version bumps, when we've done nothing internally

```suggestion
      - "**.py"
      - .github/workflows/typing.yml
      - pyproject.toml
      - uv.lock
```

dangotbanned (approved):
Thanks @FBruzzesi!

I stole the self-ref idea from (https://github.com/pola-rs/polars/blob/7b05d67937a6b715a813b65a20ee8fa8a01d4950/.github/workflows/test-python.yml)

narwhal hugging dolphin

PR #3683, opened 2026-06-12
feat(ibis): support y/q/mo/d units in dt.offset_by

### LINKED ISSUE #3681: [Enh]: support other units in expr.dt.offset_by for ibis backend

Thanks for the great library! We use narwhals at $WORK to write backend-agnostic code. One of our backends is Snowflake (via the Ibis backend).

Would it be possible to enable offsetting by other units (like month) in .dt.offset_by for the ibis backend? It appears that all of the pieces to support it are already there. For example, in the core function that does the interval

```
def timedelta_to_ibis_interval(td: timedelta) -> ibis.expr.types.temporal.IntervalScalar:
    return ibis.interval(days=td.days, seconds=td.seconds, microseconds=td.microseconds)
```

it directly uses ibis.interval, which supports nanoseconds, microseconds, milliseconds, seconds, minutes, hours, days, weeks, months, quarters, and years. Would it just be a matter of relaxing this check (https://github.com/narwhals-dev/narwhals/blob/200baa7fbbbc33ecbffef4fa1cf759d0ac050477/src/narwhals/_ibis/expr_dt.py#L66)?

```
if unit in {"y", "q", "mo", "d", "ns"}:
      msg = f"Offsetting by {unit} is not yet supported for ibis."
```

Relax the check https://github.com/narwhals-dev/narwhals/blob/200baa7fbbbc33ecbffef4fa1cf759d0ac050477/src/narwhals/_ibis/expr_dt.py#L66

Creating a custom .dt.offset_by function

I would be willing to contribute this feature.

gaoflow:
Closes #3681.

`dt.offset_by` previously raised `NotImplementedError` for `y`, `q`, `mo`, `d` and `ns` on the ibis backend. `ibis.interval` already accepts `years`/`quarters`/`months`/`days` (and the existing `UNITS_DICT_BUCKET` maps every narwhals unit to the matching kwarg), so the offset is now built directly from `(multiple, unit)` rather than routed through `timedelta_to_ibis_interval` — a `timedelta` cannot represent calendar units, which is why those were blocked. Only `ns` stays unsupported, since duckdb has no nanosecond interval resolution (`sqlglot ... doesn't support nanosecond interval resolutions`). The now-unused `timedelta_to_ibis_interval` helper is removed.

The xfail markers for `y`/`q`/`mo`/`d` on ibis are dropped from `test_offset_by` and `test_offset_by_3471`, which now pass.

### On the DST question

@MarcoGorelli — you asked to check timezone-aware DST behaviour. I did, and there is a caveat worth a decision:

* **Naive datetimes:** ibis matches polars exactly on every unit. The generated SQL is `expr + INTERVAL <n> <UNIT>`, identical to the already-supported duckdb backend.
* **tz-aware, crossing a DST boundary:** offsetting `2020-10-25 00:00 Europe/Amsterdam` by `1d` gives
  * polars → `2020-10-26 00:00+01:00` (wall-clock, instant shifts 25h)
  * **duckdb backend** → same as polars (`23:00Z`)
  * **ibis backend** → `22:00Z` (fixed 24h)

  The divergence is because ibis stores tz-aware columns as `timestamp(UTC)` and loses the original zone, so `INTERVAL 1 DAY` is added against UTC rather than wall-clock. This isn't reachable through the current test suite — `convert_time_zone` is still `not_implemented` for ibis, so all the `*_tz` / `*_dst` cases remain xfailed for ibis — but a user passing a native tz-aware ibis table would hit it.

I kept this PR to the naive case (which is correct and now tested). For tz-aware calendar offsets on ibis, the options I see are: (a) document the UTC/fixed-offset semantics, or (b) guard tz-aware + calendar-unit and raise until `convert_time_zone` lands. Happy to follow up with whichever you prefer.

### DIFF
--- a/src/narwhals/_ibis/expr_dt.py
+++ b/src/narwhals/_ibis/expr_dt.py
@@ -2,12 +2,10 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
 
+import ibis
+
 from narwhals._duration import Interval
-from narwhals._ibis.utils import (
-    UNITS_DICT_BUCKET,
-    UNITS_DICT_TRUNCATE,
-    timedelta_to_ibis_interval,
-)
+from narwhals._ibis.utils import UNITS_DICT_BUCKET, UNITS_DICT_TRUNCATE
 from narwhals._sql.expr_dt import SQLExprDateTimeNamesSpace
 from narwhals._utils import not_implemented
 
@@ -63,10 +61,10 @@ class IbisExprDateTimeNamespace(SQLExprDateTimeNamesSpace["IbisExpr"]):
     def offset_by(self, by: str) -> IbisExpr:
         interval = Interval.parse_no_constraints(by)
         unit = interval.unit
-        if unit in {"y", "q", "mo", "d", "ns"}:
-            msg = f"Offsetting by {unit} is not yet supported for ibis."
+        if unit == "ns":
+            msg = "Offsetting by nanoseconds is not yet supported for ibis."
             raise NotImplementedError(msg)
-        offset = timedelta_to_ibis_interval(interval.to_timedelta())
+        offset = ibis.interval(**{UNITS_DICT_BUCKET[unit]: interval.multiple})
         return self.compliant._with_callable(lambda expr: expr.add(offset))
 
     def replace_time_zone(self, time_zone: str | None) -> IbisExpr:
--- a/src/narwhals/_ibis/expr_dt.py
+++ b/src/narwhals/_ibis/expr_dt.py
@@ -60,12 +60,28 @@ class IbisExprDateTimeNamespace(SQLExprDateTimeNamesSpace["IbisExpr"]):
 
     def offset_by(self, by: str) -> IbisExpr:
         interval = Interval.parse_no_constraints(by)
-        unit = interval.unit
+        multiple, unit = interval.multiple, interval.unit
         if unit == "ns":
             msg = "Offsetting by nanoseconds is not yet supported for ibis."
             raise NotImplementedError(msg)
-        offset = ibis.interval(**{UNITS_DICT_BUCKET[unit]: interval.multiple})
-        return self.compliant._with_callable(lambda expr: expr.add(offset))
+        offset = ibis.interval(**{UNITS_DICT_BUCKET[unit]: multiple})
+
+        def fn(expr: ir.TimestampValue) -> ir.TimestampValue:
+            # Ibis stores timezone-aware data as UTC, so calendar offsets are
+            # applied to the underlying instant rather than wall-clock time.
+            # The result would differ from other backends across a DST
+            # transition, so refuse it rather than return a surprising value.
+            tz = getattr(expr.type(), "timezone", None)
+            if unit in {"y", "q", "mo", "d"} and tz is not None:
+                msg = (
+                    f"Offsetting timezone-aware data by {unit} is not supported "
+                    "for ibis, as the result would be incorrect across daylight "
+                    "saving time transitions."
+                )
+                raise NotImplementedError(msg)
+            return expr.add(offset)
+
+        return self.compliant._with_callable(fn)
 
     def replace_time_zone(self, time_zone: str | None) -> IbisExpr:
         if time_zone is None:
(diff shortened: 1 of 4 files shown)


MarcoGorelli:
thanks for your pr!

i think we may need to raise in the tz-aware case for ibis if they do something different

binggao1230:
Done — pushed a guard that raises `NotImplementedError` for timezone-aware data when the unit is a calendar unit (`d`/`mo`/`q`/`y`). Those are exactly the cases where ibis diverges (it offsets the underlying UTC instant, so the result differs from polars/duckdb across a DST boundary). Pure-duration units (`h` and smaller) are unambiguous, so they stay supported even on tz-aware columns. Added a test covering both the raise and the still-working `7h` case.

src/narwhals/_ibis/expr_dt.py
```diff
@@ -62,12 +60,29 @@ def truncate(self, every: str) -> IbisExpr:
...
+        kwds: dict[BucketUnit, Any] = {UNITS_DICT_BUCKET[unit]: multiple}
+        offset = ibis.interval(**kwds)
+
+        def fn(expr: ir.TimestampValue) -> ir.TimestampValue:
+            # Ibis stores timezone-aware data as UTC, so calendar offsets are
+            # applied to the underlying instant rather than wall-clock time.
+            # The result would differ from other backends across a DST
+            # transition, so refuse it rather than return a surprising value.
+            tz = getattr(expr.type(), "timezone", None)
+            if unit in {"y", "q", "mo", "d"} and tz is not None:
```
FBruzzesi, line 82: Given the comment, should a genuine UTC timezone be accepted?

src/narwhals/_ibis/expr_dt.py
```diff
@@ -62,12 +60,29 @@ def truncate(self, every: str) -> IbisExpr:
...
+                msg = (
+                    f"Offsetting timezone-aware data by {unit} is not supported "
```
FBruzzesi: Nitpick on the message: for nanoseconds the message expands the unit (checks for 'ns', report 'nanoseconds'). Should we map the unit here as well (to `UNITS_DICT_BUCKET[unit]`)?

tests/expr_and_series/dt/offset_by_test.py
```diff
@@ -266,8 +266,31 @@ def test_offset_by_date_pandas() -> None:
...
+
+    con = ibis.duckdb.connect()
+    tbl = con.create_table(
+        "offset_by_tz",
+        pl.LazyFrame({"a": [datetime(2020, 10, 25)]}).with_columns(
+            pl.col("a").dt.replace_time_zone("Europe/Amsterdam")
+        ),
+    )
+    df = nw.from_native(tbl)
+    for by in ("1d", "3mo", "2q", "1y"):
```
FBruzzesi: Can these values be passed via `pytest.mark.parametrize` rather than looping within the test?

src/narwhals/_ibis/utils.py
FBruzzesi, line 1: Nice 👌🏼

FBruzzesi (commented):
Thanks for the contribution @gaoflow - I left a couple of nitpicks.

One additional ask is to check if `ibis.interval(quarters=...)` is safe on the minimum supported ibis version, given `truncate` deliberately avoids passing quarters through; otherwise mirror its `q` to months rewrite.

binggao1230:
Thanks for the review @FBruzzesi! Addressed all three in 9c5957e:

- **Quarters on old ibis**: good catch. Rather than rely on `ibis.interval(quarters=...)`, I now mirror what `truncate` does and rewrite a quarter offset to the equivalent number of months (`q` -> `3 * multiple` months) before building the interval, so nothing version-specific is passed through. The naive results are unchanged (e.g. `offset_by("2q")` still lands on the same date as `+6mo`).

- **Error message**: it now expands the unit via `UNITS_DICT_BUCKET[unit]` (e.g. "Offsetting timezone-aware data by months ..."), consistent with the nanoseconds message.

- **Test**: switched the in-body loop to `pytest.mark.parametrize`, using `does_not_raise()` for the still-supported `7h` case.

mypy + pyright are clean and the full `offset_by` suite passes across ibis/duckdb/polars/pandas/pyarrow.

FBruzzesi (approved):
Thanks for bearing with me and addressing all the points @gaoflow 
I added a commit to add a note in the public method docstring.

MarcoGorelli (approved):
thanks!

PR #3685, opened 2026-06-12
refactor: Testing constructors with default namespace

FBruzzesi:
@dangotbanned this branch has two commits, one for typing one to add the default namespace in `frame_constructor._call__`.

Let me know what you think


tests/preserve_pandas_like_columns_name_attr_test.py
```diff
@@ -22,6 +22,9 @@ def test_ops_preserve_column_index_name(
         request.applymarker(pytest.mark.xfail)
 
     data = {"a": [1, 3, 2], "b": [4, 4, 6], "z": [7.0, 8.0, 9.0]}
+    # NOTE: the pandas-like-only `.columns.name` accesses below go unchecked because
+    # `.columns` is loosely typed on the native protocols (`main` typed `constructor`
+    # as `Callable[..., pd.DataFrame]` in this test).
     df_native = constructor(data).to_native()
     df_native.columns.name = "foo"
```
dangotbanned: Since we have quite a few `pandas`-specific tests, it may be worth adding a `Protocol` we can use for them.

<details><summary>Show <code>FrameConstructor</code></summary>
<p>

```py

from typing import Any, Protocol, TypeAlias, TypeVar

from narwhals import DataFrame, LazyFrame

FrameT_co = TypeVar(
    "FrameT_co", bound=nw.DataFrame[Any] | nw.LazyFrame[Any], covariant=True
)
Incomplete: TypeAlias = Any
"""Fill these in @FBruzzesi!"""

Data: TypeAlias = Incomplete
NarwhalsNamespace: TypeAlias = Incomplete


class FrameConstructor(Protocol[FrameT_co]):
    """Pick whatever name you like!"""

    # NOTE: Include any attributes that are referenced in a test
    # exclude any that are part of the setup
    is_eager: bool
    nan_is_null: bool
    needs_gpu: bool
    default_include: bool

    # NOTE: Would add just `__hash__` but that could be an issue w/ inheritance
    def __hash__(self) -> int: ...
    def __eq__(self, other: object) -> bool: ...

    def __call__(
        self, obj: Data, /, namespace: NarwhalsNamespace | None = None, **kwds: Any
    ) -> FrameT_co: ...

    # NOTE: Defining the docs here means:
    # - `frame_constructor` can inherit them
    # - this protocol can be used in an annotation and still provide docs
    @property
    def identifier(self) -> str:
        """Instance-level string identifier for test IDs."""
        ...

    @property
    def is_lazy(self) -> bool:
        """Whether this constructor produces a lazy native frame."""
        ...

    @property
    def is_pandas(self) -> bool:
        """Whether this is one of the pandas constructors."""
        ...

    @property
    def is_modin(self) -> bool:
        """Whether this is one of the modin constructors."""
        ...

    @property
    def is_cudf(self) -> bool:
        """Whether this is the cudf constructor."""
        ...

    @property
    def is_pandas_like(self) -> bool:
        """Whether this constructor produces a pandas-like dataframe (pandas, modin, cudf)."""
        ...

    @property
    def is_polars(self) -> bool:
        """Whether this is one of the polars constructors."""
        ...

    @property
    def is_pyarrow(self) -> bool:
        """Whether this is the pyarrow table constructor."""
        ...

    @property
    def is_dask(self) -> bool:
        """Whether this is the dask constructor."""
        ...

    @property
    def is_duckdb(self) -> bool:
        """Whether this is the duckdb constructor."""
        ...

    @property
    def is_pyspark(self) -> bool:
        """Whether this is one of the pyspark constructors."""
        ...

    @property
    def is_sqlframe(self) -> bool:
        """Whether this is the sqlframe constructor."""
        ...

    @property
    def is_ibis(self) -> bool:
        """Whether this is the ibis constructor."""
        ...

    @property
    def is_spark_like(self) -> bool:
        """Whether this constructor uses a spark-like backend (pyspark, sqlframe)."""
        ...

    @property
    def needs_pyarrow(self) -> bool:
        """Whether this constructor requires `pyarrow` to be installed."""
        ...

    @property
    def is_available(self) -> bool:
        """Whether every package this constructor needs is importable."""
        ...

```


</p>
</details> 

Then we can add aliases like this:

```py
import pandas as pd

PandasConstructor: TypeAlias = FrameConstructor[nw.DataFrame[pd.DataFrame]]
```

And fix the typing like this 😄 

```diff
def test_ops_preserve_column_index_name(
-    constructor: Constructor, request: pytest.FixtureRequest
+    constructor: PandasConstructor, request: pytest.FixtureRequest
) -> None:
```

tests/dependencies/is_into_dataframe_test.py
```diff
@@ -41,22 +41,24 @@ def drop(self, *args: Any, **kwargs: Any) -> Any: ...
...
-def test_is_into_dataframe(constructor: Constructor) -> None:
-    native_frame = constructor(data).to_native()
-    nw_frame = nw.from_native(native_frame)
-    nw_v1_frame = nw_v1.from_native(native_frame)
+def test_is_into_dataframe(nw_frame: FrameConstructor) -> None:
+    native_frame = nw_frame(data, nw).to_native()
+    frame = nw.from_native(native_frame)
+    # NOTE: `type-var` because v1 `from_native` typing does not admit `NativeIbis`,
+    # which `IntoLazyFrame` (hence the constructor return) includes.
+    nw_v1_frame = nw_v1.from_native(native_frame)  # type: ignore[type-var]
```
dangotbanned: I can't tell if the `ibis` one should be raising or not?

```suggestion
    from narwhals.dependencies import is_ibis_table

    if is_ibis_table(native_frame):
        nw_v1.from_native(native_frame)  # type: ignore[type-var]
    else:
        nw_v1_frame = nw_v1.from_native(native_frame)
```

tests/dependencies/is_into_lazyframe_test.py
```diff
@@ -42,7 +42,9 @@ def join(self, *args: Any, **kwargs: Any) -> Any: ...
 def test_is_into_lazyframe(constructor: Constructor) -> None:
     native_frame = constructor(data).to_native()
     nw_frame = nw.from_native(native_frame)
-    nw_v1_frame = nw_v1.from_native(native_frame)
+    # NOTE: `type-var` because v1 `from_native` typing does not admit `NativeIbis`,
+    # which `IntoLazyFrame` (hence the constructor return) includes.
```
dangotbanned: Same comment as [`is_into_dataframe`](https://github.com/narwhals-dev/narwhals/pull/3685#discussion_r3412782744)

tests/dependencies/is_into_dataframe_test.py
dangotbanned: > [!NOTE]
> Feel free to defer this to an issue to be resolved on `main` first 👍

Tbh it might be easier to do this with the tests split up some more.

I missed it while reviewing (#3613) - but there isn't much typing coverage being exercised.

## What?
We check this input type:

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/tests/utils.py#L48


against things that depend the same alias:

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/src/narwhals/dependencies.py#L657

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/src/narwhals/_native.py#L354

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/src/narwhals/_native.py#L297

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/src/narwhals/_native.py#L310

## Why is that bad?
We can change the definition of `IntoDataFrame` and the typing for the tests will still stay green.

## Fix?
We should be testing the actual native types, similar to:

<details><summary><code>tests.implementation_test._TYPING_ONLY_TESTS</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/8f6b947c8ab0ff3c4f9df6fd02081a8a89293b67/tests/implementation_test.py#L76-L319

</p>
</details>


We can still do the short versions for *runtime tests*, and this would be `if TYPE_CHECKING:`-only (free for `pytest`)

tests/preserve_pandas_like_columns_name_attr_test.py
dangotbanned: The nice thing about the separate `Protocol` is that you have the flexibility of defining both the narwhals & native types in one swoop 😎

tests/frame/schema_test.py
```diff
@@ -581,7 +581,9 @@ def origin_pandas_like(
         "d": [5.3, 4.99],
         "e": [datetime(2006, 1, 1), datetime(2001, 9, 3)],
     }
-    return constructor_pandas_like(data).to_native().dtypes.to_dict()  # type: ignore[no-any-return]
+    # NOTE: `attr-defined` because the `IntoDataFrame` protocol does not declare `.dtypes`.
```
dangotbanned: Same comment as [`tests/preserve_pandas_like_columns_name_attr_test.py`](https://github.com/narwhals-dev/narwhals/pull/3685#discussion_r3412726541)

tests/frame/schema_test.py
```diff
@@ -606,7 +608,8 @@ def origin_pandas_like_pyarrow(
     df_nw = nw.from_native(df_pd).with_columns(
         nw.col("f").cast(nw.Date()), nw.col("g").cast(nw.Time())
     )
-    return df_nw.to_native().dtypes.to_dict()  # type: ignore[no-any-return]
+    # NOTE: `attr-defined` for the same reason as `origin_pandas_like` above.
```
dangotbanned: Same comment as [`tests/preserve_pandas_like_columns_name_attr_test.py`](https://github.com/narwhals-dev/narwhals/pull/3685#discussion_r3412726541)

docs/api-reference/testing.md
```diff
@@ -55,13 +54,13 @@ if TYPE_CHECKING:
 
 def test_shape(nw_dataframe: DataFrameConstructor) -> None:
     data: Data = {"x": [1, 2, 3]}
-    df = nw_dataframe(data, namespace=nw)
+    df = nw_dataframe(data)  # wraps with the main `narwhals` namespace by default
```
dangotbanned: I haven't got any code block in markdown yet for #3497, but I want to use this when that comes up
- https://zensical.org/docs/authoring/code-blocks/#code-annotations

docs/api-reference/testing.md
```diff
@@ -55,13 +54,13 @@ if TYPE_CHECKING:
...
     assert df.shape == (3, 1)
 
 
 def test_laziness(nw_lazyframe: LazyFrameConstructor) -> None:
     data: Data = {"x": [1, 2, 3]}
-    lf = nw_lazyframe(data, namespace=nw_v2)
+    lf = nw_lazyframe(data, namespace=nw_v2)  # opt in to a stable namespace
```
dangotbanned: [Ditto](https://github.com/narwhals-dev/narwhals/pull/3685#discussion_r3413066240)

tests/conftest.py
```diff
@@ -123,62 +116,28 @@ def nested_dtype(request: pytest.FixtureRequest) -> NestedOrEnumDType:
...
-    ) -> DataFrame[Any]:
-        return cast("DataFrame[Any]", self._inner(obj, namespace=namespace, **kwds))
+# The following fixtures are short-name aliases of those registered in
+# `narwhals/testing/pytest_plugin.py`. Calling a constructor without an explicit
+# `namespace` defaults to the main `narwhals` namespace; tests can still pass
+# `nw_v1` / `nw_v2` explicitly to opt in to a stable namespace. The legacy pattern
+# `nw.from_native(constructor(data))` keeps working because `nw.from_native` is
+# idempotent on narwhals objects.
+# TODO(FBruzzesi): Drop these aliases once every test requests `nw_frame` /
+# `nw_dataframe` / `nw_pandas_like_frame` directly.
```
dangotbanned, line 127: Just make this a sub-issue of #3653 and save yourself from the merge conflict pains 😂

dangotbanned (commented):
Thanks for getting to this so quickly @FBruzzesi!

I have a few suggestions and 1 follow-up issue that we might wanna address on `main` instead

tests/preserve_pandas_like_columns_name_attr_test.py
FBruzzesi: Went with the "quick win" in https://github.com/narwhals-dev/narwhals/pull/3685/commits/75a581cbda130c15de92f08f70d9cafc06060f44

FBruzzesi: @dangotbanned sorry how does this differ from `frame_constructor`?

dangotbanned: > @dangotbanned sorry how does this differ from `frame_constructor`?

### (1)
It uses forward references, and they won't link on the API reference [^1]

[^1]: That was my 2nd most common issue in (https://github.com/narwhals-dev/narwhals/compare/expr-ir/docs/are-we-docs-yet...expr-ir/docs/make-mkdocstrings-happy#files_bucket)

AFAICT, `frame_constructor` isn't on the API reference either, - so maybe there isn't a way to display the docstrings currently?

### (2)
A separate `Protocol` avoids needing to resolve the `@overload`s in `frame_constructor.__call__`.

This is cheaper for a type checker [^2], because they're given `FrameT_co` upfront and it is the only return type

```py
    def __call__(
        self, obj: Data, /, namespace: NarwhalsNamespace | None = None, **kwds: Any
    ) -> FrameT_co: ...
```

[^2]: `pyright` is great at understanding `@overload`s, but they can be expensive when you use [inlay hints](https://github.com/microsoft/pylance-release/blob/main/docs/settings/python_analysis_inlayHints_variableTypes.md#performance-considerations)

### (3)
The idea was that `frame_constructor` *inherits from* the protocol.
In (https://github.com/narwhals-dev/narwhals/pull/3685#discussion_r3412726541) I left out the method impls, but you could just as easily implement them there if extra code is a concern?

The protocol also isn't too far off from being able to support `Series` 😉

FBruzzesi: Thanks for explaining @dangotbanned ! If you fancy adding a commit to address this, please go ahead! I won't have time to make code changes before mid next week most likely (but at the same time I would like to see the testing constructors moving forward 😅)

tests/dependencies/is_into_dataframe_test.py
FBruzzesi: Added the suggestion in https://github.com/narwhals-dev/narwhals/pull/3685/commits/395d591f42502101fea798ef37e796e7eabaa9f5

tests/preserve_pandas_like_columns_name_attr_test.py
FBruzzesi: Took me a while to come back to this bit but we finally have it: https://github.com/narwhals-dev/narwhals/pull/3685/commits/8afc46fb88d7b1e948e2354da3510b784b88845c

I hope it goes into the direction you were hoping for

PR #3688, opened 2026-06-15
ci: Simplify prek CI run

FBruzzesi:
# Description

It's not much, but yesterday I discovered the prek action. So we can avoid manually doing it ourselves.

From https://github.com/j178/prek-action:

> prek is always invoked as:
>
> ```terminal
> prek run --show-diff-on-failure --color=always <extra-args>
> ```


.pre-commit-config.yaml
FBruzzesi, line 1: See https://github.com/j178/prek-action/issues/62

FBruzzesi:
Nevermind! Local hooks aren't working as expected. Closing

PR #3694, opened 2026-06-19
feat: add nw.list

### LINKED ISSUE #3691: [Enh]: add nw.concat_list
## Related
- [x] https://github.com/pola-rs/polars/issues/8510

_No response_

Narwhals provides a list namespace, but does not provide a way to turn multiple columns into the list datatype.

For polars, there is pl.concat_list(). For ibis, there is ibis.array(). 

I am not as familiar with pyarrow, pandas, cudf, and modin, but for pyarrow I believe there is pa.array(). For pandas-like, I believe you could use .values.tolist(), but that converts things into a python list object which would probably be pretty slow.

Would it be acceptable for an initial implementation to only consider polars, ibis, and pyarrow?

_No response_

I would be interested in contributing this feature.

CangyuanLi:
# Description

Adds a nw.list function that constructs a list datatype from columns. It mimics duckdb's list_pack semantics. Scalars are collected into a list and lists are collected into a list of lists. 

## If you have comments or want to explain your changes, please do so in this section

I basically tried to mimic nw.struct wherever possible. For example, since nw.struct is not implemented for dask, neither is nw.list.

I also added a scalars_only flag which is false by default. This allows us to directly use concat_list for the polars backend without calling .implode().over(). It does nothing for all other backends.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue https://github.com/narwhals-dev/narwhals/issues/3691
- Closes https://github.com/narwhals-dev/narwhals/issues/3691

## Checklist

- [x] Tests added
- [x] Documented the changes


src/narwhals/functions.py
```diff
@@ -1975,3 +1975,59 @@ def struct(*exprs: IntoExpr | Sequence[IntoExpr], **named_exprs: IntoExpr) -> Ex
...
+        >>> data = {
+        ...     "a": [1, 2, 3],
+        ...     "b": [4, None, 6],
+        ...     "d": [[10, 20], None, [40, 50, 60]],
+        ...     "e": [[100], [200, 300], [None]],
+        ... }
+        >>> df = nw.from_native(pl.DataFrame(data))
+        >>> df.select(
+        ...     nw.list("a", "b").alias("list"), nw.list("d", "e").alias("list_of_lists")
+        ... )
```
MarcoGorelli: can we keep it simple and just have:
- a
- b

and use with_columns to make a `a_b` column?

MarcoGorelli (changes requested):
thanks @CangyuanLi !

my main question is - what's `scalars_only`?

CangyuanLi:
> thanks @CangyuanLi !
> 
> my main question is - what's `scalars_only`?

it is only relevant for thr polars backend and uses concat_list directly instead of the .implode().over() trick. I was worried about the potential performance impact but maybe its leaking implementation details too much?

src/narwhals/functions.py
CangyuanLi: fixed!

CangyuanLi:
The stable api tests are failing

```
=================================== short test summary info ===================================
FAILED tests/stable_api_test.py::test_stable_api_docstrings - AttributeError: module 'narwhals.stable.v2' has no attribute 'list'. Did you mean: 'List'?
FAILED tests/stable_api_test.py::test_stable_api_completeness - AssertionError: assert not {'list'}
```

would you want me to edit the tests to explicitly xfail list, or add it to the stable api?

CangyuanLi:
> > thanks @CangyuanLi !
> > my main question is - what's `scalars_only`?
> 
> it is only relevant for thr polars backend and uses concat_list directly instead of the .implode().over() trick. I was worried about the potential performance impact but maybe its leaking implementation details too much?

The performance impact is relatively minimal for concatenating two columns. It's only really much faster for very small dataframes. 

image

However, it does appear to make a pretty big difference as the number of columns to concatenate increases.

image

Timings are the median timing of 50 runs, with the CIs representing the 5th and 95th percentile. At 50 columns and 1 million rows, the median time using scalars_only=False 26.26 seconds. Using scalars_only=True, the median time is 8.9694 seconds, a roughly 2.93x speedup.

src/narwhals/_duckdb/namespace.py
```diff
@@ -252,3 +252,18 @@ def func(df: DuckDBLazyFrame) -> list[Expression]:
...
             version=version,
         )
+
+    def list(self, *exprs: DuckDBExpr, scalars_only: bool) -> DuckDBExpr:
+        version = self._version
+
+        def func(df: DuckDBLazyFrame) -> list[Expression]:
+            cols = [native_expr for expr in exprs for native_expr in expr(df)]
+            col_args = ", ".join(str(col) for col in cols)
+            return [sql_expression(f"list_pack({col_args})")]
```
MarcoGorelli: can you use `FunctionExpression` here? looks like this works

```py
In [5]: rel = duckdb.sql("""
   ...: select a, b
   ...: from values(1, 4), (2, 5) df(a, b)
   ...: """)

In [6]: rel.select(duckdb.FunctionExpression('list_pack', 'a', 'b'))
Out[6]:
┌─────────────────┐
│ list_pack(a, b) │
│     int32[]     │
├─────────────────┤
│ [1, 4]          │
│ [2, 5]          │
└─────────────────┘
```

there should be other examples of this in the narwhals codebase

src/narwhals/_pandas_like/namespace.py
```diff
@@ -404,6 +404,43 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             context=self,
         )
 
+    def list(self, *exprs: PandasLikeExpr, scalars_only: bool) -> PandasLikeExpr:
+        def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
+            try:
+                import pandas as pd  # ignore-banned-import
```
MarcoGorelli: you can put this one at the top of the file, it shouldn't be a banned-import in `_pandas_like` (there may be other places where you can put it at the top of the file in this file)

src/narwhals/_pandas_like/namespace.py
```diff
@@ -404,6 +404,43 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
...
+                    "Please install pyarrow: pip install pyarrow"
+                )
+                raise ImportError(msg) from exc
+
+            from narwhals._arrow.utils import build_list_array
+
+            align = self._series._align_full_broadcast
+            series = align(*chain.from_iterable(expr(df) for expr in exprs))
+            result = build_list_array(
+                [pa.array(s.native, from_pandas=True) for s in series]
```
MarcoGorelli: i think you can access the arrow array undereat `s` directly, check some other functions in this file for how

src/narwhals/_polars/namespace.py
```diff
@@ -209,6 +209,16 @@ def struct(self, *exprs: PolarsExpr) -> PolarsExpr:
...
 
+    def list(self, *exprs: PolarsExpr, scalars_only: bool) -> PolarsExpr:
+        pl_exprs: list[pl.Expr] = [expr._native_expr for expr in exprs]
+
+        if scalars_only:
+            to_concat = pl_exprs
+        else:
+            to_concat = [e.implode().over(pl.int_range(pl.len())) for e in pl_exprs]
+
+        return self._expr(pl.concat_list(to_concat), version=self._version)
```
MarcoGorelli: looks like `list` will get into Polars, and their next release may be `2.0`

can we do

```
if self._backend_version < (2,):
    # use your implde/over workaround here
else:
    # just use `pl.list`
```

src/narwhals/_polars/namespace.py
```diff
@@ -245,9 +255,11 @@ def __init__(self, context: _LimitedContext, /) -> None:
...
     def by_dtype(self, dtypes: Iterable[IntoDType]) -> PolarsExpr:
         native_dtypes = [
-            narwhals_to_native_dtype(dtype, self._version).__class__
-            if isinstance(dtype, type) and issubclass(dtype, DType)
-            else narwhals_to_native_dtype(dtype, self._version)
+            (
+                narwhals_to_native_dtype(dtype, self._version).__class__
+                if isinstance(dtype, type) and issubclass(dtype, DType)
+                else narwhals_to_native_dtype(dtype, self._version)
+            )
```
MarcoGorelli: unrelated change?

src/narwhals/functions.py
```diff
@@ -1975,3 +1975,52 @@ def struct(*exprs: IntoExpr | Sequence[IntoExpr], **named_exprs: IntoExpr) -> Ex
...
+def list_(*exprs: IntoExpr | Sequence[IntoExpr], scalars_only: bool = False) -> Expr:
+    """Collect columns into a list column.
+
+    Arguments:
+        *exprs: Column(s) to collect into a list column, specified as
+            positional arguments. Accepts only expression input. Strings are parsed
+            as column names, other non-expression inputs are not allowed.
+        scalars_only: If True, assume all inputs are scalar columns. This allows us to
+            take a fast path in the polars backend, but is unused for all other
+            backends.
```
MarcoGorelli: let's remove this

MarcoGorelli (changes requested):
thanks! a few comments

src/narwhals/_polars/namespace.py
CangyuanLi: I think this is a change that prek made (probably the ruff-format portion). I can revert it and commit with skip checks.

src/narwhals/_polars/namespace.py
CangyuanLi: Changed!

src/narwhals/_pandas_like/namespace.py
CangyuanLi: I changed it to access the underlying array if its arrow and use pa.array if not

src/narwhals/_pandas_like/namespace.py
CangyuanLi: I added it to the top of the file but did not touch any other functions

src/narwhals/_duckdb/namespace.py
CangyuanLi: changed!

src/narwhals/functions.py
CangyuanLi: removed

src/narwhals/_arrow/namespace.py
```diff
@@ -273,7 +293,9 @@ def _if_then_else(
         otherwise = pa.nulls(len(when), then.type) if otherwise is None else otherwise
         return pc.if_else(when.fill_null(pa.scalar(False)), then, otherwise)
 
-    def corr(self, a: ArrowExpr, b: ArrowExpr, *, method: CorrelationMethod) -> ArrowExpr:
+    def corr(
+        self, a: ArrowExpr, b: ArrowExpr, *, method: CorrelationMethod
+    ) -> ArrowExpr:
```
MarcoGorelli: can we revert this (and any other cosmetic-only changes)?

src/narwhals/_polars/namespace.py
MarcoGorelli: ruff-format is passing on main so it wouldn't have been that - you can revert this part anyway, the checks need to pass for the pr to be mergeable

src/narwhals/_arrow/namespace.py
```diff
@@ -139,7 +139,9 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
 
     def min_horizontal(self, *exprs: ArrowExpr) -> ArrowExpr:
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
-            init_series, *series = tuple(chain.from_iterable(expr(df) for expr in exprs))
+            init_series, *series = tuple(
+                chain.from_iterable(expr(df) for expr in exprs)
+            )
```
MarcoGorelli: same

src/narwhals/_arrow/namespace.py
```diff
@@ -156,7 +158,9 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
 
     def max_horizontal(self, *exprs: ArrowExpr) -> ArrowExpr:
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
-            init_series, *series = tuple(chain.from_iterable(expr(df) for expr in exprs))
+            init_series, *series = tuple(
+                chain.from_iterable(expr(df) for expr in exprs)
+            )
```
MarcoGorelli: same

src/narwhals/_arrow/namespace.py
```diff
@@ -229,7 +233,9 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
     def coalesce(self, *exprs: ArrowExpr) -> ArrowExpr:
         def func(df: ArrowDataFrame) -> list[ArrowSeries]:
             align = self._series._align_full_broadcast
-            init_series, *series = align(*chain.from_iterable(expr(df) for expr in exprs))
+            init_series, *series = align(
+                *chain.from_iterable(expr(df) for expr in exprs)
+            )
```
MarcoGorelli: same

src/narwhals/_arrow/utils.py
```diff
@@ -91,6 +91,19 @@ def is_array_or_scalar(obj: Any) -> TypeIs[ArrayOrScalar]:
...
+    # This works by using concat_arrays to vertically stack the arrays.
+    # Then we use take to grab the data by index and to horizontally
+    # pack the arrays.
+    n = len(arrays[0])
+    num_cols = len(arrays)
+    flat = pa.concat_arrays(arrays)
+    indices = [j * n + i for i in range(n) for j in range(num_cols)]
+    interleaved = pc.take(flat, pa.array(indices))
+    offsets = pa.array(range(0, n * num_cols + 1, num_cols), type=pa.int32())
+    return pa.chunked_array([pa.ListArray.from_arrays(offsets, interleaved)])
```
MarcoGorelli: nice, clever! out of interest, where does this solution come from?

MarcoGorelli (changes requested):
thanks for updating!

some comments

can you also make sure some tests include null values?

src/narwhals/_arrow/namespace.py
CangyuanLi: so strange, ruff-format is passing on my end so I'm not sure what's going on

image

but will work on reverting all of the cosmetic changes, sorry for the extra noise!

src/narwhals/_arrow/utils.py
CangyuanLi: I found concat_arrays, realized it didn't do what I wanted, then I found ListArray.from_arrays, and found it confusing, and then I asked Claude to mash them together.

CangyuanLi:
i think I fixed all the formatting errors, as well as a typing error. for the stable api tests I followed "from_dict" excluded list from the tests. I also added the nulls in the test data. With the default backend for pandas, something like [None, 5] gets converted to a float, so I only tested pyarrow-backed pandas. 

Sorry again for the formatting noise and thanks for your patience!

MarcoGorelli:
thanks for updating! there may be some coverage issue, check if all lines are covered (or use `# pragma: no cover` if necessary), and if you fetch and merge upstream/main then the unrelated ci issues should be fixed

CangyuanLi:
> thanks for updating! there may be some coverage issue, check if all lines are covered (or use `# pragma: no cover` if necessary), and if you fetch and merge upstream/main then the unrelated ci issues should be fixed

the coverage issue should be fixed! I used no cover for the branch where polars is > 2.0 since it is not out yet

CangyuanLi:
Another try to fix CI errors--old polars versions now use pl.count() instead of pl.len(). I also removed what looked to be an extraneous type ignore

MarcoGorelli (approved):
thanks @CangyuanLi ! `pl.list` has been merged in Polars, so I think we could/should go ahead with this

seriously awesome feature here



I just pushed a commit to make `build_list_array` pyarrow-native without using a python for-loop

PR #3697, opened 2026-06-20
feat: Add `maintain_order` argument to `{Expr, Series}.list.unique`

### LINKED ISSUE #3649: feat: Add `maintain_order` argument in `{Expr, Series}.list.unique`
Drive by [CI failure](https://github.com/narwhals-dev/narwhals/actions/runs/26372398962/job/77626770171?pr=3648)

Shaan-alpha:
# Description

Adds a `maintain_order` keyword argument to `Expr.list.unique` and `Series.list.unique`, matching the Polars API.

- **Polars**: forwarded natively (Polars >= 0.20.4, the minimum supported version, already accepts `maintain_order` on `list.unique`).
- **Other backends** (DuckDB, Spark-like, Ibis): their distinct functions don't guarantee the order of the unique values, so they raise `NotImplementedError` when `maintain_order=True`. The default (`maintain_order=False`) behaviour is unchanged. This mirrors how the project already handles partial support elsewhere (e.g. Ibis raising for descending `list.sort`).
- pandas-like / PyArrow already `not_implemented` `list.unique`, so they are unaffected.

Happy to extend native order-preserving support to the SQL backends in a follow-up if you'd prefer that over raising.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3649

## Checklist

- [x] Tests added
- [x] Documented the changes


Shaan-alpha:
Hi @FBruzzesi, would you have time to review this? It adds a maintain_order argument to list.unique with implementations for the DuckDB, Ibis, and PySpark backends plus tests, close in spirit to the recent Expr.list.median work you merged. As a first-time contributor I believe CI also needs a maintainer to approve the workflow run. Happy to iterate on naming or scope if you would prefer a different API. Thanks!

src/narwhals/_spark_like/expr_list.py
```diff
@@ -19,7 +19,10 @@ class SparkLikeExprListNamespace(
     def len(self) -> SparkLikeExpr:
         return self.compliant._with_elementwise(self.compliant._F.array_size)
 
-    def unique(self) -> SparkLikeExpr:
+    def unique(self, *, maintain_order: bool) -> SparkLikeExpr:
+        if maintain_order:
+            msg = "`maintain_order=True` is not supported for the Spark-like backend."
+            raise NotImplementedError(msg)
```
FBruzzesi, line 25: While both duckdb and ibis explicitly document that order is not preserved, pyspark doesn't document it. I tried to look up for some resources and the only reference I found is the following:

> it's important to note that array_distinct preserves the order of elements in the resulting array column. The first occurrence of each unique element is retained, and subsequent occurrences are removed.

Source: [Spark Reference - Introduction to the `array_distinct` function](https://www.sparkreference.com/reference/array_distinct/#:~:text=It's%20important%20to%20note%20that,the%20array_distinct%20function%20in%20PySpark)

Could you do some more digging on this case?

FBruzzesi (commented):
Thanks for the contribution @Shaan-alpha - there are a couple of CI jobs failing for now.

I also left a comment regarding the pyspark case which I would like to better investigate before declaring it not supported.

Shaan-alpha:
Thanks for the feedback @FBruzzesi!

I looked into Spark's behavior: Apache Spark's `array_distinct` function preserves element insertion order in the array (the first occurrence of each unique element is retained, and subsequent occurrences are dropped).

I've updated `SparkLikeExprListNamespace.unique` to support `maintain_order=True` via `array_distinct` without raising `NotImplementedError`, updated the test suite, merged the latest `main`, and ensured all linting checks pass.

src/narwhals/_spark_like/expr_list.py
Shaan-alpha: I looked into Spark's `array_distinct` function: in Apache Spark (PySpark SQL), `array_distinct` preserves element order (keeping the first occurrence of each unique element and removing subsequent duplicates).

I've updated `SparkLikeExprListNamespace.unique` to support `maintain_order=True` via `array_distinct` without raising `NotImplementedError`, updated the tests, merged `main`, and verified all linting/test checks pass cleanly.

src/narwhals/_compliant/any_namespace.py
FBruzzesi, line 1: I will comment it here, as I am not sure there is a better place.

You will need to change also `EagerExprListNamespace.unique` in  `narwhals._complaint.expr.py` to make the signature compliant:

<details><summary> diff </summary>

```diff
diff --git a/src/narwhals/_compliant/expr.py b/src/narwhals/_compliant/expr.py
index 362b82d85..0fd0ced0f 100644
--- a/src/narwhals/_compliant/expr.py
+++ b/src/narwhals/_compliant/expr.py
@@ -1055,8 +1055,10 @@ class EagerExprListNamespace(
     def len(self) -> EagerExprT:
         return self.compliant._reuse_series_namespace("list", "len")
 
-    def unique(self) -> EagerExprT:
-        return self.compliant._reuse_series_namespace("list", "unique")
+    def unique(self, *, maintain_order: bool) -> EagerExprT:
+        return self.compliant._reuse_series_namespace(
+            "list", "unique", maintain_order=maintain_order
+        )
 
     def contains(self, item: NonNestedLiteral) -> EagerExprT:
         return self.compliant._reuse_series_namespace("list", "contains", item=item)
```

</details>

---

You can spot this by running `make typing`

src/narwhals/_spark_like/expr_list.py
```diff
@@ -19,7 +19,7 @@ class SparkLikeExprListNamespace(
     def len(self) -> SparkLikeExpr:
         return self.compliant._with_elementwise(self.compliant._F.array_size)
 
-    def unique(self) -> SparkLikeExpr:
+    def unique(self, *, maintain_order: bool) -> SparkLikeExpr:
         return self.compliant._with_elementwise(self.compliant._F.array_distinct)
```
FBruzzesi, line 27:
```suggestion
    def unique(self, *, maintain_order: bool) -> SparkLikeExpr:
	    # NOTE: `maintain_order` is unused as `array_distinct` always maintains the order
		return self.compliant._with_elementwise(self.compliant._F.array_distinct)
```

src/narwhals/expr_list.py
```diff
@@ -44,10 +44,15 @@ def len(self) -> ExprT:
...
+    def unique(self, *, maintain_order: bool = False) -> ExprT:
         """Get the unique/distinct values in the list.
 
-        Null values are included in the result. The order of unique values is not guaranteed.
+        Null values are included in the result.
+
+        Arguments:
+            maintain_order: Keep the same order as the input data. Setting this to
+                `True` is only supported by the Polars backend; for the other
+                backends the order of the unique values is not guaranteed.
```
FBruzzesi: > is only supported by the Polars backend

this is no longer true. I think it's ok to remove it from here and keep in a `Notes:`/`Warning:` that `maintain_order` is not supported by all backends.

Same for the Series case

FBruzzesi (commented):
Thanks for addressing it @Shaan-alpha - I found a couple of (tiny) issues, but I think we are very close to merging it :)

src/narwhals/_compliant/any_namespace.py
Shaan-alpha: Done — updated `EagerExprListNamespace.unique` in `_compliant/expr.py` to take `*, maintain_order: bool` and forward it through `_reuse_series_namespace`. `make typing` (pyright, mypy, pyrefly) is now clean.

src/narwhals/_spark_like/expr_list.py
Shaan-alpha: Applied, thanks — added the `# NOTE: maintain_order is unused as array_distinct always maintains the order` comment above the return.

src/narwhals/expr_list.py
Shaan-alpha: Done for both `Expr.list.unique` and `Series.list.unique` — dropped the "only supported by the Polars backend" line and moved the caveat into a `Notes:` section, noting that backends whose native distinct does not preserve order (e.g. DuckDB, Ibis) raise `NotImplementedError`.

Shaan-alpha:
Thanks for the review @FBruzzesi! Pushed a commit addressing all three points:

- Made `EagerExprListNamespace.unique` in `_compliant/expr.py` forward `maintain_order` so the signature matches the `ListNamespace` ABC (`make typing` is clean now).
- Added the `# NOTE: maintain_order is unused …` comment on the Spark-like `unique`.
- Removed the Polars-only wording from the `Expr.list.unique` / `Series.list.unique` docstrings and moved the backend caveat into a `Notes:` section.

Verified locally: `pyright`, `mypy`, `pyrefly`, `ruff` (check + format), `darglint`, the `list/unique` test suite (polars/pyspark/sqlframe pass, duckdb/ibis xfail as expected), and the doctests all pass. Ready for another look.

FBruzzesi:
Thanks @Shaan-alpha - not sure what's going on in CI. Will come back to you once we figure it out!

FBruzzesi (approved):
Thanks for your contribution (and patience) @Shaan-alpha 🔥

PR #3709, opened 2026-06-25
feat: Add `empty_as_null` and `keep_nulls` options to `explode`

FBruzzesi:
# Description

In light of polars deprecation warnings showing up in our CI, this PR expands the `explode` signature to match polars' one. Adapted to all backends and backported for polars (safe for 1 column explode, let polars do polars internal on multiple columns).

---

Test cases were Claude assisted, trying to cover all four flag combos for single and multi columns including null-vs-empty same-index case, and shape-mismatch errors.

---

Merging this PR would make a lot of the jobs in #3711 happy

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


FBruzzesi:
CI failures are related to a change in polars error message for `get_column`

FBruzzesi:
Regarding the pyspark-connect, I guess we can lower coverage to 90%? (it's 94 now, target is 95)

MarcoGorelli (approved):
thanks @FBruzzesi !

> Regarding the pyspark-connect, I guess we can lower coverage to 90%? (it's 94 now, target is 95)

yup sure

dangotbanned:
I ported a bunch of the tests from `polars` for this that you can steal 😉 

- https://github.com/narwhals-dev/narwhals/blob/c57e72c801b5817dc9d20262029fc6c6143c31b2/tests/plan/explode_test.py
- #3347

dangotbanned:
I've marked this as high priority (https://github.com/narwhals-dev/narwhals/pull/3709#event-27279266983) because of the CI fix.

I obviously agree with adding the flags 😅 (#3347).
But would have suggested just passing `empty_as_null=True, keep_nulls=True` *now* to fix CI - and adding the feature as a follow-up.

I haven't reviewed the changes in each backend, but if @MarcoGorelli has and is happy - then maybe we just merge?

MarcoGorelli:
that's a good point thanks! sure, let's do that, that'll make this easier to revert if necessary

PR #3712, opened 2026-06-25
chore: Address various CI failures

FBruzzesi:
# Description

Fix all the issues from #3711

* polars `get_column` error message
* modin numpy deprecation warning
* pyright
* polars explode default changes warning


tests/frame/get_column_test.py
```diff
@@ -15,7 +15,7 @@ def test_get_column(constructor_eager: ConstructorEager) -> None:
         (KeyError, TypeError),
         match=(
             r"Expected str|'int' object cannot be (converted|cast)|0"
-            r"|argument 'name': 'int' object is not an instance of 'str'"
+            r"|'int' object is not an instance of 'str'"
```
dangotbanned: So I was going to suggest changing the pattern to be:

```py
    with pytest.raises((KeyError, TypeError), match=r"(str|int).+(str|int)"):
```

But that misses the `"|0"` part 😔 

Could we split out the pandas part and make this test either of these?

1. what happens for a missing column?
    i. we have no coverage for this beyond pandas 😳
2. what happens for an invalid column type?

> [!NOTE]
> I know this is an increase in scope, but the test itself keeps breaking


## Related
- #479
- #517
- #3397
- #3553
- (c7930c623d36b384f0d1f926c3d3560616d38c43)

src/narwhals/_arrow/series.py
```diff
@@ -633,7 +633,7 @@ def sample(
 
         rng = np.random.default_rng(seed=seed)
         idx = np.arange(num_rows)
-        mask = rng.choice(idx, size=n, replace=with_replacement)
+        mask = cast("_1DArray", rng.choice(idx, size=n, replace=with_replacement))
```
dangotbanned: I looked into what caused this, because `np.arange` is being inferred correctly ...

image


Our highest pin in `uv.lock` seems to be (https://github.com/numpy/numpy/releases/tag/v2.4.6).
Which gives us [these overloads](https://github.com/numpy/numpy/blob/b832a09cf2a169c833dd2371e7c07aa00b293242/numpy/random/_generator.pyi#L478-L519) and a TODO.

So I checked the latest release (https://github.com/numpy/numpy/releases/tag/v2.5.0) which is [a bit different](https://github.com/numpy/numpy/blob/610766be461774dd652558ad6691802c13409b7a/numpy/random/_generator.pyi#L690-L756) - but I'm not sure they'd help us

I wonder if I can summon @jorenham and catch us a `_Array1D` overload for `Generator.choice`? 😉

dangotbanned: Alternatively, we could avoid passing `np.ndarray`s directly and use this

```diff
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -632,9 +632,8 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
             n = int(num_rows * fraction)
 
         rng = np.random.default_rng(seed=seed)
-        idx = np.arange(num_rows)
-        mask = cast("_1DArray", rng.choice(idx, size=n, replace=with_replacement))
-        return self._with_native(self.native.take(mask))
+        indices = rng.choice(np.arange(num_rows), size=n, replace=with_replacement)
+        return self._with_native(self.native.take(pa.array(indices, type=pa.uint64())))
```

src/narwhals/_arrow/dataframe.py
```diff
@@ -811,7 +811,7 @@ def sample(
             n = int(num_rows * fraction)
         rng = np.random.default_rng(seed=seed)
         idx = np.arange(num_rows)
-        mask = rng.choice(idx, size=n, replace=with_replacement)
+        mask = cast("_1DArray", rng.choice(idx, size=n, replace=with_replacement))
```
dangotbanned: (Same ~~spiralling~~ suggestion as `ArrowSeries.sample` 😅)

dangotbanned (approved):
Thanks @FBruzzesi for keeping things working 😍

src/narwhals/_arrow/series.py
FBruzzesi: Thanks for the suggestion @dangotbanned - addressed in https://github.com/narwhals-dev/narwhals/pull/3712/commits/396229838d89051bef2a31c57330189b8c34fb3e

tests/frame/get_column_test.py
FBruzzesi: Split up the test cases in https://github.com/narwhals-dev/narwhals/pull/3712/commits/cb5d2e876dd6801f2c8b3f880de6e85395db4cc1

PR #3558, opened 2026-04-19
chore: Remove `level` from main and v2 classes init

FBruzzesi:
# Description

Not really user facing but I had it started some time ago, so why not finish it 😇

## What type of PR is this? (check all applicable)

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #2932
- Parent of #3721

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes

PR #3721, opened 2026-06-27
chore: Isolate interchange-level support

dangotbanned:
# Description
This PR started as an experiment and snowballed into an *informed refactor*™.

> [!NOTE]
> **TL;DR**: In (https://github.com/narwhals-dev/narwhals/pull/3558#discussion_r3486733917) I had no clue what interchange meant
> It now has clear boundaries, based on what people actually use
> All related tests can be run with `uv run pytest tests/interchange`

## Related issues

- #3558
- https://github.com/narwhals-dev/narwhals/pull/3558#discussion_r3477878582
- https://github.com/pola-rs/polars/pull/28152

# Experiment
To find out what *interchange* really means:
- I removed it
- Looked at what broke downstream
- Gradually added things until both our tests and downstream were green

## Break interchange (e745e375c0711d2fc153dbd957c2066725ef6522)
- Removed it entirely from main, v2
- Raising when requested for v1
- xfail 21 tests

## Identify downstream breakage
Found out we have coverage from 3 downstream packages for `eager_or_interchange_only=True`

- [Altair](https://github.com/narwhals-dev/narwhals/actions/runs/28296653471/job/83837655934?pr=3721)
- [Marimo](https://github.com/narwhals-dev/narwhals/actions/runs/28296653471/job/83837655928?pr=3721)
- [Plotly](https://github.com/narwhals-dev/narwhals/actions/runs/28296653471/job/83837655979?pr=3721)

### Altair
- `v1.from_native(eager_or_interchange_only=True)`
  - [`altair.Chart.to_dict`](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/vegalite/v6/api.py#L2107)
  - [`altair.utils.data.is_data_type`](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/data.py#L69-L73)
  - [`altair.utils.core.to_eager_narwhals_dataframe`](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/core.py#L515)
  - [`altair.utils.core.parse_shorthand`](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/core.py#L679)
- `v1.get_level`
  	- [`altair.utils.core.to_eager_narwhals_dataframe`](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/core.py#L516-L522) 
- [defers to pyarrow](https://github.com/vega/altair/blob/48b388f140c79d29056d6ea56e519b27e2ed8838/altair/utils/data.py#L419-L436)
	- They're using interchange to convert to `pa.Table` and then using narwhals as normal

### Marimo
- There's one [comment mentioning interchange](https://github.com/marimo-team/marimo/blob/c3e14e30082c397d2a560e794cfb718f461c169a/marimo/_data/preview_column.py#L238-L239)
- Didn't find any direct usage
- Everything seems to be indirectly via Altair

### Plotly
- `plotly.express._core.build_dataframe`
  - [`v1.from_native(eager_or_interchange_only=True)`](https://github.com/plotly/plotly.py/blob/5cdb6064a58ccff2a92b71f9710d83662dbbb7a5/plotly/express/_core.py#L1543-L1556)
  - [`v1.get_level`](https://github.com/plotly/plotly.py/blob/5cdb6064a58ccff2a92b71f9710d83662dbbb7a5/plotly/express/_core.py#L1553)
  - [defers to pyarrow](https://github.com/plotly/plotly.py/blob/5cdb6064a58ccff2a92b71f9710d83662dbbb7a5/plotly/express/_core.py#L1693-L1720)
  	- (**Roughly the same as Altair**)

### Summary

- Altair and Plotly could benefit from something like `from_arrow(DataFrameLike, backend="pyarrow")`
- Interchange is simply a stepping stone to pyarrow

## Re-introduce interchange
- [x] With more restrictions (https://github.com/narwhals-dev/narwhals/pull/3721/commits/4940eb33b7b9acad2f7c6dadfda4603c5bcbccc6), (https://github.com/narwhals-dev/narwhals/pull/3721/commits/3cfb767401c2c6c768ad01e1d47f00bab380f0d5)
  - Interestingly, this fixed both Altair and Plotly - but not Marimo
  - So maybe Marimo is doing something sneaky?
- [x] Add the things Marimo needs (see [failures](https://github.com/narwhals-dev/narwhals/actions/runs/28305031920/job/83859654934?pr=3721))
  - [x] Ibis (https://github.com/narwhals-dev/narwhals/pull/3721/commits/902796adb2d5ca64ecc67c611af45ce8962716cd)
  - [x] DuckDB
- [x] Isolate interchange tests (3ceb3e050c700af3b0cafaa4f01f0bcc2627aa2f), (ecfc247969a2bc7a5468020e5bd6bd127e4ecbba), (545e9a301287f9e80bd08e64f0d7d4cdf5b385ad), (fc21485128a5bb0695f073c4d9fd120e6f247779), (af940e6b1af0dc45f190d6cc3d3048947f766df6), (d7c5b450f1e597d6b94d4fbcb6743091d3223205), (f4f7d7232c36950e9ce962935b79bf6fcf243fc7)
- [x] Review what extra stuff we have that isn't used downstream
  - [x] Compliant-level methods that downstream do not use are no longer expected to work (edc6feaf88b13532120044cf7eac2cbe0b573f04)
  - [x] Do a github search dive to find other users of interchange
    - `pyfixest`
        - [`DataFrame.to_pandas`](https://github.com/py-econometrics/pyfixest/blob/27dfbbadfe29a684056bc951c4868131672ba62e/pyfixest/utils/dev_utils.py#L11)
        - They're [only using narwhals to convert to `pandas`](https://github.com/search?q=repo%3Apy-econometrics%2Fpyfixest%20%22narwhals%22&type=code) 😔
    - `vegafusion`
        - [`v1.get_level() == "full"`](https://github.com/vega/vegafusion/blob/7a40b506fb16be6e9423723d3ffa5539a9cccd15/vegafusion-python/vegafusion/runtime.py#L54)
        - Interchange is not really beyond [converting](https://github.com/vega/vegafusion/blob/7a40b506fb16be6e9423723d3ffa5539a9cccd15/vegafusion-python/vegafusion/runtime.py#L292-L394) to [specific backends](https://github.com/vega/vegafusion/blob/7a40b506fb16be6e9423723d3ffa5539a9cccd15/vegafusion-python/vegafusion/runtime.py#L584-L592)
        - **Like Altair & Plotly**, [interchange requires `pyarrow`](https://github.com/vega/vegafusion/blob/7a40b506fb16be6e9423723d3ffa5539a9cccd15/vegafusion-python/vegafusion/transformer.py#L150-L170)
  - [x] ~~Consider extending our downstream ci testing if I find someone using more than `marimo`~~ **found nothing new!**

# Possible follow-ups
Found a good point to stop, but still had a few ideas that could follow.

I'm not in any rush to do this and would be happy to lock interchange up and never look back 😄 

## Docs
Be explicit about what is supported for `ibis` and `duckdb` (and suggest using `v2` instead)

## Feat
Almost all use cases I found for interchange support would be solved by a **new** `v1` function like:

```py
def from_interchange(
    frame: DataFrameLike, *, backend: IntoBackend[EagerAllowed] = "pyarrow"
) -> nw_v1.DataFrame[pa.Table | pl.DataFrame | pd.DataFrame]: ...
```

Where the default is `"pyarrow"` because Altair, Plotly, and Vegafusion already require it for interchange support.

## Tests
- Deduplicate the test suite
    - Most `ibis` tests are 80% setup code
- Test everything against: `ibis`, `duckdb`, `pl.DataFrame.__dataframe__`, custom
    - Currently a mixed bag
    - Or remove `polars` entirely because of (https://github.com/pola-rs/polars/pull/28152)
- Separate tests for `columns`, `collect_schema`, `to_native`, `get_column`
    - They're all supported, but usually added onto some other test (or indirectly before this PR)

## CI
Following #3629, we no longer test Vegafusion in downstream ci [^1].

An alternative to reintroducing that workflow is to run Altair again, but with [Vegafusion enabled](https://altair-viz.github.io/user_guide/large_datasets.html#vegafusion-data-transformer).
That should get us some more coverage for `v1`, without needing to compile from source

[^1]: Not sure why, guessing speed?


MarcoGorelli:
thanks for looking into this

just noting that whatever solution we go with shouldn't require downstream libraries (marimo / altair / ....) to change anything on their side. i'm fairly confident that it should be possible to remove this from v2/main without affecting v1 usage at all

dangotbanned:
> just noting that whatever solution we go with shouldn't require downstream libraries (marimo / altair / ....) to change anything on their side

100% agreed @MarcoGorelli !

My plan now is to:

1. Get all downstream green (with the smallest interchange footprint possible)
    i. Altair & Plotly are all good with "just" interchange
    ii. Marimo is a WIP
2. Do a github search dive to find other users of interchange
3. Consider extending our downstream ci testing if I find someone using more than `marimo`

I'm still kinda in a fact-finding stage, but feel like I understand interchange better than yesterday 😄

src/narwhals/_interchange/dataframe.py
```diff
@@ -142,6 +142,10 @@ def schema(self) -> dict[str, DType]: ...
     def columns(self) -> list[str]:
         return list(self.column_names())
 
+    @property
+    def _native_frame(self) -> Original_co:
+        return self._df
```
dangotbanned: I added this alias temporarily so that `to_native` keeps working.

But we should probably update these guys:

https://github.com/narwhals-dev/narwhals/blob/35e16811ec683542c9468cd1f34d311829f7fb17/src/narwhals/translate.py#L111-L114

```diff
-    if isinstance(narwhals_object, BaseFrame):
-        return narwhals_object._compliant_frame._native_frame
-    if isinstance(narwhals_object, Series):
-        return narwhals_object._compliant_series.native
+    if isinstance(narwhals_object, (BaseFrame, Series)):
+        return narwhals_object._compliant.native
```

src/narwhals/_interchange/dataframe.py
```diff
@@ -75,12 +83,152 @@ def map_interchange_dtype_to_narwhals_dtype(  # noqa: C901, PLR0911, PLR0912
...
+    def __native_namespace__(self) -> ModuleType:
+        return self._implementation.to_native_namespace()
+
+    if not TYPE_CHECKING:
+
+        def __getattr__(self, name: str) -> Never:
+            raise unsupported_error(name)
+
+
+class Frame(Protocol):
```
dangotbanned: Need to remove most of this module

Didn't realise we were barely using interchange before this PR 🤦‍♂️

tests/v1_test.py
```diff
@@ -44,15 +44,28 @@
...
 
     from narwhals._typing import EagerAllowed
     from narwhals.dtypes import DType
-    from narwhals.stable.v1.typing import IntoDataFrameT
+    from narwhals.stable.v1.typing import Frame, IntoDataFrameT
     from narwhals.typing import IntoDType, _1DArray, _2DArray
     from tests.utils import Constructor, ConstructorEager
 
 
+# TODO @dangotbanned: Include duckdb after checking ci fallout for ibis
```
dangotbanned: I did that!

```suggestion
```

src/narwhals/_interchange/lazyframe.py
```diff
@@ -0,0 +1,99 @@
...
+            "__native_namespace__",
+            "_native_frame",
+            "schema",
+            "columns",
+            "collect_schema",
+        )
+    )
+
+    def __init__(self, compliant: Compliant[NativeT]) -> None:
+        self._compliant = compliant
```
dangotbanned: Fix this for old `Protocol.__init__` (https://github.com/narwhals-dev/narwhals/actions/runs/28408178674/job/84175225039?pr=3721)

dangotbanned: Forgot about (https://github.com/narwhals-dev/narwhals/pull/2294#discussion_r2014534830) 🤦‍♂️ 

Fixed: (https://github.com/narwhals-dev/narwhals/pull/3721/commits/94f2e865c7f8fe9ceeb0e2dbe46b90b16edd37b6)

dangotbanned:
If anyone can point me towards evidence of downstream use cases (that aren't in the description) that this will break **please let me know!**

I spent some time searching, but I'm open to widening the API if shown something new 🙂

dangotbanned:
> just noting that whatever solution we go with **shouldn't require** downstream libraries (marimo / altair / ....) to change anything on their side


@MarcoGorelli I'm still in agreement with you on this ...
*but* I think I've stumbled into an (existing) [^1] performance issue for `altair`.

[^1]: I'm testing this from (https://github.com/narwhals-dev/narwhals/compare/expr-ir/docs/fluff-1...expr-ir/altair-expr), which doesn't have the changes from this branch

It looks to stem from (https://github.com/vega/altair/pull/3452#pullrequestreview-2176310770), which I think was well-intentioned.

## Example
Here we have a very simple chart.

I was suprised to find that these steps *didn't* trigger `from_native`:

```py
import altair as alt
import polars as pl

df = pl.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
chart = alt.Chart(df).mark_point().encode(x="x", y="y")
```

Reading back through (https://github.com/vega/altair/pull/3452) - the reason was to avoid a copy to `pyarrow`.


This call [^2] manages to hit `from_native` **9 times** in total, with **3 different objects**:

```py
chart.to_dict()
```


[^2]: which happpens when you repr the chart

<details><summary>Show logs (polars)</summary>
<p>

image


</p>
</details> 

Now here's the same thing, but using `duckdb`:

```py
import altair as alt
import polars as pl
import duckdb

df = pl.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
chart = alt.Chart(duckdb.from_arrow(df)).mark_point().encode(x="x", y="y")
```

This call increases to **10 times** in total, with **4 different objects**:

```py
chart.to_dict()
```

<details><summary>Show logs (duckdb)</summary>
<p>

image

</p>
</details> 

## Questions
1. Do these numbers surprise you?
2. Does optimizing for `__dataframe__` make sense when we don't use it in either of these cases?


## Solution(s)?
For DuckDB there are two things that wouldn't have been available back in (https://github.com/vega/altair/pull/3452):

1. We can collect the schema (and then cache it) without loading the data
2. `__arrow_c_stream__` should ensure life begins at zero-copy 😉 

For everything else, I feel like wrapping the native object early would be the place to start?

src/narwhals/_interchange/dataframe.py
```diff
@@ -107,57 +126,61 @@ def to_pandas(self) -> pd.DataFrame:
...
+    return _should_interchange(type(obj))  # type: ignore[arg-type]
 
 
-def supports_dataframe_interchange(obj: Any) -> TypeIs[DataFrameLike]:
-    return _hasattr_static(obj, "__dataframe__")
+@lru_cache(64)
+def _should_interchange(tp_native: type[Any]) -> TypeIs[type[DataFrameLike]]:
+    if not _hasattr_static(tp_native, "__dataframe__"):
+        return (duckdb := get_duckdb()) and issubclass(tp_native, duckdb.DuckDBPyRelation)
+    has_top_level_df = (get_polars, get_pandas, get_dask_dataframe, get_modin)
```
FBruzzesi: `cudf` is missing from this list which would lead to a regression "demoting" cudf.DataFrame's to interchange level for v1.

More generally this function is an inversion of logic (from inclusion to exclusion) and seems quite hard to maintain/keep in sync, for many reasons, one of which is that we don't typically look nor touch interchange stuff anymore - what about dog-feeding `_is_native_dataframe` / `_is_<known_to_exclude>_lazyframe`?

src/narwhals/_interchange/dataframe.py
```diff
@@ -75,28 +85,37 @@ def map_interchange_dtype_to_narwhals_dtype(  # noqa: C901, PLR0911, PLR0912
     raise AssertionError(msg)
 
 
-class InterchangeFrame:
-    _version = Version.V1
+def unsupported_error(method_name: str) -> NotImplementedError:
+    msg = (
+        f"{method_name!r} is not supported for interchange-level dataframes.\n\n"
+        "Hint: you probably called `from_native` on an object which isn't fully "
+        "supported by `narwhals.stable.v1`, yet implements `__dataframe__`."
```
FBruzzesi, line 98: If this can trigger from a `DuckDBPyRelation`, then that's actually not the case: it does not implement `__dataframe__` (not sure about ibis case)

src/narwhals/_duckdb/dataframe.py
```diff
@@ -540,9 +482,6 @@ def unpivot(
 
     @requires.backend_version((1, 3))
     def with_row_index(self, name: str, order_by: Sequence[str]) -> Self:
-        if order_by is None:
-            msg = "Cannot pass `order_by` to `with_row_index` for DuckDB"
-            raise TypeError(msg)
```
FBruzzesi, line 545: How come this is the case?

FBruzzesi (commented):
Thanks @dangotbanned - I left a couple of comments, the real blocker is `_should_interchange` regression and maybe implementation.

Another doubt I have is about the introduction of two new `# pragma: no cover`:

* `_duckdb/utils.py` added `if version is Version.V1:  # pragma: no cover`
* `_duckdb/dataframe.py` added `if validate_backend_version:  # pragma: no cover`

src/narwhals/_interchange/dataframe.py
dangotbanned: > `cudf` is missing from this list which would lead to a regression "demoting" cudf.DataFrame's to interchange level for v1.

Well spotted!

So I started out thinking `cudf` supported interchange, but [removed it when I couldn't find it anywhere](https://github.com/narwhals-dev/narwhals/commit/68f8136341099c11f9c5eb198e7c116a9de6ca59#diff-78e0aa45e620281b85bc452f45d4e5d1666b8eb84435198c005551b094201787L316)

I'm [looking again]((https://github.com/search?q=repo%3Arapidsai%2Fcudf%20%22__dataframe__%22&type=code)) now and it looks like [this is the support](https://github.com/rapidsai/cudf/blob/c1063448c685761c1d2e6d68919bbb753fd17914/python/cudf/cudf/pandas/fast_slow_proxy.py#L1784)?

There is also a [mention here](https://github.com/rapidsai/cudf/blob/c1063448c685761c1d2e6d68919bbb753fd17914/ci/test_narwhals.sh#L53-L61) of narwhals and interchange:

```sh
# polars 1.40 deprecated the DataFrame interchange protocol, which narwhals
# 2.16.0 still exercises; narwhals' filterwarnings=error turns the resulting
# DeprecationWarning into a failure. These run for every constructor, so skip
# them across all runs. test_get_level also relies on the interchange protocol.
# Remove once narwhals stops using the deprecated protocol.
```

And possibly this was when it got removed:
- https://github.com/rapidsai/cudf/issues/17403

src/narwhals/_duckdb/dataframe.py
dangotbanned: Because this is now unreachable, since `v1.DataFrame.with_row_index` will always raise for duckdb/ibis.

Actually, this signature could be tightened up now too:

https://github.com/narwhals-dev/narwhals/blob/b23e39903db8e1c028960d9d40e1e0a4736c6330/src/narwhals/_compliant/dataframe.py#L281

Since `LazyFrame` doesn't accept `None`:

https://github.com/narwhals-dev/narwhals/blob/b23e39903db8e1c028960d9d40e1e0a4736c6330/src/narwhals/dataframe.py#L2571-L2573

src/narwhals/_interchange/dataframe.py
dangotbanned: This is a pretty interesting point.

What is being called *interchange-level* actually is entirely disjoint from it 😂

> `DuckDBPyRelation` it does not implement `__dataframe__`

Which was news to be me before I started this PR!

> (not sure about ibis case)

`ibis` does support it and I was using that directly in an earlier version (https://github.com/narwhals-dev/narwhals/blob/b7e533d9f2e10be02713617b3cfb556eb236da66/src/narwhals/_ibis/interchange.py)

I don't know what exactly to call this support - but I'll take any name you can think of 🙏

dangotbanned:
Thanks for the review @FBruzzesi

> Another doubt I have is about the introduction of two new `# pragma: no cover:`

I believe all of those are from [`tests/v1_test.py` diff](https://github.com/narwhals-dev/narwhals/pull/3721/files#diff-4da9e38e2e0c05c4aa26936a95b843bcc1fdd326411fb3fbe563881ab0d51516) and these commits:
- (https://github.com/narwhals-dev/narwhals/pull/3721/commits/edc6feaf88b13532120044cf7eac2cbe0b573f04)
- (https://github.com/narwhals-dev/narwhals/pull/3721/commits/395ee9da6999a18e09d09afe3c19c40f1ecfd1a9)

Essentially - I found no downstream using any of this and it has never been documented as working [^1].

[^1]: Which I have to assume is why nobody depended on/knew about it 😉 

I will add things back if needed (https://github.com/narwhals-dev/narwhals/pull/3721#issuecomment-4846436504) - but I would like to see the use case 🙂 

What we have here is what people are using, which is very similar to (https://github.com/narwhals-dev/narwhals/pull/3558#discussion_r3477878582), but also `to_native` is used by Marimo - so there are a couple extras

https://github.com/narwhals-dev/narwhals/blob/4bbdf114e109163077ab22bcab1c622f93dc3f23/src/narwhals/_interchange/lazyframe.py#L33-L38
https://github.com/narwhals-dev/narwhals/blob/4bbdf114e109163077ab22bcab1c622f93dc3f23/src/narwhals/_interchange/lazyframe.py#L42-L48
https://github.com/narwhals-dev/narwhals/blob/4bbdf114e109163077ab22bcab1c622f93dc3f23/src/narwhals/_interchange/lazyframe.py#L62-L63

src/narwhals/_interchange/dataframe.py
dangotbanned: > More generally this function is an inversion of logic (from inclusion to exclusion) and seems quite hard to maintain/keep in sync, for many reasons, one of which is that we don't typically look nor touch interchange stuff anymore

I'll do my best to walk through this tomorrow and explain how I landed here - there were a couple of concerns I was trying to balance

In terms of keeping in sync, IIUC this logic is frozen - which is why I opted to make it the case to handle first and keep everything confined to v1/"interchange"

FBruzzesi: Yes CuDF supports (or at least supported) interchange and is deprecating/ has deprecated it, but some versions we support definitely can enter such path.

dangotbanned: @FBruzzesi hopefully these guys are helpful (https://github.com/narwhals-dev/narwhals/pull/3721/files/4bbdf114e109163077ab22bcab1c622f93dc3f23..955a677a1ebd5960686a8f13ebd90b91fb91c432)

Another thing I could speak on, but didn't address in those is: *Why is this not using `is_*_dataframe`?* or *why is this checking on the type?*

They're kinda the same question really 😅

I wanted to remove the concept of interchange from `translate.py` and `v2`, where it still lingered since (#2814).

So my choices were:
1. Repeat all of `_from_native_impl` in `v1`, with the tweaks that allow "interchange"
2. Handle "interchange" in `v1`, then defer to main

I picked option **2**, since it requires less maintenance.

I also saw a way that option **2** could use caching for the very complicated checking 🙂. 
*But*, that needed to switch to using the type vs the instance and so I couldn't reuse the existing guards

src/narwhals/_duckdb/utils.py
```diff
@@ -273,7 +273,7 @@ def narwhals_to_native_dtype(  # noqa: PLR0912, C901
     if duckdb_type := NW_TO_DUCKDB_DTYPES.get(base_type):
         return duckdb_type
     if isinstance_or_issubclass(dtype, dtypes.Enum):
-        if version is Version.V1:
+        if version is Version.V1:  # pragma: no cover
```
dangotbanned, line 276: Probably should remove this branch (same reason as https://github.com/narwhals-dev/narwhals/pull/3721#discussion_r3516077780)

src/narwhals/_interchange/dataframe.py
```diff
@@ -170,17 +176,42 @@ def native(self) -> Any:
...
+    We *could* use them with interchange-level support - but we have a better option at home.
+
+    Sadly this type is [not yet](https://www.youtube.com/watch?v=xUsPD7iwETY) spellable:
+
+        (DataFrameLike & (pl.DataFrame | pd.DataFrame | pa.Table | dd.DataFrame | mpd.DataFrame | cudf.DataFrame)
+        #              |             union
+        #         intersection
+
+    We gathers all the types (that are already imported) which fit the the right-hand-side,
+    as we need to refine from the set produced by checking `__dataframe__`.
```
FBruzzesi, line 195: Best description I read in a while 😂

dangotbanned: aha thanks - part of this was also trying to make the `dask` part more visible too.


... and you highlighted a typo!

```suggestion
    We *could* use them with interchange-level support - but we have a better option at home.

    Sadly this type is [not yet](https://www.youtube.com/watch?v=xUsPD7iwETY) spellable:

        (DataFrameLike & (pl.DataFrame | pd.DataFrame | pa.Table | dd.DataFrame | mpd.DataFrame | cudf.DataFrame)
        #              |             union
        #         intersection

    We gather all the types (that are already imported) which fit the the right-hand-side,
    as we need to refine from the set produced by checking `__dataframe__`.
```

PR #3724, opened 2026-06-28
fix: Lift `sample` arguments validation, bump pyright and numpy deps

FBruzzesi:
# Description

Lift `n`/`fraction` resolution and validation for `DataFrame.sample` and `Series.sample` out of the individual backends into a shared `_resolve_sample_size` helper, so error messages are consistent across pandas, PyArrow and Polars.

Backend `sample` methods now take a single resolved `n: int`. Also bumps the pyright and numpy dev dependencies.

Related to #3711 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


tests/v1_test.py
```diff
@@ -899,6 +899,15 @@ def test_expr_sample(constructor_eager: ConstructorEager) -> None:
     assert result_expr == expected_expr
 
 
+def test_expr_sample_fraction(constructor_eager: ConstructorEager) -> None:
```
dangotbanned, line 902: I forgot about the other validation stuff, nice one that you didn't 😎

I'm curious about what `polars.Expr.sample` does (compared to our `v1`).
Does it also raise the same errors?

FBruzzesi: These are the polars exceptions on latest version:

```py
import polars as pl

frame = pl.DataFrame({"a": [1, 2, 3]})

for n in (-1, 4):
    try:
        frame.select(pl.col("a").sample(n=n))
    except Exception as exc:
        print(n, type(exc), str(exc), sep="\n")

try:
    frame.select(pl.col("a").sample(n=2, fraction=.5))
except Exception as exc:
    print(type(exc), str(exc), sep="\n")
```

```terminal
-1
<class 'polars.exceptions.InvalidOperationError'>
conversion from `i32` to `u32` failed in column 'literal' for 1 out of 1 values: [-1]

This error occurred in the following expression:
        col("a").sample([dyn int: -1])

4
<class 'polars.exceptions.ShapeError'>
cannot take a larger sample than the total population when `with_replacement=false`

This error occurred in the following expression:
        col("a").sample([dyn int: 4])

<class 'ValueError'>
cannot specify both `n` and `fraction`
```

dangotbanned: Perfect, thanks @FBruzzesi

tests/expr_and_series/sample_test.py
```diff
@@ -1,11 +1,20 @@
...
 
 import narwhals as nw
 from narwhals.exceptions import InvalidOperationError, ShapeError
 from tests.utils import ConstructorEager, assert_equal_data
 
+if TYPE_CHECKING:
+
+    class SampleKwargs(TypedDict, total=False):
+        n: int | None
+        fraction: float | None
```
dangotbanned, line 15: oh hello 😄

dangotbanned: A trick I quite like is making it a runtime `TypedDict`, and using a single `parametrize` param:

https://github.com/narwhals-dev/narwhals/blob/c57e72c801b5817dc9d20262029fc6c6143c31b2/tests/plan/ewm_mean_test.py#L14-L38

With your `SampleKwargs` it could be like:

```py
SampleKwargs(n=2, fraction= 0.5)
SampleKwargs(n=10)
```


But idk maybe more beneficial when there are more than 2 cases (and 2 parameters) 😂

FBruzzesi: This was just to avoid having `params: dict[str, Any]` nothing serious 🤣

MarcoGorelli (approved):
thanks!

PR #3732, opened 2026-06-30
fix: Explicitely pass `orient="row"` in `pl.from_numpy`

### LINKED ISSUE #3716: bug: `from_numpy` silently transposes a square array on the polars backend
`narwhals.from_numpy` documents row orientation as the only supported behavior, but on the polars backend it does not enforce it: it calls `pl.from_numpy(data, schema)` without `orient="row"`.

For a **square**, **Fortran-contiguous** array, polars then infers column orientation and the resulting frame is the **transpose** of the input.
This is easy to hit in practice because `polars.DataFrame.to_numpy()` returns a Fortran-contiguous array, so a `to_numpy()` -> `from_numpy(..., schema=cols)` round-trip on a square frame silently transposes the data.

<details><summary>Repro</summary>
<p>

```py
import narwhals as nw
import numpy as np
import polars as pl

arr = pl.DataFrame(np.arange(9, dtype=float).reshape(3, 3)).to_numpy()

out_with_schema = nw.from_numpy(arr, schema=["a", "b", "c"], backend=pl).to_numpy()
out_wo_schema = nw.from_numpy(arr, backend=pl).to_numpy()

print(arr, out_with_schema, out_wo_schema, sep="\n\n")
```

```terminal
[[0. 1. 2.]
 [3. 4. 5.]
 [6. 7. 8.]]

[[0. 3. 6.]
 [1. 4. 7.]
 [2. 5. 8.]]

[[0. 1. 2.]
 [3. 4. 5.]
 [6. 7. 8.]]
```

</p>
</details> 

---

Proposed fix is to pin

```python
return cls.from_native(pl.from_numpy(data, pl_schema, orient="row"), context=context)
```

FBruzzesi:
# Description

Fixes polars roundtrip `from_numpy(obj.to_numpy(), schema=...)` for square case

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3716


tests/frame/from_numpy_test.py
```diff
@@ -56,3 +56,26 @@ def test_from_numpy_non_eager() -> None:
...
+
+@pytest.mark.parametrize("schema", [None, ["x", "y", "z"]])
+def test_from_numpy_square(eager_backend: EagerAllowed, schema: list[str] | None) -> None:
+    # See https://github.com/narwhals-dev/narwhals/issues/3716:
+    # Fortran-contiguous square array (as returned by polars' `to_numpy`) used to be
+    # silently transposed on the polars backend, since polars infers column
+    # orientation when the schema length matches both axes.
+    rows = [[0, 1, 2], [3, 4, 5], [6, 7, 8]]
+    columns = [f"column_{i}" for i in range(3)] if schema is None else schema
+    expected = {name: values for name, *values in zip(columns, *rows, strict=True)}
```
FBruzzesi, line 69: Is this too mysterious? Goal is to transpose row values

tests/namespace_test.py
```diff
@@ -128,7 +128,7 @@ def test_namespace_from_numpy_polars() -> None:
     import polars as pl
 
     arr: _2DArray = cast("_2DArray", np.array([[5, 2, 0, 1], [1, 4, 7, 8], [1, 2, 3, 9]]))
-    columns = "a", "b", "c"
+    columns = ["a", "b", "c", "d"]
```
FBruzzesi, line 131: This was wrong to begin with: `arr` has 3 rows, 4 columns. So polars was inferring non-row oriented. Now, passing `orient="row"` make this test case fail

tests/frame/from_numpy_test.py
MarcoGorelli: in general i'd strongly prefer to keep `expected` as simple as possible and without logic

ok to then also check against something like `{name: values for name, *values in zip(columns, *rows, strict=True)}`

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #3740, opened 2026-07-01
ci: speed up pyspark connect

EdAbati:
# Description

Pyspark connect was downloading the full Spark tarball from archive.apache.org in CI which [made it very slow](https://github.com/narwhals-dev/narwhals/actions/runs/28507554213/job/84499603308?pr=3737)

In this PR we move to using the `apache/spark` official image directly, so we don't need to download anything extra. The small downside is that we now have 2 places that need to stay in sync when we bump the Spark version (`env` isn't available in `container.image`, so it can't just reference `SPARK_VERSION`).

An alternative is to revert this branch to the [first commit](https://github.com/narwhals-dev/narwhals/pull/3740/commits/6241471b83b474586b63544ac9c12311c5ee75b2). It's a workaround where we install spark from pip and download the connect start/stop scripts from GitHub (they're not shipped with the pip package). The potential problem is that it relies on an undocumented method that could break in the future.

Let me know what you prefer :)

# Performance

| Run                       | Time        |
| ------------------------- | ----------- |
| [**Average (12 months)**] | 9m 41s [^1] |
| [**This PR**]             | **2m 50s**  |

[**Average (12 months)**]: https://github.com/narwhals-dev/narwhals/actions/metrics/performance?dateRangeType=DATE_RANGE_TYPE_CUSTOM&tab=jobs&filters=workflow_file_name%3Apytest-pyspark.yml+job_name%3A%22pytest-pyspark-connect-constructor+%283.11%2C+ubuntu-latest%29%22&range=1751328000000-1759881600000
[**This PR**]: https://github.com/narwhals-dev/narwhals/actions/runs/28573417785/job/84716194861?pr=3740

[^1]: But frequently >45m

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other


.github/workflows/pytest-pyspark.yml
```diff
@@ -72,8 +73,14 @@ jobs:
       SPARK_VERSION: 4.1.0
       SPARK_PORT: 15002
       SPARK_CONNECT: true
-      SPARK_HOME: ""
     runs-on: ${{ matrix.os }}
+
+    container:
+      # env context isn't available here (https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability), so this must match SPARK_VERSION
+      image: apache/spark:4.1.0
```
FBruzzesi, line 80: Is `env.SPARK_VERSION` even used anymore?

FBruzzesi (approved):
Thanks for taking care of this @EdAbati

.github/workflows/pytest-pyspark.yml
EdAbati: Yeah in the 'install pyspark' step 😕

MarcoGorelli (approved):
sure, if this works, seems like a net positive

thanks so much @EdAbati !

dangotbanned (approved):
HUGE HUGE HUGE-ly appreciated @EdAbati!

~~Would it be possible to add a comparison of the runtime(s) to the PR description?~~ **Update: I did it**
I tried doing this yesterday, but was having trouble finding a reliable time for the old workflow 😔
(I know from experience that the current time is a waaaaaaaaay shorter, but everyone should know how much of a difference this makes 😍)

hungry narwhal

PR #3749, opened 2026-07-03
feat: stabilize dtype conversion across pandas, pyarrow, and modin test constructors

camriddell:
Happy to discuss this. Not converting dtypes is going to be a good thing to test as part of #3717

`pandas.{DataFrame,Series}.convert_dtypes(…)` will automatically downcast whole number floats (e.g. 1.0, 2.0, 3.0) to integer datatypes. The only option within that method to avoid this behavior is to provide `convert_integer=False`, which both: retains the whole number floats columns as a float dtype, but also then ignores any actual integer columns and thus converts them `int64`. 

This PR performs each conversion column-by-column and eagerly check for whole number float to control the `convert_integer` parameter. I found this to be the least invasive option while still allowing our test suite to continue passing (locally).

---

stabilize dtype conversion across pandas, pyarrow, and modin test constructors

Introduce a shared helper to detect integer-like floats (e.g. 1.0, 2.0) and use it to control pandas' `convert_dtypes(…)` to avoid downcasting floats to integers.

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [x] 🐳 Other

## Related issues

- Related issue #3717 
- Closes #\<issue number\>

## Checklist

- [x] Tests added
- [x] Documented the changes


FBruzzesi (approved):
Thanks for taking the initiative @camriddell - I have no objections on this! I wonder if the constructors in #3552 should follow the same approach. WDYT?

dangotbanned:
Seems reasonable 🙂

@camriddell have you considered raising an issue in `pandas`?
I have to imagine it's cheaper for them not to do whatever inference is required for "whole number floats" 😅

camriddell:
> Thanks for taking the initiative @camriddell - I have no objections on this! I wonder if the constructors in #3552 should follow the same approach. WDYT?

Yeah, I think those constructors should take the same approach.

FBruzzesi:
@camriddell I am not sure what's going on with the release drafter 🫥 I would try to merge and see what happens there

PR #3756, opened 2026-07-07
feat: Widen `IntoSchema` to accept `IntoDType` values and `(name, dtype)` pair sequences

### LINKED ISSUE #3257: `typing.IntoSchema` too narrow

FBruzzesi:
# Description

Expand `IntoSchema` to `Mapping[str, IntoDType] | Sequence[tuple[str, IntoDType]] | Schema`, so uninstantiated
dtypes and pair sequences in `Schema`, `from_dict`, `from_dicts`, `from_numpy`, and
`sql.table` are type safe. `Schema` instantiates dtype classes on construction (its values remain `DType`), and schemas are
normalized at the public entry points so the compliant layer and backends only ever receive mappings.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3266
- Closes #3257


FBruzzesi:
@dangotbanned altair failure is due to the change from `Iterable` to `Sequence`:

```diff
- Mapping[str, DType] | Iterable[tuple[str, DType]] | None = None
+ Mapping[str, IntoDType] | Sequence[tuple[str, IntoDType]] | None = None,
```

but wouldn't it be more correct?

dangotbanned (changes requested):
> altair failure is due to the change from `Iterable` to `Sequence`:
> but wouldn't it be more correct?

No, and I've already told you before that `Sequence` was wrong 🙃

https://github.com/narwhals-dev/narwhals/pull/3266#discussion_r2482751880

There is nothing wrong with this code:

```py
    def schema(self, name: _Dataset, /) -> nw.Schema:
        it = ((col, _FIELD_TO_DTYPE[tp_str]()) for col, tp_str in self[name].items())
        return nw.Schema(it)
```

It also looks the same as all of the code that was in narwhals before https://github.com/narwhals-dev/narwhals/pull/3756/files#diff-193a3cb187e6f95645ce64e6dfb1e99ec9a30387785f2353a946d5bd3bae7a4a

Please don't make this change to `Sequence` when `Iterable` worked before

src/narwhals/schema.py
```diff
@@ -141,8 +148,10 @@ def from_arrow(cls, schema: IntoArrowSchema, /) -> Self:
         from narwhals._arrow.utils import native_to_narwhals_dtype
 
         return cls(
-            (field.name, native_to_narwhals_dtype(field.type, cls._version))
-            for field in schema
+            {
+                field.name: native_to_narwhals_dtype(field.type, cls._version)
+                for field in schema
+            }
```
dangotbanned:
```suggestion
            (field.name, native_to_narwhals_dtype(field.type, cls._version))
            for field in schema
```

src/narwhals/schema.py
```diff
@@ -247,8 +256,10 @@ def from_polars(cls, schema: IntoPolarsSchema, /) -> Self:
         from narwhals._polars.utils import native_to_narwhals_dtype
 
         return cls(
-            (name, native_to_narwhals_dtype(dtype, cls._version))
-            for name, dtype in schema.items()
+            {
+                name: native_to_narwhals_dtype(dtype, cls._version)
+                for name, dtype in schema.items()
+            }
```
dangotbanned:
```suggestion
            (name, native_to_narwhals_dtype(dtype, cls._version))
            for name, dtype in schema.items()
```

src/narwhals/schema.py
```diff
@@ -377,6 +388,10 @@ def _from_pandas_like(
...
         impl = implementation
         return cls(
-            (name, native_to_narwhals_dtype(dtype, cls._version, impl, allow_object=True))
-            for name, dtype in schema.items()
+            {
+                name: native_to_narwhals_dtype(
+                    dtype, cls._version, impl, allow_object=True
+                )
+                for name, dtype in schema.items()
+            }
```
dangotbanned:
```suggestion
            (name, native_to_narwhals_dtype(dtype, cls._version, impl, allow_object=True))
            for name, dtype in schema.items()
```

dangotbanned (changes requested):
All of this was utilizing `Iterable[tuple[str, IntoDType]]`, in the same way as in `altair` (https://github.com/narwhals-dev/narwhals/pull/3756#pullrequestreview-4651952464).

> [!NOTE]
> I wrote both 😉  

The problem appears to be that we haven't got test coverage for it.
Could we fix that please?

Here is a superset of things to test:

<details><summary>Show <code>tests/plan/schema_test.py:test_schema</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/d73c381def7aa8ca17f3f4d6687d63d1214e36bc/tests/plan/schema_test.py#L24-L51

</p>
</details>

src/narwhals/stable/v1/__init__.py
```diff
@@ -498,12 +497,6 @@ def any_value(self, *, ignore_nulls: bool = False) -> Self:
 class Schema(NwSchema):
     _version = Version.V1
 
-    @inherit_doc(NwSchema)
-    def __init__(
-        self, schema: Mapping[str, DType] | Iterable[tuple[str, DType]] | None = None
-    ) -> None:
-        super().__init__(schema)
-
```
dangotbanned, line 500: cheeky

src/narwhals/dataframe.py
```diff
@@ -571,7 +572,8 @@ def from_dict(
 
         Arguments:
             data: Dictionary to create DataFrame from.
-            schema: The DataFrame schema as Schema or dict of {name: type}. If not
+            schema: The DataFrame schema as Schema, dict of {name: type}, or a
+                iterable of (name, type) tuples. If not
                 specified, the schema will be inferred by the native library. If
                 any `dtype` is `None`, the data type for that column will be inferred
```
dangotbanned: I suppose there are quite a few of these guys

```suggestion
            schema: The DataFrame schema as Schema, dict of {name: type}, or a
                iterable of (name, type) tuples. 
                If not specified, the schema will be inferred by the native library. 
                If any `dtype` is `None`, the data type for that column will be inferred
```

src/narwhals/dataframe.py
```diff
@@ -741,10 +747,12 @@ def from_numpy(
         if not _is_into_schema(schema):
             msg = (
                 "`schema` is expected to be one of the following types: "
-                "Mapping[str, DType] | Schema | Sequence[str]. "
+                "IntoSchema | Sequence[str]. "
```
dangotbanned: I'd prefer if this went in the opposite direction and spelled out the parts of the alias.

If you hit this error, at least one of these is probably why:
1. You aren't using a type checker
2. You didn't read the docs

So `IntoSchema` has little chance of helping this frustrating person 😭

```suggestion
                "Schema | Mapping[str, IntoDType] | Iterable[tuple[str, IntoDType]] | Sequence[str].\n"
```

I've moved `Schema` to be first because if they aren't type checking, it is the simplest one

src/narwhals/dataframe.py
```diff
@@ -741,10 +747,12 @@ def from_numpy(
...
                 f"Got {type(schema)}."
             )
             raise TypeError(msg)
+        if not (schema is None or is_sequence_of(schema, str)):
+            schema = Schema(schema)
```
dangotbanned, line 755: I was confused by the check here.
So re-read of everything else that follows and the way `Sequence[str]` is handled is quite complicated. 

But not because of this PR - so I'm gonna fix that in a follow-up.

Thanks for the unintentional nerd-snipe 😉

tests/testing/conftest.py
```diff
@@ -8,12 +8,14 @@
...
-    from narwhals.typing import IntoSchema
+    from collections.abc import Mapping
+
+    from narwhals.typing import IntoDType
     from tests.conftest import Data
 
 
 @pytest.fixture(scope="module")
-def testing_schema() -> IntoSchema:
+def testing_schema() -> Mapping[str, IntoDType]:
```
dangotbanned: If we had any `type[DType]` cases, then `IntoDType` would make more sense


```suggestion
def testing_schema() -> Mapping[str, DType]:
```

src/narwhals/typing.py
```diff
@@ -309,20 +309,41 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+IntoSchema: TypeAlias = (
+    "Mapping[str, IntoDType] | Iterable[tuple[str, IntoDType]] | Schema"
+)
 """Anything that can be converted into a Narwhals Schema.
 
-Defined by column names and their associated *instantiated* Narwhals DType.
+Defined by column names and their associated Narwhals DType, either as a
+mapping or as an iterable of `(name, dtype)` tuples.
+Data types that take no required arguments may also be passed uninstantiated,
+e.g. `nw.UInt8` instead of `nw.UInt8()`.
```
dangotbanned: Pylance recently added support for mkdocstrings links 😄

```suggestion
"""Anything that can be converted into a [`narwhals.Schema`][].

Defined by column names and their associated [`DType`][narwhals.dtypes.DType], either as a
mapping or as an iterable of `(name, dtype)` tuples.

See Also:
    [`IntoDType`][narwhals.typing.IntoDType]
```

dangotbanned: ## VSCode

image

## API Reference

image

src/narwhals/schema.py
```diff
@@ -81,11 +83,19 @@ class Schema(OrderedDict[str, "DType"]):
...
+        self,
+        schema: Mapping[str, IntoDType] | Iterable[tuple[str, IntoDType]] | None = None,
     ) -> None:
-        schema = schema or {}
-        super().__init__(schema)
+        mapping: Mapping[str, IntoDType] = dict(schema) if schema else {}
+        super().__init__(
+            (name, dtype if isinstance(dtype, DType) else dtype())
+            for name, dtype in mapping.items()
+        )
```
dangotbanned: Besides the new tests that are explicitly checking for initialized `DType`s, *on an output* - is there a reason for initializing them?

Per #3257:

> This runs fine, as expected:
> 
> ```py
> print(nw.from_dict({'a': [1,2,3]}, schema={'a': nw.Int64}, backend='pyarrow'))
> ```

## Where do we really need `DType`?
- `Schema`
  - `.dtypes`
    - We use it in 2x test modules & in `testing.assert_frame_equal`
  - ~~`.to_{arrow,pandas,polars}`~~
    - They call functions which accept `IntoDType`
- `OrderedDict`
  - `.popitem`
- `dict`
  - `.values`
  - `.items`
  - `.get`
  - `.pop`
  - `.setdefault`
  - `.__getitem__`

> [!NOTE]
> At this point I have convinced myself that doing this lazily (instead) would be a bad idea. 
> But want to leave my workings here in case Marco suggests it 😂

dangotbanned (approved):
Thanks @FBruzzesi, really appreciate you taking my feedback on board ❤️ 

Nothing here is blocking, just potential icing on the cake 

hungry narwhal

src/narwhals/schema.py
FBruzzesi: The type checker would not be happy, probably rightfully so - since `Schema(OrderedDict[str, "DType"])` the input should an instance. On main `Schema(..).dtypes()` could end up returning classes.

**main**

```py
[type(d) for d in nw.Schema({'a': nw.Int64}).dtypes()]
[narwhals.dtypes.DTypeClass]
```

**now**

```py
[type(d) for d in nw.Schema({'a': nw.Int64}).dtypes()]
[Int64]
```

FBruzzesi:
Thanks @dangotbanned - I have addressed all the comments 🙏🏼

PR #3757, opened 2026-07-07
fix(pyarrow): Support constructing `Categorical` columns from Python values

mkzung:
# Description

Constructing a `Categorical` column from Python values raises on the pyarrow backend:

```python
import narwhals as nw
nw.new_series("s", ["a", "b"], dtype=nw.Categorical, backend="pyarrow")
# pyarrow.lib.ArrowTypeError: Array chunks must all be same type
```

The same happens via `from_dict` with a partial schema and via `from_dicts`. `Categorical` is a supported pyarrow dtype (it maps to `pa.dictionary(uint32, string)`, and only `Object`/`Enum` are unsupported), but `pa.chunked_array` and `pa.Table.from_pylist` can't build a dictionary-typed array straight from Python values.

I handle the dictionary case in the shared `chunked_array` helper (build a plain array, then cast to the dictionary type) and route the `from_dict` / `from_dicts` builders through it, so all three construction paths work. `from_native` and `.cast(nw.Categorical)` already worked and are unchanged; `Enum` stays unsupported and still raises.

Added a regression test for each entry point (new_series, from_dict, from_dicts), parametrised over the eager backends.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Self-found regression; no existing issue.

## AI assistance

- [x] AI tools were used.

Claude (Anthropic, Opus) was used for implementation assistance. I reviewed every line, built narwhals locally, and ran the test suite to confirm the fix and that the new tests pass. I take responsibility for the contribution.

## Checklist

- [x] Tests added
- [ ] Documented the changes

image

### DIFF
--- a/src/narwhals/_arrow/dataframe.py
+++ b/src/narwhals/_arrow/dataframe.py
@@ -178,16 +178,23 @@ class ArrowDataFrame(
         elif pa_schema is not None and any(
             pa.types.is_dictionary(field.type) for field in pa_schema
         ):
-            # `from_pylist` cannot build dictionary (Categorical) columns, and casting
-            # string -> dictionary is unsupported on older PyArrow, so pivot the rows to
-            # columns and let `from_pydict` build them with the requested schema.
-            native = pa.Table.from_pydict(
-                {
-                    field.name: [row.get(field.name) for row in data]
+            # `from_pylist` cannot build dictionary (Categorical) columns directly. Read
+            # the rows using the dictionaries' value types (so column selection and order
+            # follow the schema, like the branch below), then cast to the requested
+            # schema. Casting string -> dictionary needs PyArrow >=15; on older versions
+            # the backend raises, which is acceptable here.
+            pre_cast_schema = pa.schema(
+                [
+                    (
+                        field.name,
+                        field.type.value_type
+                        if pa.types.is_dictionary(field.type)
+                        else field.type,
+                    )
                     for field in pa_schema
-                },
-                schema=pa_schema,
+                ]
             )
+            native = pa.Table.from_pylist(data, schema=pre_cast_schema).cast(pa_schema)
         else:
             native = pa.Table.from_pylist(data, schema=pa_schema)
         return cls.from_native(native, context=context)
--- a/src/narwhals/_arrow/dataframe.py
+++ b/src/narwhals/_arrow/dataframe.py
@@ -9,6 +9,7 @@ import pyarrow.compute as pc
 from narwhals._arrow.series import ArrowSeries
 from narwhals._arrow.utils import (
     arange,
+    chunked_array,
     concat_tables,
     narwhals_to_native_dtype,
     native_to_narwhals_dtype,
@@ -142,9 +143,9 @@ class ArrowDataFrame(
             raise NotImplementedError(msg)
         res = pa.table(
             {
-                name: pa.chunked_array(  # type: ignore[misc]
+                name: chunked_array(  # type: ignore[misc]
                     [data[name] if data else []],
-                    type=narwhals_to_native_dtype(nw_dtype, version=context._version)
+                    narwhals_to_native_dtype(nw_dtype, version=context._version)
                     if nw_dtype is not None
                     else None,
                 )
@@ -174,6 +175,19 @@ class ArrowDataFrame(
         )
         if pa_schema and not data:
             native = pa_schema.empty_table()
+        elif pa_schema is not None and any(
+            pa.types.is_dictionary(field.type) for field in pa_schema
+        ):
+            # `from_pylist` cannot build dictionary (Categorical) columns, and casting
+            # string -> dictionary is unsupported on older PyArrow, so pivot the rows to
+            # columns and let `from_pydict` build them with the requested schema.
+            native = pa.Table.from_pydict(
+                {
+                    field.name: [row.get(field.name) for row in data]
+                    for field in pa_schema
+                },
+                schema=pa_schema,
+            )
         else:
             native = pa.Table.from_pylist(data, schema=pa_schema)
         return cls.from_native(native, context=context)
--- a/src/narwhals/_arrow/dataframe.py
+++ b/src/narwhals/_arrow/dataframe.py
@@ -9,7 +9,7 @@ import pyarrow.compute as pc
 from narwhals._arrow.series import ArrowSeries
 from narwhals._arrow.utils import (
     arange,
-    chunked_array,
+    chunked_array_from_values,
     concat_tables,
     narwhals_to_native_dtype,
     native_to_narwhals_dtype,
@@ -147,7 +147,7 @@ class ArrowDataFrame(
         version = context._version
         # NOTE: stubs don't allow `ChunkedArray` values, but `pa.table` accepts them
         arrays: Mapping[str, Any] = {
-            name: chunked_array(
+            name: chunked_array_from_values(
                 [data[name] if data else []],
                 narwhals_to_native_dtype(nw_dtype, version=version)
                 if nw_dtype is not None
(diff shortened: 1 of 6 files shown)


src/narwhals/_arrow/dataframe.py
```diff
@@ -173,6 +174,19 @@ def from_dicts(
...
+            # `from_pylist` cannot build dictionary (Categorical) columns, and casting
+            # string -> dictionary is unsupported on older PyArrow, so pivot the rows to
+            # columns and let `from_pydict` build them with the requested schema.
+            native = pa.Table.from_pydict(
+                {
+                    field.name: [row.get(field.name) for row in data]
+                    for field in pa_schema
+                },
+                schema=pa_schema,
+            )
```
FBruzzesi: I am a bit concerned about performances by this kind of looping for pivoting the data, what about something along the lines of:

```suggestion
        elif pa_schema is not None and any(
            pa.types.is_dictionary(field.type) for field in pa_schema
        ):
            native = pa.Table.from_pylist(data).cast(pa_schema)
```

Any chance you can run a benchmark with a medium size dataset?

If casting is not support for older versions it might be ok to let the backend raise for those old versions

src/narwhals/_arrow/utils.py
```diff
@@ -97,6 +97,11 @@ def chunked_array(
     if isinstance(arr, pa.ChunkedArray):
         return arr
     if isinstance(arr, list):
+        if dtype is not None and pa.types.is_dictionary(dtype):
+            # `pa.chunked_array` cannot build a dictionary-typed (Categorical) array from
+            # Python values, so build each chunk from the values with the target type
+            # (casting string -> dictionary is unsupported on older PyArrow).
```
FBruzzesi: Could you add a comment stating which is the min version for which we can do: `pc.cast(pa.chunked_array(arr), dtype)` instead?

Something like `TODO(Unassigned): Once min pyarrow version is x.y.z, we can simplify via pc.cast(pa.chunked_array(arr), dtype)`

FBruzzesi (commented):
Hey @mkzung  👋🏼 first and foremost thanks for your contribution and taking the initiative. - I am curious to know how you spotted this and if you have a use case :) 

---

I left a couple of comments but it shouldn't take too long before we merge this 😇

mkzung:
Thanks! I ran into it fuzzing the new `narwhals.testing` asserts module - `nw.new_series("s", ["a", "b"], dtype=nw.Categorical, backend="pyarrow")` blew up, which led me down to the arrow construction paths. The use case is just building small Categorical columns straight from Python values (test data, or a quick frame) without going through pandas or polars first.

Good call on the loop. I switched `from_dicts` to `from_pylist` + `cast`. On 200k rows with two categorical columns and one int, the pivot loop was around 33ms and `from_pylist` + `cast` is around 16ms, so about 2x, same result.

One small difference from your snippet: I pass a schema to `from_pylist` (the dictionaries' value types) instead of a bare `from_pylist(data)`. A bare `from_pylist` infers the column order from the data, so `from_dicts` with a schema whose column order differs from the row keys would raise on the cast, and it also wouldn't handle the inconsistent-keys case `from_dicts` supports. Passing the value-typed schema keeps it aligned to the schema and mirrors the non-categorical branch just below it, and it's a touch faster too since it skips inference. I added a small pyarrow-only test for the reordered case. Happy to drop it and just let that case raise if you'd rather keep it to your exact form.

For the version comment: string to dictionary cast is supported from pyarrow 15 (13 and 14 raise), so I put that in the `chunked_array` TODO. Since `from_dicts` now goes through the cast it raises on pyarrow < 15, which you said is fine - I xfailed that test on < 15 so minimum-versions stays green, and `new_series` / `from_dict` still use the per-chunk build so they keep working on the floor.

tests/from_dicts_test.py
```diff
@@ -87,3 +87,34 @@ def test_from_dicts_inconsistent_keys(
...
+        request.applymarker(pytest.mark.xfail)
+    schema = {"a": nw.Categorical()}
+    result = nw.from_dicts(
+        [{"a": "x"}, {"a": "y"}, {"a": "x"}], backend=eager_backend, schema=schema
+    )
+    assert result.collect_schema() == schema
+    assert_equal_data(result, {"a": ["x", "y", "x"]})
+
+
+def test_from_dicts_categorical_reordered_pyarrow() -> None:
```
FBruzzesi: It might be worth checking this for all backends to ensure consistency

FBruzzesi (approved):
Thanks for bearing with me @mkzung - It's a +1 from me! Left a non-blocking comment for testing

mkzung:
Thanks! The plain categorical case already runs on every eager backend - test_from_dicts_categorical is parametrized on eager_backend, so pandas, polars and pyarrow all exercise it (xfailed on pyarrow < 15, where the string->dictionary cast doesn't exist yet).

I kept the reordered test pyarrow-only because from_dicts with a schema doesn't order columns the same way on every backend. With schema {"a": Categorical, "b": Int64} and rows keyed {"b": .., "a": ..}, pyarrow and polars follow the schema and give ["a", "b"], but pandas keeps the order the keys appear in the rows and gives ["b", "a"]. That's independent of this PR, so a shared assertion on column order would fail on pandas. Happy to add a backend-agnostic version that only checks values and dtypes rather than column order, or leave a short comment noting the difference - whichever you prefer.

FBruzzesi:
> I kept the reordered test pyarrow-only because from_dicts with a schema doesn't order columns the same way on every backend. With schema {"a": Categorical, "b": Int64} and rows keyed {"b": .., "a": ..}, pyarrow and polars follow the schema and give ["a", "b"], but pandas keeps the order the keys appear in the rows and gives ["b", "a"]. That's independent of this PR, so a shared assertion on column order would fail on pandas. Happy to add a backend-agnostic version that only checks values and dtypes rather than column order, or leave a short comment noting the difference - whichever you prefer.

That's definitely something we should address. I would suggest to:

1. Extend the test to all backend
2. xfail those that are failing it (any pandas-like)
3. Make a followup to address pandas-like behavior, and align it with polars

src/narwhals/_arrow/dataframe.py
```diff
@@ -173,6 +174,26 @@ def from_dicts(
...
+                [
+                    (
+                        field.name,
+                        field.type.value_type
+                        if pa.types.is_dictionary(field.type)
+                        else field.type,
+                    )
+                    for field in pa_schema
+                ]
+            )
```
dangotbanned: This is too complicated.

We do not need to iterate over the schema twice for an edge case

src/narwhals/_arrow/utils.py
```diff
@@ -97,6 +97,13 @@ def chunked_array(
     if isinstance(arr, pa.ChunkedArray):
         return arr
     if isinstance(arr, list):
+        if dtype is not None and pa.types.is_dictionary(dtype):
+            # `pa.chunked_array` cannot build a dictionary-typed (Categorical) array from
+            # Python values, so build each chunk from the values with the target type.
+            # TODO(unassigned): once the minimum PyArrow is >=15.0.0, simplify to
+            # `pc.cast(pa.chunked_array(arr), dtype)` (string -> dictionary cast is
+            # supported from 15.0.0).
+            return pa.chunked_array([pa.array(chunk, dtype) for chunk in arr])
```
dangotbanned: I'm not comfortable with this.

`chunked_array` is called very frequently and this is not a common need

dangotbanned (changes requested):
I do not think we can merge this PR in it's current state.

I understand @FBruzzesi has already approved, but I would like to voice this now - ahead of suggesting an alternative.

In the meantime, @mkzung could you please open an issue demonstrating the bug, and then we can look at how to fix it?

I would also like to remind you of point 5 in (https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md#ai-assisted-contributions)


> 5. When interacting in issues, pull requests, discussion, discord, etc., **do not use LLMs to speak for you**,
    except for translation or grammar edits. Human-to-human communication is foundational to open source communities.

mkzung:
Opened #3775 with a minimal repro, per your suggestion. Happy to go with whatever fix approach you'd prefer.

mkzung:
Proposed a simpler fix on #3775 (one line: declare Categorical with int32 dictionary indices), which drops both of the changes you flagged here. Happy to reduce this PR to that, or close it and reopen against the issue - whichever you prefer.

mkzung:
@dangotbanned bumping this one in case it slipped by. The int32-index fix is on a branch ready to go and works back to pyarrow 13, so nothing here waits on Arrow. I'd lean toward folding that in now and leaving the uint32 path for the Arrow PR to fix upstream, but it's your call. Happy to wait if you'd rather. Just say which way and I'll sort it.

PR #3765, opened 2026-07-09
deps: replace black with ruff in `docs` dependency group

### LINKED ISSUE #3676: Avoid `black` docs dependency
## What happened?

While working on (#3497) I've dived into the mkdocstrings/zensical docs and tried out lots of configuration.

So I come bearing confusion on this old comment 😅 (https://github.com/narwhals-dev/narwhals/pull/1608)

https://github.com/narwhals-dev/narwhals/blob/07b5d94db70e79cbc96c2e08ed6d2c497248174a/pyproject.toml#L107-L108

Well, it turns out that support for `ruff` was merged **2 days later** in (https://github.com/mkdocstrings/python/pull/216) and released shortly after in [`1.13.0`].

> [!IMPORTANT]
> All of this was from 2024

## The fix?
We'd need to remove `black` from the `docs` group, since `mkdocstrings` [will pick it first if installed].

## Related
The only other place I've seen `black` referenced is here:

<details><summary>Show <code>pre-commit-config.yaml</code></summary>
<p>

https://github.com/narwhals-dev/narwhals/blob/07b5d94db70e79cbc96c2e08ed6d2c497248174a/.pre-commit-config.yaml#L115-L121

</p>
</details> 

1. Since that hook has inline dependencies declared, removing from `docs` might be a non-issue?
2. There are alternatives to `blacken-docs`, like 
    i. Ruff's (preview) [Markdown code formatting]
    ii. Zensical's [Snippets] extension ([more docs])

[`1.13.0`]: https://github.com/mkdocstrings/python/releases/tag/1.13.0
[will pick it first if installed]: https://github.com/mkdocstrings/python/blob/dc6aa93b1c956bc02904f33d1f6f635b275175d9/src/mkdocstrings_handlers/python/_internal/rendering.py#L508-L518
[Markdown code formatting]: https://docs.astral.sh/ruff/formatter/#markdown-code-formatting
[Snippets]: https://zensical.org/docs/setup/extensions/python-markdown-extensions/#snippets
[more docs]: https://facelessuser.github.io/pymdown-extensions/extensions/snippets/

PraveenKumarSridhar:
## Summary

`mkdocstrings >= 1.13.0` supports `ruff` as a code-block formatter and
prefers `black` only when both are installed. Pinning `black` in the
`[docs]` dependency group therefore pulls a heavy, unrelated formatter
into the docs install path purely to be shadowed by `ruff` anyway.

This drops the `black>=26.3.1` pin from `pyproject.toml`'s `[docs]`
group, which removes 8 transitive packages from `uv sync --group docs`.
The pre-commit `blacken-docs` hook pins its own `black` version inline
(`additional_dependencies: black==22.12.0`), so the hook itself is
unaffected.

Fixes #3676

## Changes

- `pyproject.toml`: remove `black>=26.3.1` from the `[docs]` dependency
  group
- `uv.lock`: regenerated; `black` and its transitive deps are gone
- `tests/dependencies/docs_group_no_black_test.py`: regression test that
  asserts no entry in the `[docs]` group has name `black`

## Testing

- New test `tests/dependencies/docs_group_no_black_test.py::test_docs_group_does_not_pull_black`
  fails on `upstream/main` (caught `black>=26.3.1`) and passes on this branch.
- Full `tests/dependencies/` suite: `87 passed, 13 skipped` locally on
  `Python 3.11.5` (skips are backend-availability, not related to this change).
- Wider smoke (selected core tests): `134 passed, 13 skipped, 1 xfailed`.
- `ruff check` + `ruff format` clean on the new test file.
- `uv tree --group docs` no longer pulls `black` (verified).

```
$ uv run pytest tests/dependencies/docs_group_no_black_test.py -v
collected 1 item
tests/dependencies/docs_group_no_black_test.py::test_docs_group_does_not_pull_black PASSED
```

## AI assistance

AI tools were used. The change was reproduced, implemented, and
verified locally before submission (failing test on `upstream/main`,
passing test on this branch; lockfile regenerated and inspected).


tests/dependencies/docs_group_no_black_test.py
FBruzzesi, line 1: I don't think we need a dedicated test for it to be honest

pyproject.toml
```diff
@@ -105,7 +105,6 @@ typing = [  # keep some of these pinned and bump periodically so there's fewer s
   {include-group = "plugins"},
 ]
 docs = [
-  "black>=26.3.1",  # required by mkdocstrings_handlers
```
FBruzzesi, line 108: We need to add ruff in here: in every CI job we run, we set a global env variable (`UV_NO_DEV: 1`) which makes uv not carry `dev` dependencies, and that's the only group specifying `ruff`

uv.lock
FBruzzesi, line 1: Nice one 👌🏼

FBruzzesi (commented):
Hey @PraveenKumarSridhar 👋🏼 thanks for your contribution!
I left a couple of comments that we might need to address before merging

tests/dependencies/docs_group_no_black_test.py
PraveenKumarSridhar: Done in fd84c48.

pyproject.toml
PraveenKumarSridhar: Done in fd84c48 — added ruff to [docs] and regenerated uv.lock.

uv.lock
PraveenKumarSridhar: Thanks!

FBruzzesi (approved):
Thanks @PraveenKumarSridhar

PraveenKumarSridhar:
Quick FYI on the one red check: `altair (3.12, ubuntu-latest)` fails on the `mypy altair tests` step, but the failure is in altair's own `vegalite/v6/schema/_config.py` and `vegalite/v6/api.py` (5 `[unused-ignore]` errors on `# type: ignore` comments) — unrelated to this PR. The pytest step in the same job is green (1880 passed). #3767 looks like the active fix on the maintainer side.

Marking ready for review so this can move forward; happy to address anything else.

PraveenKumarSridhar:
Thanks for the review and the merge, @FBruzzesi! 🎉

PR #3770, opened 2026-07-10
fix: reject a negative `width` in `str.zfill` consistently across backends

mkzung:
# Description

A negative `width` or `length` reached the backends and each one did something
different with it. pandas returned the string unchanged, Polars raised
`InvalidOperationError: conversion from 'i128' to 'u64' failed ... [-1]`, and
PyArrow raised `ArrowInvalid: Negative buffer resize: -21`. `pad_start` and
`pad_end` do the same, since the argument reaches the same backend calls.

This started as a PyArrow-only fix for `zfill`. @camriddell raised the question of
whether the check belonged further up in `expr_str.py` instead, so it raises early
and every backend reports the same thing. That is what it now does. The PyArrow
backend keeps only its `width == 0` short-circuit, which is an evaluation-order
problem rather than an argument one.

The check is in both `expr_str.py` and `series_str.py` because `Series.str.zfill`
calls the compliant series directly rather than going through an expression, so
`expr_str.py` alone would not cover it. Removing either set turns tests red.

`pc.case_when` is why PyArrow crashed rather than returned: it evaluates every
branch eagerly, so `pc.utf8_lpad(remaining_chars, width - 1)` ran with a negative
pad length before any row was selected.

The tests pin the message and not just the exception type, which matters here more
than it usually would. Polars already raised `InvalidOperationError` for this, just
with its internal cast message, so without the message pinned the Polars case
passes against unfixed code. `InvalidOperationError` is also what the library
raises elsewhere for an argument rejected before anything is computed.

Tests run on every backend through the existing fixtures. The expression cases use
`constructor` rather than `constructor_eager`, so the lazy backends are covered as
well, where the check raises at `select()` rather than at `collect()`.

Reproduction, before this change:

    import narwhals as nw, pyarrow as pa
    s = nw.from_native(pa.table({"a": ["1", "-5", "abc"]}), eager_only=True)["a"]
    s.str.zfill(0)   # ArrowInvalid: Negative buffer resize: -9
    s.str.zfill(-1)  # ArrowInvalid: Negative buffer resize: -21

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- No existing issue; found while exercising the string namespace on the PyArrow backend.

## AI assistance

- [x] AI tools were used.

I used an AI coding assistant to help locate the root cause and draft the regression
test. I have read, run, and verified every line, and I take full responsibility for
the change.

## Checklist

- [x] Tests added
- [ ] Documented the changes (no doc change needed; behaviour matches the documented `str.zfill`)


MarcoGorelli:
thanks @mkzung !

>  If you would rather the Arrow backend raise on a negative width to match Polars
instead of returning unchanged, I am happy to switch.

i think i'd prefer that, could we do that instead please? thanks 🙏

mkzung:
Done. The Arrow backend now raises InvalidOperationError on a negative width, matching Polars, instead of returning unchanged, and width 0 stays a no-op. I also moved the regression tests to pyarrow only, since Polars' own zfill(0) raises on the older CI versions.

src/narwhals/_arrow/series_str.py
```diff
@@ -114,6 +115,16 @@ def to_titlecase(self) -> ArrowSeries:
         return self.with_native(pc.utf8_title(self.native))
 
     def zfill(self, width: int) -> ArrowSeries:
+        if width < 0:
+            # Polars rejects a negative zfill width (its argument is unsigned).
+            # Match that here rather than letting pyarrow's utf8_lpad(width - 1)
+            # crash later with a negative pad width.
+            msg = f"zfill does not support a negative width, got {width}."
+            raise InvalidOperationError(msg)
```
camriddell: @MarcoGorelli should we push this check up to [narwhals/expr_str.py](https://github.com/narwhals-dev/narwhals/blob/e7dd05e11f9e8cde5ece83da43d9bd622d8ec4eb/src/narwhals/expr_str.py#L556). If we do that then 
1. The error can be raised as early as possible for all backends
2. All implementations will have a uniform error message

If not, then I'm happy with the implementation here.

mkzung:
Worth adding that the inconsistency is wider than zfill. With a negative width today, pandas returns the string unchanged, Polars raises `conversion from 'i128' to 'u64' failed`, and pyarrow crashes inside `utf8_lpad`. `pad_start` and `pad_end` do the same thing - on pyarrow they leak `ArrowInvalid: Negative buffer resize`.

So moving the check up to `expr_str`/`series_str` fixes three methods at once and gives every backend the same error. I have it on a branch: validation in `Expr.str` and `Series.str`, the Arrow negative-width raise removed (the `width == 0` no-op stays, that one is a genuine pyarrow quirk), tests parametrised across backends, full suite green.

https://github.com/mkzung/narwhals/compare/main...mkzung:narwhals:fix/zfill-validate-in-public-layer

Happy to fold it into this PR, or to leave it until you and @MarcoGorelli have settled where the check belongs.

mkzung:
Moved it up to expr_str.py, so every backend now raises the same message before
anything reaches a compliant object.

I kept the guards in series_str.py as well. Series.str.zfill calls the compliant
series directly rather than building an expression, so the expression layer alone
does not cover that path. With the series-side checks removed, 8 tests in
tests/expr_and_series/str fail.

The _arrow change is still in there for the case Marco raised earlier: a negative
width returned the string unchanged on that backend instead of raising.

src/narwhals/_arrow/series_str.py
```diff
@@ -114,6 +114,10 @@ def to_titlecase(self) -> ArrowSeries:
         return self.with_native(pc.utf8_title(self.native))
 
     def zfill(self, width: int) -> ArrowSeries:
+        if width == 0:
+            # Nothing to pad, and short-circuiting avoids pc.case_when eagerly
+            # evaluating the utf8_lpad(width - 1) = utf8_lpad(-1) branch.
+            return self.with_native(self.native)
```
camriddell: Thanks for leaving the comments explaining the code. Was this also an issue for the other arrow `str` methods?

tests/expr_and_series/str/zfill_test.py
```diff
@@ -64,3 +64,32 @@ def test_str_zfill_series(
...
     assert_equal_data({"a": result}, expected)
+
+
+def test_str_zfill_zero_width_pyarrow() -> None:
+    # A zero width is a no-op (every string is already at least 0 long). This used
+    # to crash on the pyarrow backend with `ArrowInvalid: Negative buffer resize`,
+    # because pc.case_when eagerly evaluated the utf8_lpad(width - 1) = -1 branch.
+    pytest.importorskip("pyarrow")
+    result = nw.from_dict(data, backend="pyarrow")["a"].str.zfill(0)
+    assert_equal_data({"a": result}, data)
```
MarcoGorelli: shall we parametrise over all constructors, to make sure none of the others crash?

src/narwhals/series_str.py
```diff
@@ -488,6 +489,9 @@ def zfill(self, width: int) -> SeriesT:
             3    123456
             dtype: str
         """
+        if width < 0:
+            msg = f"`width` must be non-negative but got {width}"
+            raise InvalidOperationError(msg)
```
MarcoGorelli: this is repeated a few times, could/should it be a util function? like `validate_width`

mkzung: Pulled the check into a `validate_width` helper in `_utils.py`, used by zfill and both pad methods.

tests/expr_and_series/str/zfill_test.py
mkzung: The zero-width test runs over every constructor now (plus a series version) - nothing crashes, only the pandas+pyarrow xfail.

src/narwhals/_arrow/series_str.py
mkzung: Checked the other arrow methods: pad_start/pad_end don't hit it. zfill was special because it does `utf8_lpad(width - 1)`, so width=0 asked pyarrow for length -1; the pad methods pass the length straight through and pad_start(0)/pad_end(0) return the string unchanged. The negative case is handled for all three at the public layer now.

camriddell:
@mkzung looks like the above failures are from a behavior alignment correction we implement for older Polars versions.

https://github.com/narwhals-dev/narwhals/blob/34edd7eeb9b3f8cd2b1e46b913197fa3561a96d8/src/narwhals/_polars/expr.py#L464

I think if we replace the linked line `.str.zfill(width - 1)` with `.str.zfill(max(width - 1, 0))` we should be looking good on the tests and wrap this one up.

mkzung:
Done, pushed.

The reason it fires with the predicate false is that Polars builds the `then` arm before choosing, so width 0 reaches `zfill(-1)` either way. Checked against a real 1.30.0 rather than my local 1.42, where the branch never runs: all three failures reproduce there and clear with the change.

One thing I ran into while checking, separate from this PR and already on main: that branch disagrees with newer Polars when the remainder is multibyte or holds a second plus. `"+é"` at width 4 comes out `"++0é"` against `"+0é"`, and `"++7"` comes out `"++07"` against `"+0+7"`. It mixes units, since `zfill` pads to bytes and `pad_start` to characters. Happy to open a follow-up if you want it.

camriddell:
> One thing I ran into while checking, separate from this PR and already on main: that branch disagrees with newer Polars when the remainder is multibyte or holds a second plus. "+é" at width 4 comes out "++0é" against "+0é", and "++7" comes out "++07" against "+0+7". It mixes units, since zfill pads to bytes and pad_start to characters. Happy to open a follow-up if you want it.

I cannot replicate this. Can you re-confirm and if this issue persists then create an issue following the issue template (with an MRE).

mkzung:
You are right not to be able to replicate it, and one half of what I wrote was wrong.

It only happens on Polars `<= 1.30.0`, which is the only case that enters that branch at all. I should have said so. I have now checked it on a real 1.30.0 rather than by pinning the reported backend version, and the `"++7"` half does not hold: it comes out `"+0+7"` on both, matching newer Polars. Sorry for the noise on that one.

What does hold is multibyte content after a leading `+`. Same narwhals commit, same input, two Polars:

```
polars 1.42.1   "+é".zfill(4) -> "+0é"
polars 1.30.0   "+é".zfill(4) -> "++0é"
```

Filed as #3838 with the MRE and the version matrix. The cause is that the branch mixes units, `zfill` padding to bytes and `pad_start` to characters, so on ASCII they agree and on multibyte one `+` too many is prepended.

The two red checks here are `Install uv` failing to fetch on the lightgbm job and the downstream scikit-lego job, which has been red on main since April.

camriddell (approved):
My only remaining concern is that some of the inline comments are quite verbose, however I don't feel this should hold up the PR.

Changes are looking good, so I'll go ahead and approve.

@MarcoGorelli can you provide your opinion on the verbosity of the comments and request changes if you see fit to have them removed?

src/narwhals/expr_str.py
```diff
@@ -581,6 +582,7 @@ def pad_start(self, length: int, fill_char: str = " ") -> ExprT:
             |3           NaN           NaN|
             └─────────────────────────────┘
         """
+        validate_width(length, "length")
```
MarcoGorelli: `validate_width(lengh` looks quite odd, no?

MarcoGorelli (commented):
thanks for your pr

agree that the comments in the source code can be removed, as well as the extra docstring in `validate_width`, it's clear what the function does

maybe just `validate_is_non_negative` and don't make it specific to width?

MarcoGorelli:
> One thing I ran into while checking, separate from this PR and already on main

> You are right not to be able to replicate it, and one half of what I wrote was wrong.

To be completely honest, this reads to me as so AI generated that I feel like rejecting the PR in order to not encourage such behaviour

It's one to use AI to help with code, but as a reply to a PR review?

mkzung:
You're right about those replies, they weren't written by me and I shouldn't have posted them like that. I won't do it again.

The three things you asked for are in: validate_width is now validate_is_non_negative with no default for arg_name, the docstring is gone, and the comments are cut.

CI is showing action_required on every job, so it needs someone with access to release it.

FBruzzesi:
@camriddell I got to this PR only now. I added one commit to it: https://github.com/narwhals-dev/narwhals/pull/3770/commits/4526a14b2b9586282d8dcc730a2153fb9ebad62f

Let me know your thoughts about it. The commit message went terribly wrong, it should have been something along the line of "fix: unify non-negative validation across str methods and backends".

Edit: For the changes in head/slice, I think we should better align with polars. At the moment is a bag of mix behavior

camriddell:
> @camriddell I got to this PR only now. I added one commit to it: [4526a14](https://github.com/narwhals-dev/narwhals/commit/4526a14b2b9586282d8dcc730a2153fb9ebad62f)
> 
> Let me know your thoughts about it. The commit message went terribly wrong, it should have been something along the line of "fix: unify non-negative validation across str methods and backends".
> 
> Edit: For the changes in head/slice, I think we should better align with polars. At the moment is a bag of mix behavior

The error message stuff looks good.

Agreed that `head` behavior should be re-aligned to support negative indexing before landing this PR. How come we're now guarding against non-negatives there?  Noticing that was added in your commit.

FBruzzesi:
> The error message stuff looks good.

Thanks @camriddell 🙏🏼

> Agreed that `head` behavior should be re-aligned to support negative indexing before landing this PR. How come we're now guarding against non-negatives there? Noticing that was added in your commit.

~~Ok let me do some more homework before committing to that. I will take care of rolling back my changes~~

Rolled-back the changes, will follow up with `str.slice` and `str.head` behavior alignment.

Failing marimo CI is unrelated

FBruzzesi:
@camriddell I rescoped this PR as it had a clear overlap with #3909 and #3910 which are already merged in main

mkzung:
Thanks, and the rescope is right.

What is left after #3909 and #3910 is the zfill guard. str.zfill on main takes a negative width without a word, and this shares the existing pad_start/pad_end check rather than adding a second one. Happy to drop it if you would rather not carry the extra validation.

camriddell (approved):
Kind of a long comment in the `str/zfill_test.py` file, but I'd rather land this PR than go back for edits. If they truly bother us in the future, it is easy to remove those comments.

FBruzzesi:
Thanks everyone 🙌🏼 @mkzung sorry for the long wait and merging similar PRs before this one

PR #3753, opened 2026-07-05
feat: Extend plugin support for IO and `from_*` methods/functions

FBruzzesi:
# Description

The core changes are in the `eager_namespace` for the backend resolution logic, which allows for `backend=` arguments to resolve for Narwhals plugins: users can pass a plugin's entry point name, module name, or the module itself.

* IO functions `{read,scan}_{csv,parquet}` dispatch to same-named top-level plugin hooks.
* Eager constructors (`from_*`, `new_series`, `DataFrame.from_*`, `Series.from_*`) work with zero extra plugin code if the plugin's compliant namespace implements the `EagerNamespace` protocol.

---

Most of the lines changes are in `packages/test-plugin` and `tests/plugins_test.py` so that we have coverage for the functionalities. 

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3713
- Related #3042: the `narwhals.plugins` entry-point group _is_ the registration mechanism we have. Not sure we can do _much_ more than this dispatching

PR #3793, opened 2026-07-15
refactor: Dispatch IO functions via `*Namespace` methods

FBruzzesi:
# Description

Follow-up to the contract discussion in #3753 (https://github.com/narwhals-dev/narwhals/pull/3753#discussion_r3564162949)

`read_csv`, `scan_csv`, `read_parquet` and `scan_parquet` now dispatch to same-named methods on the compliant `*Namespace` instead of per-backend `if`/`elif` chains in `functions.py`.

- [x] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3753


MarcoGorelli (approved):
thanks @FBruzzesi !

src/narwhals/_compliant/namespace.py
```diff
@@ -109,6 +109,12 @@ def list(self, *exprs: CompliantExprT) -> CompliantExprT: ...
     def selectors(self) -> CompliantSelectorNamespace[Any, Any]: ...
     def struct(self, *exprs: CompliantExprT) -> CompliantExprT: ...
     def coalesce(self, *exprs: CompliantExprT) -> CompliantExprT: ...
+    # NOTE: IO methods, both take an already-normalized path (see `narwhals._utils.normalize_path`)
+    # and forward `kwds` to the native reader.
+    def scan_csv(
+        self, source: str, *, separator: str = ",", **kwds: Any
```
dangotbanned: You *could* add an alias for `str` with this comment as a docstring, linking to `narwhals._utils.normalize_path`. 

Then we'd be able to see it on hover for all IO methods 🙂

src/narwhals/_spark_like/utils.py
```diff
@@ -17,16 +17,18 @@
...
     from narwhals._compliant.typing import CompliantLazyFrameAny
-    from narwhals._spark_like.dataframe import SparkLikeLazyFrame
+    from narwhals._spark_like.dataframe import SparkLikeLazyFrame, SQLFrameDataFrame
     from narwhals._spark_like.expr import SparkLikeExpr
     from narwhals.dtypes import DType
     from narwhals.typing import IntoDType
 
     _NativeDType: TypeAlias = sqlframe_types.DataType
     SparkSession = Session[Any, Any, Any, Any, Any, Any, Any]
+    SparkReader = _BaseDataFrameReader[SparkSession, SQLFrameDataFrame, Any]
```
dangotbanned, line 31: I like it!

src/narwhals/functions.py
```diff
@@ -684,44 +657,27 @@ def read_csv(
...
             # implement `read_csv` function in the top-level namespace.
-            native_frame = native_namespace.read_csv(source=source, **kwargs)
+            native_frame: NativeDataFrame = native_namespace.read_csv(
+                source=source, **kwargs
+            )
         except AttributeError as e:
             msg = "Unknown namespace is expected to implement `read_csv` function."
             raise AttributeError(msg) from e
-    return from_native(native_frame, eager_only=True)
+        return from_native(native_frame, eager_only=True)
```
dangotbanned, line 675: Is the intention that this PR keeps this (roughly) as-is, and then in #3753 we replace/remove it?

If so, I believe that would close #2786 😄

docs/extending.md
```diff
@@ -59,7 +59,33 @@ handle plugins. For this integration to work, any plugin architecture must conta
...
+
+The Narwhals IO functions (`read_csv`, `scan_csv`, `read_parquet`, `scan_parquet`)
+dispatch to same-named methods on the compliant namespace. This is a single mechanism,
+shared by built-in backends and extensions alike: to support these functions, a
+compliant namespace implements (a subset of):
+
+- `read_csv(source, *, separator=",", **kwargs)`, returning a compliant DataFrame.
+- `read_parquet(source, **kwargs)`, returning a compliant DataFrame.
+- `scan_csv(source, *, separator=",", **kwargs)`, returning a compliant frame.
+- `scan_parquet(source, **kwargs)`, returning a compliant frame.
```
dangotbanned: We can use types for **what**, and save words for **why** (if needed):

```suggestion
```py
def read_csv(self, source: str, *, separator: str = ",", **kwds: Any) -> CompliantDataFrame: ...
def scan_csv(self, source: str, *, separator: str = ",", **kwds: Any) -> CompliantFrame: ...
def read_parquet(self, source: str, **kwds: Any) -> CompliantDataFrame: ...
def scan_parquet(self, source: str, **kwds: Any) -> CompliantFrame: ...
\`\`\`
```

I had to do some weird escaping to get the suggestion to work 😔, but it shouldn't be needed in the actual markdown

docs/extending.md
```diff
@@ -59,7 +59,33 @@ handle plugins. For this integration to work, any plugin architecture must conta
...
+
+In all cases:
+
+- `source` is a plain string: Narwhals normalizes `Path` and path-like inputs before
+  dispatching to the namespace.
```
dangotbanned: So, I know I suggested a `TypeAlias` of `str` in (https://github.com/narwhals-dev/narwhals/pull/3793#discussion_r3621528956).
But this would be another case where `NewType` fits in.

[Here's another article](https://web.archive.org/web/20230519162111/https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html) on how to spot them 🙂

I guess our issue is that every backend supports `str`, but in python we want to use `pathlib.Path`.
We can often get away with passing `pathlib.Path` to these backends anyway, but those pesky Windows path separators come along to spoil the party!

So we revert to a primitive, but do so with a wrapper that ensures it ain't just any old `str` 😎

dangotbanned (approved):
Thanks @FBruzzesi, really appreciate this!

cat-ptain on a narwhal

FBruzzesi:
Thanks @dangotbanned - addressed all your feedback in https://github.com/narwhals-dev/narwhals/pull/3793/commits/264b5a31119c0a06d3931de9da3315c9ae2b6fea - Update: except that the wrong doc file was committed, so look at https://github.com/narwhals-dev/narwhals/pull/3793/commits/2f26bee7f98c7b21620dcde55f2c954d1894bcf3 🤦🏼‍♂️

Looking forward to merge this 🔥

dangotbanned (approved):
Double approved

bullseye

Issue #3149, opened 2025-09-23
[Enh]: is_eager_allowed method for Implementation


TheSkyentist:
### We would like to learn about your use case. For example, if this feature is needed  to adopt Narwhals in an open source project, could you please enter the link to it below?

In the Astropy project we are adding a generic conversion layer using Narwhals (see astropy/astropy#18435). This interacts with the Astropy Table format, which is an astronomy oriented data structure designed to interact with domain-specific file types and data. 

We currently can only support converting to eager backends. Therefore we implement a check to ensure that the end user is converting to an eager-compatible backend. If they request a non-eager backend, we reject raise an error, otherwise the conversion goes through. 

Currently we use the `nw._utils.is_eager_allowed` which is a private/undocumented method and subject to change in the future and, I believe, would not be pinned by the API version guarantee. 

We therefore would like to expose the `is_eager_allowed` API in the Implementation class, ideally so that a user could instantiate an nw.Implementation and determine if the Implementation is eager compatible without consulting the documentation. 

### Please describe the purpose of the new feature or describe the problem to solve.

The new feature would enable users to know if the relevant implementation provides Eager support. Currently this is only exposed in a private API that is not documented. As this function is designed to take an Implementation instance and return information about it, it seems apt to have it be a class method. 

### Suggest a solution if possible.

Moving the private method in `nw._utils` to the exposed implementation class would accomplish this goal.

dangotbanned:
@TheSkyentist 

So I've spent some time with [this version](https://github.com/TheSkyentist/astropy/blob/7f3504bef2981ef6bc0d843fc46e9e1d25dbd893/astropy/table/_dataframes.py) of `astropy.table._dataframes.py` to try and understand why this is needed.

### Repro
AFAICT, this is the only case that would benefit from validation:

```py
import narwhals as nw

data = {"a": ["A", "B", "A"], "b": [1, 2, 3]}
nw.from_dict(data, backend="duckdb")
```

Currently that would raise the following error:
```py
ValueError: duckdb support in Narwhals is lazy-only, but `from_dict` is an eager-only function.

Hint: you may want to use an eager backend and then call `.lazy`, e.g.:

    nw.from_dict({'a': [1, 2]}, backend='pyarrow').lazy('duckdb')
```

[...]

PR #3794, opened 2026-07-15
feat(typing): Introduce `PluginName` NewType so plugin backends can type check

FBruzzesi:
# Description

Other prerequisite for #3753 

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related
- Rabbit hole that led us here (https://github.com/narwhals-dev/narwhals/pull/3753#discussion_r3573928175)


src/narwhals/_typing.py
```diff
@@ -91,7 +91,25 @@
...
+- a value explicitly wrapped as `PluginName("...")` is accepted.
+
+The contract is that a wrapped string **must** name an installed plugin's
+entry point in the `narwhals.plugins` group.
+
+Signatures that dispatch to plugins include it in the `IntoBackend`
+parameter, e.g. `IntoBackend[EagerAllowed | PluginName]`.
+"""
+
+BackendT = TypeVar("BackendT", bound=Backend | PluginName)
```
dangotbanned, line 115: Thank you for this!

Much better idea than I had, while getting to the same goal 🥳

**Edit**: woops that was meant to start the review

tests/plugins_test.py
```diff
@@ -8,10 +8,12 @@
 import narwhals.stable.v1.dependencies as nw_v1_dependencies
 import narwhals.stable.v2.dependencies as nw_v2_dependencies
 from narwhals import dependencies as nw_dependencies
+from narwhals._typing import PluginName
```
dangotbanned: This suggestion is only a part of it 😄

```suggestion
from narwhals.plugins import PluginName
```

I guess the main part is, do we need to make a decision *here* on this bit of (https://github.com/narwhals-dev/narwhals/pull/3753#discussion_r3573928175)?

> All that is left is to decide *where* the wrapping in `PluginName` happens
> I see 3 options for that, but it could be a mix of them too:

<details><summary>Show me the money</summary>
<p>

1. The plugin author does it explicitly, somewhere here (and/or alongside where they write `Implementation.UNKNOWN`) https://github.com/narwhals-dev/narwhals/blob/314d18539c66c0a6c966b462fb9e1c21c39b2a4b/src/narwhals/plugins.py#L107-L114
2. We wrap `EntryPoint.name` implicitly here https://github.com/narwhals-dev/narwhals/blob/314d18539c66c0a6c966b462fb9e1c21c39b2a4b/src/narwhals/plugins.py#L54-L56
3. We expose `PluginName` and users/downstream do everything imperatively

I like first choice.  
The second would block how we might solve *How do we introduce the concepts of `{Eager,Lazy}{Allowed,Only}` for plugins, when we can't define them in a `Literal`?*.  
The third I'd be okay with *as an escape hatch* - but as the only option it would be cumbersome.

</p>
</details>

Maybe the simplest thing we could do in this PR is make `PluginName` accessible from `narwhals.plugins`?
It could be defined in `_typing.py` or in `plugins.py`

src/narwhals/_typing.py
```diff
@@ -91,7 +91,25 @@
 - An Implementation, such as: `Implementation.DASK`, `Implementation.PYSPARK`, ...
 """
 
-BackendT = TypeVar("BackendT", bound=Backend)
+PluginName = NewType("PluginName", str)
+"""Name of a plugin backend's [entry point](https://packaging.python.org/en/latest/specifications/entry-points/).
```
dangotbanned: Hope this isn't too nitpik-y

```suggestion
"""Name of a plugin's [entry point](https://packaging.python.org/en/latest/specifications/entry-points/).
```

*So far* all the docs on plugins haven't actually used the word *backend*:

- https://github.com/narwhals-dev/narwhals/blob/848aec5cea715323527982e424ab2e9e7b5461e3/src/narwhals/plugins.py
- https://github.com/narwhals-dev/narwhals/blob/848aec5cea715323527982e424ab2e9e7b5461e3/docs/extending.md


No idea if that was intentional, but this suggestion would be consistent with that 🙂

src/narwhals/_typing.py
```diff
@@ -91,7 +91,25 @@
...
+
+Plugin backends are discovered at runtime, so their names cannot join the
+`Literal` unions that describe the built-in backends. `PluginName` is a
+[`NewType`](https://typing.python.org/en/latest/spec/aliases.html#newtype):
+type checkers treat it as a distinct subtype of `str`, meaning
```
dangotbanned: I think this reads like docs about `NewType` itself vs `PluginName`.

That's not to say that we shouldn't document how it works (it could help with spotting where else we can use it). 
But maybe this would make more sense in the `_typing.py` module doc instead?

src/narwhals/_typing.py
```diff
@@ -91,7 +91,25 @@
...
+
+- an arbitrary `str` is still rejected where a `backend` is expected, and
+- a value explicitly wrapped as `PluginName("...")` is accepted.
+
+The contract is that a wrapped string **must** name an installed plugin's
+entry point in the `narwhals.plugins` group.
```
dangotbanned: I think this could belong in `plugins.py` somewhere, since it is directed at plugin authors IIUC?

tests/plugins_test.py
```diff
@@ -132,3 +134,40 @@ def test_typing() -> None:
...
+        # Rejected: opaque strings do not satisfy `PluginName`.
+        nw.from_dict(data, backend=dynamic_string)  # type: ignore[arg-type]
+        nw.new_series("a", [1, 2], backend=dynamic_string)  # type: ignore[arg-type]
+        nw.Implementation.from_backend(dynamic_string)  # type: ignore[arg-type]
+
+        # Rejected: lazy-only literals on eager constructors (no regression).
+        nw.from_dict(data, backend="duckdb")  # type: ignore[arg-type]
+
+        # Rejected: `.lazy` does not dispatch to plugins (yet).
+        df.lazy(plugin_name)  # type: ignore[arg-type]
```
dangotbanned, line 173:
```suggestion
        df.lazy(plugin_name)  # type: ignore[arg-type]
        lf = df.lazy()
        # Rejected: `.collect` does not dispatch to plugins (yet).
        lf.collect(plugin_name)  # type: ignore[arg-type]
```

src/narwhals/_utils.py
```diff
@@ -140,8 +141,6 @@
         _SliceNone,
     )
 
-    UnknownBackendName: TypeAlias = str
```
dangotbanned, line 143: very nice

src/narwhals/series.py
```diff
@@ -186,7 +186,7 @@ def from_iterable(
         values: Iterable[Any],
         dtype: IntoDType | None = None,
         *,
-        backend: IntoBackend[EagerAllowed],
+        backend: IntoBackend[EagerAllowed | PluginName],
```
dangotbanned, line 189: > [!NOTE]
> Not a review, just trying to add context from our conversation on discord

If you are downstream from Narwhals and make use of our typing - we have been suggesting (https://github.com/narwhals-dev/narwhals/issues/3149#issuecomment-3334040552) you write things like:

```py
backend: IntoBackend[EagerAllowed]
```

Following this PR, that type describes **exactly the same thing**.
It does not indicate plugin support, meaning that there should be no expectation that downstream tests against plugins.

*If* downstream wants to opt-in to plugins, the update to their **typing** small: 

```diff
- backend: IntoBackend[EagerAllowed]
+ backend: IntoBackend[EagerAllowed | PluginName]
```

But it gives a very clear signal (to their users) that plugins should work. 
I *hope* that downstream (authors) will be more likely to update their **tests**, to account for plugins, when these concepts are visible in each signature with a `backend` 🙂

src/narwhals/_typing.py
```diff
@@ -91,7 +91,25 @@
...
+
```
dangotbanned, line 110: This is regarding both ([L98-L100](https://github.com/narwhals-dev/narwhals/pull/3794#discussion_r3595050338)) and ([L105-L106](https://github.com/narwhals-dev/narwhals/pull/3794#discussion_r3595067749)) comments

---

I think the doc for `PluginName` should feel kinda like [`IntoBackend`](https://narwhals-dev.github.io/narwhals/api-reference/typing/#narwhals.typing.IntoBackend) and [`EagerAllowed`](https://narwhals-dev.github.io/narwhals/api-reference/typing/#narwhals.typing.EagerAllowed)

## Example
If I were using `from_dict` and hover over the types for `backend`, I see these guys:

<details><summary><code>EagerAllowed</code></summary>
<p>

image

</p>
</details> 


<details><summary><code>PluginName</code></summary>
<p>


image

</p>
</details> 

`PluginName` provides much more detail and that itself is no problem at all! 😄 
But what is said in `EagerAllowed` feels more contextually helpful for me (`from_dict` guy)

## Suggestion/Question?
The content is good, but if you sprinkle little bits all over the place - I think it is more likely to land in front of who needs it.

These are the audiences I see being addressed in the doc currently:

- `PluginName` as the tool for plugin authors
- `PluginName` as the annotation for downstream users
- `PluginName` as the first use we have of `NewType` (optional)

Do you see other places we could reach each of those?

dangotbanned (commented):
Thanks @FBruzzesi, just a few suggestions/questions from me

this narwhal loves you

tests/plugins_test.py
FBruzzesi: That's a good point. I would also prefer option 1 if feasible. As you also mentioned, for now I will just make `PluginName` exposed in `plugins.py`

src/narwhals/_typing.py
FBruzzesi: I will do my best

src/narwhals/_typing.py
FBruzzesi: Adding as module docs - not sure where to place it to be honest! We can consider moving it to `extending.md` in #3753

src/narwhals/series.py
FBruzzesi: All this _should_ go somewhere! but where exactly? 🧐

FBruzzesi: Also made me realize that one should import from two modules:

```py
from narwhals.typing import IntoBackend, EagerAllowed
from narwhals.plugins import PluginName

...
```

src/narwhals/_typing.py
```diff
@@ -91,7 +102,17 @@
...
 """
 
-BackendT = TypeVar("BackendT", bound=Backend)
+PluginName = NewType("PluginName", str)
+"""Name of a plugin's [entry point](https://packaging.python.org/en/latest/specifications/entry-points/).
+
+- Wrap an entry point name to pass it wherever a `backend` is expected, e.g. `PluginName("my-plugin")`.
+- Add it to a signature's `IntoBackend` parameter to advertise plugin support, e.g. `IntoBackend[EagerAllowed | PluginName]`.
+
+See the `narwhals.plugins` module for how plugin authors register entry points
```
FBruzzesi, line 111: So... when I went to add the reference via `[...][]`, I realized that we have no api-reference for plugins 😭 that's a needed follow-up

src/narwhals/_typing.py
FBruzzesi: I ended up stil mentioning `backend` - I hope that's forgivable:

```md
- Wrap an entry point name to pass it wherever a `backend` is expected, e.g. `PluginName("my-plugin")`.
```

src/narwhals/series.py
dangotbanned: > All this should go somewhere! but where exactly? 🧐

We could do either the PR description or the release notes until we find somewhere else?
For release notes I mean if we had a short highlight message in this style (https://github.com/vega/altair/releases#release-v5.5.0)

**Both can work too 😅**

src/narwhals/_typing.py
dangotbanned: > I ended up stil mentioning backend - I hope that's forgivable:

how dare you 😉

tests/plugins_test.py
dangotbanned: > As you also mentioned, for now I will just make `PluginName` exposed in `plugins.py`

https://github.com/narwhals-dev/narwhals/pull/3794#discussion_r3597430820
> Also made me realize that one should import from two modules:

One solution to that is make the symbol(s) available in multiple places

<details><summary>Examples</summary>
<p>

## 1
This module defines and exports 2 protocols.

While I was *implementing* them, having to import `Implementation` from `_utils` was giving me a headache.
So I made it available here too

https://github.com/narwhals-dev/narwhals/blob/fbe56e2dc0c1e633bda580a44eab8bc2b2f0773a/src/narwhals/_plan/compliant/plugins.py#L38

## 2
This package imports and re-exports those 2 protocols, plus some other stuff
https://github.com/narwhals-dev/narwhals/blob/fbe56e2dc0c1e633bda580a44eab8bc2b2f0773a/src/narwhals/_plan/plugins/__init__.py#L26

## 3
I defined `PluginName` here. I don't have a reason, but that is what I did 😂 

https://github.com/narwhals-dev/narwhals/blob/fbe56e2dc0c1e633bda580a44eab8bc2b2f0773a/src/narwhals/_plan/typing.py#L137

I guess the only difference vs what you *had* was that this is `typing.py` vs `_typing.py`.
I probably wouldn't have blinked if the import came from anything public 😄 

</p>
</details> 

It isn't a perfect solution, as you might question which is the canonical one?
But idk, sometimes the world is a hierarchy and sometimes you have things that fit multiple boxes

src/narwhals/_typing.py
dangotbanned: > that's a needed follow-up

Agreed!

For now, something I discovered with (https://github.com/microsoft/pylance-release/issues/6611) - is that it supports lots of stuff (by default) that we either:
- haven't even got configured for `mkdocstrings`
- aren't actually possible

<details><summary>Cross ref to a function in the same module, but isn't API reference</summary>
<p>

`mkdocstrings` has a support for 2 kinds of cross-refs that we haven't got enabled.
IIRC, the syntax was different and it didn't work (yet?) in zensical. 
But it still wouldn't be able to do this wizardry 😄 

image


</p>
</details> 

<details><summary>Cross ref to 3rd-party symbols</summary>
<p>

We can do something similar by adding inventories (if they have one https://github.com/ibis-project/ibis/issues/11723)

But this ~~let's you link~~ links direct to the source code 🤯 

image


</p>
</details> 

### Try these out

```diff
diff --git a/src/narwhals/_typing.py b/src/narwhals/_typing.py
index 189a1543b..c47190a26 100644
--- a/src/narwhals/_typing.py
+++ b/src/narwhals/_typing.py
@@ -110,6 +110,16 @@ PluginName = NewType("PluginName", str)
 
 See the `narwhals.plugins` module for how plugin authors register entry points
 and the contract a wrapped name must satisfy.
+
+
+## Some examples of working cross-refs
+Try these out in your IDE
+
+- [`importlib.metadata.EntryPoint.name`][]
+- [`IntoBackend`][]
+- [`EagerAllowed`][]
+- [`narwhals.typing.LazyAllowed`][]
+- [`narwhals.plugins.Plugin`][]
 """
 
 BackendT = TypeVar("BackendT", bound=Backend | PluginName)

```

image

dangotbanned (approved):
Thanks again @FBruzzesi


choo choo narwhal

src/narwhals/_typing.py
FBruzzesi: Doesn't seem to work with modules though. Will merge without it for now and follow up with proper docs

PR #3797, opened 2026-07-16
perf: remove redundant work in eager when/then, schema resolution, selectors, and module lookups

FBruzzesi:
# Description

Each commit addresses one of the following points, and can be reviewed independently:

* Eager when/then: removed a duplicated block in `EagerNamespace.when_then`
* DuckDB schema: the property populated `_cached_native_schema` but then rebuilt the dict
* pandas-like `Series.to_numpy`: the isna().any() scan only runs when the dtype is in the nullable-dtype maps, instead of unconditionally.
* Use frozenset in selectors (`__sub__`/`__or__`/`__and__`) and in `evaluate_output_names_and_aliases`
* Cache `Version` properties
* Use cached schema for ibis `collect_schema`
* Fix dead guard in the complex pandas group-by fallback (generator is always truthy)
* sqlframe imports: the per-access `import_module("sqlframe.<dialect>.*")` is now cached keyed on dialect.

They should all be low risk with no behavior change. Run some microbenchmarks generated with claude

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [x] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## AI assistance

- [x] AI tools were used. -> for spotting some, and writing micro-benchmarks

### DIFF
--- a/src/narwhals/_compliant/namespace.py
+++ b/src/narwhals/_compliant/namespace.py
@@ -236,10 +236,6 @@ class EagerNamespace(
             if otherwise is None:
                 predicate_s, then_s = align(predicate_s, then_s)
                 result = self._if_then_else(predicate_s.native, then_s.native)
-
-            if otherwise is None:
-                predicate_s, then_s = align(predicate_s, then_s)
-                result = self._if_then_else(predicate_s.native, then_s.native)
             else:
                 otherwise_s = df._evaluate_single_output_expr(otherwise)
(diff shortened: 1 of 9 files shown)


MarcoGorelli (approved):
really nice

thanks @FBruzzesi !

Issue #3567, opened 2026-04-24
bug: fetch_rel_time_zone fails with read-only MotherDuck connection


spark-dataduck:
## Description

`fetch_rel_time_zone()` in `narwhals/_duckdb/utils.py:207` fails when the DuckDB relation is connected to MotherDuck via a **read-only** token. The error occurs because `rel.query()` internally creates a temporary view in the relation's database context, which is a write operation that fails on a read-only database.

This affects **any Narwhals operation that touches the schema** of a DuckDB relation containing a `TIMESTAMPTZ` column — including `collect_schema()`, filtering, selecting, joining, etc.

## Error

```
Invalid Input Error: Failed to create view 'duckdb_settings()':
Cannot execute statement of type "RELATION" on database "my_db" which is attached in read-only mode!
```

## Root Cause

The current implementation at [`utils.py:207-212`](https://github.com/narwhals-dev/narwhals/blob/16842cba56457dcd27e629dafaf38622f5b36084/narwhals/_duckdb/utils.py#L207):

```python
def fetch_rel_time_zone(rel: duckdb.DuckDBPyRelation) -> str:
    result = rel.query(
        "duckdb_settings()", "select value from duckdb_settings() where name = 'TimeZone'"
    ).fetchone()
    assert result is not None
    return result[0]
```

`rel.query(view_name, sql)` works by:
1. Creating a temporary view (named `"duckdb_settings()"`) inside the relation's database — **this is a write operation**
2. Running the SQL query
3. Dropping the view

When the relation belongs to a read-only MotherDuck database, step 1 fails.

Note: calling `conn.sql("SELECT value FROM duckdb_settings() WHERE name = 'TimeZone'")` directly on the connection works fine — the issue is specifically `rel.query()`.

## Call Chain

```
df.collect_schema()                                    # dataframe.py:387
  → self.schema                                        # dataframe.py:235
    → DeferredTimeZone(self.native)                    # dataframe.py:241
    → native_to_narwhals_dtype(...)                    # dataframe.py:243
      → if duckdb_dtype_id == "timestamp with time zone"  # utils.py:195
          → deferred_time_zone.time_zone               # utils.py:196
            → fetch_rel_time_zone(rel)                 # utils.py:146
              → rel.query("duckdb_settings()", ...)    # utils.py:208 ← FAILS
```

## Reproduction

**Environment:** macOS (darwin), Python 3.14.0, DuckDB 1.5.2, Narwhals 2.20.0

```python
[...]

PR #3807, opened 2026-07-20
test: Add free-threading test suite and fix thread-safety issues

FBruzzesi:
# Description

This PR adds test cases for free-threading support and fixes multiple issues, both in the test suite and in the codebase, surfaced by running the suite concurrently under [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel).

1. https://github.com/narwhals-dev/narwhals/commit/4c4578e03fe99f6254ad57c916c175d61a9678dc sets up the free-threading tests:
   * adds `tests/free_threading_test.py`, a dedicated stress-test module that runs narwhals from many threads (cold-cache dispatch, plugin discovery, shared-`Expr.over(...)` push-down, `Enum` deferred categories, shared `LazyFrame` schema, and concurrent `narwhals.sql.table`);
   * adds a second `python-314t` CI step that reruns the suite with `--parallel-threads=4`, a runtime assertion that the GIL is actually off, and a job timeout as a deadlock backstop;
   * registers the `thread_unsafe(reason)` marker so tests that cannot run concurrently opt out;
   * fixes `narwhals.sql`: the module-global DuckDB connection and time-zone object are replaced by a per-call cursor (`CONN.cursor()`), DuckDB's supported per-thread pattern.

2. https://github.com/narwhals-dev/narwhals/commit/273bf7d1e30b267e4333e608c073aefa594e0c26 fixes the issues with reused `pytest.raises` contexts: a single `pytest.raises(...)` object shared across parametrized cases is mutated concurrently under threads.

3. The remaining two commits target DuckDB issues:
   * https://github.com/narwhals-dev/narwhals/commit/69863326dd593c875913ef3d8b2c8cbcf63c89fa: improves DuckDB thread safety by using one connection per thread. Adds a `connection` keyword to `scan_csv`/`scan_parquet` so DuckDB reads can go through a caller-provided `DuckDBPyConnection` instead of the process-global default, and uses one connection per thread in the test `conftest` (relations from different connections cannot be combined, so frames built within one test must share a connection).
   * https://github.com/narwhals-dev/narwhals/commit/e83c8300f446b8dc7a746dd0a7bb180d86184642 : replaces `duckdb.sql(statement)`, which runs on the process-global default connection, with `rel.query(view, sql)`, which executes on the relation's own connection, for the asof-join, unpivot, and concat operations. `top_k` is also rewritten from a `QUALIFY` SQL string to the relational API. Note the trade-off: `rel.query` creates a view, so these operations no longer work on read-only DuckDB databases (#3567), unavoidable until the relational API gains ASOF-join / unpivot support upstream.

---

I left out `warnings.catch_warnings()`, which is also not thread safe, with the risk of filtering warnings.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other


FBruzzesi:
Apologies in advance for the direct pings

---

@ngoldbaum any chance you can skim through this work to assess the usage of `pytest-run-parallel` and the tests in [tests/free_threading_test.py](https://github.com/narwhals-dev/narwhals/pull/3807/changes#diff-8ae59f6a57a5765a72a905a34319e3fa1e6b8d50faae02078a5fdabeebd6d66b)?
@spark-dataduck could you take a look at the DuckDB changes? I am replacing `duckdb.sql(statement)` in favor of `rel.query(view, sql)` in a few places to make duckdb thread safe as suggested in [DuckDB Multiple Python Threads](https://duckdb.org/docs/lts/guides/python/multiple_threads), however as you previous brought up `rel.query` creates a view, so these operations no longer work on read-only DuckDB databases.

---

The PR description hopefully has all the details you need to understand the context 🙏🏼

ngoldbaum:
> @ngoldbaum any chance you can skim through this work to assess the usage of pytest-run-parallel and the tests in [tests/free_threading_test.py](https://github.com/narwhals-dev/narwhals/pull/3807/changes#diff-8ae59f6a57a5765a72a905a34319e3fa1e6b8d50faae02078a5fdabeebd6d66b)?

Sure, although it may be a week or two.

> I left out warnings.catch_warnings(), which is also not thread safe, with the risk of filtering warnings.

If you’re using free-threaded 3.14 or newer, warnings are thread-safe by default. If for some reason you want to run pytest-run-parallel on the GIL-enabled build (why?), you can also enable `thread_inherit_context` and `context_aware_warnings` for the pytest jobs that generate warnings in parallel. Those are enabled by default on the free-threaded build.

.github/workflows/pytest.yml
```diff
@@ -164,5 +166,18 @@ jobs:
...
+      - name: Run pytest (parallel threads)
+        env:
+          # NOTE: Clear PYTEST_ADDOPTS since xdist forks processes,
+          # and the point here is threads within one process
+          PYTEST_ADDOPTS: ""
+        run: |
+          make run-ci \
+          DEPS="--pre --group tests --group plugins --extra pandas --extra pyarrow" \
+          RUN_ONLY="--with pytest-run-parallel" \
+          CMD="pytest tests --parallel-threads=4 --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow"
```
ngoldbaum, line 189: Maybe add a `# TODO: add duckdb/polars when they add free-threaded support` comment

src/narwhals/_duckdb/utils.py
```diff
@@ -337,6 +343,26 @@ def join_column_names(*names: str) -> str:
     return ", ".join(str(col(name)) for name in names)
 
 
+def temporary_view_name() -> str:
+    """Unique name for `DuckDBPyRelation.query`'s `virtual_table_name`.
+
+    `rel.query(view, sql)` registers `rel` as a view named `view` and runs `sql` on
+    `rel`'s own connection (see [relational api]). We prefer it to `duckdb.sql(statement)`,
+    which uses the global default connection and cannot see relations from other
+    connections (see [using connections in parallel pythonprograms]).
```
ngoldbaum, line 371: Maybe it makes sense to have a helper to go along with this one that accepts two queries and verifies that they're from the same connection. I bet users will start seeing errors coming from queries on two different connections because of this.

Also are there memory consequences for this change? The paragraph below indicates that these temporary views are never dropped. If this cost is per-thread, that might add up a lot.

src/narwhals/_duckdb/utils.py
ngoldbaum: should this be using `temporary_view_name()` instead of `"duckdb_settings()"`?

src/narwhals/sql.py
```diff
@@ -22,9 +22,15 @@
...
-)
+
+
+def _cursor() -> duckdb.DuckDBPyConnection:
+    # DuckDB connections must not be used concurrently from multiple threads.
+    # Each cursor shares `CONN`'s catalog but executes independently.
+    # Citing from [Multiple Python Threads](https://duckdb.org/docs/current/guides/python/multiple_threads)
+    # > Each thread must use the `.cursor()` method to create a thread-local
+    # > connection to the same DuckDB file based on the original connection
+    return CONN.cursor()
```
ngoldbaum: I ran an AI model over this PR and it identified a crash due to this change on duckdb < 1.5:

```python
>>> import narwhals as nw
>>> from narwhals.sql import table
>>> t = table("assets", {"a": nw.Int64()})
>>> print(t.to_native().fetchall())
  File "<python-input-0>", line 5, in <module>
    print(t.to_native().fetchall())
          ~~~~~~~~~~~~~~~~~~~~~~^^
_duckdb.ConnectionException: Connection Error: Connection has already been closed
```

The fix is either to define your own thread-local cursor or bump your minimum supported duckdb to 1.5.

```diff
+_local = threading.local()
+
+
 def _cursor() -> duckdb.DuckDBPyConnection:
     # DuckDB connections must not be used concurrently from multiple threads.
     # Each cursor shares `CONN`'s catalog but executes independently.
-    return CONN.cursor()
+    # duckdb 1.4 and older don't keep cursors alive
+    if (cursor := getattr(_local, "cursor", None)) is None:
+        cursor = _local.cursor = CONN.cursor()
+    return cursor
```

Also wow an actual use for the walrus operator!

tests/free_threading_test.py
```diff
@@ -0,0 +1,191 @@
...
+    if DUCKDB_VERSION < (1, 3):
+        pytest.skip()
+    from narwhals.sql import table
+
+    def check(barrier: threading.Barrier) -> None:
+        barrier.wait()
+        name = f"tbl_{uuid.uuid4().hex}"
+        result = table(name, {"a": nw.Int64(), "b": nw.String()})
+        assert result.collect_schema() == {"a": nw.Int64(), "b": nw.String()}
+        assert name in result.to_sql()
```
ngoldbaum, line 322:
```suggestion
        assert name in result.to_sql()
        assert result.to_native().fetchall() == []
```

This would have caught the stale connection issue I commented about above

ngoldbaum:
This PR doesn't touch the docs at all. Should it? The thread-local connection behavior might be surprising.

Also I'm not sure if you're planning to do more free-threaded support work, but I also usually suggest adding explicit docs somewhere on thread safety of the library. Focus on mutable global state and mutable objects. 

Is the mutable global state in the library thread-safe? Mutable global state is things like configuration state, caches initialized at runtime, and global mutable singletons.

Are instances of mutable types defined by the library safe to share between threads without external synchronization (e.g. a lock) or coordination (e.g. an algorithm that is safe by construction, like filling disjoint parts of an array with a thread pool)? Note that, for example, mutating NumPy arrays is not thread-safe, so Pandas backed by NumPy is likely susceptible to races for any workflows that tries to write to a data frame. Pandas' arrow backend uses CoW semantics so it doesn't have that problem.

My AI model also found a decent number of additional issues and wrote some scripts to trigger them: 

**`repro_groupby_agg_race.py`** — run on 3.14t with pandas; expect `AssertionError: [10.0, 30.0]`. It passes on the GIL-enabled build.

```python
from concurrent.futures import ThreadPoolExecutor

import pandas as pd

import narwhals as nw

df = nw.from_native(pd.DataFrame({"g": [1, 1, 2, 2], "i": [2, 1, 2, 1], "v": [10.0, 20.0, 30.0, 40.0]}), eager_only=True)
gb = df.group_by("g")

def worker(i):
    for _ in range(300):
        expr, expected = (nw.col("v").sum(), [30.0, 70.0]) if i % 2 else (nw.col("v").first(order_by="i"), [20.0, 40.0])
        assert (got := gb.agg(expr).sort("g")["v"].to_list()) == expected, got

with ThreadPoolExecutor(8) as tpe:
    [f.result() for f in [tpe.submit(worker, i) for i in range(8)]]
```

**`repro_shared_connection_schema.py`** — GIL 3.14 with duckdb; expect `InvalidInputException: ... unsuccessful or closed pending query result`

```python
from concurrent.futures import ThreadPoolExecutor

import duckdb

import narwhals as nw

con = duckdb.connect()
frames = [nw.from_native(con.sql("select timestamptz '2024-01-01' as t")) for _ in range(8)]

def worker(lf):
    for _ in range(50):
        lf.collect_schema()

with ThreadPoolExecutor(8) as tpe:
    [f.result() for f in [tpe.submit(worker, lf) for lf in frames]]
```

ngoldbaum:
> **`repro_groupby_agg_race.py`** — run on 3.14t with pandas; expect `AssertionError: [10.0, 30.0]`. It passes on the GIL-enabled build.

This one triggers on the GIL-enabled build as well if you add `sys.setwitchinterval(1e-7)` to the top of the repro script, so it's not free-threaded specific per-se, it's just very unlikely with the default swtich interval.

The issue is how this cache gets filled if the `AggExpr` is shared:

https://github.com/narwhals-dev/narwhals/blob/7880b36c45051dd6dea3c1b878bdc2f913d8efec/src/narwhals/_pandas_like/group_by.py#L299

I think there can be a race to set that attribute. I'm not totally sure why the attribute is being stashed there instead of being passed along.

FBruzzesi:
Thanks for all your inputs @ngoldbaum - those are really really useful insights 🙏🏼 
I will try to address all your findings and add a dedicated documentation page for what we _can_ promise from the narwhals layer. Some questions regarding DuckDB, I don't have an answer just yet, I will do some investigation, but thank you for raising the questions in the first place

src/narwhals/sql.py
FBruzzesi: Thanks, added

src/narwhals/_duckdb/utils.py
FBruzzesi: I assume that the fixed name is what we want, a unique name per call would instead add one permanent catalog entry per `collect_schema()` on a time-zone-aware frame. I added a `with _TIME_ZONE_LOCK` so that the relation is consumed by `.fetchone()` inside the lock.

I would not expect _that_ to be the expensive operation in a query.

---

On a second thought, I think renaming it would at least be less confusing (e.g. `__narwhals_time_zone__`), since the current name reads like it's referring to the table function.

src/narwhals/_duckdb/utils.py
FBruzzesi: You're right that the views are never dropped: I couldn't find a way we can just drop the view once the relation is built.

Regarding the _same-connection helper_, similarly I couldn't find _something_ exposed by a `DuckDBPyRelation` that can be used for such task. What _we_ can do is raise a better exception by wrapping the exception into our own `catch_duckdb_exception` and explain the issue in full detail.

In both cases, I will run the questions by an agent and/or ask in a duckdb forum/github discussion to see if there is something I couldn't find from `dir(<various duckdb objects>)` and their docs 😅

tests/free_threading_test.py
```diff
@@ -0,0 +1,335 @@
...
+def test_gil_stays_disabled_on_free_threaded_build() -> None:  # pragma: no cover
+    if not sysconfig.get_config_var("Py_GIL_DISABLED"):
+        pytest.skip("not a free-threaded build")
+    # NOTE: Only reached on a free-threaded build, never on the GIL-enabled coverage jobs.
+    # Because of this, the entire function is flagged as "pragma: no cover"
+    is_gil_enabled = getattr(sys, "_is_gil_enabled", lambda: True)
+    assert not is_gil_enabled(), (
+        "The GIL was re-enabled, likely by importing an extension module "
+        "without free-threading support."
+    )
```
ngoldbaum: FWIW pytest-run-parallel includes this check out-of-the-box

ngoldbaum (commented):
I only did a brief scan but the new docs look nice and this look like it's heading in the right direction!

tests/free_threading_test.py
FBruzzesi: Thanks for the tip. Related to pytest-run-parallel behavior, should I remove the PYTHON_GIL env variable in the job? Does the plugin fail the run if the GIL gets re-enabled at runtime (naming the module responsible)?

src/narwhals/_duckdb/utils.py
FBruzzesi: Opened two discussions in duckdb-python repository: https://github.com/duckdb/duckdb-python/discussions/580 and https://github.com/duckdb/duckdb-python/discussions/581

Issue #3761, opened 2026-07-08
[Enh]: Add functionality to access categorical codes


cakedev0:
Hi, I'm a scikit-learn contributor and I'm currently working on this PR https://github.com/scikit-learn/scikit-learn/pull/34393.

I'm trying to use Narhwals as much as I can for this work, but I currently can't rely on Narwhals fully because of some limitations and behavior differences around categories.

## My need

In scikit-learn encoders, we want to avoid materializing categorical values when
the backend already stores integer categorical codes. We need:

- the full category list in category-code order;
- the per-row integer codes;
- a mask for missing values.

## Current state of things

`Series.cat.get_categories()` provides the first part, but I could not find a public Narwhals API for the codes. The only public categorical series method I found is `Series.cat.get_categories()`.

For Polars `Categorical` / `Enum` series, casting the Narwhals series to `nw.UInt32` appears to expose the physical category codes.

For pandas categoricals, the same operation either fails or returns the category values instead of the categorical codes.

I am not sure whether `Series.cast(nw.UInt32)` is intended to expose categorical
codes. If it is, pandas behavior looks inconsistent with Polars. If it is not,
would it make sense to expose an explicit categorical-code API, for instance
`Series.cat.get_codes()` or `Series.cat.to_physical()`?

## Reproducer
```python
import narwhals.stable.v2 as nw
import pandas as pd
import polars as pl

pd_native = pd.Series(pd.Categorical([2, 1], categories=[1, 2, 3]))
pd_s = nw.from_native(pd_native, allow_series=True)

print("pandas categories:", pd_s.cat.get_categories().to_numpy())
# prints: pandas categories: [1 2 3]
print("pandas cast UInt32:", pd_s.cast(nw.UInt32).to_numpy())
# prints: pandas cast UInt32: [2 1]
print("pandas native codes:", pd_native.cat.codes.to_numpy())
# prints: pandas native codes: [1 0]

pl_native = pl.Series("x", ["b", "a"], dtype=pl.Enum(["b", "a", "c"]))
pl_s = nw.from_native(pl_native, allow_series=True)

print("polars categories:", pl_s.cat.get_categories().to_numpy())
# prints: polars categories: ['b' 'a' 'c']
print("polars cast UInt32:", pl_s.cast(nw.UInt32).to_numpy())
# prints: polars cast UInt32: [0 1]
```

## Related issues / prior context

Related categorical/cast discussions AI found:

[...]

PR #3809, opened 2026-07-20
feat: add Series.factorize

camriddell:
factorize produces unique values and value mappings back to the locations of those values. Inspired by [pandas.factorize](https://pandas.pydata.org/docs/reference/api/pandas.factorize.html)

## Implementation Details

**Value Semantics**
- `uniques` is the set of distinct non-null values.
- `codes[i]` is the index into uniques for `values[i]`.
- Any value not present in uniques maps to -1 (currently only nulls).

Missing Value Semantics:
- Nulls are dropped from the unique factorized values and the corresponding mapping code is `-1`.
- NaNs are preserved (treated as actual values). Except for within pandas where it cannot distinguish between Null/NaN.

## To Do/Discuss

Pure Narwhals implementation: is it tolerable to only have this implemented at the `narwhals.functions` level and NOT within each individual backend? This function can be composed purely of pieces that already exist

Missing Values:
- Implement the `use_na_sentinel` from [pandas.factorize](https://pandas.pydata.org/docs/reference/api/pandas.factorize.html)?
  -  for this route, I'd prefer to add a different keyword argument to convey "treat Null as an ordinary value" (e.g. `drop_nulls` or `null_as_value`)
- Is `-1` an appropriate "missing" mapping value? Currently when one factorizes a Series with a Null (or NaN in pandas), the Null is automatically dropped from the unique value set and the corresponding mapping code becomes `-1`.

# Description

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3761

## AI assistance

- [x] No AI tools were used for this PR.
- [ ] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes
    
    ```bash
    ```


cakedev0:
Will this call `pd.Series.factorize` for pandas Series? Asking because it's much much faster than unique + replace_strict (almost 10x fast for categorical dtype).

> Is -1 an appropriate "missing" mapping value?

Given "codes" are supposed to be non-negative integers, I'd say this is good. Or at least, it's much better than other options that comes to my mind: NaN, MaxInt32, n_uniques + 1, ...?

src/narwhals/functions.py
```diff
@@ -2021,3 +2023,55 @@ def list_(*exprs: IntoExpr | Sequence[IntoExpr]) -> Expr:
...
+
+    Arguments:
+        values: A series to factorize.
+        sort: Whether to sort the unique values before assigning codes.
+
+    Returns:
+        - codes: An integer series where each value represents the index
+          of the corresponding value in `uniques`. Null values are encoded
+          as -1.
+        - uniques: A series containing the unique non-null values.
```
FBruzzesi: I don't think this is rendering as expected:

<details> <summary> See screenshot </summary> 

Image

</details>

```suggestion
    Returns:
        codes: An integer series where each value represents the index of the
            corresponding value in `uniques`. Null values are encoded as -1.
        uniques: A series containing the unique non-null values.
```

<details> <summary> See screenshot after </summary> 

Image

</details>

src/narwhals/functions.py
```diff
@@ -2021,3 +2023,55 @@ def list_(*exprs: IntoExpr | Sequence[IntoExpr]) -> Expr:
...
+        └─────┘
+        >>> uniques
+        ┌─────┐
+        |  a  |
+        |-----|
+        | str |
+        |-----|
+        | b   |
+        | a   |
+        └─────┘
```
FBruzzesi: This example shows everything that is confusing to me if we return a `tuple[Series, Series]` versus the dictionary mapping `value: code` directly.

Given the `uniques` order, I would expect `{"b": 0, "a": 1}`, however this is not the case as you can see from `codes`. In other words, my issue is not being able to recover the mapping in an ergonomic way from the returned series.

I would rather return the tuple of (`codes`, `mapping`). The latter could be either "code to value" or "value to code" - not sure which one is preferable

FBruzzesi (commented):
Thanks @camriddell - I left a couple of inline comments. 

On top of those:

1. I think the concern from @cakedev0 on performances is totally valid
2. Is there a reason to not expose the feature in `stable.v1`?
3. I didn't check the tests 🙈

src/narwhals/functions.py
cakedev0: This example is not running (see the red CI job). And I checked locally, a and b are swapped in `uniques`.

Personally, I think returning two series such that `uniques[codes] == values` (when no missing are present) is convenient.

cakedev0: The corrected example:

```
        >>> import polars as pl
        >>> import narwhals as nw
        >>> df = pl.DataFrame({"a": ["a", "b", "a", None]})
        >>> nw_df = nw.from_native(df)
        >>> codes, uniques = nw.factorize(nw_df["a"], sort=True)
        >>> codes
        ┌─────────────────┐
        | Narwhals Series |
        |-----------------|
        |shape: (4,)      |
        |Series: 'a' [i32]|
        |[                |
        |        0        |
        |        1        |
        |        0        |
        |        -1       |
        |]                |
        └─────────────────┘
        >>> uniques
        ┌─────────────────┐
        | Narwhals Series |
        |-----------------|
        |shape: (2,)      |
        |Series: 'a' [str]|
        |[                |
        |        "a"      |
        |        "b"      |
        |]                |
        └─────────────────┘
```

camriddell: @FBruzzesi, @cakedev0 this was a copy/paste typo on my end to build the example. The above snippet is the correct output.

returning a dict forces us to pull values out of their respective backends and all of the inputs to be hashable. This form follows the style that pandas returns so it should also be familiar.

camriddell:
- Moves `nw.factorize` to a Series method: `nw.Series.factorize`
- Moves the function logic to each respective backend, so that backend specific optimizations can be applied (e.g. use `pandas.factorize` for pandas; possible future Polars https://github.com/pola-rs/polars/pull/27045)

src/narwhals/_arrow/series.py
```diff
@@ -1038,6 +1038,19 @@ def hist_from_bin_count(
...
+        uniques = self.unique().drop_nulls()
+        if sort:
+            uniques = uniques.sort(descending=False, nulls_last=True)
+        codes = self.replace_strict(
+            old=uniques.to_list(),
+            new=[*range(len(uniques))],
+            default=-1,
+            return_dtype=dtypes.Int32(),
+        )
+        return codes, uniques
```
FBruzzesi: [pyarrow.compute.dictionary_encode](https://arrow.apache.org/docs/python/generated/pyarrow.compute.dictionary_encode.html) can probably help

src/narwhals/_pandas_like/series.py
```diff
@@ -1163,6 +1163,15 @@ def is_native_dtype_pyarrow(self, native_dtype: Any) -> bool:
         impl = self._implementation
         return get_dtype_backend(native_dtype, implementation=impl) == "pyarrow"
 
+    def factorize(self, *, sort: bool = False) -> tuple[Self, Self]:
+        pdx = self.__native_namespace__()
+        name = self.native.name
+        codes, uniques = self.native.factorize(sort=sort)
+        return (
+            self._with_native(pdx.Series(codes, name=name)),
+            self._with_native(pdx.Series(uniques, name=name)),
```
FBruzzesi: I am not the biggest fan of using the original name for both results. It's on the nose but: what about `"codes"` and `"uniques"` or with the `name` as prefix: `f"{name}_codes"` and `f"{name}_uniques"?`

src/narwhals/_polars/series.py
```diff
@@ -670,6 +670,18 @@ def last(self) -> PythonLiteral:
     def any_value(self, *, ignore_nulls: bool) -> PythonLiteral:
         return self.drop_nulls().first() if ignore_nulls else self.first()
 
+    def factorize(self, *, sort: bool = False) -> tuple[Self, Self]:
+        uniques = self.unique().drop_nulls()
+        if sort:
+            uniques = uniques.sort(descending=False, nulls_last=True)
+        codes = self.native.replace_strict(
+            old=uniques.to_list(),
+            new=[*range(len(uniques))],
```
FBruzzesi: Hopefully polars can take care of some fast(er) paths:


```suggestion
            old=uniques,
            new=range(len(uniques)),
```

FBruzzesi (commented):
Thanks again @camriddell 🙏🏼 

Left a few inline suggestions. On top of those:

1. I like the possibility in pyarrow to treat null as any other value:
	* `dictionary_encode(array, null_encoding='mask')` -> keeps null's in the indices
	* `dictionary_encode(array, null_encoding='encode')` -> encodes null values as any other value, not with a sentinel value.
	I think this is preferable, a user can always control the sentinel value to use by `null_encoding='mask'` + `fill_null(<whatever value they want>)`.
	On the variable name (`null_encoding`), and its values (`"mask"` and `"encode"`), we can discuss better options
2. As of https://github.com/narwhals-dev/narwhals/pull/3809#discussion_r3628721772, I understand the appeal of being able to do `uniques[codes] == values`. However, since this is greenfield and _very unlikely_ that polars will have the same method, I would like to have some creative freedom and propose a way to satisfy both returning uniques _and_ a mapping.
	We can return a tuple of (codes, uniques), however, we can do so in a `NamedTuple`, so that we can enrich its behavior (draft version):

	```py
	from narwhals import Series
	from typing import Any, NamedTuple, Generic
	from collections.abc import Mapping
	
	from narwhals.typing import IntoSeriesT
	
	
	class Encoded(NamedTuple, Generic[IntoSeriesT]):
	    """Result of `factorize`. Unpacks as `(codes, uniques)` like pandas."""
	 
	    codes: Series[IntoSeriesT]
	    uniques: Series[IntoSeriesT]
	 
	    @property
	    def mapping(self) -> Mapping[Any, int]:
	        """Forward map as a joinable ``(value, code)`` frame; works for any dtype."""
	        name = self.uniques.name
	        return dict(
	            self.uniques
	            .to_frame()
	            .with_row_index("code")
	            .select(name, "code")
	            .iter_rows()
	        )
	```
	
	In this way, one can both:
	
	```py
	codes, uniques = series.factorize()
	
	encoded = series.factorize()
	encoded.mapping
	```
	
	Let me know your thoughts - In general I am not the biggest fan of NamedTuple's for various reasons, yet this seem like a good case to have one ;)

src/narwhals/_pandas_like/series.py
camriddell: Yeah this was bothering me as well. I'm in favor of the plain "codes" and "uniques".

camriddell:
@FBruzzesi I think that wraps up all of your requests.

Seems like I'll need to add a back up path for an old pandas version, then we should hopefully be squared away.

src/narwhals/series.py
```diff
@@ -2946,3 +3009,26 @@ def list(self) -> SeriesListNamespace[Self]:
...
+    @property
+    def mapping(self) -> Mapping[Any, int]:
+        """Forward map as a joinable ``(value, code)`` frame; works for any dtype."""
+        name = self.uniques.name
+        return dict(
+            self.uniques.to_frame()
+            .with_row_index("code")
+            .select(name, "code")
+            .iter_rows()
+        )
```
FBruzzesi: I am basically self-reviewing this bit - I ended up suggesting:

* A wrong docstring (from another version I was writing)
* possibly overlapping name ("code")

```suggestion
    @property
    def mapping(self) -> Mapping[Any, int]:
        """Forward map `value -> code` as a dict (handy for small cardinality)."""
        uniques = self.uniques
        return dict(zip(uniques, range(len(self.uniques)), strict=True))
```

src/narwhals/series.py
```diff
@@ -2905,6 +2906,68 @@ def is_close(
...
+        The integer codes are index locations that map the unique values back to their
+        positions within the original array.
+
+        Arguments:
+            null_as_value: Whether to treat null as a regular value. When False,
+                nulls are removed from the returned unique values and the code -1 is
+                used to indicate the location of null values in the original array.
+                When True, nulls are preserved in the returned unique values and a
+                positive integer is used to indicate their location in the original
+                array.
```
FBruzzesi: I would still hold against having a sentinel value unless provided. I understand the ergonomic, but in polars spirit, I think there should be a way to preserve null values. `null_as_value` is probably not a good name for it, but we could _enrich_ what pyarrow is doing:

* `"mask"` -> preserve null values -> I would like this to be the default behavior
* `"encode"` -> same as `null_as_value = True` i.e. null's get encoded as any other value
* `<value>` -> encode with such sentinel value, equivalent to 
 	```py
	codes, _ = series.factorize("mask")
	codes.fill_null(<value)>
	```

---

I am curious to know how others feel about this

src/narwhals/series.py
```diff
@@ -2905,6 +2906,68 @@ def is_close(
...
+            sort: Whether to sort the unique values before assigning codes.
+
+        Returns:
+            codes: An integer series where each value represents the index
+                of the corresponding value in `uniques`. Null values are encoded
+                as -1.
```
FBruzzesi: Out of sync with the new behaviour

FBruzzesi (commented):
Thanks @camriddell - I did a quick iteration for the top-level behavior, I will wait for the CI to be green for all the nitty gritty details

src/narwhals/series.py
camriddell: Beyond ergonomics, preserving the null value in the returned codes breaks base pandas as the returned NaN upcasts the integer codes to become floats. I'd prefer not to have the default case create a surprise for the backend where this function is likely to be used most often.

I am fine with the remaining semantics, just would advocate to make the `"encode"` behavior the default.

src/narwhals/series.py
```diff
@@ -2905,6 +2906,68 @@ def is_close(
         result = result.rename(orig_name) if name_is_none else result
         return cast("Self", result)
 
+    def factorize(
```
dangotbanned, line 2916: Could we please mark this as unstable?

src/narwhals/series.py
```diff
@@ -2946,3 +3009,26 @@ def list(self) -> SeriesListNamespace[Self]:
     @property
     def struct(self) -> SeriesStructNamespace[Self]:
         return SeriesStructNamespace(self)
+
+
+@dataclass(frozen=True)
+class Encoded(Generic[IntoSeriesT]):
```
dangotbanned, line 3048: Could this also be marked as unstable and/or private please?

It is mostly an implementation detail

tests/v2_test.py
```diff
@@ -586,3 +586,33 @@ def test_schema_from_generator() -> None:
     )
     assert schema == nw_v2.Schema({"a": nw_v2.Int64(), "b": nw_v2.String()})
     assert schema._version is Version.V2
+
+
+@pytest.mark.parametrize(
```
dangotbanned: Per the other comments, if `factorize` is unstable, would we still have these v1, v2 tests?

src/narwhals/_arrow/series.py
```diff
@@ -1038,6 +1038,57 @@ def hist_from_bin_count(
...
+            )
+
+        sorted_uniques = pc.take(uniques, pc.sort_indices(uniques))
+        new_mapping = pc.index_in(uniques, value_set=sorted_uniques)
+        new_codes = pc.take(new_mapping, codes)
+
+        return (
+            self._with_native(pc.fill_null(new_codes, pa.scalar(-1))),
+            self._with_native(sorted_uniques),
+        )
```
dangotbanned: As somebody who has battled against `pyarrow`'s frustrating API a lot - I feel like this is doing too many steps 🤔

I'm not 100% sure what is going on here, so I don't a concrete suggestion.

*But* I do have some code that may be helpful to look at. If only inspire an *oh that's how these bits could fit together* moment 😉 

<details><summary>Show (maybe related) pyarrow functions</summary>
<p>


### Definitions
- https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/functions/_categorical.py
- https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/functions/_multiplex.py
- https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/functions/_sort.py

### Some usage
https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/dataframe.py#L301-L313

https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/functions/_aggregation.py#L152-L159

https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/series.py#L183-L191

https://github.com/narwhals-dev/narwhals/blob/c179a78e4f3d6c95530aea10a7b2f917ec54ac21/src/narwhals/_plan/arrow/functions/_lists.py#L352-L365

</p>
</details>

src/narwhals/_arrow/series.py
```diff
@@ -1038,6 +1038,57 @@ def hist_from_bin_count(
...
+                    pa.repeat(-1, len(self.native)),
+                    pa.nulls(0, type=self.native.type),
+                )
+
+            return (self._with_native(codes.cast(pa.int32())), self._with_native(uniques))
+
+        native = self.native
+        if pa.types.is_dictionary(native.type):
+            # re-encode if already dict encoded; can't be certain how the original dict encoding was done
+            native = native.cast(native.type.value_type)
```
dangotbanned, line 1080: Can't you tell by the presence of nulls?

IIRC, one version keeps nulls in the indices [^1].
If there aren't any nulls, then I would assume the kind of encoding doesn't matter

[^1]: It could also be the other property

src/narwhals/_polars/series.py
```diff
@@ -670,6 +670,27 @@ def last(self) -> PythonLiteral:
     def any_value(self, *, ignore_nulls: bool) -> PythonLiteral:
         return self.drop_nulls().first() if ignore_nulls else self.first()
 
+    def factorize(
+        self, *, null_as_value: bool = False, sort: bool = False
+    ) -> tuple[Self, Self]:
```
dangotbanned: Is [`polars.Series.to_physical`](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.to_physical.html) no good for this?
The example says:

> Replicating the pandas [pd.Series.factorize](https://pandas.pydata.org/docs/reference/api/pandas.Series.factorize.html) method

src/narwhals/_polars/series.py
```diff
@@ -670,6 +670,27 @@ def last(self) -> PythonLiteral:
...
+        uniques = self.unique() if null_as_value else self.unique().drop_nulls()
+        if sort:
+            uniques = uniques.sort(descending=False, nulls_last=True)
+
+        if null_as_value:
+            codes = self.native.replace_strict(
+                old=uniques.native, new=range(len(uniques)), return_dtype=pl.Int32()
+            )
```
dangotbanned: Have you considered starting this with a `self.native.to_frame()` and then doing all the steps with expressions?

<details><summary>Show pattern</summary>
<p>

This is based on the [magic polars does](https://github.com/pola-rs/polars/blob/8122e9326d52208593ee736b7efb93821fe95444/py-polars/src/polars/series/utils.py#L26-L109) for most `Series` methods:

https://github.com/narwhals-dev/narwhals/blob/05ce363b2c8106015f540a51cc9e0b8d332e2b8a/src/narwhals/_polars/series.py#L808-L811

Chaining `Series` methods has to do all these steps on every intermediate call. And this is just the python part 😅 

</p>
</details> 

For example, what if `native` was this?

```py
import polars as pl

N = 10_000_000
N_UNIQUE = 100

native = (
    pl.select(big_boi=pl.int_range(N), eager=False)
    .gather((pl.col.big_boi % N_UNIQUE).shuffle(1))
    .collect()
    .to_series()
)

native
```

<details><summary>Show output</summary>
<p>

```
shape: (10_000_000,)
Series: 'big_boi' [i64]
[
	28
	88
	5
	1
	71
	…
	58
	1
	23
	92
	27
]
```

</p>
</details> 

Would there be a benefit to not materializing until as late as possible?

```py
uniques = native.unique().drop_nulls().sort(nulls_last=True)
codes = native.replace_strict(uniques, range(len(uniques)), return_dtype=pl.Int32()).cast(
    pl.Int32()
)
```

dangotbanned (commented):
Thanks @camriddell, hope some of this can be helpful

src/narwhals/series.py
dangotbanned: > Beyond ergonomics, preserving the null value in the returned codes breaks ... as the returned NaN upcasts the integer codes to become floats.

Is it not possible to use [`pd.Int64Dtype`](https://pandas.pydata.org/docs/user_guide/integer_na.html)?

> I'd prefer not to have the default case create a surprise for the backend where this function is likely to be used most often.

I understand this concern, but I agree with @FBruzzesi that the default behavior should *feel like* polars.

## Third option
What if we held back on `nw.Series.factorize` *for now*, and instead introduced *how to do it*  into our docs?

This feels pretty related to [discord thread](https://discord.com/channels/1235257048170762310/1272835535190364283/1525158517680377979) and would slide in perfectly into a whole section about transitioning from `pd.Categorical` 🙂

@camriddell, correct me if I'm wrong, but that feels like something you'd have a lot to speak on 😅

src/narwhals/_arrow/series.py
camriddell: I originally had a version that dispatched based on the presense of nulls, but it felt a bit more like an implementation detail that produced a surprising behavior. Was planning to open an upstream issue.

```python
>>> import pyarrow as pa
>>> arr = pa.array([*'ab', None])
>>> arr.dictionary_encode("encode").is_null()
<pyarrow.lib.BooleanArray object at 0x7f76b91b56c0>
[
  false,
  false,
  false
]
>>> arr.dictionary_encode("mask").is_null()
<pyarrow.lib.BooleanArray object at 0x7f76b91b5c00>
[
  false,
  false,
  true
]
```

src/narwhals/_polars/series.py
camriddell: `factorize` needs to guarantee that the returned unique values and the codes align. Since `to_physical` requires a Categorical/Enum input, the global categorical cache makes the physical codes surprising. However, we could reasonably make a fast path for an `Enum` type:

<details>
<summary>pl.Categorical</summary>

```python
>>> s1 = pl.Series("X", [*"abc"]).cast(pl.Categorical)
>>> s2 = pl.Series("X", [*"abd"]).cast(pl.Categorical)
>>> s1.to_physical()
shape: (3,)
Series: 'X' [u32]
[
        0
        1
        2
]
>>> s2.to_physical()
shape: (3,)
Series: 'X' [u32]
[
        0
        1
        3
]

```

</details>

<details>
<summary>pl.Enum</summary>

```python
>>> import polars as pl
>>> e1 = pl.Enum(['a', 'b', 'c'])
>>> e2 = pl.Enum(['a', 'b', 'd'])
>>> s1 = pl.Series("X", [*"abc"]).cast(e1)
>>> s2 = pl.Series("X", [*"abd"]).cast(e2)
>>> s1.to_physical()
shape: (3,)
Series: 'X' [u8]
[
        0
        1
        2
]
>>> s2.to_physical()
shape: (3,)
Series: 'X' [u8]
[
        0
        1
        2
]

```

</details>

src/narwhals/_polars/series.py
camriddell: I had intentionally re-used our `PolarsSeries.sort` as it implements a fallback for a previous version of Polars where the `nulls_last` argument did not exist and I didn't want to duplicate that code.

> Would there be a benefit to not materializing until as late as possible?

Probably, but "as late as possible" comes pretty early as we must materialize before calling replace (otherwise `range(len(uniques))` does not work). So pretty much by using expressions we can push the `unique()` and possibly `drop_nulls` & `sort` into the same operation.

src/narwhals/series.py
camriddell: > Is it not possible to use [pd.Int64Dtype](https://pandas.pydata.org/docs/user_guide/integer_na.html)?

@dangotbanned yeah we can return an `Int64` dtype. Originally I had thought that we supported a pandas version that didn't have nullable types but I am no longer certain that is the case.

> What if we held back on nw.Series.factorize for now, and instead introduced how to do it into our docs?

I am biased as I have already sunk a few hours into this, but I would like to see it land. This is also a fairly common operation for our array-based libraries downstream users, and the backend implementations are all quite different from one another.

@FBruzzesi & @dangotbanned how do you feel about this as an API? I would love it if the SENTINEL policy was its own struct, so one could embed the sentinel value inside of it but we're using the wrong language for that.

```python
class NullEncoding(StrEnum):
    PRESERVE = auto() # keep the Nulls in the resultant codes
    ENCODE = auto()   # encode Nulls as one would like any other value
    SENTINEL = auto() # replace Nulls with a particular `sentinel` value

class Series:
    def factorize(
        self, *, sort: bool = False, null_policy: NullEncoding = NullEncoding.PRESERVE, sentinel: Any | NoDefault = NO_DEFAULT,
    ) -> Encoded[IntoSeriesT]:
        if null_policy is NullEncoding.SENTINEL:
            if sentinel is NO_DEFAULT:
                msg = "Must supply `sentinel` when null_policy=NullEncoding.SENTINEL"
                raise TypeError(msg)
        elif sentinel is not NO_DEFAULT:
            msg = f"Argument `sentinel` is ignored when null_policy={null_policy}"
            raise TypeError(msg)

        ...
```

FBruzzesi: Hey @camriddell sorry for the late feedback, I completely lost this thread!

I like having `null_policy` (and such name) as well as using `sentinel`. I would disagree on creating a custom `Enum` to expose to the user. I think `Literal[...]` is just fine.

On the larger picture of having the feature vs documenting it, I would be in favor of having it with the unstable tag. As Cam mentioned, this topic came up quite a lot recently but not only, and there is real need for it.

On the Int64 dtype for pandas: I am quite ok with that, keep in mind that one day pyarrow might be a default/requirement if I remember correctly. I am sure Marco knows much more about it.

src/narwhals/_arrow/series.py
camriddell: I tested `pc.dictionary_encoding` vs `uniques -> replace` and found that the former was faster at large dataset sizes.

<details>
<summary>Timing code</summary>

```python
import pyarrow as pa
import pyarrow.compute as pc

def factorize_via_replace(arr, *, sort=False):
    uniques = pc.unique(arr)
    if sort:
        uniques = pc.take(uniques, pc.sort_indices(uniques))

    return pc.index_in(arr, value_set=uniques), uniques

def factorize_via_dictionary(arr, sort=False):
    encoded = pc.dictionary_encode(arr, null_encoding='mask').combine_chunks()
    uniques = encoded.dictionary
    codes = pa.chunked_array(encoded.indices)

    if not sort:
        return codes, uniques

    sorted_uniques = pc.take(uniques, pc.sort_indices(uniques))
    new_mapping = pc.index_in(uniques, value_set=sorted_uniques)
    new_codes = pc.take(new_mapping, codes)

    return new_codes, sorted_uniques


import numpy as np
from numpy.random import default_rng
from string import ascii_lowercase
from timeit import timeit
from itertools import product

rng = default_rng(0)
n = 10

for sz, n_chunks, sort in product([10_000, 1_000_000], [5, 100, 10_000], [True, False]):
    n_chunks = min(n_chunks, sz)

    arr = rng.choice([*ascii_lowercase], size=(sz, 4)).view('<U4').ravel()
    arr = pa.chunked_array(np.array_split(arr, n_chunks), type=pa.string())

    c1, u1 = factorize_via_replace(arr, sort=sort)
    c2, u2 = factorize_via_dictionary(arr, sort=sort)
    assert u1.equals(u2)
    assert c1.equals(c2)

    print(f'N={sz:<10,}| Chunks={n_chunks:<10,}| sort={sort!r:<5}| Unique={len(u1):,}')
    print(f"{timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = :.6f}")
    print(f"{timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = :.6f}")
    print()
```

</details>

<details>
<summary>Timing results, stepping through dictionary_encoding is *generally* faster, but not in every scenario. ~2× faster for the largest array size (1M values, 10k chunks) when `sort=False`.</summary>

```
N=10,000    | Chunks=5         | sort=True | Unique=9,874
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.021563
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.028813

N=10,000    | Chunks=5         | sort=False| Unique=9,910
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.007005
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.003449

N=10,000    | Chunks=100       | sort=True | Unique=9,894
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.028647
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.029794

N=10,000    | Chunks=100       | sort=False| Unique=9,893
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.015438
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.006601

N=10,000    | Chunks=10,000    | sort=True | Unique=9,905
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.067098
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.241810

N=10,000    | Chunks=10,000    | sort=False| Unique=9,894
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 0.049481
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 0.222008

N=1,000,000 | Chunks=5         | sort=True | Unique=405,646
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 3.767795
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 3.885547

N=1,000,000 | Chunks=5         | sort=False| Unique=405,821
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 2.810773
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 1.740194

N=1,000,000 | Chunks=100       | sort=True | Unique=406,059
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 3.903205
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 3.496898

N=1,000,000 | Chunks=100       | sort=False| Unique=405,580
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 2.629999
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 1.398467

N=1,000,000 | Chunks=10,000    | sort=True | Unique=405,871
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 3.914798
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 3.816666

N=1,000,000 | Chunks=10,000    | sort=False| Unique=405,657
timeit(lambda: factorize_via_replace(arr, sort=sort), number=n) = 2.829571
timeit(lambda: factorize_via_dictionary(arr, sort=sort), number=n) = 1.563878
```

</summary>

camriddell: (do note that in my timing I used `combine_chunks()` as that ended up being MUCH faster than the `unify_dictionaries()` approach. So I'll need to swap that out regardless.

src/narwhals/_arrow/series.py
```diff
@@ -1043,6 +1045,58 @@ def hist_from_bin_count(
...
+        )
+        encoded: Incomplete = pc.dictionary_encode(
+            native, null_encoding=null_encoding
+        ).combine_chunks()
+        uniques, codes = encoded.dictionary, encoded.indices
+
+        if sort:
+            uniques = pc.take(uniques, pc.sort_indices(uniques))
+            mapping = pc.index_in(uniques, value_set=uniques)
+            codes = pc.take(mapping, codes)
```
FBruzzesi, line 1093: `uniques` is re-assigned to itself, so `index_in` computes an array against itself. That is always the identity permutation, so codes is never remapped while uniques has moved.

```suggestion
            sorted_uniques = pc.take(uniques, pc.sort_indices(uniques))
            mapping = pc.index_in(uniques, value_set=sorted_uniques)
            codes = pc.take(mapping, codes)
            uniques = sorted_uniques
```


Code example:


```py
import narwhals as nw

data = ['Rome','Milan','Rome','Rome','Milan','Naples']
for sort in (False, True):
	for backend in ("polars", "pandas", "pyarrow"):
    	s = nw.Series.from_iterable(name="foo", values=data, backend=backend)
		codes, _ = s.factorize(sort=sort)
		print(sort, backend, "\n", codes.to_native(), "\n")
```

```terminal
False polars 
 shape: (6,)
Series: 'codes' [i32]
[
        1
        0
        1
        1
        0
        2
] 

False pandas 
 0    0
1    1
2    0
3    0
4    1
5    2
Name: codes, dtype: Int64 

False pyarrow 
 [
  [
    0,
    1,
    0,
    0,
    1,
    2
  ]
] 

True polars 
 shape: (6,)
Series: 'codes' [i32]
[
        2
        0
        2
        2
        0
        1
] 

True pandas 
 0    2
1    0
2    2
3    2
4    0
5    1
Name: codes, dtype: Int64 

True pyarrow 
 [
  [
    0,
    1,
    0,
    0,
    1,
    2
  ]
] 
```

src/narwhals/__init__.py
```diff
@@ -106,6 +106,7 @@
     "Datetime",
     "Decimal",
     "Duration",
+    "Encoded",
```
FBruzzesi, line 109: This is exported publicly: I think it deserves to appear somewhere in the docs

src/narwhals/_arrow/series.py
```diff
@@ -1043,6 +1045,58 @@ def hist_from_bin_count(
...
         )
 
+    def factorize(self, *, sort: bool, null_policy: NullPolicy) -> tuple[Self, Self]:
+        if len(self.native) == 0:
+            codes = pa.chunked_array([[]], type=pa.int32())
+            uniques = pa.chunked_array([[]]).cast(self.native.type)
+            return (self._with_native(codes), self._with_native(uniques))
+
+        # https://github.com/apache/arrow/issues/33297; input pa.NullArray's don't dictionary_encode properly
+        if pa.types.is_null(self.native.type):
```
FBruzzesi, line 1055: I see a couple of issues in this special null handling

## Different integer dtypes

The `pa.NullArray` special case hardcodes possibly unrelated types for each branch:

* `pa.nulls(n, type=pa.int8())` for `Preserve`,
* `pa.repeat(value, n)`, hence `int64` for `Encode`,
* `pa.repeat(value, n)` for `Sentinel`, whose type follows the sentinel's Python type.

Codes dtype becomes Int8, Int64 or Int32 depending on which policy and which column, so something like `nw.concat` of two such results fails on dtype mismatch.

## Non integer dtype

Worse, a non-integer sentinel produces `String` or `Float64` codes here rather than raising, because `pa.repeat` infers from the value.

---

The normal dictionary_encode path yields `int32`, we could do the same and pin every branch to `int32`: `pa.nulls(n, type=pa.int32())`, and `pa.repeat(pa.scalar(v, type=pa.int32()), n)`. WDYT?

src/narwhals/_pandas_like/series.py
```diff
@@ -1198,6 +1200,60 @@ def is_native_dtype_pyarrow(self, native_dtype: Any) -> bool:
...
+
+        match null_policy:
+            case Preserve():
+                codes, uniques = self.native.factorize(sort=sort)
+                codes = pdx.Series(codes, dtype="Int64").mask(lambda s: s == -1)
+
+            case Encode():
+                if (
+                    self._implementation is Implementation.PANDAS
+                    and self._backend_version < (1.5,)
```
FBruzzesi, line 1236: Typo:

```suggestion
                    and self._backend_version < (1,5,)
```

src/narwhals/_polars/series.py
```diff
@@ -670,6 +672,32 @@ def last(self) -> PythonLiteral:
     def any_value(self, *, ignore_nulls: bool) -> PythonLiteral:
         return self.drop_nulls().first() if ignore_nulls else self.first()
 
+    def factorize(self, *, sort: bool, null_policy: NullPolicy) -> tuple[Self, Self]:
+
+        unique_expr = pl.col(self.native.name).unique()
```
FBruzzesi, line 677: `unique` defaults to `maintain_order=False`, which means that multiple runs can result in different outputs so nothing downstream is reproducible.

pandas and pyarrow should both return first-appearance order.

```suggestion
        unique_expr = pl.col(self.native.name).unique(maintain_order=True)
```

Worth verifying null behavior as well

tests/series_only/factorize_test.py
```diff
@@ -0,0 +1,387 @@
...
+
+@pytest.mark.parametrize(
+    ("values", "expected_uniqs", "expected_codes"),
+    [
+        ([], [], []),
+        ([*"abcabc"], ["a", "b", "c"], [0, 1, 2, 0, 1, 2]),
+        ([1, 2, 3, 2], [1, 2, 3], [0, 1, 2, 1]),
+        ([1.1, 2.2, 3.3, 2.2], [1.1, 2.2, 3.3], [0, 1, 2, 1]),
+    ],
+)
```
FBruzzesi, line 159: Every `sort=True` input is already sorted. That's probably why the arrow issue was not spotted

tests/series_only/factorize_test.py
```diff
@@ -0,0 +1,387 @@
...
+
+polars_lt_v1 = POLARS_VERSION < (1, 0, 0)
+pl_skip_reason = "replace_strict only available after 1.0"
+
+NULL_ARGS = [
+    NULL_PRESERVE := {"null_policy": "preserve"},
+    NULL_ENCODE := {"null_policy": "encode"},
+    NULL_SENTINEL := {"null_policy": "sentinel", "sentinel": -1},
+    NULL_SENTINEL := {"null_policy": "sentinel", "sentinel": -2},
```
FBruzzesi, line 23: `NULL_SENTINEL` is defined twice, so the name resolves to the -2 dict.


```suggestion
    NULL_SENTINEL_NEG1 := {"null_policy": "sentinel", "sentinel": -1},
    NULL_SENTINEL_NEG2 := {"null_policy": "sentinel", "sentinel": -2},
```

src/narwhals/stable/v2/__init__.py
FBruzzesi, line 1: Should v1 and v2 emit an unstable warning?

src/narwhals/typing.py
```diff
@@ -426,4 +445,6 @@ def Binary(self) -> type[dtypes.Binary]: ...
     "IntoSeries",
     "IntoSeriesT",
     "LazyAllowed",
+    "NullPolicy",
+    "NullPolicyLiteral",
```
FBruzzesi, line 449: Should these be documented in `typing.md` (and have a one liner docstring) if exposed here?

src/narwhals/_arrow/series.py
```diff
@@ -1043,6 +1045,58 @@ def hist_from_bin_count(
...
+                    )
+                case _:  # pragma: no cover
+                    assert_never(null_policy)
+
+            return (self._with_native(codes), self._with_native(uniques))
+
+        native = self.native
+        if pa.types.is_dictionary(native.type):
+            # re-encode if already dict encoded; can't be certain how the original dict encoding was done
+            native = native.cast(native.type.value_type)
```
FBruzzesi, line 1080: The re-encode casts a dictionary column to its `value_type` and never casts uniques back, so `uniques.dtype` is String on pyarrow while polars and pandas both return Categorical.

Consider capturing `self.native.type` before the re-encode and `uniques = uniques.cast(original_type)` before returning, and maybe add a dtype assertion in the test

src/narwhals/_pandas_like/series.py
```diff
@@ -1198,6 +1200,60 @@ def is_native_dtype_pyarrow(self, native_dtype: Any) -> bool:
...
+                codes, uniques = self.native.factorize(sort=sort)
+                if value != -1:
+                    codes = pdx.Series(codes, dtype="Int64").mask(
+                        lambda s: s == -1, value
+                    )
+            case _:  # pragma: no cover
+                assert_never(null_policy)
+
+        return (
+            self._with_native(pdx.Series(codes)),
```
FBruzzesi, line 1253: Should we attach the original index to the series? We currently create a RangeIndex

src/narwhals/_polars/series.py
```diff
@@ -670,6 +672,32 @@ def last(self) -> PythonLiteral:
     def any_value(self, *, ignore_nulls: bool) -> PythonLiteral:
         return self.drop_nulls().first() if ignore_nulls else self.first()
 
+    def factorize(self, *, sort: bool, null_policy: NullPolicy) -> tuple[Self, Self]:
```
FBruzzesi, line 675: Considering the test suite, we can add `@requires.backend_version((1,))`

src/narwhals/series.py
```diff
@@ -2909,6 +2912,94 @@ def is_close(
...
+            raise TypeError(msg)
+
+        if null_policy == "preserve":
+            policy = Preserve()
+        elif null_policy == "encode":
+            policy = Encode()
+        elif null_policy == "sentinel":
+            policy = Sentinel(sentinel)
+        else:
+            assert_never(null_policy)
```
FBruzzesi, line 2995: `assert_never` raises `AssertionError("Expected code to be unreachable, but got: ... Please report an issue at <bug_report url>")`. But `null_policy` is a plain user-supplied string with only static typing behind it, so `factorize(null_policy="perserve")` sends the user to the issue tracker for their own typo. 

Other methods raise a `ValueError` naming the valid options: I would do the same, here naming the 3 valid options and keep `assert_never` where we the value genuinely is unreachable (e.g. at the compliant level)

src/narwhals/series.py
```diff
@@ -2950,3 +3041,26 @@ def list(self) -> SeriesListNamespace[Self]:
     @property
     def struct(self) -> SeriesStructNamespace[Self]:
         return SeriesStructNamespace(self)
+
+
+@unstable
+@dataclass(frozen=True)
+class Encoded(Generic[IntoSeriesT]):
```
FBruzzesi, line 3048: I can't remember if and when we changed this from another data structure (e.g. namedtuple).

The claim "it unpacks like pandas" is currently false as we cannot indexing, `len`, or check equality like a tuple would. Only `__iter__` is defined:

```text
result[0]        -> TypeError: 'Encoded' object is not subscriptable
len(result)      -> TypeError: object of type 'Encoded' has no len()
enc_a == enc_b   -> TypeError: the truth value of a Series is ambiguous
```

I was hoping for a namedtuple, and probably lost track of the updates

src/narwhals/series.py
```diff
@@ -2950,3 +3041,26 @@ def list(self) -> SeriesListNamespace[Self]:
...
+    codes: Series[IntoSeriesT]
+    uniques: Series[IntoSeriesT]
+
+    def __iter__(self) -> Iterator[Series[IntoSeriesT]]:
+        yield self.codes
+        yield self.uniques
+
+    @property
+    def mapping(self) -> Mapping[Any, int]:
+        """Forward map as a joinable ``(value, code)`` frame; works for any dtype."""
```
FBruzzesi, line 3065: This docstring is also wrong: it raises on unhashable dtypes, and it keys nulls inconsistently

Three separate inaccuracies:

* it returns a dict, not a joinable frame;
* "any dtype" is false, since polars Struct uniques raise a bare TypeError: unhashable type: 'dict';
* the null key is None on polars/pyarrow but nan / pd.NA on pandas, so mapping[None] is missing there under null_policy="encode".

src/narwhals/series.py
```diff
@@ -2909,6 +2912,94 @@ def is_close(
...
+            |        "b"            |
+            |]                      |
+            └───────────────────────┘
+        """
+        if null_policy == "sentinel" and sentinel is NO_DEFAULT:
+            msg = "Must supply `sentinel` when null_policy='sentinel'"
+            raise TypeError(msg)
+        if null_policy != "sentinel" and sentinel is not NO_DEFAULT:
+            msg = f"Argument `sentinel` is ignored when null_policy='{null_policy}'"
+            raise TypeError(msg)
```
FBruzzesi, line 2986: Should these be `ValueError`'s? `fill_null` has similar checks all raising a `ValueError`

src/narwhals/series.py
```diff
@@ -2909,6 +2912,94 @@ def is_close(
         result = result.rename(orig_name) if name_is_none else result
         return cast("Self", result)
 
+    @unstable
+    def factorize(
```
FBruzzesi, line 2916: First off, the `null_policy` / `sentinel` design reads really nicely, much better than a boolean flag, and the three policies cover everything I'd reach for. This is a small thing on top of that.

While playing with sentinel I noticed it currently accepts any value, and I wonder whether it's worth narrowing slightly. Good news first: the values you'd actually reach for behave exactly right. -1 and -2 both round-trip cleanly on all three backends, so this is much narrower than "the sentinel needs validating".

Where it gets a bit shady is a sentinel that also happens to be a valid code. `sentinel=0` is accepted everywhere, and since 0 is a real index into uniques, the null becomes indistinguishable from a genuine value afterwards. Probably not something you'd write on purpose, though 0 _can_ feel like a natural pick for an "unknown" bucket, so I could imagine someone landing there. I think this is also the reason `pandas.factorize` restricts it to -1.

The part I found trickiest is that the safe range depends on the data, so it's hard for a caller to reason about:

```text
sentinel=5, 3 uniques:  codes=[0,1,2,5]           -> IndexError in the consumer
sentinel=5, 8 uniques:  codes=[0,1,2,3,4,5,6,7,5] -> quietly decodes as a real value
```

Same call, different outcome depending on cardinality, which makes it the kind of thing that passes on sample data.

---

Separately, there's some backend variation in what's accepted at all: 2**40 goes through on pandas but raises on polars and pyarrow, 2.5 truncates on polars/pyarrow but raises on pandas, and True becomes code 1 on polars while the other two raise. Also `sentinel=None` is accepted everywhere and gives null codes, which ends up equivalent to null_policy="preserve" even though the API asks you for a sentinel to get there.

So: would it make sense to require a negative int in `Series.factorize`, rejecting bool (since `isinstance(True, int) is True`) and anything below `-2**31`? I'd lean towards that over checking 0 <= sentinel < len(uniques) after the fact, mainly because it's checkable up front and doesn't depend on the data, and it happens to cover all the cases below in one guard. But you've thought about this API far more than I have, so if you'd rather leave it to you and just note it in the docstring, that's a perfectly reasonable call too. Curious what you think.

---

Code example:

```py
import narwhals as nw

def factorize(backend, **kwargs):
    s = nw.Series.from_iterable(name="x", values=values, backend=backend)
    codes, uniques = s.factorize(sort=True, null_policy="sentinel", **kwargs)
    return codes.to_list(), uniques.to_list()

backends = ("pandas", "polars", "pyarrow")
values = ["a", None, "b"]

print("1. Sentinel inside range(len(uniques))")
for sentinel in (-1, 0, 1):
    for backend in backends:
        codes, uniques = factorize(backend, sentinel=sentinel)
        # The guarded decoder from tests/series_only/factorize_test.py
        decoded = [uniques[c] if c >= 0 else None for c in codes]
        ok = "ok  " if decoded == values else "LOSS"
        print(f"  sentinel={sentinel:<3} {backend:8} {ok} uniques={uniques} codes={codes} -> decoded {decoded}")
    print()

print("2. same sentinel, different behaviour per backend")
for sentinel in (True, 2.5, 2**40, None):
    cells = []
    for backend in backends:
        try:
            codes, _ = factorize(backend, sentinel=sentinel)
            cells.append(f"{backend}={codes}")
        except Exception as exc:
            cells.append(f"{backend}=raises {type(exc).__name__}")
    print(f"sentinel={sentinel} " + "  ".join(cells))
```

```terminal
1. Sentinel inside range(len(uniques))
  sentinel=-1  pandas   ok   uniques=['a', 'b'] codes=[0, -1, 1] -> decoded ['a', None, 'b']
  sentinel=-1  polars   ok   uniques=['a', 'b'] codes=[0, -1, 1] -> decoded ['a', None, 'b']
  sentinel=-1  pyarrow  ok   uniques=['a', 'b'] codes=[0, -1, 1] -> decoded ['a', None, 'b']

  sentinel=0   pandas   LOSS uniques=['a', 'b'] codes=[0, 0, 1] -> decoded ['a', 'a', 'b']
  sentinel=0   polars   LOSS uniques=['a', 'b'] codes=[0, 0, 1] -> decoded ['a', 'a', 'b']
  sentinel=0   pyarrow  LOSS uniques=['a', 'b'] codes=[0, 0, 1] -> decoded ['a', 'a', 'b']

  sentinel=1   pandas   LOSS uniques=['a', 'b'] codes=[0, 1, 1] -> decoded ['a', 'b', 'b']
  sentinel=1   polars   LOSS uniques=['a', 'b'] codes=[0, 1, 1] -> decoded ['a', 'b', 'b']
  sentinel=1   pyarrow  LOSS uniques=['a', 'b'] codes=[0, 1, 1] -> decoded ['a', 'b', 'b']

2. same sentinel, different behaviour per backend
sentinel=True pandas=raises TypeError  polars=[0, 1, 1]  pyarrow=raises ArrowTypeError
sentinel=2.5 pandas=raises TypeError  polars=[0, 2, 1]  pyarrow=[0, 2, 1]
sentinel=1099511627776 pandas=[0, 1099511627776, 1]  polars=raises InvalidOperationError  pyarrow=raises ArrowInvalid
sentinel=None pandas=[0, <NA>, 1]  polars=[0, None, 1]  pyarrow=[0, None, 1]
```

FBruzzesi (commented):
Thanks @camriddell - I lost myself in a lot of details and left quite a few comments 🙈 I am sorry if that feels a bit overwhelming, yet I hope they are helpful

src/narwhals/_polars/series.py
camriddell: For performance I'd rather not enforce `maintain_order=True` here. If the user wants stability in their results, they should call `.factorize(sort=True)`

tests/series_only/factorize_test.py
camriddell: Yep, verified and added into the tests. Awesome catch.

PR #3815, opened 2026-07-24
feat: Add `{DataFrame, LazyFrame}.cast`

### LINKED ISSUE #3402: [Enh]: Add support `{DataFrame, LazyFrame}.cast(...)`
Necessity is for better users (and self) ergonomics.

---

## Backend support
- [`dask.dataframe.DataFrame.astype`](https://docs.dask.org/en/stable/generated/dask.dataframe.DataFrame.astype.html#dask.dataframe.DataFrame.astype)
- [`ibis.Table.cast`](https://ibis-project.org/reference/expression-tables#ibis.expr.types.relations.Table.cast)
- [`pandas.DataFrame.astype`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.astype.html#pandas.DataFrame.astype)
- [`polars.LazyFrame.cast`](https://docs.pola.rs/api/python/stable/reference/lazyframe/api/polars.LazyFrame.cast.html)
- [`pyarrow.Table.cast`](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.cast)
- [`pyspark.sql.DataFrame.to`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.to.html#pyspark.sql.DataFrame.to)

## Related

- https://github.com/narwhals-dev/narwhals/pull/3398#discussion_r2682943041
- #1045

---

_Originally posted by @dangotbanned in https://github.com/narwhals-dev/narwhals/pull/3398#discussion_r2682943041_

What's here *could be* both more efficient *and* shorter:

https://github.com/narwhals-dev/narwhals/blob/8fabb13e4f74be9236504b139334d6dbbb697684/narwhals/_arrow/namespace.py#L180-L194

<details><summary><b>Show <code>ArrowDataFrame.cast</code> diff</b></summary>
<p>

```diff
+++ b/narwhals/_arrow/dataframe.py
+    def cast(self, schema: IntoSchema | Iterable[tuple[str, DType]]) -> Self:
+        return self._with_native(
+            self.native.cast(Schema(schema).to_arrow()), validate_column_names=False
+        )
+

+++ b/narwhals/_arrow/namespace.py
-from narwhals._arrow.utils import cast_to_comparable_string_types
+from narwhals._arrow.utils import cast_to_comparable_string_types, concat_tables
-from narwhals.schema import Schema, combine_schemas, to_supertype
+from narwhals.schema import Schema, to_supertype, to_supertype_diagonal
```

</p>
</details> 

With those changes, we can avoid materializing as much all at once using generators:

```py
    def concat_diagonal_relaxed(self, dfs: Sequence[ArrowDataFrame], /) -> ArrowDataFrame:
        schemas = [Schema(df.schema) for df in dfs]
        supertypes = reduce(to_supertype_diagonal, schemas)
        to_concat = (
            df.cast((name, supertypes[name]) for name in schema).native
            for df, schema in zip(dfs, schemas)
        )
        return dfs[0]._with_native(concat_tables(to_concat))
```

mkzung:
# Description

Adds `DataFrame.cast` and `LazyFrame.cast`.

```python
df.cast({"a": nw.Float64, "b": nw.Int32})
```

Takes a `{column: dtype}` mapping and casts those columns, leaving the rest unchanged. It uses each backend's native frame cast where there is one (`pyarrow` `Table.cast`, `pandas`/`dask` `astype`, `polars` `cast`); the SQL backends (`duckdb`, `pyspark`) cast per column through `with_columns`, and `ibis` inherits that path. Casting a column that is not in the frame raises `ColumnNotFoundError`, the same as `select` and `drop`.

I kept it to the mapping form, which is what was proposed in the issue. Polars also takes a single dtype to cast every column, so I can add that too if you'd like it for parity.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3402

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes


src/narwhals/_arrow/dataframe.py
```diff
@@ -708,6 +700,17 @@ def rename(self, mapping: Mapping[str, str]) -> Self:
...
+    def cast(self, dtypes: Mapping[str, IntoDType]) -> Self:
+        native = self.native
+        # `with_type` keeps each field's name, nullability and metadata.
+        target = pa.schema(
+            field.with_type(narwhals_to_native_dtype(dtypes[field.name], self._version))
+            if field.name in dtypes
+            else field
+            for field in native.schema
+        )
+        return self._with_native(native.cast(target), validate_column_names=False)
```
FBruzzesi: Can you run a quick benchmark for pyarrow to assess if the implementation here is faster than:

 ```py
ns = self.__narwhals_namespace__()
return self.with_columns(
	*(ns.col(name).cast(dtype) for name, dtype in dtypes.items())
)
```

File: src/narwhals/_pandas_like/dataframe.py [hunk 2]
```diff
@@ -521,6 +525,19 @@ def rename(self, mapping: Mapping[str, str]) -> Self:
             rename(self.native, columns=mapping, implementation=self._implementation)
         )
 
+    def cast(self, dtypes: Mapping[str, IntoDType]) -> Self:
+        native = self.native
+        to_cast = {
+            name: narwhals_to_native_dtype(
+                dtype,
+                dtype_backend=get_dtype_backend(native[name].dtype, self._implementation),
```
FBruzzesi: You can grab the native type by first getting all the native types via `pd_dtypes = native.dtypes` and then `pd_dtypes[name]`, so that we avoid slicing the dataframe _just_ to get a column type

src/narwhals/_polars/dataframe.py
```diff
@@ -167,6 +168,15 @@ def __native_namespace__(self) -> ModuleType:
     def _with_native(self, df: NativePolarsFrame) -> Self:
         return self.__class__(df, version=self._version)
 
+    def cast(self, dtypes: Mapping[str, IntoDType]) -> Self:
+        native_dtypes = {
+            name: narwhals_to_native_dtype(dtype, self._version)
+            for name, dtype in dtypes.items()
+        }
+        return self._with_native(
+            self.native.cast(native_dtypes)  # pyright: ignore[reportArgumentType]
```
FBruzzesi: Is `# pyright: ignore[reportArgumentType]` an upstream polars issue?

FBruzzesi (commented):
Thanks for the contribution @mkzung - I left a few inline comments. On top of those, I wonder about the rationale regarding:

> the SQL backends (duckdb, pyspark) cast per column through with_columns, and ibis inherits that path

as listed in the issue, pyspark and ibis have their own casting method and I think we should used them instead of choosing for a user. DuckDB doesn't seem to have it, SQLFrame I didn't check. 

---

Also, there are a few conflicts to solve with the main branch before being able to merge

mkzung:
Rebased, conflicts resolved. On the comments:

- Benchmarked the pyarrow path against `with_columns(ns.col(name).cast(...))`, same speed. Kept
  the schema version since `field.with_type` preserves nullability and metadata that the
  per-column cast drops (`double not null` becomes `double`).
- pandas now reads `native.dtypes[name]` instead of slicing.
- ibis has a native `Table.cast`, so it uses that. pyspark/sqlframe/duckdb have no frame-level
  cast, so they stay on the shared path.
- The polars ignore is just a stubs gap (the mapping key is typed wider than `dict[str, ...]`).
  Moved it to `# type: ignore[arg-type]` to match the rest of the polars backend.

Also moved the LazyFrame column check to `collect_schema().names()`, since `self.columns` now
warns on lazy polars.

FBruzzesi:
Thanks for adjusting @mkzung 

> * Benchmarked the pyarrow path against `with_columns(ns.col(name).cast(...))`, same speed. Kept
>   the schema version since `field.with_type` preserves nullability and metadata that the
>   per-column cast drops (`double not null` becomes `double`).

Preserving mutability and metadata is a good point. I wonder if we should do it in any cast operation.
Regarding the benchmark, can you please share the script and the numbers with us?

>  pyspark/sqlframe/duckdb have no frame-level cast, so they stay on the shared path.

What about [`pyspark.sql.DataFrame.to`](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.DataFrame.to.html)

---

Can you also please run all the checks locally as mentioned in the contributing guide?

mkzung:
Benchmark script and numbers (pyarrow 23, 500k rows x 8 int columns, 50 runs after warm-up):

```python
import time
import numpy as np
import pyarrow as pa
import narwhals as nw

rng = np.random.default_rng(0)
tbl = pa.table({f"c{i}": rng.integers(0, 100, 500_000) for i in range(8)})
df = nw.from_native(tbl, eager_only=True)
targets = {f"c{i}": nw.Float64 for i in range(8)}

def schema_cast():
    return df.cast(targets)

def with_columns_cast():
    return df.with_columns(*(nw.col(n).cast(d) for n, d in targets.items()))

for fn in (schema_cast, with_columns_cast):
    fn()
    start = time.perf_counter()
    for _ in range(50):
        fn()
    print(f"{fn.__name__}: {(time.perf_counter() - start) / 50 * 1000:.2f} ms")
```

```
schema_cast: 1.73 ms
with_columns_cast: 1.83 ms
```

So no meaningful speed difference; the reason to keep the schema path is only the
nullability/metadata preservation.

On `DataFrame.to`: it reconciles the whole frame to the given schema, so a partial mapping
would mean rebuilding the full target schema from the current one plus the overrides, and
sqlframe does not implement `.to` at all (checked both its DuckDB and Spark flavors), so the
spark-like backend cannot rely on it across the board. I can add a pyspark-only path via a
constructed StructType if you think it is worth the special case; otherwise the shared
with_columns path keeps the three SQL backends uniform.

On doing the field-preserving cast everywhere: for the eager per-column `Expr.cast` there is
no field object to preserve (the expression builds a new array), so this only really applies
where a backend casts a whole table; happy to open an issue to track it if useful.

Checks: prek, pyright/mypy/pyrefly and the full-coverage suite were run locally before the
push; everything on the PR is green now.

FBruzzesi (approved):
@MarcoGorelli I just did one more pass and added a couple of commits: one to [simplify some bits and remove trivial inline comments](https://github.com/narwhals-dev/narwhals/pull/3815/commits/e2ccd3a25da5a668be4f8258242836e4f25de853) (some AI slop leftover), and one [to fix an edge case for pyarrow](https://github.com/narwhals-dev/narwhals/pull/3815/commits/d286b9161463e2078c1372c9591cfe40afce9f8f).

For me this is ready to merge, but happy to get a second look whenever you have the time 🙏🏼

mkzung:
The red check is not this branch. `test_corr_constant_column[pyspark[connect]]` landed on main at
13:49 on 17 Sep and passed there; my run was 15:51 and the diff touches cast only. Happy to rebase
if you want a clean run.

mkzung:
The red check is not this branch. `test_corr_constant_column[pyspark[connect]]` landed on main at
13:49 on 17 Sep and passed there; my run was 15:51 and the diff touches cast only. Happy to rebase
if you want a clean run.

Issue #3632, opened 2026-05-16
Add an AI Policy


FBruzzesi:
Opening as an issue to track progress, following the [Discord thread discussion](https://discord.com/channels/1235257048170762310/1504514041215979792/1504514056285978916).

---

> [!IMPORTANT]
>  The goal is to set sensible guardrails that protect maintainer time without categorically rejecting AI-assisted work, while keeping human contributors and human review as the priority.

---

## Proposed measures

- [x] Mandatory AI disclosure on PRs. Requires adding a PR-template field asking whether any AI tool or model was used, and which one. The point is transparency. Addressed by #3701
- [x] Responsibility clause in `CONTRIBUTING.md`. Add wording along the lines of:
    > You are responsible for every line you submit, regardless of how it was produced. Disclosure does not transfer responsibility to the tool.
	
	Addressed by #3701
- [x] Author responsiveness as the bar: PRs where the author does not engage with review feedback are closed without extended discussion. This is reactive, not preventive: we cannot screen for it upfront, but we can be explicit that PRs where the author does not respond to review feedback will be closed without extended discussion, whether or not AI was used. Addressed by #3701
- [x] Add an `AGENTS.md` (under discussion) - WIP in #3821
- [ ] Ship a user-facing `SKILL.md` files. Examples from FastAPI: [library-skills.io](https://library-skills.io/) and [FastAPI skills](https://github.com/fastapi/fastapi/blob/master/fastapi/.agents/skills/fastapi/SKILL.md)
- [ ] AI-assisted PR review as a supplement to, and never a replacement for, human review. Open questions are:
  - Cost: Do any vendors offer a free allowance for open source projects worth investigating?
    - [coderabbit](https://github.com/marketplace/coderabbitai) Pro plan is free for Open Source projects
    - [qodo](https://www.qodo.ai/) also free for OSS
  - If individual maintainers use AI personally for review, we could add a `review.md` describing what to look for and how to structure the review.
  - **Edit**: my personal stand on this is that, although useful, these kind of reviews should not run by default, but only when triggered. The main risk I see is biasing the reviewer and/or give a false sense of trust

---

Additional references:

- Thread originating PR: #3621
- Related human-contributor improvements: #3608, #3552
- A community-maintained Survey of OSS AI policies: [melissawm/open-source-ai-contribution-policies](https://github.com/melissawm/open-source-ai-contribution-policies)
- Polars contributing guide: https://docs.pola.rs/development/contributing/
[...]

PR #3821, opened 2026-07-26
docs: Add `AGENTS.md` 

FBruzzesi:
# Description

We started to observe more a more PRs which are LLM generated. As mentioned in #3632, I would not oppose to having an `AGENTS.md` file - this can point the agent to the checks it needs to run before finalizing its work.

The alternative is having myself or other reviewers checking out the PR just to realize something as simple as pre-commit hooks were not installed (see #3816) or `make typing` was not run locally.

Triggering CI for that is also a waste of resources which I would like to avoid. Consider this an additional protection layer before we do any review.

> [!TIP]
> Using "Display the rich diff" will make your life much easier in reviewing this PR 

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #3632


MarcoGorelli:
generally looks good, but about "Source layout":

- isn't this something ai agents and see for themselves?
- won't it quickly go out-of-date?

FBruzzesi:
Thanks @MarcoGorelli 

> generally looks good, but about "Source layout":
> 
> * isn't this something ai agents and see for themselves?
> * won't it quickly go out-of-date?

**Disclaimer**:  I don't know if this is a good practice or not. What I am doing at work is the following:

1. Keep a one-liner in each module `__init__.py`
2. Have a hook that automatically parses the one-liner and keep the structure in sync with it's description

On top of this I have a per-module `README.md`.

---

I am ok with removing it for now and iterate if needed

MarcoGorelli (approved):
i guess an ai agent will eventually say something if it goes out of sync 😄 

happy to go forwards

AGENTS.md
```diff
@@ -0,0 +1,216 @@
...
+  objects.
+* **100% branch coverage** is enforced by the full-coverage CI job. When a branch is genuinely
+  unreachable (e.g. gated on an unsupported backend version), mark it `# pragma: no cover` with a
+  one-line reason.
+* **Breaking changes never land in `narwhals.stable.v1` or `narwhals.stable.v2`.** New public APIs
+  land in the main `narwhals` namespace and graduate into the next stable version. See
+  [docs/backcompat.md](docs/backcompat.md), and add an entry to its `main` vs `stable.*` diff when
+  the namespaces diverge.
+
+Backend-specific rules (no pandas `apply`/`map`/`assign`/`drop`/`reset_index`/`rename`, no Polars
```
camriddell, line 77: What is the issue with each of these pandas methods? I understand `Series.{map,apply}` since that move element-wise, but `DataFrame.apply`, `.assign`, `.drop`, `.reset_index`, `.rename` are all pretty useful methods?

camriddell (commented):
Only one small question, once addressed I'm happy to approve.

AGENTS.md
FBruzzesi: Thanks Cam 🙏🏼

- `apply`: we should not have the need to use it in the first place.
- `assign`: we went a different direction in `select`/`with_columns`, I remember Marco running some benchmark to decide.
- `drop` & `reset_index`: we use them, but we are quite careful with their parameters depending on pandas version.
- `rename`: we have _our own_ `rename` wrapper: https://github.com/narwhals-dev/narwhals/blob/00dd1cfe93845f2622b4e73cfbf8963b524afaee/src/narwhals/_pandas_like/utils.py#L188-L189

In general nothing is banned, but I rather have a strong(er) guideline and a discussion afterwards than everything running through an `apply` + `lambda` 😅

camriddell (approved):

FBruzzesi:
Thanks everyone for reviewing and contributing to this. I will go ahead and merge for now, we can iterate on it if/when needed 🙏🏼

PR #3822, opened 2026-07-27
feat: Resolve plugins via `from_backend`

FBruzzesi:
# Description

Addresses https://github.com/narwhals-dev/narwhals/pull/3753#discussion_r3652844118

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


src/narwhals/_utils.py
```diff
@@ -1654,6 +1654,24 @@ def is_eager_allowed(impl: Implementation, /) -> TypeIs[_EagerAllowedImpl]:
     }
 
 
+# NOTE: Keep `TypeGuard`, not `TypeIs`, as the two narrow by different rules.
+def is_into_plugin(
+    backend: IntoBackend[Backend | PluginName],  # noqa: ARG001
+    impl: Implementation,
+    /,
+) -> TypeGuard[IntoBackend[PluginName]]:
```
FBruzzesi, line 1660: I spent an unhealthy amount of time on this. @dangotbanned I am sure this is obvious to you, but it's still a bit mysterious for me

src/narwhals/_utils.py
```diff
@@ -1654,6 +1654,24 @@ def is_eager_allowed(impl: Implementation, /) -> TypeIs[_EagerAllowedImpl]:
     }
 
 
+# NOTE: Keep `TypeGuard`, not `TypeIs`, as the two narrow by different rules.
+def is_into_plugin(
+    backend: IntoBackend[Backend | PluginName],  # noqa: ARG001
+    impl: Implementation,
```
FBruzzesi, line 1658: `impl` is a parameter rather than re-derived here since in `eager_namespace` it's already computed (it's not in `_io_method`)

src/narwhals/_utils.py
dangotbanned: I will try to work through this with you soon 🙏

MarcoGorelli (commented):
thanks @FBruzzesi 

One thing that may be worth pointing out is this: say in `read_csv`, we do:

- `impl = Implementation.from_backend(backend)`
- then, `read = _io_method(backend, "read_csv", version=Version.MAIN)`, in which we do `impl = Implementation.from_backend(backend)` again
- then, in `namespace = version.namespace.from_backend(backend).compliant`, we do `impl = Implementation.from_backend(backend)` again, a third time

is it worth caching `from_backend` at least?

src/narwhals/_namespace.py
```diff
@@ -69,6 +71,58 @@
 __all__ = ["Namespace"]
 
 
+@cache
```
FBruzzesi, line 75: We can safely use an unbounded cache, the size is constrained by `len(Implementation._member_names_) * len(Version._member_names_)`.

src/narwhals/_namespace.py
```diff
@@ -152,41 +206,16 @@ def from_backend(
...
-            from narwhals._duckdb.namespace import DuckDBNamespace
-
-            ns = DuckDBNamespace(version=version)
-        elif impl.is_dask():
-            from narwhals._dask.namespace import DaskNamespace
-
-            ns = DaskNamespace(version=version)
-        elif impl.is_ibis():
-            from narwhals._ibis.namespace import IbisNamespace
+        if is_into_plugin(backend, impl):
```
FBruzzesi, line 216: Deliberately *not* cached 👀 do you think it's _always_ safe to cache?

src/narwhals/_utils.py
```diff
@@ -395,32 +386,21 @@ def from_native_namespace(
...
-            get_duckdb(): Implementation.DUCKDB,
-            get_ibis(): Implementation.IBIS,
-            get_sqlframe(): Implementation.SQLFRAME,
-            get_pyspark_connect(): Implementation.PYSPARK_CONNECT,
-        }
-        return mapping.get(native_namespace, Implementation.UNKNOWN)
+        name = getattr(native_namespace, "__name__", "")
+        if (impl := _MODULE_NAME_TO_IMPLEMENTATION.get(name)) is not None and (
+            sys.modules.get(name) is native_namespace
+        ):
```
FBruzzesi, line 392: Not sure we can use `get_<module>` in `_MODULE_NAME_TO_IMPLEMENTATION` as the module might not be loaded when defining the dict, hence this workaround

src/narwhals/_namespace.py
MarcoGorelli: can't see why not, but it just returns `Implementation.UNKNOWN` so i think this is fine either way

FBruzzesi: Adding a the `@cache` to `_find_plugin` otherwise we need to carry another "key" in  `_compliant_namespace` to cache and get the plugin namespace, as `impl` would be UNKNOWN for all plugin's

src/narwhals/_namespace.py
```diff
@@ -152,42 +212,18 @@ def from_backend(
...
-            ns = IbisNamespace(version=version)
+            resolved = _resolve_plugin(backend)
         else:
-            msg = "Not supported Implementation"  # pragma: no cover
-            raise AssertionError(msg)
-        return cls(ns)
+            resolved = impl
+        # NOTE: `type: ignore` as `functools.cache` requires resolved to be hashable;
+        # a plugin is a module, so it always is.
+        return cls(_compliant_namespace(resolved, cls._version))  # type: ignore[arg-type]
```
FBruzzesi, line 226: Skill issue to make `Plugin` Hashable without breaking a bunch of other things

PR #3834, opened 2026-08-01
skip changelog(deps): bump GHA and comply with zizmor

FBruzzesi:
# Description

Making this happy:

```terminal
pinact run -u
uvx zizmor .
```

Closes #3833, #3832, #3831, #3830, #3829


EdAbati (approved):
Nice!

Should we add `zizmor` as pre-commit hook?

FBruzzesi:
Thanks @EdAbati - I was think to followup by adding zizmor as pre-commit and creating a pinact action/job to replace dependabot, so that we can update all versions in one PR instead of separates

PR #3619, opened 2026-05-13
ci: unpin pyright version

MarcoGorelli:
closes #3584 

# Description

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue #\<issue number\>
- Closes #\<issue number\>

## Checklist

- [ ] Code follows style guide (ruff)
- [ ] Tests added
- [ ] Documented the changes

PR #3842, opened 2026-08-09
feat: add `LazyFrame.pivot`

### LINKED ISSUE #1901: enh: add LazyFrame.pivot
This is something I'd quite like to see. In Polars too, but it's easier to implement here 😄 

There's an example in the Polars docs about how to do a lazy pivot: https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pivot.html

API-wise, we could have:
```python
    def pivot(
        self: Self,
        on: str | list[str],
        *,
        column_names: list[str],
        index: str | list[str] | None = None,
        values: str | list[str] | None = None,
        aggregate_function: Literal[
            "min", "max", "sum", "mean", "median", "len"
        ]
        | None = None,
        separator: str = "_",
    ) -> Self:
```
This is very similar to `DataFrame.pivot`, but requires specifying the output column names explicitly so that the computation can stay completely lazy

CangyuanLi:
# Description

Adds lazyframe pivot.

A couple of decisions I'm not sure about and a (maybe) big typing question I wasn't sure how to handle. 

Decision 1:

Currently, this PR only supports a scalar `on`. If `on` could be a sequence, then `on_columns` must be a dataframe. I wasn't sure how to express this cleanly across all backends, and it significantly increases the complexity. I'm hoping that a reduced version with scalar `on` would be acceptable for now, but also happy to try and support multiple `on`. Consequently, `on_columns` also cannot be a dataframe.

Decision 2:

Only pre-defined aggregations are supported, whereas polars supports expressions. I followed the convention of the existing eager pivot implementation here and excluded expression inputs.

Decision 3:

Polars pivot (both eager and lazy) has been updated with a couple of convenience parameters like column_naming. Since these don't appear in the existing eager implementation, I excluded them in the lazy one as well.

Typing Error:

EagerDataFrame inherits from CompliantDataFrame and CompliantLazyFrame. Pivot has different signatures for dataframe and lazyframe. I wasn't really sure how to handle this without making some larger changes I wasn't comfortable doing, e.g. updating the eager implementation to also have on_columns, and perhaps removing "item" from the lazy pivot aggregation options to maintain parity with the eager pivot? Advice here would be greatly appreciated!

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue https://github.com/narwhals-dev/narwhals/issues/1901
- Closes https://github.com/narwhals-dev/narwhals/issues/1901

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes

    ```bash
    ```


CangyuanLi:
nightlies 3.12 error: seems unrelated, seems something to do with pandas? affects a ton of tests
lightgbm: downloading scipy timed out, seems unrelated
scikit-lego: seems unrelated, the test is related to timgegapslit, which as far as I can tell does not use lazy pivot

src/narwhals/_duckdb/dataframe.py
```diff
@@ -556,6 +564,80 @@ def unpivot(
...
+        on: str,
+        on_columns: Sequence[Any],
+        *,
+        index: Sequence[str] | None,
+        values: Sequence[str] | None,
+        aggregate_function: LazyPivotAgg | None,
+        maintain_order: bool,
+        separator: str,
+    ) -> Self:
+        # DuckDB does not have pivot in its relational API
```
FBruzzesi, line 578: Together with the next comment/suggestion

```suggestion
```

src/narwhals/_duckdb/dataframe.py
```diff
@@ -556,6 +564,80 @@ def unpivot(
...
+            )
+            raise NotImplementedError(msg)
+
+        index, values = resolve_pivot_index_values(self.columns, on, index, values)
+
+        # Example generated query:
+        # PIVOT _rel ON "subject" IN ('maths', 'physics')
+        # USING mean("test_1") AS "test_1", mean("test_2") AS "test_2"
+        # GROUP BY "name"
+
```
FBruzzesi, line 595:
```suggestion
		# DuckDB has no pivot in its relational API, so we hand-write the query, e.g.
        # PIVOT _rel ON "subject" IN ('maths', 'physics')
        # USING mean("test_1") AS "test_1", mean("test_2") AS "test_2"
        # GROUP BY "name"
```

src/narwhals/_duckdb/dataframe.py
```diff
@@ -556,6 +564,80 @@ def unpivot(
...
+            #
+            # One value:
+            # if values=["test_1"]
+            # name | maths
+            # Cady |    98
+            #
+            # Multiple Values:
+            # if values=["test_1", "test_2"]
+            # name | maths_test_1 | maths_test_2
+            # Cady |           98 |          100
```
FBruzzesi, line 630: I would cut it to something as:


```suggestion
			# DuckDB prefixes the value name only when there is more than one value:
            # `maths` for values=["test_1"], `maths_test_1` and `maths_test_2` for
            # values=["test_1", "test_2"].```

File: src/narwhals/_duckdb/dataframe.py [hunk 4]
```diff
@@ -556,6 +564,80 @@ def unpivot(
...
+            source_name = str(on_value) if len(values) == 1 else f"{on_value}_{value}"
+            expression = col(source_name)
+            if aggregate in {"sum", "count"}:
+                # duckdb returns null for a missing pivot combination, to match polars
+                # semantics we need to 0. For the other aggregations like mean, min,
+                # max, both duckdb and polars return null.
```
FBruzzesi, line 636: I remember having a big discussion over the default we should have here 😅


```suggestion
                # DuckDB gives null for a missing combination, polars gives 0.
                # Every other aggregation returns null in both.
```

src/narwhals/_ibis/dataframe.py
```diff
@@ -407,6 +414,53 @@ def unpivot(
...
+
+        aggregate = "count" if aggregate_function == "len" else aggregate_function
+        result = self.native.pivot_wider(
+            id_cols=s.cols(*index),
+            names_from=on,
+            names=on_columns,
+            names_sep=separator,
+            values_from=values,
+            # ibis will treat missing pivot combination as null for these, instead of 0
+            # as polars does.
```
FBruzzesi, line 449:
```suggestion
            # ibis gives null for a missing combination, polars gives 0.
```

src/narwhals/_ibis/dataframe.py
```diff
@@ -407,6 +414,53 @@ def unpivot(
...
+            values_fill=0 if aggregate in {"sum", "count"} else None,
+            values_agg=aggregate,
+        )
+
+        # ibis orders pivoted columns by `on_columns` first, while polars orders
+        # them by `values` first.
```
FBruzzesi, line 455:
```suggestion
        # ibis orders the pivoted columns by `on_columns` first, polars by `values`.
```

src/narwhals/_utils.py
FBruzzesi, line 1: Nice ones, can these be re-used in the pandas case as well? (I didn't check, but I remember implementing a bunch of helper private methods)

src/narwhals/_utils.py
```diff
@@ -1263,6 +1263,33 @@ def generate_temporary_column_name(
...
+) -> list[tuple[str, Any, str]]:
+    result = []
+    for value in values:
+        for on_value in on_columns:
+            pivot_name = str(on_value)
+            output_name = (
+                separator.join((value, pivot_name)) if len(values) > 1 else pivot_name
+            )
+            result.append((value, on_value, output_name))
+    return result
```
FBruzzesi, line 1277: Disclaimer: writing directly in github, I didn't test the formatting/linting.

You can do this is a list comprehension. I tend to like them and since some python version have (much) better performance.

<details><summary> Deprecated in favor of next comment </summary>

```suggestion
	multiple_values = len(values) > 1
	return [
		(
			value,
			on_value,
			separator.join((value, str(on_value))) if multiple_values else str(on_value),
		)
		for value in values
		for on_value in on_columns
	]
```

</details>

src/narwhals/dataframe.py
```diff
@@ -3296,6 +3297,86 @@ def lazy(self) -> Self:
...
+        Examples:
+            >>> import polars as pl
+            >>> import narwhals as nw
+            >>> data = {
+            ...     "ix": [1, 1, 2, 2, 1, 2],
+            ...     "col": ["a", "a", "a", "a", "b", "b"],
+            ...     "foo": [0, 1, 2, 2, 7, 1],
+            ...     "bar": [0, 2, 0, 0, 9, 4],
+            ... }
+            >>> df_native = pl.LazyFrame(data)
```
FBruzzesi, line 3368: Can we showcase duckdb instead?

src/narwhals/_dask/dataframe.py
FBruzzesi, line 1: Making it work for dask definitely got you bonus points 🤣 🔥

src/narwhals/_ibis/dataframe.py
```diff
@@ -407,6 +414,53 @@ def unpivot(
...
+        output_names = [
+            output_name
+            for _, _, output_name in generate_pivot_column_names(
+                on_columns, values, separator=separator
+            )
+        ]
```
FBruzzesi, line 461: nitpick: you can keep this as a generator:


```suggestion
        output_names = (
            output_name
            for _, _, output_name in generate_pivot_column_names(
                on_columns, values, separator=separator
            )
        )
```

I am extremely surprise that the output naming pattern aligns perfectly! Or am I missing something?

src/narwhals/_polars/dataframe.py
```diff
@@ -764,6 +765,30 @@ def collect(
         msg = f"Unsupported `backend` value: {backend}"  # pragma: no cover
         raise ValueError(msg)  # pragma: no cover
 
+    @requires.backend_version((1, 43))
```
FBruzzesi, line 771: Should we use the old snippet in `polars.DataFrame.pivot` for the previous versions?

From _old_ polars docs:

> Note that pivot is only available in eager mode. If you know the unique column values in advance, you can use [polars.LazyFrame.groupby()](https://docs.pola.rs/api/python/version/0.20/reference/lazyframe/api/polars.LazyFrame.groupby.html#polars.LazyFrame.groupby) to get the same result as above in lazy mode:

```py
index = pl.col("col1")
columns = pl.col("col2")
values = pl.col("col3")
unique_column_values = ["x", "y"]
aggregate_function = lambda col: col.tanh().mean()
df.lazy().group_by(index).agg(
    aggregate_function(values.filter(columns == value)).alias(value)
    for value in unique_column_values
).collect()  
```

src/narwhals/_dask/dataframe.py
```diff
@@ -533,6 +540,51 @@ def unpivot(
...
+    ) -> Self:
+        if maintain_order:
+            msg = "Dask does not support maintaining row order during a pivot."
+            raise NotImplementedError(msg)
+        if aggregate_function is None or aggregate_function == "item":
+            msg = (
+                "Dask does not support pivoting without aggregation because it "
+                "cannot validate that each group contains a single value."
+            )
+            raise NotImplementedError(msg)
```
FBruzzesi, line 563: You can probably factor this out into a helper function as it's used in 3-4 places. The best name I could think of is `check_lazy_pivot_supported`, but feel free to come up with anything better:

```py
def check_lazy_pivot_supported(
    aggregate_function: LazyPivotAgg | None,
    *,
    maintain_order: bool,
) -> PivotAgg:
    """Reject the pivot options no lazy backend other than Polars can honour."""
    if maintain_order:
        msg = f"`maintain_order=True` is not supported for {implementation}."
    if aggregate_function is None or aggregate_function == "item":
        msg = (
            f"`aggregate_function={aggregate_function!r}` is not supported for "
            f"{implementation}, as it cannot validate that each group contains a "
            "single value."
        )
    return aggregate_function
```

src/narwhals/_spark_like/dataframe.py
```diff
@@ -561,6 +563,65 @@ def unpivot(
...
+            result = (
+                self.native.groupBy(*index).agg(*aggregations)
+                if index
+                else self.native.agg(*aggregations)
+            )
+            return self._with_native(result)
+
+        aggregations = [
+            getattr(self._F, aggregate_function)(value).alias(value) for value in values
+        ]
```
FBruzzesi, line 610: Didn't test it, but would the following work:

```py
aggregations = [
	(
    	self._F.count(self._F.lit(1)) if agg == "len" else getattr(self._F, agg)(value)
	).alias(value)
    for value in values
]
```

to completely remove the `if aggregate_function == "len":` block?

src/narwhals/_utils.py
FBruzzesi: On a second thought on this function, I realized that there are multiple cases in which also the native name is used and can be parsed together. I would consider creating a data structure (named tuple/dataclass) for the following:

```py
class PivotColumn(NamedTuple):
    value: str
    on_value: Any
    native_name: str
    """Name that DuckDB's `PIVOT` and Spark's `groupBy(...).pivot(...)` give this column."""
    output_name: str
    """Name Narwhals gives this column, matching Polars."""


def generate_pivot_column_names(
    on_columns: Sequence[Any], values: Sequence[str], *, separator: str
) -> list[PivotColumn]:
    single_value = len(values) == 1
    return [
        PivotColumn(
            value,
            on_value,
            str(on_value) if single_value else f"{on_value}_{value}",
            str(on_value) if single_value else separator.join((value, str(on_value))),
        )
        for value in values
        for on_value in on_columns
    ]
```

and then accessing the various attributes as needed for internal use only.

src/narwhals/_spark_like/dataframe.py
```diff
@@ -561,6 +563,65 @@ def unpivot(
             unpivoted_native_frame = unpivoted_native_frame.drop(*ids)
         return self._with_native(unpivoted_native_frame)
 
+    def pivot(
```
FBruzzesi, line 566: Nitpick: assign `F = self._F` once rather than accessing it each time

FBruzzesi (commented):
Thanks @CangyuanLi - this looks like a great first iteration. I left a few comments but nothing major, there are a lot of nitpicks and a few suggestions, some of which are code that I didn't have the chance to try directly so please double check if they are viable options.

---

On the inline comments, I assume they are AI generated and suggested to trim them quite a bit

---

**Update 2026-08-23**

I forgot to comment on the decisions in the PR description: I think they are totally valid and fair, although it can be considered a simplification, I believe it already bring us a long way. We can revisit if there are actual use cases.

On the typing issue: I am not the best person to comment (and I am currently doing so without an IDE), but that's definitely needs to be taken care of 🙈

**Update 2026-08-24**

RE: Typing

* For the test-plugin extension is enough to add `pivot = not_implemented()`
* For Dask, you can do the same as in https://github.com/narwhals-dev/narwhals/pull/3619 i.e. casting the result of a group_by + agg to a `dx.DataFrame`
* For the `PivotAgg`, I would pick one of the following:
	1. Do a prerequisite PR to add `"item"` to `PivotAgg` also for the eager case
	2. Ignore `"item"` case for now and add its support in a follow up PR

src/narwhals/_ibis/dataframe.py
CangyuanLi: It does, except for when one of the on_columns is an empty string and there are multiple values (would not have found without AI lol). I modified the code ot handle that case.

```python
import ibis
import polars as pl

data = {
    "id": [1, 1],
    "key": ["", "a"],
    "foo": [1, 2],
    "bar": [3, 4],
}

polars_result = (
    pl.LazyFrame(data)
    .pivot(
        "key",
        on_columns=["", "a"],
        index="id",
        values=["foo", "bar"],
        aggregate_function="sum",
    )
    .collect()
)

ibis_result = ibis.memtable(data).pivot_wider(
    id_cols="id",
    names_from="key",
    names=["", "a"],
    values_from=["foo", "bar"],
    values_agg="sum",
)

print("Polars columns:", polars_result.columns)
print("Ibis columns:  ", ibis_result.columns)
```

src/narwhals/_polars/dataframe.py
CangyuanLi: implemented!

src/narwhals/_utils.py
CangyuanLi: The pandas helpers seem more complex, I guess because they need to handle things like indices and multiple on columns? I can take a stab if you would like though!

src/narwhals/dataframe.py
CangyuanLi: done!

src/narwhals/_dask/dataframe.py
CangyuanLi: done, I decided to call it check_lazy_pivot_supported_options

CangyuanLi:
> Thanks @CangyuanLi - this looks like a great first iteration. I left a few comments but nothing major, there are a lot of nitpicks and a few suggestions, some of which are code that I didn't have the chance to try directly so please double check if they are viable options.
> 
> On the inline comments, I assume they are AI generated and suggested to trim them quite a bit
> 
> **Update 2026-08-23**
> 
> I forgot to comment on the decisions in the PR description: I think they are totally valid and fair, although it can be considered a simplification, I believe it already bring us a long way. We can revisit if there are actual use cases.
> 
> On the typing issue: I am not the best person to comment (and I am currently doing so without an IDE), but that's definitely needs to be taken care of 🙈
> 
> **Update 2026-08-24**
> 
> RE: Typing
> 
>     * For the test-plugin extension is enough to add `pivot = not_implemented()`
> 
>     * For Dask, you can do the same as in #3619 i.e. casting the result of a group_by + agg to a `dx.DataFrame`
> 
>     * For the `PivotAgg`, I would pick one of the following:
>       
>       1. Do a prerequisite PR to add `"item"` to `PivotAgg` also for the eager case
>       2. Ignore `"item"` case for now and add its support in a follow up PR

Thank you for the review!! I think I got to all your comments. Regarding typing, I ended up removing item from the lazy pivot since only polars supported it. If I get a burst of time / motivation I'll probably try and add it to both eager and lazy in the future haha. The typing is still not fully resolved because of `on_columns`, which does not exist in eager but must exist in lazy pivot. I think one way of handling this might be to continue omitting on_columns in the public API for eager, but requiring it in the internal API and just having it do nothing for eager. Alternatively, in polars, on_columns is now part of the eager API as well, so we could just add on_columns to eager properly.

src/narwhals/_utils.py
FBruzzesi: I am happy to keep that as a follow up

FBruzzesi:
Hey @CangyuanLi thanks for bearing with me and taking care of all the points. I will think about how to tackle the signature difference between eager and lazy pivot a bit more. Introducing `on_columns` to the eager case is an option, I didn't check it directly, but my concern is to add a lot of complexity (but maybe that's not the case). If that's the best solution, I would propose to upstream it before this one

FBruzzesi:
Ok, I _think_ there might be a much easier way for the time being and to be able to move forward with this feature while making the type checkers happy.

If we rename the compliant method to `_lazy_pivot` this would allow us to avoid the intersection with eager `pivot` and use such method to dispatch from `LazyFrame` to the compliant layers. The reason for making it private is because I would rather not expose it for plugins until we align the signature of the two pivot methods, by introducing `on_columns` for the eager case.

---

@MarcoGorelli not sure if you have a preference between:

* Moving on with this feature, **then** align signature and expose `pivot` to (lazy) plugins
* Wait until `on_columns` is available in the eager case and then ship this one

MarcoGorelli:
i think it might be good to add `on_columns` to eager first (or, as part of this pr?)

PR #3843, opened 2026-08-09
feat: add `str.strip_chars_{start,end}`

### LINKED ISSUE #3804: [Enh]: Add str.strip_chars_start / strip_chars_end

_No response_

Although stripping from start / end should be expressible via regex, it is more convenient and performant to be able to dispatch to dedicated functions. 

My specific use-case that inspired this request was porting over some Snowflake code that used LTRIM.

polars: strip_chars_start / strip_chars_end
pandas: : lstrip / rstrip
pyarrow: utf8_ltrim / utf8_rtrim
duckdb: ltrim / rtrim
pyspark: seems like for older spark versions a regex-fallback is required, but for 3.3+ has ltrim / rtrim
ibis: lstrip / rstrip only do whitespace, so need a regex fallback

I would be willing to open a PR

_No response_

_No response_

CangyuanLi:
# Description

Add strip_chars_start and strip_chars_end. For now, I have followed the pattern of strip_chars and have not aligned whitespace handling across backends. For example, for strip_chars, the existing sql / duckdb implementation uses string.whitespace. But polars uses an expanded set of whitespace characters, so behavior between duckdb and polars may be different. I preferred aligning with the existing narwhals behavior in this PR, but maybe it would be better to update all strip_chars functions to align with polars?

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue https://github.com/narwhals-dev/narwhals/issues/3804
- Closes https://github.com/narwhals-dev/narwhals/issues/3804

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes

    ```bash
    ```


src/narwhals/_spark_like/expr_str.py
```diff
@@ -14,6 +16,26 @@
 
 
 class SparkLikeExprStringNamespace(SQLExprStringNamespace["SparkLikeExpr"]):
+    def _strip_chars(self, characters: str | None, *, start: bool) -> SparkLikeExpr:
+        characters = string.whitespace if characters is None else characters
+        if not characters:  # pragma: no cover
+            # empty string means that no characters should be stripped, so we just
+            # early return
+            return self.compliant
```
FBruzzesi: I wonder if it's worth to upstream this bit to the narwhals level so that:

1. early return is shared among all backends
2. "parsing" of the characters is centralized. This will also help to avoid any drift in default behavior in a given backend
3. then compliant level would receive `characters: str` rather than `characters: str | None`

FBruzzesi (commented):
Thanks @CangyuanLi this looks great! I left a tiny comment to possibly upstream characters parsing to the narwhals level. Aside that I would consider this ready to merge!

src/narwhals/_spark_like/expr_str.py
CangyuanLi: Thank you for the review!! Made those fixes

src/narwhals/_sql/expr_str.py
FBruzzesi, line 1: As both ibis and spark-like have their own implementation, this is used only for the duckdb case. I would rather move the implementation in the duckdb namespace itself.

src/narwhals/expr_str.py
```diff
@@ -136,6 +137,56 @@ def strip_chars(self, characters: str | None = None) -> ExprT:
...
+            >>> import narwhals as nw
+            >>> df_native = pl.DataFrame({"fruits": [" apple ", "\nmango"]})
+            >>> df = nw.from_native(df_native)
+            >>> df.with_columns(
+            ...     stripped=nw.col("fruits").str.strip_chars_start()
+            ... ).to_dict(as_series=False)
+            {'fruits': [' apple ', '\nmango'], 'stripped': ['apple ', 'mango']}
+        """
+        characters = parse_str_strip_chars(characters)
+        if characters == "":
```
FBruzzesi: Nitpick: we use walrus extensively:

```suggestion
        if (characters := parse_str_strip_chars(characters)) == "":
```

src/narwhals/expr_str.py
```diff
@@ -136,6 +137,56 @@ def strip_chars(self, characters: str | None = None) -> ExprT:
...
+            >>> import narwhals as nw
+            >>> df_native = pl.DataFrame({"fruits": [" apple ", "mango\n"]})
+            >>> df = nw.from_native(df_native)
+            >>> df.with_columns(stripped=nw.col("fruits").str.strip_chars_end()).to_dict(
+            ...     as_series=False
+            ... )
+            {'fruits': [' apple ', 'mango\n'], 'stripped': [' apple', 'mango']}
+        """
+        characters = parse_str_strip_chars(characters)
+        if characters == "":
```
FBruzzesi: Same as above

```suggestion
        if (characters := parse_str_strip_chars(characters)) == "":
```

src/narwhals/series_str.py
```diff
@@ -110,6 +111,62 @@ def strip_chars(self, characters: str | None = None) -> SeriesT:
...
+            >>> s.str.strip_chars_start().to_native()  # doctest: +NORMALIZE_WHITESPACE
+            shape: (2,)
+            Series: '' [str]
+            [
+                    "apple "
+                    "mango"
+            ]
+        """
+        characters = parse_str_strip_chars(characters)
+        if characters == "":
```
FBruzzesi:
```suggestion
        if (characters := parse_str_strip_chars(characters)) == "":
```

src/narwhals/series_str.py
```diff
@@ -110,6 +111,62 @@ def strip_chars(self, characters: str | None = None) -> SeriesT:
...
+            >>> s.str.strip_chars_end().to_native()  # doctest: +NORMALIZE_WHITESPACE
+            shape: (2,)
+            Series: '' [str]
+            [
+                    " apple"
+                    "mango"
+            ]
+        """
+        characters = parse_str_strip_chars(characters)
+        if characters == "":
```
FBruzzesi:
```suggestion
        if (characters := parse_str_strip_chars(characters)) == "":
```

FBruzzesi (commented):
Thanks @CangyuanLi - I made another pass, I think we need just one more tiny iteration before merging 🙏🏼 bear with me 

One more ask on top of the inline comments: can we use a backend different from polars in _some_ docstring example? I know in the expression case we do `.to_dict(as_series=False)` before "displaying" the result, but if we were to skip that,  duckdb should show up nicely.

src/narwhals/_sql/expr_str.py
CangyuanLi: makes sense, moved!

CangyuanLi:
> Thanks @CangyuanLi - I made another pass, I think we need just one more tiny iteration before merging 🙏🏼 bear with me
> 
> One more ask on top of the inline comments: can we use a backend different from polars in _some_ docstring example? I know in the expression case we do `.to_dict(as_series=False)` before "displaying" the result, but if we were to skip that, duckdb should show up nicely.

I changed the strip_chars_start example to use the duckdb backend instead

FBruzzesi (approved):
Thanks a ton @CangyuanLi

PR #3844, opened 2026-08-10
feat(ibis): implement skew

CangyuanLi:
# Description

Implements skew for the ibis backend using a numerically stable two pass algorithm (an adaption of the one found in the polars docs here: https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.skew.html).

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Related issue https://github.com/ibis-project/ibis/issues/11176
- Closes 

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes

    ```bash
    ```


src/narwhals/_ibis/expr.py
```diff
@@ -212,6 +212,46 @@ def __neg__(self) -> Self:
         neg = cast("Callable[..., ir.Value]", operator.neg)
         return self._with_callable(neg)
 
+    def skew(self) -> Self:
+        def func(expr: ir.Value) -> ir.Value:
+            # There is a one-pass version that utilizes an identify for the variance
```
MarcoGorelli: identify?

src/narwhals/_ibis/expr.py
```diff
@@ -212,6 +212,46 @@ def __neg__(self) -> Self:
...
+            # var(x) = E[x^2] - E[x]^2. But this suffers from catastrophic cancellation
+            # for values with a large offset but small spread. So we instead use a more
+            # stable two-pass version. However, SQL engines don't allow nested
+            # reductions, we have to materialize the first pass mean :(.
```
MarcoGorelli: i don't think we should be materialising values within expressions

is there any way that this could be worked around within ibis?

MarcoGorelli (commented):
thanks for your PR!

src/narwhals/_ibis/expr.py
CangyuanLi: i meant identity haha

src/narwhals/_ibis/expr.py
CangyuanLi: The trouble I ran into is that I wasn't able to nest aggregations like AVG(AVG()). I'm not aware of a way to work around it, and I don't think implementing it at the ibis level would help either. From my understanding its a fundamental limitation of a lot of SQL engines. There is a one-pass version as the comment mentions, but its less numerically stable. Here's an implementation and a script showing the issue.

```
import duckdb
import ibis
import polars as pl
import pyarrow as pa

import narwhals as nw

SAMPLES = {
    "well-behaved": [1.0, 2.0, 3.0, 2.0, 1.0],
    "moderate cancellation": [10_001.0, 10_002.0, 10_003.0, 10_002.0, 10_001.0],
    "severe cancellation": [
        1_000_001.0,
        1_000_002.0,
        1_000_003.0,
        1_000_002.0,
        1_000_001.0,
    ],
}


def single_pass_skew(column):
    column = column.cast("float64")
    first_raw_moment = column.mean()
    second_raw_moment = (column**2).mean()
    third_raw_moment = (column**3).mean()
    second_central_moment = second_raw_moment - first_raw_moment**2
    third_central_moment = (
        third_raw_moment
        - 3 * first_raw_moment * second_raw_moment
        + 2 * first_raw_moment**3
    )
    return third_central_moment / second_central_moment**1.5


def execute_single_pass(values):
    table = ibis.memtable({"value": values})
    result = table.aggregate(skew=single_pass_skew(table.value)).execute()
    return float(result.iloc[0, 0])


def execute_two_pass(values):
    table = ibis.memtable({"value": values})
    result = (
        nw.from_native(table)
        .select(nw.col("value").skew())
        .to_native()
        .execute()
    )
    return float(result.iloc[0, 0])


def execute_duckdb(values):
    relation = duckdb.from_arrow(pa.table({"value": values}))
    result = (
        nw.from_native(relation)
        .select(nw.col("value").skew())
        .to_native()
        .fetchone()
    )
    return float(result[0])


def comparison() -> pl.DataFrame:
    rows = []
    for sample, values in SAMPLES.items():
        series = pl.Series(values)
        rows.append(
            {
                "sample": sample,
                "data": values,
                "polars": float(series.skew(bias=True)),
                "ibis_single_pass": execute_single_pass(values),
                "ibis_two_pass": execute_two_pass(values),
                "duckdb": execute_duckdb(values),
            }
        )
    return pl.DataFrame(rows)


if __name__ == "__main__":
    print(comparison())
```

```
shape: (3, 6)
┌─────────────────┬─────────────────┬──────────┬────────────────┬───────────────┬─────────────┐
│ sample          ┆ data            ┆ polars   ┆ ibis_single_pa ┆ ibis_two_pass ┆ duckdb      │
│ ---             ┆ ---             ┆ ---      ┆ ss             ┆ ---           ┆ ---         │
│ str             ┆ list[f64]       ┆ f64      ┆ ---            ┆ f64           ┆ f64         │
│                 ┆                 ┆          ┆ f64            ┆               ┆             │
╞═════════════════╪═════════════════╪══════════╪════════════════╪═══════════════╪═════════════╡
│ well-behaved    ┆ [1.0, 2.0, …    ┆ 0.343622 ┆ 0.343622       ┆ 0.343622      ┆ 0.343622    │
│                 ┆ 1.0]            ┆          ┆                ┆               ┆             │
│ moderate        ┆ [10001.0,       ┆ 0.343622 ┆ 0.343724       ┆ 0.343622      ┆ 0.343025    │
│ cancellation    ┆ 10002.0, …      ┆          ┆                ┆               ┆             │
│                 ┆ 10001.0]        ┆          ┆                ┆               ┆             │
│ severe          ┆ [1.000001e6,    ┆ 0.343622 ┆ 610.98672      ┆ 0.343622      ┆ 1955.029638 │
│ cancellation    ┆ 1.000002e6, …   ┆          ┆                ┆               ┆             │
│                 ┆ 1.0…            ┆          ┆                ┆               ┆             │
└─────────────────┴─────────────────┴──────────┴────────────────┴───────────────┴─────────────
```

to be fair, duckdb's native skewness function also suffers from catastrophic cancellation. surprisingly I couldn't find any issues or documentation warning about this. I definitely would prefer the two-pass version, maybe with a docstring or performance warning that for some backends you are going to have to pay the cost of double-materialization for accuracy? My hope is that skew is such rarely-needed functionality that this would be OK(I never thought about it until a teammate needed it for some population distance metric I never heard of).

CangyuanLi: The current two-pass implementation generates SQL like this--so still a single query actually. 

```
SELECT
  CASE
    WHEN COUNT("t0"."value") = 0
    THEN NULL
    WHEN COUNT("t0"."value") = 1
    THEN CAST('nan' AS DOUBLE)
    WHEN COUNT("t0"."value") = 2
    THEN 0.0
    ELSE (
      (
        AVG(
          POWER(
            (
              "t0"."value" - (
                SELECT
                  *
                FROM (
                  SELECT
                    AVG("t1"."_narwhals_skew_input") AS "_mean"
                  FROM (
                    SELECT
                      "t0"."value" AS "_narwhals_skew_input"
                    FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                  ) AS "t1"
                ) AS "t2"
              )
            ),
            3
          )
        ) - (
          AVG(
            "t0"."value" - (
              SELECT
                *
              FROM (
                SELECT
                  AVG("t1"."_narwhals_skew_input") AS "_mean"
                FROM (
                  SELECT
                    "t0"."value" AS "_narwhals_skew_input"
                  FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                ) AS "t1"
              ) AS "t2"
            )
          ) * 3 * AVG(
            POWER(
              (
                "t0"."value" - (
                  SELECT
                    *
                  FROM (
                    SELECT
                      AVG("t1"."_narwhals_skew_input") AS "_mean"
                    FROM (
                      SELECT
                        "t0"."value" AS "_narwhals_skew_input"
                      FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                    ) AS "t1"
                  ) AS "t2"
                )
              ),
              2
            )
          )
        )
      ) + (
        (
          POWER(
            AVG(
              "t0"."value" - (
                SELECT
                  *
                FROM (
                  SELECT
                    AVG("t1"."_narwhals_skew_input") AS "_mean"
                  FROM (
                    SELECT
                      "t0"."value" AS "_narwhals_skew_input"
                    FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                  ) AS "t1"
                ) AS "t2"
              )
            ),
            3
          )
        ) * 2
      )
    ) / (
      POWER(
        (
          AVG(
            POWER(
              (
                "t0"."value" - (
                  SELECT
                    *
                  FROM (
                    SELECT
                      AVG("t1"."_narwhals_skew_input") AS "_mean"
                    FROM (
                      SELECT
                        "t0"."value" AS "_narwhals_skew_input"
                      FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                    ) AS "t1"
                  ) AS "t2"
                )
              ),
              2
            )
          ) - (
            POWER(
              AVG(
                "t0"."value" - (
                  SELECT
                    *
                  FROM (
                    SELECT
                      AVG("t1"."_narwhals_skew_input") AS "_mean"
                    FROM (
                      SELECT
                        "t0"."value" AS "_narwhals_skew_input"
                      FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
                    ) AS "t1"
                  ) AS "t2"
                )
              ),
              2
            )
          )
        ),
        1.5
      )
    )
  END AS "value"
FROM "ibis_pandas_memtable_7l3iqqyginavrbnfm5lbzcvu5m" AS "t0"
```

FBruzzesi: There is an issue in group-by/over contexts by using the `origin = input_table.aggregate(_mean=input_column.mean())["_mean"].as_scalar()` approach as it creates a detached relation rooted at the whole base table

<details><summary>Repro</summary>
<p>


```py
import ibis
import narwhals as nw

t = ibis.memtable({
    "g": ["a"] * 5 + ["b"] * 5,
    "value": [
        1e9 + 1,
        1e9 + 2,
        1e9 + 3,
        1e9 + 2,
        1e9 + 1,
        -1e9 + 1,
        -1e9 + 2,
        -1e9 + 3,
        -1e9 + 2,
        -1e9 + 1
    ],
})
df = nw.from_native(t)

# true skew for both groups is 0.343622 (same shape, different offset)
print(df.group_by("g").agg(nw.col("value").skew()).to_native().execute())
#    g        value
# 0  b          NaN
# 1  a -134217728.0

print(df.with_columns(nw.col("value").skew().over("g")).to_native().execute())
# every row: -134217728.0 or NaN
```

</p>
</details> 

---

I would be in favor of initially accepting a single-pass formula with numerical instability together with:

1. Testing + documenting which backends suffer from this issue
2. Possibly raise a feature request upstream

> [!IMPORTANT]
> Maybe the solution/fix to the above issue is simpler than what I imagine 

Let me know what your thoughts are 🙏🏼

CangyuanLi: Ah thats a good find, I didn't even consider the group by case. Looking at the rank implementation for ibis, it seems like there's support for taking a different path if we are in an over context using `window_f`. But this wouldn't work for group by is my understanding, as we cannot detect that we are in a group-by context. I couldn't find a way to make it work without explicitly knowing what context we are in :(. 

I think I can implement it similar to rank where skew is only allowed in over or un-aggregated usage but is rejected by group-by. But I can also implement the single-pass version and maybe raise a warning + add documentation for the numerically unstable backends. Let me know what you would prefer!

FBruzzesi: Yes that's correct. For the `over` case, we can just add:

```py
def window_f(_df: IbisLazyFrame, _inputs: IbisWindowInputs) -> Sequence[ir.Value]:
	raise NotImplementedError(...)
```

Detecting a group by can't be done. However, what we could do is to "track" the information in the expression, and check for it in `.agg`

<details><summary>Sketched diff </summary>
<p>

**src/narwhals/_ibis/expr.py b/src/narwhals/_ibis/expr.py**:

```diff
diff --git a/src/narwhals/_ibis/expr.py b/src/narwhals/_ibis/expr.py
index 384f2ce8f..219f8afc8 100644
--- a/src/narwhals/_ibis/expr.py
+++ b/src/narwhals/_ibis/expr.py
@@ -52,8 +52,30 @@ if TYPE_CHECKING:
     IbisWindowInputs = WindowInputs[ir.Value]
 
 
+def unsupported_partitioned_reduction_msg(name: str, /) -> str:
+    """Message for a reduction whose `group_by`/`.over()` support is blocked."""
+    return (
+        f"`{name}` is not supported for the Ibis backend inside "
+        "`group_by(...).agg(...)` or `.over(...)`: its computation needs an "
+        "intermediate value derived from the whole frame, which can't be scoped "
+        "to a partition without materializing it per group."
+    )
+
+
 class IbisExpr(SQLExpr["IbisLazyFrame", "ir.Value"]):
     _implementation = Implementation.IBIS
+    # Set (to the reduction's name, e.g. "skew") on the result of a reduction
+    # whose flat `_call` is only correct outside `group_by`/`.over()` (see
+    # `unsupported_partitioned_reduction_msg`). `.over()` is caught by the
+    # reduction's own `window_function`; `group_by(...).agg(...)` never calls
+    # `window_function`, so `IbisGroupBy.agg` checks this attribute explicitly.
+    _partitioned_reduction_name: str | None = None
 
     def __init__(
         self,
@@ -88,6 +110,16 @@ class IbisExpr(SQLExpr["IbisLazyFrame", "ir.Value"]):
 
         return self._window_function or default_window_func
 
+    def _with_alias_output_names(self, func: AliasNames | None, /) -> Self:
+        # `SQLExpr._with_alias_output_names` (used by e.g. `.alias()`, `.name.*`)
+        # builds a fresh instance and doesn't know about
+        # `_partitioned_reduction_name`; without carrying it forward,
+        # `skew().alias(...)` inside `group_by(...).agg(...)` would silently
+        # skip the guard in `IbisGroupBy.agg`.
+        result = super()._with_alias_output_names(func)
+        result._partitioned_reduction_name = self._partitioned_reduction_name
+        return result
+
     def _window_expression(
         self,
         expr: ir.Value,
@@ -251,7 +283,15 @@ class IbisExpr(SQLExpr["IbisLazyFrame", "ir.Value"]):
                 else_=biased_population_skewness,
             )
 
-        return self._with_callable(func)
+        def window_f(_df: IbisLazyFrame, _inputs: IbisWindowInputs) -> Sequence[ir.Value]:
+            raise NotImplementedError(unsupported_partitioned_reduction_msg("skew"))
+
+        result = self._with_callable(func, window_f)
+        result._partitioned_reduction_name = "skew"
+        return result
 
     def quantile(
         self, quantile: float, interpolation: RollingInterpolationMethod
```

**src/narwhals/_ibis/group_by.py b/src/narwhals/_ibis/group_by.py**
```diff
diff --git a/src/narwhals/_ibis/group_by.py b/src/narwhals/_ibis/group_by.py
index 9b4d25a68..2bfefe84d 100644
--- a/src/narwhals/_ibis/group_by.py
+++ b/src/narwhals/_ibis/group_by.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from narwhals._ibis.expr import unsupported_partitioned_reduction_msg
 from narwhals._sql.group_by import SQLGroupBy
 
 if TYPE_CHECKING:
@@ -26,6 +27,13 @@ class IbisGroupBy(SQLGroupBy["IbisLazyFrame", "IbisExpr", "ir.Value"]):
         self._compliant_frame = frame.drop_nulls(self._keys) if drop_null_keys else frame
 
     def agg(self, *exprs: IbisExpr) -> IbisLazyFrame:
+        for expr in exprs:
+            if expr._partitioned_reduction_name is not None:
+                raise NotImplementedError(
+                    unsupported_partitioned_reduction_msg(
+                        expr._partitioned_reduction_name
+                    )
+                )
         native = self.compliant.native
         return self.compliant._with_native(
             native.group_by(self._keys).aggregate(*self._evaluate_exprs(exprs))
```

</p>
</details>

@MarcoGorelli I will let you weigh on this one - recently I have been facing a bunch of issue with numerical stability/precision issues on other libraries, so I share the feeling of not wanting to ship something known a priori to be unstable

MarcoGorelli:
thanks for your pr

could this be solved by addressing https://github.com/ibis-project/ibis/issues/11176 and implementing skew in ibis? has a pr been opened over there?

PR #3854, opened 2026-08-14
fix: `is_in` propagates nulls and ignores nulls in the member collection

### LINKED ISSUE #3851: [Bug]: is_in emits False for null input on pyarrow but null on polars, changing filter results
## Describe the bug

`is_in` disagrees between polars and pyarrow on what a **null input** produces: polars emits null, pyarrow emits `False`. `filter(~col.is_in(...))` therefore returns a different number of rows depending on the backend.

Unlike #3850 this is not the pandas-numpy null representation — the declared narwhals dtype is `Int64` on every backend below, and `pandas[pyarrow]` (a fully nullable backend) sides with pyarrow.

```python
import pandas as pd
import polars as pl
import pyarrow as pa
import narwhals as nw

data = {"a": [1, 2, None, 4]}

for name, native in {
    "polars": pl.DataFrame(data),
    "pyarrow": pa.table(data),
    "pandas[pyarrow]": pd.DataFrame(data).convert_dtypes(dtype_backend="pyarrow"),
}.items():
    mask = df.select(nw.col("a").is_in([1, 2]).alias("m"))["m"].to_list()
    filt = df.filter(~nw.col("a").is_in([1, 2]))["a"].to_list()
    print(f"{name:16} mask={mask}  filter(~is_in)={filt}")
```

## Actual results

```
polars           mask=[True, True, None,  False]  filter(~is_in)=[4]
pyarrow          mask=[True, True, False, False]  filter(~is_in)=[None, 4]
pandas[pyarrow]  mask=[True, True, False, False]  filter(~is_in)=[<NA>, 4]
```

One row difference in the filter, silently.

## Expected results

Consistent across backends. Following polars, which narwhals generally takes as the reference semantics:

```
mask=[True, True, None, False]   filter(~is_in)=[4]
```

Null means "unknown", so whether an unknown value is in `[1, 2]` is also unknown, and an unknown mask value should not survive `filter` under either polarity.

## Notes on fixing it

pyarrow cannot express this natively. `SetLookupOptions` only offers `skip_nulls`, and neither setting emits a null:

```
skip_nulls=False (default)  -> null input matched like a regular value -> False
skip_nulls=True             -> null input always fails matching        -> False
```

So matching polars means restoring the null explicitly after the lookup, roughly:

```python
pc.if_else(pc.is_null(arr), pa.scalar(None, pa.bool_()), pc.is_in(arr, value_set=...))
```

which also makes the result dtype nullable boolean, as polars already gives.

## Current test coverage

`tests/expr_and_series/is_in_test.py` has no null cases — `test_expr_is_in`, `test_expr_is_in_empty_list`, `test_ser_is_in`, `test_is_in_other` and `test_filter_is_in_with_series` all use non-null data. So the current divergence is unpinned rather than deliberate, which is why I am filing rather than assuming it is a documented trade-off. If it turns out to be intentional, a test asserting the pyarrow behaviour would be worth adding either way so it stops being accidental.

I am happy to open a PR for the pyarrow-side fix plus null cases in the test file, if you agree polars is the semantics to match here.

## narwhals.show_versions()

```
```

ghostiee-11:
# Description

`is_in` returned `False` for a null input on PyArrow and pandas where Polars returns null, and the backends also disagreed on whether a null in the member collection matches, so `filter(~col.is_in(...))` gave a different row count depending on the backend.

Fixed by restoring nulls after the lookup on PyArrow and pandas-like/dask, and by dropping nulls from the collection on spark-like/ibis; DuckDB was already correct and is unchanged, as is pandas with classical NumPy dtypes, which cannot hold a null in a boolean column.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3851

## AI assistance

- [x] AI tools were used.

I used Claude Code (Opus) for research and scaffolding. I chose the approach and scope, and measured every claim myself before writing it up. I can explain and defend each line in review.

## Checklist

- [x] Tests added
- [x] Documented the changes

### DIFF
--- a/docs/concepts/boolean.md
+++ b/docs/concepts/boolean.md
@@ -58,6 +58,10 @@ so we hope that the classical NumPy dtypes not supporting null values will just
 be a temporary legacy pandas issue which will eventually go
 away anyway.
 
+The same rule applies to `nw.col('a').is_in([1, 2])`: a null input gives a null result.
+A null in the collection being searched is ignored, so `nw.col('a').is_in([1, 2])` and
+`nw.col('a').is_in([1, 2, None])` always produce the same result.
+
 ```python exec="yes" source="above" session="boolean"
 from narwhals.typing import FrameT
 
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -570,7 +570,12 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
             value_set: ArrayOrChunkedArray = other
         else:
             value_set = pa.array(other)
-        return self._with_native(pc.is_in(self.native, value_set=value_set))
+        result = pc.is_in(self.native, value_set=value_set)
+        # `SetLookupOptions` can't express "null in, null out", so restore nulls
+        # ourselves. This also stops a null in `value_set` from matching null inputs.
+        return self._with_native(
+            pc.if_else(pc.is_null(self.native), lit(None, pa.bool_()), result)
+        )
 
     def arg_true(self) -> Self:
         import numpy as np  # ignore-banned-import
--- a/src/narwhals/_dask/expr.py
+++ b/src/narwhals/_dask/expr.py
@@ -16,7 +16,11 @@ from narwhals._dask.utils import (
 )
 from narwhals._expression_parsing import evaluate_nodes, evaluate_output_names_and_aliases
 from narwhals._pandas_like.expr import window_kwargs_to_pandas_equivalent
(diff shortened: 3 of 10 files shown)


MarcoGorelli:
thanks for your pr

tbh this is a bit too ai generated, and i don't think you've followed all the checklist (e.g. posting screenshot of tests passing locally)

the code comments are also way too verbose

closing for now then

PR #3855, opened 2026-08-14
fix: use exact quantile for median on the pyarrow backend

### LINKED ISSUE #3853: [Bug]: median() is approximate on the pyarrow backend and returns the lower value for two-element input
## Describe the bug

`median()` is exact on the pandas and polars backends and approximate on pyarrow, with no way to request an exact result.

`src/narwhals/_arrow/series.py:350` calls `pc.approximate_median`, and `src/narwhals/_arrow/group_by.py:36` maps `"median"` to `"approximate_median"`. The approximation comes from pyarrow, but the same `median()` call returns a different number depending on which frame the caller passed in.

For a two-element input it returns the lower value rather than the mean, deterministically and regardless of input order.

```python
import narwhals as nw
import pandas as pd
import polars as pl
import pyarrow as pa

for name, native in {
    "pandas": pd.DataFrame({"v": [5.0, 1.0]}),
    "polars": pl.DataFrame({"v": [5.0, 1.0]}),
    "pyarrow": pa.table({"v": [5.0, 1.0]}),
}.items():
    value = nw.from_native(native, eager_only=True).select(nw.col("v").median()).item()
    print(f"{name:8} {value}")
```

The same divergence appears through `group_by().agg(nw.col("v").median())`.

pyarrow's function on its own, for reference:

```python
import pyarrow as pa
import pyarrow.compute as pc
import statistics

for values in ([5.0, 1.0], [2.0, 1.0], [10.0, 20.0]):
    print(values, pc.approximate_median(pa.array(values)).as_py(), statistics.median(values))
```

## Expected results

```
pandas   3.0
polars   3.0
pyarrow  3.0
```

## Actual results

```
pandas   3.0
polars   3.0
pyarrow  1.0
```

Measured across input sizes:

| input | pyarrow `median()` | exact |
| --- | --- | --- |
| `[5.0, 1.0]` | 1.0 | 3.0 |
| `[2.0, 1.0]` | 1.0 | 1.5 |
| `[10.0, 20.0]` | 10.0 | 15.0 |
| 100 random floats | 52.234649 | 52.139905 |
| 1000 random floats | 49.489255 | 49.344791 |

n=3, 4, 5 and 10 came out exact in my testing.

## Two notes that may help

`pc.quantile(q=0.5)` is exact on pyarrow, including the two-element case, so there may be a drop-in alternative:

```python
import pyarrow as pa
import pyarrow.compute as pc

print(pc.quantile(pa.array([5.0, 1.0]), q=0.5).to_pylist()[0])  # 3.0
```

The existing tests cannot catch this. `tests/expr_and_series/median_test.py` uses `{"a": [3, 8, 2, None], "b": [5, 5, None, 7], "z": [7.0, 8.0, 9.0, None]}`, and every column there has three non-null values, so the even-length case where `approximate_median` diverges is never exercised.

I searched the issue list for `median` and `approximate_median` before filing and did not find this reported.

## Please run narwhals.show_versions() and enter the output below.

```shell
System:
    python: 3.12.13 | packaged by conda-forge | (main, Mar  5 2026, 17:06:14) [Clang 19.1.7 ]
   machine: macOS-26.2-arm64-arm-64bit

Python dependencies:
     narwhals: 2.17.0
        numpy: 2.4.2
       pandas: 3.0.1
      pyarrow: 23.0.1
       polars: 1.43.2
         dask: 2024.10.0
       duckdb: 1.4.4
```
Disclosure per CONTRIBUTING.md: I hit this while integrating narwhals into another project, and used AI assistance (Claude) to investigate it and prepare this report. Every output pasted above was produced by running the snippets against the versions listed, not reconstructed. I also see point 5 asks contributors not to have LLMs speak for them in issues, so I am flagging the assistance up front rather than leaving it implicit, and any follow-up discussion will be in my own words.

ghostiee-11:
# Description

`median()` on the PyArrow backend called `pc.approximate_median`, a t-digest estimate. For a two element input it returns the lower value rather than the mean, so `median()` of `[5.0, 1.0]` gave `1.0` where pandas and Polars give `3.0`. It also drifts on larger inputs (about 0.09 on 100 random floats, 0.14 on 1000).

Switched to `pc.quantile(q=0.5, interpolation="linear")`, which is exact. This is the same call the existing `quantile` implementation already uses a few hundred lines below, so it is not a new pattern. `None` in and empty input behave as before.

It is also faster, measured from n=2 up to n=10M:

| n | `approximate_median` | `quantile` |
| --- | --- | --- |
| 2 | 162.5 us | 18.3 us |
| 100 | 18.5 us | 8.1 us |
| 10,000 | 496.2 us | 135.6 us |
| 1,000,000 | 57.6 ms | 19.8 ms |
| 10,000,000 | 896.2 ms | 236.7 ms |

Grouped median is not addressed here. PyArrow has no exact `hash_` quantile kernel to switch to, only `hash_approximate_median` and `hash_tdigest`, so `group_by().agg(col.median())` stays approximate on that backend. Asked about reopening apache/arrow#28985 for that.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3853

## AI assistance

- [x] AI tools were used.

Claude Code was used for research and scaffolding.

## Checklist

- [x] Tests added
- [ ] Documented the changes

### DIFF
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -346,8 +346,11 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
 
+        # `approximate_median` is a t-digest estimate: for a two element input
+        # it returns the lower value rather than the mean. `quantile` is exact.
         return maybe_extract_py_scalar(
-            pc.approximate_median(self.native), _return_py_scalar
+            pc.quantile(self.native, q=0.5, interpolation="linear")[0],
+            _return_py_scalar,
         )
 
     def min(self, *, _return_py_scalar: bool = True) -> Any:
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -346,11 +346,8 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
 
-        # `approximate_median` is a t-digest estimate: for a two element input
-        # it returns the lower value rather than the mean. `quantile` is exact.
         return maybe_extract_py_scalar(
-            pc.quantile(self.native, q=0.5, interpolation="linear")[0],
-            _return_py_scalar,
+            pc.quantile(self.native, q=0.5, interpolation="linear")[0], _return_py_scalar
         )
 
     def min(self, *, _return_py_scalar: bool = True) -> Any:
--- a/tests/expr_and_series/median_test.py
+++ b/tests/expr_and_series/median_test.py
@@ -45,11 +45,13 @@ def test_median_series(
 def test_median_group_by(
     constructor: Constructor, request: pytest.FixtureRequest
 ) -> None:
-    # PyArrow has no exact hashed quantile kernel, only `hash_approximate_median` and
-    # `hash_tdigest`, so a grouped median stays approximate there.
-    # https://github.com/apache/arrow/issues/28985
     if "pyarrow_table" in str(constructor):
-        request.applymarker(pytest.mark.xfail)
+        reason = (
+            "PyArrow has no exact hashed quantile kernel, only `hash_approximate_median` "
+            " and `hash_tdigest`, so a grouped median stays approximate there. "
+            "See: https://github.com/apache/arrow/issues/28985"
+        )
+        request.applymarker(pytest.mark.xfail(reason=reason))
     df = nw.from_native(constructor(group_data))
     result = df.group_by("g").agg(nw.col("a").median()).sort("g")
     assert_equal_data(result, {"g": [1, 2], "a": [1.5, 15.0]})
--- a/tests/expr_and_series/median_test.py
+++ b/tests/expr_and_series/median_test.py
@@ -8,12 +8,15 @@ import narwhals as nw
 from narwhals.exceptions import InvalidOperationError
 from tests.utils import Constructor, ConstructorEager, assert_equal_data
 
+# `a` has an even number of non-null values, where an approximate median disagrees
+# with the exact one.
 data = {
-    "a": [3, 8, 2, None],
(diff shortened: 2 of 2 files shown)


src/narwhals/_arrow/series.py
```diff
@@ -346,8 +346,11 @@ def median(self, *, _return_py_scalar: bool = True) -> float:
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
 
+        # `approximate_median` is a t-digest estimate: for a two element input
+        # it returns the lower value rather than the mean. `quantile` is exact.
```
FBruzzesi: I think it's ok to let these kind of comments out. If anything, you can add the issue number/link

tests/expr_and_series/median_test.py
```diff
@@ -39,6 +39,34 @@ def test_median_series(
     assert_equal_data({col: [result]}, {col: [expected]})
 
 
+# Every column in `data` has an odd number of non-null values. An approximate
+# median agrees with the exact one there, so the columns below have an even
+# number of non-null values instead.
+even_data = {"a": [5.0, 1.0, None], "b": [1.0, 2.0, None], "z": [10.0, None, 20.0]}
```
FBruzzesi: Rather than adding a new set of data, I would change the `data` to make one column have an even number of non-null values.

FBruzzesi (commented):
Thanks for your contribution @ghostiee-11 

1. pre-commit hooks are failing, while the scikit-lego CI failure is unrelated
2. from the discussion in the issue, the group by / over context would not be fixed by this PR. I think it's worth to add a test case that shows that (i.e. in which pyarrow would xfail and explain the reason why that's the case)

FBruzzesi (approved):
Thanks @ghostiee-11 🚀

ghostiee-11:
Thanks @FBruzzesi for reviewing and fixing :)

PR #3856, opened 2026-08-16
fix: honor schema column selection/order in pandas-like from_dicts

### LINKED ISSUE #3837: bug: `from_dicts` schema does not select or order columns on pandas-like
## Describe the bug

`from_dicts` with an explicit `schema` treats it as the schema on Polars and PyArrow, and as a dtype hint on pandas-like. The schema does not select columns, does not order them, and does not add missing ones.

| given | polars | pyarrow | pandas |
|---|---|---|---|
| row keys in a different order than the schema | `['a', 'b']` | `['a', 'b']` | `['b', 'a']` |
| a key in the data that the schema omits | `['a', 'b']` | `['a', 'b']` | `['a', 'b', 'c']` |
| a key in the schema that the data omits | `['a', 'b']` | `['a', 'b']` | `['a']` |

```python
import narwhals as nw

cases = {
    "reordered keys":      ([{"b": 1, "a": "x"}, {"b": 2, "a": "y"}], {"a": nw.String(), "b": nw.Int64()}),
    "extra key in data":   ([{"a": "x", "b": 1, "c": 9}],             {"a": nw.String(), "b": nw.Int64()}),
    "key missing in data": ([{"a": "x"}],                             {"a": nw.String(), "b": nw.Int64()}),
}
for label, (data, schema) in cases.items():
    print(label, "schema:", list(schema))
    for backend in ("polars", "pyarrow", "pandas"):
        print("   ", backend, nw.from_dicts(data, backend=backend, schema=schema).columns)
```

## Expected results

The schema decides the columns, their order and their dtypes on every backend, as it already does on Polars and PyArrow.

## Actual results

```
reordered keys schema: ['a', 'b']
    polars ['a', 'b']
    pyarrow ['a', 'b']
    pandas ['b', 'a']
extra key in data schema: ['a', 'b']
    polars ['a', 'b']
    pyarrow ['a', 'b']
    pandas ['a', 'b', 'c']
key missing in data schema: ['a', 'b']
    polars ['a', 'b']
    pyarrow ['a', 'b']
    pandas ['a']
```

## Where it comes from

`PandasLikeDataFrame.from_dicts` builds with `DataFrame.from_records(data)`, which takes the columns from the rows, and then applies the schema with `astype`, which changes dtypes and neither reorders nor selects.

A reindex to the schema's keys between those two steps covers all three rows of the table. I can open a PR if that is the shape you want; the empty-data branch already builds from the schema, so only the populated branch needs it.

Found while widening a test in #3757 across backends, as suggested there.

## narwhals.show_versions()

```
narwhals: 2.23.0 | pandas: 2.3.3 | pyarrow: 23.0.1 | polars: 1.42.1 | duckdb: 1.5.4 | numpy: 2.2.6 | python: 3.10
```

aleks-drozy:
## Summary

Fixes #3837.

`from_dicts` with an explicit `schema` treats the schema as the source of truth for column selection, order, and dtypes on Polars and PyArrow, but on the pandas-like backend it only used the schema as a dtype hint.

Root cause: `PandasLikeDataFrame.from_dicts` builds the native frame with `DataFrame.from_records(data)`, which derives the column set/order purely from the row dicts' key order. The following `native.astype(native_schema)` only casts dtypes for whichever schema keys already happen to be present — it never selects, drops extras, adds missing columns, or reorders to match the schema's key order.

| given | polars | pyarrow | pandas (before) | pandas (after) |
|---|---|---|---|---|
| row keys in a different order than the schema | `['a', 'b']` | `['a', 'b']` | `['b', 'a']` | `['a', 'b']` |
| a key in the data that the schema omits | `['a', 'b']` | `['a', 'b']` | `['a', 'b', 'c']` | `['a', 'b']` |
| a key in the schema that the data omits | `['a', 'b']` | `['a', 'b']` | `['a']` | `['a', 'b']` |

## Fix

In the populated-data branch, `reindex` the native frame to `schema`'s columns before the existing `astype` dtype-casting step, so newly-added missing columns still get their dtype cast. The empty-data branch already builds directly from `schema`, so it needs no change.

## Test plan

- [x] Extended `tests/frame/from_dicts_test.py` with `test_from_dicts_schema_reordered_keys`, `test_from_dicts_schema_extra_key`, and `test_from_dicts_schema_missing_key`, parametrized via the `eager_backend` fixture (pandas/pyarrow/polars).
- [x] Confirmed the 6 new pandas-backend cases fail on `main` and pass after the fix; polars/pyarrow were already passing (used as the reference behavior).
- [x] `uv run pytest tests/frame -q` — 2195 passed, 292 skipped, 72 xfailed (no regressions).
- [x] `uv run ruff check` / `uv run ruff format --check` on touched files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/narwhals/_pandas_like/dataframe.py
+++ b/src/narwhals/_pandas_like/dataframe.py
@@ -207,13 +207,9 @@ class PandasLikeDataFrame(
             native = DataFrame.from_records(data)
         else:
             native = DataFrame.from_dict({col: [] for col in schema})
-        if schema and data:
-            # `DataFrame.from_records` derives columns from the row dicts'
-            # keys, ignoring `schema`. Reindex so the result matches
-            # `schema`'s column selection and order, adding any missing
-            # columns (filled with null) and dropping any extra ones.
-            native = native.reindex(columns=list(schema))
         if schema:
+            # https://github.com/narwhals-dev/narwhals/issues/3837
+            native = native.reindex(columns=list(schema))
             backends: Iterable[DTypeBackend]
             if data:
                 backends = iter_dtype_backends(native.dtypes, implementation)
--- a/src/narwhals/_pandas_like/dataframe.py
+++ b/src/narwhals/_pandas_like/dataframe.py
@@ -203,18 +203,14 @@ class PandasLikeDataFrame(
         implementation = context._implementation
         ns = implementation.to_native_namespace()
         DataFrame = cast("type[pd.DataFrame]", ns.DataFrame)
-        if data or not schema:
-            native = DataFrame.from_records(data)
-        else:
-            native = DataFrame.from_dict({col: [] for col in schema})
+        # NOTE: `DataFrame.from_records` returns *zero* rows when every mapping is
+        # empty, while the constructor keeps `len(data)` (empty) rows.
+        native = DataFrame(data)
         if schema:
-            # https://github.com/narwhals-dev/narwhals/issues/3837
+            # Missing columns become all-null, extra ones are dropped, and the schema
+            # order wins: https://github.com/narwhals-dev/narwhals/issues/3837
             native = native.reindex(columns=list(schema))
-            backends: Iterable[DTypeBackend]
-            if data:
-                backends = iter_dtype_backends(native.dtypes, implementation)
-            else:
-                backends = (None for _ in range(len(schema)))
+            backends = iter_dtype_backends(native.dtypes, implementation)
             native_schema = {
                 key: narwhals_to_native_dtype(
                     dtype,
--- a/src/narwhals/_pandas_like/dataframe.py
+++ b/src/narwhals/_pandas_like/dataframe.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from collections.abc import Callable, Iterable, Iterator, Mapping, Sequence
 from itertools import chain, product
-from typing import TYPE_CHECKING, Any, Literal, cast, overload
+from typing import TYPE_CHECKING, Any, Literal, TypedDict, cast, overload
 
 import numpy as np
 
@@ -71,6 +71,9 @@ if TYPE_CHECKING:
     )
 
     Constructor: TypeAlias = Callable[..., pd.DataFrame]
+
+    class CopyKwargs(TypedDict, total=False):
+        copy: bool
 
 
 CLASSICAL_NUMPY_DTYPES: frozenset[np.dtype[Any]] = frozenset(
@@ -207,9 +210,14 @@ class PandasLikeDataFrame(
         # empty, while the constructor keeps `len(data)` (empty) rows.
         native = DataFrame(data)
         if schema:
+            kwds: CopyKwargs = (
+                {"copy": False}
+                if implementation.is_pandas() and implementation._backend_version() < (3,)
+                else {}
+            )
             # Missing columns become all-null, extra ones are dropped, and the schema
             # order wins: https://github.com/narwhals-dev/narwhals/issues/3837
-            native = native.reindex(columns=list(schema))
+            native = native.reindex(columns=list(schema), **kwds)
             backends = iter_dtype_backends(native.dtypes, implementation)
             native_schema = {
                 key: narwhals_to_native_dtype(
@@ -221,7 +229,7 @@ class PandasLikeDataFrame(
                 for ((key, dtype), backend) in zip(schema.items(), backends, strict=False)
                 if dtype is not None
             }
-            native = native.astype(native_schema)
+            native = native.astype(native_schema, **kwds)
         return cls.from_native(native, context=context)
 
     @staticmethod
--- a/src/narwhals/_pandas_like/dataframe.py
+++ b/src/narwhals/_pandas_like/dataframe.py
@@ -207,6 +207,12 @@ class PandasLikeDataFrame(
             native = DataFrame.from_records(data)
         else:
             native = DataFrame.from_dict({col: [] for col in schema})
+        if schema and data:
+            # `DataFrame.from_records` derives columns from the row dicts'
+            # keys, ignoring `schema`. Reindex so the result matches
+            # `schema`'s column selection and order, adding any missing
+            # columns (filled with null) and dropping any extra ones.
+            native = native.reindex(columns=list(schema))
         if schema:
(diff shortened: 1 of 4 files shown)


src/narwhals/_pandas_like/dataframe.py
```diff
@@ -207,6 +207,12 @@ def from_dicts(
             native = DataFrame.from_records(data)
         else:
             native = DataFrame.from_dict({col: [] for col in schema})
+        if schema and data:
+            # `DataFrame.from_records` derives columns from the row dicts'
+            # keys, ignoring `schema`. Reindex so the result matches
+            # `schema`'s column selection and order, adding any missing
+            # columns (filled with null) and dropping any extra ones.
+            native = native.reindex(columns=list(schema))
```
FBruzzesi: Couple of comments here:

1. Shouldn't this happen regardless of having data? i.e. in the `if schema` block below?
2. Can you trim the inline comments? I think it's ok to keep, but linking the originating issue is enough

tests/frame/from_dicts_test.py
FBruzzesi, line 1: All these new tests are pretty similar, you can parametrize them into one

FBruzzesi (commented):
Thanks for your contribution @aleks-drozy 

Couple of considerations on top of the two code comments:

* Can you add test cases for:
	1. `nw.from_dicts([{}, {}], {"a": nw.Int64()}, backend=...)`?
	2. `nw.from_dicts([{"a": 1, "b": True}, {"a": 2}], {"a": nw.Int64(), "b": nw.Boolean()}, backend=...)`?
* `from_dict` has all the same defects: would you be interested in a dedicated followup?

src/narwhals/_pandas_like/dataframe.py
```diff
@@ -203,16 +203,14 @@ def from_dicts(
         implementation = context._implementation
         ns = implementation.to_native_namespace()
         DataFrame = cast("type[pd.DataFrame]", ns.DataFrame)
-        if data or not schema:
-            native = DataFrame.from_records(data)
-        else:
-            native = DataFrame.from_dict({col: [] for col in schema})
+        # NOTE: `DataFrame.from_records` returns *zero* rows when every mapping is
+        # empty, while the constructor keeps `len(data)` (empty) rows.
+        native = DataFrame(data)
```
FBruzzesi, line 211: @MarcoGorelli I couldn't find the reason why we were not using the DataFrame constructor directly.
I changed this in order to allow to preserve two (empty) rows when passing `[{}, {}]`

tests/frame/from_dicts_test.py
```diff
@@ -37,6 +37,67 @@ def test_from_dicts_schema(eager_backend: EagerAllowed) -> None:
...
+    result = nw.DataFrame.from_dicts(
+        [{"a": True}, {}], backend="pandas", schema={"a": nw.Boolean()}
+    )
+    assert_equal_data(result, {"a": [True, True]})
+
+    # integer casting is stricter, `pandas.errors.IntCastingNaNError` is surfaced as-is
+    with pytest.raises(ValueError, match="Cannot convert non-finite values"):
+        nw.DataFrame.from_dicts(
+            [{"a": 1}, {}], backend="pandas", schema={"a": nw.Int64()}
+        )
```
FBruzzesi, line 98: @MarcoGorelli thoughts on this behavior?

src/narwhals/_pandas_like/dataframe.py
MarcoGorelli: 😄 nice, simpler like this

tests/frame/from_dicts_test.py
MarcoGorelli: not sure there's much we can do here 🤷 if pandas defaulted to nullable types, then those would be used

i think this is ok here anyway

FBruzzesi: right! I tried to document the behavior a bit more. If we are ok with that, I think we are ready to merge. I will wait for your pass as I am biased due to my edits/commits 😅

src/narwhals/dataframe.py
```diff
@@ -634,8 +634,12 @@ def from_dicts(
         """Instantiate DataFrame from a sequence of dictionaries representing rows.
 
         Notes:
+            Keys missing from some (or all) rows are filled with null.
+
             For pandas-like dataframes, conversion to schema is applied after dataframe
-            creation.
+            creation, and missing keys are filled with `NaN`. A `schema` requesting a dtype
```
MarcoGorelli: I think we should say "null", as it's not always `NaN` (e.g. it could be `NaT`)

src/narwhals/_pandas_like/dataframe.py
```diff
@@ -203,16 +203,14 @@ def from_dicts(
...
         if schema:
-            backends: Iterable[DTypeBackend]
-            if data:
-                backends = iter_dtype_backends(native.dtypes, implementation)
-            else:
-                backends = (None for _ in range(len(schema)))
+            # Missing columns become all-null, extra ones are dropped, and the schema
+            # order wins: https://github.com/narwhals-dev/narwhals/issues/3837
+            native = native.reindex(columns=list(schema))
```
MarcoGorelli: can we pass `copy=False` if `backend_version < (3,)`?

same for `native = native.astype(native_schema)` a few lines down (admittedly that was already present)

src/narwhals/functions.py
```diff
@@ -337,8 +337,12 @@ def from_dicts(
     """Instantiate DataFrame from a sequence of dictionaries representing rows.
 
     Notes:
+        Keys missing from some (or all) rows are filled with null.
+
         For pandas-like dataframes, conversion to schema is applied after dataframe
-        creation.
+        creation, and missing keys are filled with `NaN`. A `schema` requesting a dtype
```
MarcoGorelli: same

MarcoGorelli (commented):
thanks both!

just some comments

src/narwhals/_pandas_like/dataframe.py
FBruzzesi: Addressed in https://github.com/narwhals-dev/narwhals/pull/3856/commits/496d6210f2098b60f1863a5e9fea841e91f7e2f7

MarcoGorelli (approved):
amazing thank you

PR #3857, opened 2026-08-16
fix: decode dictionary-encoded columns before sorting in pyarrow backend

### LINKED ISSUE #3841: [Bug]: `DataFrame.sort()` fails on dictionary-encoded pyarrow columns
### Describe the bug

Sorting a narwhals `DataFrame` wrapping a pyarrow `Table` fails with `ArrowNotImplementedError` when the sorted column is dictionary-encoded (pyarrow's standard representation for categorical-like data). The underlying `pyarrow.Table.sort_by()` compute kernel doesn't support dictionary type directly, and narwhals' arrow backend (`narwhals/_arrow/dataframe.py`, `sort()`) calls it without decoding first.

(I'm still working on a narwhals-based fast path for dataframe-like inputs in scikit-learn encoders)

```python
import pyarrow as pa
import narwhals as nw

arr = pa.array(["dog", "cat", "cat", "bird", "dog"]).dictionary_encode()
table = pa.table({"col": arr})

df = nw.from_native(table)
df.sort("col")
```

### Expected results

No errors / df gets sorted.

### Actual results

Fails with:

```
  File "narwhals/dataframe.py", line 1859, in sort
    return super().sort(by, *more_by, descending=descending, nulls_last=nulls_last)
  File "narwhals/dataframe.py", line 278, in sort
    self._compliant_frame.sort(*by, descending=descending, nulls_last=nulls_last)
  File "narwhals/_arrow/dataframe.py", line 500, in sort
    self.native.sort_by(keyed),
  File "pyarrow/table.pxi", line 2153, in pyarrow.lib._Tabular.sort_by
  File "pyarrow/compute.py", line 271, in wrapper
    return func.call(args, options, memory_pool)
  ...
pyarrow.lib.ArrowNotImplementedError: dictionary<values=string, indices=int32, ordered=0>
```

```shell
System:
    python: 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0]
executable: /home/arthur/scikit-learn/sklearn-env/bin/python
   machine: Linux-6.17.0-1030-oem-x86_64-with-glibc2.39

Python dependencies:
     narwhals: 2.24.0
        numpy: 2.5.1
       pandas: 3.0.3
        modin: 
         cudf: 
      pyarrow: 25.0.0
      pyspark: 
       polars: 1.42.1
         dask: 
       duckdb: 
         ibis: 
     sqlframe:
```

aleks-drozy:
## Summary

`ArrowDataFrame.sort()` calls `pa.Table.sort_by()` / `pc.sort_indices()` directly on the native table. PyArrow's sort compute kernel has no implementation for `dictionary<...>`-typed columns, so sorting a narwhals `DataFrame` wrapping a pyarrow `Table` on a dictionary-encoded ("categorical-like") column raises `ArrowNotImplementedError` instead of sorting.

Fixes #3841

## Fix

In `ArrowDataFrame.sort()`, detect dictionary-encoded sort keys and decode them (`pc.cast` to the dictionary's value type) before computing the sort order, then reorder the *original* (still dictionary-encoded) table with `.take()`. This sorts lexicographically on the decoded value, matching pandas/polars categorical sort semantics, as confirmed in the issue discussion, while keeping the output's dictionary encoding intact. Both the `pyarrow<25` and `>=25` code paths (which differ in how `null_placement` is passed) are handled.

## Test plan

- [x] Added `test_sort_pyarrow_dictionary` in `tests/frame/sort_test.py`, reproducing the exact repro from the issue plus a secondary non-key column, descending order, and dtype-preservation checks.
- [x] Confirmed the new test **fails** on `main` with the reported `ArrowNotImplementedError` (verified by temporarily stashing the fix).
- [x] Confirmed the new test **passes** with the fix applied.
- [x] Ran `tests/frame/sort_test.py`, `tests/frame/top_k_test.py`, `tests/series_only/sort_test.py` — all passing, no regressions.
- [x] Ran the full pyarrow-backend test slice (`pytest tests/ -k pyarrow`) — 3709 passed, 221 skipped, 255 xfailed, no failures.
- [x] Manually verified correct behavior (order, descending, null placement, multi-key sort mixing dictionary/non-dictionary columns, dtype preservation) against both `pyarrow==24.0.0` and `pyarrow==25.0.1`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/narwhals/_arrow/dataframe.py
+++ b/src/narwhals/_arrow/dataframe.py
@@ -10,6 +10,7 @@ from narwhals._arrow.series import ArrowSeries
 from narwhals._arrow.utils import (
     arange,
     concat_tables,
+    is_dictionary,
     narwhals_to_native_dtype,
     native_to_narwhals_dtype,
     repeat,
@@ -487,17 +488,42 @@ class ArrowDataFrame(
             ]
 
         null_placement = "at_end" if nulls_last else "at_start"
+        native = self.native
+
+        dict_keys = [key for key in by if is_dictionary(native.schema.field(key).type)]
+        if dict_keys:
+            # `sort_by`/`sort_indices` have no kernel for dictionary-encoded columns
+            # (https://github.com/apache/arrow/issues/29887). Sort on the decoded
+            # values instead - lexicographically, matching `pandas`/`polars`
+            # categorical sort semantics - then reorder the original table so the
+            # output keeps its dictionary encoding.
+            sort_source = native
+            for key in dict_keys:
+                value_type = sort_source.schema.field(key).type.value_type
+                sort_source = sort_source.set_column(
+                    sort_source.schema.get_field_index(key),
+                    key,
+                    pc.cast(sort_source[key], value_type),
+                )
+            if self._backend_version < (25,):  # pragma: no cover
+                indices = pc.sort_indices(
+                    sort_source, sort_keys=sorting, null_placement=null_placement
+                )
+            else:
+                keyed = [(key, order, null_placement) for key, order in sorting]
+                indices = pc.sort_indices(sort_source, sort_keys=keyed)  # type: ignore[arg-type]
+            return self._with_native(native.take(indices), validate_column_names=False)
 
         if self._backend_version < (25,):  # pragma: no cover
             return self._with_native(
-                self.native.sort_by(sorting, null_placement=null_placement),
+                native.sort_by(sorting, null_placement=null_placement),
                 validate_column_names=False,
             )
         # `null_placement` in `SortOptions` is deprecated since 25.0.0;
         # it must be specified per sort key instead.
         keyed = [(key, order, null_placement) for key, order in sorting]
         return self._with_native(
-            self.native.sort_by(keyed),  # type: ignore[arg-type]
+            native.sort_by(keyed),  # type: ignore[arg-type]
             validate_column_names=False,
         )
 
--- a/tests/frame/sort_test.py
+++ b/tests/frame/sort_test.py
@@ -31,3 +31,20 @@ def test_sort_nulls(
     df = nw.from_native(constructor(data))
     result = df.sort("b", descending=True, nulls_last=nulls_last)
     assert_equal_data(result, expected)
+
+
+def test_sort_pyarrow_dictionary() -> None:
+    # https://github.com/narwhals-dev/narwhals/issues/3841
+    pytest.importorskip("pyarrow")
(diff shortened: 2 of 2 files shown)


FBruzzesi (commented):
Thanks for the contribution @aleks-drozy 
The fix is correct for what it covers: I checked nulls, descending, multi-key, chunked columns with per-chunk dictionaries, non-string dictionaries and empty tables, and all behave. Two things make me want a different shape before merging, though.

1. `top_k` crashes on the same input as `pc.select_k_unstable` on a dictionary column does not raise, it segfaults (tested on pyarrow 25):
    ```py
	import pyarrow as pa
	import pyarrow.compute as pc

	t = pa.table({"c": pa.array(["dog", "cat", "bird"]).dictionary_encode()})
	pc.select_k_unstable(t, 2, [("c", "descending")])
	```
2. `Series.sort()`, `Series.rank()` and `DataFrame.with_row_index(order_by=...)` all raise the same error of `DataFrame.sort` - I would rather fix it once everywhere.
3. Decoding every row is the expensive way to get the order. Since only the order matters, ranking the (small) dictionary once and remapping the indices turns the sort key into uint64: on 5M rows and 30 categories that is 664 ms → 41 ms, with the non-dictionary paths unchanged.

---

I will open a PR with an alternative implementation covering all of the above on happy to add you as co-author

aleks-drozy:
Thanks for the thorough look, @FBruzzesi — appreciate you checking nulls/descending/multi-key/chunked/non-string-dict/empty-table cases. Makes sense to fold this into a single fix covering `top_k`, `Series.sort`, `Series.rank`, and `with_row_index(order_by=...)` together rather than patching each call site separately, and the rank-and-remap approach is clearly the better perf story. Happy to be added as co-author — go ahead and open your PR, I'll close this one out once it's up.

FBruzzesi:
Closed by #3862 - @aleks-drozy Marco merged without adding you as co-author 🫥 I am sorry for the inconvenience. We'll make sure this does not happen again

MarcoGorelli:
oops missed this, thought it might have been added already, sorry

note that you add a co-author by editing the commit message when opening a pr https://docs.github.com/en/pull-requests/how-tos/commit-changes/creating-a-commit-with-multiple-authors

aleks-drozy:
No worries at all, thanks both for following up. Appreciate the pointer on the co-author trailer, good to know for next time. All good here, nothing further needed.

PR #3859, opened 2026-08-18
fix: `is_in` was not preserving nulls where possible

### LINKED ISSUE #3851: [Bug]: is_in emits False for null input on pyarrow but null on polars, changing filter results

MarcoGorelli:
closes #3851

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## AI assistance

- [ ] AI tools were used.

## Checklist

- [ ] Tests added
- [ ] Documented the changes

    ```bash
    ```

### DIFF
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -570,7 +570,9 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
             value_set: ArrayOrChunkedArray = other
         else:
             value_set = pa.array(other)
-        return self._with_native(pc.is_in(self.native, value_set=value_set))
+        native = self.native
+        is_in = pc.is_in(native, value_set=value_set)
+        return self._with_native(pc.if_else(native.is_null(), None, is_in))
 
     def arg_true(self) -> Self:
         import numpy as np  # ignore-banned-import
--- a/tests/expr_and_series/is_in_test.py
+++ b/tests/expr_and_series/is_in_test.py
@@ -1,13 +1,27 @@
 from __future__ import annotations
 
 import re
+from typing import Any
 
 import pytest
 
 import narwhals as nw
+from tests.conftest import (
+    dask_lazy_p1_constructor,
+    dask_lazy_p2_constructor,
+    modin_constructor,
+    pandas_constructor,
+)
 from tests.utils import Constructor, ConstructorEager, assert_equal_data
 
 data = {"a": [1, 4, 2, 5]}
+
+NON_NULLABLE_CONSTRUCTORS = [
+    pandas_constructor,
+    dask_lazy_p1_constructor,
+    dask_lazy_p2_constructor,
+    modin_constructor,
+]
 
 
 def test_expr_is_in(constructor: Constructor) -> None:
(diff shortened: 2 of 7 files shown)


tests/expr_and_series/is_in_test.py
```diff
@@ -50,3 +64,67 @@ def test_filter_is_in_with_series(constructor_eager: ConstructorEager) -> None:
...
+    # https://github.com/narwhals-dev/narwhals/issues/3851
+    data_with_null = {"a": [1, 2, None, 4]}
+    df = nw.from_native(constructor(data_with_null))
+    result = df.filter(~nw.col("a").is_in([1, 2]))
+
+    expected: dict[str, list[Any]]
+    if any(constructor is c for c in NON_NULLABLE_CONSTRUCTORS):
+        expected = {"a": [None, 4]}
+    else:
+        expected = {"a": [4]}
```
MarcoGorelli, line 96: it's unfortunate that the results don't totally match here, but as per https://narwhals-dev.github.io/narwhals/concepts/boolean/, i don't think there's much we can do about it

src/narwhals/_arrow/series.py
```diff
@@ -570,7 +570,9 @@ def is_in(self, other: Any) -> Self:
             value_set: ArrayOrChunkedArray = other
         else:
             value_set = pa.array(other)
-        return self._with_native(pc.is_in(self.native, value_set=value_set))
+        native = self.native
+        is_in = pc.is_in(native, value_set=value_set)
+        return self._with_native(pc.if_else(native.is_null(), None, is_in))
```
FBruzzesi, line 575: You can most likely use the `skip_nulls` flag from [`pc.is_in`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.is_in.html)

tests/expr_and_series/is_in_test.py
FBruzzesi: I think it's worth adding a note somewhere in the docstrings - the issue was about this exact behavior, and it's not fully fixed.

FBruzzesi (approved):
Thanks @MarcoGorelli - I left a tiny suggestion.

I think we should consider following up with adding the `nulls_equal` flag in `is_in` (as in polars [is_in](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_in.html))

src/narwhals/_arrow/series.py
MarcoGorelli: i think that does something different, e.g.
```py
In [12]: pc.is_in(pa.array([1, 2, None]), pa.array([1, 2, 3]), skip_nulls=True)
Out[12]:
<pyarrow.lib.BooleanArray object at 0x74aa51c6b3a0>
[
  true,
  true,
  false
]
```
whereas what we want is
```
[
  true,
  true,
  null
]
```

tests/expr_and_series/is_in_test.py
MarcoGorelli: good call thanks!

MarcoGorelli:
thanks for your review!

yup, valid follow-up!

FBruzzesi (approved):
Thanks for addressing @MarcoGorelli

PR #3865, opened 2026-08-19
fix: null propagation for str boolean methods

### LINKED ISSUE #3850: [Bug]: string predicates collapse null to False on the pandas backend, so filter(~...) keeps null rows
## Describe the bug

A string predicate (`str.starts_with`, `str.contains`, ...) applied to a `String` column containing nulls produces a mask whose null is **collapsed to `False`** on the pandas backend, while polars and pyarrow keep it null. Negating that mask therefore turns the null row into `True`, and `filter` silently keeps a row the other backends drop.

The input column is `String` on every backend below — this is not a case of handing narwhals an odd dtype.

```python
import pandas as pd
import polars as pl
import pyarrow as pa
import narwhals as nw

data = {"s": ["x", "y", None, "z"]}

for name, native in {
    "polars": pl.DataFrame(data),
    "pyarrow": pa.table(data),
    "pandas": pd.DataFrame(data),
    "pandas[object]": pd.DataFrame(data).astype(object),
    "pandas[pyarrow]": pd.DataFrame(data).astype("string[pyarrow]"),
}.items():
    print(name, df.filter(~nw.col("s").str.starts_with("x"))["s"].to_list())
```

## Actual results

```
filter(~col.str.starts_with('x'))
    polars           String   -> ['y', 'z']
    pyarrow          String   -> ['y', 'z']
    pandas           String   -> ['y', nan, 'z']      <- keeps the null row
    pandas[object]   String   -> TypeError: bad operand type for unary ~: 'float'
    pandas[pyarrow]  String   -> ['y', 'z']
```

Same shape for `str.contains`. The mask itself is where it goes wrong:

```
the produced mask:
    polars           dtype=Boolean  [True, False, None,  False]
    pyarrow          dtype=Boolean  [True, False, None,  False]
    pandas           dtype=Boolean  [True, False, False, False]   <- null became False
    pandas[object]   dtype=Object   [True, False, nan,   False]   <- and not even Boolean
    pandas[pyarrow]  dtype=Boolean  [True, False, <NA>,  False]
```

So there are two distinct failures from one root:

1. **pandas (numpy-backed, the default in pandas 3)** — the null in the mask becomes `False`. `~False` is `True`, so the null row survives `filter`. **Silent wrong result**, no warning.
2. **pandas with `object` dtype** — the mask comes back as `Object` rather than `Boolean` and `~` raises `TypeError: bad operand type for unary ~: 'float'`. The plain-truth version, `filter(col.str.starts_with("x"))`, instead raises `ValueError: Cannot use ignore_nulls=False in all_horizontal for non-nullable ...`.

`pandas[pyarrow]` is correct throughout, which is consistent with this being about the numpy/object mask not being able to hold a null.

## Expected results

`['y', 'z']` on every backend, and a `Boolean` mask carrying a null in the third position — matching polars, pyarrow and `pandas[pyarrow]`.

Null is not `False`: a row whose value is unknown is not known to not start with `"x"`, so it should be dropped by `filter` under either polarity. That is what the other three backends do.

## Why this one bites

The pandas path is the default one, and the failure is silent. A dedup/exclusion filter like `~col.str.starts_with(prefix)` quietly returns extra rows on pandas and the right rows on polars — which is precisely the class of divergence narwhals exists to remove. The `object` variants at least fail loudly.

## Notes

- Reproduced on `main` (2.24.0 install from a clean `pip install -e .`).
- I have not looked at whether the same collapse affects the other `str.*` predicates beyond `starts_with`/`ends_with`/`contains`, or the non-string predicates (`is_in`, comparisons) on a numpy-backed nullable column — worth checking as part of any fix, since the root looks like "boolean result on a numpy-backed pandas column cannot represent null" rather than anything string-specific.
- Happy to open a PR if you can point me at the preferred shape — casting the mask to `boolean` (nullable) on the pandas-like backend is the obvious move, but that changes the output dtype users see, so it seems like your call rather than mine.

## narwhals.show_versions()

```
```

MarcoGorelli:
closes #3850 

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## AI assistance

- [ ] AI tools were used.

## Checklist

- [ ] Tests added
- [ ] Documented the changes

    ```bash
    ```

### DIFF
--- a/src/narwhals/_pandas_like/series_str.py
+++ b/src/narwhals/_pandas_like/series_str.py
@@ -30,6 +30,8 @@ class PandasLikeSeriesStringNamespace(
                 result = result.astype(bool, copy=False)
             elif not all_valid:
                 result = result.where(not_na, False).astype(bool)
+            else:  # pragma: no cover
+                pass
         return self.with_native(result)
 
     def len_chars(self) -> PandasLikeSeries:
--- a/src/narwhals/_pandas_like/series_str.py
+++ b/src/narwhals/_pandas_like/series_str.py
@@ -23,10 +23,12 @@ class PandasLikeSeriesStringNamespace(
             result.dtype == object
             and get_dtype_backend(self.native.dtype, self.implementation) is None
         ):
-            if self.implementation._backend_version() < (3,):
+            notna = result.notna()
+            all_valid = notna.all()
+            if not all_valid and self.implementation._backend_version() < (3,):
                 result.where(result.notna(), False, inplace=True)
                 result = result.astype(bool, copy=False)
-            else:
+            elif not all_valid:
                 result = result.where(result.notna(), False).astype(bool)
         return self.with_native(result)
 
--- a/src/narwhals/_pandas_like/series_str.py
+++ b/src/narwhals/_pandas_like/series_str.py
@@ -23,7 +23,11 @@ class PandasLikeSeriesStringNamespace(
             result.dtype == object
             and get_dtype_backend(self.native.dtype, self.implementation) is None
         ):
-            result = result.where(result.notna(), False).astype(bool)
+            if self.implementation._backend_version() < (3,):
+                result.where(result.notna(), False, inplace=True)
+                result = result.astype(bool, copy=False)
+            else:
+                result = result.where(result.notna(), False).astype(bool)
         return self.with_native(result)
 
     def len_chars(self) -> PandasLikeSeries:
--- a/src/narwhals/_pandas_like/series_str.py
+++ b/src/narwhals/_pandas_like/series_str.py
@@ -23,13 +23,13 @@ class PandasLikeSeriesStringNamespace(
             result.dtype == object
             and get_dtype_backend(self.native.dtype, self.implementation) is None
         ):
-            notna = result.notna()
-            all_valid = notna.all()
+            not_na = result.notna()
+            all_valid = not_na.all()
             if not all_valid and self.implementation._backend_version() < (3,):
-                result.where(result.notna(), False, inplace=True)
+                result.where(not_na, False, inplace=True)
                 result = result.astype(bool, copy=False)
             elif not all_valid:
-                result = result.where(result.notna(), False).astype(bool)
+                result = result.where(not_na, False).astype(bool)
         return self.with_native(result)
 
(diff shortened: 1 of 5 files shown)


src/narwhals/_pandas_like/series_str.py
```diff
@@ -16,6 +17,19 @@
 class PandasLikeSeriesStringNamespace(
     PandasLikeSeriesNamespace, StringNamespace["PandasLikeSeries"]
 ):
+    def _with_native_bool(self, result: Any) -> PandasLikeSeries:
+        # See https://narwhals-dev.github.io/narwhals/concepts/boolean/.
+        if (
+            result.dtype == object
+            and get_dtype_backend(self.native.dtype, self.implementation) is None
+        ):
```
FBruzzesi, line 25: I wonder if there is a _fast_ way to check if any element is None/NaN/NaT in the condition and only in such case run `isna` + `where` (I am afraid of adding some overhead here)

MarcoGorelli: yeah i guess we could bring part of it out

src/narwhals/_pandas_like/series_str.py
```diff
@@ -16,6 +17,21 @@
...
+            notna = result.notna()
+            all_valid = notna.all()
+            if not all_valid and self.implementation._backend_version() < (3,):
+                result.where(result.notna(), False, inplace=True)
+                result = result.astype(bool, copy=False)
+            elif not all_valid:
+                result = result.where(result.notna(), False).astype(bool)
```
FBruzzesi:
```suggestion
            notna = result.notna()
            all_valid = notna.all()
            if not all_valid and self.implementation._backend_version() < (3,):
                result.where(notna, False, inplace=True)
                result = result.astype(bool, copy=False)
            elif not all_valid:
                result = result.where(notna, False).astype(bool)
```

FBruzzesi (approved):
Thanks for the fix 🚀

MarcoGorelli:
thanks for your review!

PR #3866, opened 2026-08-19
fix: follow left-hand-index alignment rule for pandas-like `concat`

### LINKED ISSUE #3864: `nw.concat` and `maybe_get/set_index`
Hi, this is sort of a collection of issues/questions but I'm reporting them together to share to full picture.

I'm currently trying to figure out how to best handle pandas indices in the context of `nw.concat` for certain cases. Consider the situation where I have some function `process` that transforms a given dataframe into another but *not* using index-preserving dataframe operations (the specifics don't matter, but in my case this happens because I'm transitioning to `torch` for the computation).

The problem I run into is that the resulting pandas execution will differ from other backends if I feed the resulting dataframe together with the original one into `nw.concat`:

```python
import narwhals as nw
import pandas as pd


def process(df: nw.DataFrame) -> nw.DataFrame:
    return nw.from_dict(
        {"B": list(range(len(df)))}, backend=nw.get_native_namespace(df)
    )


df1 = pd.DataFrame({"A": [1, 2, 3]}, index=[10, 20, 30])
df1_nw = nw.from_native(df1)
df2_nw = process(df1_nw)

print("\n-- concat without index alignment --")
print(nw.to_native(nw.concat([df1_nw, df2_nw], how="horizontal")))
```

That is, for polars I'll get the "expected" compact 3-row dataframe while for pandas I get:
```
-- concat without index alignment --
      A    B
10  1.0  NaN
20  2.0  NaN
30  3.0  NaN
0   NaN  0.0
1   NaN  1.0
2   NaN  2.0
```

### Questions
1. Do you see a better way how to handle this situation, i.e. how would you approach it?
2. One easy way to avoid losing the index would be if there existed some function that would let me "swap out" the dataframe content of the original one, instead of creating a new one. In particular, that would allow to both a) keep the index and b) keep the backend information. But I guess that functionality does not exist? The closest I could find is `replace_strict`, but that only works for `Series`.
3. The other alternative I see is manually handle the index by calling `maybe_get_index`/`maybe_set_index`, like outlined in the method below. However, when I tried this, I was very surprised that the two functions are not symmetric in terms of types! That is, I would have expected that `maybe_set_index` also accepts an interable like `Series.from_iterable` does and I can easily nest them like `nw.maybe_set_index(..., index=nw.maybe_get_index(...))`, but that does not seem to be the case.

Any thoughts here?

```python
def process_preserving_index(df: nw.DataFrame) -> nw.DataFrame:
    idx = nw.maybe_get_index(df)
    result = process(df)
    if idx is not None:
        idx_series = nw.from_native(pd.Series(idx), series_only=True)
        result = nw.maybe_set_index(result, index=idx_series)
    return result
```

MarcoGorelli:
# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3864

### DIFF
--- a/src/narwhals/_pandas_like/namespace.py
+++ b/src/narwhals/_pandas_like/namespace.py
@@ -340,12 +340,10 @@ class PandasLikeNamespace(
             [df.set_axis(arange(len(df)), axis=VERTICAL) for df in dfs],
         )
         concatenated = self._concat_by_index(reset_dfs)
-        ret = set_index(
-            cast("pd.DataFrame", concatenated),
-            target_index,
-            implementation=self._implementation,
+        # Value of type variable "NativeNDFrameT" of "set_index" cannot be "NativeDataFrameT"
+        return set_index(  # type: ignore[type-var]
+            concatenated, target_index, implementation=self._implementation
         )
-        return cast("NativeDataFrameT", ret)
 
     def _concat_vertical(self, dfs: Sequence[NativeDataFrameT], /) -> NativeDataFrameT:
         cols_0 = dfs[0].columns
--- a/src/narwhals/_pandas_like/namespace.py
+++ b/src/narwhals/_pandas_like/namespace.py
@@ -505,7 +505,12 @@ class PandasLikeNamespace(
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             a_series = df._evaluate_single_output_expr(a)
             b_series = df._evaluate_single_output_expr(b)
-            _df = self._concat_by_index([a_series.native, b_series.native])
+            _df = self._concat_by_index(
+                cast(
+                    "list[NativeSeriesT]",
+                    [df._extract_comparand(a_series), df._extract_comparand(b_series)],
+                )
+            )
             corr = _df.corr(method=method).iloc[0, [1]]  # type: ignore[union-attr]
             return [
                 PandasLikeSeries(
@@ -525,7 +530,16 @@ class PandasLikeNamespace(
             a_series = df._evaluate_single_output_expr(a)
             b_series = df._evaluate_single_output_expr(b)
             _df = cast(
-                "pd.DataFrame", self._concat_by_index([a_series.native, b_series.native])
+                "pd.DataFrame",
+                self._concat_by_index(
+                    cast(
+                        "list[NativeSeriesT]",
+                        [
+                            df._extract_comparand(a_series),
+                            df._extract_comparand(b_series),
+                        ],
+                    )
+                ),
             )
             n = _df.count(axis=1).eq(2).sum()
             if ddof == 0 and n == 1:
--- a/src/narwhals/_pandas_like/namespace.py
+++ b/src/narwhals/_pandas_like/namespace.py
@@ -505,12 +505,11 @@ class PandasLikeNamespace(
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             a_series = df._evaluate_single_output_expr(a)
             b_series = df._evaluate_single_output_expr(b)
-            _df = self._concat_by_index(
-                cast(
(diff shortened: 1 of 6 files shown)


src/narwhals/_pandas_like/namespace.py
```diff
@@ -319,6 +331,23 @@ def _concat_horizontal(
...
+    def _concat_horizontal(self, dfs: Sequence[NativeDataFrameT], /) -> NativeDataFrameT:
+        # Follow the left-hand-rule documented in `docs/concepts/pandas_index.md`.
+        target_index = dfs[0].index
+        arange = import_array_module(self._implementation).arange
+
+        def reset(df: NativeDataFrameT) -> NativeDataFrameT:
+            # Value of type variable "NativeNDFrameT" of "set_index" cannot be "NativeDataFrameT"
+            return set_index(  # type: ignore[type-var]
+                df, arange(len(df)), implementation=self._implementation
+            )
```
FBruzzesi, line 344: I have a few questions here:
1. Why can't we directly use `target_index` and avoid _one more_ `set_index` in the return statement?
2. Given we can't go with 1., can `arange(len(df))` be computed once outside the helper function?
3. Similarly, can we assign `self._implementation` to a variable once and re-use it?
4. Why do we need the helper function to begin with?

FBruzzesi (commented):
Thanks @MarcoGorelli - I left some questions 🙈

src/narwhals/_pandas_like/namespace.py
MarcoGorelli: thanks for your review!

1. because they might not all be the same length 😭 

https://github.com/narwhals-dev/narwhals/blob/e585f929ff0f60492293efd62e50490f5b3b50a8/tests/frame/concat_test.py#L42-L64

2. we're doing `reset(df) for df in dfs` so i don't think so, they're not necessarily the same length

3. good point, done!

4. i just thought it looked cleaner, this commit shows the diff https://github.com/narwhals-dev/narwhals/pull/3866/changes/fcb435798c4cc1a7f3dd477a07d8d562f306f2ca

FBruzzesi: ahhh I see! I also double checked in polars: from the next release (v2?) that behavior is deprecated, and they suggest to use `how="horizontal_extend"`.

```py
df1 = pl.DataFrame({"a": [1,2,3]})
df2 = pl.DataFrame({"b": [100,200]})

pl.concat([df1, df2], how="horizontal")
<ipython-input-11-11860e3b23ce>:1: DeprecationWarning: the default behavior of `how='horizontal'` for `concat` is deprecated and will require equal heights in the next breaking release. Use `how='horizontal_extend'` to keep the current behavior.
(Deprecated in version 1.42.1)
  pl.concat([df1, df2], how="horizontal")

shape: (3, 2)
┌─────┬──────┐
│ a   ┆ b    │
│ --- ┆ ---  │
│ i64 ┆ i64  │
╞═════╪══════╡
│ 1   ┆ 100  │
│ 2   ┆ 200  │
│ 3   ┆ null │
└─────┴──────┘
```

It's just food for thoughts, but I wonder if it's worth to add something that we should deprecate soon. As you know, the correctness of `concat` has been on my radar for a while now

MarcoGorelli: thanks! yeah it's definitely time to revisit it (but that's on me - will get back to it next!)

FBruzzesi (approved):
Thanks @MarcoGorelli - I will try to open an issue this week to track the work on concat correctness. In the meanwhile, I am happy to ship this

MarcoGorelli:
thanks for your review!

PR #3869, opened 2026-08-21
docs: shorter AGENTS.md

EdAbati:
# Description

The best practice is usually to keep the AGENT.md small and add links and pointers to the full docs. The idea is that the agent will look for more details/docs only when required.

Here I propose to:
- remove the repo tree structure and only summarise the content in `## Code structure and design` (agents tend to list folders and content out of the box anyway)
- shrink the "harness" section: Makefile and contributing.md have already a lot of details, if anything is missing we should add it in the other files (so it is useful for humans too :) )
- remove "code style": prek already gives the feedback the agent need to fix formatting/typing/linting etc. I have kept the note about "inline comments" (which is only relevant for AI agents)
- In the PR paragraph I think we should start with the rule that the AI cannot open PRs. I added the link to the contributing guide which already lists the naming convention

useful sources:
- "Evaluating AGENTS.md: Are Repository-Level Context Files Helpful for Coding Agents?" https://arxiv.org/html/2602.11988v1
- https://www.philschmid.de/writing-good-agents

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## AI assistance

- [x] AI tools were used.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -164,18 +164,17 @@ Default to none. Only add one when the *why* is non-obvious
 already says, and never to narrate a diff. Comments that merely repeat the following
 line are a common tell of unreviewed AI-assisted output and are removed on sight in review.
 
-## Pull requests
-
-Title must start with a [conventional commit](https://www.conventionalcommits.org/) type: `build`,
-`chore`, `ci`, `depr`, `docs`, `feat`, `fix`, `perf`, `refactor`, `release`, `test` (append `!` for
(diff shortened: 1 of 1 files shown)


FBruzzesi (approved):
Thanks for taking care of this @EdAbati  🚀

PR #3890, opened 2026-08-31
feat: add `is_not_null`, `is_not_nan` and `is_infinite` for `Expr` and `Series`

### LINKED ISSUE #3028: Support remaining `Expr.is_<...>` mask methods
## Description
We have *almost all* of these supported, and completing that group seems pretty realistic to me.

They *might* all be possible with a default implementation in `CompliantColumn` (#2967) that just `__invert__`s another method?

`is_not_null` is the only one I'd push for being *directly* implemented for each backend for performance (see [discord comment](https://discord.com/channels/1235257048170762310/1272835922924273694/1397243168273076375))

## Overview

| Method                                                                                                                     | Supported | Related                                                                                                       |
| -------------------------------------------------------------------------------------------------------------------------- | --------- | ------------------------------------------------------------------------------------------------------------- |
| [`is_between`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_between.html)               | ✅         | (#609), (#1672)                                                                                               |
| [`is_close`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_close.html)                   | ✅         | (#2962)                                                                                                       |
| [`is_duplicated`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_duplicated.html)         | ✅         | (#187)                                                                                                        |
| [`is_finite`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_finite.html)                 | ✅         | (#1341), (#1825)                                                                                              |
| [`is_first_distinct`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_first_distinct.html) | ✅         | (#187)                                                                                                        |
| [`is_in`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_in.html)                         | ✅         |                                                                                                               |
| [`is_infinite`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_infinite.html)             | ❌         | (https://github.com/narwhals-dev/narwhals/pull/2962#discussion_r2262187495)                                   |
| [`is_last_distinct`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_last_distinct.html)   | ✅         | (#187)                                                                                                        |
| [`is_nan`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_nan.html)                       | ✅         | (#1625)                                                                                                       |
| [`is_not_nan`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_not_nan.html)               | ❌         |                                                                                                               |
| [`is_not_null`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_not_null.html)             | ❌         | ([discord comment](https://discord.com/channels/1235257048170762310/1272835922924273694/1397243168273076375)) |
| [`is_null`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_null.html)                     | ✅         |                                                                                                               |
| [`is_unique`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_unique.html)                 | ✅         | (#187)                                                                                                        |

sneha4175:
# Description

This PR completes the family of `is_<...>` mask methods tracked in #3028 by adding
`is_not_null`, `is_not_nan` and `is_infinite` to both `Expr` and `Series`, across all
backends. Polars already exposes all three, so the goal here is to bring the other
backends up to parity and match the Polars API.

### Design

- **`is_not_null` / `is_not_nan`** are added as inversion defaults on `CompliantColumn`
  (`return ~self.is_null()` / `return ~self.is_nan()`), following the existing precedent of
  `is_duplicated = ~self.is_unique()`. Because every compliant expression and eager series
  inherits `CompliantColumn`, this single default covers the lazy backends for free.
  As suggested in the issue, `is_not_null` additionally gets a native fast path on the
  eager backends (`Series.notna()` for pandas-like, `pc.is_valid` for PyArrow), wired
  through `EagerExpr` via `_reuse_series`.

- **`is_infinite`** is implemented per backend rather than as an inversion, because it is
  **not** simply the negation of `is_finite`: `NaN` is neither finite nor infinite, so
  `~is_finite()` would wrongly report `NaN` as infinite. Native primitives are used where
  available (`pc.is_inf`, `numpy.isinf`, DuckDB `isinf`, `dask.array.isinf`, Ibis
  `isinf`, and an explicit `±inf` comparison for Spark-like), each preserving nulls the
  same way the neighbouring `is_finite` implementation does.

- **Polars** routes `is_not_nan`/`is_infinite` through small version-guarded wrappers
  (like the existing `is_nan`/`is_finite`) so that null values are preserved on
  `polars < 1.18`. `is_not_null` is added to the inherited-methods passthrough.

Semantics were validated against Polars as the reference across every installed backend.

- [x] ✨ Feature
- [x] ✅ Test
- [x] 📝 Documentation

## Related issues

- Closes #3028

## AI assistance

- [x] AI tools were used.

Claude (Claude Code) was used to help navigate the backend layout, draft the
implementations/tests, and iterate to green. I have read through the full diff, understand
the cross-backend null/NaN semantics involved, and take responsibility for every line.

## Checklist

- [x] Tests added
- [x] Documented the changes

### Local test run

New tests pass across every backend I have installed locally
(pandas / pandas[nullable] / pandas[pyarrow] / pyarrow / polars[eager] / polars[lazy] /
dask / duckdb / sqlframe / ibis):

```
$ uv run pytest tests/expr_and_series/is_not_null_test.py \
    tests/expr_and_series/is_not_nan_test.py \
    tests/expr_and_series/is_infinite_test.py \
    --constructors="pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],dask,duckdb,sqlframe,ibis"

======================== 97 passed, 3 xfailed in 0.70s =========================
```

The full suite is green as well, and the changed files are at 100% line/branch coverage:

```
$ uv run pytest tests/ --constructors="pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],dask,duckdb,sqlframe,ibis" --cov=narwhals

15785 passed, 315 skipped, 1016 xfailed
TOTAL coverage 99% (no missing lines in any file touched by this PR)
```

`ruff`, `mypy`, `pyright`, `pyrefly`, the doctests, and the `check_api_reference` /
`check_docstrings` utilities all pass locally.

The 3 xfails are the existing "`is_nan` on a non-float column" behaviour, which the
`is_not_nan` inversion inherits (DuckDB/PySpark/Ibis don't raise), mirroring
`is_nan_test.py`.


src/narwhals/_polars/series.py
```diff
@@ -379,6 +380,23 @@ def is_finite(self) -> Self:
...
+            select = pl.when(self.native.is_not_null()).then(native_is_not_nan)
+            return self._with_native(pl.select(select)[self.name])
+        return self._with_native(native_is_not_nan)
+
+    def is_infinite(self) -> Self:
+        native_is_infinite = self.native.is_infinite()
+        if self._backend_version < (1, 18):  # pragma: no cover
+            select = pl.when(self.native.is_not_null()).then(native_is_infinite)
+            return self._with_native(pl.select(select)[self.name])
+        return self._with_native(native_is_infinite)
```
FBruzzesi, line 398: `is_not_nan` wraps polars error while `is_infinite` does not. I assume it's a consequence of how `is_finite` is implemented? I would consider this to be a pre-existing issue for `is_finite` (and worth fixing, either here or in a followup)

src/narwhals/_compliant/column.py
```diff
@@ -102,7 +102,14 @@ def is_between(
     def is_duplicated(self) -> Self:
         return ~self.is_unique()
 
+    def is_not_nan(self) -> Self:
+        return ~self.is_nan()
+
+    def is_not_null(self) -> Self:
+        return ~self.is_null()
```
FBruzzesi, line 109: These are fine, however the error raised for `is_not_nan` on a non-numeric column is suggesting the wrong thing, namely the one from the `is_nan`, suggesting to use `is_null` instead. We should suggest `is_not_null`

src/narwhals/_polars/series.py
```diff
@@ -107,6 +107,7 @@
         "is_first_distinct",
         "is_in",
         "is_last_distinct",
+        "is_not_null",
```
FBruzzesi, line 110: Also requires to add `Method[Self]` declaration at the end of the class

tests/expr_and_series/is_infinite_test.py
```diff
@@ -0,0 +1,88 @@
...
+    assert_equal_data(result, expected)
+
+
+@pytest.mark.filterwarnings("ignore:invalid value encountered in cast")
+def test_is_infinite_series(constructor_eager: ConstructorEager) -> None:
+    if "polars" in str(constructor_eager) or "pyarrow_table" in str(constructor_eager):
+        expected = {"a": [False, True, False, None]}
+    elif (
+        "pandas_constructor" in str(constructor_eager)
+        or "dask" in str(constructor_eager)
```
FBruzzesi, line 52: This is a dead condition as the `constructor_eager` can never be the dask constructor

src/narwhals/_dask/expr.py
```diff
@@ -630,6 +630,11 @@ def is_finite(self) -> Self:
 
         return self._with_callable(da.isfinite)
 
+    def is_infinite(self) -> Self:
+        import dask.array as da
+
+        return self._with_callable(da.isinf)
+
```
FBruzzesi, line 637: Should dask also have a is_not_nan "fast path" as pandas does?

src/narwhals/_spark_like/expr.py
```diff
@@ -306,6 +306,19 @@ def _is_finite(expr: Column) -> Column:
 
         return self._with_elementwise(_is_finite)
 
+    def is_infinite(self) -> Self:
+        def _is_infinite(expr: Column) -> Column:
+            # A value is infinite if it is +inf or -inf, while NaN and finite values
+            # are not. NULLs should be preserved.
+            is_infinite_condition = (expr == self._F.lit(float("inf"))) | (
+                expr == self._F.lit(float("-inf"))
+            )
```
FBruzzesi, line 315: Comparing with `is_finite`: the leading `~self._F.isnan(expr)` works also as a type gate. `isnan` has no overload for non-numeric input, so a string column fails at bind time. `is_infinite` uses only `==`, which Spark resolves by casting the string side to double.

The resulting difference is:

> is_finite    on a string column -> BinderException: No function matches 'isnan(VARCHAR)'
> is_infinite  on a string column -> ConversionException: Could not convert string 'x' to FLOAT

Both error, but the second is a data-dependent conversion failure rather than a schema-level rejection.

I think changing the condition to:

```py
is_infinite_condition = ~self._F.isnan(expr) & (
    (expr == self._F.lit(float("inf"))) | (expr == self._F.lit(float("-inf")))
)
```

is enough as a guard and `isnan` is never true for +/-inf, so this is a no-op on numeric input and makes it match with `is_finite`'s failure mode on everything else.

tests/expr_and_series/is_not_null_test.py
```diff
@@ -0,0 +1,34 @@
...
+    data_na = {"a": [None, 3, 2], "z": [7.0, None, None]}
+    expected = {"a": [False, True, True], "z": [True, False, False]}
+    df = nw.from_native(constructor_eager(data_na), eager_only=True)
+    result = {"a": df["a"].is_not_null(), "z": df["z"].is_not_null()}
+
+    assert_equal_data(result, expected)
+
+
+def test_not_null_is_inverse_of_null(constructor: Constructor) -> None:
+    # `is_not_null` should always be the exact negation of `is_null`.
```
FBruzzesi, line 26: unneeded inline comment

tests/expr_and_series/is_not_null_test.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import annotations
+
+import narwhals as nw
+from tests.utils import Constructor, ConstructorEager, assert_equal_data
+
+
+def test_not_null(constructor: Constructor) -> None:
+    data_na = {"a": [None, 3, 2], "z": [7.0, None, None]}
```
FBruzzesi, line 8: You can define the data outside the functions, it's the same in every test

src/narwhals/_spark_like/expr.py
```diff
@@ -306,6 +306,19 @@ def _is_finite(expr: Column) -> Column:
 
         return self._with_elementwise(_is_finite)
 
+    def is_infinite(self) -> Self:
+        def _is_infinite(expr: Column) -> Column:
+            # A value is infinite if it is +inf or -inf, while NaN and finite values
+            # are not. NULLs should be preserved.
```
FBruzzesi, line 312: Comment is probably not needed. Code is pretty much self explainatory

tests/expr_and_series/is_not_nan_test.py
```diff
@@ -0,0 +1,142 @@
...
+    constructor: Constructor, request: pytest.FixtureRequest
+) -> None:
+    pytest.importorskip("pyarrow")
+
+    if (
+        ("pyspark" in str(constructor))
+        or "duckdb" in str(constructor)
+        or "ibis" in str(constructor)
+    ):
+        request.applymarker(pytest.mark.xfail)
```
FBruzzesi, line 108: Can you add the xfail reason here?

tests/expr_and_series/is_not_nan_test.py
```diff
@@ -0,0 +1,142 @@
...
+    from narwhals.exceptions import InvalidOperationError
+
+    data = {"a": ["x", "y"]}
+    df = nw.from_native(constructor(data))
+
+    exc = (
+        ArrowNotImplementedError
+        if "pyarrow_table" in str(constructor)
+        else InvalidOperationError
+    )
```
FBruzzesi, line 120: I think the arrow exception import should be conditional, otherwise this would fail if we run polars only (without any other dependency).

Something along the following lines:

```py
exc: type[Exception]

if "pyarrow_table" in str(constructor):
    from pyarrow.lib import ArrowNotImplementedError

    exc = ArrowNotImplementedError
else:
    exc = InvalidOperationError
```

src/narwhals/series.py
```diff
@@ -2319,6 +2365,34 @@ def is_finite(self) -> Self:
         """
         return self._with_compliant(self._compliant_series.is_finite())
 
+    def is_infinite(self) -> Self:
+        """Returns a boolean Series indicating which values are infinite.
+
+        Warning:
+            Different backend handle null values differently. `is_infinite` will return
```
FBruzzesi, line 2372:
```suggestion
            Different backends handle null values differently. `is_infinite` will return
```

tests/expr_and_series/is_infinite_test.py
```diff
@@ -0,0 +1,88 @@
...
+    if any(
+        x in str(constructor)
+        for x in ("polars", "pyarrow_table", "duckdb", "pyspark", "ibis")
+    ):
+        expected = {"a": [False, True, False, None]}
+    elif any(
+        x in str(constructor) for x in ("pandas_constructor", "dask", "modin_constructor")
+    ):
+        expected = {"a": [False, True, False, False]}
+    else:  # pandas_nullable_constructor, pandas_pyarrow_constructor, modin_pyarrrow_constructor
```
FBruzzesi, line 37: Doesn't really matter but:

```suggestion
    else:  # pandas_nullable_constructor, pandas_pyarrow_constructor, modin_pyarrow_constructor
```

And same a few lines below

FBruzzesi (commented):
Thanks for the contribution @sneha4175 🙏🏼

I left a few inline comments. On top of those `docs/concepts/null_handling.md` should be updated with a new entry in the table ("is not")

PR #3896, opened 2026-09-03
fix(test): Fix interchange and concat tests

FBruzzesi:

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

---

Related to polars v2.0.0rc: use pyarrow for interchange and enlarge the matching message for polars concat error. The CI will still fail with a couple of errors left (Update 2 errors left out of 10 for the nightly build)

The `__contains__` and `is_in` errors are a bit more complex and I am planning to take care separately. Most importantly, we might need to decide if we want to fully align with polars v2 behavior

### DIFF
--- a/tests/frame/concat_test.py
+++ b/tests/frame/concat_test.py
@@ -102,7 +102,7 @@ def test_concat_vertical(constructor: Constructor) -> None:
 
     with pytest.raises(
         (Exception, TypeError),
-        match=r"unable to vstack|inputs should all have the same schema",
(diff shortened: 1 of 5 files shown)


MarcoGorelli (approved):
thanks @FBruzzesi !

PR #3902, opened 2026-09-04
fix: Deprecate `cat.get_categories` and stop relying on Polars' removed method

### LINKED ISSUE #3895: cat.get_categories() on a polars backend hits a method deprecated in polars 1.44 / removed in polars 2.0
### Summary

`Series.cat.get_categories()` / `Expr.cat.get_categories()` are passed straight through to polars, where the method is deprecated as of polars 1.44 and is documented for removal in polars 2.0. Any narwhals user calling `cat.get_categories()` on a polars-backed object gets a `DeprecationWarning` they cannot act on, and will get an `AttributeError` once polars 2.0 lands.

The polars deprecation message:

> `cat.get_categories()` is deprecated. To get the distinct values present in a Categorical column, use `Expr.unique()`. For the fixed category list of an Enum, use its `dtype.categories`. This method will be removed in Polars 2.0.

The passthrough is here — `PolarsSeriesCatNamespace` has no body, so the attribute resolves onto the native polars namespace via `__getattr__`:

https://github.com/narwhals-dev/narwhals/blob/main/narwhals/_polars/series.py (`class PolarsSeriesCatNamespace(PolarsSeriesNamespace, PolarsCatNamespace[PolarsSeries, pl.Series]): ...`)

### Reproducible example

```python
import warnings

import narwhals as nw
import polars as pl

print(f"narwhals {nw.__version__}, polars {pl.__version__}")

s_cat = pl.Series(["a", "b", "b", "c"], dtype=pl.Categorical)
s_enum = pl.Series(["Panda", "Polar"], dtype=pl.Enum(["Polar", "Panda", "Brown"]))

for name, s in [("Categorical", s_cat), ("Enum", s_enum)]:
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter("always")
        out = nw.from_native(s, series_only=True).cat.get_categories().to_list()
    print(f"\n{name}: {out}")
    for warning in w:
        print(f"  {warning.category.__name__}: {warning.message}")

# Workaround for Enum on the main namespace: read the dtype's category list.
print(
    "\nEnum via dtype.categories (narwhals):",
    list(nw.from_native(s_enum, series_only=True).dtype.categories),
)

# ...but the same workaround is unavailable under narwhals.stable.v1:
try:
    list(nw_v1.from_native(s_enum, series_only=True).dtype.categories)
except AttributeError as e:
    print("Enum via dtype.categories (narwhals.stable.v1):", type(e).__name__, e)
```

Output:

```
narwhals 2.25.0, polars 1.44.1

Categorical: ['a', 'b', 'c']
  DeprecationWarning: `cat.get_categories()` is deprecated. To get the distinct values present in a Categorical column, use `Expr.unique()`. For the fixed category list of an Enum, use its `dtype.categories`. This method will be removed in Polars 2.0.

Enum: ['Polar', 'Panda', 'Brown']
  DeprecationWarning: `cat.get_categories()` is deprecated. To get the distinct values present in a Categorical column, use `Expr.unique()`. For the fixed category list of an Enum, use its `dtype.categories`. This method will be removed in Polars 2.0.

Enum via dtype.categories (narwhals): ['Polar', 'Panda', 'Brown']
Enum via dtype.categories (narwhals.stable.v1): AttributeError 'narwhals.stable.v1._dtypes.Enum' object has no attribute '_cached_categories'
```

### Possible implementation for the polars backend

Give `PolarsSeriesCatNamespace` / the expression equivalent a real `get_categories()` rather than letting it fall through, following what polars itself recommends:

* **Enum** — take the fixed list off the native dtype: `pl.Series(dtype.categories)` (an Enum's `categories` is already a `pl.Series`), so unused categories and declared order are preserved. This matches the current behaviour exactly.
* **Categorical** — polars no longer keeps a category list separate from the values present, so `s.unique(maintain_order=True).drop_nulls().cast(pl.String)` is the equivalent. This also matches current behaviour for the cases we tested.

Gating on the polars version (`get_categories` is only deprecated from 1.44) would keep older polars working unchanged.

### Secondary: `narwhals.stable.v1` blocks the documented workaround

The `AttributeError` in the last line of the reprex is a separate wrinkle. `narwhals/stable/v1/_dtypes.py` defines:

```python
class Enum(NwEnum):
    def __init__(self) -> None:
        super(NwEnum, self).__init__()
```

which deliberately skips `NwEnum.__init__`, so `_cached_categories` / `_delayed_categories` are never assigned and the inherited `categories` property raises `AttributeError` on a private slot. I understand v1 treats `Enum` as category-less by design, but the practical effect is that a stable.v1 user cannot follow polars' recommended workaround at all, and the failure mode is an internal-looking `AttributeError` rather than a clear message. A `NotImplementedError` (or similar) explaining "use `narwhals.Enum` for category access" would at least be diagnosable.

### Context

Found in [py-shiny](https://github.com/posit-dev/py-shiny), where we call `cat.get_categories()` to serialize a data frame column's categories for the client ([`_tbl_data.py`](https://github.com/posit-dev/py-shiny/blob/main/shiny/render/_data_frame_utils/_tbl_data.py)). Ideally, we'd like to notspecial-case polars in our own code.

Happy to open a PR if the approach above sounds right.

### Versions

```
narwhals 2.25.0
polars 1.44.1
python 3.14.1
```

FBruzzesi:
# Description

Deprecates `cat.get_categories` in the main namespace, where it is implemented as `unique().drop_nulls().cast(String)` until removal. `stable.{v1,v2}` keep dispatching to each backend's native implementation.

* For `Expr` I added a new private `_version` attribute to branch conditionally on the version
* For Polars backend I switched to the workaround Polars itself suggests so we never touch the method they deprecated in 1.44 and remove in 2.0. That might have some performance cost.

While doing that, I spotted an (**existing**) issue with selectors, namely: `narwhals.stable.v*.selectors` return main-namespace `Selectors`, which made stable selectors silently behave like main-namespace expressions.

~~So this PR also fixes that (I can upstream it separately if more convenient, and stack this one on top of it)~~ Moved into a separate PR to allow a more focused and manageable review. See #3904

---

Closes #3895.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


src/narwhals/expr_cat.py
```diff
@@ -17,9 +21,24 @@ def __init__(self, expr: ExprT) -> None:
...
+            are actually present, in no guaranteed order.
+
+        Note:
+            In `narwhals.stable.v1` and `narwhals.stable.v2` this stays available, and
+            keeps dispatching to each backend's native implementation.
+
         Examples:
             >>> import pandas as pd
-            >>> import narwhals as nw
+            >>> import narwhals.stable.v1 as nw
```
FBruzzesi, line 41: Otherwise the warning is issued

src/narwhals/series_cat.py
```diff
@@ -12,16 +16,37 @@ def __init__(self, series: SeriesT) -> None:
...
+            the values which are actually present, in order of appearance.
+
+        Note:
+            In `narwhals.stable.v1` and `narwhals.stable.v2` this stays available, and
+            keeps dispatching to each backend's native implementation.
+
         Examples:
             >>> import pandas as pd
-            >>> import narwhals as nw
+            >>> import narwhals.stable.v2 as nw
```
FBruzzesi, line 36: Otherwise the warning is issued

src/narwhals/stable/v1/selectors.py
FBruzzesi, line 1: First implementation was _much more compact_ (see https://github.com/narwhals-dev/narwhals/pull/3902/commits/5a6b822ffacf7d7484246f38325de73e36aadacf). `typing` was failing with a warning. It did feel like a pyrefly shortcoming but didn't went down such rabbithole.

Same applies to `src/narwhals/stable/v2/selectors.py`

MarcoGorelli (approved):
thanks @FBruzzesi !

PR #3904, opened 2026-09-04
fix(typing): Stableify selectors

FBruzzesi:
# Description

Spotted while working on #3902 

`narwhals.stable.v*.selectors` return main-namespace `Selectors`, which made stable selectors silently behave like main-namespace expressions.

---

Comment ported from #3902:

> First implementation was _much more compact_ (see https://github.com/narwhals-dev/narwhals/pull/3902/commits/5a6b822ffacf7d7484246f38325de73e36aadacf). `typing` was failing with a warning. It did feel like a pyrefly shortcoming but didn't went down such rabbithole.
>
> Same applies to `src/narwhals/stable/v2/selectors.py`

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other


MarcoGorelli (approved):
thanks @FBruzzesi !

PR #2889, opened 2025-07-26
fix: `.clip` should not treat 0 and None as same

FBruzzesi:
## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

Spotted while narwhalifying great-tables

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [ ] Documented the changes

PR #3906, opened 2026-09-05
fix(pyarrow): keep nulls through clip instead of replacing them with a bound

winklemad:
# Description

On the pyarrow backend, `clip` (and `clip_lower` / `clip_upper`) replaced a **null value in the data** with a clip bound instead of preserving it:

```python
import narwhals as nw, pyarrow as pa
s = nw.from_native(pa.chunked_array([[1, None, 3]]), series_only=True)
s.clip(0, 2).to_list()
# main:   [1, 2, 3]     <- the null became the upper bound
# polars: [1, None, 2]  <- null preserved
```

The cause is that pyarrow's `min_element_wise` / `max_element_wise` skip nulls by default, so `min_element_wise([null], 2)` returns `2`. Passing `skip_nulls=False` keeps a null as null, matching Polars and pandas.

This is distinct from #2889, which was about `None` **bounds** (unbounded clip); here the null is in the data. The existing `clip` tests use no nulls in the data, so this was uncovered.

- [x] 🐛 Bug Fix

## Related issues

- None

## AI assistance

- [x] AI tools were used.

I used Claude Code to help locate the divergence (a differential comparison of pyarrow vs pandas/polars) and draft the fix and test. I reviewed every line, reproduced the bug and the fix locally, and take full responsibility for the change.

## Checklist

- [x] Fix applied to `clip`, `clip_lower` and `clip_upper` (the whole class) in the pyarrow backend
- [x] Regression test `test_clip_preserves_null` added; fails on `main` for pyarrow, passes after; full `clip_test.py` green
- [x] `ruff check` and `ruff format` clean

### DIFF
--- a/src/narwhals/_sql/expr.py
+++ b/src/narwhals/_sql/expr.py
@@ -576,7 +576,6 @@ class SQLExpr(LazyExpr[SQLLazyFrameT, NativeExprT], Protocol[SQLLazyFrameT, Nati
         def _clip(
             expr: NativeExprT, lower_bound: NativeExprT, upper_bound: NativeExprT
         ) -> NativeExprT:
-            # `least` / `greatest` skip nulls, so guard so a null stays null.
             clipped = self._function(
                 "greatest", self._function("least", expr, upper_bound), lower_bound
             )
--- a/src/narwhals/_sql/expr.py
+++ b/src/narwhals/_sql/expr.py
@@ -576,9 +576,11 @@ class SQLExpr(LazyExpr[SQLLazyFrameT, NativeExprT], Protocol[SQLLazyFrameT, Nati
         def _clip(
             expr: NativeExprT, lower_bound: NativeExprT, upper_bound: NativeExprT
         ) -> NativeExprT:
-            return self._function(
+            # `least` / `greatest` skip nulls, so guard so a null stays null.
+            clipped = self._function(
                 "greatest", self._function("least", expr, upper_bound), lower_bound
             )
+            return self._when(self._function("isnull", expr), self._lit(None), clipped)
 
         return self._with_elementwise(
             _clip,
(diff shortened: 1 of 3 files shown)


tests/expr_and_series/clip_test.py
```diff
@@ -67,3 +67,10 @@ def test_clip_invalid(constructor: Constructor) -> None:
     df = nw.from_native(constructor({"a": [1, 2, 3], "b": [4, 5, 6]}))
     with pytest.raises(MultiOutputExpressionError):
         df.select(nw.col("a").clip(nw.all(), nw.col("a", "b")))
+
+
+def test_clip_preserves_null(constructor_eager: ConstructorEager) -> None:
+    # A null value must stay null through clip, not be replaced by a bound.
+    df = nw.from_native(constructor_eager({"a": [1, None, 3, -4, None]}), eager_only=True)
+    result = {"a": df["a"].clip(lower_bound=0, upper_bound=2)}
+    assert_equal_data(result, {"a": [1, None, 2, 0, None]})
```
FBruzzesi: Can you test against `constructor` so that we can guarantee that lazy backends also behave the same?

FBruzzesi (commented):
Thanks @winklemad for spotting and fixing the issue. Left a comment to extend the test to lazy backends as well. Once we do that, it's ready to merge

winklemad:
Good call — widening the test to `constructor` caught that the SQL backends (duckdb, and any other `least`/`greatest`-based one) had the same bug: those functions skip nulls too, so clip was replacing a null with the bound there as well. Guarded `clip`, `clip_lower` and `clip_upper` in `_sql/expr.py` with a null check, and the test now runs across every backend (dask was already fine). Full clip suite is green.

src/narwhals/_sql/expr.py
```diff
@@ -576,9 +576,11 @@ def clip(self, lower_bound: Self, upper_bound: Self) -> Self:
         def _clip(
             expr: NativeExprT, lower_bound: NativeExprT, upper_bound: NativeExprT
         ) -> NativeExprT:
-            return self._function(
+            # `least` / `greatest` skip nulls, so guard so a null stays null.
+            clipped = self._function(
```
MarcoGorelli:
```suggestion
            clipped = self._function(
```

MarcoGorelli (approved):
thanks @winklemad ! and thanks @FBruzzesi for review

winklemad:
Thanks @FBruzzesi @MarcoGorelli 🙏 — and appreciate the nudge to widen the test to `constructor`; that's what surfaced the same null-dropping in the `least`/`greatest` SQL path, so it ended up a better fix for it.

PR #3908, opened 2026-09-06
fix(typing): Stableify `concat` with the stable frame classes

### LINKED ISSUE #3897: [Bug]: `nw.concat` typing
### Describe the bug

Hi, not sure if this is an actual typing bug or some limitation of the typing system, but I'll leave it to the experts to decide 🙃 

```python
from typing import TypeVar

import narwhals.stable.v2 as nw
import pandas as pd

_FrameT = TypeVar("_FrameT", nw.Series, nw.DataFrame)


def identity(x: _FrameT) -> _FrameT:
    return x


df = nw.from_native(pd.DataFrame({"a": [1, 2, 3]}), eager_only=True)

# nw.from_native -> narwhals.stable.v2.DataFrame  ✓ accepted
identity(df)

# nw.concat -> narwhals.dataframe.DataFrame[Any]  ✗ rejected by type checkers
# even though at runtime it's the same object as nw.DataFrame
identity(nw.concat([df], how="horizontal"))
```

### Expected results

No error thrown.

### Mypy
```
error: Value of type variable "_FrameT" of "identity" cannot be "DataFrame[Any]"  [type-var]
```

### Pyrefly
```
ERROR `DataFrame[Any]` is not assignable to any of constraints `Series[Unknown]`, `DataFrame[Unknown]` of type variable `_FrameT` [bad-specialization]
  --> nar.py:20:9
   |
20 | identity(nw.concat([df], how="horizontal"))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
 INFO 1 error
```

```shell
narwhals==2.25.0
```

heyilingran:
## Summary

`concat` in `narwhals.stable.v1` / `stable.v2` was re-exported from the main namespace, so type checkers saw its result as `narwhals.dataframe.DataFrame[Any]` instead of the stable classes. Passing the result to a function typed with the stable classes (e.g. through a constrained `TypeVar` over `nw.DataFrame` / `nw.Series`) was rejected by mypy and pyrefly, even though the runtime object is a stable instance.

## Root cause

Both stable namespaces imported and re-exported `narwhals.functions.concat` unchanged, leaving its annotations tied to the main namespace.

## Fix

- Define `concat` in both stable namespaces, wrapping the main implementation with `_stableify`.
- Use each namespace's existing `FrameT` in the signature `Iterable[FrameT] -> FrameT`, preserving generic callers without separate eager/lazy overloads.
- Include the concat docstring in both wrappers, keeping v2 consistent with the main namespace.

## Tests

- Consolidate regression checks into each namespace's existing `test_concat`.
- Check inferred return types before `isinstance` narrows them.
- Use the same `Series` / `DataFrame` constrained TypeVar in v1 and v2 for the issue's `identity(concat(...))` reproduction.
- Exercise generic `FrameT` and `Sequence[FrameT]` callers with both eager and lazy frames, checking types and data.

Local validation on `ed446cf`, after merging `main` at `6bb4f7c`:

- All `prek run --all-files` checks pass.
- mypy, pyright, and the default project-scoped pyrefly check report no errors (see the source-level follow-up below).
- `pytest tests/v1_test.py tests/v2_test.py tests/stable_api_test.py tests/namespace_test.py tests/frame/concat_test.py`: **458 passed, 16 skipped, 9 xfailed**.
- `pytest src --doctest-modules`: **484 passed**.
- Dynamic documentation generation and `zensical build --clean --strict` pass.
- Full suite with the Makefile's constructor list, slow tests, coverage, and four workers: **14,650 passed, 4 failed, 478 skipped, 864 xfailed**. The four failures are SQLFrame timezone assertions in this Windows environment. An independent, unmodified `main` checkout at `6bb4f7c` produces the same four failures and test counts. Both runs fall below the 100% coverage gate, with the same 79 uncovered statements and six partially covered branches.

Fixes #3897

## Source-level pyrefly follow-up

Commit `c8c5d2b` adds an explicit `cast` to each stable `concat` return. Both the underlying concatenation and `_stableify` preserve the eager/lazy kind represented by the constrained `FrameT`. Mypy already infers this relationship, so a per-line `redundant-cast` annotation handles that checker difference.

The default `pyrefly check` targets `tests`; it did not report the two `bad-return` errors found by directly checking the stable source files. After this fix, both concat errors are gone. Direct source checking still reports two pre-existing `DataFrame.to_dict` override errors, also present on unmodified `main`. The result is the same when targeting Python 3.12.

Mypy, pyright, default pyrefly, public type completeness, pre-commit checks, the 458 targeted tests, and 484 doctests pass. The full suite retains the same 14,650 passes and four baseline SQLFrame timezone failures described above.

## AI assistance

- [x] AI tools were used: Codex (GPT-6) assisted with implementation and validation.

### DIFF
--- a/src/narwhals/stable/v1/__init__.py
+++ b/src/narwhals/stable/v1/__init__.py
@@ -969,7 +969,8 @@ def concat(items: Iterable[FrameT], *, how: ConcatMethod = "vertical") -> FrameT
     Raises:
         TypeError: The items to concatenate should either all be eager, or all lazy
     """
-    return _stableify(nw_f.concat(items, how=how))
+    # Pyrefly needs this cast; concat and _stableify preserve the eager/lazy kind.
+    return cast("FrameT", _stableify(nw_f.concat(items, how=how)))  # type: ignore[redundant-cast]
 
 
 @deprecate_native_namespace(required=True)
--- a/src/narwhals/stable/v1/__init__.py
+++ b/src/narwhals/stable/v1/__init__.py
@@ -969,7 +969,10 @@ def concat(items: Iterable[FrameT], *, how: ConcatMethod = "vertical") -> FrameT
     Raises:
         TypeError: The items to concatenate should either all be eager, or all lazy
     """
-    # Pyrefly needs this cast; concat and _stableify preserve the eager/lazy kind.
+    # `pyrefly` rejects the union `_stableify` returns while `FrameT` is still
+    # unsolved; `mypy` solves per-constraint and calls the cast redundant, hence
+    # the ignore. CI cannot catch removal of either, because `pyrefly check` only
+    # covers `tests`. Verify with: pyrefly check src/narwhals/stable/v1/__init__.py
     return cast("FrameT", _stableify(nw_f.concat(items, how=how)))  # type: ignore[redundant-cast]
 
 
--- a/src/narwhals/stable/v1/__init__.py
+++ b/src/narwhals/stable/v1/__init__.py
@@ -25,7 +25,7 @@ from narwhals._utils import (
 from narwhals.dataframe import DataFrame as NwDataFrame, LazyFrame as NwLazyFrame
 from narwhals.exceptions import InvalidIntoExprError, NarwhalsUnstableWarning
 from narwhals.expr import Expr as NwExpr
-from narwhals.functions import _new_series_impl, concat, show_versions
+from narwhals.functions import _new_series_impl, show_versions
 from narwhals.schema import Schema as NwSchema
 from narwhals.series import Series as NwSeries
 from narwhals.stable.v1 import dependencies, dtypes, selectors
@@ -108,6 +108,7 @@ if TYPE_CHECKING:
     )
     from narwhals.dataframe import MultiColSelector, MultiIndexSelector
     from narwhals.typing import (
+        ConcatMethod,
         FileSource,
         IntoDType,
         IntoExpr,
@@ -123,6 +124,7 @@ if TYPE_CHECKING:
     T = TypeVar("T", default=Any)
     P = ParamSpec("P")
     R = TypeVar("R")
+    _FrameT = TypeVar("_FrameT", "DataFrame[Any]", "LazyFrame[Any]")
 
 
 # NOTE legit
@@ -946,6 +948,24 @@ class Then(nw_f.Then, Expr):
 
 def when(*predicates: IntoExpr | Iterable[IntoExpr]) -> When:
     return When.from_when(nw_f.when(*predicates))
+
+
+@overload
+def concat(
+    items: Iterable[DataFrame[Any]], *, how: ConcatMethod = "vertical"
+) -> DataFrame[Any]: ...
+
+
+@overload
+def concat(
+    items: Iterable[LazyFrame[Any]], *, how: ConcatMethod = "vertical"
+) -> LazyFrame[Any]: ...
+
+
+def concat(
+    items: Iterable[_FrameT], *, how: ConcatMethod = "vertical"
+) -> DataFrame[Any] | LazyFrame[Any]:
+    return _stableify(nw_f.concat(items, how=how))
 
 
 @deprecate_native_namespace(required=True)
--- a/src/narwhals/stable/v1/__init__.py
+++ b/src/narwhals/stable/v1/__init__.py
@@ -62,6 +62,7 @@ from narwhals.stable.v1.dtypes import (
 )
 from narwhals.stable.v1.typing import (
     DataFrameT,
+    FrameT,
     IntoDataFrameT,
     IntoFrame,
     IntoLazyFrameT,
@@ -124,7 +125,6 @@ if TYPE_CHECKING:
     T = TypeVar("T", default=Any)
     P = ParamSpec("P")
     R = TypeVar("R")
-    _FrameT = TypeVar("_FrameT", "DataFrame[Any]", "LazyFrame[Any]")
 
 
 # NOTE legit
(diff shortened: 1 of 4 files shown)


src/narwhals/stable/v1/__init__.py
```diff
@@ -948,6 +950,24 @@ def when(*predicates: IntoExpr | Iterable[IntoExpr]) -> When:
...
+@overload
+def concat(
+    items: Iterable[DataFrame[Any]], *, how: ConcatMethod = "vertical"
+) -> DataFrame[Any]: ...
+
+
+@overload
+def concat(
+    items: Iterable[LazyFrame[Any]], *, how: ConcatMethod = "vertical"
+) -> LazyFrame[Any]: ...
```
FBruzzesi, line 962: I don't think these overloads are correct as it would allow `concat([pandas_df, pyarrow_df])` which is rejected at runtime

src/narwhals/stable/v1/__init__.py
```diff
@@ -948,6 +950,24 @@ def when(*predicates: IntoExpr | Iterable[IntoExpr]) -> When:
...
+
+
+def concat(
+    items: Iterable[_FrameT], *, how: ConcatMethod = "vertical"
```
FBruzzesi, line 966: You can use `FrameT` coming from `narwhals.stable.v1.typing` rather than re-defining it.
So the signature becomes:

```py
from narwhals.stable.v1.typing import FrameT

...

def concat(items: Iterable[FrameT], *, how: ConcatMethod = "vertical") -> FrameT: ...
```

src/narwhals/stable/v1/__init__.py
```diff
@@ -948,6 +950,24 @@ def when(*predicates: IntoExpr | Iterable[IntoExpr]) -> When:
...
+) -> DataFrame[Any] | LazyFrame[Any]:
+    return _stableify(nw_f.concat(items, how=how))
```
FBruzzesi, line 968: Docstrings are missing

src/narwhals/stable/v2/__init__.py
FBruzzesi, line 1: Same comments are per v1 case

tests/v1_test.py
```diff
@@ -1247,3 +1250,35 @@ def test_schema_from_generator() -> None:
...
+    https://github.com/narwhals-dev/narwhals/issues/3897
+    """
+    pytest.importorskip("pandas")
+    import pandas as pd
+
+    df = nw_v1.from_native(pd.DataFrame({"a": [1, 2, 3]}), eager_only=True)
+    result = nw_v1.concat([df], how="horizontal")
+    assert isinstance(result, nw_v1.DataFrame)
+    if TYPE_CHECKING:
+        assert_type(result, nw_v1.DataFrame[Any])
```
FBruzzesi, line 1267: The `assert_type` here is a no-op: the `isinstance` on the line above already narrows `result` to `nw_v1.DataFrame[Any]`, so the assertion is checking the narrowed type rather than what `concat` returned.

Swapping the order makes it real:

```suggestion
    if TYPE_CHECKING:
        # NOTE: must precede the `isinstance` below, which narrows `result`.
        assert_type(result, nw_v1.DataFrame[Any])
    assert isinstance(result, nw_v1.DataFrame)
```

tests/v2_test.py
```diff
@@ -586,3 +586,35 @@ def test_schema_from_generator() -> None:
...
+    https://github.com/narwhals-dev/narwhals/issues/3897
+    """
+    pytest.importorskip("pandas")
+    import pandas as pd
+
+    df = nw_v2.from_native(pd.DataFrame({"a": [1, 2, 3]}), eager_only=True)
+    result = nw_v2.concat([df], how="horizontal")
+    assert isinstance(result, nw_v2.DataFrame)
+    if TYPE_CHECKING:
+        assert_type(result, nw_v2.DataFrame[Any])
```
FBruzzesi, line 603: Same as v1 case

tests/v2_test.py
```diff
@@ -586,3 +586,35 @@ def test_schema_from_generator() -> None:
...
+    A constrained TypeVar over stable `Series` / `DataFrame` must accept the
+    result of stable `concat`.
+    """
+    pytest.importorskip("pandas")
+    import pandas as pd
+
+    df = nw_v2.from_native(pd.DataFrame({"a": [1, 2, 3]}), eager_only=True)
+
+    def identity(x: FrameT) -> FrameT:
+        return x
```
FBruzzesi, line 618: This `FrameT` is `narwhals.stable.v2.typing.FrameT`, which is constrained to `DataFrame` / `LazyFrame`. The docstring just above (lines 609-610) says `Series` / `DataFrame`, and that is also what the issue's repro uses. The v1 twin you added at `tests/v1_test.py:57` defines the `Series` / `DataFrame` one locally, so the two tests are checking different things under the same name and docstring.

Both catch the regression today, so nothing is broken, but I would mirror v1 here so the pair stays in sync and the test matches its docstring

FBruzzesi (changes requested):
Thanks for picking this up @heyilingran and for `test_concat_typevar` in particular: I verified it  fails on `main` under both mypy and pyright, so it is a genuine regression guard.

Two things before this can go in.

### Tests

* The `assert_type` in `test_concat_typing` is a no-op in both files, because the `isinstance` above it narrows `result` first. Inline suggestions below; reordering takes the live assertion count from 2 to 4.
* `test_concat_typing` and the existing `test_concat` differ only in `how=`. If you would rather not grow the file, folding the new `assert_type` and the `identity(...)` call into the existing `test_concat` would give the same coverage in one test per namespace instead of three.
 
### Typing

The overload shape regresses two generic-caller patterns that type-check on `main` today under pyright and pyrefly, including the `FrameT` idiom we document on `FrameT` itself, and it accepts `concat([pandas_df, pyarrow_df])`, which `main` rejects and which raises `AttributeError` at runtime. Details in a separate comment.

tests/v2_test.py
heyilingran: Thanks for the thorough review. I really appreciate the detailed feedback — it helped me understand the issues much more clearly. I’ll go through the comments and update the PR accordingly.

heyilingran:
Thanks again for the careful review, @FBruzzesi. I've pushed an update using the existing stable `FrameT` in both namespaces, moved the type assertions before `isinstance`, and folded the regression checks into the existing concat tests. The v1/v2 Series/DataFrame reproduction now matches, and both wrappers have docstrings. I also added eager/lazy generic-caller checks.

The type checkers, pre-commit checks, targeted tests, doctests, and strict docs build pass. The full suite still has four SQLFrame timezone failures and falls below the coverage gate locally; I confirmed the same failures and uncovered counts on unmodified `main` in the same environment. I've updated the PR description with the details.

Your comments helped me catch the generic-caller regression and the ineffective type assertions. Thank you for taking the time to explain them.

FBruzzesi (approved):
Thanks for bearing with me @heyilingran - I also added a tiny commit to validate mixing eager and lazy in concat is flagged by the type checkers

PR #3909, opened 2026-09-06
fix: `str.zfill(0)` on pyarrow-backed data; cover zfill sign-only and width-1 edge cases in tests

jonasdedden:
# Description

Adds the two `str.zfill` edge cases that pass uniformly today: sign-only/empty strings with width 3 (`-`/`+`/`""` -> `-00`/`+00`/`000`, a branch that returned null on some backends during narwhals-daft differential testing) and width 1 (`""` -> `"0"`, everything else unchanged).

Now covered: width 0 is pinned to polars (input unchanged) with a fix for the pyarrow backend, whose `utf8_lpad` rejected the negative width on an untaken branch; non-ASCII width 3 is pinned to polars (byte counting, input unchanged) with `xfail` for the character-counting backends.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
  
  `pytest tests/expr_and_series/str/zfill_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` -> 29 passed. Also green with `--constructors=ibis` and `--constructors=dask`.


FBruzzesi:
Hey @jonasdedden thanks for the contribution. I skimmed through all the test PRs you submitted (the one ready for review at least). Each description seems to end with a list of genuine cross-backend divergences found during differential testing, then discarding them. I would argue to go a different way about them, namely add them in the test with expected output matching polars, and then you have two options:

1. xfail the backends that differ (PR will be kept as testing only), and open an issue per item (or one umbrella issue with a checklist)
2. adjust the behavior of failing backends (PR will be a fix as well)

In here, what I refer is exactly:

> Not included: width 0 (pyarrow raises ArrowInvalid while others return input unchanged) and non-ASCII width 3 (日本: polars counts bytes and returns it unchanged, pandas/pyarrow/duckdb count characters and return 0日本). Both need a semantic decision first.

The semantic decision is "best effort" to align with polars behavior

jonasdedden:
@FBruzzesi thanks for the remark! :heart: 

To give a bit of context: For `narwhals-daft`, I let Fable generate a large test harness of 3000+ tests (partly) not currently part of `narwhals` to uncover inconsistencies between polars and Daft behavior. The PRs I opened pretty much are the set of tests that *were* uncovering inconsistencies between polars and Daft, but *not* failed with other backends currently natively (i.e. plugin-less) supported by narwhals. I didn't want to introduce failing tests into the narwhals suite.

But through your remark and a more thorough look into the narwhals test suite, it became clear that this isn't really to bad, so I'm following your suggestion! :) 

I'm currently going over all PRs and implement missing tests, and if the fix to a specific backend is trivial, I implement it right away too. 

The PRs still in Draft mode basically only are because I didn't want to spam you too much, but they should be review- and (if okay) mergable.

jonasdedden:
Okay, these PRs now also got "Bug fix" checked as they change stuff in `src/` too:

- #3909 
- #3910 
- #3912 
- #3916 
- #3918
- #3919

EDIT: If important for review triage, these PRs are tests only:

- #3911 
- #3913 
- #3914 
- #3915 
- #3917 

Reason for some of these still being Draft is that I'm not allowed to have too many PRs open in this repository.

jonasdedden:
Okay, with my original 11 PRs, the mentioned  previously omitted additions to each of them a few minutes ago and [this](https://github.com/narwhals-dev/narwhals/pull/3913/commits/58c1fef838d4c0dbad25c113f290c67ef073f5d4) additional commit now, all test shortcomings of `narwhals` compared to the ~3000 test cases now should be covered by PRs!

src/narwhals/_arrow/series_str.py
```diff
@@ -120,6 +120,10 @@ def to_titlecase(self) -> ArrowSeries:
         return self.with_native(pc.utf8_title(self.native))
 
     def zfill(self, width: int) -> ArrowSeries:
+        if width <= 0:
+            # `pc.utf8_lpad` rejects negative widths even on untaken branches,
+            # and no string is shorter than a non-positive width anyway.
```
FBruzzesi: You can remove these kind of inline comments 😉

tests/expr_and_series/str/zfill_test.py
FBruzzesi, line 1: Thanks! These are really nice to have edge cases. What I would suggest to do is to define all the cases upfront at the top of the file via `pytest.param` with `data` and `expected` as entries. In this way we can both track the "case" we are testing ("sign_only", "width_1", etc..) via the `id` argument, and use `pytest.mark.parametrize` to define one test only. By using parametrize we can avoid doing the same checks in multiple similar shaped functions and everything collapses into the `test_str_zfill` test

FBruzzesi (changes requested):
Thanks for following up on everything and explaining how you came up with all these tests. In general the more tests we have the better. They have been extremely useful both in detecting each library shortcomings _and_ spotting regressions.

Coming to _this specific PR_ (I still need to check implementation for the others), I am suggesting to trim inline naive comments and exploit pytest parametrization whenever the shape of a test is the same.

---

Update: Some CI job is failing for either old polars or modin with pyarrow dtypes. In both cases we can add a fast path for `width=0` as for pyarrow

tests/expr_and_series/str/zfill_test.py
jonasdedden: Done, thanks for the suggestion!

src/narwhals/_arrow/series_str.py
jonasdedden: Done ;)

jonasdedden:
~Now CI should hopefully be green~ :sweat_smile:  EDIT: Now, maybe? :pray: 

Btw., I checked for similar test errors in all other PRs and also removed silly/unnecessary comments accordingly. Hopefully things are now good/better across the board.

FBruzzesi (approved):
Thanks a ton @jonasdedden 🚀

PR #3910, opened 2026-09-06
fix: reject invalid `str.{pad_start,pad_end}` arguments consistently; cover pad length-zero and unicode edge cases in tests

jonasdedden:
# Description

Covers two `str.pad_start`/`pad_end` cases that hold on all backends: length 0 returns input unchanged (per docstring), and unicode input already exactly `length` characters long stays unchanged.

Multi-character and empty `fill_char`, and negative `length`, are now validated in the public `Expr.str`/`Series.str` accessors, so every backend raises the same `ValueError` up-front. Previously each backend either raised something different, silently returned the input unchanged, or (dask) deferred the error to collect time.

Two user-visible changes for polars, which previously validated natively:

- negative `length` raised `InvalidOperationError`, now `ValueError`
- errors surface at expression construction rather than at evaluation, so
  `nw.col("a").str.pad_start(-1)` raises immediately instead of at `select`/`collect`

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
  
  `pytest tests/expr_and_series/str/pad_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` -> 68 passed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/src/narwhals/_ibis/expr_str.py
+++ b/src/narwhals/_ibis/expr_str.py
@@ -7,7 +7,7 @@ import ibis
 from ibis.expr.datatypes import Timestamp
 
 from narwhals._sql.expr_str import SQLExprStringNamespace
-from narwhals._utils import _is_naive_format, not_implemented, validate_pad_arguments
+from narwhals._utils import _is_naive_format, not_implemented
 
 if TYPE_CHECKING:
     from collections.abc import Callable
@@ -97,8 +97,6 @@ class IbisExprStringNamespace(SQLExprStringNamespace["IbisExpr"]):
         )
 
     def pad_start(self, length: int, fill_char: str) -> IbisExpr:
-        validate_pad_arguments("pad_start", length, fill_char)
-
         def _pad_start(expr: ir.StringColumn) -> ir.Value:
             padded = expr.lpad(length, fill_char)
             return ibis.cases((expr.notnull(), padded))
@@ -106,8 +104,6 @@ class IbisExprStringNamespace(SQLExprStringNamespace["IbisExpr"]):
         return self.compliant._with_callable(_pad_start)
 
     def pad_end(self, length: int, fill_char: str) -> IbisExpr:
-        validate_pad_arguments("pad_end", length, fill_char)
-
         def _pad_end(expr: ir.StringColumn) -> ir.Value:
             padded = expr.rpad(length, fill_char)
             return ibis.cases((expr.notnull(), padded))
--- a/src/narwhals/_ibis/expr_str.py
+++ b/src/narwhals/_ibis/expr_str.py
@@ -97,6 +97,13 @@ class IbisExprStringNamespace(SQLExprStringNamespace["IbisExpr"]):
         )
 
     def pad_start(self, length: int, fill_char: str) -> IbisExpr:
+        if len(fill_char) != 1:
+            msg = f"`str.pad_start` only supports single-character `fill_char`, got {fill_char!r}."
+            raise ValueError(msg)
+        if length < 0:
+            msg = f"`str.pad_start` length must be non-negative, got {length}."
+            raise ValueError(msg)
+
         def _pad_start(expr: ir.StringColumn) -> ir.Value:
             padded = expr.lpad(length, fill_char)
             return ibis.cases((expr.notnull(), padded))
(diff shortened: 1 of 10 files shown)


src/narwhals/_ibis/expr_str.py
```diff
@@ -97,13 +97,27 @@ def to_time(self, format: str | None) -> IbisExpr:
...
 
         return self.compliant._with_callable(_pad_start)
 
     def pad_end(self, length: int, fill_char: str) -> IbisExpr:
+        if len(fill_char) != 1:
+            msg = f"`str.pad_end` only supports single-character `fill_char`, got {fill_char!r}."
+            raise ValueError(msg)
+        if length < 0:
+            msg = f"`str.pad_end` length must be non-negative, got {length}."
+            raise ValueError(msg)
```
FBruzzesi: I would upstream these checks via a helper method in `StringNamespace` from which all other `<Backend>StringNamespace`'s inherit from, except polars. In this way:

* polars does not duplicate work
* In all other compliant level's we can do:
    ```py
    def pad_end(self, length: int, fill_char: str) -> ...:
        self.validate_inputs(fill_char, length, method_name)
        ...
    ```
	**Important**: Please come up with a much better method name

FBruzzesi (commented):
Thanks @jonasdedden - Left a suggestion to avoid repeating ourselves in so many places

src/narwhals/_ibis/expr_str.py
jonasdedden: Done in d957107a. Note I did not put the helper on `StringNamespace` itself as it's a Protocol. I introduced a `ValidatesPadArgs` mixin; let me know what you think.

Two nice side effects, since dask and pandas-like/pyarrow now validate up-front as well:

* dask raises eagerly instead of at collect time, so the `xfail` in `test_pad_invalid_fill_char` is gone
* old pandas no longer silently accepts a negative length, so the `PANDAS_VERSION < (3,)` `xfail` in `test_pad_negative_length_raises` is gone too

jonasdedden: Replaced it with a bare Callable in https://github.com/narwhals-dev/narwhals/pull/3910/commits/b54d71edb2f6125f24fe7c9e3d3714f6daf597fb, let me know which variant you prefer.

FBruzzesi (commented):
Hey @jonasdedden - apologies for the slow feedback loops. It's a bit though to keep up with everything these days!

That said, I think the callable/pure function approach is the right one. One quick check I did was to benchmark the cost of moving it to the top level (i.e. in `series_str.py` and `expr_str.py`) before dispatching to the compliant level. The cost of the check is fixed (locally an average of ~32ns), and polars `pad_start` on a 100k rows run in ~818 us, this validation was ~0.00397% of the total cost 😅

Considering that, I would be in favor of moving the checks to the compliant level to simplify things

jonasdedden:
Thanks and no worries about the delay!

Moved to the top level: `validate_pad_arguments` now runs in `ExprStringNamespace.pad_start/pad_end` and `SeriesStringNamespace.pad_start/pad_end`, and is gone from all five compliant backends. I got the same result as your benchmark: 45.5 ns/call against 2294 µs for `pad_start` over 100k polars rows.

Nice side effect: the PR no longer touches any backend file. Two user-visible changes worth flagging, both from polars now going through our validation too:

- Exception type changes for polars on negative length: `InvalidOperationError` -> `ValueError`. `fill_char` was already `ValueError`, just with polars' own message.
- Errors surface at expression construction rather than at evaluation, so `nw.col("a").str.pad_start(-1)` raises immediately instead of at select/collect.

Since the message is now identical on every backend, the tests could drop the broad Exception and use `match=`, and I added a Series-path case since that accessor validates too.

FBruzzesi (approved):
Thanks @jonasdedden 🚀

FBruzzesi (approved):
Thanks @jonasdedden 🚀

PR #3911, opened 2026-09-06
fix: `str.replace(n=0)` on pandas-like backends; cover replace regex and `n` edge cases in tests

jonasdedden:
# Description

Parametrized `replace_edge_data` plus expr/series tests for the `str.replace` cases that agree across eager backends: literal replacement on non-ASCII text (`ααα-x`, `héllo wörld` - byte-vs-char offsets), a regex whose match text also
occurs earlier (`abcx abc` with `abc\b`), replacement containing the pattern with `n=2` (`aa` -> `abab`), `n=0` unchanged, and empty-pattern prepend. Expr tests xfail for `pyspark`/`duckdb`/`ibis` (replace_all-only), matching existing tests. The Greek-α row carries `# noqa: RUF001` (precedent in `to_uppercase_to_lowercase_test.py`): the confusable characters are the point.

Now covered, pinned to polars with `xfail` for the rest: `$1` capture-group expansion (pandas-likes and pyarrow splice it literally), backslash-one without a group (stays literal on polars, rejected elsewhere), and null replacement values (polars leaves the input unchanged, pandas-likes and pyarrow only accept strings). Regex-engine differences are intentionally not fixed here. The stale "n>1 with regex raises" claim from the gap notes is dropped: it only holds for multivalue replacements.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
  
  `pytest tests/expr_and_series/str/replace_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` -> 202 passed, 84 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/tests/expr_and_series/str/replace_test.py
+++ b/tests/expr_and_series/str/replace_test.py
@@ -189,24 +189,49 @@ def test_str_replace_edge_series_scalar(
     assert_equal_data({"a": result_series}, expected)
 
 
-def test_str_replace_group_expansion_series_scalar(
+replace_value_edge_cases = [
+    pytest.param(
+        "(b)",
+        "[$1]",
+        {"a": ["aa[b]"]},
+        "no capture-group expansion",
+        id="group_expansion",
+    ),
+    pytest.param(
+        "b", "\\1", {"a": ["aa\\1"]}, "invalid replacement string", id="backslash_literal"
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    ("pattern", "value", "expected", "reason"), replace_value_edge_cases
+)
+def test_str_replace_value_edge_series_scalar(
     constructor_eager: ConstructorEager,
     request: pytest.FixtureRequest,
+    pattern: str,
+    value: str,
+    expected: dict[str, list[str]],
+    reason: str,
 ) -> None:
-    # `$1` in a regex replacement expands the capture group, matching polars.
-    # pandas-likes and pyarrow splice it in literally instead.
+    # Special replacement strings follow polars; pandas-likes and pyarrow differ.
     if any(x in str(constructor_eager) for x in ("pandas", "modin", "cudf", "pyarrow")):
-        request.applymarker(pytest.mark.xfail(reason="no capture-group expansion"))
+        request.applymarker(pytest.mark.xfail(reason=reason))
     df = nw.from_native(constructor_eager({"a": ["aab"]}), eager_only=True)
-    result_series = df["a"].str.replace(
-        pattern="(b)", value="[$1]", n=1, literal=False
-    )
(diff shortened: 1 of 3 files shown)


tests/expr_and_series/str/replace_test.py
```diff
@@ -159,6 +196,187 @@ def test_str_replace_expr_scalar(
...
+    ("data", "pattern", "value", "n", "literal", "expected"), replace_edge_data
+)
+def test_str_replace_edge_series_scalar(
+    constructor_eager: ConstructorEager,
+    request: pytest.FixtureRequest,
+    data: dict[str, list[str]],
+    pattern: str,
+    value: str,
+    n: int,
+    literal: bool,  # noqa: FBT001
```
FBruzzesi:
```suggestion
    n: int,
    *,
    literal: bool,
```

tests/expr_and_series/str/replace_test.py
```diff
@@ -159,6 +196,187 @@ def test_str_replace_expr_scalar(
...
+    ("data", "pattern", "value", "n", "literal", "expected"), replace_edge_data
+)
+def test_str_replace_edge_expr_scalar(
+    constructor: Constructor,
+    request: pytest.FixtureRequest,
+    data: dict[str, list[str]],
+    pattern: str,
+    value: str,
+    n: int,
+    literal: bool,  # noqa: FBT001
```
FBruzzesi:
```suggestion
    n: int,
    *,
    literal: bool,
```

FBruzzesi (commented):
Thanks @jonasdedden , for the (`n=0`, pandas-like) case, can't we add the check as you did in other PRs to make it a no-op?

tests/expr_and_series/str/replace_test.py
jonasdedden: Thanks!

tests/expr_and_series/str/replace_test.py
jonasdedden: Thanks!

jonasdedden:
> Thanks @jonasdedden , for the (n=0, pandas-like) case, can't we add the check as you did in other PRs to make it a no-op?

Yes, and that cleared some test up too

FBruzzesi (approved):
Thanks @jonasdedden , I added a tiny commit to apply the same fix to dask 🙏🏼

PR #3912, opened 2026-09-06
fix(pyarrow): use native `{starts,ends}_with` kernels, fixing empty-suffix `ends_with`; cover empty-pattern and null expr-pattern search in tests

jonasdedden:
# Description

`contains("")` matches every non-null row with null preserved (pandas-conditional expected, as in existing tests), `starts_with("")` likewise, and a column `contains` pattern holding null yields null for that row (xfailed where expression patterns are unsupported, as in existing tests).

Now covered: empty `ends_with` is pinned to `True` for non-null rows with a fix for the pyarrow backend (its negative-zero slice compared the whole string), and invalid regex is pinned to raising (generic `Exception`, each backend has its own type, lazy backends raise at collect time).

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
  
  `pytest tests/expr_and_series/str/contains_test.py tests/expr_and_series/str/starts_with_ends_with_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` -> 155 passed, 52 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/src/narwhals/_arrow/series_str.py
+++ b/src/narwhals/_arrow/series_str.py
@@ -58,29 +58,14 @@ class ArrowSeriesStringNamespace(ArrowSeriesNamespace, StringNamespace["ArrowSer
         if not isinstance(prefix_native, pa.StringScalar):
             msg = "`.str.starts_with` only supports str prefix values for pyarrow backend"
             raise TypeError(msg)
-        return self.with_native(
-            pc.equal(
-                self.slice(0, len(prefix_native.as_py())).native,
-                lit(prefix_native.as_py()),
-            )
-        )
+        return self.with_native(pc.starts_with(self.native, prefix_native.as_py()))
 
     def ends_with(self, suffix: ArrowSeries) -> ArrowSeries:
         _, suffix_native = extract_native(self.compliant, suffix)
         if not isinstance(suffix_native, pa.StringScalar):
             msg = "`.str.ends_with` only supports str suffix values for pyarrow backend"
             raise TypeError(msg)
-        if suffix_native.as_py() == "":
-            # Every string ends with the empty string, but `-0 == 0`, so the
-            # slice below would compare the whole string against `""`.
-            # `match_substring` with an empty pattern preserves nulls instead.
-            return self.with_native(pc.match_substring(self.native, ""))
-        return self.with_native(
-            pc.equal(
-                self.slice(-len(suffix_native.as_py()), None).native,
-                lit(suffix_native.as_py()),
-            )
-        )
+        return self.with_native(pc.ends_with(self.native, suffix_native.as_py()))
 
     def contains(self, pattern: ArrowSeries, *, literal: bool) -> ArrowSeries:
         _, pattern_native = extract_native(self.compliant, pattern)
--- a/src/narwhals/_arrow/series_str.py
+++ b/src/narwhals/_arrow/series_str.py
@@ -70,6 +70,11 @@ class ArrowSeriesStringNamespace(ArrowSeriesNamespace, StringNamespace["ArrowSer
         if not isinstance(suffix_native, pa.StringScalar):
             msg = "`.str.ends_with` only supports str suffix values for pyarrow backend"
             raise TypeError(msg)
+        if suffix_native.as_py() == "":
+            # Every string ends with the empty string, but `-0 == 0`, so the
+            # slice below would compare the whole string against `""`.
+            # `match_substring` with an empty pattern preserves nulls instead.
+            return self.with_native(pc.match_substring(self.native, ""))
         return self.with_native(
             pc.equal(
                 self.slice(-len(suffix_native.as_py()), None).native,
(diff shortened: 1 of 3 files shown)


tests/expr_and_series/str/contains_test.py
```diff
@@ -242,3 +242,27 @@ def test_series_contains_literal_vs_regex(constructor_eager: ConstructorEager) -
...
     assert_equal_data(result, expected)
+
+
+def test_expr_contains_invalid_regex(constructor: Constructor) -> None:
+    df = nw.from_native(constructor({"pets": ["cat", "dog"]}))
+    expr = nw.col("pets").str.contains("(", literal=False)
+    # Broad `Exception`: error types differ per backend.
+    if isinstance(df, nw.LazyFrame):
+        with pytest.raises(Exception):  # noqa: B017, PT011
+            df.select(expr).lazy().collect()
```
FBruzzesi: Why do we need `.lazy` call if `isinstance(df, nw.LazyFrame)`?

src/narwhals/_arrow/series_str.py
```diff
@@ -70,6 +70,11 @@ def ends_with(self, suffix: ArrowSeries) -> ArrowSeries:
...
             msg = "`.str.ends_with` only supports str suffix values for pyarrow backend"
             raise TypeError(msg)
+        if suffix_native.as_py() == "":
+            # Every string ends with the empty string, but `-0 == 0`, so the
+            # slice below would compare the whole string against `""`.
+            # `match_substring` with an empty pattern preserves nulls instead.
+            return self.with_native(pc.match_substring(self.native, ""))
         return self.with_native(
             pc.equal(
                 self.slice(-len(suffix_native.as_py()), None).native,
```
FBruzzesi: Should we use `match_substring` for everything by adding an end anchor? Something along the lines of:

```py
return self.with_native(
    pc.match_substring(self.native, f"{suffix_native.as_py()}$")
)
```

I assume it's more efficient than slicing + comparing if that works?

FBruzzesi: Much simpler, TIL that [`pc.starts_with`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.starts_with.html) and [`pc.ends_with`](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ends_with.html) actually exist for quite some time. Not sure why we are not using them, it would be nice if you could edit both. That's also some nice speedup

tests/expr_and_series/str/starts_with_ends_with_test.py
```diff
@@ -135,6 +135,20 @@ def test_starts_with_null(constructor: Constructor) -> None:
     assert_equal_data(result, expected)
 
 
+@pytest.mark.parametrize("method", ["contains", "starts_with", "ends_with"])
```
FBruzzesi, line 138: Can you add a comment in `str/contains_test.py` that the empty case in tested here for convenience?

tests/expr_and_series/str/contains_test.py
```diff
@@ -242,3 +242,27 @@ def test_series_contains_literal_vs_regex(constructor_eager: ConstructorEager) -
...
+    else:
+        with pytest.raises(Exception):  # noqa: B017, PT011
+            df.select(expr)
```
FBruzzesi: We have a (tests) utils: `maybe_collect`, which literally is does `df.collect() if isinstance(df, nw.LazyFrame) else df`. I think you can use it and keep a single raising context.

On the context itself, `pytest.raises(Exception)` also passes if a backend raises `NotImplementedError`, or if the test itself develops a typo and raises `ColumnNotFoundError`. Consider assert not `isinstance(exc_info.value, NotImplementedError)`, or the per-backend context idiom already used by `test_to_datetime_tz_aware` (I know if a bit of a headache)

FBruzzesi (commented):
Thanks @jonasdedden , left a couple of suggestions 🙏🏼

src/narwhals/_arrow/series_str.py
jonasdedden: Used that instead, much much simpler

tests/expr_and_series/str/starts_with_ends_with_test.py
jonasdedden: Added 
```
# NOTE: the empty-pattern case lives in `test_empty_pattern` in
# `starts_with_ends_with_test.py` for convenience.
```

tests/expr_and_series/str/contains_test.py
jonasdedden: For now I went for `assert not isinstance(...)`, as the per-backend specific variant is ~25LOC+ at least in my implementation. Let me know whether you still want to have it more specific.

FBruzzesi (approved):
Thanks a ton @jonasdedden

PR #3913, opened 2026-09-06
test: cover to_datetime date-only, all-null and trailing cases

jonasdedden:
# Description

Explicit `"%Y-%m-%d"` parses to midnight (with a `Datetime` schema assertion); an explicit offset format on an all-null column yields all nulls (values only, since time units and time zones differ by backend; ibis is skipped because it cannot construct an all-null column); trailing input with an explicit format raises on all backends, asserted with a generic `Exception` plus a `B017, PT011` noqa and justification comment (each backend has its own error type, so the pinned contract is only "raises rather than prefix-parses").

Now covered: fractional-second formats pin polars dtypes (millis and micros variants) with `xfail` for every other backend, and unparseable input (explicit or inferred format) is pinned to raising via generic `Exception`. Strict cross-backend error-type and time-unit assertions remain open.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added

    ```bash
    ```

Note: the full-suite command above was not run. Targeted run on the CI constructor set instead: `pytest tests/expr_and_series/str/to_datetime_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` gives 97 passed, 8 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/tests/expr_and_series/str/to_datetime_test.py
+++ b/tests/expr_and_series/str/to_datetime_test.py
@@ -275,24 +275,45 @@ def test_to_datetime_all_null_offset_format(constructor: Constructor) -> None:
 
 
 @pytest.mark.parametrize(
-    ("format", "time_unit"),
-    [("%Y-%m-%dT%H:%M:%S%.3f", "ms"), ("%Y-%m-%dT%H:%M:%S%.f", "us")],
+    ("format", "time_unit", "value", "expected"),
+    [
+        (
+            "%Y-%m-%dT%H:%M:%S%.3f",
+            "ms",
+            "2020-01-01T12:34:56.123",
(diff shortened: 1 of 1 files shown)


jonasdedden:
@FBruzzesi why exactly is this PR considered blocked? :thinking:  EDIT: Ah sorry, because of https://github.com/pola-rs/polars/issues/29247.

Issue #3900, opened 2026-09-04
Decide on a cross-backend coercion for `is_in` & `__contains__`


FBruzzesi:
## Description

Polars 2.0 is making coercion casts for `is_in()` strict instead of lossy: operands are only coerced when the cast is lossless, and an `InvalidOperationError` is raised otherwise.

Every other backend still coerces silently, and Narwhals passes the call straight through. The result is that identical Narwhals code returns a boolean on one backend and raises on another.

Short term we let the Polars error propagate and adapt the tests (@MarcoGorelli 's [discord comment](https://discord.com/channels/1235257048170762310/1284535800633229312/1545387689790935050). This issue tracks the underlying inconsistency, which I would not want to leave unresolved.

## What Polars 2 actually enforces

Observed on `2.0.0rc1` by comparing pair of dtypes:

| column dtype | accepted value dtype |
|---|---|
| any integer | any integer, signed or unsigned, any width |
| `Float32` / `Float64` | any float |
| `Decimal` | any numeric, integers and floats included |
| `String` | `String`, `Categorical`, `Enum` |
| `Categorical` | `Categorical`, `String` (but not `Enum`) |
| `Datetime` | `Datetime`, only where the time unit converts losslessly (`ms` into `us` is fine, `us` into `ms` is not) |
| anything else | the same dtype |

## Open questions

- Do we want to follow polars strict behavior? For eager backends we can achieve that by checking in narwhals, also for polars. That could mean diverging between (polars) Series and Expr for pre v2 versions.
- If strict, do we accept that behaviour is driven by the *dtype* rather than the *values*, which means non-nullable pandas diverges whenever nulls force a column to `Float64`?
- Do lazy backends stay unvalidated (and therefore inconsistent with eager ones), or do we thread the schema into `is_in`?
- Whatever we pick, `Expr.is_in` / `Series.is_in` / `Series.__contains__` should say so in their docstrings.

## Experimented so far

I tried to implement a strict version in [fix/polars-v2-is_in](https://github.com/narwhals-dev/narwhals/tree/fix/polars-v2-is_in), in which lazy backends are left unvalidated.

Implementing `DType.from_python` (see https://github.com/narwhals-dev/narwhals/issues/2264) first would greatly help.

MarcoGorelli:
my suggestion would be to just let it vary by backend, we (narwhals) just don't make any guarantees

FBruzzesi:
[...]

PR #3915, opened 2026-09-06
test: cover list.contains numeric coercion and all-null inner

jonasdedden:
# Description

Numeric width and family coercion (`2.0` matches ints, `1.5` does not, `300` overflows `Int8`, `1` matches a `Float64` list) and an all-null inner list containing nothing. Same xfail pattern as the existing tests for `dask`/`modin`/`cudf`/`pyarrow`/`pandas`; ibis is skipped for the all-null fixture because it cannot construct one.

Now covered, pinned to polars: a `None` item matches null-holding lists (SQL backends return null everywhere instead, hence `xfail`), mismatched bool or str items on int lists raise (SQL backends coerce instead, hence `xfail`), and precision-mismatched datetime items raise (SQL backends coerce instead, hence `xfail`).

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added

    ```bash
    ```

Note: the full-suite command above was not run. Targeted run on the CI constructor set instead: `pytest tests/expr_and_series/list/contains_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` gives 13 passed, 16 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/tests/expr_and_series/list/contains_test.py
+++ b/tests/expr_and_series/list/contains_test.py
@@ -7,7 +7,7 @@ import pytest
 
 import narwhals as nw
 from narwhals.exceptions import InvalidOperationError
-from tests.utils import assert_equal_data
+from tests.utils import POLARS_VERSION, assert_equal_data
 
 if TYPE_CHECKING:
     from tests.utils import Constructor, ConstructorEager
@@ -96,7 +96,9 @@ def test_contains_numeric_coercion_expr(
 def test_contains_none_item_expr(
     request: pytest.FixtureRequest, constructor: Constructor
 ) -> None:
-    # SQL backends return null for every row instead.
+    # SQL backends return null for every row instead. Old polars does too.
+    if "polars" in str(constructor) and POLARS_VERSION < (1, 24, 0):
+        request.applymarker(pytest.mark.xfail(reason="old polars null item"))
     if any(
         backend in str(constructor)
         for backend in ("dask", "modin", "cudf", "pyarrow", "pandas")
@@ -132,6 +134,13 @@ def test_contains_invalid_item_raises(
     item: bool | str | datetime,
 ) -> None:
     # Mismatched items raise, matching polars; SQL backends coerce instead.
+    # Old polars coerces precision-mismatched datetimes instead of raising.
+    if (
+        "datetime_precision" in request.node.callspec.id
+        and "polars" in str(constructor)
+        and POLARS_VERSION < (1, 28, 0)
+    ):
+        request.applymarker(pytest.mark.xfail(reason="old polars precision"))
     if any(
         backend in str(constructor)
         for backend in ("dask", "modin", "cudf", "pyarrow", "pandas")
--- a/tests/expr_and_series/list/contains_test.py
+++ b/tests/expr_and_series/list/contains_test.py
@@ -41,39 +41,39 @@ def test_contains_series(
     assert_equal_data({"a": result}, expected)
 
 
-coercion_cases = [
-    pytest.param(
-        {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
-        nw.Int64(),
-        2.0,
-        [True, None, False, False, False, False],
-        id="float_matches_int",
-    ),
-    pytest.param(
-        {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
-        nw.Int64(),
-        1.5,
-        [False, None, False, False, False, False],
-        id="non_integer",
-    ),
-    pytest.param(
-        {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
-        nw.Int8(),
-        300,
-        [False, None, False, False, False, False],
-        id="overflow",
-    ),
-    pytest.param(
-        {"a": [[1.0, 2.0], [3.0], None, []]},
-        nw.Float64(),
-        1,
-        [True, False, None, False],
-        id="int_matches_float",
-    ),
-]
-
-
-@pytest.mark.parametrize(("data", "inner", "item", "expected"), coercion_cases)
+@pytest.mark.parametrize(
+    ("data", "inner", "item", "expected"),
+    [
+        pytest.param(
+            {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
+            nw.Int64(),
+            2.0,
+            [True, None, False, False, False, False],
+            id="float_matches_int",
+        ),
+        pytest.param(
+            {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
+            nw.Int64(),
+            1.5,
+            [False, None, False, False, False, False],
+            id="non_integer",
+        ),
+        pytest.param(
+            {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
+            nw.Int8(),
+            300,
+            [False, None, False, False, False, False],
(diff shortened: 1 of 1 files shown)


tests/expr_and_series/list/contains_test.py
```diff
@@ -29,11 +35,137 @@ def test_contains_expr(request: pytest.FixtureRequest, constructor: Constructor)
...
+    data: dict[str, list[list[int | float | None] | None]],
+    inner: IntoDType,
+    item: float,
+    expected: list[bool | None],
+) -> None:
+    # Polars 2.0 requires an explicit cast rather than lossily coercing to `Float64`.
+    # `overflow` compares two integer types, so it still resolves.
+    if (
+        "polars" in str(constructor)
+        and POLARS_VERSION >= (2,)
```
jonasdedden, line 90: @FBruzzesi / @MarcoGorelli important:

Before version 2.0, Polars would perform this check by lossily coercing the operands to Float64. However, since Polars 2.0, for is_in() it is required to explicitly cast (one of) the operands to a compatible type.

Do we want to either:

1. Follow polars 2.0 => validate up front and raise on mixed numeric kinds across all backends. Consistent and explicit, but a breaking change for existing pandas/pyarrow/SQL users. 
2. Keep coercing => narwhals casts the item to the column's type before dispatching, so polars 2.0 keeps working and all backends agree. Preserves today's behaviour, but re-introduces the lossiness polars just removed.

MarcoGorelli: i'm inclined to keep it simple and unspecified, just let each backend do its thing here

we've already got a docs note warning about this

FBruzzesi: Tracking the discussion in #3900 - I personally more in favor in aligning behavior, but that's a breaking changes and, as Marco said, for now we decided to just document the divergence

jonasdedden: Ah sorry, didn't see that basically exactly this issue was already tracked somewhere.

For my understanding: Would you be fine then with removing this specific test entirely from this PR to keep the scope clearer? Or do you want me to have this current state as-is, basically just a test for documenting purposes?

tests/expr_and_series/list/contains_test.py
```diff
@@ -1,24 +1,30 @@
...
 
 data = {"a": [[2, 2, 3, None, None], None, []]}
 expected = {"a": [True, None, False]}
 
+UNSUPPORTED_EAGER_BACKENDS = ("cudf", "modin", "pandas", "pyarrow")
+"""Eager backends where `list.contains` is not implemented."""
+
+UNSUPPORTED_BACKENDS = ("dask", *UNSUPPORTED_EAGER_BACKENDS)
+"""Backends where `list.contains` is not implemented."""
+
```
FBruzzesi, line 26: As Marco used to say: "in for a penny in for a pound"

This could also be useful across the file:

```python
SQL_BACKENDS = ("duckdb", "ibis", "pyspark", "sqlframe")
"""Backends which coerce the item to the inner dtype instead of matching polars."""
```

tests/expr_and_series/list/contains_test.py
```diff
@@ -29,11 +35,137 @@ def test_contains_expr(request: pytest.FixtureRequest, constructor: Constructor)
...
     df = nw.from_native(constructor_eager(data), eager_only=True)
     result = df["a"].cast(nw.List(nw.Int32())).list.contains(2)
     assert_equal_data({"a": result}, expected)
+
+
+@pytest.mark.parametrize(
+    ("data", "inner", "item", "expected"),
+    [
+        pytest.param(
+            {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]},
```
FBruzzesi: Worth defining it once?

```py
INT_LIST_DATA = {"a": [[2, 2, 3, None, None], None, [], [None], [1], [0, -1]]}
```

tests/expr_and_series/list/contains_test.py
```diff
@@ -29,11 +35,137 @@ def test_contains_expr(request: pytest.FixtureRequest, constructor: Constructor)
...
+        ),
+    ],
+)
+def test_contains_invalid_item_raises(
+    request: pytest.FixtureRequest,
+    constructor: Constructor,
+    data: dict[str, list[list[int | str | datetime] | None]],
+    inner: IntoDType,
+    *,
+    item: bool | str | datetime,
```
FBruzzesi: That's probably a job for:

```suggestion:
    item: NonNestedLiteral,
```

coming from `narwhals.typing` module

tests/expr_and_series/list/contains_test.py
```diff
@@ -29,11 +35,137 @@ def test_contains_expr(request: pytest.FixtureRequest, constructor: Constructor)
...
+    if any(x in str(constructor) for x in ("duckdb", "sqlframe", "ibis", "pyspark")):
+        request.applymarker(pytest.mark.xfail(reason="mismatched item coerced"))
+    df = nw.from_native(constructor(data))
+    expr = nw.col("a").cast(nw.List(inner)).list.contains(item)
+    if isinstance(df, nw.LazyFrame):
+        with pytest.raises(InvalidOperationError):
+            df.select(expr).lazy().collect()
+    else:
+        with pytest.raises(InvalidOperationError):
+            df.select(expr)
```
FBruzzesi: Use `maybe_collect` in here

tests/expr_and_series/list/contains_test.py
FBruzzesi: For now I would only consider adding a comment referencing the issue, something along the lines of:

```py
# Mixing numeric kinds is deliberately unspecified, see
# https://github.com/narwhals-dev/narwhals/issues/3900. This pins what each backend
# does today so a future change is visible, it is not a guarantee.
```

on top of the existing comment describing polars v2 behavior

FBruzzesi (commented):
Thanks @jonasdedden , just a few minor comments 🙌🏼

tests/expr_and_series/list/contains_test.py
jonasdedden: Added the comment referencing #3900 above the Polars 2.0 note, so it's clear this records what each backend does today rather than a guarantee.

tests/expr_and_series/list/contains_test.py
jonasdedden: Done. INT_LIST_DATA is defined once and reused by the three integer-list cases.

tests/expr_and_series/list/contains_test.py
jonasdedden: Done. item is typed as NonNestedLiteral now, which also let me drop the keyword-only *.

tests/expr_and_series/list/contains_test.py
jonasdedden: Done

tests/expr_and_series/list/contains_test.py
jonasdedden: Done. SQL_BACKENDS now replaces the inline tuples, and a small xfail_unsupported helper removes the repeated unsupported-backend checks.

PR #3916, opened 2026-09-06
fix: align pandas-like and Dask `concat_str` boolean rendering

jonasdedden:
# Description

Cover single-expression passthrough (null preserved), null literals under the default `ignore_nulls=False`, and all-null input with `ignore_nulls=True`.

This PR also proposes lowercase boolean rendering in `concat_str` for pandas-like and Dask backends, matching Polars while leaving actual strings unchanged. Tests cover nullable booleans, both null-handling modes, and multiple Dask partitions.

**Still to discuss:** should Narwhals support concatenating booleans with strings and standardize the result, or leave support and rendering to each backend? The boolean implementation and tests here represent a proposal, not an established API guarantee. Mixed object-dtype boolean cases remain `xfail`d.

Rebased onto `main`, retaining the upstream fixes from #3939 and #3941 and removing obsolete pandas/PyArrow null-related `xfail`s. Added regression coverage for literal and aggregation broadcasting. Dask retains its existing string representation and pre-cast null masks to avoid an empty Arrow-string metadata failure on pandas 2.0.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

Follow-up from narwhals-daft differential testing and the review of this PR.

Related: #3939 and #3941. The Dask trailing-separator and `mean_horizontal` fixes from #3967 and #3968 are already included through the rebase.

## AI assistance

- [x] AI tools were used.

Delta was used for implementation, regression tests, and review.

## Checklist

- [x] Tests added
- [ ] Documented the changes (boolean-concatenation semantics remain under discussion)

```bash
CI=true GITHUB_ACTIONS=true PYTEST_XDIST_AUTO_NUM_WORKERS=8 \
make test-full-coverage
```

### DIFF
--- a/src/narwhals/_pandas_like/namespace.py
+++ b/src/narwhals/_pandas_like/namespace.py
@@ -370,10 +370,17 @@ class PandasLikeNamespace(
         self, *exprs: PandasLikeExpr, separator: str, ignore_nulls: bool
     ) -> PandasLikeExpr:
         string = self._version.dtypes.String()
+        boolean = self._version.dtypes.Boolean()
 
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             expr_results = [s for _expr in exprs for s in _expr(df)]
-            series = [s.cast(string) for s in expr_results]
+            # Polars renders booleans lowercase, pandas `astype(str)` does not.
+            series = [
+                s.cast(string).str.to_lowercase()
+                if s.dtype == boolean
+                else s.cast(string)
+                for s in expr_results
+            ]
             null_mask = [s.is_null() for s in expr_results]
 
             if not ignore_nulls:
--- a/tests/expr_and_series/concat_str_test.py
+++ b/tests/expr_and_series/concat_str_test.py
@@ -72,6 +72,24 @@ def test_concat_str_with_lit(constructor: Constructor) -> None:
     assert_equal_data(result, expected)
 
 
+def test_concat_str_single_expr(constructor: Constructor) -> None:
+    # A single expression is passed through (null preserved).
+    df = nw.from_native(constructor({"b": ["a", None]}))
+    result = df.select(nw.concat_str(["b"], separator=", ").alias("out"))
+    assert_equal_data(result, {"out": ["a", None]})
+
+
+def test_concat_str_null_literal(constructor: Constructor) -> None:
+    # A null literal poisons the row when `ignore_nulls=False` (default).
+    df = nw.from_native(constructor({"b": ["a", None]}))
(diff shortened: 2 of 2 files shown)


tests/expr_and_series/concat_str_test.py
```diff
@@ -72,6 +78,70 @@ def test_concat_str_with_lit(constructor: Constructor) -> None:
...
+        pytest.param(
+            {"b": ["a", None]},
+            ["b", nw.lit(None, dtype=nw.String())],
+            {"out": [None, None]},
+            id="null_literal",
+        ),
+        pytest.param(
+            {"bo": [True, None], "b": ["x", "y"]},
+            ["bo", "b"],
+            {"out": ["true x", None]},
```
jonasdedden, line 94: Just FYI: This is documenting current behavior, but I don't know whether we *want* to assert this. Dunno whether there is some value in a string concatenation of bools (that are lowercased) and strings.

src/narwhals/_pandas_like/namespace.py
```diff
@@ -370,10 +370,17 @@ def concat_str(
...
         string = self._version.dtypes.String()
+        boolean = self._version.dtypes.Boolean()
 
         def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
             expr_results = [s for _expr in exprs for s in _expr(df)]
-            series = [s.cast(string) for s in expr_results]
+            # Polars renders booleans lowercase, pandas `astype(str)` does not.
+            series = [
+                s.cast(string).str.to_lowercase()
+                if s.dtype == boolean
```
FBruzzesi:
```suggestion
                if s.dtype.is_boolean()
```

_and_ remove line 373

FBruzzesi:
FYI spotted a couple of upstream issues while reviewing this one. I will work on them.

* [x] #3939 
* [x] #3941

FBruzzesi:
Hey @jonasdedden we should have unblocked some of the issues found in `concat_str` for pandas, which however led to a conflict with your changes.

If you fancy it, Dask will have the very same problem, and it would be nice if you could fix that as well 🙏🏼

jonasdedden:
@FBruzzesi I'm closing this PR since by now the scope creep was a bit too much. Please find much smaller PRs that are a result of this one here:

- #3978
- #3979 (for this one it's unclear whether we want to have str concatenation of booleans or not)
- #3980
- #3981

PR #3914, opened 2026-09-06
test: cover to_time custom formats, schema and invalid raises

jonasdedden:
# Description

Non-default formats (`"12-34"` with `"%H-%M"`, `"01:30 PM"` with `"%I:%M %p"`), including a lazy `collect_schema() == Time` assertion (the plugin had no `Time` mapping, which the old tests never caught), plus unparseable inputs raising a generic `Exception` (backends without `Time` support also raise, so no xfail is needed there).

Follow-up adds the series variants of the custom-format cases with nulls preserved.

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

None. Test-only follow-up from narwhals-daft differential testing.

## AI assistance

- [ ] No AI tools were used for this PR.
- [x] AI tools were used.

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes (N/A, test-only, no behavior change)
- [ ] If this is your first PR to narwhals, attach a screenshot of `pytest` passing locally (not CI):

    ```bash
    PYTEST_ADDOPTS="--numprocesses=logical" \
    make run-ci DEPS="--extra pandas --extra dask --group core-tests --group sklearn --group plugins" \
    CMD="pytest tests --cov=src --cov=tests --runslow --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe"
    ```

Note: the full-suite command above was not run. Targeted run on the CI constructor set instead: `pytest tests/expr_and_series/str/to_time_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` gives 66 passed, 5 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

PR #3919, opened 2026-09-06
test: test cover quantile boundaries, invalid raises, median over

jonasdedden:
# Description

`quantile(0.0/1.0)` as min/max, out-of-bounds quantile raising a generic `Exception` (per-backend error types differ; lazy and eager paths handled separately), and `median().over(...)` mirroring the existing quantile-over coverage (xfailed for `pyarrow_table`, which returns `1.0` instead of `1.5`). The boundaries test asserts dask's multi-partition `NotImplementedError`, like the existing quantile test does.

Now covered, pinned to polars: all-null and empty median and `any_value` yield null (via float casts, since pandas cannot cast all-`None` object columns to int), `NaN` sorts last in `quantile` (with `xfail` where backends drop it, excluding sqlframe which transpiles fine), and infinities order normally (with `xfail` where plain or nullable pandas return `NaN` or `None`). Boolean median stays unsupported, per maintainer decision: the coercion experiment is dropped entirely.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added

    ```bash
    ```

Note: the full-suite command above was not run. Targeted run on the CI constructor set instead: `pytest tests/expr_and_series/quantile_test.py tests/expr_and_series/median_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` gives 142 passed, 15 xfailed. Also green with `--constructors=ibis` and `--constructors=dask`.

### DIFF
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -35,6 +35,7 @@ from narwhals._utils import (
     not_implemented,
 )
 from narwhals.dependencies import is_numpy_array_1d
+from narwhals.dtypes import Boolean, Float64
 from narwhals.exceptions import InvalidOperationError, ShapeError
 
 if TYPE_CHECKING:
@@ -343,6 +344,8 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
         return maybe_extract_py_scalar(pc.mean(self.native), _return_py_scalar)
 
     def median(self, *, _return_py_scalar: bool = True) -> float:
+        if self.dtype == Boolean:
+            return self.cast(Float64()).median(_return_py_scalar=_return_py_scalar)
         if not self.dtype.is_numeric():
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -35,7 +35,6 @@ from narwhals._utils import (
     not_implemented,
 )
 from narwhals.dependencies import is_numpy_array_1d
-from narwhals.dtypes import Boolean, Float64
 from narwhals.exceptions import InvalidOperationError, ShapeError
 
 if TYPE_CHECKING:
@@ -344,14 +343,16 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
         return maybe_extract_py_scalar(pc.mean(self.native), _return_py_scalar)
 
     def median(self, *, _return_py_scalar: bool = True) -> float:
-        if self.dtype == Boolean:
-            return self.cast(Float64()).median(_return_py_scalar=_return_py_scalar)
-        if not self.dtype.is_numeric():
+        if (dtype := self.dtype).is_boolean():
+            native = self.cast(self._version.dtypes.Float64()).native
+        elif dtype.is_numeric():
+            native = self.native
+        else:
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
 
         return maybe_extract_py_scalar(
-            pc.quantile(self.native, q=0.5, interpolation="linear")[0], _return_py_scalar
+            pc.quantile(native, q=0.5, interpolation="linear")[0], _return_py_scalar
         )
 
     def min(self, *, _return_py_scalar: bool = True) -> Any:
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -343,16 +343,12 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
         return maybe_extract_py_scalar(pc.mean(self.native), _return_py_scalar)
 
     def median(self, *, _return_py_scalar: bool = True) -> float:
-        if (dtype := self.dtype).is_boolean():
-            native = self.cast(self._version.dtypes.Float64()).native
-        elif dtype.is_numeric():
-            native = self.native
-        else:
+        if not self.dtype.is_numeric():
             msg = "`median` operation not supported for non-numeric input type."
             raise InvalidOperationError(msg)
 
         return maybe_extract_py_scalar(
-            pc.quantile(native, q=0.5, interpolation="linear")[0], _return_py_scalar
+            pc.quantile(self.native, q=0.5, interpolation="linear")[0], _return_py_scalar
         )
 
     def min(self, *, _return_py_scalar: bool = True) -> Any:
--- a/src/narwhals/_dask/expr.py
+++ b/src/narwhals/_dask/expr.py
@@ -23,6 +23,7 @@ from narwhals._utils import (
     generate_temporary_column_name,
     not_implemented,
 )
+from narwhals.dtypes import Boolean
 from narwhals.exceptions import InvalidOperationError
 
 if TYPE_CHECKING:
@@ -289,7 +290,9 @@ class DaskExpr(
 
         def func(s: dx.Series) -> dx.Series:
             dtype = native_to_narwhals_dtype(s.dtype, self._version, Implementation.DASK)
-            if not dtype.is_numeric():
+            if dtype == Boolean:
+                s = s.astype("float64")
+            elif not dtype.is_numeric():
                 msg = "`median` operation not supported for non-numeric input type."
                 raise InvalidOperationError(msg)
             return s.median_approximate().to_series()
--- a/src/narwhals/_dask/expr.py
+++ b/src/narwhals/_dask/expr.py
@@ -289,9 +289,7 @@ class DaskExpr(
 
         def func(s: dx.Series) -> dx.Series:
             dtype = native_to_narwhals_dtype(s.dtype, self._version, Implementation.DASK)
-            if dtype.is_boolean():
-                s = s.astype("float64")
-            elif not dtype.is_numeric():
+            if not dtype.is_numeric():
                 msg = "`median` operation not supported for non-numeric input type."
                 raise InvalidOperationError(msg)
             return s.median_approximate().to_series()
--- a/src/narwhals/_dask/expr.py
+++ b/src/narwhals/_dask/expr.py
@@ -23,7 +23,6 @@ from narwhals._utils import (
     generate_temporary_column_name,
     not_implemented,
 )
-from narwhals.dtypes import Boolean
 from narwhals.exceptions import InvalidOperationError
 
 if TYPE_CHECKING:
@@ -290,7 +289,7 @@ class DaskExpr(
 
         def func(s: dx.Series) -> dx.Series:
             dtype = native_to_narwhals_dtype(s.dtype, self._version, Implementation.DASK)
-            if dtype == Boolean:
+            if dtype.is_boolean():
                 s = s.astype("float64")
             elif not dtype.is_numeric():
                 msg = "`median` operation not supported for non-numeric input type."
--- a/src/narwhals/_pandas_like/series.py
+++ b/src/narwhals/_pandas_like/series.py
@@ -540,9 +540,7 @@ class PandasLikeSeries(EagerSeries[Any]):
         return self.native.mean()
 
     def median(self) -> float:
-        if (dtype := self.dtype).is_boolean():
-            return self.cast(self._version.dtypes.Float64()).native.median()
-        if not dtype.is_numeric():
+        if not self.dtype.is_numeric():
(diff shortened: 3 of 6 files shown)


src/narwhals/_arrow/series.py
```diff
@@ -343,6 +344,8 @@ def mean(self, *, _return_py_scalar: bool = True) -> float:
         return maybe_extract_py_scalar(pc.mean(self.native), _return_py_scalar)
 
     def median(self, *, _return_py_scalar: bool = True) -> float:
+        if self.dtype == Boolean:
+            return self.cast(Float64()).median(_return_py_scalar=_return_py_scalar)
```
FBruzzesi: Using dtypes like this is not _really_ safe since they depend on the version (main, v1, v2 namespaces).
What you should do is accessing the dtypes from the version. Also I would avoid the recursion for the extra check. What about:


```py
def median(self, *, _return_py_scalar: bool = True) -> float:
    if (dtype:=self.dtype).is_boolean():
        native = self.cast(self._version.dtypes.Float64()).native
    elif dtype.is_numeric():
        native = self.native
    else:
        msg = "`median` operation not supported for non-numeric input type."

    return maybe_extract_py_scalar(
            pc.quantile(native, q=0.5, interpolation="linear")[0], _return_py_scalar
        )
```

src/narwhals/_dask/expr.py
```diff
@@ -289,7 +290,9 @@ def median(self) -> Self:
 
         def func(s: dx.Series) -> dx.Series:
             dtype = native_to_narwhals_dtype(s.dtype, self._version, Implementation.DASK)
-            if not dtype.is_numeric():
+            if dtype == Boolean:
+                s = s.astype("float64")
+            elif not dtype.is_numeric():
```
FBruzzesi: Similarly as above, prefer using `dtype.is_boolean()`

src/narwhals/_pandas_like/series.py
```diff
@@ -540,6 +540,8 @@ def mean(self) -> float:
         return self.native.mean()
 
     def median(self) -> float:
+        if self.dtype == Boolean:
+            return self.cast(Float64()).median()
```
FBruzzesi: As for pyarrow, use `dtype.is_boolean()` and avoid recurring

tests/expr_and_series/median_test.py
```diff
@@ -57,6 +57,46 @@ def test_median_group_by(
...
+
+
+@pytest.mark.parametrize(
+    "data", [{"a": [None, None]}, {"a": []}], ids=["all_null", "empty"]
+)
+@pytest.mark.filterwarnings("ignore::RuntimeWarning")
+def test_median_null_or_empty(
+    constructor: Constructor, data: dict[str, list[int]]
+) -> None:
+    # Degenerate-slice `RuntimeWarning`s (old pandas on all-null input) are expected.
```
FBruzzesi: Can you move the comment inline and capture the warning with more precise capture if possible - I can't manage to do it all the times, but you should be able to add the library it comes from at the end. There are examples in pyproject.toml

tests/expr_and_series/quantile_test.py
```diff
@@ -51,6 +57,60 @@ def test_quantile_expr(
...
+def test_quantile_out_of_bounds_raises(constructor: Constructor) -> None:
+    df = nw.from_native(constructor({"a": [1, 2, 3]}))
+    expr = nw.col("a").quantile(1.5, "linear")
+    # Broad `Exception`: error types differ per backend.
+    with pytest.raises(Exception):  # noqa: B017, PT011
+        maybe_collect(df.select(expr))
+
+
+def test_quantile_nan(constructor: Constructor, request: pytest.FixtureRequest) -> None:
+
```
FBruzzesi:
```suggestion
```

tests/expr_and_series/quantile_test.py
```diff
@@ -51,6 +57,60 @@ def test_quantile_expr(
...
+        "duckdb" in str(constructor) or "sqlframe" in str(constructor)
+    ) and DUCKDB_VERSION < (1, 3, 1):
+        request.applymarker(pytest.mark.xfail(reason="old duckdb NaN handling"))
+    df = nw.from_native(constructor({"a": [1.0, float("nan"), 2.0]}))
+    result = df.select(nw.col("a").quantile(0.5, "linear").alias("q"))
+    assert_equal_data(result, {"q": [2.0]})
+
+
+def test_quantile_inf(constructor: Constructor, request: pytest.FixtureRequest) -> None:
+
```
FBruzzesi:
```suggestion
```

tests/expr_and_series/median_test.py
```diff
@@ -57,6 +57,46 @@ def test_median_group_by(
     assert_equal_data(result, {"g": [1, 2], "a": [1.5, 15.0]})
 
 
+def test_median_over(constructor: Constructor, request: pytest.FixtureRequest) -> None:
+    # `median` composes with `over`, like `quantile` already does.
```
FBruzzesi: move this comment into `skip(reason=...)`

FBruzzesi (commented):
Thanks @jonasdedden 🙏🏼 left some suggestions. The dtypes module is not the easiest to reason about the first time you see it due to the various namespace 😅

tests/expr_and_series/median_test.py
jonasdedden: This wasn't a comment why that DuckDB test has to be skipped, but just a silly redudant docstring. The DuckDB skip now has a proper reason though.

tests/expr_and_series/quantile_test.py
jonasdedden: Done

tests/expr_and_series/quantile_test.py
jonasdedden: Done

tests/expr_and_series/median_test.py
jonasdedden: Added a `@pytest.mark.filterwarnings("ignore:Mean of empty slice:RuntimeWarning:numpy")`

src/narwhals/_arrow/series.py
jonasdedden: Implemented as suggested! Thanks!

src/narwhals/_dask/expr.py
jonasdedden: Done

src/narwhals/_pandas_like/series.py
jonasdedden: Done

tests/expr_and_series/median_test.py
FBruzzesi: whops my bad!

FBruzzesi (approved):
Thanks for adjusting @jonasdedden 🚀

Approving but I have a couple of remarks:

1. We cannot merge until the PR stays in draft/not ready for review. Edit: At the same time, we updated some settings to avoid more that 2 PRs ready for review from the same author
2. We are casting to Boolean to Float64, yet it's safe to convert to any less memory hungry numeric dtype, wouldn't it? Not sure we have tests for UInt8/Int8, but it might be worth checking

jonasdedden:
> We cannot merge until the PR stays in draft/not ready for review

I'm unfortunately still getting "Author has reached the open pull request limit". This means I only could make this PR ready for review if I undo this for another PR. 

If you have a candidate PR for me which I should convert to Draft, because there still are issues with it, please let me know. Dunno who could overwrite the setting that I can have more PRs ready for review.

FBruzzesi:
> If you have a candidate PR for me which I should convert to Draft

I would suggest to do so for #3913 and #3914 as I would like to wait for polars team to reply to the issue I opened.

> Dunno who could overwrite the setting that I can have more PRs ready for review.

We just recently added it: to be totally honest opening 11 PRs in 20 minutes was something we were not expecting, and that could set a precedent that we don't really want to have. I know you are coming from a good place and with interest in the project as well as targeted contributions, however thanks to the current tools if _anyone_ can do it, that can be problematic for maintainers. We set the limit to 3 "ready for review" PRs per authors without writing access

jonasdedden:
> to be totally honest opening 11 PRs in 20 minutes was something we were not expecting

Yeah, this didn't really "organically" happen of course :sweat_smile: My mode of working nowadays is: Try out a new thing/feature with a library, probably stumble upon a dozen different issues, collect everything in detail and review it, and then as a last step, open all contributions in one big swoop.

11 PRs is a bit much, yes, which also was the reason why I opened a majority of the PRs in Draft mode from the get-go :sweat_smile: 

This PR is ready now :pray:

FBruzzesi:
All good, no worries! Did you have the chance to try:

> We are casting to Boolean to Float64, yet it's safe to convert to any less memory hungry numeric dtype, wouldn't it? Not sure we have tests for UInt8/Int8, but it might be worth checking

?

jonasdedden:
I think Float64 makes at least some sense, since `median()` outputs Float64/Double for numeric columns across all backends. For pandas and polars, even for boolean columns.

```Python
import duckdb
import pandas as pd
import polars as pl
import pyarrow as pa
import pyarrow.compute as pc

print("--- median over Int8 column: output dtype ---")
print("pandas:          ", pd.DataFrame({"a": [1, 0, 1]}, dtype="Int8").median().dtype)
print("pandas[pyarrow]: ", pd.DataFrame({"a": [1, 0, 1]}, dtype="int8[pyarrow]").median().dtype)
print("polars:          ", pl.DataFrame({"a": [1, 0, 1]}, schema={"a": pl.Int8}).median().schema)
s = pc.quantile(pa.array([1, 0, 1], type=pa.int8()), q=0.5)[0]
print("pyarrow:         ", s.type)
print("duckdb:          ", duckdb.sql("describe select median(a) from (values (1::TINYINT), (0), (1)) t(a)").fetchall()[0][1])
print("--- median over Boolean column: output dtype ---")
print("pandas:          ", pd.DataFrame({"a": [True, False, True]}).median().dtype)
try:
    print("pandas[pyarrow]: ", pd.DataFrame({"a": [True, False, True]}, dtype="boolean[pyarrow]").median().dtype)
except Exception as e:
    print("pandas[pyarrow]:  RAISED", type(e).__name__)
print("polars:          ", pl.DataFrame({"a": [True, False, True]}).median().schema)
try:
    print("pyarrow:         ", pc.quantile(pa.array([True, False, True]), q=0.5)[0].type)
except Exception as e:
    print("pyarrow:          RAISED", type(e).__name__)
print("duckdb:          ", duckdb.sql("describe select median(a) from (values (true), (false), (true)) t(a)").fetchall()[0][1])
```
=>
```
--- median over Int8 column: output dtype ---
pandas:           Float64
pandas[pyarrow]:  double[pyarrow]
polars:           Schema({'a': Float64})
pyarrow:          double
duckdb:           DOUBLE
--- median over Boolean column: output dtype ---
pandas:           float64
pandas[pyarrow]:  RAISED TypeError
polars:           Schema({'a': Float64})
pyarrow:          RAISED ArrowNotImplementedError
duckdb:           BOOLEAN
```

MarcoGorelli:
thanks for the pr

>  enable median for boolean series

why do we need this? can't users cast by themselves?

jonasdedden:
> > enable median for boolean series
> 
> why do we need this? can't users cast by themselves?

I'm totally opinion-less about this. That pandas and polars support this natively whilst other backends could be brought to supporting it through a cast, was just something that came out when differential testing for `narwhals-daft`. Totally happy to drop this entirely again if it doesn't seem useful.

MarcoGorelli:
sure, i'd suggest dropping it then, i'd prefer to keep narwhals kind of minimal

jonasdedden:
Dropped the median of a boolean column feature in https://github.com/narwhals-dev/narwhals/pull/3919/commits/a9369704e51d2a6039699febf45edf9af10c283a. This PR reverted back to being test-only.

FBruzzesi:
Thanks for chiming in @MarcoGorelli

> why do we need this? can't users cast by themselves?

The boundary between "aligning all backend behavior with polars" and "let each backend do whatever they do natively" is not always clear to me.

That said, I also don't have strong opinions in this case, so merging the tests 🙏🏼

PR #3920, opened 2026-09-06
fix(pyarrow): `shift` preserves length when `n` exceeds the series length

winklemad:
# Description

`ArrowSeries.shift(n)` padded the result with `n` nulls without capping `n` at the series length, so shifting by more than the length grew the series instead of returning an all-null series of the same length. For example, `shift(3)` on a two-row pyarrow series returned three rows.

The same series is used by the rolling functions — `rolling_sum`, `rolling_mean`, `rolling_var` and `rolling_std` all shift by `window_size` internally — so a window larger than the series desynced the chunk lengths and raised `ArrowInvalid: Array arguments must all be the same length` instead of returning nulls:

```python
import narwhals as nw
import pyarrow as pa

s = nw.from_native(pa.chunked_array([[1]]), series_only=True)

s.rolling_sum(window_size=2)   # main: ArrowInvalid  ->  this PR: [None]
s.shift(3).to_list()           # main: [None, None, None]  ->  this PR: [None]
```

polars and pandas both return an all-null series of the original length here, so this brings the pyarrow backend in line with them.

The fix caps the null padding at the series length (`min(n, length)`), which keeps the existing behaviour for every in-range shift untouched. Added a regression test for `shift` beyond the length across the eager backends, and one for `rolling_sum` with a window longer than the series.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Found while testing the pyarrow backend; no existing issue.

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [ ] Documented the changes (no public API change; behaviour now matches polars/pandas)

### DIFF
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -374,10 +374,15 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
         return self._with_native(self.native.drop_null())
 
     def shift(self, n: int) -> Self:
+        # Cap the number of null padding values at the length of the series:
+        # shifting by more than the length must still return a series of the
+        # same length (all null), and must not desync the array lengths that
+        # rolling windows rely on.
+        length = len(self.native)
         if n > 0:
-            arrays = [nulls_like(n, self), *self.native[:-n].chunks]
+            arrays = [nulls_like(min(n, length), self), *self.native[:-n].chunks]
         elif n < 0:
-            arrays = [*self.native[-n:].chunks, nulls_like(-n, self)]
+            arrays = [*self.native[-n:].chunks, nulls_like(min(-n, length), self)]
         else:
             return self._with_native(self.native)
         return self._with_native(pa.concat_arrays(arrays))
--- a/src/narwhals/_arrow/series.py
+++ b/src/narwhals/_arrow/series.py
@@ -374,17 +374,22 @@ class ArrowSeries(EagerSeries["ChunkedArrayAny"]):
         return self._with_native(self.native.drop_null())
 
     def shift(self, n: int) -> Self:
+        if n == 0:
+            return self._with_native(self.native)
+        length = len(self.native)
         # Cap the number of null padding values at the length of the series:
-        # shifting by more than the length must still return a series of the
+        # shifting by at least the length must still return a series of the
         # same length (all null), and must not desync the array lengths that
         # rolling windows rely on.
-        length = len(self.native)
+        null_array = nulls_like(min(abs(n), length), self)
+        if abs(n) >= length:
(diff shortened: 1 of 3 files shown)


src/narwhals/_arrow/series.py
```diff
@@ -374,10 +374,15 @@ def drop_nulls(self) -> Self:
...
+        # shifting by more than the length must still return a series of the
+        # same length (all null), and must not desync the array lengths that
+        # rolling windows rely on.
+        length = len(self.native)
         if n > 0:
-            arrays = [nulls_like(n, self), *self.native[:-n].chunks]
+            arrays = [nulls_like(min(n, length), self), *self.native[:-n].chunks]
         elif n < 0:
-            arrays = [*self.native[-n:].chunks, nulls_like(-n, self)]
+            arrays = [*self.native[-n:].chunks, nulls_like(min(-n, length), self)]
```
FBruzzesi: Thanks! Could you check if the array slicing is a free op when `abs(n)>length`? If that's not the case, I would suggest to have a fast path; something along the following lines should suffice:

```py
null_size = abs(n)
null_array = nulls_like(null_size, self)
if null_size >= length:
    return self._with_native(null_array)
if n > 0:
    arrays = [null_array, *self.native[:-n].chunks]
elif n < 0:
    arrays = [*self.native[-n:].chunks, null_array]
else:
    return self._with_native(self.native)
return self._with_native(pa.concat_arrays(arrays))
```

FBruzzesi (commented):
Thanks @winklemad for spotting the issue and coming up with a fix! Left a question and possible suggestion 🙏🏼

src/narwhals/_arrow/series.py
winklemad: Good idea — added the fast path. One small tweak: returning `nulls_like(abs(n))` directly would give the wrong length when `abs(n) > length` (e.g. `shift(3)` on a 2-row series would come back with 3 rows), so I cap the null size at the length — `nulls_like(min(abs(n), length))` — and return it directly when `abs(n) >= length`.

On the slicing question: with the fast path that case never reaches the slice/concat, so it is a no-op now. For `abs(n) < length` the pyarrow slice is a cheap zero-copy view, so the remaining path stays as-is.

FBruzzesi: Awesome, thanks for bearing with me (and spotting the wrong suggestion 🙈)

FBruzzesi (approved):
Thanks again @winklemad

PR #3924, opened 2026-09-07
fix: use `dtype.categories` for `pl.Enum` in `get_categories`

### LINKED ISSUE #3923: Preserve category order in `cat.get_categories` for `pl.Enum`
I think that for `pl.Enum` at least, we should preserve the category order

MarcoGorelli:
closes #3923 

this also fixes the shiny downstream ci failure

# Description

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #\<issue number\>

## AI assistance

- [ ] AI tools were used.

## Checklist

- [ ] Tests added
- [ ] Documented the changes

    ```bash
    ```

### DIFF
--- a/src/narwhals/_polars/utils.py
+++ b/src/narwhals/_polars/utils.py
@@ -369,14 +369,10 @@ def native_get_categories(native: pl.Series) -> pl.Series:
     # See https://github.com/narwhals-dev/narwhals/issues/3895.
     #
     # For `Enum`, the declared categories are already unique, ordered, and
-    # null-free, so when every declared category is actually present in the
-    # data we can return `dtype.categories` directly instead of paying for
-    # `unique().drop_nulls().cast(String)`.
+    # null-free, so we can return `dtype.categories` directly.
     dtype = native.dtype
     if isinstance(dtype, pl.Enum):
-        categories = dtype.categories
-        if native.drop_nulls().n_unique() == len(categories):
-            return categories
+        return dtype.categories
     return native.unique(maintain_order=True).drop_nulls().cast(pl.String)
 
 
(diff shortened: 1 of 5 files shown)


src/narwhals/_polars/utils.py
```diff
@@ -363,6 +363,23 @@ def zfill(self, width: int) -> CompliantT: ...
...
+    #
+    # For `Enum`, the declared categories are already unique, ordered, and
+    # null-free, so when every declared category is actually present in the
+    # data we can return `dtype.categories` directly instead of paying for
+    # `unique().drop_nulls().cast(String)`.
+    dtype = native.dtype
+    if isinstance(dtype, pl.Enum):
+        categories = dtype.categories
+        if native.drop_nulls().n_unique() == len(categories):
+            return categories
```
FBruzzesi: Ahhh that's a good catch! Polars deprecation warning suggestion is to return `dtype.categories`, should we do that regardless all categories being present?

FBruzzesi (commented):
Thanks @MarcoGorelli that's a good catch! Left a comment just to double make sure we are returning the correct items. Go ahead if that's the case ;)

src/narwhals/_polars/utils.py
MarcoGorelli: yeah maybe

there's probably no 100% correct solution here 😄 😿 so let's bias in favour of simplicity / polars' own message

PR #3929, opened 2026-09-08
fix: align `__mod__` and `__rmod__` behavior with negative operands across backends

ptimizeroracle:
# Description

`%` on the SQL backends (duckdb, sqlframe, pyspark) returned C-style remainders
for negative operands, while narwhals follows Python and Polars, where the
result takes the sign of the divisor.

```python
import narwhals as nw
df = nw.from_native(duckdb_lazy({"a": [-7, 7, -7]}))
df.select(nw.col("a") % 3)
# main:    {"a": [-1, 1, -1]}
# this PR: {"a": [2, 1, 2]}
```

The existing `test_mod` generates negative operands with hypothesis, but it
only runs on `constructor_eager`, so the SQL-lazy backends never saw it. The
new `test_mod_negative_operands` / `test_rmod_negative_operands` use the plain
`constructor` fixture with deterministic negative cases for both directions.

the identity lives in one helper, `floor_mod` under `_utils.py`, typed with a
`SupportsFloorMod` protocol so it works for any native expression with
`__mod__` and `__add__`. `DuckDBExpr` and `SparkLikeExpr` pass it straight
into `_with_binary`, `IbisExpr` wraps it in a closure for the `NumericValue`
casts. polars versions below 0.20.8 also took the dividend's sign, so
`PolarsExpr` and `PolarsSeries` route through `floor_mod` when
`BACKEND_VERSION < (0, 20, 8)` and use the native `__mod__` otherwise

the old `_sql` modulo methods are removed, all subclasses now override
`__mod__`/`__rmod__` directly. eager polars below 0.20.8 also gets the
corrected modulo behavior, other eager backends are unchanged

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- none filed, the divergence was found by differential probing of the backends

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes

    ```bash
    ```

### DIFF
--- a/src/narwhals/_duckdb/expr.py
+++ b/src/narwhals/_duckdb/expr.py
@@ -49,6 +49,22 @@ if TYPE_CHECKING:
 class DuckDBExpr(SQLExpr["DuckDBLazyFrame", "Expression"]):
     _implementation = Implementation.DUCKDB
 
+    def __mod__(self, other: Self) -> Self:
+        def _mod(expr: Expression, other: Expression) -> Expression:
+            # DuckDB's `%` follows C sign semantics (sign of the dividend),
+            # while narwhals follows Python and Polars (sign of the divisor).
+            # The double-modulus identity `((a % b) + b) % b` restores
+            # floor-modulus.
+            return ((expr % other) + other) % other
+
+        return self._with_binary(_mod, other)
+
+    def __rmod__(self, other: Self) -> Self:
+        def _rmod(expr: Expression, other: Expression) -> Expression:
+            return ((other % expr) + expr) % expr
+
+        return self._with_binary(_rmod, other).alias("literal")
+
     def __init__(
         self,
         call: EvalSeries[DuckDBLazyFrame, Expression],
--- a/src/narwhals/_duckdb/expr.py
+++ b/src/narwhals/_duckdb/expr.py
@@ -21,7 +21,7 @@ from narwhals._duckdb.utils import (
     window_expression,
 )
 from narwhals._sql.expr import SQLExpr
-from narwhals._utils import NO_DEFAULT, Implementation, Version, extend_bool
+from narwhals._utils import NO_DEFAULT, Implementation, Version, extend_bool, floor_mod
 
 if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
@@ -52,16 +52,14 @@ class DuckDBExpr(SQLExpr["DuckDBLazyFrame", "Expression"]):
     def __mod__(self, other: Self) -> Self:
         def _mod(expr: Expression, other: Expression) -> Expression:
             # DuckDB's `%` follows C sign semantics (sign of the dividend),
-            # while narwhals follows Python and Polars (sign of the divisor).
-            # The double-modulus identity `((a % b) + b) % b` restores
-            # floor-modulus.
-            return ((expr % other) + other) % other
+            # `floor_mod` restores Python's (sign of the divisor).
+            return floor_mod(expr, other)
 
         return self._with_binary(_mod, other)
 
     def __rmod__(self, other: Self) -> Self:
         def _rmod(expr: Expression, other: Expression) -> Expression:
-            return ((other % expr) + expr) % expr
+            return floor_mod(other, expr)
 
         return self._with_binary(_rmod, other).alias("literal")
 
--- a/src/narwhals/_duckdb/expr.py
+++ b/src/narwhals/_duckdb/expr.py
@@ -50,12 +50,7 @@ class DuckDBExpr(SQLExpr["DuckDBLazyFrame", "Expression"]):
     _implementation = Implementation.DUCKDB
 
     def __mod__(self, other: Self) -> Self:
-        def _mod(expr: Expression, other: Expression) -> Expression:
-            # DuckDB's `%` follows C sign semantics (sign of the dividend),
-            # `floor_mod` restores Python's (sign of the divisor).
-            return floor_mod(expr, other)
-
-        return self._with_binary(_mod, other)
+        return self._with_binary(floor_mod, other)
 
     def __rmod__(self, other: Self) -> Self:
         def _rmod(expr: Expression, other: Expression) -> Expression:
--- a/src/narwhals/_polars/expr.py
+++ b/src/narwhals/_polars/expr.py
@@ -311,18 +311,20 @@ class PolarsExpr:
         return self._with_native(result)
 
     def __mod__(self, other: Any) -> Self:
-        if self._backend_version < (0, 20, 8):
-            # Polars 0.20.7 and earlier follows C sign semantics for `%`
-            # (sign of the dividend), restore Python's floor-modulus.
-            return self._with_native(floor_mod(self.native, extract_native(other)))
-        return self._with_native(self.native.__mod__(extract_native(other)))
+        native = (
+            floor_mod(self.native, extract_native(other))
+            if BACKEND_VERSION < (0, 20, 8)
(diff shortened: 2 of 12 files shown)


MarcoGorelli:
thanks for your PR!

looks like there's some ci failures

ptimizeroracle:
Thanks! All three addressed in f1eec34:

- typing: the emulation now lives in `DuckDBExpr` and `SparkLikeExpr` with concrete native expression types, instead of chaining dunders on the `NativeExprT` protocol (mirrors how `__floordiv__` is already overridden in `_spark_like`)
- old versions: polars `%` followed C sign semantics before 0.20.8 (verified: 0.20.4 gives [-1, 1, -1], 0.20.8 gives [2, 1, 2]), so the new tests now skip `POLARS_VERSION < (0, 20, 8)`, matching the minimum and pretty-old pins of 0.20.4
- full checklist command locally: 14932 passed, coverage 99.89%

ptimizeroracle:
Follow-up in f463d2b for the two remaining jobs:

- ibis constructor: `IbisExpr` was inheriting the generic `%` passthrough, so it had the same trunc semantics. It now has the same `((a % b) + b) % b` override (verified locally with the ibis constructor, 88 passed)
- full-coverage: the `_sql` `__mod__`/`__rmod__` bodies became dead code once duckdb, spark_like and ibis all override them, so they are removed; the base `CompliantExpr` string-dispatch versions still cover the eager backends

ptimizeroracle:
@MarcoGorelli the ci failures are fixed, all 44 checks are green now

also corrected the description, the modulo overrides live in the three backend classes and the old shared methods are removed

could you take another look when you have a moment

FBruzzesi (changes requested):
Thanks for your contribution @ptimizeroracle 

I have a couple of asks before moving forward with this PR:

1. Since the `((a % b) + b) % b` formula is used in multiple places, can we factor it out in a helper function under `_utils.py` along the lines of:

    ```py
    def adjusted_mod(dividend, divisor):
        """Adapt '%' ops that follow C sign semantics (sign of the dividend),
        to python and polars one (sign of the divisor)
        by using the double-modulus identity `((a % b) + b) % b`
        """
        return ((dividend % divisor) + divisor) % divisor
    ```
    * docstring needs improvement and, probably, the function name as well.
    * you need to define a protocol for dividend and divisor to make it type safe
2. I suggesting to define it in `_utils.py` so that you can use it also for polars old versions. You can see in polars `expr.py` and `series.py` how we deal with other cases for which old versions need patching
3. In the tests, you should move all the inline comments explaining why something is skipped/xfailed into the skip/xfail `reason` argument.

ptimizeroracle:
thanks for the review, all three points addressed in 8f11f20

1. the identity now lives in a `floor_mod` helper under `_utils.py`, typed with a `SupportsFloorMod` protocol (`__mod__` and `__add__` with a same-typed operand) so it works for duckdb `Expression`, pyspark `Column`, ibis `Value` and polars `Expr`/`Series` alike. kept the name `floor_mod` since the distinction being restored is `math.fmod`-style (c sign) vs python `%` (floor), happy to rename

2. `PolarsExpr` and `PolarsSeries` now use it for polars<0.20.8, following the same version-guard shape as the existing `__rfloordiv__` patch. the new tests now run on old polars instead of being skipped: verified on a polars 0.20.4 venv, `arithmetic_test.py` is 86 passed there, and 561 passed across pandas/pandas[nullable]/pandas[pyarrow]/pyarrow/polars/dask/duckdb/sqlframe on current deps, pyright and ruff clean

3. skip/xfail explanations moved into `reason=` arguments, and the polars<0.20.8 skips are gone entirely since the behavior is now patched

src/narwhals/_duckdb/expr.py
```diff
@@ -49,6 +49,20 @@
 class DuckDBExpr(SQLExpr["DuckDBLazyFrame", "Expression"]):
     _implementation = Implementation.DUCKDB
 
+    def __mod__(self, other: Self) -> Self:
+        def _mod(expr: Expression, other: Expression) -> Expression:
+            # DuckDB's `%` follows C sign semantics (sign of the dividend),
+            # `floor_mod` restores Python's (sign of the divisor).
+            return floor_mod(expr, other)
+
+        return self._with_binary(_mod, other)
```
FBruzzesi: Couple of follow up suggestions:

1. You can trim all these inline comments everywhere / for all backend cases, they are already listed in `SupportsFloorMod`, and clear from calling `floor_mod`
2. I wonder if the following would work fine for the `__mod__` case:
	
	```py
	def __mod__(self, other: Self) -> Self:
        return self._with_binary(floor_mod, other)
	```

Both comment apply elsewhere

src/narwhals/_polars/expr.py
```diff
@@ -311,8 +311,19 @@ def __rfloordiv__(self, other: Any) -> Self:
         return self._with_native(result)
 
     def __mod__(self, other: Any) -> Self:
+        if self._backend_version < (0, 20, 8):
+            # Polars 0.20.7 and earlier follows C sign semantics for `%`
+            # (sign of the dividend), restore Python's floor-modulus.
+            return self._with_native(floor_mod(self.native, extract_native(other)))
         return self._with_native(self.native.__mod__(extract_native(other)))
```
FBruzzesi: We should update it elsewhere for sure, but at least for new cases let's do it directly: you can use `BACKEND_VERSION` already imported instead of accessing `self._backend_version`.

Also, I would be in favor of using ternary expressions:

```py
def __mod__(self, other: Any) -> Self:
    native = (
        floor_mod(self.native, extract_native(other))
        if BACKEND_VERSION < (0, 20, 8)
        else self.native.__mod__(extract_native(other))
    )
    return self._with_native(native)
```

---

The same comment applies for all other polars cases

FBruzzesi (commented):
Thanks @ptimizeroracle - I did another pass and left a couple of suggestions, after those, and fixing the CI jobs failures, we should be ready to merge!

Good job on the adjustments so far 🚀

ptimizeroracle:
all addressed in a614030

- comments trimmed everywhere, `SupportsFloorMod`'s docstring is now the only place explaining the semantics
- `floor_mod` is passed directly to `_with_binary` for `__mod__` on duckdb and spark_like. this needed its params renamed to `expr`/`other` since `_with_binary` calls the op with `other` as a keyword, ibis keeps a closure because of the `NumericValue` casts
- polars now uses `BACKEND_VERSION` with ternary expressions in both `PolarsExpr` and `PolarsSeries`
- ci: the prek failure was ruff format (long xfail-reason lines, now wrapped), the typing failure was mypy flagging the `floor_mod` result casts as redundant on ci's 3.12 env, casts dropped

local verification: full suite 14932 passed across pandas/pandas[nullable]/pandas[pyarrow]/pyarrow/polars/dask/duckdb/sqlframe, arithmetic tests 86 passed on a polars 0.20.4 venv, ruff check and format clean, mypy clean on the touched files

ptimizeroracle:
@FBruzzesi both points from your second pass are in a614030: `floor_mod` goes through `_with_binary` directly on duckdb and spark_like (ibis keeps a closure for the NumericValue casts), and polars uses the `BACKEND_VERSION` ternaries you sketched. the ci failures are fixed too, all 44 jobs green at head. ready for another look whenever you have a moment

FBruzzesi (approved):
Thanks a ton @ptimizeroracle - you have my +1 🙌🏼

PR #3939, opened 2026-09-11
fix: preserve nulls when casting to `String` on pandas pre v3.0.0

FBruzzesi:
# Description

Before pandas 3, `nw.String()` maps to builtin `str` and `astype(str)` renders nulls as the strings 'None' / 'nan' / 'NaT', so a null survived a cast as data. `PandasLikeSeries.cast` now restores the null mask when the result comes back as object, keying off the result dtype rather than a version check so modin and cudf are covered too.

`concat_str` needs a matching `fill_null("")`: with the nulls genuinely null, adding them raises `TypeError`.

Spotted while reviewing #3916 

@MarcoGorelli this would be a breaking change?!

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

### DIFF
--- a/src/narwhals/_pandas_like/series.py
+++ b/src/narwhals/_pandas_like/series.py
@@ -12,6 +12,7 @@ from narwhals._pandas_like.series_str import PandasLikeSeriesStringNamespace
 from narwhals._pandas_like.series_struct import PandasLikeSeriesStructNamespace
 from narwhals._pandas_like.utils import (
     NUMPY_VERSION,
+    PANDAS_VERSION,
     align_and_extract_native,
     binary_string_sum_fallback,
     broadcast_series_to_index,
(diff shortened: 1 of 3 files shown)


src/narwhals/_pandas_like/series.py
```diff
@@ -324,7 +325,17 @@ def cast(self, dtype: IntoDType) -> Self:
...
+        result = self.native.astype(pd_dtype)
+        if (
+            pd_dtype is str
+            and PANDAS_VERSION < (3,)
+            and (null_mask := self.native.isna()).any()
+        ):
+            # NOTE: Before pandas 3, `astype(str)` renders nulls as `'None'` / `'nan'`
+            # instead of keeping them null. Newer versions cast to a string dtype which
+            # preserves them, so there's nothing to restore.
+            result = result.where(~null_mask, None)
```
MarcoGorelli: can we also:
- check backend is pandas
- do `result.where(~null_mask, None, inplace=True)`, seeing as `result` is already a copy (due to `astype`)?

MarcoGorelli (approved):
thanks! just a minor comment

and happy to just treat this as a bug fix

src/narwhals/_pandas_like/series.py
FBruzzesi: Thanks @MarcoGorelli 

* Added a guard for pandas or modin (as `test_cast_string_keeps_nulls[modin[pyarrow]-date]` would fail otherwise). I swear I tried to run cudf for 15 mins on kaggle and only got installation issues. I asked claude and it pointed me to a line in [libcudf strings convert API](https://docs.rapids.ai/api/libcudf/stable/group__strings__convert), stating:

> Any null entries will result in corresponding null entries in the output column.

* Applied the suggestion

MarcoGorelli: thanks! all good then

PR #3918, opened 2026-09-06
fix: raise `DuplicateError` on duplicate `struct` field names; cover nested struct schema round trip in tests

jonasdedden:
# Description

Nested `struct(struct("a","b").alias("inner"),"c")` with a `collect_schema()` round trip (previously `List`/`Struct` mapped to `Unknown` on some backends), following the file's existing `UNSUPPORTED_BACKENDS`/`maybe_skip` pattern.

Now covered: duplicate field names are pinned to raising `DuplicateError`, with checks added to the pandas-like, pyarrow, DuckDB, ibis and Spark-like backends (which previously kept one field silently or errored late).

## What type of PR is this? (check all applicable)

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [X] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

None. Test-only follow-up from narwhals-daft differential testing.

## AI assistance

- [ ] No AI tools were used for this PR.
- [x] AI tools were used.

## Checklist

- [x] Code follows style guide (ruff)
- [x] Tests added
- [x] Documented the changes (N/A, test-only, no behavior change)
- [ ] If this is your first PR to narwhals, attach a screenshot of `pytest` passing locally (not CI):

    ```bash
    PYTEST_ADDOPTS="--numprocesses=logical" \
    make run-ci DEPS="--extra pandas --extra dask --group core-tests --group sklearn --group plugins" \
    CMD="pytest tests --cov=src --cov=tests --runslow --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],dask,duckdb,sqlframe"
    ```

Note: the full-suite command above was not run. Targeted run on the CI constructor set instead: `pytest tests/expr_and_series/struct_test.py --constructors=pandas,pandas[nullable],pandas[pyarrow],pyarrow,polars[eager],polars[lazy],duckdb,sqlframe` gives 103 passed. Also green with `--constructors=ibis` and `--constructors=dask`.

PR #3940, opened 2026-09-11
fix(tests): PySpark tests fixup

FBruzzesi:
# Description

Follow up from https://github.com/narwhals-dev/narwhals/pull/3918#issuecomment-5635260850


jonasdedden:
Thanks so much for working on this, and sorry for introducing the test failure! :see_no_evil:

FBruzzesi:
All good, I should have tested. And we should have a way to manually trigger the job

Issue #3932, opened 2026-09-09
[Enh]: Document and test chDB interoperability through Arrow


ShawnChen-Sirius:
## Context

[chDB DataStore](https://github.com/chdb-io/chdb) exposes the Arrow C Stream interface. Its results can already be passed to Narwhals with:

```python
nw.from_arrow(datastore, backend="pyarrow")
```

This allows Narwhals-based libraries to consume chDB results without converting through pandas.

## Scope of this issue

This proposal covers Arrow interoperability only.

- [ ] Add an interoperability test using `nw.from_arrow`
- [ ] Cover nullable values, unsigned integers, strings, nested types, empty results, and row order
- [ ] Add a short chDB example to the Arrow interoperability documentation
- [ ] Clearly document that the conversion executes and materializes the chDB result
- [ ] Check conversion performance on representative result sizes

After conversion, subsequent Narwhals operations run on the selected Arrow backend. They are not pushed down to chDB.

## Possible follow-up: native chDB backend

If there is community interest, we can explore a native Narwhals backend for chDB as a separate follow-up.

A native backend would keep Narwhals expressions lazy and translate operations such as filtering, projection, sorting, and aggregation into chDB queries. This would:

- avoid materializing unnecessary intermediate results
- reduce data transfer by reading only the required rows and columns
- keep query planning and execution inside chDB
- make ClickHouse execution available to libraries built on Narwhals
- improve performance for larger-than-memory and multi-step analytical workloads

We would appreciate feedback on the initial Arrow integration and whether native chDB execution would be useful to the Narwhals community.

PR #3944, opened 2026-09-13
docs: document chDB interoperability through Arrow

devtechedge:
chDB results implement the Arrow PyCapsule Interface, so they can be passed to `nw.from_arrow` directly.

This is now documentation only, following the review feedback: the chDB test module and the `chdb` test dependency are dropped, since chDB's own PyCapsule support is outside Narwhals' scope and reading objects that implement `__arrow_c_stream__` is already covered independently in CI.

The example sits in its own section under PyCapsule Interface, and notes that `chdb.query` executes and materializes the result, so subsequent operations run on the backend passed to `backend` and are not pushed down to chDB.

Towards #3932


FBruzzesi:
Thanks for your contribution @devtechedge 

I would be ok(ish) mentioning chDB as an example of the PyCapsule Interface interoperability in a dedicated section (i.e. under a `####` section level).

On the other hand, I am against adding it as a dependency and running it in the test suite: we are basically testing that:

* chDB supports PyCapsule Interface, which is outside narwhals scope to do
* Narwhals can read an object that implements arrow_c_stream, which we already test in our CI independently.

devtechedge:
Thanks for the review, @FBruzzesi, the scope call makes sense.

chDB's PyCapsule support is theirs to test, and Narwhals already covers `__arrow_c_stream__` ingestion on its own.

Dropped the dependency and `tests/from_arrow_chdb_test.py`, and moved the example into a dedicated `#### chDB` section under PyCapsule Interface.

Kept the note that `chdb.query` executes and materializes the result, because that is the part users tend to get wrong.

This is one file and 17 lines now.

PR #3955, opened 2026-09-17
fix: align str.slice negative offsets and float->int casts across backends

ptimizeroracle:
grouped parity fixes for `str.slice` and float->int casts, found with a differential matrix across backends

str.slice(offset, length):
- negative offset combined with an explicit length returned empty strings on pandas, pyarrow and dask: stop = offset + length lands before the end-relative start, e.g. slice(-3, 3) on "fdas" gave "" instead of "das". all three now slice from offset then head-slice the result, matching polars
- on the shared SQL backend, a negative offset on strings shorter than |offset| computed a non-positive substring start and DuckDB silently dropped a character: slice(-3) on "ab" returned "a". the start is now clamped to 1 via greatest
- a negative length was inconsistent everywhere: polars raised InvalidOperationError, the other backends silently returned empty or wrong strings. the stable Expr and Series apis now raise a ValueError saying length must be non-negative, before any backend is reached

float->int casts:
- pyarrow raised `ArrowInvalid` on any fractional part, both for the pyarrow constructor and pandas with dtype_backend="pyarrow". both now truncate toward zero (1.7 -> 1, -2.5 -> -2), matching pandas-numpy and polars. the pyarrow constructor passes safe=False for exactly the float->int case, the pandas path truncates through numpy before converting to the arrow dtype
- DuckDB, sqlframe and ibis round natively (1.7 -> 2) instead of truncating; the pyspark constructors truncate like pandas and polars. the new truncation test xfails for the rounding three, marked with reason. that divergence looks like it needs a semantics ruling rather than a silent override of engine casts, so i left it visible instead of papering over it

tests: new negative-offset cases including strings shorter than |offset|, the negative-length ValueError, and the float->int truncation case with the duckdb/sqlframe xfail. full suite 14927 passed across nine constructors locally, typing and lint clean on the touched files


FBruzzesi:
Hey @ptimizeroracle I didn't check the PR details, but from its title and description I understand that it has two different scopes, and they seem quite unrelated to me. Could you please split them into two separate PRs/repurpose this one for a single task, so that we can discuss in a more focused way?

ptimizeroracle:
the only failing check is test_corr_constant_column[pyspark[connect]], which is the pre-existing main failure from #3957 (added in bd1e4e0, unrelated to this diff). everything else is green

FBruzzesi:
Hey @ptimizeroracle a few sparse comments:

1. After the latest commits to main there are merge conflicts that need to be resolved.
2. I still think that this PR does too much in a single pass and it should be split into two, one for each scope.
3. Thanks for spotting the CI issue and reporting it. I have a PR open to address it.

ptimizeroracle:
split into #3969 (str.slice) and #3970 (float->int cast), both rebased onto current main and adopting validate_str_non_negative from the zfill work. thanks for the review and for the pointer on scope

Issue #3948, opened 2026-09-15
[Enh]: `read_*` and `scan_*` do not support file-like objects (buffers), unlike `write_*`


AdrianSosic:
### Please describe the purpose of the new feature or describe the problem to solve.

```python
df.write_parquet(io.BytesIO())  # works
df.write_csv(io.StringIO())     # works
```

However, the corresponding `read_*` and `scan_*` functions only accept file paths:

```python
nw.read_parquet(io.BytesIO(...), backend=...)  # TypeError
nw.read_csv(io.StringIO(...), backend=...)     # TypeError
nw.scan_parquet(io.BytesIO(...), backend=...)  # TypeError
nw.scan_csv(io.StringIO(...), backend=...)     # TypeError
```

This is asymmetric – the underlying backends all support buffer reading natively (pandas, polars, and pyarrow all accept `BytesIO`/`StringIO` for both reading and writing).

### Suggest a solution if possible.

`nw.read_*(buf, backend=...)` and `nw.scan_*(buf, backend=...)` should work wherever the corresponding `df.write_*(buf)` works.

MarcoGorelli:
thanks for the request

I'm open to this. IIRC duckdb doesn't support it, but whatever, we can support it where possible and open feature requests to places that don't (yet?) support it

AdrianSosic:
This would be great, since my current workaround is to write to a temporary file 😅

00200200:
I've started working on this.

FBruzzesi:
BTW, is this a duplicate of #3101 ? 👀
It's ok to keep it open and close both with #3956

00200200:
Yes — same request as #3101. #3956 should close both.

FBruzzesi:
Closed by #3956

Issue #3101, opened 2025-09-06
Allow `IO` objects in `{read,scan}_*` functions


dangotbanned:
An idea for solving in  
- https://github.com/narwhals-dev/narwhals/issues/3064#issuecomment-3262222775

[`pl.read_parquet`](https://docs.pola.rs/api/python/stable/reference/api/polars.read_parquet.html) and [`pl.scan_parquet`](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_parquet.html) seems to be the most permissive on `source`

https://github.com/pola-rs/polars/blob/b8bfb07a4a37a8d449d6d1841e345817431142df/py-polars/polars/_typing.py#L332-L341

```py
FileSource: TypeAlias = Union[
    str,
    Path,
    IO[bytes],
    bytes,
    list[str],
    list[Path],
    list[IO[bytes]],
    list[bytes],
]
```

[`pl.read_csv`](https://docs.pola.rs/api/python/stable/reference/api/polars.read_csv.html) is the most restrictive, but [`pl.scan_csv`](https://docs.pola.rs/api/python/stable/reference/api/polars.scan_csv.html) seems to support more?

FBruzzesi:
Closed by #3956

PR #3956, opened 2026-09-17
feat: accept file-like objects in `{read,scan}_*`

### LINKED ISSUE #3064: [Enh]: Allow `read_` and `scan_` functions to read from path-like or IO objects

00200200:
# Description

`read_csv` / `read_parquet` / `scan_csv` / `scan_parquet` only accepted paths. `BytesIO` / `StringIO` blew up in `normalize_path`, even though `write_*` already takes buffers.

I pass file-like objects through to pandas, Polars, and PyArrow. DuckDB / Ibis / Dask / Spark still raise a clear `TypeError` (they need a path). Polars `scan_*` of a buffer is eager `read_*` then `.lazy()`. PyArrow CSV text buffers get encoded to binary.

- [ ] 💾 Refactor
- [x] ✨ Feature
- [ ] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3064, #3101, #3948

## AI assistance

- [x] AI tools were used.

Used editor AI; I reviewed the diff and ran the tests.

## Checklist

- [x] Tests added
- [x] Documented the changes

```bash
uv run pytest tests/read_scan_test.py
```

208 passed, 2 skipped.


src/narwhals/_utils.py
```diff
@@ -2247,9 +2266,11 @@ def normalize_path(source: FileSource, /) -> NormalizedPath:
     # If we stringify that, we get:
     #     `'\\narwhals\\narwhals\\_utils.py'`
     # Which contains 2x `"\n"` characters
-    def normalize_path(source: FileSource, /) -> NormalizedPath:
+    def normalize_path(source: FileSource, /) -> NormalizedSource:
         from narwhals.typing import NormalizedPath
 
+        if is_file_like(source):
+            return source
         return NormalizedPath(Path(source).as_posix())
```
FBruzzesi, line 2273: Not sure where to comment this exactly, but I would do the following for simplification and naming consistency:

1. Rename both `normalize_path` to `_normalize_path`
2. Add a new helper function outside the platform branching which is:
    ```py
    def normalize_source(source: FileSource, /) -> NormalizedSource:
        return source if is_file_like(source) else _normalize_path(source)
    ```

That would then require a refactor/renaming from `normalize_path`  to `normalize_source` in a few places

docs/extending.md
```diff
@@ -67,31 +67,36 @@ shared by built-in backends and extensions alike: to support these functions, a
...
-  Narwhals has already normalized `Path` and path-like inputs before dispatching to the
+- `source` is a [`NormalizedSource`][narwhals.typing.NormalizedSource]: either a
+  normalized path string ([`NormalizedPath`][narwhals.typing.NormalizedPath], a `str`
+  [`NewType`](https://docs.python.org/3/library/typing.html#newtype)) or a file-like
+  object (`io.BytesIO` / `io.StringIO`). Narwhals has already normalized `Path` and
+  path-like inputs, and passed file-like objects through, before dispatching to the
   namespace.
+- File-like objects are supported where the native reader accepts them (pandas, Polars,
+  PyArrow, and pandas-like). Namespaces whose native readers require a path should
+  reject them with `TypeError`.
```
FBruzzesi: Couple of suggestions:

1. The `make docs-build` (and the same check we do in CI) would fail since `NormalizedSource` and `NormalizedPath` are not in the API reference page
2. I would suggest a tiny rephrasing itself:

```md
- `source` is a `NormalizedSource`: either a `NormalizedPath` (a `str`
  [`NewType`](https://docs.python.org/3/library/typing.html#newtype) tagging that
  Narwhals has already normalized `Path` and path-like inputs), or a file-like object
  such as `io.BytesIO` / `io.StringIO`, passed through untouched.
- A namespace whose native reader cannot take a file-like object should reject it with
  `TypeError`, via `narwhals._utils.ensure_path_source`.
```

src/narwhals/_utils.py
```diff
@@ -2234,11 +2236,28 @@ def validate_separators(
...
+
+def ensure_path_source(source: NormalizedSource, backend: str, /) -> NormalizedPath:
+    """Reject file-like objects for backends that only accept paths."""
+    if is_file_like(source):
+        msg = (
+            f"Reading from file-like objects is not supported for the {backend} backend.\n\n"
+            "Hint: pass a file path, or use an eager backend such as pandas, polars, or pyarrow."
+        )
+        raise TypeError(msg)
+    return source
```
FBruzzesi: I would suggest to pass backend as either an Implementation object (which `__str__` is `return str(self.value)`) or a PluginName so that we can allow plugins to use this with a dedicated name

```suggestion
def ensure_path_source(
    source: NormalizedSource, backend: Implementation | PluginName, /
) -> NormalizedPath:
    """Reject file-like objects for backends whose native reader requires a path."""
    if is_file_like(source):
        msg = (
            f"Reading from a file-like object is not supported for the {backend} backend.\n\n"
            "Hint: use 'pandas', 'polars' or 'pyarrow', or write the buffer to a file first."
        )
        raise TypeError(msg)
    return source
```

src/narwhals/functions.py
```diff
@@ -749,7 +754,9 @@ def read_parquet(
     """Read into a DataFrame from a parquet file.
 
     Arguments:
-        source: Path to a file.
+        source: Path to a file, or a file-like object such as `io.BytesIO`.
+            File-like objects are supported for eager backends (pandas, Polars,
+            PyArrow, and pandas-like).
```
FBruzzesi: `read` is already eager only, for lazy backends one must use `scan`. I would trim this line completely

```suggestion
```

src/narwhals/_arrow/namespace.py
```diff
@@ -67,9 +68,12 @@ def read_csv(
                 raise TypeError(msg)
         else:
             kwds["parse_options"] = csv.ParseOptions(delimiter=separator)
+        # PyArrow's CSV reader accepts paths and binary streams, not text buffers.
+        if isinstance(source, TextIOBase):
+            source = BytesIO(source.read().encode())  # pyright: ignore[reportAttributeAccessIssue]
```
FBruzzesi: Disclaimer: never used `parquet.csv.read_csv` directly but I double checked its argument. It's possible to pass `encoding` in its `read_options`. I understand it's niche, but since we allow it, I think we should use it also in the `.encode` here?

```suggestion
       # `pyarrow.csv` reads binary streams, but not text ones. Re-encode with the
        # declared encoding, otherwise pyarrow decodes our bytes with the wrong codec.
        if isinstance(source, TextIOBase):
            encoding = getattr(kwds.get("read_options"), "encoding", "utf8")
            source = BytesIO(source.read().encode(encoding))  # pyright: ignore[reportAttributeAccessIssue]
```

src/narwhals/_polars/namespace.py
```diff
@@ -116,22 +128,33 @@ def from_numpy(
...
         return self._dataframe.from_native(native, context=self)
 
     def scan_csv(
-        self, source: NormalizedPath, *, separator: str = ",", **kwds: Any
+        self, source: NormalizedSource, *, separator: str = ",", **kwds: Any
     ) -> PolarsLazyFrame:
-        native = pl.scan_csv(source, separator=separator, **kwds)
+        # Polars lazy scans need a path; file-like objects are read eagerly.
+        native = (
+            pl.read_csv(source, separator=separator, **kwds).lazy()
```
FBruzzesi, line 144: Seems like polars can do this natively since v1.7.0 (at least for `IO[bytes]`, see https://github.com/pola-rs/polars/pull/18532, please verify that version supports `IO[str]` as well) 

I would define an additional helper method:

```py
def _read_eagerly(self, source: NormalizedSource, /) -> bool:
        """`pl.scan_*` only accepts a file-like object from `1.7.0` onwards."""
        return is_file_like(source) and self._backend_version < (1, 7)
```

then do:

```py
native = (
    pl.read_csv(source, separator=separator, **kwds).lazy()
    if self._read_eagerly(source)
    else pl.scan_csv(source, separator=separator, **kwds)
)
```

and 

```py
native = (
    pl.read_parquet(cast("Any", source), **kwds).lazy()
    if self._read_eagerly(source)
    else pl.scan_parquet(cast("Any", source), **kwds)
)
```

src/narwhals/_utils.py
FBruzzesi: I would suggest to pass backend as either an `Implementation` object (which `__str__` is `return str(self.value)`) or a `PluginName` so that we can allow plugins to use this with a dedicated name

```suggestion
def ensure_path_source(
    source: NormalizedSource, backend: Implementation | PluginName, /
) -> NormalizedPath:
    """Reject file-like objects for backends whose native reader requires a path."""
    if is_file_like(source):
        msg = (
            f"Reading from a file-like object is not supported for the {backend} backend.\n\n"
            "Hint: use 'pandas', 'polars' or 'pyarrow', or write the buffer to a file first."
        )
        raise TypeError(msg)
    return source
```

that will require do change how you pass the backend in a few places

src/narwhals/typing.py
```diff
@@ -365,20 +365,36 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+FileSource: TypeAlias = "str | os.PathLike[str] | IO[bytes] | IO[str]"
+"""Path to a file, or a file-like object.
 
-Either a string or an object that implements [`__fspath__`], such as [`pathlib.Path`].
+Either a string, an object that implements [`__fspath__`] (such as [`pathlib.Path`]),
+or a file-like object such as [`io.BytesIO`] / [`io.StringIO`].
+
+File-like objects are forwarded to backends that support them (pandas, Polars,
+PyArrow, and pandas-like). Lazy-only backends such as DuckDB, Ibis, Dask, and
+Spark-like currently require a file path.
```
FBruzzesi:
```suggestion
Only pandas (and pandas-like), Polars and PyArrow accept a file-like object.
The lazy-only backends require a path, and raise `TypeError` otherwise.
```

src/narwhals/typing.py
```diff
@@ -365,20 +365,36 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+[`io.BytesIO`]: https://docs.python.org/3/library/io.html#io.BytesIO
+[`io.StringIO`]: https://docs.python.org/3/library/io.html#io.StringIO
 """
 
 NormalizedPath = NewType("NormalizedPath", str)
-"""A [`FileSource`][narwhals.typing.FileSource] normalized via `narwhals._utils.normalize_path`.
+"""A path-like [`FileSource`][narwhals.typing.FileSource] normalized via `narwhals._utils.normalize_path`.
+
+Path-like inputs become a plain string. File-like objects are not wrapped in this
+type — see [`NormalizedSource`][narwhals.typing.NormalizedSource].
```
FBruzzesi: I would honestly cut these lines since the type below is defined starting from this type and adding IO ones

src/narwhals/_duckdb/namespace.py
FBruzzesi, line 1: The duckdb case is actually a bit of a mindbending one: I would avoid raising a `TypeError` ourselves since it supports file-like objects, but only if fsspec is installed. For csv from our minimum supported version (v1.1), for parquet since very recently (v1.5.4). This means we need to add some special casing for it.

Consider the following script:

```py
from io import BytesIO
from typing import Any

import duckdb
import polars as pl

data = {"a": [1, 2, 3], "b": [4.5, 6.7, 8.9], "z": ["x", "y", "w"]}


def csv() -> BytesIO:
    buf = BytesIO()
    pl.DataFrame(data).write_csv(buf)
    buf.seek(0)
    return buf


def parquet() -> BytesIO:
    buf = BytesIO()
    pl.DataFrame(data).write_parquet(buf)
    buf.seek(0)
    return buf


def show(label: str, fn: Any) -> None:
    try:
        print(f"{label:<44} {fn()}")
    except Exception as exc:  # noqa: BLE001
        print(f"{label:<44} {type(exc).__name__}: {str(exc).splitlines()[0]}")


print(f"duckdb {duckdb.__version__}\n")

show("duckdb.read_csv(BytesIO)", lambda: duckdb.read_csv(csv()).fetchall())
show("duckdb.read_parquet(BytesIO)", lambda: duckdb.read_parquet(parquet()).fetchall())
```

no fsspec:

```terminal
$ uv run --isolated --with duckdb --with polars t.py
duckdb 1.5.5

duckdb.read_csv(BytesIO)                     InvalidInputException: Invalid Input Error: This operation could not be completed because required module 'fsspec' is not installed
duckdb.read_parquet(BytesIO)                 InvalidInputException: Invalid Input Error: This operation could not be completed because required module 'fsspec' is not installed
```

With fsspec and various duckdb versions:

```terminal
$ uv run --isolated --with duckdb==1.1.0 --with polars --with fsspec t.py
duckdb 1.1.0

duckdb.read_csv(BytesIO)                     [(1, 4.5, 'x'), (2, 6.7, 'y'), (3, 8.9, 'w')]
duckdb.read_parquet(BytesIO)                 TypeError: read_parquet(): incompatible function arguments. The following argument types are supported:
```

```terminal
$ uv run --isolated --with duckdb==1.5.3 --with polars --with fsspec t.py
duckdb 1.5.3

duckdb.read_csv(BytesIO)                     [(1, 4.5, 'x'), (2, 6.7, 'y'), (3, 8.9, 'w')]
duckdb.read_parquet(BytesIO)                 TypeError: read_parquet(): incompatible function arguments. The following argument types are supported:
```

```terminal
$ uv run --isolated --with duckdb==1.5.4 --with polars --with fsspec t.py
duckdb 1.5.4

duckdb.read_csv(BytesIO)                     [(1, 4.5, 'x'), (2, 6.7, 'y'), (3, 8.9, 'w')]
duckdb.read_parquet(BytesIO)                 [(1, 4.5, 'x'), (2, 6.7, 'y'), (3, 8.9, 'w')]
```

---

This actually requires to change all other statements in which we mention that lazy-only backends do not support file-like objects

tests/read_scan_test.py
FBruzzesi, line 1: I have a few suggestions for the test suite, but I will keep them outside of the current review as the main codebase needs some adjustments to begin with. In general they are more stylistic and personal preference more than anything else

FBruzzesi (changes requested):
Hey @00200200 thanks for your contribution, I think this PR goes in the right direction but currently leaves a few details out that we need to address

00200200:
Pushed a follow-up that should cover the review notes:

- `normalize_path` is now `_normalize_path`, with `normalize_source` in front
- `ensure_path_source` takes `Implementation | PluginName`
- PyArrow CSV re-encodes text buffers using `read_options.encoding`
- Polars `scan_*` uses native file-like support from 1.7 (`_read_eagerly` below that)
- DuckDB no longer raises `TypeError` itself; file-likes go to the native reader (`fsspec` / version as upstream)
- docs + API reference members for `NormalizedPath` / `NormalizedSource`

Tests for path-only lazy backends still expect `TypeError`; DuckDB file-like coverage is separate.

docs/api-reference/typing.md
FBruzzesi, line 1: Thanks for adding these 🙏🏼

src/narwhals/_duckdb/namespace.py
```diff
@@ -75,13 +75,13 @@ def _lazyframe(self) -> type[DuckDBLazyFrame]:
...
+        self, source: NormalizedSource, *, separator: str = ",", **kwds: Any
     ) -> DuckDBLazyFrame:
         validate_separators(separator, ("delimiter", "delim", "sep"), kwds)
         native = duckdb.read_csv(source, delimiter=separator, **kwds)
         return self._lazyframe.from_native(native, context=self)
 
-    def scan_parquet(self, source: NormalizedPath, **kwds: Any) -> DuckDBLazyFrame:
+    def scan_parquet(self, source: NormalizedSource, **kwds: Any) -> DuckDBLazyFrame:
         native = duckdb.read_parquet(source, **kwds)
         return self._lazyframe.from_native(native, context=self)
```
FBruzzesi, line 94: For scan_parquet I would add a version check:

```suggestion
    def scan_parquet(self, source: NormalizedSource, **kwds: Any) -> DuckDBLazyFrame:
        if is_file_like(source) and (version := self._backend_version) < (1, 5, 4):
            from narwhals._utils import parse_version
            msg = (
                "`scan_parquet` from a file-like object is only available in "
                f"'duckdb>=1.5.4', found version {parse_version(version)!r}."
            )
            raise NotImplementedError(msg)
        native = duckdb.read_parquet(source, **kwds)
        return self._lazyframe.from_native(native, context=self)
```

(formatting needed)

src/narwhals/functions.py
```diff
@@ -697,7 +698,10 @@ def scan_csv(
     a csv file eagerly and then converts the resulting dataframe to a lazyframe.
 
     Arguments:
-        source: Path to a file.
+        source: Path to a file, or a file-like object such as `io.StringIO` /
+            `io.BytesIO`. File-like objects are supported for pandas, Polars,
+            PyArrow, pandas-like backends, and DuckDB (when `fsspec` is installed).
+            Ibis, Dask, and Spark-like backends require a file path.
```
FBruzzesi: Since the same explanation is in `FileSource` and that's now exposed in the API reference, we can link it rather than repeat ourselves:

```suggestion
        source: Path to a file, or a file-like object such as `io.StringIO` /
            `io.BytesIO`. See [`FileSource`][narwhals.typing.FileSource] for which
            backends accept a file-like object.
```

src/narwhals/functions.py
```diff
@@ -822,7 +826,10 @@ def scan_parquet(
         ```
 
     Arguments:
-        source: Path to a file.
+        source: Path to a file, or a file-like object such as `io.BytesIO`.
+            File-like objects are supported for pandas, Polars, PyArrow,
+            pandas-like backends, and DuckDB (when `fsspec` is installed).
+            Ibis, Dask, and Spark-like backends require a file path.
```
FBruzzesi: Same as for `scan_csv`:


```suggestion
        source: Path to a file, or a file-like object such as `io.BytesIO`.
            See [`FileSource`][narwhals.typing.FileSource] for which backends
            accept a file-like object.
```

src/narwhals/typing.py
```diff
@@ -365,20 +365,34 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
+"""Path to a file, or a file-like object.
 
-Either a string or an object that implements [`__fspath__`], such as [`pathlib.Path`].
+Either a string, an object that implements [`__fspath__`] (such as [`pathlib.Path`]),
+or a file-like object such as [`io.BytesIO`] / [`io.StringIO`].
+
+Only pandas (and pandas-like), Polars and PyArrow accept a file-like object.
+DuckDB also accepts one when `fsspec` is installed (CSV from the minimum
+supported DuckDB version; Parquet from 1.5.4). The other lazy-only backends
+require a path, and raise `TypeError` otherwise.
```
FBruzzesi:
```suggestion
pandas (and pandas-like), Polars and PyArrow accept a file-like object. DuckDB
does too, but only with `fsspec` installed, and `scan_parquet` additionally needs
`duckdb>=1.5.4`. Dask, Ibis and Spark-like require a path and raise `TypeError`
for anything else.
```

src/narwhals/typing.py
```diff
@@ -365,20 +365,34 @@ def Binary(self) -> type[dtypes.Binary]: ...
...
 IntoPandasSchema: TypeAlias = Mapping[str, PandasLikeDType]
 
-FileSource: TypeAlias = "str | os.PathLike[str]"
-"""Path to a file.
+FileSource: TypeAlias = "str | os.PathLike[str] | IO[bytes] | IO[str]"
+"""Path to a file, or a file-like object.
 
-Either a string or an object that implements [`__fspath__`], such as [`pathlib.Path`].
+Either a string, an object that implements [`__fspath__`] (such as [`pathlib.Path`]),
+or a file-like object such as [`io.BytesIO`] / [`io.StringIO`].
```
FBruzzesi: I know this is pre-existing issue, but the links are not working. If I try to build the docs locally, this is what I see:

Image

Could you fix it at least for this type as we are touching it in this PR?

src/narwhals/stable/v2/__init__.py
FBruzzesi, line 1: Docstrings need to match the ones in `functions.py` otherwise a test will fail

tests/read_scan_test.py
FBruzzesi, line 1: Ok I don't think I have a much better way to comment of this than showing how I would rework it via its diff:

```diff
diff --git a/tests/read_scan_test.py b/tests/read_scan_test.py
index 8e1a2e6f0..d4467497c 100644
--- a/tests/read_scan_test.py
+++ b/tests/read_scan_test.py
@@ -39,6 +39,8 @@ skipif_pandas_lt_1_5 = pytest.mark.skipif(
 )
 lazy_core_backend = pytest.mark.parametrize("backend", ["duckdb", "ibis", "sqlframe"])
 spark_like_backend = pytest.mark.parametrize("backend", ["pyspark", "sqlframe"])
+# Backends whose native reader cannot take a file-like object.
+path_only_backend = pytest.mark.parametrize("backend", ["dask", "ibis", "sqlframe"])
 
 
 class MockPathLike:
@@ -252,84 +254,90 @@ def test_scan_csv_raise_on_conflicting_separator(
         nw.scan_csv("unused.csv", backend=backend, separator="|", **kwargs)
 
 
-def _csv_string_buffer() -> StringIO:
-    buf = StringIO()
-    pl.DataFrame(data).write_csv(buf)
+def _csv_buffer(
+    into: type[StringIO | BytesIO] = BytesIO, /, *, separator: str = ","
+) -> StringIO | BytesIO:
+    buf = into()
+    pl.DataFrame(data).write_csv(buf, separator=separator)
     buf.seek(0)
     return buf
 
 
-def _csv_bytes_buffer() -> BytesIO:
-    buf = BytesIO()
-    pl.DataFrame(data).write_csv(buf)
-    buf.seek(0)
-    return buf
-
-
-def _parquet_bytes_buffer() -> BytesIO:
+def _parquet_buffer() -> BytesIO:
     buf = BytesIO()
     pl.DataFrame(data).write_parquet(buf)
     buf.seek(0)
     return buf
 
 
-@pytest.mark.parametrize("backend", ["pandas", "polars", "pyarrow"])
-@pytest.mark.parametrize("into_buffer", [_csv_string_buffer, _csv_bytes_buffer])
+@pytest.mark.parametrize("into", [StringIO, BytesIO])
 def test_read_csv_file_like(
-    backend: Literal["pandas", "polars", "pyarrow"], into_buffer: Callable[[], Any]
+    eager_backend: EagerAllowed, into: type[StringIO | BytesIO]
 ) -> None:
-    pytest.importorskip(backend)
-    assert_equal_eager(nw.read_csv(into_buffer(), backend=backend))
+    assert_equal_eager(nw.read_csv(_csv_buffer(into), backend=eager_backend))
+    assert_equal_eager(
+        nw.read_csv(
+            _csv_buffer(into, separator="|"), backend=eager_backend, separator="|"
+        )
+    )
 
 
-@pytest.mark.parametrize("backend", ["pandas", "polars", "pyarrow"])
-@pytest.mark.parametrize("into_buffer", [_csv_string_buffer, _csv_bytes_buffer])
+@pytest.mark.parametrize("into", [StringIO, BytesIO])
 def test_scan_csv_file_like(
-    backend: Literal["pandas", "polars", "pyarrow"], into_buffer: Callable[[], Any]
+    eager_backend: EagerAllowed, into: type[StringIO | BytesIO]
 ) -> None:
-    pytest.importorskip(backend)
-    assert_equal_lazy(nw.scan_csv(into_buffer(), backend=backend))
+    assert_equal_lazy(nw.scan_csv(_csv_buffer(into), backend=eager_backend))
+
+
+def test_read_csv_file_like_pyarrow_encoding() -> None:
+    # A text buffer is re-encoded before reaching `pyarrow.csv`, so it has to use
+    # whatever `read_options.encoding` declares, else the values come back mojibake.
+    from pyarrow import csv
+
+    result = nw.read_csv(
+        StringIO("z\nrené\n"),
+        backend="pyarrow",
+        read_options=csv.ReadOptions(encoding="latin-1"),
+    )
+    assert result.rows() == [("rené",)]
 
 
 @skipif_pandas_lt_1_5
-@pytest.mark.parametrize("backend", ["pandas", "polars", "pyarrow"])
-def test_read_parquet_file_like(backend: Literal["pandas", "polars", "pyarrow"]) -> None:
-    pytest.importorskip(backend)
-    assert_equal_eager(nw.read_parquet(_parquet_bytes_buffer(), backend=backend))
+def test_read_parquet_file_like(eager_backend: EagerAllowed) -> None:
+    assert_equal_eager(nw.read_parquet(_parquet_buffer(), backend=eager_backend))
 
 
 @skipif_pandas_lt_1_5
-@pytest.mark.parametrize("backend", ["pandas", "polars", "pyarrow"])
-def test_scan_parquet_file_like(backend: Literal["pandas", "polars", "pyarrow"]) -> None:
-    pytest.importorskip(backend)
-    assert_equal_lazy(nw.scan_parquet(_parquet_bytes_buffer(), backend=backend))
+def test_scan_parquet_file_like(eager_backend: EagerAllowed) -> None:
+    assert_equal_lazy(nw.scan_parquet(_parquet_buffer(), backend=eager_backend))
 
 
-@pytest.mark.parametrize("backend", ["ibis", "sqlframe"])
-def test_scan_csv_file_like_unsupported(backend: Literal["ibis", "sqlframe"]) -> None:
+@path_only_backend
+def test_scan_csv_file_like_unsupported(backend: _LazyOnly) -> None:
     pytest.importorskip(backend)
     with pytest.raises(TypeError, match="file-like"):
-        nw.scan_csv(_csv_string_buffer(), backend=backend)
+        nw.scan_csv(_csv_buffer(StringIO), backend=backend)
 
 
-@pytest.mark.parametrize("backend", ["ibis", "sqlframe"])
-def test_scan_parquet_file_like_unsupported(backend: Literal["ibis", "sqlframe"]) -> None:
+@path_only_backend
+def test_scan_parquet_file_like_unsupported(backend: _LazyOnly) -> None:
     pytest.importorskip(backend)
     with pytest.raises(TypeError, match="file-like"):
-        nw.scan_parquet(_parquet_bytes_buffer(), backend=backend)
+        nw.scan_parquet(_parquet_buffer(), backend=backend)
 
 
-def test_scan_csv_file_like_duckdb() -> None:
+@pytest.mark.parametrize("into", [StringIO, BytesIO])
+def test_scan_csv_file_like_duckdb(into: type[StringIO | BytesIO]) -> None:
     pytest.importorskip("duckdb")
     pytest.importorskip("fsspec")
-    assert_equal_lazy(nw.scan_csv(_csv_bytes_buffer(), backend="duckdb"))
-    assert_equal_lazy(nw.scan_csv(_csv_string_buffer(), backend="duckdb"))
+    assert_equal_lazy(nw.scan_csv(_csv_buffer(into), backend="duckdb"))
 
 
-@skipif_pandas_lt_1_5
 def test_scan_parquet_file_like_duckdb() -> None:
     pytest.importorskip("duckdb")
     pytest.importorskip("fsspec")
     if DUCKDB_VERSION < (1, 5, 4):
-        pytest.skip("read_parquet(file-like) requires duckdb>=1.5.4")
-    assert_equal_lazy(nw.scan_parquet(_parquet_bytes_buffer(), backend="duckdb"))
+        with pytest.raises(NotImplementedError, match=r"duckdb>=1\.5\.4"):
+            nw.scan_parquet(_parquet_buffer(), backend="duckdb")
+        return
+    assert_equal_lazy(nw.scan_parquet(_parquet_buffer(), backend="duckdb"))

```

As you can see there are a few simplification thanks to the fact that we can use the dataframe constructor fixtures directly, removes the `@skipif_pandas_lt_1_5` marker when pandas is not used, and simplify the `_csv_buffer` builder.

FBruzzesi (commented):
Thanks a ton for adjusting so promptly. I just did one more review pass. I think we should be ready after these comments are addressed 😇

00200200:
Pushed a follow-up for the second pass:

- DuckDB `scan_parquet` raises `NotImplementedError` for file-like objects on `duckdb<1.5.4`
- `scan_csv` / `scan_parquet` docs now point at `FileSource` instead of repeating backend rules
- `FileSource` uses inline links so they render in the API docs
- file-like tests follow the suggested rework (`eager_backend`, shared buffer helpers, path-only backends including dask)

00200200:
Follow-up for the second pass: `FileSource` now uses inventory cross-refs so the stdlib links resolve, `read_csv`/`read_parquet` point at `FileSource` the same way `scan_*` do, and the DuckDB parquet floor uses `requires._unparse_version`. Tests already follow the `path_only_backend` / `_csv_buffer` layout from the review diff.

00200200:
Tox was failing on the StringIO CSV fixtures: older Polars `write_csv` emits bytes and raises if that lands in a text buffer. The helper now writes to `BytesIO` first and only then wraps a `StringIO`.

FBruzzesi:
don't worry about pyspark connect job failure as it's unrelated

FBruzzesi (approved):
Thanks a ton @00200200 you have my +1

FBruzzesi:
@MarcoGorelli I am pretty happy with the state of this - let me know if you are ok with merging or want to take the time to review it yourself as well

MarcoGorelli:
thanks both!

from a quick glance it's ok, feel free to merge if happy!

PR #3958, opened 2026-09-18
fix: correct multibyte handling in polars zfill sign compat branch

### LINKED ISSUE #3838: bug: `zfill` differs between Polars <=1.30 and newer for multibyte strings with a leading plus
## Describe the bug

`zfill` gives a different answer on Polars `<= 1.30.0` than on newer Polars, for strings that start with `+` and hold a multibyte character. Same narwhals, same input, two results.

The compatibility branch in `_polars/expr.py` that re-pads a leading `+` on old Polars mixes two units: `str.zfill` pads to a count of **bytes**, `str.pad_start` pads to a count of **characters**. On ASCII the two agree and the branch is correct; on multibyte content it prepends one `+` too many.

Only `<= 1.30.0` is affected, which is why this does not reproduce on a current install.

```python
import polars as pl
import narwhals as nw

s = nw.from_native(pl.Series("a", ["+é", "+12"]), series_only=True)
print(pl.__version__, [s.str.zfill(w).to_list() for w in (3, 4, 5)])
```

Run it twice, once with `polars==1.30.0` and once with a newer Polars.

## Expected results

The same answer on both, since narwhals is what is being asked:

```
['+é', '+0é', '+00é']      # for "+é" at widths 3, 4, 5
```

## Actual results

```
polars 1.42.1  [['+é', '++7', '+12', '+01'], ['+0é', ...], ['+00é', ...]]
polars 1.30.0  [['++é', '++7', '+12', '+01'], ['++0é', ...], ['++00é', ...]]
                  ^^                            ^^^
```

Pure-ASCII input agrees on both, including `"++7"`, `"+12"` and `"+1"`.

## Please run narwhals.show_versions() and enter the output below.

```
narwhals: main @ 7880b36c | polars: 1.30.0 and 1.42.1 | python: 3.10
```

## Relevant log output

The branch, at `src/narwhals/_polars/expr.py`:

```python
length = self.native.str.len_chars()          # characters
...
self.native.str.slice(1, length)
    .str.zfill(width - 1)                     # pads to a byte count
    .str.pad_start(width, plus)               # pads to a character count
```

For `"+é"` at width 4: `slice` gives `"é"`, `zfill(3)` gives `"0é"` because `"é"` is two bytes, and `pad_start(4, "+")` then sees two characters and adds two `+`.

Raised from https://github.com/narwhals-dev/narwhals/pull/3770#issuecomment-5184135525. Happy to open a PR, though a correct fix looks like reworking the branch rather than patching the concat: prepending exactly one `+` fixes the multibyte case but the re-slicing approach has its own edges.

Aniket-a14:
# Description

Fixes the pre-1.30 polars `zfill` compat branch (`PolarsExprStringNamespace.zfill` in `src/narwhals/_polars/expr.py`) for strings that start with `+` and contain a multibyte character.

The branch re-pads a leading `+` for older polars by:
1. stripping the sign and zero-filling the remainder with `.str.zfill(width - 1)` (byte-based padding, matching native `zfill` semantics), then
2. re-adding the sign with `.str.pad_start(width, plus)` (character-based padding).

Mixing a byte-based step with a character-based step is fine for ASCII, but for multibyte content it prepends one extra `+`. For example, `"+é".zfill(4)` on old polars returned `"++0é"` instead of `"+0é"`.

The fix replaces the character-based `pad_start` re-attachment with a plain string concatenation of the sign, so the whole operation stays consistent with `zfill`'s own byte-based semantics — matching what a correct/native `zfill` call produces on newer polars versions.

I verified this against a real `polars==1.30.0` install (the affected version) in an isolated venv: the old code reproduces the bug reported in the issue (`+é` → `++é`/`++0é`/`++00é` for widths 3/4/5), and the fixed code produces the correct `+é`/`+0é`/`+00é`, matching current (>1.30) native `zfill` output exactly, with no change for the existing ASCII test cases.

Note: the affected branch is gated behind `backend_version <= (1, 30, 0)` and is excluded from the local coverage report via the project's `backend_version <` coverage-exclude pattern, since polars pinned for local dev (1.44.x) never takes that branch. It is exercised for real in CI's `extreme-pretty-old-versions` / `extreme-minimum-versions` jobs, which pin `polars==0.20.4`.

## Steps or code to reproduce (from the issue)

```python
import polars as pl
import narwhals as nw

s = nw.from_native(pl.Series("a", ["+é", "+12"]), series_only=True)
print(pl.__version__, [s.str.zfill(w).to_list() for w in (3, 4, 5)])
```

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3838

## AI assistance

- [x] AI tools were used.

This PR was written with Claude Code (Anthropic, Sonnet 5). I reviewed the diff, reproduced the original bug and verified the fix against a real `polars==1.30.0` install, and ran the full local test/lint/typing suite myself before opening this PR.

## Checklist

- [x] Tests added
- [x] Documented the changes

I don't have a GUI to take a screenshot, so here is the raw terminal output of the exact command from this template, run locally (not in CI):

```
$ PYTEST_ADDOPTS="--numprocesses=logical" \
...
================================ tests coverage ================================
______________ coverage: platform darwin, python 3.11.16-final-0 _______________

Name                    Stmts   Miss Branch BrPart  Cover   Missing
-------------------------------------------------------------------
tests/version_test.py      22      1      0      0    95%   32
-------------------------------------------------------------------
TOTAL                   28577      1   2900      0    99%

454 files skipped due to complete coverage.
Required test coverage of 80.0% reached. Total coverage: 100.00%
=============== 15441 passed, 248 skipped, 996 xfailed in 40.03s ===============
```

`make typing` (mypy, pyright, pyrefly) and `ruff format`/`ruff check` also pass locally on the changed files.


Aniket-a14:
Hey @FBruzzesi can you review this pr?

tests/expr_and_series/str/zfill_test.py
```diff
@@ -33,6 +33,11 @@
         id="width_0",
     ),
     pytest.param({"a": ["日本", None]}, 3, {"a": ["日本", None]}, id="non_ascii"),
+    pytest.param({"a": ["+é", "+12"]}, 3, {"a": ["+é", "+12"]}, id="multibyte_plus_w3"),
+    pytest.param({"a": ["+é", "+12"]}, 4, {"a": ["+0é", "+012"]}, id="multibyte_plus_w4"),
+    pytest.param(
+        {"a": ["+é", "+12"]}, 5, {"a": ["+00é", "+0012"]}, id="multibyte_plus_w5"
+    ),
```
FBruzzesi: I would keep using `"non_ascii"` prefix in the id name to simplify the filtering below (ie. keep the `"non_ascii" in request.node.callspec.id` condition). I think I found another edge case to add here:



```suggestion
    # A `+` in front of a multibyte character: Polars pads to a byte count, every other
    # backend to a character count.
    pytest.param({"a": ["+é", "+12"]}, 3, {"a": ["+é", "+12"]}, id="non_ascii_plus_w3"),
    pytest.param({"a": ["+é", "+12"]}, 4, {"a": ["+0é", "+012"]}, id="non_ascii_plus_w4"),
    pytest.param(
        {"a": ["+é", "+12"]}, 5, {"a": ["+00é", "+0012"]}, id="non_ascii_plus_w5"
    ),
    # Only the first character is a sign; the second one is padded like any other.
    pytest.param(
        {"a": ["+-1", "-+1", "+-", "-+"]},
        4,
        {"a": ["+0-1", "-0+1", "+00-", "-00+"]},
        id="nested_sign",
    ),
```

src/narwhals/_polars/expr.py
```diff
@@ -474,11 +474,7 @@ def zfill(self, width: int) -> PolarsExpr:
             starts_with_plus = self.native.str.starts_with(plus)
             native_result = (
                 pl.when(starts_with_plus & less_than_width)
-                .then(
-                    self.native.str.slice(1, length)
-                    .str.zfill(width - 1)
-                    .str.pad_start(width, plus)
-                )
+                .then(plus + self.native.str.slice(1, length).str.zfill(width - 1))
```
FBruzzesi: I believe this fix is incomplete and it has a couple of bugs:

1. "+" followed by "-" is still wrong as `plus + rest.str.zfill(width - 1)` delegates the remainder to old polars zfill, which itself special-cases a leading -. So the zeros land after the inner - instead of before it. Examples: `"+-1".zfill(4)` gives `+-01`, while it should be `+0-1`. `"+-".zfill(3)` gives `+-0`, while it should be `+0-`.
2. `plus + expr` renames the output column to "literal" instead of maintaining the original column name.

One _hack_ I found to fix both is the following:

```python
@requires.backend_version((0, 20, 5))
def zfill(self, width: int) -> PolarsExpr:
    if width == 0:
        return self.compliant._with_native(self.native)

    native_result = self.native.str.zfill(width)
    if self.compliant._backend_version <= (1, 30, 0):
        plus, minus = "+", "-"
        sign, rest = self.native.str.slice(0, 1), self.native.str.slice(1)
        native_result = (
            pl.when(self.native.str.starts_with(plus))
            .then(sign + (minus + rest).str.zfill(width).str.slice(1))
            .otherwise(native_result)
        )

    return self.compliant._with_native(native_result)
```

FBruzzesi (changes requested):
Thanks for your contribution @Aniket-a14 - I found a couple of issues in the rewrite, thus left a couple of comments.

If you want to test your changes locally, you need to install a polars version that enters such conditional, i.e. a version before v1.30.0, otherwise you won't see the issues

tests/expr_and_series/str/zfill_test.py
Aniket-a14: Thanks, applied: ids renamed to `non_ascii_*` (original filter condition restored) and the `nested_sign` case added.

src/narwhals/_polars/expr.py
Aniket-a14: Good catch on both, adopted your approach (`+-1` -> `+0-1`, column name preserved) and verified the tests against a real polars 1.30.0 install.

Aniket-a14:
Hey @FBruzzesi  thanks for the suggestions, can you review them?

FBruzzesi (approved):
Thanks for bearing with me and for your patience @Aniket-a14 

Happy to merge as soon as all checks pass

PR #3959, opened 2026-09-18
fix: round half to even in `round` for PyArrow and SQL backends

### LINKED ISSUE #3698: [Bug]: round() half-tie behavior differs across backends; docstring is out of date for Polars
## Describe the bug

`round()` on values exactly halfway between two integers gives different results across backends: pandas and Polars round half-to-even, while PyArrow and DuckDB round half-away-from-zero. The `round` docstring documents this as pandas (even) vs "Polars and Arrow" (away from 0), but current Polars (1.41.2) rounds half-to-even — so the docstring is now wrong about Polars. The real split is `{pandas, Polars}` half-to-even vs `{PyArrow, DuckDB}` half-away-from-zero.

```python
import narwhals as nw
import polars as pl, pandas as pd, pyarrow as pa, duckdb

vals = [0.5, 1.5, 2.5, 3.5, -0.5, -2.5]
print("polars :", nw.from_native(pl.Series("a", vals), series_only=True).round(0).to_list())
print("pandas :", nw.from_native(pd.Series(vals, name="a"), series_only=True).round(0).to_list())
print("pyarrow:", nw.from_native(pa.chunked_array([vals]), series_only=True).round(0).to_list())
rel = duckdb.sql("SELECT * FROM (VALUES (0.5),(1.5),(2.5),(3.5),(-0.5),(-2.5)) t(a)")
print("duckdb :", nw.from_native(rel).select(nw.col("a").round(0)).to_native().fetchall())
```

## Expected results

Either the backends agree, or the docstring describes the real behavior. The PyArrow path hardcodes `round_mode="half_towards_infinity"` (`src/narwhals/_arrow/series.py:496`). Two options, and I'm happy to PR whichever you prefer:

1. Align PyArrow (and DuckDB) to half-to-even so all backends match pandas/Polars, plus update the docstring and add a half-tie test.
2. Keep per-backend behavior and just correct the docstring to `{pandas, Polars}` = half-to-even, `{PyArrow, DuckDB}` = away from 0.

The existing test (`tests/expr_and_series/round_test.py`) computes expected values with Python's built-in `round()` (half-to-even) and uses no half-ties (`[2.12345, 2.56789, 3.901234]`), so this isn't currently covered.

## Actual results

```
polars : [0.0, 2.0, 2.0, 4.0, -0.0, -2.0]   # half-to-even
pandas : [0.0, 2.0, 2.0, 4.0, -0.0, -2.0]   # half-to-even
pyarrow: [1.0, 2.0, 3.0, 4.0, -1.0, -3.0]   # half-away-from-zero
duckdb : [(Decimal('1'),), (Decimal('2'),), (Decimal('3'),), (Decimal('4'),), (Decimal('-1'),), (Decimal('-3'),)]   # half-away-from-zero
```

Docstring (`src/narwhals/expr.py`, `round` Notes): "Polars and Arrow round away from 0 (e.g. -0.5 to -1.0, 0.5 to 1.0, 1.5 to 2.0, 2.5 to 3.0, etc..)." — but Polars gives `0.5 -> 0.0` and `2.5 -> 2.0`.

## narwhals.show_versions()

```
narwhals: 2.22.1 | pandas: 3.0.3 | pyarrow: 24.0.0 | polars: 1.41.2 | duckdb: 1.5.4 | numpy: 2.4.6 | python: 3.12.13 | machine: macOS-arm64
```

ethanstoner:
# Description

`round` breaks ties differently depending on the backend. pandas, Polars and Dask round half to even, while PyArrow, DuckDB, SQLFrame, Ibis and PySpark round half away from zero. The docstrings also say Polars rounds away from zero, which is wrong.

This makes every backend round half to even, which is the default-alignment approach sueun-dev proposed in #3698 and @MarcoGorelli okayed. (sueun-dev offered to open a PR back in June; thanks for the groundwork and the reproducer.) It doesn't add a `mode=` parameter; that can be a follow-up.

| Backend | 0.5, 1.5, 2.5, -0.5, -2.5 before | after |
|---|---|---|
| pandas / Polars / Dask | 0, 2, 2, -0, -2 | unchanged |
| PyArrow | 1, 2, 3, -1, -3 | 0, 2, 2, -0, -2 |
| DuckDB, SQLFrame, Ibis | 1, 2, 3, -1, -3 | 0, 2, 2, 0, -2 |
| PySpark | 1, 2, 3, -1, -3 | 0, 2, 2, 0, -2 |

Changes:
- **PyArrow:** `round_mode="half_to_even"` instead of `"half_towards_infinity"`.
- **SQL backends** (`_sql/expr.py`, shared by DuckDB, SQLFrame and Ibis): SQL `round` plus a tie correction. When the value is exactly halfway and the rounded result is odd, it takes the other neighbor. I didn't use DuckDB's `round_even` because it goes through `x/2` and `power`, which turns every input into DOUBLE (`round_even(12345678901234567::BIGINT, 0)` gives `1.2345678901234568e+16`). A small `_cast_like` hook keeps Decimal precision exact on DuckDB 1.3+ and Ibis.
- **PySpark:** native `bround`. SQLFrame stays on the shared path because its DuckDB dialect has no `bround`.
- **Docstrings** in `expr.py` and `series.py` now describe the actual behavior.
- **Tests:** half-tie cases at 0, 1 and 2 decimals, the issue's DuckDB Decimal reproducer, and a check that integer dtypes stay integer.

Caveat: on DuckDB 1.1/1.2 the values are right, but a tied Decimal result can come back with wider precision, since `cast_to_type` only exists from 
pytest-local-pass-cropped
1.3.

Full local suite (the command below): 15493 passed, 0 failed. As an extra check, about 7,700 values at 0-3 decimals matched Polars exactly on DuckDB and SQLFrame.

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [x] 📝 Documentation
- [x] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3698

## AI assistance

- [x] AI tools were used.

I used Claude Code (Claude Opus 5) to write the fix, the tests and a first draft of this description. I read the whole diff and ran the full test suite locally before opening this.

## Checklist

- [x] Tests added
- [x] Documented the changes

   
pytest-local-pass-cropped


src/narwhals/_duckdb/expr.py
FBruzzesi, line 1: DuckDB natively implements [`round_even`](https://duckdb.org/docs/lts/sql/functions/numeric#round_evenv-numeric-s-integer), no need to dispatch to the SQL implementation

src/narwhals/_spark_like/expr.py
FBruzzesi, line 1: SQLFrame also supports `bround` for at least some backends. If that's not the case for DuckDB, I would suggest to open an issue/feature request upstream in SQLFrame asking for it since DuckDB supports even rounding natively, and we can use `self._F.bround` for all spark-like backends

src/narwhals/_sql/expr.py
FBruzzesi, line 1: Given the previous two comments, the only SQL-like backend that does not seem to implement even rounding natively is ibis, therefore I would move the dedicated implementation in its namespace/module.

That said, maybe this is what you are doing here, so apologies if I am misreading something. While searching for native implementation, I stumbled upon the feature request in DuckDB and especially a comment that implements it in SQL:

```sql
round_even(x, n) AS
	case ((abs(x) * power(10, n+1)) % 10)
	when 5 then round(x/2, n) * 2
	else round(x, n)
	end
```

which seems a bit less complex than what you have here so far?

See [DuckDB#2348(Support Bankers rounding)](https://github.com/duckdb/duckdb/issues/2348#issuecomment-930413410)

FBruzzesi (changes requested):
Thanks for your contribution @ethanstoner - I believe we can simplify things a bit using native functionalities. Left comments for each file where this is relevant

ethanstoner:
Thanks — I measured all three against DuckDB 1.5.5 and SQLFrame 4.4.0 before changing anything. The native paths don't hold up for the two backends the shared correction exists for, so the rounding logic is unchanged; `df527a3c` only writes the rationale into the code and pins the strongest piece of evidence as a test.

**1. `round_even` returns `DOUBLE` for every input type.**

| input | n | `round_even` | `round` |
|---|---|---|---|
| `9007199254740993::BIGINT` | 0 | `DOUBLE` 9007199254740992.0 | `BIGINT` 9007199254740993 |
| `123456789012345678::BIGINT` | 0 | `DOUBLE` 1.2345678901234568e+17 | `BIGINT` 123456789012345678 |
| `25::BIGINT` | -1 | `DOUBLE` 20.0 | `BIGINT` 30 |
| `2.5::DECIMAL(18,1)` | 0 | `DOUBLE` 2.0 | `DECIMAL(18,0)` 3 |

Two separate problems: `Int64`/`Decimal` come back `Float64`, and integers above 2\*\*53 are corrupted — 9007199254740993 returns 9007199254740992 on a call that should be the identity, since rounding an integer to 0 decimals is a no-op.

Dtype preservation isn't something this PR introduced. On `main` (ddfde629), DuckDB `round` already returns `Int32` for ints and `Decimal(2, 0)` for decimals, and polars, pandas and pyarrow all map `Int64 -> Int64`. Substituting native `round_even` into `_duckdb/expr.py` and running the round tests on the duckdb constructor gives `3 failed, 6 passed, 6 skipped` — the two `test_round_keeps_integer_dtype` cases and the decimal case. Every half-tie *value* test passes, so `round_even` gets the values right; it's the types and the large integers that break.

`cast_to_type(round_even(x, n), x)` doesn't rescue it either: the `DOUBLE` round trip happens before the cast, so `9007199254740993::BIGINT` still comes back 9007199254740992, and `2.5::DECIMAL(18,1)` casts back to the input scale `DECIMAL(18,1)` rather than `round`'s `DECIMAL(18,0)`.

What's in the PR substitutes a value only on an actual tie and otherwise returns `round`'s output untouched, so non-tie values never pass through `DOUBLE`.

**2. SQLFrame doesn't expose `bround` on duckdb or postgres.** `hasattr(sqlframe.<backend>.functions, "bround")` on 4.4.0: duckdb **False**, postgres **False**, bigquery/snowflake/spark/redshift True. `test-full-coverage` passes `--constructors=...,sqlframe`, which runs through `DuckDBSession`, so `self._F.bround` for all spark-like backends would raise `AttributeError` in CI rather than degrade quietly. PySpark's own `bround` is already used on the branch, gated on `is_pyspark()` / `is_pyspark_connect()`.

Your upstream suggestion is the right long-term fix, though — `sqlframe` already has `bround_using_half_even` for snowflake, so the machinery is there. Happy to open that feature request on `eakmanrq/sqlframe` as a follow-up, which would let the correction be dropped later.

**3. That makes the set needing the fallback {DuckDB, sqlframe-duckdb, sqlframe-postgres, ibis}, not {ibis}** — moving it under `_ibis/` would mean duplicating it for the other three, which is why it stayed in `_sql/expr.py`.

Your SQL is correct: I checked it against Python's `ROUND_HALF_EVEN` across 19 cases — ties and non-ties, positive and negative, `n` = -1, 0, 1, 2 — and it agrees on every one, and it is genuinely more compact than what's in the PR. The blocker is the same as (1): `power(10, n+1)` is `DOUBLE`, so the arithmetic promotes and the result degrades identically — `25::BIGINT` at `n=-1` gives `DOUBLE` 20.0, `2.5::DECIMAL(18,1)` gives `DOUBLE` 2.0.

So the real question is yours to settle: if `round` may return `Float64` on these backends, your whole line of suggestions is right and the shared implementation can go away. If it has to preserve dtype and exactness, the tie-only correction is what's left. Say which you want and I'll follow it.

`df527a3c` adds the two measured constraints as a comment above `SQLExpr.round`, a note on why the oddness test avoids `%` (sign behavior differs across these backends — see the existing `__mod__` override), and `test_round_keeps_large_integers_exact`, which was confirmed to fail when native `round_even` is patched in.

Local results: 14,520 passed / 302 skipped / 710 xfailed across duckdb, sqlframe, pandas (plain, nullable, pyarrow), pyarrow and polars eager+lazy; `pytest src --doctest-modules` 484 passed; mypy, pyright and pyrefly clean; `prek` clean on the changed files. ibis, pyspark and dask aren't installed in this environment, so CI covers those.

Written with Claude Code (Claude Opus 5) on @ethanstoner's behalf; the measurements above were run locally against the versions named.

FBruzzesi:
Hi @ethanstoner - first and foremost, I would really like to interact with a human as mentioned in our contributing guide:  [ai assisted contributions, point 5](https://github.com/narwhals-dev/narwhals/blob/main/CONTRIBUTING.md#ai-assisted-contributions), as I can prompt an agent myself if I need to.

Established that you are ok with that; I would try the SQL approach I was proposing which seems like a huge simplification (if there are no numerical issues), and let the dtype be for now. I would maybe cast to Int64, but I assume that can be problematic.

In general, opening feature requests or issues upstream is preferable:

* SQLFrame is a clear case, for which we asked many times and got feature shipped quite fast
* duckdb has a longer process, but the overflow might be an issue and the dtype might be a feature request?

PR #3960, opened 2026-09-18
fix: avoid `DIVIDE_BY_ZERO` in PySpark `corr` on constant columns

### LINKED ISSUE #3957: bug: `test_corr_constant_column` fails on `pyspark[connect]`: ANSI DIVIDE_BY_ZERO on a constant column
`test_corr_constant_column[pyspark[connect]]` fails on current main (bd1e4e0) with:

```
pyspark.errors.exceptions.connect.ArithmeticException: [DIVIDE_BY_ZERO] Division by zero.
Use `try_divide` to tolerate divisor being 0 and return NULL instead.
If necessary set "spark.sql.ansi.enabled" to "false" to bypass this error. SQLSTATE: 22012
```

first seen on my PR #3955's ci (job 105252246416), which merges current main. it's not caused by that PR's diff (str.slice and cast only): the failing test was added in bd1e4e0 itself, and main's own `PyTest PySpark constructor` run last passed on 735d819, one commit before bd1e4e0 landed. so any PR opened from current main should hit the same failure

the same test passes on the regular pyspark constructor, so the difference is presumably connect running under spark ansi mode, where dividing by the zero standard deviation of a constant column raises instead of returning null. i haven't verified the ansi hypothesis, only the failing run and the bisect-by-history

job log: https://github.com/narwhals-dev/narwhals/actions/runs/35236089804/job/105252246416

FBruzzesi:
# Description

Spark's `corr` only guards the empty/single-row cases, so a zero-variance column hits an unguarded divide and raises under ANSI mode (default since Spark 4). Compute it from guarded `covar_pop`/`stddev_pop` aggregates instead, matching the null result other backends return.

Same idea of what we already do in `cov`

- [ ] 💾 Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 🔧 Optimization
- [ ] 📝 Documentation
- [ ] ✅ Test
- [ ] 🐳 Other

## Related issues

- Closes #3957


FBruzzesi:
This is a totally new error:

```terminal
FAILED tests/translate/from_native_test.py::test_pyspark_connect_deps_2517[pyspark[connect]] - pyspark.errors.exceptions.base.PySparkRuntimeError: [JAVA_GATEWAY_EXITED] Java gateway process exited before sending its port number.
===== 1 failed, 3006 passed, 909 skipped, 157 xfailed in 134.20s (0:02:14) =====
```

See https://github.com/narwhals-dev/narwhals/actions/runs/35383813932/job/105725928247?pr=3960

---

Re-triggered and it run successfully, that's a new flaky one?! Never seen that failing before

MarcoGorelli (approved):
thanks @FBruzzesi !



just one thing to check - is this doing the same thing? just checking that pyspark's `corr` doesn't do any stability adjustments or anything like that (in which case, i think it would be fine to just let it raise and xfail the test)

FBruzzesi:
Thanks @MarcoGorelli - it's hard to say from the docs and jumping into the implementation it ends into java world. We have a `test_corr_numerical_stability` test which I am not sure is capturing all the edge cases though.

MarcoGorelli:
i asked claude to investigate and it came back with

>   What I found checking Spark's source
> 
>   This settles the question in the PR's favor. Spark's Corr, CovPopulation, and StddevPop/StddevSamp
>   aggregates all share the same single-pass, Welford-style moment-tracking base
>   (CentralMomentAgg/Covariance, implementing Meng 2015's "Simpler Online Updates for Arbitrary-Order
>   Central Moments"):
>   - covar_pop = ck / n
>   - stddev_pop = sqrt(m2 / n)
>   - native corr = ck / sqrt(xMk * yMk) — computed from the same ck/xMk/yMk accumulators, just without
>     the intermediate /n (it cancels algebraically: (ck/n) / (sqrt(xMk/n)·sqrt(yMk/n)) =
>     ck/sqrt(xMk·yMk)).
> 
>   So the PR's covar_pop(a,b) / (stddev_pop(a)*stddev_pop(b)) is mathematically identical to native
>   corr, built from the exact same numerically-stable single-pass accumulators — not a naive
>   two-pass/sum-of-squares formula that would reintroduce catastrophic cancellation. The only difference
>   is one extra division-then-multiply-back step, which adds negligible rounding versus what
>   test_corr_numerical_stability (offset-by-1e9 case) is checking, and that test already passes on all
>   three PySpark CI variants.
> 
>   Bottom line: the PR is correct and there's no numerical-stability regression — FBruzzesi's
>   uncertainty in that comment can be put to rest.

so i think it's fine to merge

PR #3966, opened 2026-09-20
fix: preserve all-null rows in `concat_str` for pyarrow backend

### LINKED ISSUE #3965: [Bug]: PyArrow `concat_str(ignore_nulls=True)` drops rows where every input is null
### Describe the bug

On the PyArrow backend, a row whose inputs are all null disappears from the output instead of producing an empty string, so the result is shorter than the frame. Selecting it alongside any other column then raises `ShapeError`.

The cause is upstream: `pc.binary_join_element_wise(..., null_handling="skip")` is not length-preserving. Its docstring says nulls "are skipped", but when every input in a row is null the kernel emits no element at all.

```python
import narwhals as nw

data = {
    "a": ["x", None, "z", None, "p"],
    "b": ["1", "2", None, None, "q"],
    "c": ["A", "B", "C", None, None],
}
expr = nw.concat_str(nw.col("a"), nw.col("b"), nw.col("c"), separator="-", ignore_nulls=True)

for backend in ("polars", "pyarrow"):
    frame = nw.from_dict(data, backend=backend)
    result = frame.select(x=expr)
    print(result, "\n", f"{result.shape=}", "\n")
```

```terminal
$ python t.py

┌──────────────────┐
|Narwhals DataFrame|
|------------------|
|  shape: (5, 1)   |
|  ┌───────┐       |
|  │ x     │       |
|  │ ---   │       |
|  │ str   │       |
|  ╞═══════╡       |
|  │ x-1-A │       |
|  │ 2-B   │       |
|  │ z-C   │       |
|  │       │       |
|  │ p-q   │       |
|  └───────┘       |
└──────────────────┘ 
 result.shape=(5, 1) 

┌────────────────────────────────┐
|       Narwhals DataFrame       |
|--------------------------------|
|pyarrow.Table                   |
|x: string                       |
|----                            |
|x: [["x-1-A","2-B","z-C","p-q"]]|
└────────────────────────────────┘ 
 result.shape=(4, 1) 
```

Without narwhals:

```python
import pyarrow as pa, pyarrow.compute as pc

a, b = pa.array(["x", None]), pa.array(["1", None])
print(pc.binary_join_element_wise(a, b, "-", null_handling="skip"))   # length 1: ['x-1']
print(pc.binary_join_element_wise(a, b, "-", null_handling="emit_null"))  # length 2: ['x-1', None]
```

### Expected results

The row is kept, with the empty string that Polars, DuckDB and the pandas-like backends produce:

```
['x-1-A', '2-B', 'z-C', '', 'p-q']
```

### Possible workaround?

A candidate workaround, giving exact Polars parity is to make all-null rows non-null before the join, so the kernel keeps them, and let "skip" produce the empty string:

```python
all_null = reduce(pc.and_, [pc.is_null(c) for c in cols])
patched = [pc.if_else(all_null, "", cols[0]), *cols[1:]]
pc.binary_join_element_wise(*patched, separator, null_handling="skip")
# ['x-1-A', '2-B', 'z-C', '', 'p-q'], length 5
```

It costs one extra pass over the inputs to build the mask. I hope we can come up with some better ideas 🤣

ddy314:
# Description

PyArrow's join kernel drops all-null rows with `null_handling="skip"`, shortening `concat_str(ignore_nulls=True)` results. Replace the first input with an empty string on those rows so the join preserves their position.

Use the shared cross-backend regression from #3967 and remove its PyArrow xfail. The existing string-type tests also pass with `pc.if_else(all_null, "", first)`.

## What type of PR is this?

- [x] Bug Fix
- [x] Test

## Related issues

Closes #3965.

## AI assistance

- [x] AI tools were used.

Codex prepared the code and replies and ran local validation. The account owner authorized submission and this review follow-up.

## Checklist

- [x] Code follows style guide (all prek hooks passed).
- [x] Shared cross-backend regression enabled.
- [x] Pyright, mypy, and pyrefly passed.
- [x] 484 doctests passed.
- [x] Initial local test screenshot attached below.

After merging main and addressing review: 86 concat_str tests passed across nine backend configurations. The full suite passed with 15,791 passed, 248 skipped, 962 xfailed and 100% source/test branch coverage. CI=true and GITHUB_ACTIONS=true were set locally to exercise the version-check branch.

Initial submission's local test screenshot (the current results are listed above):

Initial local pytest output


src/narwhals/_arrow/namespace.py
```diff
@@ -245,11 +245,19 @@ def func(df: ArrowDataFrame) -> list[ArrowSeries]:
...
                 *self.extract_native(*series), separator=separator
             )
+            columns = list(it)
+            if ignore_nulls:
+                # Arrow's "skip" kernel drops rows when every input is null.
+                and_: Incomplete = pc.and_
+                all_null = reduce(and_, (pc.is_null(column) for column in columns))
+                columns[0] = pc.if_else(
+                    all_null, pa.scalar("", type=columns[0].type), columns[0]
+                )
```
FBruzzesi: If I can suggest a slightly different approach for naming variables:

```suggestion
            rest = it
            if ignore_nulls:
                # `null_handling="skip"` emits no element at all for an all-null row,
                # so the output would be shorter than the input. Blanking the
                # first value on those rows keeps them and yields "" like Polars.
                and_: Incomplete = pc.and_
                all_null = reduce(and_, (pc.is_null(c) for c in (first, *rest)))
                first = pc.if_else(all_null, "", first)
```

and then line 260: `concat_str(first, *rest, separator_scalar, null_handling=null_handling),`

Regarding using `pa.scalar(...)` it might not be needed, but please double check

tests/expr_and_series/concat_str_test.py
```diff
@@ -72,6 +72,29 @@ def test_concat_str_with_lit(constructor: Constructor) -> None:
...
+    result = df.with_columns(
+        joined=nw.concat_str("a", "b", separator=separator, ignore_nulls=ignore_nulls)
+    )
+    expected = (
+        ["", f"x{separator}y", "", "", "z", ""]
+        if ignore_nulls
+        else [None, f"x{separator}y", None, None, None, None]
+    )
+    assert_equal_data(result.select("joined"), {"joined": expected})
+    assert result.to_native().schema.field("joined").type == dtype
```
FBruzzesi: This is a no-go test as we need to check that all the backends behave the same. I suggested one in https://github.com/narwhals-dev/narwhals/pull/3967#discussion_r4057530133 that assess both issues we currently have reported in `concat_str`. If you want to test with more parameters/arguments, that's welcome, but make sure we are not testing the same thing somewhere else. This issue is dedicated to all column being empty in one row, to me it shouldn't really matter which dtype and separator it's used.

Each PR will have to xfail the other backends (#3967 xfail pyarrow case, here xfail pandas-like and dask cases), and solve conflict if the other is merged first.

FBruzzesi (changes requested):
Hi @ddy314 thanks for your contribution and taking care of the issue. I left a couple of suggestions. The main one regarding the test suite itself

tests/expr_and_series/concat_str_test.py
FBruzzesi: Ok, #3967 has been merged onto main, so if you merge main into your branch you should get a failing test. There you should remove the xfail statement looking like the following:

```py
if ignore_nulls and "pyarrow_table" in str(constructor):  # pragma: no cover
    request.applymarker(
        pytest.mark.xfail(reason="pyarrow all-null concat_str row pending #3965")
    )
```

src/narwhals/_arrow/namespace.py
ddy314: Updated to first/rest in 3c2d24bc and removed pa.scalar. pc.if_else(all_null, "", first) passes the existing string and large_string type tests, along with the shared null-row regression.

tests/expr_and_series/concat_str_test.py
ddy314: Merged main, removed the PyArrow xfail from the shared test added in #3967, and deleted my backend-specific regression. All 86 concat_str cases pass across the nine backend configurations. The full suite passes with 15,791 passed, 248 skipped, 962 xfailed and 100% coverage; typing, prek and 484 doctests also pass.

FBruzzesi (approved):
Thanks a ton for bearing with me @ddy314

PR #3967, opened 2026-09-20
fix: drop trailing separator in `concat_str(..., ignore_nulls=True)` for pandas-like and Dask backends

### LINKED ISSUE #3962: [Bug]: `concat_str(ignore_nulls=True)` keeps the separator before a trailing null on pandas-like and Dask
### Describe the bug

With `ignore_nulls=True`, a null in the last position leaves a dangling separator at the end of the string on the pandas-like and Dask backends while Polars drops it.

Nulls in any other position are handled correctly, which is why this went unnoticed: each separator is blanked using the mask of the operand to its **left** (`null_mask[:-1]`), so an interior null drops the separator that follows it. A trailing null has no following separator, so the one before it survives.

```python
import narwhals as nw

data = {
    "a": ["x", None, "z", None, "p"],
    "b": ["1", "2", None, None, "q"],
    "c": ["A", "B", "C", None, None],
}
expr = nw.concat_str(nw.col("a"), nw.col("b"), nw.col("c"), separator="-", ignore_nulls=True)

for backend in ("polars", "pyarrow", "pandas"):
    frame = nw.from_dict(data, backend=backend)
    print(frame.select(x=expr).to_native(), end="\n\n")    
```

```terminal
shape: (5, 1)
┌───────┐
│ x     │
│ ---   │
│ str   │
╞═══════╡
│ x-1-A │
│ 2-B   │
│ z-C   │
│       │
│ p-q   │
└───────┘

pyarrow.Table
x: string
----
x: [["x-1-A","2-B","z-C","p-q"]]

       x
0  x-1-A
1    2-B
2    z-C
3       
4   p-q-
```

00200200:
## Description

In `concat_str(..., ignore_nulls=True)`, the separator between non-null values was previously derived by checking if the preceding column had a non-null value (`~null_mask[:-1]`). When the subsequent column was null but the preceding column was non-null, a dangling separator was still appended before the empty string, causing outputs like `'x-1-'` instead of `'x-1'`.

This updates both the pandas-like and Dask implementations so that separator \(i\) (placed before item \(i\)) is included if and only if there was at least one preceding non-null value (`has_prev`) AND current item \(i\) is non-null (`~null_mask[i]`).

- [x] 🐛 Bug Fix

## Related issues

- Closes #3962

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes


src/narwhals/_dask/namespace.py
```diff
@@ -265,10 +265,14 @@ def func(df: DaskLazyFrame) -> list[dx.Series]:
...
-                    for nm in null_mask[:-1]
-                )
+                has_prev = ~null_mask[0]
+                separators = []
+                for nm in null_mask[1:]:
+                    sep_mask = has_prev & (~nm)
+                    separators.append(
+                        sep_mask.map({True: separator, False: ""}, meta=str)
+                    )
+                    has_prev = has_prev | (~nm)
```
FBruzzesi: Couple of nitpicks here due to variable naming and inverting once:

```suggestion
                seen = ~null_mask[0]
                separators = []
                for nm in null_mask[1:]:
                    not_null = ~nm
                    separators.append(
                        (seen & not_null).map({True: separator, False: ""}, meta=str)
                    )
                    seen = seen | not_null
```

On top of this, I would also consider removing the reduce step and accumulate/fold in the for loop to avoid carrying what would be temporary information.

I was working on it locally, and my entire `else` block looked like the following:

```py
init_value, *values = [
    s.where(~nm, "") for s, nm in zip(series, null_mask, strict=True)
]

# A separator goes before a value only if that value and some
# earlier one are both non-null, so `seen` accumulates the latter.
seen = ~null_mask[0]
result = init_value
for nm, value in zip(null_mask[1:], values, strict=True):
    not_null = ~nm
    separators = (seen & not_null).map(
        {True: separator, False: ""}, meta=str
    )
    result = result + separators + value
    seen = seen | not_null
```

src/narwhals/_pandas_like/namespace.py
```diff
@@ -402,7 +402,11 @@ def func(df: PandasLikeDataFrame) -> list[PandasLikeSeries]:
                         dtype=init_value.native.dtype,
                     )
                 )
-                separators = (sep_array.zip_with(~nm, "") for nm in null_mask[:-1])
+                has_prev = ~null_mask[0]
+                separators = []
+                for nm in null_mask[1:]:
+                    separators.append(sep_array.zip_with(has_prev & (~nm), ""))
+                    has_prev = has_prev | (~nm)
```
FBruzzesi: Same nitpick as in dask, I would suggest renaming some variables, inverting series once and avoid storing temporary series:

```diff
diff --git a/src/narwhals/_pandas_like/namespace.py b/src/narwhals/_pandas_like/namespace.py
index 25a6e70a0..61851f364 100644
--- a/src/narwhals/_pandas_like/namespace.py
+++ b/src/narwhals/_pandas_like/namespace.py
@@ -402,12 +402,14 @@ class PandasLikeNamespace(
                         dtype=init_value.native.dtype,
                     )
                 )
-                separators = (sep_array.zip_with(~nm, "") for nm in null_mask[:-1])
-                result = reduce(
-                    operator.add,
-                    (s + v for s, v in zip(separators, values, strict=True)),
-                    init_value,
-                )
+                # A separator goes before a value only if that value and some
+                # earlier one are both non-null, so `seen` accumulates the latter.
+                seen = ~null_mask[0]
+                result = init_value
+                for nm, value in zip(null_mask[1:], values, strict=True):
+                    not_null = ~nm
+                    result = result + sep_array.zip_with(seen & not_null, "") + value
+                    seen = seen | not_null
 
             return [result]
```

tests/expr_and_series/concat_str_test.py
```diff
@@ -176,3 +176,18 @@ def test_concat_str_with_lit_and_nulls(
     assert_equal_data(
         result.select("r"), {"r": expected if ignore_nulls else expected_nulls}
     )
+
+
+def test_concat_str_ignore_nulls_trailing_null(constructor: Constructor) -> None:
+    # https://github.com/narwhals-dev/narwhals/issues/3962
+    if not any(name in str(constructor) for name in ("pandas", "modin", "cudf", "dask")):
+        pytest.skip("regression is specific to pandas-like and Dask backends")
```
FBruzzesi: This is a no-go for me. We need to test all the backend result in the same output.

I would suggest to add the following test, which will temporarily fail for pyarrow (until #3965 is addressed) thus it needs its own xfail

```py
@pytest.mark.parametrize(
    ("ignore_nulls", "expected"),
    [
        (True, ["x-1-A", "2-B", "z-C", "", "p-q"]),
        (False, ["x-1-A", None, None, None, None]),
    ],
)
def test_concat_str_nulls_in_every_position(
    constructor: Constructor, *, ignore_nulls: bool, expected: list[str | None]
) -> None:
    # A trailing null must not leave a dangling separator behind (#3962), and an
    # all-null row must stay in the output as an empty string (#3965).
    data = {
        "i": [0, 1, 2, 3, 4],
        "a": ["x", None, "z", None, "p"],
        "b": ["1", "2", None, None, "q"],
        "c": ["A", "B", "C", None, None],
    }
    expr = nw.concat_str("a", "b", "c", separator="-", ignore_nulls=ignore_nulls)
    result = df.with_columns(r=expr).sort("i").select("r")
    assert_equal_data(result, {"r": expected})
```

FBruzzesi (changes requested):
Hey @00200200 thanks for you contribution. I left suggestions for both the codebase and the test suite

src/narwhals/_dask/namespace.py
00200200: Updated in 64e389d: the Dask path now uses `seen`/`not_null`, folds directly in the loop, and avoids carrying a temporary separators collection.

src/narwhals/_pandas_like/namespace.py
00200200: Updated in 64e389d: the pandas-like path mirrors the Dask shape, inverts each mask once, and folds separators/values directly without storing temporary series.

tests/expr_and_series/concat_str_test.py
00200200: Updated in 64e389d: replaced the backend-specific skip with the parametrized all-backend test you suggested, including the native pyarrow xfail for the #3965 all-null-row case. Verified with pandas, pandas[pyarrow], dask, pyarrow, polars[eager], and polars[lazy].

00200200:
Thanks for the thorough review and clean suggestions @FBruzzesi! Applied the refactoring to both pandas-like and Dask namespaces (iterative accumulation without storing intermediate series), and updated the test suite to the parameterized regression test with xfail markers for `pyarrow_table` (pending #3965) and `polars < 1.0` (which also exhibited the trailing separator).

FBruzzesi (approved):
Amazing! Thanks @00200200 - sorry you had to face the polars v2.0rc2 breaking changes

Issue #3963, opened 2026-09-19
[Bug]: Dask raises when a scalar-like expression is an operand


FBruzzesi:
### Describe the bug

On the Dask backend an expression the plan treats as scalar-like (a literal, an aggregation) is materialized by `DaskExpr.broadcast()` as a `dask.dataframe.Scalar`:

```python
def broadcast(self) -> Self:
    def func(df: DaskLazyFrame) -> list[dx.Series]:
        return [result.loc[0][0] for result in self(df)]
```

A `Scalar` has none of the Series API. Any code path that calls a Series method on an expression result before putting it through `align_series_full_broadcast` therefore raises as soon as one of its operands is a literal or an aggregation.

`mean_horizontal` is the clearest instance: it calls `.fillna(0)` and `.isna()` on the raw results and aligns only afterwards, where `sum_horizontal`, `min_horizontal` and `max_horizontal` align first and work.

> [!IMPORTANT]
>  This is a shape of bug rather than a single call site, so it is worth auditing `_dask/namespace.py` and `_dask/expr.py` for the same ordering. `Expr.clip` is a second instance, reported separately because it also raises a backend-comparison question.

### Steps or code to reproduce the bug

```python
import narwhals as nw
import dask.dataframe as dd
import pandas as pd

df = nw.from_native(dd.from_pandas(pd.DataFrame({"a": [1.0, 2.0, None, 3.0]}), npartitions=1))
df.select(x=nw.mean_horizontal(nw.lit(1), nw.col("a")))
```

> AttributeError: 'Scalar' object has no attribute '_validate_axis'. Did you mean: '_validate_chunks'?

Every other backend agrees:

```
[1.0, 1.5, 1.0, 2.0]   # polars, pandas, pandas[pyarrow], pyarrow, duckdb
```

PR #3968, opened 2026-09-20
fix: align series before `fillna` and `isna` in dask `mean_horizontal`

00200200:
## Description

In `DaskNamespace.mean_horizontal`, operands can be scalar-like expressions (such as literals or aggregations) materialized as `dask.dataframe.Scalar`. Calling `.fillna(0)` and `.isna()` on those before `align_series_full_broadcast` raised an AttributeError because `Scalar` does not implement those Series methods.

This change aligns the expression results using `align_series_full_broadcast` first (turning any scalars into broadcasted Series), before calling `.fillna(0)` and `.isna()`. This matches the alignment pattern used by `min_horizontal` and `max_horizontal`, and enables removing the `xfail` marker on `test_meanh_with_literal` for Dask.

- [x] 🐛 Bug Fix

## Related issues

- Related #3963

## AI assistance

- [x] AI tools were used.

## Checklist

- [x] Tests added
- [x] Documented the changes


src/narwhals/_dask/namespace.py
```diff
@@ -200,10 +200,9 @@ def concat(
     def mean_horizontal(self, *exprs: DaskExpr) -> DaskExpr:
         def func(df: DaskLazyFrame) -> list[dx.Series]:
             expr_results = [s for _expr in exprs for s in _expr(df)]
-            series = align_series_full_broadcast(df, *(s.fillna(0) for s in expr_results))
-            non_na = align_series_full_broadcast(
-                df, *(1 - s.isna() for s in expr_results)
-            )
+            aligned = align_series_full_broadcast(df, *expr_results)
+            series = [s.fillna(0) for s in aligned]
+            non_na = [1 - s.isna() for s in aligned]
```
FBruzzesi: Fancy keeping these as generators? Shouldn't change much to be fair 😇


```suggestion
            series = (s.fillna(0) for s in aligned)
            non_na = (1 - s.isna() for s in aligned)
```

FBruzzesi (approved):
Really really happy to see such tests passing 🔥 Thanks @00200200 

Just FYI, I am going to change the PR description as I would rather keep the issue open to see if there is a more structural way of doing this rather than addressing it case by case

PR #3969, opened 2026-09-20
fix: make str.slice with negative offsets and negative length consistent

ptimizeroracle:
found with a differential matrix across backends while grouping fixes for #3955, split out per review:

negative offset combined with an explicit length returned empty strings on pandas, pyarrow and dask: stop = offset + length lands before the end-relative start, e.g. `slice(-3, 3)` on `"fdas"` gave `""` instead of `"das"`. all three now slice from offset then head-slice the result, matching polars

on the shared SQL backend, a negative offset on strings shorter than |offset| computed a non-positive substring start and DuckDB silently dropped a character: `slice(-3)` on `"ab"` returned `"a"`. the start is now clamped to 1 via `greatest`

a negative length was inconsistent everywhere: polars raised InvalidOperationError, the other backends silently returned empty or wrong strings. the stable `Expr.str.slice` and `Series.str.slice` now validate it up-front with `validate_str_non_negative` (same pattern as `str.zfill`'s width), so every backend gets the same error

tests: negative-offset cases including strings shorter than |offset|, the length=0 pin stays, negative-length ValueError for both the Expr and Series paths


ptimizeroracle:
test-sdist failed with 'Unexpected packages/ (workspace) directory inside the sdist'. the diff here touches only src and test files, the sdist config is unchanged from main, and the same check passed on my other pr (#3970) built from the same base an hour earlier, so it looks like a packaging/runner flake rather than something introduced here. would a re-run be ok?

FBruzzesi:
Hey @ptimizeroracle first and foremost, thanks for splitting the PR into two. May I ask why there are changes unrelated to the `str.slice` functionality? The changes are touching many string namespace methods, namely:`str.{starts_with, ends_with, replace, zfill}`, which seem completely unrelated to the `str.slice` issue. That's why you are seeing some many job failures, including the sdist one (from [the job logs](https://github.com/narwhals-dev/narwhals/actions/runs/35538990665/job/106153067698?pr=3969)):

```terminal
FAILED tests/expr_and_series/str/starts_with_ends_with_test.py::test_empty_pattern[pyarrow-ends_with] - AssertionError: Mismatch at index 0, key match: False != True
Expected: {'match': [True, None, True]}
Got: {'match': [False, None, True]}
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill_series[pyarrow-width_0] - pyarrow.lib.ArrowInvalid: Negative buffer resize: -18
FAILED tests/expr_and_series/str/zfill_test.py::test_str_zfill[pyarrow-width_0] - pyarrow.lib.ArrowInvalid: Negative buffer resize: -18
```

ptimizeroracle:
you were right, and thank you for the precise diagnosis. the cause: when splitting the pr i checked out whole files from the older combined branch, which silently reverted main's newer starts_with/ends_with, replace and zfill changes living in the same files. fixed in ac995aa: the backend files are main's versions again with only the slice methods changed, verified by diffing against main (only slice hunks remain). the zfill and starts_with failures disappear locally, full suite 15777 passed across nine constructors. the sdist job's packages/ leak should also follow, as it was building from the reverted tree

src/narwhals/_pandas_like/series_str.py
```diff
@@ -89,8 +89,13 @@ def contains(self, pattern: PandasLikeSeries, *, literal: bool) -> PandasLikeSer
...
     def slice(self, offset: int, length: int | None) -> PandasLikeSeries:
-        stop = offset + length if length is not None else None
-        return self.with_native(self.native.str.slice(start=offset, stop=stop))
+        # Slice from `offset` (which pandas handles natively, including negatives),
+        # then head-slice the result so that a negative offset combined with an
+        # explicit length doesn't compute a nonsense stop index.
+        result = self.native.str.slice(start=offset)
+        if length is not None:
+            result = result.str[:length]
+        return self.with_native(result)
```
FBruzzesi: The two-step fixes `offset < 0 <= offset + length`, but it loses the length reduction Polars applies when offset reaches before the start of the string. For example, `slice(-3, 2)` on "ab" now gives "ab" where Polars gives "a", and main got that case right. One native call with an adjusted stop covers both (needs `from narwhals._utils import str_slice_stop`):

```py
        stop = str_slice_stop(offset, length)
        return self.with_native(self.native.str.slice(start=offset, stop=stop))
```

src/narwhals/_arrow/series_str.py
```diff
@@ -76,10 +76,13 @@ def contains(self, pattern: ArrowSeries, *, literal: bool) -> ArrowSeries:
...
-        return self.with_native(
-            pc.utf8_slice_codeunits(self.native, start=offset, stop=stop)
-        )
+        # Same two-step as the pandas backend: `utf8_slice_codeunits` handles a
+        # negative start natively, but start + length can compute a stop index
+        # that lands before the (end-relative) start.
+        result = pc.utf8_slice_codeunits(self.native, start=offset)
+        if length is not None:
+            result = pc.utf8_slice_codeunits(result, start=0, stop=length)
+        return self.with_native(result)
```
FBruzzesi: Same as the pandas comment: `slice(-3, 2)` on "ab" now returns "ab" instead of "a". `utf8_slice_codeunits` follows Python slice semantics, so the shared stop works here too (needs `str_slice_stop` added to the existing `narwhals._utils` import):

```py
        stop = str_slice_stop(offset, length)
        return self.with_native(
            pc.utf8_slice_codeunits(self.native, start=offset, stop=stop)
        )
```

src/narwhals/_dask/expr_str.py
```diff
@@ -89,11 +89,17 @@ def _contains(expr: dx.Series, pattern: dx.Series) -> dx.Series:
...
+            # with an explicit length would otherwise compute a nonsense stop.
+            result = expr.str.slice(  # pyright: ignore[reportAttributeAccessIssue]
+                start=offset
             )
-        )
+            if length is not None:
+                result = result.str[:length]
+            return result
+
+        return self.compliant._with_callable(_slice)
```
FBruzzesi: Same issue as pandas. With the single-call version, the nested function and the `pyright: ignore` go away as well:

```py
        stop = str_slice_stop(offset, length)
        return self.compliant._with_callable(
            lambda expr: expr.str.slice(start=offset, stop=stop)
        )
```

src/narwhals/_sql/expr_str.py
FBruzzesi: Clamping the start to 1 is the right fix for `length=None`. With an explicit length, though, the count also needs to shrink by the overhang. Otherwise `slice(-3, 2)` on "ab" gives "ab" (Polars: "a") and `slice(-10, 3)` on "fdas" gives "fda" (Polars: "").

Handling `offset >= 0` first with an early return keeps the negative branch readable. `least`, `greatest` and `add` are already mapped for spark-like and ibis:

```py
            str_length = self._function("length", expr)
            count = self._lit(length) if length is not None else str_length
            if offset >= 0:
                return self._function("substr", expr, self._lit(offset + 1), count)

            # SQL engines disagree on a non-positive `substr` start, so clamp it to 1
            # and, like Polars, shorten `length` by however far `offset` overhangs.
            start_index = self._function("add", str_length, self._lit(offset))
            start = self._function(
                "greatest", self._lit(1), self._function("add", start_index, self._lit(1))
            )
            if length is not None:
                overhang = self._function("least", self._lit(0), start_index)
                count = self._function(
                    "greatest", self._lit(0), self._function("add", count, overhang)
                )
            return self._function("substr", expr, start, count)
```

tests/expr_and_series/str/slice_test.py
```diff
@@ -43,3 +45,31 @@ def test_str_slice_series(
...
+@pytest.mark.parametrize(
+    ("offset", "length", "expected"),
+    [
+        # Negative offset: slice starts |offset| from the end and runs forward.
+        (-3, 3, {"a": ["das", "ab", ""]}),
+        # Length overhanging the end of the string.
+        (-3, 10, {"a": ["das", "ab", ""]}),
+        # Negative offset with no length: to the end, also for strings shorter
+        # than |offset| (DuckDB used to return one character too few).
+        (-3, None, {"a": ["das", "ab", ""]}),
```
FBruzzesi: These cases all have `length >= abs(offset)`, so they never reach the case where Polars shortens length. The cases below fail on the current branch for every non-Polars backend (expected values come from native Polars). Could you add them? The per-case comments can probably go, since the values speak for themselves:

```suggestion
        (-3, None, {"a": ["das", "ab", ""]}),
        (-3, 3, {"a": ["das", "ab", ""]}),
        (-3, 10, {"a": ["das", "ab", ""]}),
        # When `offset` reaches before the start, Polars shortens `length` by the overhang.
        (-3, 2, {"a": ["da", "a", ""]}),
        (-3, 1, {"a": ["d", "", ""]}),
        (-5, 2, {"a": ["f", "", ""]}),
        (-10, 3, {"a": ["", "", ""]}),
```

tests/expr_and_series/str/slice_test.py
```diff
@@ -43,3 +45,31 @@ def test_str_slice_series(
...
+    ],
+)
+def test_str_slice_negative_offset_short_strings(
+    constructor: Constructor, offset: int, length: int | None, expected: Any
+) -> None:
```
FBruzzesi: `test_str_slice_negative_offset` might be a better name now that it covers more than short strings. The `expected: Any` annotations in this file should also be `dict[str, list[str]]`.

FBruzzesi (changes requested):
Hey @ptimizeroracle thanks for splitting this out and for the clean-up after the last round, the diff is now nicely scoped to `str.slice`.

Rejecting a negative `length` up-front is a good call. Polars already raises there, and it reuses the `zfill` validation pattern. The DuckDB non-positive `substr` start is a real bug too.

I checked each backend against native Polars and found one more thing to handle before merging. When a negative `offset` reaches before the start of the string, Polars doesn't just clamp the start to 0. It also **shortens `length` by the overhang**:

```py
import polars as pl
pl.Series(["ab"]).str.slice(-3, 2).to_list()  # starts 1 char before "a" -> keeps 2 - 1 = 1 char
['a']
pl.Series(["fdas"]).str.slice(-10, 3).to_list()  # overhang 6 >= length 3
['']
```

The two-step approach (slice from `offset`, then take the first `length` characters) and the SQL `greatest(1, ...)` clamp keep the full `length`, so these cases now differ from Polars:

| call | input | Polars | this PR (pandas, pyarrow, dask) | this PR (duckdb, sqlframe, ibis) | `main` (pandas, pyarrow, dask) |
| --- | --- | --- | --- | --- | --- |
| `slice(-3, 2)` | `"ab"` | `"a"` | `"ab"` | `"ab"` | `"a"` |
| `slice(-3, 1)` | `"ab"` | `""` | `"a"` | `"a"` | `""` |
| `slice(-5, 2)` | `"fdas"` | `"f"` | `"fd"` | `"fd"` | `"f"` |
| `slice(-10, 3)` | `"fdas"` | `""` | `"fda"` | `"fda"` | `""` |

The last column matters most. On `main`, pandas, pyarrow and dask already got these right, because a negative Python `stop` (`offset + length < 0`) clamps the same way Polars does. The original bug only shows up when `offset < 0 <= offset + length`, where `stop` wraps around past zero. So for those three backends the smallest fix is to keep the single native call and set `stop = None` in that one case:

```py
def str_slice_stop(offset: int, length: int | None) -> int | None:
    """Translate Polars' `str.slice(offset, length)` into a Python-slice `stop`."""
    if length is None or offset < 0 <= offset + length:
        return None
    return offset + length
```

Putting it in `narwhals/_utils.py` lets pandas-like, pyarrow and dask share it instead of each carrying its own copy of the logic. The inline suggestions below assume that helper exists.

On a grid of 10 offsets x 6 lengths over `["abcde", "ab", "", None, "héllo"]`, this plus the SQL change suggested below gives 0 mismatches against native Polars. The branch as it stands has 11 mismatches per backend. The current tests all use `length >= abs(offset)`, which is why CI didn't catch it; I've suggested extra cases on the test file.

ptimizeroracle:
applied all suggestions, thanks for the thorough review and for catching the overhang cases — the grid made it easy to verify.

- `str_slice_stop` added to `narwhals/_utils.py`, and pandas-like, pyarrow and dask now use a single native call with it (the nested function and the pyright ignore are gone)
- sql backends rewritten with the early return for `offset >= 0`, the `greatest(1, ...)` start clamp and the `least(0, start_index)` overhang shortening
- tests: overhang cases added, renamed to `test_str_slice_negative_offset`, annotations are `dict[str, list[str]]` and the per-case comments dropped

i also ran the parity grid locally, slightly wider than yours (offsets -10..10, lengths None,0,1,2,3,5,10, over ["abcde","ab","","héllo"]): 147 cases per backend, 0 mismatches against native polars on pandas, pandas[pyarrow], pyarrow, dask, duckdb and sqlframe

ptimizeroracle:
pulled your commit locally, all nine constructors green on the slice tests, ci fully green too. good catch on pre-0.20.17 polars having the same overhang bug — the reverse-slice workaround is a neat trick, and it means narwhals now returns consistent results even where the reference backend itself was wrong

FBruzzesi (approved):
Thanks for bearing with me @ptimizeroracle

PR #3972, opened 2026-09-21
fix: align series before clip, fill_null, and horizontal reductions in dask

### LINKED ISSUE #3963: [Bug]: Dask raises when a scalar-like expression is an operand

00200200:
Closes #3963.

### Description
In the Dask backend, when an expression result is scalar-like, it evaluates to a `dask.dataframe.Scalar`. Calling Series methods like `.clip()` or `.fillna()` directly raises errors (`TypeError: clip() got an unexpected keyword argument 'axis'` or `AttributeError: 'Scalar' object has no attribute '_validate_axis'`).

This PR audits and fixes the operand alignment order across:
- `DaskExpr.clip`, `DaskExpr.clip_lower`, and `DaskExpr.clip_upper`: evaluate single output bounds and align all operands with `align_series_full_broadcast(df, s, lower, upper)`.
- `DaskExpr.fill_null`: when `value is not None`, align `s` and `value` before calling `.fillna()`.
- `DaskNamespace.all_horizontal` and `DaskNamespace.any_horizontal`: align series with `align_series_full_broadcast` before applying `.fillna()` on non-bool types when `ignore_nulls=True`.

### Verification
- Added `test_clip_expr_with_aggregations` in `tests/expr_and_series/clip_test.py`.
- Added `all_agg_first` and `any_agg_first` cases in `tests/expr_and_series/horizontal_broadcasts_test.py`.
- Verified 100% tests passing across all constructors and linters (`ruff check`, `ruff format`).


00200200:
Pushed f3a442c2 for the random-versions failure: the Polars `<0.20.6` concat_str fallback now inserts separators only when a previous non-null value has been seen and the current value is non-null, matching the pandas-like/Dask behavior. Reproduced the failing case with `polars==0.20.5`; local targeted test passed (`4 passed`), and targeted ruff passed.

00200200:
Small follow-up: amended the same fix to 6e3e3894 to include the ruff-format change that `prek` requested. Local targeted test still passes (`4 passed`) and targeted ruff passes.

00200200:
Rebased onto main to resolve the conflict. The Polars `<0.20.6` `concat_str` separator fix I'd added here is already on main via #3981 (same logic), so I dropped that commit and this PR is back to only the dask scalar-operand alignment. In `horizontal_broadcasts_test.py` both sides had added cases to the DuckDB-aggregation set, so it now keeps all four. The affected test files pass locally (209 passed).

